Issue #4, opened 2024-08-06
Access Control (RBAC system)


miguelmtzinf:
Investigate and implement the RBAC system.

Try to simplify Aave V3 roles system, but keeping and enhancing funcionality.

- Work with 1 single ACL contract (same as V3 or [FEI](https://github.com/fei-protocol/fei-protocol-core/blob/develop/contracts/core/Permissions.sol))

CheyenneAtapour:
#### Research Findings:

* Most protocols use OpenZeppelin libraries; each having different granularities of access control
* Many protocols use their own implementation of access control, which is effectively a derivation or extension of an OpenZeppelin. For example, a contract may be delegated to to check whether a user can call a particular function on a particular target
* Regardless of the different approaches, the fundamental and core functionality of access control is checking whether an address can or cannot call a particular function signature on a particular contract address. In almost all instances, this information is stored in some kind of mapping

#### Implementation Approach:

* I will draft a couple implementations that keep all desired behaviors from the Aave V3 access control system, while trying to reduce the code size

miguelmtzinf:
reference: https://forum.openzeppelin.com/t/complex-conditional-logic-for-function-access-with-accessmanaged/44414

PR #16, opened 2024-08-14
Solmate Auth Access Control Implementation

CheyenneAtapour:
This PR showcases an approach to implementing Aave V4's access control requirements using Solmate's RolesAuthority library. We define one contract containing all the required roles from Aave V3, and may add more as needed during development of Aave V4. Note that we do not define role admins here, since it is not a feature natively supported, but it could be possible to do so. Aave V4 contracts are to inherit from this contract, and the default admin can set the capabilities of each role. Note that in this case, it is not required to have different modifiers, since the only required modifier is requiresAuth and we can have multiple roles able to call certain function signatures on certain contracts, removing the requirement for joint roles like onlyRiskOrPoolManager in the Open Zeppelin implementation.

Closes https://github.com/aave/aave-v4/issues/4

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.8.0;
 
 import {SafeERC20} from '../dependencies/openzeppelin/SafeERC20.sol';
 import {IERC20} from '../dependencies/openzeppelin/IERC20.sol';
-import {AccessManaged} from './AccessManaged.sol';
+import {Auth, Authority} from '../dependencies/solmate/Auth.sol';
 import {WadRayMath} from './WadRayMath.sol';
 import {SharesMath} from './SharesMath.sol';
 import {MathUtils} from './MathUtils.sol';
@@ -11,7 +11,7 @@ import {IBorrowModule} from './IBorrowModule.sol';
 
 import 'forge-std/console2.sol';
 
-contract LiquidityHub is AccessManaged {
+contract LiquidityHub is Auth {
   using SafeERC20 for IERC20;
   using WadRayMath for uint256;
   using SharesMath for uint256;
@@ -62,7 +62,7 @@ contract LiquidityHub is AccessManaged {
   // asset id => user address => user data
   mapping(uint256 => mapping(address => UserConfig)) public users;
 
-  constructor(address admin) AccessManaged(admin) {}
+  constructor(address admin, address authority) Auth(admin, Authority(authority)) {}
 
   function getReserve(uint256 assetId) external view returns (Reserve memory) {
     return reserves[assetId];
--- a/tests/BaseTest.t.sol
+++ b/tests/BaseTest.t.sol
@@ -10,6 +10,7 @@ import 'src/contracts/IBorrowModule.sol';
 import 'src/contracts/WadRayMath.sol';
 import 'src/contracts/SharesMath.sol';
 import 'src/contracts/MathUtils.sol';
+import 'src/contracts/AccessManaged.sol';
 import 'src/dependencies/openzeppelin/IERC20.sol';
 import './mocks/ERC20Mock.sol';
 import './Utils.t.sol';
@@ -46,6 +47,8 @@ abstract contract BaseTest is Test, Events {
   IERC20 internal dai;
   IERC20 internal usdt;
 
+  AccessManaged internal accessManager;
+
   LiquidityHub hub;
   BorrowModule bm;
 
@@ -55,7 +58,8 @@ abstract contract BaseTest is Test, Events {
 
   function setUp() public virtual {
     vm.startPrank(ADMIN);
-    hub = new LiquidityHub(ADMIN);
+    accessManager = new AccessManaged(ADMIN);
+    hub = new LiquidityHub(ADMIN, address(accessManager));
     bm = new BorrowModule();
     usdc = new ERC20Mock();
     dai = new ERC20Mock();
--- a/tests/invariant/LiquidityHubHandler.t.sol
+++ b/tests/invariant/LiquidityHubHandler.t.sol
@@ -5,6 +5,7 @@ import 'forge-std/Test.sol';
 
 import 'src/contracts/LiquidityHub.sol';
 import 'src/contracts/BorrowModule.sol';
+import 'src/contracts/AccessManaged.sol';
 import 'src/dependencies/openzeppelin/IERC20.sol';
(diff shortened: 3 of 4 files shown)


CheyenneAtapour:
Note: While developing Aave V4, we may simply add new required roles to AccessManagerContract.sol. Any function requiring auth may simply use the `requiresAuth` modifier. We only need to ensure to grant the corresponding roles to the appropriate addresses during contract deployment. This can be done (and adjusted) at any point after the contracts are deployed.

src/contracts/AccessManagerContract.sol
```diff
@@ -0,0 +1,24 @@
...
+  uint8 public constant FLASH_BORROWER = 5;
+  uint8 public constant BRIDGE = 6;
+  uint8 public constant ASSET_LISTER = 7;
+  uint8 public constant POOL_CONFIGURATOR = 8;
+  uint8 public constant ISOLATED_COLLATERAL_SUPPLIER = 9;
+  uint8 public constant FUNDS_MANAGER = 10;
+  uint8 public constant MARKET_OWNER = 11;
+  uint8 public constant EMISSION_MANAGER = 12;
+  uint8 public constant REWARDS_MANAGER = 13;
+  uint8 public constant INCENTIVES_CONTROLLER = 14;
```
miguelmtzinf: Moving this to an external library to be used as roles dictionary, would provide complete flexibility.
Otherwise, if there is a new role in the system, we would need to either recreate this `AccessManagerContract` with the new role or just keep it as it is (so this contract won't have all roles).

CheyenneAtapour: Yeah, great point! - Done now

remappings.txt
miguelmtzinf, line 1: remove these dependencies and copy them to the dependencies folder instead please

src/contracts/AccessManagerContract.sol
```diff
@@ -0,0 +1,16 @@
...
+import { RolesAuthority } from '@solmate/auth/authorities/RolesAuthority.sol';
+
+/**
+ * @title AccessManagerContract
+ * @author Aave Labs
+ * @notice Handles access management for Aave V4
+ * @dev Contracts requiring access control should inherit from this contract
+ * @dev Restricted functions should use the `requiresAuth` modifier
+ */
+contract AccessManagerContract is RolesAuthority {
```
miguelmtzinf: Let's rename! what about `AccessManaged`? open to suggestions :D

src/contracts/AccessManagerContract.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { Authority, Auth } from '@solmate/auth/Auth.sol';
```
miguelmtzinf:
```suggestion
import { Authority } from '@solmate/auth/Auth.sol';
```

src/contracts/AccessManagerContract.sol
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * @title AccessManagerContract
+ * @author Aave Labs
+ * @notice Handles access management for Aave V4
+ * @dev Contracts requiring access control should inherit from this contract
+ * @dev Restricted functions should use the `requiresAuth` modifier
+ */
+contract AccessManagerContract is RolesAuthority {
+  constructor() RolesAuthority(msg.sender, Authority(address(this))) {}
```
miguelmtzinf: we usually do not use `msg.sender` as owner, because we may end up using factories for deployments.
let's create 2 address params as inputs: `owner` and `authority`.

src/contracts/AccessManagerContract.sol
miguelmtzinf: I d inherit from `Auth` only, and have a separate contract as authority so we can swap this contract in the future

miguelmtzinf:
I would like you to go through the alternatives:
- Auth from Solmate: https://github.com/transmissions11/solmate/blob/main/src/auth/Auth.sol
- Auth from OZ: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/manager/AccessManaged.sol 
- Authority from Solmate: https://github.com/transmissions11/solmate/blob/main/src/auth/authorities/RolesAuthority.sol
- Authority from OZ: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/manager/AccessManager.sol

In my opinion:
- Solmate Auth is very lightweight. However, didn't find audit reports of it.
- OZ Auth feels more consolidated and clean to me. However, we do not need the schedule/delay thing (although I think we can just omit that).
- Solmate Authority looks good, but we do not really need the `isPublicCapabilty` and adds 1 sload per call... That feature allows to change from permissioned to permissionless right? If the assumption is that contracts will be upgradeable, we do not need it at all. Otherwise, good to keep it.
- OZ Authority seems overkill to me, too much stuff there.

src/contracts/AccessManagerContract.sol
CheyenneAtapour: Too similar to OZ I think. I was thinking `PermissionsManager` ?

remappings.txt
CheyenneAtapour: Done

src/contracts/AccessManagerContract.sol
CheyenneAtapour: Done

src/contracts/AccessManagerContract.sol
CheyenneAtapour: I originally was trying to tackle that kind of setup in a different PR here: https://github.com/aave/aave-v4/pull/456 but can combine and just proceed here

CheyenneAtapour: But please note that this implementation will also allow us to swap out this contract in the future, because there is a `setAuthority` function which allows us to change the authority contract. I just added a test for this. However, I do think you're right it would be cleaner as a separate contract

CheyenneAtapour: I did `AccessManaged` after all

CheyenneAtapour: This commit has it with the `LiquidityHub` being it's own authority, just for reference: [b754444](https://github.com/aave/aave-v4/commit/b754444ac3ece3fbd5ad824ce0d53cd04e19ce7f)

Will proceed with making a separate authority contract now

CheyenneAtapour:
Proceeding with open zeppelin implementation for now

PR #17, opened 2024-08-14
OpenZeppelin Access Control Implementation

CheyenneAtapour:
This PR showcases an approach to implementing Aave V4's access control requirements using OpenZeppelin's Access Control contracts. Specifically, the approach is to have one contract which defines all the roles that we will need for Aave V4, taking into consideration all the roles which were required in Aave V3, and have each contract inherit from this contract, so that we do not have duplicated role declarations or modifiers.

Closes https://github.com/aave/aave-v4/issues/4

### DIFF
--- a/src/contracts/AccessManagerContract.sol
+++ b/src/contracts/AccessManagerContract.sol
@@ -16,12 +16,7 @@ import {AccessControl} from '@openzeppelin/contracts/access/AccessControl.sol';
  */
 contract AccessManagerContract is AccessControl {
   bytes32 public constant RESERVE_CONTROLLER = keccak256('RESERVE_CONTROLLER');
-  bytes32 public constant INTEREST_RATE_CONTROLLER = keccak256('INTEREST_RATE_CONTROLLER');
-  bytes32 public constant RISK_CONTROLLER = keccak256('RISK_CONTROLLER');
-
-  bytes32 public constant INTEREST_RATE_CONTROLLER_ADMIN =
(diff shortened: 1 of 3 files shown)


CheyenneAtapour:
Note: As we develop V4, we can add more roles and modifiers as needed to the `AccessManager.sol` contract. Then, we can simply append the appropriate modifier to any function that requires authentication

src/contracts/AccessManagerContract.sol
```diff
@@ -0,0 +1,173 @@
...
+  bytes32 public constant EMISSION_MANAGER_ADMIN = keccak256('EMISSION_MANAGER_ADMIN');
+  bytes32 public constant REWARDS_MANAGER = keccak256('REWARDS_MANAGER');
+  bytes32 public constant REWARDS_MANAGER_ADMIN = keccak256('REWARDS_MANAGER_ADMIN');
+  bytes32 public constant INCENTIVES_CONTROLLER = keccak256('INCENTIVES_CONTROLLER');
+  bytes32 public constant INCENTIVES_CONTROLLER_ADMIN = keccak256('INCENTIVES_CONTROLLER_ADMIN');
+
+  /**
+   * @dev Only pool manager can call functions marked by this modifier.
+   */
+  modifier onlyPoolManager() {
```
miguelmtzinf, line 51: having a modifier per role is clearer but it won't provide the flexibility/simplicity we are aiming for imho

miguelmtzinf:
Closing this one in favor of https://github.com/aave/aave-v4/pull/16

PR #18, opened 2024-08-27
Add Reserve Liquidity Premium to supply side

miguelmtzinf:
Notes: https://www.notion.so/aave/Liquidity-Premium-a70261fc71a14d77a807859a653db5eb


miguelmtzinf:
Based on our analysis, we are following an static approach where it's possible to "rebalance" the risk premium of the user under certain conditions. Created this issue https://github.com/aave/aave-v4/issues/20 for that.

Issue #21, opened 2024-09-23
Basic Liquidation Engine


miguelmtzinf:
We need to implement the liquidation mechanism. We want to have the V3 implementation to liquidate undercollateralized positions to start out with.
In addition, instead of allowing the liquidator to liquidate up to 50% if HF goes below certain threshold, we want allow the liquidator to liquidate enough assets so the HF goes back to 1 (or slightly higher).

yan-man:
in this PR we will tackle the basic liquidation mechanism similar to v3. In https://github.com/aave/aave-v4/issues/37 will be enhancement to liquidate enough to restore HF to 1

yan-man:
liquidation validation:

Asset Active (general on/off switch, enforce 0 suppliers when turning off)
- no liquidations possible

Asset Paused (if paused, no more actions, meant to be temporary)
- no liquidations possible

Asset Freeze (if frozen, no more supply or borrow)
- liq possible 

I see v3 only checks active/paused, will do the same: 
` require(vars.collateralReserveActive && vars.principalReserveActive, Errors.RESERVE_INACTIVE);
    require(!vars.collateralReservePaused && !vars.principalReservePaused, Errors.RESERVE_PAUSED);
`

PR #23, opened 2024-09-30
Add basic liquidation engine

yan-man:
- Basic initial version of liquidation engine, based on AaveV3
- Calculate collateral liquidation based on debt
- still need to resolve HF calculations later based on borrow module implementation

incorporates changes from https://github.com/aave/aave-v4/pull/24, make sure to merge after that PR

closes https://github.com/aave/aave-v4/issues/21


yan-man:
will need to be updated if/when borrow module is updated

yan-man:
see https://github.com/aave/aave-v4/pull/54

Issue #27, opened 2024-10-14
Add Abstraction for Liquidity Hub


yan-man:
Remove dependencies for liquidity hub on borrow module. Liquidity hub allows different borrow modules to draw liquidity. Borrowing happens through BM rather than LH directly. BM should keep track of its own interest rates.

PR #29, opened 2024-10-15
Liquidity Hub Abstraction

yan-man:
- abstract liquidity hub from borrow module 
- tests involving most basic bm mock - as a credit line with fixed interest rate
- resolve calculations for accrual of interest

closes https://github.com/aave/aave-v4/issues/27

### DIFF
--- a/tests/BaseTest.t.sol
+++ b/tests/BaseTest.t.sol
@@ -12,10 +12,10 @@ import 'src/contracts/SharesMath.sol';
 import 'src/contracts/MathUtils.sol';
 import 'src/dependencies/openzeppelin/IERC20.sol';
 import 'src/interfaces/IBorrowModule.sol';
+import 'src/contracts/DefaultReserveInterestRateStrategy.sol';
 import './mocks/MockERC20.sol';
 import './mocks/MockPriceOracle.sol';
 import './mocks/MockBorrowModuleCreditLine.sol';
-import '../src/contracts/DefaultReserveInterestRateStrategy.sol';
 
 import './Utils.t.sol';
 


src/contracts/LiquidityHub.sol
```diff
@@ -230,25 +236,13 @@ contract LiquidityHub {
...
-    // reset allowance
-    IERC20(reservesList[assetId]).forceApprove(reserve.config.borrowModule, 0);
+    // directly transfer funds to bm so that allowance doesn't need to be reset
+    IERC20(reservesList[assetId]).safeTransfer(reserve.config.borrowModule, amount);
 
     emit Draw(assetId, reserve.config.borrowModule, amount);
   }
 
-  function repay(uint256 assetId, uint256 amount, address onBehalfOf) external {}
+  function restore(uint256 assetId, uint256 amount, address onBehalfOf) external {}
```
miguelmtzinf: funny enough there could be a usecase for `onBehalfOf`, but I think we can remove for now, and leave it as comment

src/contracts/LiquidityHub.sol
```diff
@@ -272,7 +266,7 @@ contract LiquidityHub {
     require(reserve.totalAssets >= amount, 'NOT_AVAILABLE_LIQUIDITY');
   }
 
-  function _validateBorrow(Reserve storage reserve, uint256 amount) internal view {
+  function _validateDrawLiquidity(Reserve storage reserve, uint256 amount) internal view {
```
miguelmtzinf, line 270: there must be a `drawCap` per BM (no cap if 0)

src/interfaces/IBorrowModule.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IBorrowModule {
+  event Borrowed(uint256 assetId, address user, uint256 amount);
+  event Repaid(uint256 assetId, address user, uint256 amount);
+
+  function calculateInterestRates() external pure returns (uint256);
```
miguelmtzinf: let's add a simple dev comment specifying we work with bps

tests/mocks/MockBorrowModuleCreditLine.sol
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  function borrow(uint256 assetId, uint256 amount) external {
+    ILiquidityHub(liquidityHub).draw(assetId, amount);
+  }
+
+  function repay(uint256 assetId, uint256 amount, address onBehalfOf) external {}
+
+  function getInterestRate() public pure returns (uint256) {
+    return 0.05e27; // 5% IR
```
miguelmtzinf: we work with bps

tests/mocks/MockBorrowModuleCreditLine.sol
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  function calculateInterestRates() public pure returns (uint256) {
+    // borrowRate
+    return 0;
```
miguelmtzinf: this can be the same as `getInterestRate`in this case, since it's fixed.

src/contracts/LiquidityHub.sol
yan-man: will resolve in https://github.com/aave/aave-v4/pull/30 and be merged into this one

src/contracts/LiquidityHub.sol
yan-man: right now it is based on a single draw cap per reserve, which is fine if a single bm is allowed per reserve. But logic needs to be updated if we start to allow multiple bms per reserve

src/interfaces/IBorrowModule.sol
yan-man: where do you think we should have the conversion be to ray for calculations involving the IR? (because MathUtils.calculateLinearInterest uses ray) 

edit: decided on 22.10 call to leave conversion in the bm

src/contracts/BorrowModule.sol
```diff
@@ -79,16 +88,43 @@ contract BorrowModule is IBorrowModule {
...
         )
       );
   }
 
+  function borrow(uint256 assetId, uint256 amount) external {
+    Reserve storage r = reserves[assetId];
+    _validateBorrow(r, amount);
+    // TODO HF check
+
+    ILiquidityHub(liquidityHub).draw(assetId, amount);
```
miguelmtzinf, line 105: the `assetId` in BM may be not the same as the one in LH, but good for now.

src/interfaces/IDefaultInterestRateStrategy.sol
```diff
@@ -15,8 +15,8 @@ interface IDefaultInterestRateStrategy is IReserveInterestRateStrategy {
    * @param assetId address of the reserve that has new interest rate data set
    * @param optimalUsageRatio The optimal usage ratio, in bps
    * @param baseVariableBorrowRate The base variable borrow rate, in bps
-   * @param variableRateSlope1 The slope of the variable interest curve, before hitting the optimal ratio, in bps
-   * @param variableRateSlope2 The slope of the variable interest curve, after hitting the optimal ratio, in bps
+   * @param variableRateSlope1 The slope of the variable interest curve, before hitting the optimal ratio, in bps interest rate per bps usage ratio
```
miguelmtzinf, line 18: what do you mean here?

yan-man: the units of slope are not bps - they're a change in y divided by change in x. So that is change in base variable borrow rate, in bps (the y) divided by change in usage ratio, in bps (the x).

src/contracts/BorrowModule.sol
yan-man: Ah figured we would use the same assetId across the protocol. Ok will keep this in mind for later

src/interfaces/IDefaultInterestRateStrategy.sol
miguelmtzinf: nope, they are basically rates, in bps

src/contracts/BorrowModule.sol
CheyenneAtapour: I also think it'd be easier if we did just use LH as source of truth for asset IDs

src/contracts/BorrowModule.sol
```diff
@@ -79,16 +88,43 @@ contract BorrowModule is IBorrowModule {
...
+  // TODO: onBehalfOf
+  function repay(uint256 assetId, uint256 amount) external {
+    ILiquidityHub(liquidityHub).restore(assetId, amount);
+
+    emit Repaid(assetId, msg.sender, amount);
+  }
+
+  function getInterestRate(uint256 assetId) public view returns (uint256) {
+    // read from state, convert to ray
+    return reserves[assetId].borrowRate * 1e23;
```
CheyenneAtapour, line 125: There's one uniform IR strategy right, so should we also just fetch that from LH?

tests/BaseTest.t.sol
```diff
@@ -7,13 +7,15 @@ import 'forge-std/console2.sol';
...
-import 'src/contracts/IBorrowModule.sol';
 import 'src/contracts/WadRayMath.sol';
 import 'src/contracts/SharesMath.sol';
 import 'src/contracts/MathUtils.sol';
 import 'src/dependencies/openzeppelin/IERC20.sol';
+import 'src/interfaces/IBorrowModule.sol';
 import './mocks/MockERC20.sol';
 import './mocks/MockPriceOracle.sol';
+import './mocks/MockBorrowModuleCreditLine.sol';
+import '../src/contracts/DefaultReserveInterestRateStrategy.sol';
```
CheyenneAtapour: The smallest nit: can just do 'src/...' like the others

tests/unit/BorrowModuleCreditLineTest.t.sol
```diff
@@ -0,0 +1,417 @@
...
+      .calculateLinearInterest(
+        IBorrowModule(address(bmcl)).getInterestRate(daiId),
+        uint40(daiData1.lastUpdateTimestamp)
+      )
+      .rayMul(daiData1.totalDrawn);
+
+    // User1 draw 25% of dai reserve liquidity for borrow module
+    vm.prank(USER1);
+    IBorrowModule(address(bmcl)).borrow(daiId, drawnAmounts[1]);
+    // User1 draw 20% of dai reserve liquidity for borrow module
```
CheyenneAtapour: User2 right?

src/contracts/BorrowModule.sol
yan-man: hm my understanding was that each BM could have a different IR strategy configured, and I just stored it into state here so that it wouldn't need to read from an external contract on each invocation, but the state gets updated on each action. Perhaps this all changes with the latest design decisions cc @miguelmtzinf

src/interfaces/IDefaultInterestRateStrategy.sol
yan-man: but it is correct that we have an x-axis of usage ratio in BPS and y-axis of borrow rate BPS?

miguelmtzinf:
Merging this branch, so we can keep building on https://github.com/aave/aave-v4/pull/42

PR #30, opened 2024-10-18
Liquidity Hub / Borrow Module Interest Rate Integration

yan-man:
- incorporate IR (interest rate) strategy into borrow module
- resolve calculations involving IR
- apply this to basic bm credit line with a fixed IR

### DIFF
--- a/src/contracts/BorrowModule.sol
+++ b/src/contracts/BorrowModule.sol
@@ -100,6 +100,7 @@ contract BorrowModule is IBorrowModule {
   function borrow(uint256 assetId, uint256 amount) external {
     Reserve storage r = reserves[assetId];
     _validateBorrow(r, amount);
+    // TODO HF check
 
     ILiquidityHub(liquidityHub).draw(assetId, amount);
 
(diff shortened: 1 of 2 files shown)


src/contracts/BorrowModule.sol
```diff
@@ -174,12 +143,14 @@ contract BorrowModule is IBorrowModule {
...
   }
 
-  function getInterestRate() public pure returns (uint256) {
+  function getInterestRate(uint256 assetId) public pure returns (uint256) {
     // TODO: call from interest rate strategy
     return 0;
   }
 
-  function calculateInterestRates() public pure returns (uint256) {
+  function calculateInterestRates(
```
miguelmtzinf: this is actually part of the IR contract

src/contracts/BorrowModule.sol
```diff
@@ -174,12 +143,14 @@ contract BorrowModule is IBorrowModule {
     });
   }
 
-  function getInterestRate() public pure returns (uint256) {
+  function getInterestRate(uint256 assetId) public pure returns (uint256) {
     // TODO: call from interest rate strategy
```
miguelmtzinf: nope, just read from storage

src/contracts/BorrowModule.sol
yan-man: essentially just left this to convert BIPs (returned from IR contract) to ray for calcs. Should IR contract be returning rate in ray?

yan-man: resolved on call - do conversion for now in the borrow module from BIPS to ray

PR #42, opened 2024-11-17
Rebuild LH and BM based on last design

### LINKED ISSUE #31: Finalize first implementation of BM
Implement remaining functions in BM
Every action should follow the same approach:
1. Validate action
2. Refresh risk premium of the user and aggregated risk premium
3. Execute corresponding action in LH by passing the agg weighted risk premium
4. Accounting (using resulting shares from LH)
5. Event

All information about the IR and exchange rate between shares and assets is controlled by LH.
This info is used in the steps above (1 and 2 mostly), to convert shares to assets and viceversa.
The LH should provide 2 getter functions, to provide conversions and up-to-date balances.
- getter for conversions: `convertAssetsToShares(amount, rounding)` and `convertSharesToAssets(amount, rounding)`
- getter for up-to-date balance: same as the previous, but we compute the interest accrual right before (totalAssets will increase a bit)

We can skip riskPremium logic stuff for now.

Let's do these 2 things here too:
- Rename Borrow Module (BM) to Spoke across the code.
- Fix tests (specially unit tests, we can comment or skip invariant ones for now)

miguelmtzinf:
Closes #31

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -17,17 +17,12 @@ contract LiquidityHub is ILiquidityHub {
   using WadRayMath for uint256;
   using SharesMath for uint256;
 
+  // TODO: update name of this struct to reference the asset/reserve?
   struct Spoke {
     uint256 totalShares;
     uint256 drawnShares;
     // TODO: lastUpdateTimestamp?
-    SpokeConfig config;
-  }
-
-  // TODO: borrow cap per spoke
-  struct SpokeConfig {
-    uint256 drawCap; // asset denominated
-    uint256 supplyCap; // asset denominated
+    DataTypes.SpokeConfig config;
   }
 
   struct Asset {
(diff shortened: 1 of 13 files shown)


src/contracts/LiquidityHub.sol
```diff
@@ -3,337 +3,358 @@ pragma solidity ^0.8.0;
...
     uint256 supplyCap;
-    uint256 borrowCap;
-    uint256 liquidityPremium; // in bps, so 10000 is 100.00%
-    // uint256 eModeCategory; // TODO eMode
-    // uint256 debtCeiling; // TODO isolation mode
-  }
-
-  struct UserConfig {
-    uint256 shares;
+    address irStrategy;
```
yan-man: why not also include borrowCap here @CheyenneAtapour ?

src/interfaces/ILiquidityHub.sol
```diff
@@ -7,6 +7,37 @@ pragma solidity ^0.8.0;
  * @notice Basic interface for LiquidityHub
  */
 interface ILiquidityHub {
-  function draw(uint256 assetId, uint256 amount) external;
-  function restore(uint256 assetId, uint256 amount) external;
+  function draw(
+    uint256 assetId,
+    address onBehalfOf,
```
CheyenneAtapour: @yan-man Just a quick one: This needs to be `to` instead of `onBehalfOf` (it's correct already on LiquidityHub`)

PR #43, opened 2024-11-18
Resolve conflicts from Main branch

yan-man:
- resolve conflicts, make sure contracts compile properly
- resolve tests later in a different PR

### DIFF
--- a/src/contracts/BorrowModule.sol
+++ b/src/contracts/BorrowModule.sol
@@ -111,8 +111,8 @@ contract BorrowModule is IBorrowModule {
     emit Withdraw(assetId, msg.sender, amount);
   }
 
-  // TODO: On behalf of and referral code
   function borrow(uint256 assetId, uint256 amount) external {
+    // TODO: On behalf of and referral code
     Reserve storage r = reserves[assetId];
     _validateBorrow(r, amount);
     // TODO HF check


src/contracts/BorrowModule.sol
```diff
@@ -14,39 +14,51 @@ contract BorrowModule is IBorrowModule {
...
   using SafeERC20 for IERC20;
 
+  // debt balances, fetches indexes from liquidity layer
+
+  // keep collateral configuration
+  // By using BorrowModule, LPs can choose which collaterals are used to borrow their assets
+
+  // keep hooks to be executed by LiquidityHub when there is supply/withdraw actions
+
+  // fetch liquidity from liquidityHub
```
CheyenneAtapour: We can remove these comments

src/contracts/BorrowModule.sol
```diff
@@ -14,39 +14,51 @@ contract BorrowModule is IBorrowModule {
...
   address public liquidityHub;
+  address public interestRateStrategy;
```
CheyenneAtapour: Borrow module does not need interest rate strategy here. Will be fetched from LH

src/contracts/BorrowModule.sol
```diff
@@ -14,39 +14,51 @@ contract BorrowModule is IBorrowModule {
...
   struct Reserve {
     uint256 id;
     address asset;
-    // uint256 totalDebt;
-    // uint256 lastUpdateIndex;
-    // uint256 lastUpdateTimestamp;
+    uint256 totalDebt;
+    uint256 lastUpdateIndex;
+    uint256 lastUpdateTimestamp;
+    uint256 borrowRate;
```
CheyenneAtapour: Please revert this change. We are keeping these commented out, and borrowRate is coming from LH

src/contracts/BorrowModule.sol
```diff
@@ -14,39 +14,51 @@ contract BorrowModule is IBorrowModule {
...
     ReserveConfig config;
   }
 
   struct ReserveConfig {
     uint256 lt;
     uint256 lb; // TODO: liquidationProtocolFee
+    uint256 rf;
```
CheyenneAtapour: This will be dealt with on LH side, please revert

src/contracts/BorrowModule.sol
```diff
@@ -14,39 +14,51 @@ contract BorrowModule is IBorrowModule {
...
 
   // reserve id => user address => user data
   mapping(uint256 => mapping(address => UserConfig)) public users;
   // reserve id => reserveData
   mapping(uint256 => Reserve) public reserves;
 
-  constructor(address liquidityHubAddress) {
+  constructor(address liquidityHubAddress, address interestRateStrategyAddress) {
     liquidityHub = liquidityHubAddress;
+    interestRateStrategy = interestRateStrategyAddress;
```
CheyenneAtapour: Don't need interestRateStrategy here, will fetch from LH

src/contracts/BorrowModule.sol
```diff
@@ -60,9 +72,13 @@ contract BorrowModule is IBorrowModule {
   }
 
   function getUserDebt(uint256 assetId, address user) external view returns (uint256) {
+    // TODO: Instead use a getter from liquidity hub to get up-to-date user debt (with accrued debt)
+    return _getUserDebt(assetId, user);
+  }
+
+  function _getUserDebt(uint256 assetId, address user) internal view returns (uint256) {
```
CheyenneAtapour: I was a fan of getting rid of this intermediary function because it isn't so long. If we do want to keep it, we also need to respect solidity guidelines, and put it after all external functions. But I prefer to not have it

src/contracts/BorrowModule.sol
```diff
@@ -132,14 +151,15 @@ contract BorrowModule is IBorrowModule {
   // TODO: Implement repay, calls liquidity hub restore method
   // TODO: onBehalfOf
   function repay(uint256 assetId, uint256 amount) external {
-    ILiquidityHub(liquidityHub).restore(assetId, amount);
+    ILiquidityHub(liquidityHub).restore(assetId, amount, 0);
```
CheyenneAtapour, line 133: Can add a TODO to calculate risk premium and pass here to LH

src/contracts/BorrowModule.sol
```diff
@@ -132,14 +151,15 @@ contract BorrowModule is IBorrowModule {
...
 
     emit Repaid(assetId, msg.sender, amount);
   }
 
   // TODO: Needed?
   function getInterestRate(uint256 assetId) public view returns (uint256) {
-    return ILiquidityHub(liquidityHub).getInterestRate(assetId);
+    // read from state, convert to ray
+    return reserves[assetId].borrowRate * 1e23;
```
CheyenneAtapour: Please revert this change, borrow rate will not be stored on BM, instead fetched from LH

src/contracts/BorrowModule.sol
```diff
@@ -185,7 +205,78 @@ contract BorrowModule is IBorrowModule {
     );*/
   }
 
+  // TODO: access control
+  function updateInterestRateStrategy(address newInterestRateStrategy) external {
+    interestRateStrategy = newInterestRateStrategy;
+  }
+
```
CheyenneAtapour: Don't need this, will be on LH side

src/contracts/BorrowModule.sol
```diff
@@ -185,7 +205,78 @@ contract BorrowModule is IBorrowModule {
...
+          uint40(reserve.lastUpdateTimestamp),
+          block.timestamp
+        )
+      ) +
+      amount;
+
+    reserve.lastUpdateTimestamp = block.timestamp;
+
+    // TODO: update index
+  }
```
CheyenneAtapour: Both these update state and accrue interest will happen on LH, please remove from here

src/contracts/LiquidityHub.sol
```diff
@@ -9,6 +9,7 @@ import {MathUtils} from './MathUtils.sol';
 import {ILiquidityHub} from '../interfaces/ILiquidityHub.sol';
 import {IReserveInterestRateStrategy} from '../interfaces/IReserveInterestRateStrategy.sol';
 import {DataTypes} from '../libraries/types/DataTypes.sol';
+import {IBorrowModule} from '../interfaces/IBorrowModule.sol';
```
CheyenneAtapour: Unused, please remove

src/contracts/BorrowModule.sol
yan-man: will user level interest accrual also happen in the LH? or just spoke level

src/contracts/BorrowModule.sol
yan-man: removed for now

src/contracts/BorrowModule.sol
CheyenneAtapour: So the shares on the Spoke and LH are exactly 1:1, so we only need the actual accrual to happen on LH side. The Spokes should just keep track of their users' shares, and can then distribute interest accordingly to users.

Side point: Can just think of the LH lending to "users" being the Spokes. Whatever users are using the Spokes are abstracted from the point of view of LH. Spokes will just handle everything related to users on Spoke level

src/contracts/BorrowModule.sol
```diff
@@ -60,31 +60,15 @@ contract BorrowModule is IBorrowModule {
...
     // TODO: Instead use a getter from liquidity hub to get up-to-date user debt (with accrued debt)
-    return
-      u.balance.rayMul(
-        MathUtils.calculateCompoundedInterest(
-          getInterestRate(assetId),
-          uint40(u.lastUpdateTimestamp),
-          block.timestamp
-        )
-      );
+    return 0;
```
CheyenneAtapour: I think you can leave the above calc in for now, instead of returning 0, because it serves as a good base. Just needs to be adjusted a bit

src/contracts/BorrowModule.sol
```diff
@@ -60,31 +60,15 @@ contract BorrowModule is IBorrowModule {
...
     // TODO: Instead use a getter from liquidity hub to get up-to-date reserve debt (with accrued debt)
-    return
-      r.totalDebt.rayMul(
-        MathUtils.calculateCompoundedInterest(
-          getInterestRate(assetId),
-          uint40(r.lastUpdateTimestamp),
-          block.timestamp
-        )
-      );
+    return 0;
```
CheyenneAtapour: Can leave the above calc in for now, similar to above

src/contracts/BorrowModule.sol
```diff
@@ -115,13 +99,14 @@ contract BorrowModule is IBorrowModule {
     emit Withdraw(assetId, msg.sender, amount);
   }
 
+  // TODO: On behalf of and referral code
   function borrow(uint256 assetId, uint256 amount) external {
-    // TODO: On behalf of and referral code
```
CheyenneAtapour: Small nit: I like having the todos as you do, above the function as you have it, but we just kept them within the function as our convention

src/contracts/BorrowModule.sol
yan-man: I removed it because we don't have `u.lastUpdateTimestamp` anymore, but will just use 0 there and leave the rest

src/contracts/BorrowModule.sol
yan-man: ditto

src/contracts/BorrowModule.sol
yan-man: I am agnostic, so I'll keep that in mind going forward

Issue #41, opened 2024-11-17
Implement risk premium in LH


miguelmtzinf:
Implement incremental weighted average of risk premiums across BMs
Every time a LH action happens:
- we aggregate the risk premium into the incremental weighted average risk premium
- we compute the new final borrow rate based on the current IR and the weighted average risk premium.

Blocked by #33

CanonicalJP:
Broke down into https://github.com/aave/aave-v4/issues/59 https://github.com/aave/aave-v4/issues/60 https://github.com/aave/aave-v4/issues/61 https://github.com/aave/aave-v4/issues/62 https://github.com/aave/aave-v4/issues/63

Issue #40, opened 2024-11-17
Implement risk premium in BM


miguelmtzinf:
Add liquidity premium parameter in BM
- Implement incremental weighted average of risk premium in BM
- Every time a user action happens, user's risk premium updates, and we compute the incremental weighted average.
- The new computed weighted average risk premium is passed to the LH for action execution.

Blocked by #33

CanonicalJP:
Replaced by #134 #135 #136

Issue #35, opened 2024-11-17
Calculation of user HF and risk premium


miguelmtzinf:
Add a user-facing function to set an asset as collateral.
- The user must have supplied the asset already

Create `getter` to compute HF of the user by iterating through user's position

Create `getter` to compute risk premium of the user
- weight = Amount of collateral * Price of the asset
- value = Liquidity premium of the asset

yan-man:
getter to compute risk premium of the user
- depends on the LP parameter in spoke, blocked by https://github.com/aave/aave-v4/issues/40

yan-man:
> Add a user-facing function to set an asset as collateral.

Also covered in https://github.com/aave/aave-v4/issues/25

PR #46, opened 2024-11-20
Implementation of Spoke/Liquidity Hub

yan-man:
Closes initial part of https://github.com/aave/aave-v4/issues/31
Will resolve tests after this is closed

Implement remaining functions in BM

- [x] Every action should follow the same approach:
- Validate action
- Refresh risk premium of the user and aggregated risk premium
- Execute corresponding action in LH by passing the agg weighted risk premium
- Accounting (using resulting shares from LH)
- Event
- All information about the IR and exchange rate between shares and assets is controlled by LH.

This info is used in the steps above (1 and 2 mostly), to convert shares to assets and viceversa.
- The LH should provide 2 getter functions, to provide conversions and up-to-date balances.
- [x] getter for conversions: convertAssetsToShares(amount, rounding) and convertSharesToAssets(amount, rounding)
- [x] getter for up-to-date balance: same as the previous, but we compute the interest accrual right before (totalAssets will increase a bit)
We can skip riskPremium logic stuff for now.

Let's do these 2 things here too:

- [x] Rename Borrow Module (BM) to Spoke across the code.

merge https://github.com/aave/aave-v4/pull/43 first

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -117,7 +117,7 @@ contract Spoke is ISpoke {
     emit Withdrawn(assetId, msg.sender, amount);
   }
 
-  function borrow(uint256 assetId, address onBehalfOf, uint256 amount) external {
+  function borrow(uint256 assetId, address to, uint256 amount) external {
     // TODO: referral code
     // TODO: onBehalfOf with credit delegation
     Reserve storage r = reserves[assetId];
@@ -127,14 +127,14 @@ contract Spoke is ISpoke {
     (, uint256 newAggregatedRiskPremium) = _refreshRiskPremium();
     uint256 userShares = ILiquidityHub(liquidityHub).draw(
       assetId,
-      onBehalfOf,
+      to,
       amount,
       newAggregatedRiskPremium
     );
     // debt still goes to original msg.sender
     users[assetId][msg.sender].debtShares += userShares;
 
-    emit Borrowed(assetId, onBehalfOf, amount);
+    emit Borrowed(assetId, to, amount);
   }
 
   function repay(uint256 assetId, uint256 amount) external {
--- a/tests/mocks/MockSpokeCreditLine.sol
+++ b/tests/mocks/MockSpokeCreditLine.sol
@@ -82,7 +82,7 @@ contract MockSpokeCreditLine is ISpoke {
       );
   }
 
-  function borrow(uint256 assetId, address onBehalfOf, uint256 amount) external {
+  function borrow(uint256 assetId, address to, uint256 amount) external {
     Reserve storage r = reserves[assetId];
     _validateBorrow(r, amount);
     // TODO: decide if state should be updated before or after liquidity hub call
@@ -90,12 +90,12 @@ contract MockSpokeCreditLine is ISpoke {
     _updateState(r, assetId, amount, msg.sender);
 
     // TODO: risk premium; to
-    ILiquidityHub(liquidityHub).draw(assetId, onBehalfOf, amount, 0);
+    ILiquidityHub(liquidityHub).draw(assetId, to, amount, 0);
 
     // keep liquidity in borrow module
-    IERC20(reserves[assetId].asset).safeTransfer(onBehalfOf, amount);
+    IERC20(reserves[assetId].asset).safeTransfer(to, amount);
 
-    emit Borrowed(assetId, onBehalfOf, amount);
+    emit Borrowed(assetId, to, amount);
   }
 
   // TODO: Implement repay, calls liquidity hub restore method


src/contracts/LiquidityHub.sol
```diff
@@ -81,7 +82,43 @@ contract LiquidityHub is ILiquidityHub {
...
+
+  function getInterestRate(uint256 assetId) public view returns (uint256) {
+    return assets[assetId].currentBorrowRate;
+  }
+
+  /**
+   * @param assetId The asset id
+   * @return The total balance of a given asset, either in shares or in assets
+   */
+  function getUpdatedAssetBalance(uint256 assetId, bool inShares) external returns (uint256) {
```
DhairyaSethi: maybe we rename this function to `updateAndGetAssetBalance` so it's abundantly clear it updates state by accruing interest, UI could staticcall in this for value
also nit (could be done for later): i think we should split it in two functions as `updateAndGetAssetBalance` & `updateAndGetAsset(In?)Shares` instead of the bool to maintain single responsibility principle
wdyt?

DhairyaSethi: or we modify our style guidelines to follow sablier's slightly and pass functions arguments verbosely wherever it's used like `liquidityHub.getUpdatedAssetBalance({ assetId: assetId, inShares: false })` for clarity on the boolean args

regardless this nit, we can note and do later after inital PoC

src/contracts/LiquidityHub.sol
```diff
@@ -129,21 +166,22 @@ contract LiquidityHub is ILiquidityHub {
 
   /// @dev risk premium is calculated from the spoke and passed upon every action
   function supply(uint256 assetId, uint256 amount, uint256 riskPremium) external returns (uint256) {
```
DhairyaSethi, line 136: perhaps we could store this `risk premium weighted sum` (this would be per asset or per spoke, [see](https://github.com/aave/aave-v4/pull/46/files#r1851557520)) on all of these actions in this PR itself? 
when using this weighted sum, for HF calc as per #35, we require a call BM regardless for user's pooled assets so i'm thinking we store the aggregated risk premium weighted sum per asset 
wdyt?

src/contracts/LiquidityHub.sol
```diff
@@ -81,7 +82,43 @@ contract LiquidityHub is ILiquidityHub {
     uint256 assetId,
     address spoke
   ) external view returns (SpokeConfig memory) {
-    return spokeAssetConfigs[assetId][spoke];
+    return spokes[assetId][spoke].config;
+  }
+
+  function getInterestRate(uint256 assetId) public view returns (uint256) {
+    return assets[assetId].currentBorrowRate;
```
DhairyaSethi: this needs to take into account the risk premium for the asset as well, perhaps we take into account here itself? (acc to [this](https://www.notion.so/aave/Liquidity-Premium-a70261fc71a14d77a807859a653db5eb?pvs=4#793982a0760748e780ebe66509f5add4)) would need to store the risk premium per asset on LH instead of per spoke

src/contracts/Spoke.sol
```diff
@@ -170,15 +193,41 @@ contract BorrowModule is IBorrowModule {
...
+  }
+
+  /**
+  @return uint256 new risk premium
+  @return uint256 new aggregated risk premium
+  */
+  function _updateState() internal returns (uint256, uint256) {
+    // TODO: update state - debt shares
+
+    // TODO: refresh risk premium of user, specific assets user has supplied
```
DhairyaSethi, line 229: this would be the risk premium of the spoke as a whole & not the user, right? since this is passed to LH and i believe it can work with just the RP of the Spoke and Spoke handles all user level values

src/contracts/LiquidityHub.sol
```diff
@@ -17,23 +17,24 @@ contract LiquidityHub is ILiquidityHub {
...
+    // TODO: lastUpdateTimestamp?
+    SpokeConfig config;
+  }
 
+  // TODO: borrow cap per spoke
   struct SpokeConfig {
     uint256 drawCap; // asset denominated
-    uint256 drawnShares;
     uint256 supplyCap; // asset denominated
-    uint256 shares;
```
CheyenneAtapour, line 29: Just to confirm: `shares` is now `totalShares` on the spoke, right?

src/contracts/LiquidityHub.sol
CheyenneAtapour: I think actually this should be the base interest rate that we should be fetching from liquidity hub. But indeed, we also need the calculation using the running weighted avg library to calculate IR with risk premium

DhairyaSethi: since LH is storing the aggregated RP of an asset (since multiple spokes can affect this), it makes sense to add it here or expose another getter and rename this to baseIR or something

DhairyaSethi: also, the final IR would unfortunately require the fetching the asset price, however this method is called, i believe the asset price would already be cached up the stack - so i see us optimizing this calculation down the line

src/contracts/LiquidityHub.sol
CheyenneAtapour: I agree with Dhairya on the rename because the way it's written, sounds like it should just be a view function. Prefixing it with `update` makes it clear it's changing the state

CheyenneAtapour:
Can you state in the description of this PR the exact issue # you are tackling?

src/contracts/LiquidityHub.sol
CheyenneAtapour: It doesn't matter to me which PR we implement the risk premium weighted sum on. But shouldn't the calculation be per user, not per asset?

DhairyaSethi:
+1 on the above, for example: if we're tackling storing RP and updating in a separate PR, we should track that in a separate issue (tho i think we can do storing here & tackle updating later after the lib interface is approved & merged)

src/contracts/LiquidityHub.sol
DhairyaSethi: on the LH it should be per asset imo

src/contracts/LiquidityHub.sol
```diff
@@ -129,21 +166,22 @@ contract LiquidityHub is ILiquidityHub {
...
+
     Asset storage asset = assets[assetId];
-    SpokeConfig storage spoke = spokeAssetConfigs[assetId][msg.sender];
+    Spoke storage spoke = spokes[assetId][msg.sender];
 
+    _validateSupply(asset, spoke, amount);
     // Update indexes and IRs
     _updateState(asset, spoke.drawnShares, riskPremium, amount, 0);
 
-    _validateSupply(asset, spoke, amount);
```
CheyenneAtapour, line 138: I think we do actually need to `_updateState` before validating because of such a scenario:

1. I deposit 1M USDC
2. 10 year pass by
3. The interest has made it so now we are above the supply cap
4. The new supply attempt should fail

src/contracts/LiquidityHub.sol
```diff
@@ -159,17 +197,18 @@ contract LiquidityHub is ILiquidityHub {
...
-    Asset storage asset = assets[assetId];
-    SpokeConfig storage spoke = spokeAssetConfigs[assetId][msg.sender];
+    // TODO: authorization - only spokes
 
-    _updateState(asset, spoke.drawnShares, riskPremium, 0, amount);
+    Asset storage asset = assets[assetId];
+    Spoke storage spoke = spokes[assetId][msg.sender];
 
     _validateWithdraw(asset, spoke, amount);
+    _updateState(asset, spoke.drawnShares, riskPremium, 0, amount);
```
CheyenneAtapour, line 173: These update state changes need to be reverted, see above. Miguel and I intentionally put them in the order they are in (update state, then validate)

src/contracts/LiquidityHub.sol
```diff
@@ -178,27 +217,27 @@ contract LiquidityHub is ILiquidityHub {
     return sharesAmount;
   }
 
-  // TODO: authorization - only spokes
   function draw(
     uint256 assetId,
-    address to,
+    address onBehalfOf,
```
CheyenneAtapour: This is not drawing on behalf of. That would be like if debt was delegated. This is just allowing the user to draw liquidity to a different address than their collateral came from. Should indeed stay as `to`

src/contracts/LiquidityHub.sol
```diff
@@ -178,27 +217,27 @@ contract LiquidityHub is ILiquidityHub {
...
+    // TODO: authorization - only spokes
+
     Asset storage asset = assets[assetId];
-    SpokeConfig storage spoke = spokeAssetConfigs[assetId][msg.sender];
+    Spoke storage spoke = spokes[assetId][msg.sender];
 
+    _validateDraw(asset, amount, spoke.config.drawCap);
     _updateState(asset, spoke.drawnShares, riskPremium, 0, amount);
 
-    _validateDraw(asset, amount, spoke.drawCap);
```
CheyenneAtapour, line 193: Actually do need update state, and then validate, see above

src/contracts/LiquidityHub.sol
```diff
@@ -178,27 +217,27 @@ contract LiquidityHub is ILiquidityHub {
...
-    asset.totalDrawnShares += sharesAmount;
+    uint256 sharesAmount = convertAssetsToShares(assetId, amount, true);
+    asset.drawnShares += sharesAmount;
     spoke.drawnShares += sharesAmount;
 
-    IERC20(assetsList[assetId]).safeTransfer(to, amount);
+    IERC20(assetsList[assetId]).safeTransfer(onBehalfOf, amount);
 
-    emit Draw(assetId, to, amount);
+    emit Draw(assetId, msg.sender, onBehalfOf, amount);
```
CheyenneAtapour: All the `onBehalfOf` should be reverted back to `to`

src/contracts/LiquidityHub.sol
```diff
@@ -208,15 +247,16 @@ contract LiquidityHub is ILiquidityHub {
     uint256 amount,
     uint256 riskPremium
   ) external returns (uint256) {
-    Asset storage asset = assets[assetId];
-    SpokeConfig storage spoke = spokeAssetConfigs[assetId][msg.sender];
+    // TODO: authorization - only spokes
 
-    _updateState(asset, spoke.drawnShares, riskPremium, amount, 0);
```
CheyenneAtapour, line 214: updateState does indeed need to go first (see above)

src/contracts/LiquidityHub.sol
```diff
@@ -63,15 +64,15 @@ contract LiquidityHub is ILiquidityHub {
 
   function getSpokeDrawnLiquidity(uint256 assetId, address spoke) public view returns (uint256) {
```
CheyenneAtapour: We should move these view functions below the external ones, in accordance with style guide https://docs.soliditylang.org/en/v0.8.21/style-guide.html#order-of-functions

src/contracts/Spoke.sol
```diff
@@ -79,14 +79,21 @@ contract BorrowModule is IBorrowModule {
...
+
   function supply(uint256 assetId, uint256 amount) external {
     Reserve storage r = reserves[assetId];
 
     _validateSupply(r, amount);
 
-    // TODO: Refresh risk premium of user, and pass into following function
-    uint256 newRiskPremium = 0;
-    uint256 userShares = ILiquidityHub(liquidityHub).supply(assetId, amount, newRiskPremium);
+    (, uint256 newAggregatedRiskPremium) = _updateState();
```
CheyenneAtapour: I think we should call this updateState function something like `refreshRiskPremium` instead for clarity

src/contracts/Spoke.sol
```diff
@@ -95,38 +102,54 @@ contract BorrowModule is IBorrowModule {
...
+    );
     users[assetId][msg.sender].supplyShares -= userShares;
 
-    emit Withdraw(assetId, msg.sender, amount);
+    emit Withdrawn(assetId, msg.sender, amount);
   }
 
-  function borrow(uint256 assetId, uint256 amount) external {
-    // TODO: On behalf of and referral code
+  function borrow(uint256 assetId, address onBehalfOf, uint256 amount) external {
```
CheyenneAtapour: I think `onBehalfOf` requires debt delegation, which we can tackle later, so can leave this out for now. Or leave this in, but leave the todo

src/contracts/Spoke.sol
```diff
@@ -170,15 +193,41 @@ contract BorrowModule is IBorrowModule {
...
   }
 
   function _validateBorrow(Reserve storage reserve, uint256 amount) internal view {
     require(reserve.config.borrowable, 'RESERVE_NOT_BORROWABLE');
   }
+
+  function _validateRepay(uint256 assetId, UserConfig storage user, uint256 amount) internal view {
+    require(
+      ILiquidityHub(liquidityHub).convertSharesToAssets(assetId, user.debtShares, true) >= amount,
+      'INSUFFICIENT_BALANCE'
```
CheyenneAtapour: Unsure this is the best error message. This is if they are repaying more than their debt. Can say instead something like `REPAY_EXCEEDS_DEBT`

src/contracts/Spoke.sol
CheyenneAtapour: The user involved in this action is what we recalculate the risk premium for. Then we update the risk premium for the entire spoke (weighted running average of risk premium of all users)

src/interfaces/ILiquidityHub.sol
```diff
@@ -9,17 +9,40 @@ pragma solidity ^0.8.0;
 interface ILiquidityHub {
   function draw(
     uint256 assetId,
-    address to,
+    address onBehalfOf,
```
CheyenneAtapour, line 12: Go back to `to`, will do `onBehalfOf` later on, as it requires credit delegation

src/contracts/Spoke.sol
DhairyaSethi: ah yep, i thought this was LH. you're right

tests/mocks/MockSpokeCreditLine.sol
```diff
@@ -82,20 +82,20 @@ contract MockBorrowModuleCreditLine is IBorrowModule {
       );
   }
 
-  function borrow(uint256 assetId, uint256 amount) external {
+  function borrow(uint256 assetId, address onBehalfOf, uint256 amount) external {
```
CheyenneAtapour: `to` instead of `onBehalfOf`

src/contracts/LiquidityHub.sol
yan-man: yes, shares -> totalShares, just to keep naming consistency with the Asset struct

src/contracts/LiquidityHub.sol
yan-man: agreed on both points, was debating between making 2 functions out of it but that makes sense for clarity

src/contracts/LiquidityHub.sol
yan-man: good point, was more re-arranging for consistency. So I'll update state first for each of the LH methods

src/contracts/LiquidityHub.sol
yan-man: I'm in favor of renaming this to baseIR and creating another getter later for the final IR that incorporates RP

DhairyaSethi: sgtm

src/contracts/LiquidityHub.sol
yan-man: decided on slack - will leave for separate PR and will be risk premium aggregated over all users* (with all of their collateral assets) and then aggregated per spoke

yan-man:
> +1 on the above, for example: if we're tackling storing RP and updating in a separate PR, we should track that in a separate issue (tho i think we can do storing here & tackle updating later after the lib interface is approved & merged)

yes @DhairyaSethi the RP related issues should be in: 
- https://github.com/aave/aave-v4/issues/40
- https://github.com/aave/aave-v4/issues/41

Issue #33, opened 2024-11-17
Library for incremental weighted averaged


miguelmtzinf:
Implement library to add and remove elements from an incremental weighted average

Create a standalone library with a function to calculate a new weighted average based on:
i) the current weighted average
ii) the current sum of weights
iii) the new item to add/remove
iv) the weight of the item to add/remove

Create unit tests and *fuzz tests*.

miguelmtzinf:
Created an initial library and run some tests against it some time ago, and found some challenges with numbers size and for items removal.
Please, check the draft implementation as it can serve as a base: https://gist.github.com/miguelmtzinf/989281f437c9f8061b5eb88fdd697327

Also, make sure the logic for adding and removing items from the weighted average is correct. There could be some off math there (not sure if subtracting the weight from the sum of weights when removing an item from the weighted average is correct).

CheyenneAtapour:
Just want to mention that this is a common math problem, and we can likely find an existing library/implementation for it. Probably just need to take care to ensure rounding issues are resolved properly (there are libraries for this too if needed)

DhairyaSethi:
@miguelmtzinf we loose precision because of the division, i haven't a way around this. the only way to preserve precision is to instead keep track of the (`weighted sum`, `sum of weights`) instead of (`weighted avg`, `sum of weights`) and do the devision on demand. we need to make everywhere this is used; the numerator doesn't overflow (if it can i've added PoC for 512 bit arithmetic) [conversation](https://avaraverse.slack.com/archives/C04HCBMEEBZ/p1732108656633489)

another idea i'm looking into is using fixed point math on the average, which is removed whenever we need to use the average

DhairyaSethi:
we've gone with the FixedPoint arithmetic & merged to unblock other tasks
- have to fix `Rad` naming
- alternative lib research is currently at https://github.com/aave/aave-v4/tree/tmp/weighted-avg-research which contain maintaining sum instead of avg in 512 bit precision & maintaining avg in 512 bit precision for the [liquidity rate case](https://www.notion.so/Incremental-Weighted-Average-Usage-1469d63a22de80d3aebdedae4de6deb2?pvs=4#1469d63a22de809da0e1e6d04110038e) if that's needed onchain 

keeping issue open until discussion on this is finalized and naming todo is fixed

DhairyaSethi:
@CanonicalJP this issue is not completed yet, subject to change based on the latest spec on usage of the lib (we may need to switch to 512 precision version)

[...]

PR #47, opened 2024-11-20
incremental weighted average

DhairyaSethi:
for #33


src/contracts/MathUtils.sol
```diff
@@ -96,4 +96,63 @@ library MathUtils {
...
+   */
+  function subtractFromWeightedAverage(
+    uint256 currentWeightedSum,
+    uint256 currentSumWeights,
+    uint256 newValue,
+    uint256 newValueWeight
+  ) internal view returns (uint256, uint256, uint256) {
+    // newWeightedSum, newSumWeights, weightedAvg
+    if (currentSumWeights == newValueWeight) return (0, 0, 0);
+    if (currentSumWeights < newValueWeight) revert();
```
yan-man, line 154: perhaps revert with a custom error / require?

src/contracts/MathUtils.sol
```diff
@@ -96,4 +96,63 @@ library MathUtils {
...
+    uint256 currentSumWeights,
+    uint256 newValue,
+    uint256 newValueWeight
+  ) internal view returns (uint256, uint256, uint256) {
+    // newWeightedSum, newSumWeights, weightedAvg
+    if (currentSumWeights == 0) {
+      return (newValue * newValueWeight, newValueWeight, newValue);
+    }
+
+    uint256 newSumWeights = currentSumWeights + newValueWeight;
```
yan-man, line 125: perhaps if newValueWeight == 0 we can short circuit and not go thru this calc, as it won't change current

src/contracts/MathUtilsExtended.sol
```diff
@@ -0,0 +1,178 @@
+// SPDX-License-Identifier: BUSL-1.1
```
yan-man: perhaps can add tests for extended lib? or is the plan to wait until it is confirmed to be needed first

DhairyaSethi: to the latter; yeah i think so, and if we were to go ahead with it i would refactor it to not use the struct (bc solidity stores that in memory & that achieves no purpose to us, much cheaper to keep stuff in the stack) so the interface would change as well. the core logic remains the same across libs, that's why i didn't add duplicate tests yet

src/contracts/MathUtils.sol
DhairyaSethi: yep! [f7ac505](https://github.com/aave/aave-v4/pull/47/commits/f7ac5059facb36107bd1f8e6df4f3b85289c73b9)

src/contracts/MathUtils.sol
DhairyaSethi: yep wanted to add but keeping the lib generic for now, we can add an error depending on the style we choose to go with for this codebase (for ex: i really like the new require(cond, customError))

yan-man: agreed, and would prefer the customError over strings generally

DhairyaSethi:
@yan-man let's change base branch to [feat/rebuild](https://github.com/aave/aave-v4/tree/feat/rebuild) & merge?

yan-man:
@DhairyaSethi I think it can go directly into main, is there any reason to go into feat/rebuild first?

DhairyaSethi:
@yan-man yea actually if we're merging #42 rn after merged your #50 rn, it can go in main itself

Issue #32, opened 2024-11-17
Complete LH and BM interfaces


miguelmtzinf:
Complete interfaces.
Move events to interfaces.
We can skip natspec docs in LH for now, but let's put inherit in LH implementation.

PR #50, opened 2024-11-21
Finalize first implementation of Spoke/Liquidity Hub - resolve tests

yan-man:
- Fix tests for spoke/LH implementation (specially unit tests, we can comment or skip invariant ones for now)
- resolve interfaces for LH, spoke

closes https://github.com/aave/aave-v4/issues/31, https://github.com/aave/aave-v4/issues/32

### DIFF
--- a/src/contracts/DefaultReserveInterestRateStrategy.sol
+++ b/src/contracts/DefaultReserveInterestRateStrategy.sol
@@ -6,8 +6,6 @@ import {Errors} from '../libraries/helpers/Errors.sol';
 import {IDefaultInterestRateStrategy} from '../interfaces/IDefaultInterestRateStrategy.sol';
 import {IReserveInterestRateStrategy} from '../interfaces/IReserveInterestRateStrategy.sol';
 
-import 'forge-std/console2.sol';
-
 /**
  * @title DefaultReserveInterestRateStrategy contract
  * @author Aave Labs
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -10,8 +10,6 @@ import {ISpoke} from '../interfaces/ISpoke.sol';
 import {IReserveInterestRateStrategy} from '../../src/interfaces/IReserveInterestRateStrategy.sol';
 import {DataTypes} from '../libraries/types/DataTypes.sol';
 
-import 'forge-std/console2.sol';
-
 contract Spoke is ISpoke {
   using WadRayMath for uint256;
   using SafeERC20 for IERC20;


src/contracts/Spoke.sol
```diff
@@ -89,11 +91,14 @@ contract Spoke is ISpoke {
     _validateSupply(r, amount);
 
     (, uint256 newAggregatedRiskPremium) = _refreshRiskPremium();
+    IERC20(r.asset).safeTransferFrom(msg.sender, address(this), amount);
```
yan-man: thoughts? should we just transferFrom user -> LH rather than from -> spoke -> LH. Need to update method input args and approval flow then

@DhairyaSethi @CheyenneAtapour

src/contracts/Spoke.sol
```diff
@@ -142,9 +147,12 @@ contract Spoke is ISpoke {
     // TODO: onBehalfOf
 
     UserConfig storage u = users[assetId][msg.sender];
+    Reserve storage r = reserves[assetId];
     _validateRepay(assetId, u, amount);
 
     (, uint256 newAggregatedRiskPremium) = _refreshRiskPremium();
+    IERC20(r.asset).safeTransferFrom(msg.sender, address(this), amount);
```
yan-man: ditto

src/contracts/Spoke.sol
DhairyaSethi: yeah i don't think funds should ever live on the spoke, we can either: 
user approves spoke, spoke transfers from user to LH (we should add an invariant later that there is always a callback updating internal accounting on transfers to lh initiated by spoke) 
or 
user approves LH which pulls funds & updates accounting in the same supply callback (need to pass user address to LH)

src/contracts/Spoke.sol
```diff
@@ -89,11 +91,14 @@ contract Spoke is ISpoke {
...
+    IERC20(r.asset).approve(liquidityHub, amount);
     uint256 userShares = ILiquidityHub(liquidityHub).supply(
       assetId,
       amount,
       newAggregatedRiskPremium
     );
+    IERC20(r.asset).approve(liquidityHub, 0);
```
DhairyaSethi: not needed regardless as the LH should spend the allowance, but i think we should change transfer flow anyway (comment above)

tests/unit/SpokeTest.t.sol
```diff
@@ -0,0 +1,281 @@
...
+
+    assertEq(dai.balanceOf(USER1), amount, 'wrong user token balance pre-supply');
+    assertEq(dai.balanceOf(address(hub)), 0, 'wrong hub token balance pre-supply');
+    assertEq(dai.balanceOf(address(spoke1)), 0, 'wrong spoke token balance pre-supply');
+    assertEq(userData.supplyShares, 0, 'wrong user shares pre-supply');
+    assertEq(userData.debtShares, 0, 'wrong user shares pre-supply');
+
+    vm.startPrank(USER1);
+    IERC20(dai).approve(address(spoke1), amount);
+    vm.expectEmit(true, true, true, false, address(spoke1));
```
DhairyaSethi:
```suggestion
   vm.expectEmit(address(spoke1));
```
why aren't we checking the 'data' field, ie amount? think we should

tests/unit/SpokeTest.t.sol
```diff
@@ -0,0 +1,281 @@
...
+    deal(address(dai), USER2, daiAmount);
+    Utils.spokeSupply(vm, hub, spoke1, daiId, USER2, daiAmount, USER2);
+
+    // USER1 borrow half of dai reserve liquidity
+    Utils.borrow(vm, spoke1, daiId, USER1, drawAmount, USER1);
+
+    // spoke1 restore half of drawn dai liquidity
+    vm.startPrank(USER1);
+    IERC20(address(dai)).approve(address(spoke1), restoreAmount);
+    vm.expectEmit(true, true, true, true, address(spoke1));
```
DhairyaSethi:
```suggestion
    vm.expectEmit(address(spoke1));
```
achieves the same result and is cleaner, same for all other expectEmits

tests/Utils.t.sol
```diff
@@ -21,6 +23,19 @@ library Utils {
...
+  function draw(
+    Vm vm,
+    LiquidityHub hub,
+    uint256 assetId,
+    address user,
+    uint256 amount,
+    address onBehalfOf
+  ) internal {
+    vm.startPrank(user);
+    hub.draw(assetId, user, amount, 0);
```
DhairyaSethi: this is incorrect, draw is only called by a spoke and not the user (see auth todo as well in LiquidityHub.draw)
all user interactions are through spoke and never LH, no?

tests/Utils.t.sol
```diff
@@ -21,6 +23,19 @@ library Utils {
     vm.stopPrank();
```
DhairyaSethi, line 38: user supplies through spoke now in the current design

tests/unit/SpokeTest.t.sol
```diff
@@ -0,0 +1,281 @@
...
+    // USER2 supply dai
+    deal(address(dai), USER2, daiAmount);
+    Utils.spokeSupply(vm, hub, spoke1, daiId, USER2, daiAmount, USER2);
+
+    Spoke.UserConfig memory user1Data = spoke1.getUser(ethId, USER1);
+    Spoke.UserConfig memory user2Data = spoke1.getUser(daiId, USER2);
+
+    assertEq(
+      user1Data.supplyShares,
+      ILiquidityHub(address(hub)).convertAssetsToShares(ethId, ethAmount, true),
```
DhairyaSethi: it's difficult to follow what the unnamed boolean args do, we're only using this paradigm here so we know if roundUp but i think we should move rename this function itself as `convertAssetsToSharesUp(ethId, ethAmount)` and down respectively  for clarity and verbosity 
or we be verbose in how we pass arguments, like `convertAssetsToShares({ assetId: ethId, amount: ethAmount, roundUp: true })`
would really prefer the former. wdyt?

tests/Utils.t.sol
```diff
@@ -3,9 +3,11 @@ pragma solidity ^0.8.0;
 
 import 'forge-std/Test.sol';
 import 'src/contracts/LiquidityHub.sol';
+import 'src/contracts/Spoke.sol';
 import 'src/dependencies/openzeppelin/IERC20.sol';
 
 library Utils {
```
DhairyaSethi, line 9: not sure how useful these utils functions are if they're only abstracting away an approve and direct method call w/o a need to mutate args

tests/unit/SpokeTest.t.sol
```diff
@@ -0,0 +1,281 @@
...
+    uint256 daiAssetId = 0;
+    hub.addAsset(
+      LiquidityHub.AssetConfig({decimals: 18, active: true, irStrategy: address(irStrategy)}),
+      address(dai)
+    );
+    spoke1.addReserve(
+      daiAssetId,
+      Spoke.ReserveConfig({lt: 0, lb: 0, borrowable: true, collateral: false}),
+      address(dai)
+    );
```
DhairyaSethi: we could abstract these actions in a util contract as they're repeated actions, `addAssetAndUpdateReserve(dai)` with overrides for strategy and config. wdyt - can we done at a later point as well, i just feel it helps organise the tests well

src/contracts/LiquidityHub.sol
```diff
@@ -66,7 +65,11 @@ contract LiquidityHub is ILiquidityHub {
     return assets[assetId];
   }
 
-  function getSpokeAssetConfig(
+  function getSpoke(uint256 assetId, address spoke) external view returns (Spoke memory) {
```
CheyenneAtapour, line 57: @yan-man I remember our discussion about this. I think it does need to be `SpokeAssetConfig` because we implicitly have a different spoke config for each different asset. It doesn't make sense here to query asset just to get the spoke. The assetId is relevant specifically to get the config. So I think it's clearer if this change is reverted and we just have a `getSpokeAssetConfig`

src/contracts/LiquidityHub.sol
```diff
@@ -106,26 +109,35 @@ contract LiquidityHub is ILiquidityHub {
...
     assets[assetId].config = AssetConfig({
       decimals: params.decimals,
       active: params.active,
-      supplyCap: params.supplyCap,
       irStrategy: params.irStrategy
     });
   }
 
-  function updateSpokeConfig(uint256 assetId, address spoke, uint256 drawCap) external {
+  function addSpoke(uint256 assetId, SpokeConfig memory params, address spoke) external {
```
CheyenneAtapour: This is fine, just makes me rethink, and I'm wondering if we can later on refactor the data structure `spokes`. I left a note about it on notion. Here we must update for all `assetId`s the spoke supports every time we add a new spoke

src/contracts/Spoke.sol
CheyenneAtapour: Yeah, I had actually thought about this too while we were writing it. I think this is something we add to the notion doc and discuss as an optimization / design decision later on

src/interfaces/ILiquidityHub.sol
```diff
@@ -24,6 +24,9 @@ interface ILiquidityHub {
 
   function getBaseInterestRate(uint256 assetId) external view returns (uint256);
 
+  // TODO: addAsset?
+  // TODO: addSpoke?
```
CheyenneAtapour: I think you can just go ahead and add these 2. If we want to simplify it later, it's easy to just remove them

tests/Utils.t.sol
CheyenneAtapour: this is correct

tests/unit/SpokeTest.t.sol
CheyenneAtapour: I agree with Dhairya. I think 2 dedicated functions: 1 for rounding up, and 1 for rounding down would be clearer

src/contracts/LiquidityHub.sol
yan-man: but I created this helper because there is no other means of getting individual spoke data, ie in tests, without querying asset.

But do you just mean reverting the naming, because I changed get `getSpokeAssetConfig` to `getSpokeConfig` so the functionality is still here. I can revert the name

src/contracts/LiquidityHub.sol
yan-man: will add a method to accept multiple assetIds/configs for addSpoke

src/contracts/Spoke.sol
yan-man: Agreed with Dhairya's suggestions

ok so I'll leave this as is and we can optimize later if it's been noted in notion

src/contracts/Spoke.sol
yan-man: should technically be spent by LH but wanted to reset for safety. Will leave as is like above and we can choose the optimization later

tests/Utils.t.sol
yan-man: yes, but it is just terminology then. user is spoke that invokes the draw, will update references

DhairyaSethi: i think it's dangerous to move forward with this terminology, since in the LH code we say onlyCallable by spoke
i understand spoke is a user in LH's context but i think it's well established that users will only be interacting through spoke and never LH so it doesn't make sense to refer to entities interacting with LH as user

tests/Utils.t.sol
yan-man: I think helpful for adjacent tests that just need some quick setup. I'll leave as is for now because Miguel created it initially

tests/Utils.t.sol
yan-man: updated terminology

src/contracts/Spoke.sol
DhairyaSethi: it's not really an optimization question, if LH doesn't spend allowance on supply then there's something wrong 
~~and anyway, if a user wants to max approve a spoke, then this action negates that & they'll have to max approve again (bc during max/infinite approvals, allowance doesn't get spent)~~ sorry this is approval by spoke to let LH spend. if we decide to transfer directly from user to LH, then should rm this

tests/unit/SpokeTest.t.sol
yan-man: Screenshot 2024-11-22 at 08 41 54

hm the docs say it should work but actually when I try this I get an error on compilation

yan-man: I'll leave it for now and add a todo for later

src/contracts/Spoke.sol
DhairyaSethi: guys i think we should update this here itself if we're aligned this isn't the best aproach haha,
let's go with suggestion 1 for now imo
- it's a relatively simple change, like you mentioned just the method args to transfer to LH, and approval to LH on tests instead of spoke
- we're leaving a bunch of things for future refactors where here atleast, since liquidity hub is supposed to manage all liq, it doesnt make sense for spoke to hold assets

DhairyaSethi: wdyt? if you feel strongly about leaving for later i understand ofcourse :)

src/contracts/LiquidityHub.sol
yan-man: added todo, also it is noted in notion

tests/unit/SpokeTest.t.sol
DhairyaSethi: forge-std dep was out of date, pushed an update for that, it should work now

tests/Utils.t.sol
DhairyaSethi: sure, sounds good!

src/contracts/Spoke.sol
yan-man: ok will implement option1 and we can still see about optimization later if needed

tests/unit/SpokeTest.t.sol
yan-man: thanks it works now

src/contracts/DefaultReserveInterestRateStrategy.sol
```diff
@@ -6,6 +6,8 @@ import {Errors} from '../libraries/helpers/Errors.sol';
 import {IDefaultInterestRateStrategy} from '../interfaces/IDefaultInterestRateStrategy.sol';
 import {IReserveInterestRateStrategy} from '../interfaces/IReserveInterestRateStrategy.sol';
 
+import 'forge-std/console2.sol';
```
DhairyaSethi: let's remove?

src/contracts/Spoke.sol
```diff
@@ -10,6 +10,8 @@ import {ISpoke} from '../interfaces/ISpoke.sol';
 import {IReserveInterestRateStrategy} from '../../src/interfaces/IReserveInterestRateStrategy.sol';
 import {DataTypes} from '../libraries/types/DataTypes.sol';
 
+import 'forge-std/console2.sol';
```
DhairyaSethi: can remove

Issue #34, opened 2024-11-17
AssetId and ReserveId management in Spoke


miguelmtzinf:
Implement a mapping in BM to match `assetId` with `reserveId`
- The asset listed in LH is just `asset`
- The asset listed in BM is called `reserve`
- Currently we are assuming both ids matches, but that's not the case.
- Users will come with an specific `reserveId` and we need to find the corresponding `assetId` to invoke LH with.

CheyenneAtapour:
So each BM (spoke) will have a mapping from it's `reserveId`'s to the LH's `assetId`'s. The LH does not need to care about what `reserveId`'s each BM is using

yan-man:
also: assigning reserveId as the latest reserveCount in Spoke

CanonicalJP:
@miguelmtzinf can this one be closed?

miguelmtzinf:
@CanonicalJP no, still relevant for the Spoke implementation imo

yan-man:
I think this can be closed now @miguelmtzinf

CanonicalJP:
Confirmed with @miguelmtzinf that this one is done

Issue #12, opened 2024-08-06
Enforce HF calculations in Spoke


miguelmtzinf:
Functions to calculate the HF of the user based on its position.
- supply does not force HF calc because it always increases it
- withdraw enforces HF calc
- borrow enforces HF calc
- repay does not force HF calc (if not repaying with aTokens)

We must evaluate if we want to change HF presentation: using percentage instead of bps. We should aim for simplicity because conversations can be made at UI level.

yan-man:
> Functions to calculate the HF of the user based on its position.

duplicated in https://github.com/aave/aave-v4/issues/35

PR #51, opened 2024-11-25
Calculation of User Health Factor / Risk Premium

yan-man:
- [x] Add a user-facing function to set an asset as collateral. (The user must have supplied the asset already)
- [x] Create getter to compute HF of the user by iterating through user's position
- [x] Create getter to compute risk premium of the user: (LP is assumed constant across all assets until https://github.com/aave/aave-v4/issues/40 is completed)
weight = Amount of collateral * Price of the asset
value = Liquidity premium of the asset
- [x] add setter for updating reserve config in spoke 
- [x] add revert tests for Spoke
- [x] add revert tests for LH
- [x] implement oracle definition into Spoke constructor

closes https://github.com/aave/aave-v4/issues/35

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -339,14 +339,8 @@ contract Spoke is ISpoke {
           vars.assetPrice *
           ILiquidityHub(liquidityHub).convertSharesToAssetsDown(
             vars.assetId,
-            u.supplyShares.rayMul(
-              MathUtils.calculateCompoundedInterest(
-                getInterestRate(vars.assetId),
-                uint40(0),
-                block.timestamp
-              )
-            )
-          ); // TODO: use lastUpdatedTimestamp in interest math, make sure collateral includes accrued supply interest
+            _calculateAccruedInterest(vars.assetId, u.supplyShares)
+          );
         vars.liquidityPremium = 1; // TODO: get LP from LH
         vars.totalCollateralInBaseCurrency += vars.userCollateralInBaseCurrency;
         vars.avgLiquidationThreshold += vars.userCollateralInBaseCurrency * r.config.lt;
@@ -357,13 +351,7 @@ contract Spoke is ISpoke {
         ? vars.assetPrice *
           ILiquidityHub(liquidityHub).convertSharesToAssetsUp(
             vars.assetId,
-            u.debtShares.rayMul(
-              MathUtils.calculateCompoundedInterest(
-                getInterestRate(vars.assetId),
-                uint40(0),
-                block.timestamp
-              )
-            ) // TODO: use lastUpdatedTimestamp in interest math, make sure total debt includes accrued debt interest
+            _calculateAccruedInterest(vars.assetId, u.debtShares)
           )
         : 0;
 
@@ -391,5 +379,16 @@ contract Spoke is ISpoke {
       vars.userRiskPremium,
       vars.healthFactor
     );
+  }
+
+  function _calculateAccruedInterest(
+    uint256 assetId,
+    uint256 shares
+  ) internal view returns (uint256) {
+    // TODO: use lastUpdatedTimestamp in interest math, make sure total shares includes accrued interest
+    return
+      shares.rayMul(
+        MathUtils.calculateCompoundedInterest(getInterestRate(assetId), uint40(0), block.timestamp)
+      );
   }
 }


src/contracts/PercentageMath.sol
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: BUSL-1.1
```
yan-man, line 1: added this file from v3

src/contracts/Spoke.sol
```diff
@@ -26,7 +28,7 @@ contract Spoke is ISpoke {
   }
 
   struct ReserveConfig {
-    uint256 lt;
+    uint256 lt; // 1e4 == 100%, BPS
```
yan-man, line 32: assumed this based on v3 @CheyenneAtapour @DhairyaSethi. Correct?

src/contracts/Spoke.sol
```diff
@@ -1,17 +1,21 @@
...
+import {MathUtils} from 'src/contracts/MathUtils.sol';
+import {PercentageMath} from 'src/contracts/PercentageMath.sol';
+import {ILiquidityHub} from 'src/interfaces/ILiquidityHub.sol';
+import {ISpoke} from 'src/interfaces/ISpoke.sol';
+import {IReserveInterestRateStrategy} from 'src/interfaces/IReserveInterestRateStrategy.sol';
+import {IPriceOracle} from 'src/interfaces/IPriceOracle.sol';
+import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+
+import 'forge-std/console2.sol';
 contract Spoke is ISpoke {
```
CheyenneAtapour, line 15: small nit: New line of whitespace before contract declaration

CheyenneAtapour:
Just a general note: This branch name could be shorter, like `feat/calc-hf-rp`

src/contracts/Spoke.sol
CheyenneAtapour: I think that's correct, yeah

src/contracts/Spoke.sol
```diff
@@ -158,6 +183,23 @@ contract Spoke is ISpoke {
...
+    (, , , uint256 userRiskPremium, ) = _calculateUserAccountData(user);
+    return userRiskPremium;
+  }
+
+  function getHealthFactor(address user) external view returns (uint256) {
+    (, , , , uint256 healthFactor) = _calculateUserAccountData(user);
+    return healthFactor;
+  }
+
+  function setUsingAsCollateral(uint256 assetId, bool usingAsCollateral) external {
```
CheyenneAtapour, line 195: I think it would be good to allow specifying the amount of asset to set as collateral as well. Can have an option for supplying `type(uint256).max` to use the total amount of the supplied asset as collateral.

But maybe this makes things more complicated. Will add it as a question to the notion doc

DhairyaSethi: that seems like an additional feature tbh, rn if a user wants to only specify use X% of totalSupplied as collateral, they'll have to supply 100-X% through another account. it would complicate things slightly, i say let's go with this for now and note your suggestion as a good feature. wdyt

src/contracts/Spoke.sol
```diff
@@ -212,6 +271,7 @@ contract Spoke is ISpoke {
 
   function _validateBorrow(Reserve storage reserve, uint256 amount) internal view {
     require(reserve.config.borrowable, 'RESERVE_NOT_BORROWABLE');
+    // TODO: validation on HF to allow borrowing amount
```
CheyenneAtapour, line 276: Perhaps this can be done on this PR as well?

src/contracts/Spoke.sol
```diff
@@ -35,28 +39,36 @@ contract Spoke is ISpoke {
...
+    uint256 avgLiquidationThreshold;
+    uint256 userRiskPremium;
+    uint256 healthFactor;
+  }
+
   // reserve id => user address => user data
   mapping(uint256 => mapping(address => UserConfig)) public users;
   // reserve id => reserveData
   mapping(uint256 => Reserve) public reserves;
+  uint256[] public reservesList; // assetIds
```
DhairyaSethi, line 64: should maintain a reserveCount as well, and should start assigning reserveId as the latest reserveCount - like in v3 - to handle cases when a reserve is offboarded and rm it from this array 
can be a future todo for now

src/contracts/Spoke.sol
```diff
@@ -234,4 +294,84 @@ contract Spoke is ISpoke {
...
+
+  function _borrowing(uint256 assetId, address user) internal view returns (bool) {
+    return users[assetId][user].debtShares > 0;
+  }
+
+  function _calculateUserAccountData(
+    address user
+  ) internal view returns (uint256, uint256, uint256, uint256, uint256) {
+    CalculateUserAccountDataVars memory vars;
+    vars.reservesListLength = reservesList.length;
```
DhairyaSethi: `reservesListLength` is just used once, think we can rm putting in memory (vars) and just cache reservesList.length on the stack

src/contracts/Spoke.sol
```diff
@@ -234,4 +294,84 @@ contract Spoke is ISpoke {
...
+      ? type(uint256).max
+      : vars.userRiskPremium.wadDiv(vars.totalCollateralInBaseCurrency);
+
+    vars.healthFactor = vars.totalDebtInBaseCurrency == 0
+      ? type(uint256).max
+      : (vars.totalCollateralInBaseCurrency.percentMul(vars.avgLiquidationThreshold)).wadDiv(
+        vars.totalDebtInBaseCurrency
+      ); // HF of 1 -> 1e18
+
+    // console2.log('avgLiquidationThreshold %e', vars.avgLiquidationThreshold);
```
CheyenneAtapour: Can remove these console logs when everything is finalized

tests/unit/HealthFactorTest.t.sol
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  function test_getHealthFactor_single_borrowed_asset() public {
+    uint256 daiId = 0;
+    uint256 ethId = 1;
+    uint256 usdcId = 2;
+    uint256 daiAmount = 10_000e18; // 10k dai -> $10k
+    uint256 ethAmount = 10e18; // 10 eth -> $20k
+    // total collateral -> $30k
+    uint256 usdcBorrowAmount = 15_000e18; // 15k usdc -> $15k
```
CheyenneAtapour, line 175: Kind of a small nit (I'd say like optional to fix), but this one made me double-take. The real `usdc` has only 6 decimals. Can you use e.g. `gho` here instead, which has 18 decimals?

Or if you can make your mock erc20 use 6 decimals, that'd also be nice

wbtc too -> usually 8 decimals

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -500,6 +526,37 @@ contract LiquidityHubTest is BaseTest {
...
+
+    // spoke1 draw all of dai reserve liquidity
+    Utils.draw(vm, hub, daiId, address(spoke1), amount, address(spoke1));
+
+    vm.prank(address(spoke1));
+    vm.expectRevert(TestErrors.SUPPLIED_AMOUNT_EXCEEDED);
+    hub.withdraw(daiId, address(spoke1), amount, 0);
+  }
+
+  function test_revert_withdraw_asset_not_active() public {
```
CheyenneAtapour: We can leave these for now, but might need to revisit when we consider other cases like `frozen`, `paused` etc. I think e.g. if it's inactive but not paused, users can still withdraw

tests/unit/SpokeCreditLineTest.t.sol
```diff
@@ -355,11 +355,11 @@ contract SpokeCreditLineTest is BaseTest {
   // }
 
   // function skip_test_revert_update_reserve() public {
-  //   uint256 invalidReserveId = 3;
+  //   uint256 invalidassetId = 3;
```
CheyenneAtapour: This is spoke side, so should be called `reserve`, right?

tests/unit/SpokeTest.t.sol
```diff
@@ -199,6 +283,34 @@ contract SpokeTest is BaseTest {
     assertEq(user1Data.supplyShares, 0, 'wrong user supply shares post-withdraw');
     assertEq(user1Data.debtShares, 0, 'wrong user debt shares post-withdraw');
   }
+
+  function test_revert_repay_repay_exceeds_debt() public {
```
CheyenneAtapour: change to say `repay` only once

`function test_revert_repay_exceeds_debt()`

tests/unit/UserRiskPremiumTest.t.sol
```diff
@@ -0,0 +1,254 @@
...
+        baseVariableBorrowRate: 500, // 5.00%
+        variableRateSlope1: 500, // 5.00%
+        variableRateSlope2: 500 // 5.00%
+      })
+    );
+  }
+
+  function test_getUserRiskPremium_no_supplied() public view {
+    uint256 userRiskPremium = ISpoke(spoke1).getUserRiskPremium(USER1);
+    assertEq(userRiskPremium, type(uint256).max, 'wrong user risk premium');
```
CheyenneAtapour: This convention is probably fine for now, but also we could have the user risk premium just be `0` in this case, right? That would fit the math definition

src/contracts/Spoke.sol
yan-man: was doing this just for the console2, but will make sure it doesn't happen on any others

yan-man:
> Just a general note: This branch name could be shorter, like `feat/calc-hf-rp`

just used the default naming from github on this branch when I created branch from issue but will keep in mind for next time

tests/unit/SpokeTest.t.sol
yan-man: was using the convention `test_functionName_errorMsg`, and the method is "repay" but the error is "REPAY_EXCEEDS_DEBT" so that's why it's there twice. Would prefer to keep it this way for consistency, thoughts? wdyt tie breaker @DhairyaSethi

tests/unit/LiquidityHubTest.t.sol
yan-man: ok will leave this as is for now, but I'll add some todo comments

tests/unit/SpokeTest.t.sol
DhairyaSethi: haha, i do like being consistent. my two cents would either we add an addendum to this (for word clashing), something like `test_revert_repay_with_repay_excess_debt()` or change the style to `test_functionName_revertsWith_errorMsg()`

src/contracts/Spoke.sol
yan-man: ok for now just initialized reserveCount but added some todo comments to tackle this during https://github.com/aave/aave-v4/issues/34

src/contracts/Spoke.sol
yan-man: agreed with you both, it is a potentially good feature but additional, so let's see what Emilio thinks and think through it before implementing

src/contracts/Spoke.sol
yan-man: will leave that for this ticket which does the enforcement on each of the spoke methods: https://github.com/aave/aave-v4/issues/12

tests/unit/HealthFactorTest.t.sol
yan-man: I'm going to leave this as is but add some todo comments about in BaseTest / mock erc20 (to be updated to accept decimals in constructor rather than default to 18)

tests/unit/SpokeTest.t.sol
yan-man: I think `test_functionName_revertsWith_errorMsg ` makes sense, and helps group tests together to all start with test_functionName, and clearly states the error msg. Will opt for this.

tests/unit/UserRiskPremiumTest.t.sol
yan-man: agreed, 0 makes more sense

src/contracts/Spoke.sol
CheyenneAtapour: Ah nice, ok!

src/contracts/LiquidityHub.sol
```diff
@@ -393,12 +394,12 @@ contract LiquidityHub is ILiquidityHub {
...
-        cumulated,
-        totalDrawn,
-        (cumulated - totalDrawn)
-      );
+      // console2.log(
+      //   'cumulated: %e, drawn: %e, cumulatedInterest: %e',
+      //   cumulated,
+      //   totalDrawn,
+      //   (cumulated - totalDrawn)
+      // );
```
CheyenneAtapour: Can remove these console.log before merging

src/contracts/Spoke.sol
```diff
@@ -319,8 +321,8 @@ contract Spoke is ISpoke {
     address user
   ) internal view returns (uint256, uint256, uint256, uint256, uint256) {
     CalculateUserAccountDataVars memory vars;
-    vars.reservesListLength = reservesList.length;
-    while (vars.i < vars.reservesListLength) {
+    uint256 reservesListLength = reservesList.length;
```
CheyenneAtapour, line 324: Is it necessary to store this in a variable rather than just use it directly?

DhairyaSethi: yeah it's a very standard gas optimization to avoid doing an sload on each iteration of the while loop (when exit condition is checked)

yan-man: yeah `.length` will be executed on each iteration without the caching

src/contracts/Spoke.sol
```diff
@@ -234,4 +296,82 @@ contract Spoke is ISpoke {
...
+
+      UserConfig memory u = getUser(vars.assetId, user);
+      Reserve memory r = getReserve(vars.assetId);
+
+      vars.assetPrice = IPriceOracle(oracle).getAssetPrice(vars.assetId);
+
+      if (_usingAsCollateral(vars.assetId, user)) {
+        vars.userCollateralInBaseCurrency =
+          vars.assetPrice *
+          ILiquidityHub(liquidityHub).convertSharesToAssetsDown(vars.assetId, u.supplyShares);
```
DhairyaSethi: don't we also need to include the yield on supplyShares?

yan-man: I was assuming that accruing interest on Spoke should update user's supplyShares, so supplyShares should already include yield

yan-man: updated supplyShares to also account for accrued interest, same for debtShares

DhairyaSethi: for visibility: we discussed this doesn't work bc it wouldn't accrue yield for all the previously assets whenever this internal method is called, fix on the way

PR #55, opened 2024-12-03
Implement Risk Premium in LH

### LINKED ISSUE #41: Implement risk premium in LH

CheyenneAtapour:
When a user action occurs (on spoke, which percolates to LH), 

1. We aggregate the received `newRiskPremium` parameter into our incremental weighted average risk premium.
2. We compute the new final borrow rate based on the current IR and the weighted average risk premium

closes #41

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -173,7 +173,12 @@ contract LiquidityHub is ILiquidityHub {
   // /////
 
   /// @dev risk premium is calculated from the spoke and passed upon every action
-  function supply(uint256 assetId, uint256 amount, uint256 riskPremium) external returns (uint256) {
+  function supply(
+    uint256 assetId,
+    uint256 amount,
+    uint256 riskPremium,
+    address supplier
+  ) external returns (uint256) {
     // TODO: authorization - only spokes
 
     Asset storage asset = assets[assetId];
@@ -198,8 +203,7 @@ contract LiquidityHub is ILiquidityHub {
     _updateBorrowRate(asset, riskPremium, amount, 0);
 
     // TODO: fee-on-transfer
-    // instead transferred by spoke from user to LH
-    // IERC20(assetsList[assetId]).safeTransferFrom(msg.sender, address(this), amount);
+    IERC20(assetsList[assetId]).safeTransferFrom(supplier, address(this), amount);
 
     emit Supply(assetId, msg.sender, amount);
 
@@ -278,12 +282,15 @@ contract LiquidityHub is ILiquidityHub {
    * @param amountFromPremium The amount to repay from premium interest
    * @param amountFromBase The amount to repay from base interest
    * @param riskPremium The aggregated risk premium of the calling spoke
+   * @param repayer The address who is trying to settle the credit line
+   * @return The amount of shares restored
    */
   function restore(
     uint256 assetId,
     uint256 amountFromPremium,
     uint256 amountFromBase,
-    uint256 riskPremium
+    uint256 riskPremium,
+    address repayer
   ) external returns (uint256) {
     // TODO: authorization - only spokes
 
@@ -307,6 +314,9 @@ contract LiquidityHub is ILiquidityHub {
     spoke.drawnShares -= sharesAmount;
 
     _updateBorrowRate(asset, riskPremium, amount, 0);
+
+    // TODO: fee-on-transfer, we receive at least `amount`
+    IERC20(assetsList[assetId]).safeTransferFrom(repayer, address(this), amount);
 
     emit Restore(assetId, msg.sender, amount);
 
@@ -543,7 +553,6 @@ contract LiquidityHub is ILiquidityHub {
       drawnShares: 0,
       config: DataTypes.SpokeConfig({supplyCap: params.supplyCap, drawCap: params.drawCap})
     });
-
     emit SpokeAdded(assetId, spoke);
   }
 }
--- a/tests/unit/LiquidityHubBorrowRateTest.t.sol
+++ b/tests/unit/LiquidityHubBorrowRateTest.t.sol
@@ -186,18 +186,24 @@ contract UserRiskPremiumTest is BaseTest {
   }
 
   function test_LHBorrowRate_Supply() public {
-    vm.prank(address(spoke1));
-    hub.supply(daiAssetId, 1000e18, 0);
+    deal(address(dai), address(spoke1), 1000e18);
+
+    vm.startPrank(address(spoke1));
+    LiquidityHub.Spoke memory test = hub.getSpoke(daiAssetId, address(spoke1));
+    dai.approve(address(hub), 1000e18);
+    hub.supply(daiAssetId, 1000e18, 0, address(spoke1));
     // No change to risk premium, so borrow rate is just the base rate
     assertEq(_getBaseBorrowRate(daiAssetId), _getBorrowRate(daiAssetId));
+    vm.stopPrank();
   }
 
   function test_LHBorrowRate_Borrow() public {
     // Spoke 1's first borrow should adjust the overall borrow rate with a risk premium of 10%
     uint256 newRiskPremium = 1e3;
-    deal(address(dai), address(hub), 1000e18);
+    deal(address(dai), address(spoke1), 1000e18);
     vm.startPrank(address(spoke1));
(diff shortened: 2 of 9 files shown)


src/contracts/LiquidityHub.sol
```diff
@@ -33,6 +33,11 @@ contract LiquidityHub is ILiquidityHub {
     DataTypes.AssetConfig config;
   }
 
+  struct WeightedAvg {
+    uint256 spokeBorrowRate;
+    uint256 amtDrawn; // TODO: I think this is just drawnShares on spoke
```
yan-man: nit: naming consistency, perhaps something `drawnAssets` to go along with `drawnShares` in other places

src/contracts/LiquidityHub.sol
```diff
@@ -400,13 +417,59 @@ contract LiquidityHub is ILiquidityHub {
     }
   }
 
-  function _calculateWeightedInterestRate(
-    uint256 borrowRate,
+  // TODO: We can fetch D_s,i, but perhaps better to just have spoke pass the updated value?
```
yan-man: see Notion thread, perhaps better to fetch D_s,i given we're already in LH here and it is stored vs allowing spoke to pass it

src/contracts/LiquidityHub.sol
```diff
@@ -400,13 +417,59 @@ contract LiquidityHub is ILiquidityHub {
...
     uint256 newRiskPremium,
-    uint256 spokeDrawnLiquidity
-  ) internal returns (uint256) {
-    // TODO: Add new value risk premium to weighted average
-    // TODO: Calculate final rate based on borrow rate and weighted average risk premium across spokes
+    uint256 newRiskPremiumWeight
+  ) internal {
+    // Check our saved risk premiums from this spoke to see if there is a change, if not, skip
+    if (
+      newRiskPremium == lastSpokeBorrowRate[msg.sender][assetId].spokeBorrowRate &&
```
yan-man: naming I think is a bit confusing because `R_s,i` is called both a risk premium and borrow rate. Would be good to settle on the terminology

src/contracts/LiquidityHub.sol
CheyenneAtapour: This is done

src/contracts/LiquidityHub.sol
```diff
@@ -353,19 +423,42 @@ contract LiquidityHub is ILiquidityHub {
...
+      uint256 currentAccruedBase = cumulatedBase - totalDrawnBase;
+
+      asset.totalAssetsBase = cumulatedBase;
+      asset.drawnSharesBase = cumulatedBase.toSharesDown(
+        asset.totalAssetsBase,
+        asset.totalSharesBase
+      );
+
+      // TODO: Double check math to add 1 (and rest of below) and also put into a library -> percentmul and fromRad
+      asset.totalPremium += (currentAccruedBase * (wAvgBR[asset.id].spokeBR / 1e28)) / 1e4;
```
DhairyaSethi, line 461: as mentioned on fri, consider using `.fromRad` and `.percentMul` helpers to abstract the divisions

src/contracts/LiquidityHub.sol
```diff
@@ -353,19 +423,42 @@ contract LiquidityHub is ILiquidityHub {
...
+      uint256 totalDrawnBase = convertSharesToAssetsUp(asset.id, asset.drawnSharesBase);
+      if (totalDrawnBase == 0) return; // No interest to accrue if no liquidity drawn
+      uint256 cumulatedBase = totalDrawnBase.rayMul(
+        MathUtils.calculateLinearInterest(baseBorrowRate, uint40(asset.lastUpdateTimestamp))
+      ); // TODO rounding
+
+      // Amount accrued since last action
+      uint256 currentAccruedBase = cumulatedBase - totalDrawnBase;
+
+      asset.totalAssetsBase = cumulatedBase;
```
yan-man: I think the problem lies here. `totalAssetsBase` is set to `cumulatedBase`, but `cumulatedBase` is the cumulated DRAWN base that is accumulated. i think perhaps it should be `asset.totalAssetsBase += currentAccruedBase` @CheyenneAtapour

yan-man: see https://github.com/aave/aave-v4/tree/feat/lh-rp-debug

src/contracts/LiquidityHub.sol
```diff
@@ -279,10 +341,15 @@ contract LiquidityHub is ILiquidityHub {
...
-    return assets[assetId].currentBorrowRate;
+    return assets[assetId].baseBorrowRate;
   }
 
-  // TODO: separate getter method for final IR that incorporates risk premium
+  function getInterestRate(uint256 assetId) public view returns (uint256) {
+    return
+      assets[assetId].baseBorrowRate +
+      (assets[assetId].baseBorrowRate * (wAvgBR[assetId].spokeBR.fromRad())) /
+      1e4;
```
DhairyaSethi, line 361:
```suggestion
      assets[assetId].baseBorrowRate.percentMul(wAvgBR[assetId].spokeBR).fromRad();
```
(`fromRad()` is after to keep maximum precision); similar feedback to everywhere we're doing `/ 1e4`

tests/unit/LiquidityHubBorrowRateTest.t.sol
```diff
@@ -0,0 +1,470 @@
...
+    rpSpoke2 = bound(rpSpoke2, 0, 99999);
+    drawSpoke2 = bound(drawSpoke2, 1, 1e40);
+
+    rpSpoke3 = bound(rpSpoke3, 0, 99999);
+    drawSpoke3 = bound(drawSpoke3, 1, 1e40);
+
+    deal(address(dai), address(hub), 6e40);
+
+    vm.startPrank(address(spoke1));
+    hub.supply(daiAssetId, 2e40, 0);
```
DhairyaSethi: i think we can put the args in a struct to avoid stack too deep, see how much of this struct can be re-used across tests. idt it's a big deal for having various structs in these test files, because this way the supply amount can be fuzzed as well instead of hardcoding to 2e40

tests/unit/LiquidityHubBorrowRateTest.t.sol
```diff
@@ -0,0 +1,470 @@
...
+    hub.supply(daiAssetId, 1000e18, 0);
+    hub.draw(daiAssetId, address(spoke1), 100e18, newRiskPremium);
+    vm.stopPrank();
+    uint256 borrowRate = _getBorrowRate(daiAssetId);
+    uint256 baseBorrowRate = _getBaseBorrowRate(daiAssetId);
+    assertEq(borrowRate, baseBorrowRate + (newRiskPremium * baseBorrowRate) / 1e4);
+  }
+
+  function test_LHBorrowRate_BorrowFuzz(uint256 newRiskPremium) public {
+    newRiskPremium = bound(newRiskPremium, 0, 99999);
```
DhairyaSethi, line 215: risk premium is a bps and max value should be `100_00` (100%), this is ceiled at 999.99% which is incorrect; pls fix.
we would probably have this implicitly checked on spoke side but i believe it's worth having a max value check on hub as well. wdyt

src/contracts/LiquidityHub.sol
```diff
@@ -190,14 +218,21 @@ contract LiquidityHub is ILiquidityHub {
...
+    // Accrue interest before validating action
+    _accrueAssetInterest(asset, asset.baseBorrowRate);
     _validateWithdraw(asset, spoke, amount);
 
     uint256 sharesAmount = convertAssetsToSharesDown(assetId, amount);
+    // TODO: On a withdraw, how do we know which shares (base or premium) to withdraw from? - Same as restore?
+    // It's just from base (total assets) because risk premium portion only relates to debt. Risk premium never available
+    asset.totalSharesBase -= sharesAmount;
     asset.totalShares -= sharesAmount;
+    asset.totalAssetsBase -= amount;
```
miguelmtzinf, line 234: This is problematic, because assets could have grown because of the premium, and `totalAssetsBase` can be not enough for withdrawals.
More on this [here](https://github.com/aave/aave-v4/issues/66)

Issue #56, opened 2024-12-05
Repo LICENSE


CanonicalJP:
We keep the LICENSE as `All Rights Reserved © Aave Labs` while defining the definitive LICENSE for when the codebase goes public. 

Having community contributions during development adds proprietary complexity that needs further analysis.

PR #57, opened 2024-12-05
Update LICENSE

CanonicalJP:
Change License to a more restrictive one while the final one is being defined. 

Associated with https://github.com/aave/aave-v4/issues/56


CanonicalJP:
I agreed with Legal to use this one, same as in https://github.com/aave/interface/blob/main/LICENSE.md. I'll keep working with legal on the definite one before going public. 

Because of community contributions, this will be a first in licensing for us and it's still not clear what approach to use. Most likely something like Uni's Business Source License but with a different edge cause of the these contributions.

Issue #53, opened 2024-11-27
Update token transfer flow in LH `supply`


yan-man:
Currently, during `supply` within LH, it is assumed that the spoke has already transferred the tokens from User to LH. Instead, based on Nov 27 call, the LH should do the `transferFrom` and pull the tokens from the User. 

- tests updated so User approves LH rather than Spoke during a `supply` on Spoke
- update LH `supply` method so that the Spoke passes the User address
- During supply, LH directly `transferFrom` provided User address

DhairyaSethi:
thinking about this change from a trust-relationship standpoint. in case of a compromised spoke, they can pass the address of any user arbitrarily (who have pre-approved the hub to spend) to pull tokens from

yan-man:
agreed on the trust assumption

Issue #71, opened 2025-01-15
Refactor Test Utils & Setup


DhairyaSethi:
1. `vm` does not need to be passed in Utils.supply()
https://github.com/aave/aave-v4/blob/4e457d9fd07bbe60578c4630552f9926d77cccbf/tests/Utils.t.sol#L27-L34
this can be calculated inside the lib since `vm` is always present at a deterministic address which is 
```solidity
Vm internal constant vm = Vm(address(uint160(uint256(keccak256("hevm cheat code")))));
```

2. Contracts and additional cast to same interface is being done which can be avoided 
https://github.com/aave/aave-v4/blob/4e457d9fd07bbe60578c4630552f9926d77cccbf/tests/BaseTest.t.sol#L90
https://github.com/aave/aave-v4/blob/4e457d9fd07bbe60578c4630552f9926d77cccbf/tests/unit/LiquidityHubTest.t.sol#L39
(same with liquidity hub and rest)

3. Replace warp + blocktimestamp usage in the tests with `skip` & `vm.getBlockTimestamp()` see related issues: https://github.com/foundry-rs/foundry/issues/8102#issuecomment-2154610488

yan-man:
- opt for vm.assume/bound over short circuiting

yan-man:
@DhairyaSethi please update 2) then:

https://github.com/aave/aave-v4/pull/115/files/ae80875fe66951c90c279ef0985d50fba2077be4#r1928333779

DhairyaSethi:
point 2 was meant to be used as a general example of the problem

PR #76, opened 2025-01-16
Resolve supply token transfer flow

yan-man:
Currently, during supply within LH, it is assumed that the spoke has already transferred the tokens from User to LH. Instead, based on Nov 27 call, the LH should do the transferFrom and pull the tokens from the User.

- tests updated so User approves LH rather than Spoke during a supply on Spoke
- update LH supply method so that the Spoke passes the User address
- During supply, LH directly transferFrom provided User address

closes https://github.com/aave/aave-v4/issues/53

### DIFF
--- a/tests/unit/LiquidityHubTest.t.sol
+++ b/tests/unit/LiquidityHubTest.t.sol
@@ -100,9 +100,9 @@ contract LiquidityHubTest is BaseTest {
 
     skip(20);
 
-    deal(address(dai), address(USER1), 1_000_000e18);
+    deal(address(dai), USER1, 1_000_000e18);
     vm.prank(USER1);
-    IERC20(dai).approve(address(hub), 1_000_000e18);
+    dai.approve(address(hub), 1_000_000e18);
   }
 
   function test_supply_revertsWith_ERC20InsufficientAllowance() public {
@@ -139,7 +139,7 @@ contract LiquidityHubTest is BaseTest {
     _updateSupplyCap(daiId, address(spoke1), amount - 1);
 
     vm.expectRevert(TestErrors.SUPPLY_CAP_EXCEEDED);
-    hub.supply(daiId, amount, 0, address(USER1));
+    hub.supply(daiId, amount, 0, USER1);
   }
 
   function test_first_supply() public {
@@ -149,7 +149,7 @@ contract LiquidityHubTest is BaseTest {
     LiquidityHub.Asset memory reserveData = hub.getAsset(assetId);
     LiquidityHub.Spoke memory spokeData = hub.getSpoke(assetId, address(spoke1));
 
-    uint256 initialUserBalance = dai.balanceOf(address(USER1));
+    uint256 initialUserBalance = dai.balanceOf(USER1);
 
     assertEq(reserveData.totalShares, 0, 'wrong reserve shares pre-supply');
     assertEq(reserveData.totalAssets, 0, 'wrong reserve assets pre-supply');
@@ -158,7 +158,7 @@ contract LiquidityHubTest is BaseTest {
     vm.startPrank(address(spoke1));
     vm.expectEmit(address(hub));
     emit Supply(assetId, address(spoke1), amount);
-    hub.supply(assetId, amount, 0, address(USER1));
+    hub.supply(assetId, amount, 0, USER1);
     vm.stopPrank();
 
     reserveData = hub.getAsset(assetId);
@@ -177,7 +177,7 @@ contract LiquidityHubTest is BaseTest {
     );
     assertEq(spokeData.drawnShares, 0, 'wrong spoke shares post-supply');
     assertEq(
-      dai.balanceOf(address(USER1)),
+      dai.balanceOf(USER1),
       initialUserBalance - amount,
       'wrong spoke token balance post-supply'
     );
@@ -234,7 +234,7 @@ contract LiquidityHubTest is BaseTest {
     emit Transfer(USER1, address(hub), amount);
     vm.expectEmit(address(hub));
     emit Supply(assetId, spoke, amount);
-    hub.supply(assetId, amount, 0, address(USER1));
+    hub.supply(assetId, amount, 0, USER1);
     vm.stopPrank();
   }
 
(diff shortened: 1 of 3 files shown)


src/contracts/LiquidityHub.sol
```diff
@@ -173,7 +173,12 @@ contract LiquidityHub is ILiquidityHub {
   // /////
 
   /// @dev risk premium is calculated from the spoke and passed upon every action
-  function supply(uint256 assetId, uint256 amount, uint256 riskPremium) external returns (uint256) {
+  function supply(
+    uint256 assetId,
+    uint256 amount,
+    uint256 riskPremium,
+    address user
```
DhairyaSethi: wdyt about renaming this to supplier?

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -100,25 +101,46 @@ contract LiquidityHubTest is BaseTest {
...
     uint256 amount = 100e18;
 
     _updateActive(daiId, false);
 
+    deal(address(dai), address(spoke1), amount);
     vm.prank(address(spoke1));
+    IERC20(dai).approve(address(hub), amount);
     vm.expectRevert(TestErrors.ASSET_NOT_ACTIVE);
-    hub.supply(daiId, amount, 0);
+    hub.supply(daiId, amount, 0, address(spoke1));
```
DhairyaSethi: technically spoke would never approve to spend its tokens, i see the workaround. this is nbg, but how about we structure the test for hub here such that we pass user address here and we max approve user to hub in the setup? 
the new allowance test is ok, can reset allowance to put in new test contract. 
wdyt?

yan-man: > can reset allowance to put in new test contract.

what do you mean here?

> we pass user address here and we max approve user to hub in the setup

yea I think this is a good suggestion

src/contracts/LiquidityHub.sol
yan-man: done

src/interfaces/ILiquidityHub.sol
```diff
@@ -21,7 +21,12 @@ interface ILiquidityHub {
     uint256 amountFromBase,
     uint256 riskPremium
   ) external returns (uint256);
-  function supply(uint256 assetId, uint256 amount, uint256 riskPremium) external returns (uint256);
+  function supply(
+    uint256 assetId,
+    uint256 amount,
+    uint256 riskPremium,
+    address user
```
CheyenneAtapour: A small nit, but do we wanna call this variable `supplier` as well to stay consistent between interface and actual implementation?

DhairyaSethi: yeah meant that we rename it to supplier everywhere it was introduced

DhairyaSethi: yeah meant that we rename it to supplier everywhere it was introduced

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: the user approves hub on all assets with uint.max allowance. helps keep susequent tests more to the point readable since theyre supposed to just be about borrow rate

src/interfaces/ILiquidityHub.sol
yan-man: yep absolutely, this is a miss from when I updated from Dhairya's suggestion

yan-man: done

tests/unit/LiquidityHubTest.t.sol
yan-man: done

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -98,6 +99,27 @@ contract LiquidityHubTest is BaseTest {
     MockPriceOracle(address(oracle)).setAssetPrice(daiCreditLineAssetId, 1e8);
 
     vm.warp(block.timestamp + 20);
+
+    deal(address(dai), address(USER1), type(uint256).max);
```
DhairyaSethi: bruh this is a recipe for disaster, let's not mint this much to the user bc:
- they can't receive more tokens, it'll overflow 
- balance change checks will be difficult 

let's mint them like 100k or something

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -98,6 +99,27 @@ contract LiquidityHubTest is BaseTest {
     MockPriceOracle(address(oracle)).setAssetPrice(daiCreditLineAssetId, 1e8);
 
     vm.warp(block.timestamp + 20);
```
DhairyaSethi: it is not recommended to use `warp` + `block.timestamp`, compiler can optimise the block.timestamp out 
it's best to use `skip` when advancing and `vm.getBlockTimestamp()` when current time is needed
i've added to issues here: https://github.com/aave/aave-v4/issues/71, we can tackle this seperately or here, up to you

DhairyaSethi: related issues https://github.com/foundry-rs/foundry/issues/8102#issuecomment-2154610488

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -108,38 +130,35 @@ contract LiquidityHubTest is BaseTest {
 
     vm.prank(address(spoke1));
     vm.expectRevert(TestErrors.ASSET_NOT_ACTIVE);
-    hub.supply(daiId, amount, 0);
+    hub.supply(daiId, amount, 0, address(USER1));
```
DhairyaSethi: no need to cast USER1 to address

tests/unit/LiquidityHubTest.t.sol
yan-man: I'll update it here but if it appears anywhere else will leave others for that next issue. Also updated our v4 conventions notion

done

tests/unit/LiquidityHubTest.t.sol
yan-man: done

tests/unit/LiquidityHubTest.t.sol
yan-man: done

PR #77, opened 2025-01-17
Resolve restore token flow

### LINKED ISSUE #75: Update token transfer flow in `restore/repay`
Currently, during supply within LH, it is assumed that the spoke has already transferred the tokens from User to LH. Instead, based on Nov 27 call, the LH should do the transferFrom and pull the tokens from the User.
 
- see https://github.com/aave/aave-v4/issues/53
- same updates for LH restore method and Spoke repay method

DhairyaSethi:
closes #75

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -177,7 +177,7 @@ contract Spoke is ISpoke {
       amountFromPremium: 0,
       amountFromBase: amount,
       riskPremium: newAggregatedRiskPremium,
-      settler: msg.sender
+      repayer: msg.sender
     });
     users[assetId][msg.sender].debtShares -= userShares;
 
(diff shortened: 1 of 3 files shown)


src/contracts/Spoke.sol
```diff
@@ -171,14 +171,14 @@ contract Spoke is ISpoke {
...
-      0,
-      amount,
-      newAggregatedRiskPremium
-    );
+    uint256 userShares = ILiquidityHub(liquidityHub).restore({
+      assetId: assetId,
+      amountFromPremium: 0,
+      amountFromBase: amount,
+      riskPremium: newAggregatedRiskPremium,
+      settler: msg.sender
```
yan-man: not a big deal, but what do you think of repayer? restorer? vs settler

DhairyaSethi: _settled_ on `repayer` lol [ba8b8d8](https://github.com/aave/aave-v4/pull/77/commits/ba8b8d84a47e8ce7dbcabf6fa5f72dc7e1e3a12a)

Issue #98, opened 2025-01-21
Short term bugfix


yan-man:
Update current version of DefaultReserveInterestRateStrategy to return values in ray instead of bps. Don't change underlying logic of calculations.

DhairyaSethi:
as mentioned in the title, short term fix. issue to be tracked in #95

Issue #68, opened 2025-01-14
Revert comments re: IDefaultInterestRateStrategy


yan-man:
Previously, this comment was changed from:
"The slope of the variable interest curve, after hitting the optimal ratio, in bps"

to 

"The slope of the variable interest curve, before hitting the optimal ratio, in bps interest rate per bps usage ratio"

Incorrect, should be reverted.
reference: https://github.com/aave/aave-v4/pull/29#discussion_r1816761244

PR #99, opened 2025-01-21
Short term bugfix for DefaultReserveInterestRateStrategy

yan-man:
- `forge test --mc DefaultReserveInterestRateStrategyTest` to see resolved tests

Update DefaultReserveInterestRateStrategy to return values in Ray so that it matches expected based on IReserveInterestRateStrategy. Short term bugfix to resolve output formats. Address longer-term solution (updating contract altogether to adapt to V2) in future.

closes https://github.com/aave/aave-v4/issues/68
closes https://github.com/aave/aave-v4/issues/98


DhairyaSethi:
feels like an ok patch while we focus on p1 stuff. can we pls update the comment on #114 to reflect this, maybe move it to the top of the file now to remind us to update this contract 
not able to reference it since it's not part of the changes 
@yan-man

Issue #100, opened 2025-01-22
fix _addSpoke


yan-man:
DataTypes.SpokeConfig definition is out of order (drawCap / supplyCap). Use struct initialization with named fields instead to clarify. 

Test failures appearing after initial refactoring in dev branch: https://github.com/aave/aave-v4/commit/c31134dfc216f81bf87026d0dfc7e479bd52f512

PR #101, opened 2025-01-22
Fix _addSpoke struct order

yan-man:
- `forge test --mt test_addSpokes` to see resolved tests
- consistent naming `params` -> `config`

Closes https://github.com/aave/aave-v4/issues/100

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -78,7 +78,7 @@ contract LiquidityHub is ILiquidityHub {
   // Governance
   // /////
 
-  function addAsset(DataTypes.AssetConfig memory params, address asset) external {
+  function addAsset(DataTypes.AssetConfig memory config, address asset) external {
     // TODO: AccessControl
     assetsList.push(IERC20(asset));
     _assets[assetCount] = Asset({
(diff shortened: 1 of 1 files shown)


src/contracts/LiquidityHub.sol
```diff
@@ -439,7 +439,7 @@ contract LiquidityHub is ILiquidityHub {
       baseBorrowIndex: WadRayMath.RAY,
       riskPremiumRad: 0,
       lastUpdateTimestamp: block.timestamp,
-      config: DataTypes.SpokeConfig(params.supplyCap, params.drawCap)
+      config: DataTypes.SpokeConfig({drawCap: config.drawCap, supplyCap: config.supplyCap})
```
DhairyaSethi, line 442: thx for the catch! here i think we can continue to avoid using named function args for brevity since variable names are descriptive enough. 
i acknowledge that had i kept using named args here the bug wouldn't have surfaced haha 

lgtm otherwise

yan-man: less about descriptiveness (bc I agree with you) but with the named fields we just don't need to worry about order then, which is safer imo.

DhairyaSethi: sure ok, i'd still prefer it being cleaner but not a strong opinion. let's ship it.

Issue #78, opened 2025-01-17
LH RP - supply implementation


yan-man:
`supply` method implementation on LiquidityHub

Issue #104, opened 2025-01-22
Fix `convertToAsssets{Down,Up}` and `convertToShares{Down,Up}` - Implementation


DhairyaSethi:
These methods currently return the last stored value but should accrue latest state and convert based on latest exchange ratio (similar to AToken.balanceOf)
Useful for external collaterboation 

Tests should not rely on these methods until fixed

yan-man:
Discussed on 1.22.25 - agree with Miguel, I think it makes sense to have separate getters. 1 for accrued/latest state values, and 1 for existing values in storage.

DhairyaSethi:
what's the usecase for an external method returning existing values in storage? @yan-man @miguelmtzinf

miguelmtzinf:
basically that, just to fetch stored data @DhairyaSethi . Can be used by integrators

PR #108, opened 2025-01-23
LH `supply` implementation and resolve LH tests

yan-man:
`forge test --mc LiquidityHubTest` to see passing tests
- resolve some accounting on supply method implementation
- resolve remaining LH tests to get passing CI
- next PR: add more tests/assertions on remaining LH draw, restore, withdraw flow using current implementation

closes https://github.com/aave/aave-v4/issues/79
closes https://github.com/aave/aave-v4/issues/78

### DIFF
--- a/tests/unit/LiquidityHubTest.t.sol
+++ b/tests/unit/LiquidityHubTest.t.sol
@@ -5,7 +5,7 @@ import '../BaseTest.t.sol';
 import {IERC20Errors} from 'src/dependencies/openzeppelin/IERC20Errors.sol';
 import {Asset, SpokeData} from 'src/contracts/LiquidityHub.sol';
 
-contract LiquidityHubTest_ToMigrate is BaseTest {
+contract LiquidityHubTest is BaseTest {
   using SharesMath for uint256;
   using WadRayMath for uint256;
 


src/contracts/AssetLogic.sol
```diff
@@ -9,6 +9,8 @@ import {SharesMath} from 'src/contracts/SharesMath.sol';
 import {PercentageMath} from 'src/contracts/PercentageMath.sol';
 import {WadRayMath} from 'src/contracts/WadRayMath.sol';
 
+import 'forge-std/console2.sol';
```
DhairyaSethi: let's not commit this on the src files

src/contracts/LiquidityHub.sol
```diff
@@ -12,6 +12,8 @@ import {SharesMath} from 'src/contracts/SharesMath.sol';
 import {MathUtils} from 'src/contracts/MathUtils.sol';
 import {PercentageMath} from 'src/contracts/PercentageMath.sol';
 
+import 'forge-std/console2.sol';
```
DhairyaSethi: same as above

tests/Utils.t.sol
```diff
@@ -32,8 +34,10 @@ library Utils {
     address user,
     address onBehalfOf
   ) internal {
-    vm.prank(user);
+    vm.startPrank(user);
```
DhairyaSethi, line 35: no need to start and stop prank when there's just one operation below, just pranking is enough and cleaner 
same applied for the hub.supply below, had kept start/stop prank incase we needed to add more functions in between

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -251,112 +305,177 @@ contract LiquidityHubTest_ToMigrate is BaseTest {
...
+    skip(1e4);
 
     assetData = hub.getAsset(assetId);
     spokeData = hub.getSpoke(assetId, address(spoke1));
 
+    // hub
+    assertEq(hub.getTotalAssets(assetId), amount, 'wrong total assets post-skip');
+    // asset
+    assertEq(assetData.availableLiquidity, amount, 'wrong asset availableLiquidity post-skip');
+    assertEq(assetData.baseDebt, 0, 'wrong asset baseDebt post-skip');
```
DhairyaSethi, line 387: these checks would make more sense if there was a user action on the hub which does an internal accounting update, and then we can check the desired changes that there is no change int he index etc since no drawn liquidity
so until we add an `updateIndexes` equivalent from compound, or `updateRiskPremium` for user based on internal accounting, we can do another supply. this is also imp bc we know `getAsset` or `getSpoke` currently does not do any interest accrual itself (maybe it also should?), just reads from state. related issue to fixed with #104

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -251,112 +305,177 @@ contract LiquidityHubTest_ToMigrate is BaseTest {
...
-    // uint256 spoke2SupplyShares = 1; // minimum for 1 share
-    // uint256 spoke2SupplyAssets = ILiquidityHub(address(hub)).convertToAssetsDown(
-    //   assetId,
-    //   spoke2SupplyShares
-    // );
+    // state update due to operation
+    // TODO helper for reserve state update
+    uint256 spoke2SupplyShares = 1; // minimum for 1 share
+    uint256 spoke2SupplyAssets = ILiquidityHub(address(hub)).convertToAssetsDown(
+      assetId,
```
DhairyaSethi: no need to cast hub again to ILiquidityHub

tests/Utils.t.sol
yan-man: this i specifically changed because vm.prank was causing me all the issues with insufficient approvals and I didn't get why. But it's bc the call to hub.assetsList(assetId) consumed the prank and was making it so the approve method after wasn't going through properly

DhairyaSethi: ohh right that's right, i overlooked. i've banged my around this issue as well, hub.assetLists does consume the prank

tests/unit/LiquidityHubTest.t.sol
yan-man: yeah another supply is used to trigger update (was already done this way previously in this test)

src/contracts/LiquidityHub.sol
```diff
@@ -168,16 +168,15 @@ contract LiquidityHub is ILiquidityHub {
...
     });
 
-    asset.availableLiquidity += amount;
-
     // todo: Mitigate inflation attack (burn some amount if first supply)
     uint256 sharesAmount = asset.convertToSharesDown(amount);
-    require(sharesAmount > 0, 'INVALID_AMOUNT');
+    require(sharesAmount > 0, 'INVALID_SHARES_AMOUNT');
 
+    asset.availableLiquidity += amount;
```
yan-man, line 179: note @DhairyaSethi moved accounting so that shares calc is based on existing assets

src/contracts/LiquidityHub.sol
```diff
@@ -207,11 +206,9 @@ contract LiquidityHub is ILiquidityHub {
     asset.updateBorrowRate({liquidityAdded: 0, liquidityTaken: amount});
     _updateRiskPremiumAndBaseDebt(asset, spoke, riskPremiumRad, 0); // no base debt change
 
-    asset.availableLiquidity -= amount;
-
     uint256 sharesAmount = asset.convertToSharesDown(amount);
-
     asset.suppliedShares -= sharesAmount;
+    asset.availableLiquidity -= amount;
```
yan-man, line 211: same

src/contracts/LiquidityHub.sol
DhairyaSethi: intentionally not on 'convertToAssets' used above, it's expected to be used internally and is just supposed to be abstraction of current stored exchange rate math (w/o accruing)

DhairyaSethi: #111

yan-man: I think partially unrelated to #111 because it is merely an accounting order issue rather than a shares/asset conversion issue. I guess relates to using an already-cached total asset value instead of an updated total asset value. less about accruing interest on the base debt, but the change in added liquidity side

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -828,69 +983,97 @@ contract LiquidityHubTest_ToMigrate is BaseTest {
...
     // );
     // assertEq(spoke2EthData.totalShares, 0, 'wrong spoke2 total eth shares post-restore');
     // assertEq(spoke2EthData.drawnShares, 0, 'wrong spoke2 drawn eth shares post-restore');
     // assertEq(
     //   spoke2DaiData.totalShares,
-    //   ILiquidityHub(address(hub)).convertToSharesDown(daiId, daiAmount),
+    //   hub.convertToSharesDown(daiId, daiAmount),
     //   'wrong spoke2 total dai shares post-restore'
     // );
     // assertEq(spoke2DaiData.drawnShares, 0, 'wrong spoke2 drawn dai shares post-restore');
```
CheyenneAtapour, line 1068: Do we need all this commented code for another PR migration, or can we remove here?

yan-man: yes because need next PR: add more tests/assertions on remaining LH draw, restore, withdraw flow using current implementation

src/contracts/LiquidityHub.sol
DhairyaSethi: yeah kinda the main point being here we use this method after already accruing debt and saving to state such the value can be reused 
so it's needed as an internal helper

PR #109, opened 2025-01-23
Refactor Borrow Rate Tests

### LINKED ISSUE #110: LH RP - Migrate borrow rate tests
With the new algorithm, we have to refactor our test suite. This issue covers the test suite defined in file LiquidityHubBorrowRateTest.t.sol testing the calculation of the borrow rate using the risk premium and base borrow rate

PR here: https://github.com/aave/aave-v4/pull/109

CheyenneAtapour:
Closes #110 

Depends on #106


tests/unit/LiquidityHubBorrowRateTest.t.sol
```diff
@@ -187,11 +23,11 @@ contract UserRiskPremiumTest_ToMigrate is BaseTest {
   }
 
   function test_LHBorrowRate_Supply() public {
-    deal(address(dai), address(spoke1), 1000e18);
+    deal(address(tokenList.dai), address(spoke1), 1000e18);
 
     vm.startPrank(address(spoke1));
     SpokeData memory test = hub.getSpoke(daiAssetId, address(spoke1));
-    dai.approve(address(hub), 1000e18);
+    IERC20(tokenList.dai).approve(address(hub), 1000e18);
```
DhairyaSethi: no need to cast to `IERC20`

tests/unit/LiquidityHubBorrowRateTest.t.sol
```diff
@@ -200,194 +36,197 @@ contract UserRiskPremiumTest_ToMigrate is BaseTest {
...
     baseBorrowRate = _getBaseBorrowRate(daiAssetId);
-    assertEq(borrowRate, baseBorrowRate + (newRiskPremium * baseBorrowRate) / 1e4);
+    assertEq(borrowRate, baseBorrowRate + (newRiskPremium.radToRay().rayMul(baseBorrowRate)));
     vm.stopPrank();
   }
 
   function test_LHBorrowRate_BorrowTwice() public {
-    uint256 newRiskPremium = 1e3;
-    deal(address(dai), address(spoke1), 1000e18);
+    uint256 newRiskPremium = uint256(1e3).bpsToRad();
```
DhairyaSethi: can we use 1_00 notation instead of scientific for bps values? makes it more obvious what % are. ngd

tests/unit/LiquidityHubBorrowRateTest.t.sol
```diff
@@ -200,194 +36,197 @@ contract UserRiskPremiumTest_ToMigrate is BaseTest {
...
-    assertEq(borrowRate, baseBorrowRate + (newRiskPremium2 * baseBorrowRate) / 1e4);
+    assertEq(borrowRate, baseBorrowRate + (newRiskPremium2.radToRay().rayMul(baseBorrowRate)));
     vm.stopPrank();
   }
 
   function test_LHBorrowRate_BorrowTwiceFuzz(uint256 newRiskPremium) public {
     newRiskPremium = bound(newRiskPremium, 0, 99999);
-    uint256 firstRiskPremium = 1e3;
-    deal(address(dai), address(spoke1), 1000e18);
+    newRiskPremium = newRiskPremium.bpsToRad();
```
DhairyaSethi: here i think you should instead add this to bound 
newRiskPremium = bound(newRiskPremium, 0 uint256(1000_00 - 1).bpsToRad());

this way newRiskPremium can be bw 999.999...% instead of 999.99%

also we should maybe make a const on what max possible value could be, maybe as a todo for later. wdyt?

tests/unit/LiquidityHubBorrowRateTest.t.sol
```diff
@@ -200,194 +36,197 @@ contract UserRiskPremiumTest_ToMigrate is BaseTest {
...
-    dai.approve(address(hub), 1000e18);
+    IERC20(tokenList.dai).approve(address(hub), 1000e18);
     hub.supply(daiAssetId, 1000e18, 0, address(spoke2));
     hub.draw(daiAssetId, address(spoke2), 100e18, rpSpoke2);
     borrowRate = _getBorrowRate(daiAssetId);
     baseBorrowRate = _getBaseBorrowRate(daiAssetId);
-    assertEq(borrowRate, baseBorrowRate + ((rpSpoke1 + rpSpoke2) * baseBorrowRate) / 2e4);
+    assertEq(
+      borrowRate,
+      baseBorrowRate + ((rpSpoke1 + rpSpoke2).radToRay().rayMul(baseBorrowRate)).rayDiv(2e27)
```
DhairyaSethi: imo let's not do constants instead let's do 2 * WadRayMath.RAY

tests/unit/LiquidityHubBorrowRateTest.t.sol
yan-man: > make a const on what max possible value could be

regardless of if we set a min/max cap, just as a test helper you mean @DhairyaSethi ?

tests/unit/LiquidityHubBorrowRateTest.t.sol
yan-man: agreed

tests/unit/LiquidityHubBorrowRateTest.t.sol
```diff
@@ -200,194 +36,197 @@ contract UserRiskPremiumTest_ToMigrate is BaseTest {
 
   function test_LHBorrowRate_Borrow() public {
     // Spoke 1's first borrow should adjust the overall borrow rate with a risk premium of 10%
-    uint256 newRiskPremium = 1e3;
-    deal(address(dai), address(spoke1), 1000e18);
+    uint256 newRiskPremium = uint256(10_00).bpsToRad();
+    deal(address(tokenList.dai), address(spoke1), 1000e18);
     vm.startPrank(address(spoke1));
-    dai.approve(address(hub), 1000e18);
+    IERC20(tokenList.dai).approve(address(hub), 1000e18);
```
yan-man: same no need to cast

tests/unit/LiquidityHubBorrowRateTest.t.sol
```diff
@@ -200,194 +36,197 @@ contract UserRiskPremiumTest_ToMigrate is BaseTest {
...
     hub.supply(daiAssetId, 1000e18, newRiskPremium, address(spoke1));
     borrowRate = _getBorrowRate(daiAssetId);
     baseBorrowRate = _getBaseBorrowRate(daiAssetId);
-    assertEq(borrowRate, baseBorrowRate + (newRiskPremium * baseBorrowRate) / 1e4);
+    assertEq(borrowRate, baseBorrowRate + (newRiskPremium.radToRay().rayMul(baseBorrowRate)));
     vm.stopPrank();
   }
 
   function test_LHBorrowRate_BorrowAndSupplyFuzz(uint256 newRiskPremium) public {
```
yan-man: minor on fuzz naming conventions, so it's easier to isolate/ignore fuzz tests: https://www.notion.so/aave/V4-Code-Conventions-17a9d63a22de80898fb0ee1f6df6038b?pvs=4#17b9d63a22de8067a975e52104fcce03

tests/unit/LiquidityHubBorrowRateTest.t.sol
DhairyaSethi: yeah for general bounds, can be a test helper. imo, we can make bound helpers in base test, something like boundRiskPremium(rp) and use everywhere needed

PR #111, opened 2025-01-23
poc: preview assets

DhairyaSethi:
should refactor lib to do inflight calculations in memory and commit to state at the end
will need to look if this is more expensive than current approach on `dev` with storage points not pulling the whole struct in memory

tmp solution for #104

### DIFF
--- a/src/contracts/AssetLogic.sol
+++ b/src/contracts/AssetLogic.sol
@@ -57,12 +57,8 @@ library AssetLogic {
     return shares.toAssetsDown(asset.totalAssets(), asset.totalShares());
   }
 
-  // todo carry out mul in rad for precision
-  function getInterestRate(Asset storage asset) internal view returns (uint256) {
-    return
-      asset.baseBorrowRate.percentMul(
-        PercentageMath.PERCENTAGE_FACTOR + asset.riskPremiumRad.radToBps()
-      );
+  function getInterestRate(Asset storage asset) external view returns (uint256) {
+    return asset.baseBorrowRate.radMul(WadRayMath.RAD + asset.riskPremiumRad);
   }
 
   function updateBorrowRate(
--- a/src/contracts/DefaultReserveInterestRateStrategy.sol
+++ b/src/contracts/DefaultReserveInterestRateStrategy.sol
@@ -6,6 +6,7 @@ import {DataTypes} from '../libraries/types/DataTypes.sol';
 import {Errors} from '../libraries/helpers/Errors.sol';
 import {IDefaultInterestRateStrategy} from '../interfaces/IDefaultInterestRateStrategy.sol';
 import {IReserveInterestRateStrategy} from '../interfaces/IReserveInterestRateStrategy.sol';
+import {WadRayMath} from './WadRayMath.sol';
 
 // TODO: update this contract to based on DefaultReserveInterestRateStrategyV2 in aave-v3-origin
 
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -157,7 +157,7 @@ contract LiquidityHub is ILiquidityHub {
   function supply(
     uint256 assetId,
     uint256 amount,
-    uint256 riskPremium,
+    uint256 riskPremiumRad,
     address supplier
   ) external returns (uint256, uint256) {
     // TODO: authorization - only spokes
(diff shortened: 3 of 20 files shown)


src/contracts/AssetLogic.sol
```diff
@@ -101,23 +104,32 @@ library AssetLogic {
...
+    (uint256 accruedBaseDebt, uint256 accruedOutstandingPremium) = asset.previewAccruedDebt(
+      cumulatedBaseInterest
+    );
+    if (accruedBaseDebt > 0) asset.baseDebt += accruedBaseDebt;
+    if (accruedOutstandingPremium > 0) asset.outstandingPremium += accruedOutstandingPremium;
+    asset.baseBorrowIndex = nextBaseBorrowIndex;
+    asset.lastUpdateTimestamp = block.timestamp;
+  }
 
+  function previewAccruedDebt(
```
yan-man, line 113: this and accrueInterest should be _functionName format for internal method (and all other internal methods here too)

src/contracts/AssetLogic.sol
```diff
@@ -101,23 +104,32 @@ library AssetLogic {
...
     uint256 existingBaseDebt = asset.baseDebt;
     // no interest to accrue since no liquidity has been drawn
-    if (existingBaseDebt == 0) return;
+    if (existingBaseDebt == 0) return (0, 0);
 
     // can use `cumulatedBaseInterest` instead of `indexRatio` since LH base debt is
     // accrued on each index update
-    uint256 cumulatedBaseDebt = asset.baseDebt.rayMul(cumulatedBaseInterest);
-
+    uint256 accruedBaseDebt = asset.baseDebt.rayMul(cumulatedBaseInterest - WadRayMath.RAY);
```
yan-man, line 125: re-use cached `existingBaseDebt` in place of `asset.baseDebt`

src/contracts/AssetLogic.sol
yan-man: and rename to previewAccruedDebtAndPremium or similar?

src/contracts/AssetLogic.sol
```diff
@@ -85,7 +88,7 @@ library AssetLogic {
   // @dev Utilizes existing `asset.baseBorrowRate` & `asset.baseBorrowIndex`
   // @return cumulatedBaseInterest (in ray)
   // @return nextBaseBorrowIndex (in ray)
-  function previewNextBorrowIndex(Asset storage asset) external view returns (uint256, uint256) {
+  function previewNextBorrowIndex(Asset storage asset) internal view returns (uint256, uint256) {
     uint256 elapsed = block.timestamp - asset.lastUpdateTimestamp;
```
yan-man, line 88: cache asset.lastUpdateTimestamp to be re-used?

src/contracts/AssetLogic.sol
DhairyaSethi: idt underscore convention should apply to libs, it's kinda implied from whether to lib is used purely internally (inlined) or externally. this lib is currently doing both, will be going to make it strictly either

src/contracts/AssetLogic.sol
```diff
@@ -101,23 +104,32 @@ library AssetLogic {
...
     // accrue premium interest on the accrued base interest
-    asset.outstandingPremium += (cumulatedBaseDebt - existingBaseDebt).percentMul(
-      asset.riskPremiumRad.radToBps()
-    );
-    asset.baseDebt = cumulatedBaseDebt;
-    asset.baseBorrowIndex = nextBaseBorrowIndex;
-    asset.lastUpdateTimestamp = block.timestamp;
+    uint256 accruedOutstandingPremium = accruedBaseDebt.percentMul(asset.riskPremiumRad.radToBps());
+
+    return (accruedBaseDebt, accruedOutstandingPremium);
```
yan-man, line 129: maybe if we returned existingBaseDebt here, would save an sload in accrueInterest

src/contracts/AssetLogic.sol
DhairyaSethi: yeah was kinda leaving the optimsations out to keep with existing flow. instead of passing values (not super clean) i would prefer to do a rewrite and most of these preview calcs in memory (there's a todo for since i wrote this lib last week)

DhairyaSethi:
closing in favour of impl at #174

Issue #60, opened 2025-01-11
LH RP - Supply & Withdraw


CheyenneAtapour:
* Implement supply and withdraw function on the liquidity hub using our new algorithm for dealing with risk premium, interest accrual, assets and shares on spoke and on liquidity hub
* Ensure the math works appropriately, keeping base interest accrued separately from risk premium interest accrued, and that shares and assets are handled appropriately for both the spoke side and liquidity hub side

PR #115, opened 2025-01-24
LH `withdraw` refactor and implementation

yan-man:
- resolve withdraw tests + impl
- polish some supply tests
- refactor and optimize utils

closes https://github.com/aave/aave-v4/issues/81
closes https://github.com/aave/aave-v4/issues/80
closes https://github.com/aave/aave-v4/issues/71
closes https://github.com/aave/aave-v4/issues/60

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -236,7 +236,6 @@ contract LiquidityHub is ILiquidityHub {
     _updateRiskPremiumAndBaseDebt(asset, spoke, riskPremiumRad, int256(amount)); // base debt added
 
     asset.availableLiquidity -= amount;
-    // TODO: increase base debt on asset/spoke
 
     assetsList[assetId].safeTransfer(to, amount);
 
@@ -274,7 +273,6 @@ contract LiquidityHub is ILiquidityHub {
     _updateRiskPremiumAndBaseDebt(asset, spoke, riskPremiumRad, -int256(baseDebtRestored));
 
     asset.availableLiquidity += amount;
-    // TODO: decrease base debt on asset/spoke
 
     assetsList[assetId].safeTransferFrom(repayer, address(this), amount);
 


tests/unit/LiquidityHubTest.t.sol
```diff
@@ -213,26 +217,74 @@ contract LiquidityHubTest is BaseTest {
...
+      hub.convertToSharesUp(assetId, amount),
+      'wrong asset suppliedShares post-supply'
+    );
+    assertEq(assetData.availableLiquidity, amount, 'wrong asset availableLiquidity post-supply');
+    assertEq(assetData.baseDebt, 0, 'wrong asset baseDebt post-supply');
+    assertEq(assetData.outstandingPremium, 0, 'wrong asset outstandingPremium post-supply');
+    assertEq(assetData.baseBorrowIndex, WadRayMath.RAY, 'wrong asset baseBorrowIndex post-supply');
+    assertEq(
+      assetData.baseBorrowRate,
+      uint256(500).bpsToRay(),
```
CheyenneAtapour: Should we use same convention as mentioned in previous PRs here, and do `uint256(5_00).bpsToRay()`?

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -198,6 +194,14 @@ contract LiquidityHubTest is BaseTest {
     assertEq(assetData.riskPremiumRad, 0, 'wrong asset riskPremiumRad post-supply');
     assertEq(assetData.lastUpdateTimestamp, 1, 'wrong asset lastUpdateTimestamp post-supply');
     // spoke
+    assertEq(
+      assetData.baseBorrowRate,
+      uint256(500).bpsToRay(),
```
CheyenneAtapour: convention for bps 5_00

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -482,81 +535,249 @@ contract LiquidityHubTest is BaseTest {
...
+      assertEq(assetData.baseDebt, 0, 'wrong asset baseDebt post-supply');
+      assertEq(assetData.outstandingPremium, 0, 'wrong asset outstandingPremium post-supply');
+      assertEq(
+        assetData.baseBorrowIndex,
+        WadRayMath.RAY,
+        'wrong asset baseBorrowIndex post-supply'
+      );
+      assertEq(
+        assetData.baseBorrowRate,
+        uint256(500).bpsToRay(),
```
CheyenneAtapour: bps convention

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -482,81 +535,249 @@ contract LiquidityHubTest is BaseTest {
...
+      assertEq(asset.balanceOf(address(spoke1)), 0, 'wrong spoke token balance post-supply');
+      assertEq(
+        asset.balanceOf(address(hub)),
+        hub.getTotalAssets(assetId),
+        'wrong hub token balance post-supply'
+      );
+      assertEq(asset.balanceOf(USER1), 0, 'wrong user token balance post-supply');
+
+      // time flies
+      uint256 elapsedTime = (i % 2 == 0 ? elapsedTimeChange : elapsedTimeChange * 2) % 30 days; // randomize, 30 days max
```
CheyenneAtapour: Probably fine, but the readability of this is strange. Could just have this as a parameter to this test which gets fuzzed. Or perhaps @DhairyaSethi has a better solution?

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -482,81 +535,249 @@ contract LiquidityHubTest is BaseTest {
...
+      assetData.availableLiquidity,
+      p.totalAssets,
+      'wrong asset availableLiquidity post-supply'
+    );
+    assertEq(assetData.baseDebt, 0, 'wrong asset baseDebt post-supply');
+    assertEq(assetData.outstandingPremium, 0, 'wrong asset outstandingPremium post-supply');
+    assertEq(assetData.baseBorrowIndex, WadRayMath.RAY, 'wrong asset baseBorrowIndex post-supply');
+    assertEq(
+      assetData.baseBorrowRate,
+      uint256(500).bpsToRay(),
```
CheyenneAtapour: bps convention

src/interfaces/IPriceOracle.sol
```diff
@@ -13,4 +13,10 @@ interface IPriceOracle {
    * @return The price of the asset
    */
   function getAssetPrice(uint256 assetId) external view returns (uint256);
+
+  /**
+   * @param assetId The id of the asset
+   * @param price The price of the asset
+   */
+  function setAssetPrice(uint256 assetId, uint256 price) external;
```
DhairyaSethi: let's definitely not add this here, on tests let `oracle` remain a type of `MockPriceOracle` so we can use this
this will not be used in src files

tests/invariant/LiquidityHubHandler.t.sol
```diff
@@ -70,14 +70,14 @@ contract LiquidityHubHandler is Test {
   }
 
   function supply(uint256 assetId, address user, uint256 amount, address onBehalfOf) public {
-    if (user == address(hub) || user == address(0)) return;
-    if (onBehalfOf == address(0)) return;
+    vm.assume(user != address(hub) && user != address(0));
+    vm.assume(onBehalfOf != address(0));
```
DhairyaSethi: can merge assumes into one

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: yeah let's stick with that imo

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -241,8 +293,8 @@ contract LiquidityHubTest is BaseTest {
     uint256 amount,
     address onBehalfOf
   ) public {
-    if (spoke == address(hub) || spoke == address(0)) return;
-    if (onBehalfOf == address(0)) return;
+    vm.assume(spoke != address(hub) && spoke != address(0));
+    vm.assume(onBehalfOf != address(0));
```
DhairyaSethi: merge

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -294,11 +346,14 @@ contract LiquidityHubTest is BaseTest {
     hub.supply(assetId, amount, 0, USER1);
   }
 
-  function test_supply_with_increased_index() public {
+  function skip_test_supply_with_increased_index() public {
```
DhairyaSethi: let's start using `vm.skip` to skip tests, it's more evident how many are pending in the trace and a smaller chance of forgetting to rename tests

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: yeah, would rather just add a pseudo-random `randomizer` helper method in BaseTest to re-use here

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -482,81 +535,249 @@ contract LiquidityHubTest is BaseTest {
...
+
+      // time flies
+      uint256 elapsedTime = (i % 2 == 0 ? elapsedTimeChange : elapsedTimeChange * 2) % 30 days; // randomize, 30 days max
+      skip(elapsedTime);
+
+      p.userShares = 1; // minimum for 1 share
+      p.userAssets = p.userShares.toAssetsUp(hub.getTotalAssets(assetId), assetData.suppliedShares);
+
+      p.totalAssets += p.userAssets;
+      p.suppliedShares += p.userShares;
```
DhairyaSethi, line 700: so userAssets would be incorrect here, no? bc as of now hub.getTotalAssets does not accrue debt (we just advanced in time)
this works here bc there is no debt but good keep in mind bc you were suggesting two methods (getTotalAssets, getTotalAccruedAssets - which i still don't see the reason - miguel also mentioned he doesn't know exactly how integrators will use stale data) so this might get burried
a todo maybe here to note is fine imo, and everywhere else relevant

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -601,41 +849,176 @@ contract LiquidityHubTest is BaseTest {
...
+    assertEq(assetData.baseDebt, 0, 'wrong asset baseDebt post-withdraw');
+    assertEq(assetData.outstandingPremium, 0, 'wrong asset outstandingPremium post-withdraw');
+    assertEq(
+      assetData.baseBorrowIndex,
+      WadRayMath.RAY,
+      'wrong asset baseBorrowIndex post-withdraw'
+    );
+    assertEq(
+      assetData.baseBorrowRate,
+      uint(5_00).bpsToRay(),
```
DhairyaSethi: nit: uint256 acc to style convention

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -601,41 +849,176 @@ contract LiquidityHubTest is BaseTest {
...
+      'wrong asset baseBorrowRate post-withdraw'
+    );
+    assertEq(assetData.riskPremiumRad, 0, 'wrong asset riskPremiumRad post-withdraw');
+    assertEq(assetData.lastUpdateTimestamp, 1, 'wrong asset lastUpdateTimestamp post-withdraw');
```
DhairyaSethi: don't use 1, that's forge thing, cache timestamp at operation with `vm.getBlockTimestamp()`

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -601,41 +849,176 @@ contract LiquidityHubTest is BaseTest {
...
+      assetData.outstandingPremium,
+      'wrong spoke outstandingPremium post-withdraw'
+    );
+    assertEq(
+      spokeData.baseBorrowIndex,
+      assetData.baseBorrowIndex,
+      'wrong spoke baseBorrowIndex post-withdraw'
+    );
+    assertEq(spokeData.riskPremiumRad, 0, 'wrong spoke riskPremiumRad post-withdraw');
+    assertEq(spokeData.lastUpdateTimestamp, 1, 'wrong spoke lastUpdateTimestamp post-withdraw');
```
DhairyaSethi: same

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -601,41 +849,176 @@ contract LiquidityHubTest is BaseTest {
...
+    assertEq(
+      spokeData.suppliedShares,
+      assetData.suppliedShares,
+      'wrong spoke suppliedShares pre-withdraw'
+    );
+    assertEq(spokeData.baseDebt, assetData.baseDebt, 'wrong spoke baseDebt pre-withdraw');
+    assertEq(
+      spokeData.outstandingPremium,
+      assetData.outstandingPremium,
+      'wrong spoke outstandingPremium pre-withdraw'
```
DhairyaSethi: all of these checks seem redundant bc we already tested that in the test above, this test is just testing event emission
we don't need to keep copy pasting all asserts in every unit tests, it sacrifises readability. unit tests should be focussed on what they're trying to test
pls see https://book.getfoundry.sh/tutorials/best-practices#tests

cc @CheyenneAtapour 
lmk your thoughts guys

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -652,34 +1035,70 @@ contract LiquidityHubTest is BaseTest {
...
+    );
+    assertEq(assetData.availableLiquidity, amount, 'wrong asset availableLiquidity pre-withdraw');
+    assertEq(assetData.baseDebt, 0, 'wrong asset baseDebt pre-withdraw');
+    assertEq(assetData.outstandingPremium, 0, 'wrong asset outstandingPremium pre-withdraw');
+    assertEq(assetData.baseBorrowIndex, WadRayMath.RAY, 'wrong asset baseBorrowIndex pre-withdraw');
+    assertEq(
+      assetData.baseBorrowRate,
+      uint(5_00).bpsToRay(),
+      'wrong asset baseBorrowRate pre-withdraw'
+    );
```
DhairyaSethi: imo these tests would be more useful if there was just one open debt position open and then it checks supply shares calculation 
most of the asserts in this test file could be testing important changes in the system, a few that test nothing is incrued when no debt is fine

debt flow is working, how about just initiate one debt position for this asset and then calc the user assets amounts is accurate
it's fine to keep this focussed and do on borrow tests maybe

also same comment as above applies, this can JUST check function reverts when zero are supplied 
this file has a lot of uneccesary volume that could be cleaned up

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -38,7 +38,7 @@ contract LiquidityHubTest is BaseTest {
       spokeConfigs,
       reserveConfigs
     );
-    MockPriceOracle(address(oracle)).setAssetPrice(daiAssetId, 1e8);
```
DhairyaSethi, line 41: let's keep oracle type as `MockPriceOracle` in BaseTest so we don't re-cast, all of IPriceOracle revelant methods are present in that contract

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: since it reverts we know there aren't any state changes

src/interfaces/IPriceOracle.sol
yan-man: ok pls update this issue then: https://github.com/aave/aave-v4/issues/71

tests/unit/LiquidityHubTest.t.sol
yan-man: this i disagree, bc first unit test also tests event emission. And here want to test fuzzed inputs and resultant values. Instead I'll change the name of this test into generic `test_withdraw_fuzz`

tests/unit/LiquidityHubTest.t.sol
yan-man: > debt flow is working

not fully confirmed working yet, so far just supply/withdraw, but will add these debt position tests after draw unit tests are done and debt flow is confirmed

but agree will focus this test on just the revert

tests/unit/SpokeCreditLineTest.t.sol
```diff
@@ -60,10 +60,10 @@ contract SpokeCreditLineTest is BaseTest {
     );
     MockPriceOracle(address(oracle)).setAssetPrice(daiCreditLineAssetId, 1e8);
```
DhairyaSethi: no i meant cast it to mockPriceOracle in BaseTest itself so we don't need to recast here. we'll always use it the setAssetPrice method

yan-man: ah gotcha

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: liquidityhubRiskPremium tests i wrote utilise and rely heavily on the draw flow

yan-man: i'll add it into next pr

Issue #91, opened 2025-01-20
LH RP - restore tests


DhairyaSethi:
depends on #79, #81 and #89

Issue #61, opened 2025-01-11
LH RP - Draw & Restore


CheyenneAtapour:
* Implement Draw and Restore functions on the Liquidity Hub, using our new algorithm for risk premium and keeping base interest accrued separately from risk premium interest accrued
* Ensure that shares and assets math works out appropriately on both Liquidity Hub side and Spoke side
* There is a special requirement that the spoke should pass as parameters the interest being repaid corresponding to base interest and (separately) the interest being repaid corresponding to risk premium interest. Currently the https://github.com/aave/aave-v4/pull/55 branch has this interface

PR #128, opened 2025-01-24
Resolve LH `draw`/`restore`

yan-man:
- refactor LH tests to migrate to baseTest initEnv
- update some natspec
- resolve existing and add new unit/fuzz tests to draw/restore
- update test file structure

closes https://github.com/aave/aave-v4/issues/91
closes https://github.com/aave/aave-v4/issues/89
closes https://github.com/aave/aave-v4/issues/61

### DIFF
--- a/tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
@@ -433,7 +433,7 @@ contract LiquidityHubRestoreTest is LiquidityHubBaseTest {
     );
     uint256 cumulatedBaseDebt = drawAmount.rayMul(cumulatedBaseInterest);
     uint256 accruedPremium = (cumulatedBaseDebt - drawAmount).radMul(riskPremiumRad);
-    vm.assume(accruedPremium > 0);
+    vm.assume(accruedPremium > 0); // accrued premium can round to 0 in edge case - ex. (cumulatedBaseDebt - drawAmount) = 1, riskPremiumRad = 1
 
     // alice restore invalid amount > drawn amount AND premium
     vm.expectRevert(TestErrors.INVALID_RESTORE_AMOUNT);


src/contracts/LiquidityHub.sol
```diff
@@ -215,21 +236,31 @@ contract LiquidityHub is ILiquidityHub {
 
     emit Withdraw(assetId, msg.sender, to, amount);
 
-    return sharesAmount;
+    return (nextBaseBorrowIndex, sharesAmount);
```
DhairyaSethi: it was decided that spoke will query the next borrow index through previewNextBorrow index, since it needs it for calculating its rp (need to accrue debt) so this was to be removed from supply as well

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -271,8 +269,158 @@ contract LiquidityHubTest is BaseTest {
...
+    vm.assume(user != address(hub) && user != address(0) && user != address(spoke1));
+
+    assetId = bound(assetId, 0, hub.assetCount() - 2);
+    amount = bound(amount, 1, type(uint128).max);
+    amount2 = bound(amount2, 1, type(uint128).max);
+
+    IERC20 asset = hub.assetsList(assetId);
+    deal(address(asset), user, amount);
+    vm.prank(user);
+    asset.approve(address(hub), amount);
```
DhairyaSethi: all user's have been dealt and infinite approved hub on all assets in BaseTest setUp to make tests cleaner, these parts can be removed/moved to BaseTest setUp

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -230,36 +176,48 @@ contract LiquidityHubTest is BaseTest {
...
-
-    Asset memory assetData = hub.getAsset(assetId);
-    SpokeData memory spokeData = hub.getSpoke(assetId, address(spoke1));
+    vm.startPrank(spoke);
+    vm.expectEmit(address(asset));
+    emit Transfer(user, address(hub), amount);
+    vm.expectEmit(address(hub));
+    emit Supply(assetId, spoke, amount);
+    hub.supply(assetId, amount, 0, user);
+    vm.stopPrank();
```
DhairyaSethi: the `prank`s here have no effect on the `expectEmit`s, can be removed. all characterestics of an event (emitter, log0,1,2,3, data) are being checked already w/o needing the pranks

yan-man: i can move it down but pranks still needed for the actual hub.supply

tests/unit/LiquidityHubTest.t.sol
yan-man: initial `deal` with a static amount doesn't make as much sense in some of these test situations because there might be fuzzing on supply amounts, etc. that might be greater than the initial `deal` amount so there still might be more dealing necessary. 

also I'm using USER1, USER2 instead of Alice, Bob, etc. 

I'll do an inf approval in the base test setup, but leave the dealing in each test as needed

src/contracts/LiquidityHub.sol
```diff
@@ -149,7 +149,16 @@ contract LiquidityHub is ILiquidityHub {
...
+  /**
+   * @notice Supply asset on behalf of user.
+   * @dev Only callable by spokes.
+   * @param assetId The asset id.
+   * @param amount The amount of asset to withdraw.
+   * @param riskPremiumRad The aggregated risk premium of the calling spoke.
+   * @param supplier The address which is supplying the asset (user).
+   * @return The new base borrow index.
+   * @return The sharesAmount supplied.
+   */
```
CheyenneAtapour: I personally am in favor of no periods because we are mixing phrases in here as well (describing the parameters), which shouldn't have periods. I think it just looks cleaner unless we have multiple sentences, then we can add periods

src/contracts/LiquidityHub.sol
```diff
@@ -218,18 +236,28 @@ contract LiquidityHub is ILiquidityHub {
...
+  /**
+   * @notice Draw debt on behalf of user.
+   * @dev Only callable by spokes.
+   * @param assetId The asset id.
+   * @param to The address to draw debt to (user).
+   * @param amount The amount of debt to draw.
+   * @param riskPremiumRad The aggregated risk premium of the calling spoke.
+   * @return The new base borrow index.
+   * @return The amount of debt drawn.
+   */
```
CheyenneAtapour: Also, these should prob be on interface and we inherit doc here

src/contracts/LiquidityHub.sol
```diff
@@ -241,32 +269,34 @@ contract LiquidityHub is ILiquidityHub {
...
-   * @param riskPremiumRad The aggregated risk premium of the calling spoke
-   * @param repayer The address who is trying to settle the credit line
-   * @return The amount of shares restored
+   * @notice Repays debt on behalf of user.
+   * @dev Only callable by spokes.
+   * @dev Interest is always paid off first from premium, then from base.
+   * @param assetId The asset id.
+   * @param amount The amount to repay.
+   * @param riskPremiumRad The aggregated risk premium of the calling spoke.
+   * @param repayer The address who is trying to settle the credit line.
```
CheyenneAtapour: How about something like "The address repaying the debt"?

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -300,50 +236,162 @@ contract LiquidityHubTest is BaseTest {
...
-    );
+    amount2 = bound(amount2, 1, type(uint128).max);
 
     IERC20 asset = hub.assetsList(assetId);
     deal(address(asset), USER1, amount);
-    vm.prank(USER1);
-    asset.approve(address(hub), amount);
 
-    vm.startPrank(spoke);
+    IERC20 asset2 = hub.assetsList(assetId + 1);
```
CheyenneAtapour: Might be nice to just have a variable to store this asset2's id, to make the subsequent tests look a bit less strange with the `assetId + 1`

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -370,141 +416,150 @@ contract LiquidityHubTest is BaseTest {
...
+      daiData.suppliedShares
     );
-    assertEq(dai.balanceOf(address(spoke1)), 0, 'wrong spoke token balance post-supply');
-    assertEq(dai.balanceOf(address(hub)), amount, 'wrong hub token balance post-supply');
-    assertEq(dai.balanceOf(USER1), 0, 'wrong user token balance post-supply');
+    uint256 initialSupplyShares = daiData.suppliedShares;
 
-    // Time flies, no interest acc
-    skip(1e4);
+    deal(address(tokenList.dai), USER2, supply2Amount);
```
CheyenneAtapour: Btw, we decided to `deal` and `approve` in the base test, so maybe better to move those there if needed

tests/unit/LiquidityHubTest.t.sol
CheyenneAtapour: You can change the deal amount on base test to be sufficient for your use case and that won't affect the other tests

src/contracts/LiquidityHub.sol
```diff
@@ -149,7 +149,16 @@ contract LiquidityHub is ILiquidityHub {
   // Users
   // /////
 
-  /// @dev risk premium is calculated on the spoke and passed upon every action
+  /**
+   * @notice Supply asset on behalf of user.
+   * @dev Only callable by spokes.
+   * @param assetId The asset id.
+   * @param amount The amount of asset to withdraw.
```
miguelmtzinf: supply*

src/contracts/LiquidityHub.sol
```diff
@@ -149,7 +149,16 @@ contract LiquidityHub is ILiquidityHub {
...
+   * @param riskPremiumRad The aggregated risk premium of the calling spoke.
```
miguelmtzinf: better to clarify it's the `new` one?

src/contracts/LiquidityHub.sol
```diff
@@ -149,7 +149,16 @@ contract LiquidityHub is ILiquidityHub {
...
+   * @param supplier The address which is supplying the asset (user).
```
miguelmtzinf: do we wanna say it's the address to pull assets from?

src/contracts/LiquidityHub.sol
```diff
@@ -149,7 +149,16 @@ contract LiquidityHub is ILiquidityHub {
...
+   * @return The new base borrow index.
+   * @return The sharesAmount supplied.
```
```suggestion
   * @return The amount of shares supplied.
```

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: yeah, let's move it down. the prank block inside with expectEmits inside has no effect

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: yeah exactly^

src/contracts/LiquidityHub.sol
DhairyaSethi: it may not repay entire amt

src/contracts/LiquidityHub.sol
```diff
@@ -188,13 +197,22 @@ contract LiquidityHub is ILiquidityHub {
     return (nextBaseBorrowIndex, sharesAmount);
   }
 
-  // TODO: Be able to pass max(uint) as amount to withdraw all or accept number of shares
+  /**
+   * @notice Withdraw supplied asset on behalf of user.
+   * @dev Only callable by spokes.
+   * @param assetId The asset id.
+   * @param to The address to withdraw on behalf of.
```
```suggestion
   * @param to The address to transfer the assets to.
```

src/contracts/LiquidityHub.sol
```diff
@@ -188,13 +197,22 @@ contract LiquidityHub is ILiquidityHub {
...
+   * @param amount The amount of asset to withdraw.
+   * @param riskPremiumRad The aggregated risk premium of the calling spoke.
+   * @return The sharesAmount withdrawn.
+   */
   function withdraw(
     uint256 assetId,
     address to,
```
miguelmtzinf: probably better to move the `to` to the end, so all functions' params share the same order
<assetId, amount, riskPremium, x>

src/contracts/LiquidityHub.sol
miguelmtzinf: i d change to be more generic, something like `the address to pull assets from`

src/contracts/LiquidityHub.sol
```diff
@@ -188,13 +197,22 @@ contract LiquidityHub is ILiquidityHub {
...
 
-  // TODO: Be able to pass max(uint) as amount to withdraw all or accept number of shares
+  /**
+   * @notice Withdraw supplied asset on behalf of user.
+   * @dev Only callable by spokes.
+   * @param assetId The asset id.
+   * @param to The address to withdraw on behalf of.
+   * @param amount The amount of asset to withdraw.
+   * @param riskPremiumRad The aggregated risk premium of the calling spoke.
+   * @return The sharesAmount withdrawn.
```
miguelmtzinf: return the `nextBorrowIndex` too

DhairyaSethi: spoke would have already fetched it using previewBorrowIndex to calculate its rp which was passed here, idt we need to pass again & should rm this addition being added everywhere else

tests/Utils.t.sol
```diff
@@ -30,15 +30,16 @@ library Utils {
     uint256 assetId,
     address spoke,
     uint256 amount,
+    uint256 riskPremiumRad,
     address user,
-    address onBehalfOf
+    address onBehalfOf // todo: implement
```
miguelmtzinf: rename to `to`

tests/BaseTest.t.sol
```diff
@@ -112,6 +112,11 @@ abstract contract BaseTest is Test, Events {
...
   uint256 internal usdxAssetId = 1;
 
+  uint256 internal mintAmount_USDX = 100_000e6;
+  uint256 internal mintAmount_DAI = 100_000e18;
+  uint256 internal mintAmount_WBTC = 100e8;
+  uint256 internal mintAmount_weth = 100e18;
+
   struct TokenList {
     WETH9 weth;
     TestnetERC20 usdx;
```
miguelmtzinf, line 125: should we add a sub-struct here?
```
struct Token { IERC20 token, uint256 id }
struct TokenList {
 Token weth,
 Token usdx,
 Token dai,
 Token wbtc
}
```

or just a new property per token
```
struct TokenList {
 WETH9 weth,
 uint256 wethId,
 TestnetERC20 usdx,
 uint256 usdxId,
 TestnetERC20 dai,
 uint256 daiId,
 TestnetERC20 wbtc,
 uint256 wbtcId
}
```

DhairyaSethi: initially didn't go with approach 1 bc weth's methods are not generic (deposit/withdraw) but i suppose we'd hardly use them in our test suite, `tokenList.weth.token.approve` `tokenList.weth.id` seems fine, but my preference is current approach or approach 2

tests/BaseTest.t.sol
```diff
@@ -112,6 +112,11 @@ abstract contract BaseTest is Test, Events {
   uint256 internal wethAssetId = 0;
   uint256 internal usdxAssetId = 1;
 
+  uint256 internal mintAmount_USDX = 100_000e6;
+  uint256 internal mintAmount_DAI = 100_000e18;
+  uint256 internal mintAmount_WBTC = 100e8;
+  uint256 internal mintAmount_weth = 100e18;
```
DhairyaSethi:
```suggestion
  uint256 internal mintAmount_WETH = 100e18;
```

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: @yan-man i think it's better if we stick with `alice`, `bob`, `carol` over `USER1`, `USER2` etc throughout the testsuite, more readable and less error prone to typos (only 1 char difference bw users) and aligned with v3 testsuite

src/contracts/LiquidityHub.sol
DhairyaSethi: @yan-man this is the case with all other methods as well, pls don't add borrowIndex as a return param there

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1567,4 +3005,54 @@ contract LiquidityHubTest is BaseTest {
     spokeConfig.supplyCap = newSupplyCap;
     hub.updateSpokeConfig(assetId, spoke, spokeConfig);
   }
+
+  /// @dev spoke1 (USER1) supplies dai, spoke2 (USER2) supplies weth, spoke1 (USER1) draws dai
```
DhairyaSethi: it's a dangerous inference from this that each spoke will have will different users, it can be user agonostic or let's have same tests where the same user does all of actions through different spokes

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1567,4 +3005,54 @@ contract LiquidityHubTest is BaseTest {
...
+  function _setUpIncreasedIndex(
```
DhairyaSethi: how about something like
```suggestion
  function _seedHubAndDrawLiqudity(
```
or 
```suggestion
  function _seedDaiWethAndDrawDai(
```

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1085,426 +1527,1431 @@ contract LiquidityHubTest is BaseTest {
...
+      0,
+      'wrong hub dai outstandingPremium post-restore'
+    );
+    assertEq(
+      hubData.daiData.baseBorrowIndex,
+      WadRayMath.RAY,
+      'wrong hub dai baseBorrowIndex post-restore'
+    );
+    assertEq(hubData.daiData.baseBorrowRate, rate, 'wrong hub dai baseBorrowRate post-restore');
+    assertEq(hubData.daiData.riskPremiumRad, 0, 'wrong hub dai riskPremiumRad post-restore');
```
DhairyaSethi: doing this way is risky - putting the data in memory, quering latest data and placing in struct and then asserting against this struct; primarily because these default asserts will be false positives if you forget to load this key of the struct
solidity defaults all values to 0 instead of undefined (like in js)

then you can just do `hub.getAsset(dai).riskPremiumRad` and so on everywhere

DhairyaSethi: takes the same amount of characters roughly

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1085,426 +1527,1431 @@ contract LiquidityHubTest is BaseTest {
...
 
-    // spoke1 supply eth
-    deal(address(eth), USER1, ethAmount);
+    uint256 rate = uint256(15_00).bpsToRay();
+
+    vm.mockCall(
+      address(irStrategy),
+      IReserveInterestRateStrategy.calculateInterestRates.selector,
+      abi.encode(rate)
+    );
```
DhairyaSethi: so i did this in my test to focus more on the risk premium calc and make the test calc easier to reason about, here i don't think we should be mocking but instead of we can fetch the new rate from storage imo

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1085,426 +1527,1431 @@ contract LiquidityHubTest is BaseTest {
...
+      hubData.daiData.lastUpdateTimestamp,
+      'wrong spoke1 dai lastUpdateTimestamp post-restore'
+    );
+    // spoke2 dai
+    assertEq(
+      hubData.spoke2DaiData.suppliedShares,
+      hubData.daiData.suppliedShares,
+      'wrong spoke2 total dai shares post-restore'
+    );
+    assertEq(hubData.spoke2DaiData.baseDebt, 0, 'wrong spoke2 base dai debt post-restore');
```
DhairyaSethi: `wrong` everywhere seems redundant and can be removed imo, we see this message in a revert trace when assertions fail so we know it's wrong

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1085,426 +1527,1431 @@ contract LiquidityHubTest is BaseTest {
...
+
+    skip(skipTime);
+
+    uint256 cumulatedBaseInterest = MathUtils.calculateLinearInterest(
+      rate,
+      uint40(daiData.lastUpdateTimestamp)
+    );
+    uint256 accruedBaseDebt = drawAmount.rayMul(cumulatedBaseInterest) - drawAmount;
+    uint256 accruedPremium = accruedBaseDebt.radMul(riskPremiumRad);
+    restoreAmount = bound(restoreAmount, 0, accruedPremium); // within accrued premium
```
DhairyaSethi:
```suggestion
    restoreAmount = bound(restoreAmount, 1, accruedPremium); // within accrued premium
```
wouldn't call restore with 0 amount

DhairyaSethi: also there should be a check to disallow, since we'll be adding a new method to re-trigger calc and accrue debt outside of this

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1085,426 +1527,1431 @@ contract LiquidityHubTest is BaseTest {
...
+
+    deal(address(tokenList.dai), USER1, restoreAmount);
+
+    vm.startPrank(address(spoke1));
+    hub.restore({assetId: daiAssetId, amount: restoreAmount, riskPremiumRad: 0, repayer: USER1});
+    vm.stopPrank();
```
DhairyaSethi: can just be a prank, same everywhere else relevant

DhairyaSethi: nothing inadvertantly consumes it

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1085,426 +1527,1431 @@ contract LiquidityHubTest is BaseTest {
...
+    uint256 skipTime,
+    uint256 rate,
+    uint256 riskPremiumRad,
+    uint256 restoreAmount
+  ) public {
+    uint256 daiAmount = 100e18;
+    uint256 wethAmount = 10e18;
+
+    drawAmount = bound(drawAmount, 1, 100) * 1e18; // within supplied dai amount
+    skipTime = bound(skipTime, 1, 365 * 10) * 1 days; // 1 day to 10 years
```
DhairyaSethi: if the multiplier (1e18, 1 days) is inside the maxLimit in the bound, we get more granular and interesting bounded fuzz values. this way you destroy the fuzzing engine's ability to find edge cases by rounding it out
ex
fuzz engine wants to try a draw amount 99.999...., but with your bound it's resticted up to zero decimals and can only do 99, same with skip time

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1085,426 +1527,1431 @@ contract LiquidityHubTest is BaseTest {
...
+    rate = bound(rate, 0, 200_00).bpsToRay(); // 0% to 200%
```
DhairyaSethi: same here

tests/unit/LiquidityHubTest.t.sol
DhairyaSethi: and you've anyway stopped the prank so it's not passed to hub.supply

src/contracts/LiquidityHub.sol
yan-man: This one I will leave as we had it in v4 conventions from @miguelmtzinf: https://www.notion.so/V4-Code-Conventions-17a9d63a22de80898fb0ee1f6df6038b?d=17c9d63a22de8082a635001c1cfe10d0&pvs=4#17b9d63a22de8056bdd0fba177d430a3, can update if he agrees

src/contracts/LiquidityHub.sol
yan-man: @DhairyaSethi see other comment: https://github.com/aave/aave-v4/pull/128#discussion_r1932216409

src/contracts/LiquidityHub.sol
yan-man: @DhairyaSethi I am going to leave all the return signatures as is on LH, and then when we figure out spoke side can adjust as needed. Also left a note in the Spoke issues

tests/unit/LiquidityHubTest.t.sol
yan-man: wdym, check to disallow which part?

DhairyaSethi: disallow 0 restore

src/contracts/LiquidityHub.sol
DhairyaSethi: it makes more sense to leave it as is (w/o changes by adding return in this PR) and then add if needed later

tests/BaseTest.t.sol
miguelmtzinf: trying to resolve the identifiers issue.
not a strong preference

yan-man: I'd prob prefer approach 1 or current approach as approach 2 seems a bit messy to me. thoughts @CheyenneAtapour ?

tests/unit/LiquidityHubTest.t.sol
yan-man: can add disallow 0 restore when needed

DhairyaSethi: it's needed now lmao

yan-man: yeah I see

tests/BaseTest.t.sol
```diff
@@ -84,6 +84,8 @@ abstract contract BaseTest is Test, Events {
   using WadRayMath for uint256;
   using SharesMath for uint256;
 
+  uint256 internal constant MAX_SUPPLY_AMOUNT = 1e45; // TODO update if needed when precision issues resolved
```
DhairyaSethi: this is 1e30, 1e45 is max value (amount * price) 
todo is not relevant to this but rather min allowed amount

src/contracts/LiquidityHub.sol
```diff
@@ -278,7 +272,7 @@ contract LiquidityHub is ILiquidityHub {
 
     emit Restore(assetId, msg.sender, amount);
 
-    return amount;
+    return (nextBaseBorrowIndex, amount);
```
DhairyaSethi: let's not add this change here but in the spoke pr as & when needed (it's not since spoke has already prefetched this)

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -86,425 +33,488 @@ contract LiquidityHubTest is BaseTest {
...
     IERC20 asset = hub.assetsList(assetId);
 
-    deal(address(asset), user, amount);
+    vm.startPrank(address(spoke1));
+    vm.expectEmit(address(asset));
+    emit Transfer(alice, address(hub), amount);
+    vm.expectEmit(address(hub));
+    emit Supply(assetId, address(spoke1), amount);
+    hub.supply({assetId: assetId, amount: amount, riskPremiumRad: riskPremiumRad, supplier: alice});
+    vm.stopPrank();
```
DhairyaSethi: pranks not needed on expect emits, can just be one prank above hub.supply, it's cleaner and more apparent where the external call is

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -627,67 +632,49 @@ contract LiquidityHubTest is BaseTest {
     Asset memory assetData;
     SpokeData memory spokeData;
 
+    uint256 runningBalance = asset.balanceOf(alice);
+
     for (uint256 i = 0; i < 5; i++) {
```
DhairyaSethi: instead of 5 let's do hub.assetCount like above

tests/unit/LiquidityHubTest.t.sol
```diff
@@ -1567,4 +2844,52 @@ contract LiquidityHubTest is BaseTest {
     spokeConfig.supplyCap = newSupplyCap;
     hub.updateSpokeConfig(assetId, spoke, spokeConfig);
   }
+
+  /// @dev spoke1 (alice) supplies dai, spoke2 (bob) supplies weth, spoke1 (alice) draws dai
+  function _supplyHubAndDrawLiquidity(
```
DhairyaSethi: `_supplyAndDrawLiquidity` makes more sense imo, wdyt? `_supplyOnHub..` gives more clarity if you prefer this naming

src/contracts/LiquidityHub.sol
```diff
@@ -278,7 +272,7 @@ contract LiquidityHub is ILiquidityHub {
 
     emit Restore(assetId, msg.sender, amount);
 
-    return amount;
+    return nextBaseBorrowIndex;
```
DhairyaSethi: acc to natspec this is amount of debt restored & not nextBorrowIndex, let's stick w that for now

src/contracts/LiquidityHub.sol
yan-man: leaving as is until it is resolved in conventions

tests/unit/LiquidityHubTest.t.sol
yan-man: no, this is not looping over assets, just multiple instances of skip

tests/BaseTest.t.sol
yan-man: created an issue for this until we come to decision: https://github.com/aave/aave-v4/issues/133

tests/unit/LiquidityHubTest.t.sol
yan-man: want to isolate these unit tests away from `ReserveInterestRateStrategy` calcs (esp as that contract will get updated), so want to continue mocking this

tests/unit/LiquidityHubTest.t.sol
yan-man: > if you forget to load this key of the struct

like `hubData.daiData` - then this will have a compilation error rather than default to 0. So disagree about that part as it applies to this test. 

But I do think it's a good general point, there is risk in other places where hubData is re-loaded multiple times in the same test after certain actions so it would be accidentally testing previous values if the call gets re-arranged somehow. So I'll update those to specifically reference hub data loaded at a certain point.

DhairyaSethi: > like hubData.daiData - then this will have a compilation error rather than default to 0. So disagree about that part as it applies to this test.

yeah of course, i meant if you forget to load any particular key, hubData.daiData.riskPremiumRad for example, that'll default to zero

tests/unit/LiquidityHubTest.t.sol
yan-man: I want to keep the accounting assertions separate between users/spokes rather than mixing them. I think important part here is just that spokes differ, more so than users.

tests/BaseTest.t.sol
CheyenneAtapour: Current approach or approach 1 is my vote. Originally I was leaning against the substruct approach in 1, just because I feel it's simpler to have a flatter data structure. But I do see we have a few other tokens that we could consolidate into this tokenList data type, and in that case, the approach 1 would scale better

src/contracts/LiquidityHub.sol
```diff
@@ -265,8 +258,9 @@ contract LiquidityHub is ILiquidityHub {
     Asset storage asset = _assets[assetId];
     SpokeData storage spoke = _spokes[assetId][msg.sender];
 
-    _accrueInterest(asset, spoke); // accrue interest before validating action
-    _validateRestore(asset, amount, spoke.baseDebt);
+    uint256 nextBaseBorrowIndex = _accrueInterest(asset, spoke); // accrue interest before validating action
+
+    _validateRestore(asset, amount, spoke.baseDebt + spoke.outstandingPremium);
```
CheyenneAtapour, line 262: Almost the smallest possible nit: But can we keep the whitespace consistent with the above functions which have the accrue interest line immediately followed by validate restore? I think it also reads nicely that way given the comment about validating the action. However, we can probably get rid of that comment at some point

tests/unit/LiquidityHub/LiquidityHubTest.Supply.t.sol
```diff
@@ -0,0 +1,727 @@
...
+  }
+
+  function test_supply_revertsWith_invalid_shares_amount() public {
+    uint256 assetId = 0;
+    uint256 amount = 1;
+
+    // update storage slots to create 0 shares calc
+    bytes32 baseSlot = keccak256(abi.encode(uint256(assetId), uint256(0))); // key: assetId, slot: 0, ie _assets mapping, dai assetId key
+    vm.store(address(hub), bytes32(uint256(baseSlot) + 1), bytes32(uint256(1))); // suppliedShares slot
+    vm.store(address(hub), bytes32(uint256(baseSlot) + 2), bytes32(uint256(WadRayMath.RAD))); // availableLiquidity slot
```
DhairyaSethi: should never hardcode slots, prefer using [checked_write](https://book.getfoundry.sh/reference/forge-std/std-storage#simple-example), this isn't scalable especially when we optimise the structs and these values may not end up taking a full slot as well. i'd say we can remove this test or if we must keep it, either cause the trigger condition another way or read Asset struct, modify needed values, and checked_write it again 
wdyt @miguelmtzinf?

tests/unit/LiquidityHub/LiquidityHubAccrueInterestTest.t.sol
```diff
@@ -0,0 +1,252 @@
...
+    assertEq(daiInfo.riskPremiumRad, riskPremium);
+    assertEq(daiInfo.outstandingPremium, (totalBase - borrowAmount).radMul(riskPremium));
+  }
+
+  function test_accrueInterest_fuzz_ChangingBorrowRate(
+    uint256 borrowAmount,
+    uint40 elapsed,
+    uint256 riskPremium
+  ) public {
+    elapsed = uint40(bound(elapsed, 1, type(uint40).max / 3));
```
miguelmtzinf, line 184: why this restriction only here?

tests/unit/LiquidityHub/LiquidityHubInterestRateTest.t.sol
```diff
@@ -0,0 +1,252 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/BaseTest.t.sol';
+import {SpokeData} from 'src/contracts/LiquidityHub.sol';
+import {Asset} from 'src/contracts/LiquidityHub.sol';
+import {Utils} from 'tests/Utils.t.sol';
+
+contract LiquidityHubAccrueInterestTest is BaseTest {
```
miguelmtzinf: there is something weird here. shouldn't this be `LiquidityHubInterestRateTest`?

yan-man: typo, will update

tests/unit/LiquidityHub/LiquidityHubAccrueInterestTest.t.sol
yan-man: migrated from @CheyenneAtapour test

tests/unit/LiquidityHub/LiquidityHubInterestRateTest.t.sol
```diff
@@ -6,7 +6,7 @@ import {SpokeData} from 'src/contracts/LiquidityHub.sol';
 import {Asset} from 'src/contracts/LiquidityHub.sol';
 import {Utils} from 'tests/Utils.t.sol';
 
-contract LiquidityHubAccrueInterestTest is BaseTest {
+contract LiquidityHubInterestRateTest is BaseTest {
```
miguelmtzinf: not only a name change. it seems it's duplicated code (same as  `LiquidityHubAccrueInterestTest`)

yan-man: this is my fault when migrating Cheyenne's tests over. There were 2 contracts in 1 file and I split it into 2 files and duplicated instead of moving the tests properly

fixed

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
```diff
@@ -0,0 +1,939 @@
...
+    });
+
+    // alice restore invalid amount > drawn amount AND premium
+    vm.expectRevert(TestErrors.INVALID_RESTORE_AMOUNT);
+
+    vm.prank(address(spoke1));
+    hub.restore({assetId: daiAssetId, amount: drawAmount + 1, riskPremiumRad: 0, repayer: alice});
+  }
+
+  function test_restore_revertsWith_invalid_restore_amount0() public {
```
miguelmtzinf: should we use numbers in words rather than digits always?

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
```diff
@@ -0,0 +1,939 @@
...
+      riskPremiumRad: uint256(5_00).bpsToRad(),
+      user: bob,
+      to: address(spoke2)
+    });
+
+    uint256 cumulatedBaseInterest = MathUtils.calculateLinearInterest(
+      rate,
+      uint40(spoke1DaiData.lastUpdateTimestamp)
+    );
+    uint256 cumulatedBaseDebt = drawAmount.rayMul(cumulatedBaseInterest);
```
miguelmtzinf, line 264: should we validate that effectively `premium > 0` and `cumulatedDebt > 0`? same in other fuzz tests above

yan-man: i think this one just needs validation on cumulated debt (and premium 0). But yes for premium test can also validate premium is > 0.

in this case we can't use bound or vm assume so will short circuit. Unless there's a different suggestion cc @DhairyaSethi

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
yan-man: updated and added to conventions doc

LiquidityHubInterestRateTest.t.sol
```diff
@@ -0,0 +1,524 @@
...
+    hub.draw(daiAssetId, address(spoke3), drawSpoke3, rpSpoke3);
+    vm.stopPrank();
+
+    uint256 borrowRate = _getBorrowRate(daiAssetId);
+    uint256 baseBorrowRate = _getBaseBorrowRate(daiAssetId);
+    uint256 newRp = hub.getAsset(daiAssetId).riskPremiumRad;
+    assertEq(borrowRate, baseBorrowRate + (newRp.radToRay().rayMul(baseBorrowRate)));
+  }
+
+  // TODO: Test via calling functions on spokes - after spoke side is implemented
```
miguelmtzinf: should this be placed in another file like `SpokeInterestRateTest`? even if IR getter is in LH

LiquidityHubInterestRateTest.t.sol
```diff
@@ -0,0 +1,524 @@
...
+  function _getBaseBorrowRate(uint256 assetId) internal view returns (uint256) {
+    return hub.getBaseInterestRate(assetId);
+  }
+
+  function _getBorrowRate(uint256 assetId) internal view returns (uint256) {
+    return hub.getInterestRate(assetId);
+  }
+}
+
+contract LiquidityHubAccrueInterestTest is BaseTest {
```
miguelmtzinf: remove

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
yan-man: hm well actually the test should still fail in the same way even if cumulated debt is 0

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
```diff
@@ -0,0 +1,939 @@
...
+    skip(skipTime);
+
+    uint256 cumulatedBaseInterest = MathUtils.calculateLinearInterest(
+      rate,
+      uint40(daiData.lastUpdateTimestamp)
+    );
+    uint256 accruedBaseDebt = drawAmount.rayMul(cumulatedBaseInterest) - drawAmount;
+    uint256 accruedPremium = accruedBaseDebt.radMul(riskPremiumRad);
+
+    if (accruedPremium == 0) {
```
miguelmtzinf: should we change it to an assert an enforce `riskPremiumRad > 0` ?

yan-man: yes that would help to bound RP > 0 (has been updated now). But I think even with that bound, there could be small enough other fuzzed values to still end up with 0 accrued premium. 

If we assert, the test will fail in that case. I don't think that is desired right?

tests/unit/LiquidityHub/LiquidityHubTest.Supply.t.sol
```diff
@@ -0,0 +1,727 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import './LiquidityHubBaseTest.t.sol';
+import {IERC20Errors} from 'src/dependencies/openzeppelin/IERC20Errors.sol';
+import {Asset, SpokeData} from 'src/contracts/LiquidityHub.sol';
+
+contract LiquidityHubSupplyTest is LiquidityHubBaseTest {
+  using SharesMath for uint256;
+  using WadRayMath for uint256;
```
miguelmtzinf, line 10: are these still needed? they are in `LiquidityHubBaseTest` already

tests/unit/LiquidityHub/LiquidityHubTest.Supply.t.sol
miguelmtzinf: 100%
we can just inflate the exchange rate so 1 asset is 0 shares for example - realistic

tests/unit/LiquidityHub/LiquidityHubTest.Supply.t.sol
```diff
@@ -0,0 +1,727 @@
...
+    uint256 amount = 100e18;
+
+    _updateActive(daiAssetId, false);
+
+    vm.prank(address(spoke1));
+    vm.expectRevert(TestErrors.ASSET_NOT_ACTIVE);
+    hub.supply(daiAssetId, amount, 0, alice);
+  }
+
+  function test_supply_revertsWith_supply_cap_exceeded() public {
```
miguelmtzinf, line 38: missing the case where supply cap is exceeded due to interest (same for draw)

tests/unit/LiquidityHub/LiquidityHubTest.Withdraw.t.sol
```diff
@@ -0,0 +1,727 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import './LiquidityHubBaseTest.t.sol';
+import {IERC20Errors} from 'src/dependencies/openzeppelin/IERC20Errors.sol';
+import {Asset, SpokeData} from 'src/contracts/LiquidityHub.sol';
+
+contract LiquidityHubWithdrawTest is LiquidityHubBaseTest {
+  using SharesMath for uint256;
+  using WadRayMath for uint256;
```
miguelmtzinf, line 10: same

tests/unit/LiquidityHub/LiquidityHubAccrueInterestTest.t.sol
CheyenneAtapour: I did it like this just because time is skipped twice

tests/unit/LiquidityHub/LiquidityHubTest.Supply.t.sol
yan-man: although it's inheriting, getting errors if these are removed from the child contracts

tests/unit/LiquidityHub/LiquidityHubTest.Supply.t.sol
yan-man: added

tests/unit/LiquidityHub/LiquidityHubTest.Supply.t.sol
yan-man: updated

tests/unit/LiquidityHub/LiquidityHubTest.Supply.t.sol
DhairyaSethi: lib casting doesn't fall down the inheritance chain, will have to add again

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
```diff
@@ -0,0 +1,944 @@
...
+    uint256 drawAmount,
+    uint256 skipTime,
+    uint256 rate
+  ) public {
+    uint256 daiAmount = 100e18;
+    uint256 wethAmount = 10e18;
+
+    drawAmount = bound(drawAmount, 1, daiAmount); // within supplied dai amount
+    skipTime = bound(skipTime, 1, 365 * 10 * 1 days); // 1 sec to 10 years
+    rate = bound(rate, 0, 200_00).bpsToRay(); // .1% to 200%
```
DhairyaSethi: isn't this 0% to 200%

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
```diff
@@ -0,0 +1,944 @@
...
+      rate,
+      uint40(spoke1DaiData.lastUpdateTimestamp)
+    );
+    uint256 cumulatedBaseDebt = drawAmount.rayMul(cumulatedBaseInterest);
+
+    // alice restore invalid amount > drawn amount AND premium
+    vm.expectRevert(TestErrors.INVALID_RESTORE_AMOUNT);
+
+    vm.prank(address(spoke1));
+    hub.restore({
```
DhairyaSethi, line 271: another style guideline suggestion, if reverting, always put expectRevert before the call
prank does not get consumed on with an expect revert (unless there's an external view call, in which case consider using startPrank)
```solidity
    // alice restore invalid amount > drawn amount AND premium
    vm.prank(address(spoke1));
    vm.expectRevert(TestErrors.INVALID_RESTORE_AMOUNT);
    hub.restore({ .. });
```

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
DhairyaSethi: yeah, pls no case **ever** exists to short circuit a test

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
miguelmtzinf: yeah, hesitant on this one
having external calls in the `expectRevert` can consume the single prank and this goes easily unnoticed

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
yan-man: updated rate bounds

miguelmtzinf: maybe better to fuzz between 0 and the maximum (iirc 1000%)?

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
yan-man: pls add to conventions doc to discuss there 

added

DhairyaSethi:
@yan-man could we rename test files to `LiquidityHub.<action>.t.sol` instead of `LiquidityHubTest.<action>.t.sol` as [originally proposed](https://avaraverse.slack.com/archives/C04HCBMEEBZ/p1738061958756739)? 
also would prefer folder names to not need caps so either have `test/unit/hub/LiquidityHub.<action>.t.sol` or `test/unit/liquidityHub/LiquidityHub.<action>.t.sol`, no strong opinion

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
yan-man: nvm, I was wrong, will use vm assume here. Thanks @DhairyaSethi

yan-man:
> @yan-man could we rename test files to `LiquidityHub.<action>.t.sol` instead of `LiquidityHubTest.<action>.t.sol` as [originally proposed](https://avaraverse.slack.com/archives/C04HCBMEEBZ/p1738061958756739)? also would prefer folder names to not need caps so either have `test/unit/hub/LiquidityHub.<action>.t.sol` or `test/unit/liquidityHub/LiquidityHub.<action>.t.sol`, no strong opinion

disagree, all other tests have `Test` in the file name besides the lib ones. Should also be discussed in conventions, and have consistency

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
DhairyaSethi: > change it to an assert

better to vm assume

tests/unit/LiquidityHub/LiquidityHubTest.Restore.t.sol
```diff
@@ -433,7 +433,7 @@ contract LiquidityHubRestoreTest is LiquidityHubBaseTest {
     );
     uint256 cumulatedBaseDebt = drawAmount.rayMul(cumulatedBaseInterest);
     uint256 accruedPremium = (cumulatedBaseDebt - drawAmount).radMul(riskPremiumRad);
-    vm.assume(accruedPremium > 0 && cumulatedBaseDebt > 0);
+    vm.assume(accruedPremium > 0);
```
miguelmtzinf: there will be always accruedPremium no?

yan-man: no I think with fuzzed inputs low enough, with rounding it can end up with accruedPremium = 0

ex. (cumulatedBaseDebt - drawAmount) = 1, riskPremiumRad = 1, uint256(1).radMul(uint256(1)) ends up 0

miguelmtzinf: ah, good catch! can you write a comment there pls?

yan-man: added

PR #146, opened 2025-02-06
Test Spoke User RP

### LINKED ISSUE #143: Spoke RP - User Risk Premium Unit Tests

CheyenneAtapour:
Adds unit and fuzz tests for the Spoke side calculation of User Risk Premium

Also includes some basic wip tests for the following tests (scope reduced for this PR because it was getting long, the remainder of following tests will come in subsequent PR):
* Interest accrual
* Weighted Average Spoke Risk Premium

Closes
#143

### DIFF
--- a/tests/unit/Spoke/Spoke.Repay.t.sol
+++ b/tests/unit/Spoke/Spoke.Repay.t.sol
@@ -6,12 +6,6 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
 contract SpokeRepayTest is SpokeBase {
   using PercentageMath for uint256;
 
-  function setUp() public override {
-    super.setUp();
-    // Set Weth LP to nonzero
-    updateLiquidityPremium(spoke1, wethReserveId(spoke1), 10_00);
-  }
-
   function test_repay_same_block() public {
     uint256 daiSupplyAmount = 100e18;
     uint256 wethSupplyAmount = 10e18;
--- a/tests/unit/Spoke/Spoke.Supply.t.sol
+++ b/tests/unit/Spoke/Spoke.Supply.t.sol
@@ -6,12 +6,6 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
 contract SpokeSupplyTest is SpokeBase {
   using WadRayMath for uint256;
 
-  function setUp() public override {
-    super.setUp();
-    // Set Weth LP to nonzero
-    updateLiquidityPremium(spoke1, wethReserveId(spoke1), 10_00);
-  }
-
   function test_supply_revertsWith_ReserveNotListed() public {
     uint256 reserveId = spoke1.reserveCount() + 1; // invalid reserveId
     uint256 amount = 100e18;
--- a/tests/unit/Spoke/Spoke.Withdraw.t.sol
+++ b/tests/unit/Spoke/Spoke.Withdraw.t.sol
@@ -7,12 +7,6 @@ contract SpokeWithdrawTest is SpokeBase {
   using WadRayMath for uint256;
   using PercentageMath for uint256;
 
-  function setUp() public override {
-    super.setUp();
-    // Set Weth LP to nonzero
-    updateLiquidityPremium(spoke1, wethReserveId(spoke1), 15_00);
-  }
-
   function test_withdraw_revertsWith_ReserveNotActive() public {
     uint256 daiReserveId = daiReserveId(spoke1);
     uint256 amount = 100e18;


src/contracts/LiquidityHub.sol
```diff
@@ -430,6 +430,9 @@ contract LiquidityHub is ILiquidityHub {
 
   function _addSpoke(uint256 assetId, DataTypes.SpokeConfig memory config, address spoke) internal {
     require(spoke != address(0), 'INVALID_SPOKE');
+    uint256 currentAssetBaseBorrowIndex = _assets[assetId].previewNextBorrowIndex();
+    // is zero when asset.baseBorrowIndex == 0 (ie asset not registered)
+    require(currentAssetBaseBorrowIndex != 0, 'INVALID_ASSET');
```
DhairyaSethi: cleaner check would be `_assets[assetId].{baseBorrowIndex,lastUpdateTimestamp} != 0`
especially since index `currentAssetBaseBorrowIndex` isn't used

src/contracts/Spoke.sol
```diff
@@ -38,6 +39,11 @@ contract Spoke is ISpoke {
     bool collateral;
   }
 
+  struct ReservePremium {
+    uint256 reserveId;
+    uint256 liquidityPremium;
+  }
```
```suggestion
  struct ReservePremium {
    uint128 reserveId;
    uint128 liquidityPremium;
  }
```
we're packing them like this below, better to pack here even tho this is tmp

src/contracts/Spoke.sol
```diff
@@ -393,41 +398,59 @@ contract Spoke is ISpoke {
...
+    // TODO: Optimize this
+    // Sort array of user collaterals by liquidity premium
+    Arrays.sort(packedUserCollaterals, _compareLp);
+
     // While the tempDebt variable is non-zero, loop over collateral reserves, adding up weighted risk premium, and subtract corresponding amt from tempDebt
-    for (uint256 i; i < reservesListLength; ++i) {
-      reserveId = reservesList[i];
-      if (!_usingAsCollateral(userData[reserveId])) continue;
+    for (uint256 i; i < userCollateralAssets; ++i) {
+      reserveId = packedUserCollaterals[i] >> 128;
```
DhairyaSethi: can extract liq premium from here(`packedUserCollaterals`) instead of fetching from storage again below

tests/BaseTest.t.sol
```diff
@@ -322,10 +322,10 @@ abstract contract BaseTest is Test, Events {
     );
     spokeInfo[spoke1].usdx.liquidityPremium = usdxConfig.liquidityPremium;
 
+    hub.addSpoke(usdxAssetId, spokeConfig, address(spoke1));
```
DhairyaSethi: imo it's easier to read assets are added in the same order as setup above and we rather update the individual values by having wbtc premium higher to achieve out of order liq premium asset list

tests/unit/Spoke/SpokeAccrueInterestTest.t.sol
```diff
@@ -44,4 +45,121 @@ contract SpokeAccrueInterestTest is BaseTest {
...
+    uint256 baseBorrowRate = hub.getBaseInterestRate(daiAssetId);
+
+    // Time passes
+    skip(elapsed);
+
+    // Bob does a supply through same spoke to accrue interest
+    Utils.spokeSupply(spoke1, spokeInfo[spoke1].dai.reserveId, bob, 1e18, bob);
+
+    Spoke.Reserve memory daiInfo = spoke1.getReserve(spokeInfo[spoke1].dai.reserveId);
+    Asset memory daiAssetInfo = hub.getAsset(daiAssetId);
```
DhairyaSethi: pls don't use these getters as discussed today, let's use specific exposed getters for debt & rp

src/contracts/Spoke.sol
```diff
@@ -393,41 +398,59 @@ contract Spoke is ISpoke {
...
     for (uint256 i; i < reservesListLength; ++i) {
       reserveId = reservesList[i];
-      tempDebt += userData[reserveId].baseDebt * IPriceOracle(oracle).getAssetPrice(reserveId);
+      tempDebt +=
+        userData[reserveId].baseDebt *
+        IPriceOracle(oracle).getAssetPrice(_reserves[reserveId].assetId);
+
+      if (_usingAsCollateral(userData[reserveId])) {
+        // Pack (reserveId, liquidityPremium) into a single uint256
+        userCollaterals[userCollateralAssets++] =
```
DhairyaSethi: imo packing/unpacking logic can be abstracted even tho this is tmp

src/contracts/Spoke.sol
DhairyaSethi: actually this is to be removed since we manually pack

src/contracts/LiquidityHub.sol
CheyenneAtapour: what if we add a new asset at the same block as a new spoke?

tests/BaseTest.t.sol
CheyenneAtapour: I now used 1 uniform order for these assets throughout this function

tests/unit/Spoke/SpokeUserRiskPremiumTest.t.sol
```diff
@@ -0,0 +1,1032 @@
...
+  }
+
+  function test_getUserRiskPremium_multi_asset_collateral_weth_partial_cover() public {
+    TestInfo memory params;
+    params.daiReserveId = spokeInfo[spoke1].dai.reserveId;
+    params.usdxReserveId = spokeInfo[spoke1].usdx.reserveId;
+    params.wethReserveId = spokeInfo[spoke1].weth.reserveId;
+
+    params.daiSupplyAmount = 2000e18;
+    params.usdxSupplyAmount = 2000e18;
```
DhairyaSethi: decimal normalisation is missing on spoke, once fixed, pls refactor to make sure decimal representation in tests is accurate (here usdx has 6 decimals, rather than 18)

src/contracts/LiquidityHub.sol
DhairyaSethi: asset's index default value is 1 Ray, lastUpdateTimestamp is current timestamp, they'd be non zero even if spoke's added on the same block. not sure if i'm missing something?

src/contracts/Spoke.sol
```diff
@@ -421,50 +429,69 @@ contract Spoke is ISpoke {
...
     uint256 userSupply;
 
-    // Add up user debt for each reserve, including price
+    // Add up user debt for each reserve, including price. Store user collaterals
     for (uint256 i; i < reservesListLength; ++i) {
       reserveId = reservesList[i];
-      tempDebt += userData[reserveId].baseDebt * IPriceOracle(oracle).getAssetPrice(reserveId);
+      tempDebt +=
+        userData[reserveId].baseDebt *
+        IPriceOracle(oracle).getAssetPrice(_reserves[reserveId].assetId);
```
yan-man: if oracles are defined in each spoke, should we instead change the oracle to accept reserveId instead of assetId?

src/contracts/Spoke.sol
```diff
@@ -421,50 +429,69 @@ contract Spoke is ISpoke {
...
-  ) internal returns (uint256) {
+  ) internal view returns (uint256) {
     uint256 reservesListLength = reservesList.length;
+    uint256[] memory userCollaterals = new uint256[](reservesListLength);
 
     // Variable to decrement as we count up user RP
     uint256 tempDebt = 0;
     uint256 newUserRiskPremium = 0;
     uint256 collateralValue = 0;
+    uint256 userCollateralAssets = 0;
```
yan-man: can we rename to something that connotes count, like `numUserCollateralAssets`

src/contracts/Spoke.sol
CheyenneAtapour: I think it is easier to re-use oracles if they have a consistent id system, which the liquidity hub's `assetId` system is. Otherwise, we would have to potentially deploy more oracles for the more spokes we have. For example, 2 spokes might have a diff `reserveId` for usdc, but if we use `assetId` for both, we can re-use the oracle

src/contracts/Spoke.sol
```diff
@@ -421,50 +429,69 @@ contract Spoke is ISpoke {
...
-    for (uint256 i; i < reservesListLength; ++i) {
-      reserveId = reservesList[i];
-      if (!_usingAsCollateral(userData[reserveId])) continue;
+    for (uint256 i; i < userCollateralAssets; ++i) {
+      reserveId = _unpackReserveId(packedUserCollaterals[i]);
 
       // Convert user's supply shares for this reserve to collateral value
       userSupply =
         liquidityHub.convertToAssetsDown(
           _reserves[reserveId].assetId,
```
yan-man: cache `_reserves[reserveId].assetId` to be re-used

tests/unit/Spoke/SpokeAccrueInterestTest.t.sol
```diff
@@ -17,31 +18,171 @@ contract SpokeAccrueInterestTest is BaseTest {
...
+    Utils.spokeSupply(spoke1, daiReserveId, bob, amount * 2, bob);
+    Utils.spokeBorrow(spoke1, daiReserveId, bob, amount, bob);
+
+    uint256 baseBorrowRate = hub.getBaseInterestRate(daiAssetId);
+
+    // 1 year passes
+    skip(365 days);
+
+    // Bob does a supply through same spoke to accrue interest
+    Utils.spokeSupply(spoke1, daiReserveId, bob, 1e18, bob);
```
yan-man: can we add another test except without this trivial supply action? Every assert should still hold now with accrued getters except for `getAssetLastUpdate` I think?

tests/unit/Spoke/SpokeUserRiskPremiumTest.t.sol
```diff
@@ -0,0 +1,1032 @@
...
+      (params.daiSupplyAmount *
+        oracle.getAssetPrice(daiAssetId) +
+        usdxLpContributionAmount *
+        oracle.getAssetPrice(usdxAssetId));
+
+    assertEq(spoke1.getUserRiskPremium(bob), expectedUserRiskPremium, 'user risk premium');
+  }
+
+  /// @dev Supply with a high value lp asset so it's ignored in rp calcs, but allows user to borrow large amounts.
+  /// @dev Borrow with any asset because only it's value is important, it's lp is ignored.
```
yan-man:
```suggestion
  /// @dev Borrow with any asset because only its value is important, its lp is ignored.
```

tests/unit/HealthFactorTest.t.sol
yan-man, line 1: rename test file names (`HealthFactor.t.sol`) as per: https://www.notion.so/V4-Code-Conventions-17a9d63a22de80898fb0ee1f6df6038b?d=19e9d63a22de8055b310001c8a67b735&pvs=4#18a9d63a22de8001a3d2d542d8da828b

tests/unit/Spoke/Spoke.Supply.t.sol
```diff
@@ -458,15 +458,15 @@ contract SpokeSupplyTest is SpokeBase {
...
 
     // alice supplies usdx as collateral, borrows dai
     _executeSpokeSupplyAndBorrow({
       spoke: spoke1,
       collateral: TestReserve({
-        reserveId: wethReserveId(spoke1),
+        reserveId: usdxReserveId(spoke1),
         supplier: alice,
-        supplyAmount: 100e18,
+        supplyAmount: 100e8,
```
DhairyaSethi: usdx has 6 decimals, wbtc has 8

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+    uint256 startTime;
+    uint256 reserveDebt;
+    uint256 reservePremium;
+    uint256 spokeDebt;
+    uint256 spokePremium;
+    uint256 assetDebt;
+    uint256 assetPremium;
+  }
+
+  struct ReserveLP {
```
DhairyaSethi: unused? 
imo wrt sorting we should not use lib but instead pack in this struct & write a naive but correct sorting algo in the tests

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+    setUsingAsCollateral(spoke1, bob, params.daiReserveId, true);
+    Utils.spokeBorrow(spoke1, params.daiReserveId, bob, params.borrowAmount, bob);
+
+    if (_normalizedValue(params.borrowAmount, daiAssetId) > 0) {
+      // With single collateral, user rp will match liquidity premium of collateral
+      assertEq(spoke1.getUserRiskPremium(bob), params.daiLP, 'user risk premium');
+    } else {
+      // Dust amount borrowed, no risk premium
+      assertEq(spoke1.getUserRiskPremium(bob), 0, 'user risk premium');
+    }
```
DhairyaSethi: this is no longer the case, right? since min bound is 1, it should always count in the rp

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+
+    uint256 userRiskPremium = spoke1.getUserRiskPremium(bob);
+
+    if (_normalizedValue(params.borrowAmount, daiAssetId) > 0) {
+      // With single collateral, user rp will match liquidity premium of collateral
+      assertEq(userRiskPremium, params.daiLP, 'user risk premium');
+    } else {
+      // Dust amount borrowed, no risk premium
+      assertEq(userRiskPremium, 0, 'user risk premium');
+    }
```
DhairyaSethi: same

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+
+    // Bob supply weth into spoke1
+    Utils.spokeSupply(spoke1, params.wethReserveId, bob, params.wethSupplyAmount, bob);
+    setUsingAsCollateral(spoke1, bob, params.wethReserveId, true);
+
+    // Bob draw 2000 total dai + usdx
+    Utils.spokeBorrow(spoke1, params.daiReserveId, bob, params.daiSupplyAmount, bob);
+    Utils.spokeBorrow(spoke1, params.usdxReserveId, bob, params.usdxSupplyAmount, bob);
+
+    DataTypes.UserPosition memory userPosition = spoke1.getUserPosition(params.daiReserveId, bob);
```
DhairyaSethi: thought we weren't going to use getters for stored values? for ex below we use the internal helper `getUserInfo`

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+      setUsingAsCollateral(spoke3, bob, params.usdxReserveId, true);
+    }
+
+    // Bob supply weth into spoke3
+    Utils.spokeSupply(spoke3, params.wethReserveId, bob, params.wethSupplyAmount, bob);
+    setUsingAsCollateral(spoke3, bob, params.wethReserveId, true);
+
+    // Bob draw weth
+    if (params.wethBorrowAmount > 0) {
+      Utils.spokeBorrow(spoke3, params.wethReserveId, bob, params.wethBorrowAmount, bob);
```
DhairyaSethi: should we add an assume that supplied value > borrowed value? otherwise this will fail once HF checks get merged in

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+
+    // Bob draw dai2
+    if (params.dai2BorrowAmount > 0) {
+      Utils.spokeBorrow(spoke2, params.dai2ReserveId, bob, params.dai2BorrowAmount, bob);
+    }
+
+    uint256 expectedUserRiskPremium = _calculateExpectedUserRP(bob, spoke2);
+
+    // wbtc, weth, dai, and usdx will each cover part of the debt
+    assertEq(spoke2.getUserRiskPremium(bob), expectedUserRiskPremium, 'user risk premium');
```
DhairyaSethi:
```suggestion
    // wbtc, weth, dai, and usdx will each cover part of the debt
    assertEq(spoke2.getUserRiskPremium(bob), _calculateExpectedUserRP(bob, spoke2), 'user risk premium');
```

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1305 @@
...
+
+  function _calculateExpectedUserRP(address user, ISpoke spoke) internal view returns (uint256) {
+    uint256 assetId;
+    uint256 totalDebt;
+    uint256 suppliedReservesCount;
+    uint256 userRP;
+    DataTypes.UserPosition memory userPosition;
+
+    // Find all reserves user has supplied, adding up total debt
+    for (uint256 i; i < spoke.reserveCount(); ++i) {
```
DhairyaSethi: can we do this just for additional clarity?
```suggestion
    for (uint256 reserveId; reserveId < spoke.reserveCount(); ++reserveId) {
```

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
DhairyaSethi: same for all instances below with `spoke.getUserPosition`

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: The weth supplied amount here is always `MAX_SUPPLY_AMOUNT` and the weth borrowed amount is at most half of that, so should be fine imo

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: it's used in my calculate function below

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: yep, that was a remnant of the past that got overlooked

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
DhairyaSethi: this struct is not i think? could you link where exactly

tests/Base.t.sol
```diff
@@ -565,16 +565,35 @@ abstract contract Base is Test {
...
+    (asset.baseDebt, asset.outstandingPremium) = hub.getAssetDebt(assetId);
+    asset.baseBorrowIndex = hub.getAsset(assetId).baseBorrowIndex;
+    asset.baseBorrowRate = hub.getBaseInterestRate(assetId);
+    asset.riskPremium = hub.getAssetRiskPremium(assetId);
+    asset.lastUpdateTimestamp = hub.getAsset(assetId).lastUpdateTimestamp;
+    asset.config = hub.getAssetConfig(assetId);
+    return asset;
+  }
+
+  function getAsset(ISpoke spoke, uint256 reserveId) internal view returns (uint256, IERC20) {
```
yan-man: can you rename this something else? purposefully didn't want this as `getAsset` bc it conflicts with the LH method

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+  }
+
+  struct ReserveLP {
+    uint256 reserveId;
+    uint256 lp;
+  }
+
+  function setUp() public override {
+    super.setUp();
+    super.initEnvironment();
```
yan-man: can remove super

tests/unit/Spoke/SpokeAccrueInterest.t.sol
yan-man, line 1: `MAX_BPS` var seems unused 

can also inherit from SpokeBase and then remove the setUp

is the main pt of this PR the RP tests rather than the accrue interest tests? I think could also link it to an issue for accrue, or remove accrue tests from this PR

tests/unit/Spoke/SpokeAccrueInterest.t.sol
```diff
@@ -26,22 +27,132 @@ contract SpokeAccrueInterestTest is Base {
...
 
     // Timestamp doesn't update when no interest accrued
-    assertEq(daiInfo.lastUpdateTimestamp, vm.getBlockTimestamp(), 'lastUpdateTimestamp');
+    assertEq(daiInfo.lastUpdateTimestamp, lastUpdate, 'lastUpdateTimestamp');
     assertEq(daiInfo.baseDebt, 0, 'baseDebt');
-    assertEq(daiInfo.riskPremium, 0, 'riskPremium');
     assertEq(daiInfo.outstandingPremium, 0, 'outstandingPremium');
   }
+
+  function test_accrueInterest_BorrowAndWait() public {
```
yan-man, line 44: i think can also get versions of these with premium > 0. 

why is premium 0 here actually? shouldnt bob have a user RP. Is it bc he hasnt set dai as collateral

tests/unit/Spoke/SpokeAccrueInterest.t.sol
```diff
@@ -26,22 +27,132 @@ contract SpokeAccrueInterestTest is Base {
...
+    assertEq(daiReserveInfo.outstandingPremium, 0, 'outstandingPremium');
+
+    // LH checks
+    assertEq(daiAssetInfo.baseDebt, totalBase, 'asset base debt');
+    assertEq(daiAssetInfo.riskPremium, 0);
+    assertEq(daiAssetInfo.outstandingPremium, 0);
+    assertEq(daiAssetInfo.lastUpdateTimestamp, lastUpdate);
+  }
+
+  function test_accrueInterest_fuzz_BorrowAndWait(uint40 elapsed) public {
```
yan-man: nbd but don't think we need this one given you have the next one with both inputs fuzzed

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+  }
+
+  function test_getUserRiskPremium_no_collateral() public view {
+    uint256 userRiskPremium = spoke1.getUserRiskPremium(bob);
+    assertEq(userRiskPremium, 0, 'user risk premium');
```
yan-man: can we add an assertion for initial condition, ie no collateral

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+
+    // Bob supply usdx into spoke1
+    Utils.spokeSupply(spoke1, params.usdxReserveId, bob, params.usdxSupplyAmount, bob);
+    setUsingAsCollateral(spoke1, bob, params.usdxReserveId, true);
+
+    // Bob supply weth into spoke1
+    Utils.spokeSupply(spoke1, params.wethReserveId, bob, params.wethSupplyAmount, bob);
+    setUsingAsCollateral(spoke1, bob, params.wethReserveId, true);
+
+    // Bob draw 2000 total dai + usdx
```
yan-man:
```suggestion
    // Bob draw 4000 total dai + usdx
```

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+
+  function test_getUserRiskPremium_multi_asset_collateral_weth_partial_cover() public {
+    TestInfo memory params;
+    params.daiReserveId = spokeInfo[spoke1].dai.reserveId;
+    params.usdxReserveId = spokeInfo[spoke1].usdx.reserveId;
+    params.wethReserveId = spokeInfo[spoke1].weth.reserveId;
+
+    params.daiSupplyAmount = 2000e18;
+    params.usdxSupplyAmount = 2000e6;
+    params.wethSupplyAmount = 1e18;
```
yan-man: can you add a comment with the weth price so some of these calcs are clearer on how much is covered, maybe at the top so it doesnt have to be added everywhere in the tests

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    userPosition = getUserInfo(spoke1, bob, params.wethReserveId);
+    assertEq(userPosition.baseDebt, params.wethBorrowAmount);
+
+    userPosition = getUserInfo(spoke1, alice, params.wethReserveId);
+    assertEq(
+      userPosition.suppliedShares,
+      hub.convertToSharesDown(wethAssetId, params.wethSupplyAmount)
+    );
+
+    // Dai and usdx will each cover half the debt
```
yan-man: can you add some comment to clarify bc the dai covers the first half fully bc it has a lower LP, and then usdx follows up

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+
+    // Weth covers half the debt, dai covers the rest
+    assertEq(
+      spoke1.getUserRiskPremium(bob),
+      _calculateExpectedUserRP(bob, spoke1),
+      'user risk premium'
+    );
+  }
+
+  function test_getUserRiskPremium_two_assets_equal_parts() public {
```
yan-man: can you rename these "asset" tests to "reserves" to clarify it's a spoke action

tests/Base.t.sol
miguelmtzinf: `getAssetByReserveId`?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -232,20 +232,20 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiBorrowAmount = daiSupplyAmount / 2;
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, wethReserveId(spoke1), bob, wethSupplyAmount, bob);
```
miguelmtzinf: all these tests now fail because we were assuming that weth has non-zero liqPremium.
I d prefer keep using spoke1, just because the rest of the tests are using spoke1 and this slight change can bring confusion and harder to mantain.
Few alternatives:
1. update or mock liq premium of WETH at start
2. change to wbtc coll

tests/Base.t.sol
```diff
@@ -222,26 +222,26 @@ abstract contract Base is Test {
...
+      liquidityPremium: 10_00,
       borrowable: true,
       collateral: true
     });
-    DataTypes.ReserveConfig memory usdxConfig = DataTypes.ReserveConfig({
-      lt: 0.78e4,
+    DataTypes.ReserveConfig memory wbtcConfig = DataTypes.ReserveConfig({
+      lt: 0.75e4,
       lb: 0,
       liquidityPremium: 50_00,
```
miguelmtzinf: i d change it to smt more realistic, like 5%?

tests/unit/Spoke/Spoke.Supply.t.sol
```diff
@@ -379,7 +379,7 @@ contract SpokeSupplyTest is SpokeBase {
   }
 
   function test_supply_index_increase_with_premium() public {
-    _increaseReserveIndex(spoke1, daiReserveId(spoke1));
+    _increaseReserveIndex(spoke2, daiReserveId(spoke2));
```
miguelmtzinf: same thing here. I mean, if the goal is to use an asset with non-zero liqPremium as collateral, we don't need to change to another spoke? just use another asset as coll?

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {Arrays} from 'src/dependencies/openzeppelin/Arrays.sol';
+import {KeyValueListInMemory} from 'src/contracts/KeyValueListInMemory.sol';
+import 'tests/Base.t.sol';
+import {Spoke} from 'src/contracts/Spoke.sol';
+import {WadRayMath} from 'src/contracts/WadRayMath.sol';
```
DhairyaSethi: i've always wanted to suggest import order should follow style guidelines in one of your PRs 😆

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
miguelmtzinf: let's add the case where it's supplied but not activated as collateral

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    super.initEnvironment();
+  }
+
+  function test_getUserRiskPremium_no_collateral() public view {
+    uint256 userRiskPremium = spoke1.getUserRiskPremium(bob);
+    assertEq(userRiskPremium, 0, 'user risk premium');
+  }
+
+  function test_getUserRiskPremium_single_asset_collateral() public {
+    uint256 daiReserveId = spokeInfo[spoke1].dai.reserveId;
```
miguelmtzinf: can we use the getters pls? `daiReserveId(spoke1)` I think

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    uint256 daiAmount = 100e18;
+
+    // Bob supply dai into spoke1
+    Utils.spokeSupply(spoke1, daiReserveId, bob, daiAmount, bob);
+    setUsingAsCollateral(spoke1, bob, daiReserveId, true);
+
+    uint256 userRiskPremium = spoke1.getUserRiskPremium(bob);
+    assertEq(userRiskPremium, 0, 'user risk premium');
```
miguelmtzinf:
```suggestion
    assertEq(spoke1.getUserRiskPremium(bob), 0, 'user risk premium');
```
do we wanna make it shorter? same above

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+import {Spoke} from 'src/contracts/Spoke.sol';
+import {WadRayMath} from 'src/contracts/WadRayMath.sol';
+
+contract SpokeUserRiskPremiumTest is Base {
+  using SharesMath for uint256;
+  using WadRayMath for uint256;
+  using PercentageMath for uint256;
+  using KeyValueListInMemory for KeyValueListInMemory.List;
+
+  struct TestInfo {
```
miguelmtzinf: feels a bit unnecessary to have all reserves here in 1 single struct.
Can we just simplify with the variables needed and instantiate 1 single struct per reserve in each test?
example [here](https://github.com/aave/aave-v4/blob/e2383b1e3ca8497c2baeb639df5d67232631d4d9/tests/unit/Spoke/Spoke.Repay.t.sol#L1114)

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    // Bob draw dai
+    Utils.spokeBorrow(spoke1, params.daiReserveId, bob, params.borrowAmount, bob);
+
+    uint256 userRiskPremium = spoke1.getUserRiskPremium(bob);
+
+    // With single collateral, user rp will match liquidity premium of collateral
+    assertEq(userRiskPremium, params.daiLP, 'user risk premium');
+
+    // Supplying more risky asset (usdx) should not impact user risk premium
+    Utils.spokeSupply(spoke1, params.usdxReserveId, bob, additionalSupplyAmount, bob);
```
miguelmtzinf: we gotta activate as collateral

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+
+    // Bob supply weth into spoke1
+    Utils.spokeSupply(spoke1, params.wethReserveId, bob, params.wethSupplyAmount, bob);
+    setUsingAsCollateral(spoke1, bob, params.wethReserveId, true);
+
+    // Bob draw 2000 total dai + usdx
+    Utils.spokeBorrow(spoke1, params.daiReserveId, bob, params.daiBorrowAmount, bob);
+    Utils.spokeBorrow(spoke1, params.usdxReserveId, bob, params.usdxBorrowAmount, bob);
+
+    // Weth is enough to cover the total debt
```
miguelmtzinf, line 188: can we assert this?

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+
+  // TODO: Fuzz test showing 2 diff users borrowing the same 2 assets, and show their own risk premiums are calculated and applied correctly
+
+  function _normalizedValue(uint256 amount, uint256 assetId) internal view returns (uint256) {
+    return
+      (amount * oracle.getAssetPrice(assetId) * WadRayMath.WAD) /
+      (10 ** hub.getAssetConfig(assetId).decimals);
+  }
+
+  function _calculateExpectedUserRP(address user, ISpoke spoke) internal view returns (uint256) {
```
miguelmtzinf: should we move this to any Base test file?

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    assertEq(
+      rateChecks.assetPremium,
+      debts.bobUsdxPremiumDebtAfter + debts.aliceUsdxPremiumDebtAfter,
+      'hub asset outstanding premium after accrual'
+    );
+  }
+
+  // TODO: Fuzz test showing 2 diff users borrowing the same 2 assets, and show their own risk premiums are calculated and applied correctly
+
+  function _normalizedValue(uint256 amount, uint256 assetId) internal view returns (uint256) {
```
miguelmtzinf: not v clear what this does by the name

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
miguelmtzinf: i dn't leave numbers in the comment because of maintainability

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+
+    // Bob supply usdx into spoke1
+    Utils.spokeSupply(spoke1, params.usdxReserveId, bob, params.usdxSupplyAmount, bob);
+    setUsingAsCollateral(spoke1, bob, params.usdxReserveId, true);
+
+    // Alice supply weth into spoke1
+    Utils.spokeSupply(spoke1, params.wethReserveId, alice, params.wethSupplyAmount, alice);
+    setUsingAsCollateral(spoke1, alice, params.wethReserveId, true);
+
+    // Bob draw $4000 total in weth
```
miguelmtzinf: same about numbers

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    assertEq(userPosition.baseDebt, 0);
+
+    userPosition = getUserInfo(spoke1, bob, params.wethReserveId);
+    assertEq(userPosition.baseDebt, params.wethBorrowAmount);
+
+    userPosition = getUserInfo(spoke1, alice, params.wethReserveId);
+    assertEq(
+      userPosition.suppliedShares,
+      hub.convertToSharesDown(wethAssetId, params.wethSupplyAmount)
+    );
```
miguelmtzinf: these asserts are probably not needed. There are some like these in other tests, can you check and remove?

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+
+    // Dai and usdx will each cover half the debt
+    assertEq(
+      spoke1.getUserRiskPremium(bob),
+      _calculateExpectedUserRP(bob, spoke1),
+      'user risk premium'
+    );
+  }
+
+  function test_getUserRiskPremium_fuzz_two_assets_diff_amounts(uint256 daiSupplyAmount) public {
```
miguelmtzinf: i think having the next one is just enough? or you wanted to target an specific scenario here

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    uint256 baseDebt = params.wbtcBorrowAmount;
+    uint256 originalBaseDebt = params.wbtcBorrowAmount;
+    (uint256 actualBaseDebt, uint256 actualPremium) = spoke3.getUserDebt(params.wbtcReserveId, bob);
+    uint256 startTime = vm.getBlockTimestamp();
+
+    assertEq(baseDebt, actualBaseDebt, 'user base debt');
+    assertEq(actualPremium, 0, 'user outstanding premium');
+
+    // Wait a year
+    skip(365 days);
```
miguelmtzinf, line 865: shouldn't we check `getUserRiskPremium` after time skip?
the following asserts are just about debt accrual

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    (rateChecks.actualBaseDebt, rateChecks.actualPremium) = spoke3.getUserDebt(
+      params.wethReserveId,
+      bob
+    );
+
+    assertEq(rateChecks.baseDebt, rateChecks.actualBaseDebt, 'user base debt');
+    assertEq(rateChecks.actualPremium, 0, 'user outstanding premium');
+
+    // Wait a year
+    skip(365 days);
```
miguelmtzinf, line 1015: same here
is the intention to recalc rp after debt accrual?

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1606 @@
...
+    // Asset outstanding premium should be the sum of both users' outstanding premium
+    assertEq(
+      rateChecks.assetPremium,
+      debts.bobUsdxPremiumDebtAfter + debts.aliceUsdxPremiumDebtAfter,
+      'hub asset outstanding premium after accrual'
+    );
+  }
+
+  // TODO: Fuzz test showing 2 diff users borrowing the same 2 assets, and show their own risk premiums are calculated and applied correctly
+
```
miguelmtzinf, line 1505: missing tests for:
- user risk premium decreases after repayment
- user risk premium increases after withdrawal (or removal of collateral)
- user risk premium changes because of collateral accrual (no debt change)
- user risk premium changes because of debt accrual (no collateral change)
- user risk premium changes because of coll and debt accrual (this is what's done above?)

Lmk wdyt and we can create a separate ticket for this

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
DhairyaSethi: also `Arrays` is unused

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: My apologies, duplicate naming

tests/unit/Spoke/SpokeAccrueInterest.t.sol
CheyenneAtapour: The PR description mentions that some of the accrue interest tests were here. But main point is RP. The other tests will come in later PRs, as the description explains

tests/unit/Spoke/SpokeAccrueInterest.t.sol
CheyenneAtapour: Yes, these will def come in a future PR. Good catch about the risk premium though. I'm changing this reserve to WETH (which has 0 lp) and also setting as collateral

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: Miguel has said not to put numbers in the comments. I've removed the numbers from comments now

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: Added

tests/unit/Spoke/SpokeAccrueInterest.t.sol
CheyenneAtapour: Removed

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Yeah, this was tricky cause of the discrepancy in base test. I did the approach to change liq premium at the start

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: I simplified this test info struct. But also should probably do it with the other structs too

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: Is it not sufficient that I assert the user's risk premium is equivalent to the LP of weth? But yeah I can add

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: It was clearer I was targeting a specific scenario when I had the math directly in the code without the helper. But the following fuzz test does indeed cover this case anyway. So I'll remove

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: The user risk premium shouldn't change with just a time skip and no action, but I can indeed assert that is the case

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: No change in RP as mentioned, so no need for the re-calc. But I am ensuring this RP is applied properly when accruing interest

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: Separate ticket for this, as well as 2 reserves 2 users fuzzed, and then can add more spokes etc.

CheyenneAtapour: Issue created https://github.com/orgs/aave/projects/14/views/1?pane=issue&itemId=100805824&issue=aave%7Caave-v4%7C244

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
CheyenneAtapour: Changed name

tests/unit/Spoke/SpokeAccrueInterest.t.sol
```diff
@@ -1,47 +1,117 @@
...
-    initEnvironment();
-  }
-
   function test_accrueInterest_NoActionTaken() public {
-    DataTypes.Reserve memory daiInfo = spoke1.getReserve(spokeInfo[spoke1].dai.reserveId);
-    assertEq(daiInfo.lastUpdateTimestamp, 0);
-    assertEq(daiInfo.baseDebt, 0);
-    assertEq(daiInfo.outstandingPremium, 0);
-    assertEq(daiInfo.riskPremium, 0);
+    DataTypes.Reserve memory wethInfo = getReserveInfo(spoke1, spokeInfo[spoke1].weth.reserveId);
```
miguelmtzinf:
```suggestion
    DataTypes.Reserve memory wethInfo = getReserveInfo(wethReserveId(spoke1));
```

miguelmtzinf: do we wanna do this across this file?

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -246,4 +248,75 @@ contract SpokeBase is Base {
         collData.config.lt
       ) + 1;
   }
+
+  function _getNormalizedReserveValue(
+    uint256 amount,
+    uint256 assetId
```
miguelmtzinf: can we flip the order and add a 1-line docs comment above? we must say we return the value in WAD terms right

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1638 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+import {Spoke} from 'src/contracts/Spoke.sol';
+import {WadRayMath} from 'src/contracts/WadRayMath.sol';
```
miguelmtzinf: are these needed? they are imported already in Base i think. Otherwise, i d move them there

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1638 @@
...
+    uint256 aliceDaiSupplyAmount;
+    uint256 bobUsdxBorrowAmount;
+    uint256 bobUsdxSupplyAmount;
+    uint256 aliceUsdxBorrowAmount;
+    uint256 aliceUsdxSupplyAmount;
+  }
+
+  function test_getUserRiskPremium_no_collateral() public {
+    // Assert Bob has no collateral
+    for (uint256 i = 0; i < spoke1.reserveCount(); i++) {
```
DhairyaSethi:
```suggestion
    for (uint256 reserveId = 0; i < spoke1.reserveCount(); ++reserveId) {
```
nit for consistency

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1638 @@
...
+  struct Amounts {
+    uint256 bobDaiBorrowAmount;
+    uint256 bobDaiSupplyAmount;
+    uint256 aliceDaiBorrowAmount;
+    uint256 aliceDaiSupplyAmount;
+    uint256 bobUsdxBorrowAmount;
+    uint256 bobUsdxSupplyAmount;
+    uint256 aliceUsdxBorrowAmount;
+    uint256 aliceUsdxSupplyAmount;
+  }
```
miguelmtzinf: can we follow similar approach to `ReserveInfoLocal`?
or just build 1 single struct for 3 things (rate, premium/baseDebt after and borrow/supply amount)?
also, premium/debtAfter after are expected values that we calculate? If so I d consider renaming.

tests/unit/Spoke/SpokeBase.t.sol
miguelmtzinf: actually leaning towards renaming to smt like `getReserveValueInBaseCurrency`
similar to `_getUserDebtInBaseCurrency`

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
```diff
@@ -0,0 +1,1638 @@
...
+    setUsingAsCollateral(spoke1, bob, wethInfo.reserveId, true);
+
+    // Bob draw dai + usdx
+    Utils.spokeBorrow(spoke1, daiInfo.reserveId, bob, daiInfo.borrowAmount, bob);
+    Utils.spokeBorrow(spoke1, usdxInfo.reserveId, bob, usdxInfo.borrowAmount, bob);
+
+    // Weth is enough to cover the total debt
+    assertGe(
+      _getNormalizedReserveValue(wethInfo.supplyAmount, wethAssetId),
+      _getNormalizedReserveValue(daiInfo.borrowAmount + usdxInfo.borrowAmount, daiAssetId),
```
miguelmtzinf: don't we need to separate this? as they don't have same token decimals

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
miguelmtzinf: actually yes, nbd

tests/unit/Spoke/SpokeUserRiskPremium.t.sol
miguelmtzinf: but the user rp is impacted by coll and debt interest accrual no?
the storage value shouldn't change, but the `getUserRiskPremium` calculates the new risk premium from scratch

CheyenneAtapour: I can recalculate to check the math, but the true value that interest would be accrued on is the user risk premium at the time of previous user action

CheyenneAtapour: imho, it seems an awkward place to check this calc, because this new calculated risk premium is not being used at all, but I will add it

miguelmtzinf: hmm I see your point. Maybe makes more sense to have in a separate test file, specific to in-flight calcs or alongside the refresh fn

Issue #141, opened 2025-01-30
Keeping Assets Sorted by Liquidity Premium


CheyenneAtapour:
Currently the algorithm to compute an individual user's risk premium requires us to sort all assets by liquidity premium. Even though there may not be many assets on a spoke, there is a possible optimization we can use. 

Given that it's expected both new asset listings and governance changing the liquidity premium of an asset will likely not happen super often, it might be beneficial to ensure our maintained list of assets on-chain is sorted by liquidity premium. If we do this, then it is not required to sort the list upon every user action (because any user action forces a recalculation of that user's risk premium).

The proposal here is to modify the addAsset() function in the Liquidity Hub to maintain sorted order by liquidity premium, and to ensure the same behavior when we write the function allowing governance to update an asset's liquidity premium.

**Question:** Do we plan on modifying an individual user's risk premium ever? We can discuss how we tackle that if so, and perhaps if there are any incompatibilities

avniculae:
We can duplicate the fields from the Reserve data that are used in the rp algorithm, such that we can keep the list sorted while at the same time not increasing the number of sload/gas consumed

PR #149, opened 2025-02-10
feat: sorted packed list lib

DhairyaSethi:
for #141 

also contains alternative approach analysis & research
after deciding, rm unneeded


DhairyaSethi:
note: we're going with run time sort for now

PR #167, opened 2025-02-13
PoC: failing asset borrow index

yan-man:
Additional issues found, related to when `lastUpdateTimestamp` is updated

To see bug, uncomment lines shown below in LH and run `forge test -vvv --mt test_accrueInterest_dynamicTime_scenario2`.
- test_accrueInterest_dynamicTime_scenario2 also has comments describing the issue

(Assume each time pass is 1 year; 10% IR)

```
Scenario:

t0: init asset1, init spoke1, no actions
asset:
- index = 1 RAY
- lastUpdateTimestamp initialized

t1: spoke1 supplies/draws
asset:
- index = 1.1 RAY
- lastUpdateTimestamp stays at t0

t2: add spoke4, which draws
asset: 
- index = 1.32 RAY (ie 1.1 RAY * 120%). Incorrect bc previewNextBorrowIndex calcs based on existing lastUpdateTimestamp (still at t0). 
- So cumulatedBaseInterest appears to accrue 10% over 2 years (t0 -> t2, 120%) instead of correctly from t1 -> t2. 
- Correct value = 1.21 RAY (ie 1.1 RAY * 110%)
```

Proposed fix: 
- in LH _accrueInterest, when spoke/asset index are updated in first draw, lastUpdateTimestamp should also be updated. When adding this fix, it is resolved. 
Screenshot 2025-02-12 at 21 12 40

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -394,13 +394,13 @@ contract LiquidityHub is ILiquidityHub {
 
     if (isDrawAndExistingSpokeBaseDebtZero) {
       spoke.baseBorrowIndex = nextBaseBorrowIndex;
-      spoke.lastUpdateTimestamp = block.timestamp;
+      spoke.lastUpdateTimestamp = block.timestamp; // comment this line to see failing test
       // for the case where:
       // all asset debt is repaid, and then a new draw happens with some accrual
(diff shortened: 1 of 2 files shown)


yan-man:
will implement fix in separate PR

PR #172, opened 2025-02-14
Generalized scenario test framework applied to borrow index

### LINKED ISSUE #177: Create scenario testing framework
Generalized scenario testing framework to allow structured assertions at various points in time across multiple actions

yan-man:
- create initial scenario testing framework, applied to borrow index testing
- refactor previous scenario tests for borrow index into this framework (rename prior scenario4 test to scenario3)
- also fix the initial spoke borrow index values - initialize spoke indexes to 0 explicitly

To see scenario test logs:
- adjust `isPrintLogs` to `true`
- run `forge test --mt test_borrowIndexScenario -vvv`

closes #177 
closes https://github.com/aave/aave-v4/issues/179

### DIFF
--- a/tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
+++ b/tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
@@ -46,7 +46,7 @@ abstract contract LiquidityHubScenarioBaseTest is BaseTest {
   }
 
   uint256[] internal timestamps;
-  AssetDatas[NUM_SPOKES] internal assets;
+  AssetDatas[NUM_ASSETS] internal assets;
   SpokeDatas[NUM_SPOKES] internal spokes;
   SpokeActionAssetIds[NUM_SPOKES] internal spokeActions;
   SpokeAmounts[NUM_SPOKES] internal spokeAmounts;


tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario3.t.sol
```diff
@@ -0,0 +1,542 @@
...
+      assertEq(
+        spokes.spoke1.t3_i.lastUpdateTimestamp,
+        timeAt(Stages.t1),
+        't3_i Spoke1 lastUpdateTimestamp'
+      );
+
+      // spoke4
+      // spoke index is out of sync with asset index
+      // because spoke index is set to asset's next borrow index
+      // assertEq(
```
yan-man: @DhairyaSethi this is the only weird thing, where spoke being added won't have index aligning with asset index at this point

DhairyaSethi: follows my message on tg from yesterday. it can be init to zero (along with last update timestamp) to keep in line with spoke's accounting of itself, or we can have an assumption that addReserve & addSpoke are to be done by the dao in one txn.
we could also keep last stored, will mean spoke's accounting is not in sync until the first spoke action

DhairyaSethi: @miguelmtzinf wdyt?

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
```diff
@@ -0,0 +1,416 @@
...
+      );
+    }
+  }
+
+  function printInitialLog(Stages stage) internal override {
+    if (stage == Stages.t0) {
+      console.log('----- t0_i -----');
+
+      console.log('Asset borrow index %e', assets.wethData.t0_i.baseBorrowIndex);
+      console.log('Asset base debt %e', assets.wethData.t0_i.baseDebt);
```
DhairyaSethi: btw you can also do the following to fix the decimals for good QoL
```suggestion
      console.log('Asset borrow index %27e', assets.wethData.t0_i.baseBorrowIndex);
      console.log('Asset base debt %18e', assets.wethData.t0_i.baseDebt);
```
which leads to a suggestion of making the scenarios token agnostic maybe such that they run for any set of tokens (and hence decimals) w/o needing a copy pasta. wdyt?

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario3.t.sol
yan-man: after discussion 2.17.25, we understand that spoke indexes won't always align with asset indexes anyway (after actions start occurring, as only spoke performing action will align with asset index). Also, we will define initial spoke indexes to 0 (with timestamp 0). Asset index will start from index of 1 RAY / current block timestamp

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
yan-man: good idea for token agnostic tests and also to add qol for borrow index. I think it doesn't make as much sense for base debt because that is dependent on the decimals for the token at hand. 

added: 
- https://github.com/aave/aave-v4/pull/172/commits/dca3836f8c7119bfbb12ce3d010116ee05969497
- https://github.com/aave/aave-v4/pull/172/commits/6001370833791d080374ee59db73cbdfa4843379

src/contracts/LiquidityHub.sol
```diff
@@ -44,6 +44,9 @@ contract LiquidityHub is ILiquidityHub {
   using AssetLogic for Asset;
   using SpokeDataLogic for SpokeData;
 
+  uint256 public constant DEFAULT_ASSET_INDEX = WadRayMath.RAY;
+  uint256 public constant DEFAULT_SPOKE_INDEX = 0;
```
yan-man, line 48: wanted to set this explicitly so it is clear this is a conscious decision

src/contracts/LiquidityHub.sol
```diff
@@ -87,7 +90,7 @@ contract LiquidityHub is ILiquidityHub {
       availableLiquidity: 0,
       baseDebt: 0,
       outstandingPremium: 0,
-      baseBorrowIndex: WadRayMath.RAY,
+      baseBorrowIndex: DEFAULT_ASSET_INDEX,
```
DhairyaSethi, line 93: this should remain 1 Ray, no? (on asset)

yan-man: yes but just wanted to set it explicitly in a constant to make it clear, thoughts? do you think unnecessary

src/contracts/Spoke.sol
```diff
@@ -275,6 +276,7 @@ contract Spoke is ISpoke {
     reservesList.push(reserveCount);
     reserve.assetId = assetId;
     reserve.asset = asset;
+    reserve.baseBorrowIndex = DEFAULT_SPOKE_INDEX;
     reserve.config = ReserveConfig({
       lt: params.lt,
       lb: params.lb,
```
DhairyaSethi: how about we initialise with a struct only here such that no fields are missed (also show `lastUpdateTimestamp` is also explicitly set to 0) like everywhere else
idk why we were doing it like this before

src/contracts/LiquidityHub.sol
DhairyaSethi: ohh my bad, i misread `DEFAULT_ASSET_INDEX` as 0, lil confusing. yes sure, personally would prefer 1 Ray directly here (since it's the exception) but not a strong preference. this way works too

src/contracts/Spoke.sol
yan-man: agree 

fixed here: https://github.com/aave/aave-v4/pull/172/commits/f9567130293bed43bb54af09d1fb6a7deb6bdf5a

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
```diff
@@ -0,0 +1,427 @@
...
+
+  // Scenario:
+  // t0: asset added, spoke1 added, spoke1 draws
+  // t1: spoke4 is added; spoke4 draws
+  // t2: spoke4 trivial supply action to trigger accrual
+
+  // Assumptions:
+  // - constant 10% IR
+  // - 1 year between each action
+  // - single asset (weth)
```
DhairyaSethi, line 25: i'm thinking the nice part about this structure is that we can make everything dynamic. like with the asset id change, how about we fuzz these values as well, wdyt? i'm thinking another test which sets these internal values (by making them state vars) after bounds to run: for `assetId, baseBorrowRate, delay`
this way we could rule out edge cases or precision issues with values also. wdyt? 
sorry didn't suggest before

DhairyaSethi: could also do with supply/borrow amounts, but more caveats there bc supplied amount >= sum of borrowed amounts to start

DhairyaSethi: i was doing this in the tests before (setting draw amount by the fuzzed supply amount) https://github.com/aave/aave-v4/blob/feat/scenario-test-framework/tests/unit/LiquidityHub/LiquidityHubRiskPremium.t.sol#L169
but maybe here it's better that there are min bounds to supply amount (basically the other way around, borrowed amounts restrict supply amounts)

yan-man: agree but want to leave additional tests to next PR, this one to introduce the framework and migrating existing scenario tests into it.

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
```diff
@@ -0,0 +1,427 @@
...
+        spokes.spoke1.t2_f.lastUpdateTimestamp,
+        spokes.spoke1.t0_f.lastUpdateTimestamp,
+        't2_f Spoke1 lastUpdateTimestamp'
+      );
+
+      // spoke4
+      assertEq(
+        spokes.spoke4.t2_f.baseBorrowIndex,
+        assets.assetData0.t2_f.baseBorrowIndex,
+        't2_f Spoke4 base debt'
```
DhairyaSethi:
```suggestion
        't2_f Spoke4 index'
```

src/contracts/Spoke.sol
```diff
@@ -266,26 +267,34 @@ contract Spoke is ISpoke {
     ReserveConfig memory params,
     address asset
   ) external returns (uint256) {
-    Reserve storage reserve = _reserves[reserveCount];
+    uint256 reserveCountTmp = reserveCount;
```
DhairyaSethi: final nit: can we name this reserveCount with an underscore suffix or prefix

also if we're doing a temp var here, can just return that since return is post increment (returns current value)

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
yan-man: https://github.com/aave/aave-v4/pull/172/commits/96b960d650f7a3643a8df348b60ea15c4defc0c9

src/contracts/Spoke.sol
yan-man: prefer `_reserveCount` as it's akin to an internal var. 

however, don't think we can return `_reserveCount` because we need to increment in storage then return the new val in the: `return reserveCount++;` action?

https://github.com/aave/aave-v4/pull/172/commits/a53d55891bc5569fade618a6a5bd9a6b4b09705a

DhairyaSethi: return reserveCount++ returns the current value and then increments in storage imo

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
```diff
@@ -0,0 +1,186 @@
...
+        printFinalLog(stage);
+      }
+      skipTime(stage);
+      postcondition(stage);
+
+      stage = Stages(uint256(stage) + 1);
+    }
+  }
+
+  function timeAt(Stages stage) internal view returns (uint256) {
```
DhairyaSethi: possibly cleaner to have this return a uint40 
saves casting while using as an arg on calcSimpleInterest and assertEq's will automatically cast it to uint256

src/contracts/Spoke.sol
```diff
@@ -266,26 +267,34 @@ contract Spoke is ISpoke {
     ReserveConfig memory params,
     address asset
   ) external returns (uint256) {
-    Reserve storage reserve = _reserves[reserveCount];
+    uint256 _reserveCount = reserveCount;
```
CheyenneAtapour, line 270: The reason for this is it's just cheaper to read from a cached local variable than the global one in storage, right?

CheyenneAtapour:
Overall, I really like the approach here @yan-man . I wonder if in future iterations we can make it even easier to define and test scenarios like these. Maybe like a JSON is inputted that creates the scenario somehow, or code is generated from some input. We can continue to think about it

src/contracts/Spoke.sol
DhairyaSethi: yes

src/contracts/Spoke.sol
```diff
@@ -266,26 +267,34 @@ contract Spoke is ISpoke {
...
-    reserve.assetId = assetId;
-    reserve.asset = asset;
-    reserve.config = ReserveConfig({
-      lt: params.lt,
-      lb: params.lb,
-      liquidityPremium: params.liquidityPremium,
-      borrowable: params.borrowable,
-      collateral: params.collateral
+    reservesList.push(_reserveCount);
+    _reserves[_reserveCount] = Reserve({
```
miguelmtzinf, line 278: we don't need to explicitly initialize these variables right? is it intentional?

DhairyaSethi: yeah just for verbosity https://github.com/aave/aave-v4/pull/172#discussion_r1958747063

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
```diff
@@ -0,0 +1,186 @@
...
+import 'tests/BaseTest.t.sol';
+import {Asset, SpokeData} from 'src/contracts/LiquidityHub.sol';
+
+abstract contract LiquidityHubScenarioBaseTest is BaseTest {
+  bool internal isPrintLogs = false;
+
+  // t0_i: prior to action
+  // t0_f: after action
+  struct Timestamps {
+    uint256 t0_i;
```
miguelmtzinf: why not using 2 arrays? same for the rest of structs

miguelmtzinf: what do `_i` and `_f` refer to? also, why the need of the split if for example `t0_f == t1_i` ?

DhairyaSethi: initial & final. i'd raised the same q, the idea is to have more granularity to assert things right before/after (even tho it's technically possible regardless since ti_f == ti+1_i like you said) 

re arrays, agree yeah that would be cleaner i think

src/contracts/Spoke.sol
yan-man: agreed (returning current value), are you suggesting to rather do:
```
reserveCount++;
return _reserveCount;
```

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
yan-man: yes initial and final, I specifically did this because t0_f is ex. at t0, at the end of the action, while still at t0. But t1_i is after time has skipped so it is at t1. I did this to be more granular because when we have getters/etc which include accrual calcs it's helpful to be able to check things at both points in time (pre/post accrual after time passes).

I can update to 2 arrays instead. edit: it is much cleaner! 

fixed: https://github.com/aave/aave-v4/pull/172/commits/6736ff60d69a7e6cabdc3c3d3b0ecae78876e208

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/172/commits/6736ff60d69a7e6cabdc3c3d3b0ecae78876e208

src/contracts/Spoke.sol
DhairyaSethi: yeah i'd do the increment inside the array allocation, think it'd be cheaper. it still doesn't look very pretty

yan-man: fixed: https://github.com/aave/aave-v4/pull/172/commits/9ceb918c7613af2f84099913695a9e5865798768

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
DhairyaSethi: ty! still doesn't use arrays tho? i think @miguelmtzinf suggestion was a time, amount etc array as well such that we don't need t0, t1, t2 keys in the struct, rather t[0], t[1] etc

miguelmtzinf: yes, exactly

yan-man: You guys are right, even cleaner! 

fixed: https://github.com/aave/aave-v4/pull/172/commits/4f60053be2a82c93be742b0e597922596f14f50e

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
```diff
@@ -0,0 +1,128 @@
...
+pragma solidity ^0.8.0;
+
+import 'tests/BaseTest.t.sol';
+import {Asset, SpokeData} from 'src/contracts/LiquidityHub.sol';
+
+abstract contract LiquidityHubScenarioBaseTest is BaseTest {
+  bool internal isPrintLogs = false;
+
+  struct Timestamps {
+    uint256[10] t;
```
miguelmtzinf: do we wanna have here the 2 arrays for `i` and `t` instead?

tests/unit/LiquidityHub/LiquidityHubAccrueInterestTest.t.sol
```diff
@@ -13,6 +13,92 @@ contract LiquidityHubAccrueInterestTest is BaseTest {
 
   uint256 public constant MAX_BPS = 999_99;
 
+  struct SpokeDataLocal {
```
miguelmtzinf, line 16: do we wanna refactor here too?

yan-man: ah good call 

fixed: https://github.com/aave/aave-v4/pull/172/commits/9387789d863d6b6f563031c10de31f4c67a86975

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
```diff
@@ -0,0 +1,455 @@
...
+      // spoke1
+      assertEq(
+        spokes_i.spoke[0].t[0].baseBorrowIndex,
+        hub.DEFAULT_SPOKE_INDEX(),
+        't0_i Spoke1 index'
+      );
+      assertEq(spokes_i.spoke[0].t[0].baseDebt, 0, 't0_i Spoke1 base debt');
+      assertEq(spokes_i.spoke[0].t[0].lastUpdateTimestamp, 0, 't0_i Spoke1 lastUpdateTimestamp');
+    } else if (stage == Stages.t1) {
+      assets_i.assetData[0].t[1] = hub.getAsset(assetId);
```
DhairyaSethi: hmm i would refactor `Stage` struct since it can basically be replaced by a uint which is passed as the time key (0, 1, 2, etc)

DhairyaSethi: to keep it distinct i would use a custom type over this uint
https://docs.soliditylang.org/en/latest/types.html#user-defined-value-types

yan-man: this one followed solidity state machine pattern so thought more verbose might be clearer than just time key: https://docs.soliditylang.org/en/latest/common-patterns.html#state-machine

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  struct SpokeDatas {
+    SpokeData[10] t;
```
DhairyaSethi: how about we use a global constant for all the 10's (array sizes)?

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
DhairyaSethi: yeah the name can be `stage` itself, but the guideline there is when your stages have distinct names, that's when an enum is useful (bc under the hood it's just a uint8 array representation)
since our stages look like t0, t1 etc, it's cleaner to have them as an array directly w/o loosing verbosity imo

yan-man: eh I think it's negligible, plus with the custom types then comparisons on equality become a pain with wrapping and unwrapping. I'll just leave this as enum

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/172/commits/ec192bc32209973f16433243ad6bc694979a2b9b

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/172/commits/ec192bc32209973f16433243ad6bc694979a2b9b

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
DhairyaSethi: fair. pro is it makes the stage struct cleaner since it doesn't t0-t10 keys, and can reuse the array size int that everyone uses. (then pro is that in future tests, if it needs to do more than 10 time intervals, then just update one global value and you're good to go (over adding more keys in the struct) 

cmp should just be t[0] == Stage.wrap(0) 
or inverse

yan-man: If I did it like this, is this cmp correct?

```
type Stage is uint8;
...
 Stage[NUM_TIMESTAMPS] internal stages;
...
function exec(Stage stage)
...
Stage.unwrap(stage) == Stage.unwrap(stages[0])
```

DhairyaSethi: you can extend == as an operator on the type

DhairyaSethi: so it'd be 
```solidity
using {eq as ==} for Stage global;

function eq(Stage a, Stage b) pure returns (bool) {
    return Stage.unwrap(a) == Stage.unwrap(b);
}

...

if(Stage.wrap(stage) == stages[0]) { }

```

DhairyaSethi: whoops, looks like we were too quick to merge haha

PR #174, opened 2025-02-17
feat: up to date getters

### LINKED ISSUE #104: Fix `convertToAsssets{Down,Up}` and `convertToShares{Down,Up}` - Implementation

DhairyaSethi:
closes #104 

future gas opt: don't update baseDebt & outstandingPremium if they haven't changed 

open q: do we think we need a getter for stored values? 
- at the moment, get asset & get spoke return last stored values to support this design. if we all agree that's not needed, we can add accrued debt & premium there (unneeded tho imo) - there is still a discrepency

TODO in seperate PR: up to date getters for debt on spoke & hub

### DIFF
--- a/src/contracts/AssetLogic.sol
+++ b/src/contracts/AssetLogic.sol
@@ -85,7 +85,7 @@ library AssetLogic {
   // @return nextBaseBorrowIndex (in ray)
   function previewNextBorrowIndex(Asset storage asset) internal view returns (uint256) {
     uint256 lastUpdateTimestamp = asset.lastUpdateTimestamp;
-    if (block.timestamp == lastUpdateTimestamp) {
+    if (lastUpdateTimestamp == block.timestamp) {
       return asset.baseBorrowIndex;
     }
 
@@ -115,7 +115,7 @@ library AssetLogic {
     uint256 existingBaseDebt = asset.baseDebt;
     uint256 existingOutstandingPremium = asset.outstandingPremium;
 
-    if (existingBaseDebt == 0 || block.timestamp == asset.lastUpdateTimestamp) {
+    if (existingBaseDebt == 0 || asset.lastUpdateTimestamp == block.timestamp) {
       return (existingBaseDebt, existingOutstandingPremium);
     }
 
--- a/src/contracts/SpokeDataLogic.sol
+++ b/src/contracts/SpokeDataLogic.sol
@@ -13,7 +13,7 @@ library SpokeDataLogic {
 
   // @dev Utilizes existing `spoke.baseBorrowIndex` & `spoke.riskPremiumRad`
   function accrueInterest(SpokeData storage spoke, uint256 nextBaseBorrowIndex) internal {
-    if (block.timestamp == spoke.lastUpdateTimestamp) {
+    if (spoke.lastUpdateTimestamp == block.timestamp) {
       return;
     }
 


src/contracts/AssetLogic.sol
```diff
@@ -94,26 +97,36 @@ library AssetLogic {
...
+  ) internal view returns (uint256, uint256) {
+    uint256 existingBaseDebt = asset.baseDebt;
+    uint256 existingOutstandingPremium = asset.outstandingPremium;
 
-      // accrue premium interest on the accrued base interest
-      asset.outstandingPremium += (cumulatedBaseDebt - existingBaseDebt).radMul(
-        asset.riskPremiumRad
-      );
-      asset.baseDebt = cumulatedBaseDebt;
+    if (existingBaseDebt == 0 || block.timestamp == asset.lastUpdateTimestamp) {
```
miguelmtzinf:
```suggestion
    if (existingBaseDebt == 0 || asset.lastUpdateTimestamp == block.timestamp) {
```
nit - feels more consistent if inverting the second leg (comparing stored value vs external value)

src/contracts/AssetLogic.sol
```diff
@@ -80,7 +83,7 @@ library AssetLogic {
 
   // @dev Utilizes existing `asset.baseBorrowRate` & `asset.baseBorrowIndex`
   // @return nextBaseBorrowIndex (in ray)
-  function previewNextBorrowIndex(Asset storage asset) external view returns (uint256) {
+  function previewNextBorrowIndex(Asset storage asset) internal view returns (uint256) {
     uint256 lastUpdateTimestamp = asset.lastUpdateTimestamp;
     if (block.timestamp == lastUpdateTimestamp) {
```
miguelmtzinf: same here

src/contracts/AssetLogic.sol
```diff
@@ -94,26 +97,36 @@ library AssetLogic {
...
-    if (block.timestamp == asset.lastUpdateTimestamp) {
-      return;
-    }
+  function accrueInterest(Asset storage asset, uint256 nextBaseBorrowIndex) internal {
+    (uint256 cumulatedBaseDebt, uint256 cumulatedOutstandingPremium) = asset.previewInterest(
+      nextBaseBorrowIndex
+    );
 
-    uint256 existingBaseDebt = asset.baseDebt;
+    asset.baseDebt = cumulatedBaseDebt;
```
yan-man, line 105: i think perhaps could use another escape to avoid updating storage if `if (existingBaseDebt == 0 || block.timestamp == asset.lastUpdateTimestamp)` as in previewInterest

src/contracts/AssetLogic.sol
yan-man: I think we should add this as a convention generally too - value to be tested comes first? (also applies in the assertEq in tests)

(added to notion to discuss)

src/contracts/AssetLogic.sol
DhairyaSethi: i can inline everything all the time (by making previewInterest & previewNextBorrowRate private) such that optimiser adds this opt, at the cost of larger bytecode. wdyt? cc @miguelmtzinf 
imo instead of having duplicated if, it's better compute once and pass that bool around the methods, strictly gas wise, but looks uglier. open to suggestions

src/contracts/AssetLogic.sol
DhairyaSethi: [38c3064](https://github.com/aave/aave-v4/pull/174/commits/38c306430c1e4e9d8848dafc43345e94420bfd8a)

src/contracts/AssetLogic.sol
DhairyaSethi: sounds good! [38c3064](https://github.com/aave/aave-v4/pull/174/commits/38c306430c1e4e9d8848dafc43345e94420bfd8a)

DhairyaSethi:
@yan-man agree that optimisations possible, trying to ship smallest code diff such that we focus most on test suite and do optimisations later. let's discuss the one suggested on the comment, very open to suggestions on immediately obvious optimisations which could go in now

src/contracts/AssetLogic.sol
miguelmtzinf: both ugly. Let's keep it as it is for now. We can optimize more later

miguelmtzinf:
lfg, thanks guys!

PR #26, opened 2024-10-04
Explicit rounding on Math lib

DhairyaSethi:
addresses #22 

Adds `up` & `down` rounding direction on `wadMul`, `wadDiv`, `rayMul`, & `rayDiv` in WadRayExtended (kept separate to not disrupt existing flow, can replace/append to existing lib) and on `percentMul` & `percentDiv` on PercentageMath + tests

Changes from wayRayMath
- add assembly-safe annotation to yul block; minor opt 
- update overflow checks since we don't add den/2 anymore 
(lib essentially looks like the soulmate implementation now (tested differentially against))

minor nit
```
 iszero(iszero(x)) === gt(x, 0);
 for push0 supported chains, gt(x, 0) is cheaper,
 otherwise they're the same cost but iszero(iszero(x)) takes lesser stack space - preferred by solady
```

PR #183, opened 2025-02-18
risk premium in bps

DhairyaSethi:
### Goal
the idea is to always use rp in bps terms (since it's used as a value in two running weighted averages)

### Features
- using uint32 for riskPremium passed by spoke to hub (which means that the maximum supported rp is 4294967295 (2**32-1) bps or 42949672.95%.), primarily to enforce that we never accidentally send a Ray value during testing
- added bound checks for this to be under 1k bps as a security consideration (open to discussion)
- on usage, we do `riskPremium.fromRay().percentMul(val)` instead of `riskPremium.bpsToRay().percentMul(val)` because bpsToRay adds 1e27/1e4, ie 1e23 digits of precision which is not enough for loss. 1e27 is accurate till ~0.015% for extreme cases (1e27 * 1e4 / 1e30 (max weight)). we can add a 1e36 multiplier as discussed before as well
- also enhances the weighted average library test suite

### DIFF
--- a/tests/unit/LiquidityHub/LiquidityHubRiskPremium.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHubRiskPremium.t.sol
@@ -304,7 +304,7 @@ contract LiquidityHubRiskPremium_VariableTimeAndConstantRiskPremium is
   }
 
   function test_multipleDrawWhileAccruingInterestWithChangingRate() public {
-    uint256 timeToSkip = 365 days; // todo fuzz this and rate
+    uint256 timeToSkip = 365 days;
     uint256 rate = uint256(15_00).bpsToRay();
     uint40 lastUpdateTimestamp = uint40(vm.getBlockTimestamp());
 
@@ -393,8 +393,7 @@ contract LiquidityHubRiskPremium_VariableTimeAndConstantRiskPremium is
     uint256 timeToSkip,
     uint256 baseBorrowRate
   ) public {
-    // todo: minDrawAmount temp workaround
-    p = bound({input: p, minDrawAmount: daiAmount / 1e9, maxDrawAmount: daiAmount});
+    p = bound({input: p, minDrawAmount: daiAmount, maxDrawAmount: daiAmount});
     timeToSkip = bound(timeToSkip, 1 days, 100_000 days);
     baseBorrowRate = bound(baseBorrowRate, uint256(1).bpsToRay(), uint256(100_00).bpsToRay());
     uint40 lastUpdateTimestamp = uint40(vm.getBlockTimestamp());
@@ -473,7 +472,7 @@ contract LiquidityHubRiskPremium_VariableTimeAndConstantRiskPremium is
         p.drawAmount.spoke3 *
         p.riskPremium.spoke3) /
       totalBaseDebt;
-    assertApproxEqAbs(hub.getAsset(daiAssetId).baseDebt, totalBaseDebt, 2);
-    assertApproxEqAbs(hub.getAsset(daiAssetId).riskPremium.derayify(), expectedRiskPremium, 3);
+    assertApproxEqAbs(hub.getAsset(daiAssetId).baseDebt, totalBaseDebt, 1);
+    assertApproxEqAbs(hub.getAsset(daiAssetId).riskPremium.derayify(), expectedRiskPremium, 1);
   }
 }
--- a/tests/unit/MathUtils.t.sol
+++ b/tests/unit/MathUtils.t.sol
@@ -201,7 +201,7 @@ contract MathUtilsWeightedAverage is Test {
     }
 
     assertEq(currentSumWeights, calcSumWeights);
-    assertApproxEqAbs((currentWeightedAvg / precision), calcWeightedAvg, 2);
+    assertApproxEqAbs((currentWeightedAvg / precision), calcWeightedAvg, 1);
   }
 
   function _runWeightedAverageAdd(
@@ -236,7 +236,7 @@ contract MathUtilsWeightedAverage is Test {
     }
 
     assertEq(currentSumWeights, calcSumWeights);
-    assertApproxEqAbs((currentWeightedAvg / precision), calcWeightedAvg, 2);
+    assertApproxEqAbs((currentWeightedAvg / precision), calcWeightedAvg, 1);
 
     return (currentWeightedAvg, currentSumWeights);
   }
@@ -293,7 +293,7 @@ contract MathUtilsWeightedAverage is Test {
     }
 
     assertEq(currentSumWeights, calcSumWeights);
-    assertApproxEqAbs(currentWeightedAvg / precision, calcWeightedAvg, 2);
+    assertApproxEqAbs(currentWeightedAvg / precision, calcWeightedAvg, 1);
   }
 
   function _boundAndSplitArray(


src/contracts/LiquidityHub.sol
```diff
@@ -460,4 +466,9 @@ contract LiquidityHub is ILiquidityHub {
 
     return baseDebtRestored;
   }
+
+  function _toBoundedRay(uint32 bps) internal pure returns (uint256) {
+    require(bps <= 1000_00, 'INVALID_BPS');
+    return uint256(bps) * 1e27;
+  }
```
DhairyaSethi: open to suggestions on renaming this/other ways to do sanity bound checks on the spoke input (ready to argue why this is not unnecessary)

yan-man: I think only that:
- max BPS being 1000_00 is 1000%, ie 10x baseDebt correct?
- thoughts on moving this bound check into a lib, perhaps wadraymath?

DhairyaSethi: yes to both, had a todo to move to lib after we reach consensus on this check (on the hub, remember)

src/contracts/WadRayMath.sol
```diff
@@ -165,78 +126,17 @@ library WadRayMath {
...
-      b := mul(a, RAD)
-
-      if iszero(eq(div(b, RAD), a)) {
-        revert(0, 0)
-      }
-    }
-  }
-
   /**
    * @dev Truncates number from Rad, loosing denominator precision
```
miguelmtzinf: typo, `from Ray`

src/contracts/AssetLogic.sol
```diff
@@ -58,7 +58,11 @@ library AssetLogic {
   }
 
   function getInterestRate(Asset storage asset) external view returns (uint256) {
-    return asset.baseBorrowRate.radMul(WadRayMath.RAD + asset.riskPremiumRad);
+    // @dev we truncate fromRay before `percentMul` as we only have accurate data until bps
+    return
+      asset.baseBorrowRate.percentMul(
+        PercentageMath.PERCENTAGE_FACTOR + asset.riskPremium.fromRay()
```
miguelmtzinf: I d rename `fromRay` to something more obvious rather than adding a comment above (on every usage?).

src/contracts/LiquidityHub.sol
```diff
@@ -17,7 +17,8 @@ struct SpokeData {
   uint256 baseDebt; // asset
   uint256 outstandingPremium; // asset
   uint256 baseBorrowIndex; // in ray
-  uint256 riskPremiumRad; // weighted average risk premium in rad (bps value with extra `rad` precision)
+  // todo add ray to this name?
+  uint256 riskPremium; // weighted average risk premium in ray
```
miguelmtzinf, line 20: Idt is necessary as long is well documented.

src/contracts/LiquidityHub.sol
miguelmtzinf: it makes sense to do our own version of SafeCast a la OZ
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/SafeCast.sol#L560

src/contracts/LiquidityHub.sol
yan-man: agree, and not a fan of units in the var name also

src/contracts/LiquidityHub.sol
miguelmtzinf: as an alternative, we can create a new lib with constants to refer to variables ranges/decimals. Constants can be informative only, and functions validation helpers.
Think it probably makes more sense this approach, wdyt

src/contracts/Spoke.sol
```diff
@@ -383,7 +383,7 @@ contract Spoke is ISpoke {
       newUserRiskPremium,
       baseDebtChange
     );
-    return newAggregatedRiskPremium;
+    return uint32(newAggregatedRiskPremium.fromRay());
```
miguelmtzinf: why you don't wanna return as `uint256`? probably cheaper

src/contracts/Spoke.sol
```diff
@@ -617,10 +617,18 @@ contract Spoke is ISpoke {
...
+      user.outstandingPremium += (cumulatedBaseDebt - existingBaseDebt).percentMul(
+        user.riskPremium.fromRay()
+      );
     }
 
     user.baseBorrowIndex = nextBaseBorrowIndex;
     user.lastUpdateTimestamp = block.timestamp;
   }
+
+  function _fromBoundedRay(uint256 value) internal pure returns (uint32) {
```
miguelmtzinf: unused

src/contracts/LiquidityHub.sol
DhairyaSethi: yep enforcing limits to 1k bps rather than uint32.max (ie ~437M bps) 
this is the only place where we do this, any other amounts to limit?

also side note, 1k bps was a consequence before, with this model _we can_ support higher values 
(moved to lib btw [277511f](https://github.com/aave/aave-v4/pull/183/commits/277511f7fab9cff75dd4d6973a976e5f61717fd5))

src/dependencies/openzeppelin/EnumerableSet.sol
miguelmtzinf, line 1: oversight?

src/interfaces/ILiquidityHub.sol
```diff
@@ -46,14 +46,14 @@ interface ILiquidityHub {
    * @dev Only callable by spokes.
    * @param assetId The asset id.
    * @param amount The amount of debt to draw.
-   * @param riskPremiumRad The new aggregated risk premium of the calling spoke.
+   * @param riskPremium The new aggregated risk premium of the calling spoke.
```
yan-man: worth it to add "ray" units description in the natspecs here? 

`The new aggregated risk premium (in RAY) of the calling spoke`

src/contracts/Spoke.sol
DhairyaSethi: bc this value is passed to hub, and i liked a solidity restriction of the type such that it's ranged are more apparently clear 
why cast here instead of before sending? - cleaner to do here since it only needs one cast in the code. open to suggestions

src/interfaces/ILiquidityHub.sol
DhairyaSethi: THIS value is supposed to be in bps haha

DhairyaSethi: agree btw, added [9c06e07](https://github.com/aave/aave-v4/pull/183/commits/9c06e0754d4b0524ab477f7f51341eed9c30ae5f)

src/dependencies/openzeppelin/EnumerableSet.sol
DhairyaSethi: used as a helper in wavg tests, you'll see ;)
(can rm btw and revert to using a native implementation, added just bc it made tests look slightly cleaner with less overhead to think about)

src/contracts/AssetLogic.sol
DhairyaSethi: comment is more about why truncating before the mul (bc best practise to always divide at the end)
contenders: `truncateFromRay()`, `castFromRay()`
open to suggestions, i like `fromRay()` the most so far

src/contracts/WadRayMath.sol
yan-man: also: losing vs loosing

src/contracts/LiquidityHub.sol
yan-man: I think this commit link not working

src/contracts/AssetLogic.sol
yan-man: Mildly prefer truncateFromRay as the truncation part seems important here

src/contracts/LiquidityHub.sol
```diff
@@ -30,7 +30,7 @@ struct Asset {
   uint256 outstandingPremium; // asset
   uint256 baseBorrowIndex; // in ray
   uint256 baseBorrowRate; // in ray
-  uint256 riskPremiumRad; // in rad
+  uint256 riskPremium; // weighted average risk premium in ray
```
yan-man: worth it to clarify that this wAvg RP is across spokes and the one on Spoke is across its users?

src/contracts/LiquidityHub.sol
DhairyaSethi: thanks, updated. commit hash changed on the force push after rebasing latest changes

src/contracts/Spoke.sol
yan-man: agree that if you were to cast, this would seem the spot for it 

but is the uint32 range utilized in any way to bound or is it purely design choice on your part? If latter would prefer uint256 so we don't need to also bind the wadray lib to uint32

src/contracts/LiquidityHub.sol
```diff
@@ -239,7 +240,7 @@ contract LiquidityHub is ILiquidityHub {
     _validateDraw(asset, amount, spoke.config.drawCap);
 
     asset.updateBorrowRate({liquidityAdded: 0, liquidityTaken: amount});
-    _updateRiskPremiumAndBaseDebt(asset, spoke, riskPremiumRad, int256(amount)); // base debt added
+    _updateRiskPremiumAndBaseDebt(asset, spoke, riskPremium.toBoundedRay(), int256(amount)); // base debt added
```
yan-man: and just to clarify, are we planning to change the amount var to be uint256 instead of int256 and add a boolean, in line with previous conventions? forgot where we landed on that

src/contracts/AssetLogic.sol
DhairyaSethi: hm yeah, naming should also somewhat be in sync with its sister method (currently called `toBoundedRay()`) wdyt? 

the op can only do truncation or rounding: i think `fromRay()` here can implicitly mean truncation as opposed to rounding since throughout the codebase, on any rounding operation, we specify direction (`toSharesUp()` etc) 
ps there is also an open PR from ages ago which adds this convention to the existing percent math & wadRay libs lol https://github.com/aave/aave-v4/pull/26

src/contracts/LiquidityHub.sol
DhairyaSethi: > add a boolean

not this 
i think the idea is to replace the int with two uint's: namely `debtAdded` and `debtRemoved` similar to the convention set with`liquidityAdded` & `liquidityRemoved` for calculateBorrowRate params on the IR strategy. this is still an open ticket to implement, yes, hasn't been prioritised yet

src/contracts/Spoke.sol
DhairyaSethi: the idea is to have metal shoes so we don't accidentally shoot ourselves in the foot: by strictly enforcing the riskPremium while being passed to hub is a bps value (so limited to uint32 range) but while storing & utilising in the weighted average calcs (basically everywhere else in the code) is a uint256 (ray precision) 

note: this is an extra pre-caution since we have that bound check in `toBoundedRay()`

DhairyaSethi: gas wise zero benefits atm since while being passed in calldata it won't be compressed and will still occupy the full 32 bytes (and solc with add some additional sanity checks)
but i like the distinction since we have one variable being used as two different types in the same codebase, it makes sense to have the distinct types so it's always apparent which 'form' risk premium is in

src/contracts/LiquidityHub.sol
DhairyaSethi: yep good call, will add [8135465](https://github.com/aave/aave-v4/pull/183/commits/81354653aa06e699dfaa211fc744b88132db3080)

src/contracts/LiquidityHub.sol
yan-man: ah yes good call, that was the future change

src/contracts/AssetLogic.sol
yan-man: > can implicitly mean truncation

i'm good with this if others agree (and perhaps add some docs on it to specify)

CheyenneAtapour: My 2 cents here: converting shares and assets, there are only 2 possibilities, so it's very clear what you are converting to / from. Could even name it `truncateToBps()`. I think if the name of the function is crystal clear, we don't need to worry about adding additional comments, docs, or keeping in mind implicit conventions. Even if the conventions are clear to us, it's better that it's crystal clear to people outside our org that want to integrate, build on top, etc.

Additionally, the benefit of `truncateToBps()` and `toBoundedRay()` is that they both convey what the final "type" of the value is, which I personally want to know while following code

DhairyaSethi: 'truncateToBps()' isn't great bc the operation performs de-ray-ification. it works w/o any assumptions of the "unit" of original data (it does a.div(1e27) -> no bps values here) 

it's just about specifying how it does this, truncate or round. fromRay naturally means div by 1e27, rounding has additional logic (add 1 if remainder decimals were > 0.5 in the simplest case)

ofc agree that having immediate clarity is optimal

> converting shares and assets, there are only 2 possibilities

not really, you can have different rounding techniques while going from or the other (strict down, up, down if rem < 0.5 etc), or also truncate here. this is a separate tangent than what the original comparison was made to infer  

> both convey what the final "type" of the value is

ok so the idea is specifically not restrict the final type and more importantly, to add precision **rather than convert between "types"**. if that was the original motivation, the names & impl would be `bpsToRay()` and `rayToBps()` (pls read description for the reasoning, tldr is we don't get enough precision by converting)

CheyenneAtapour: Ok, yeah good points. By "final value" I just meant return value of the function, in that the suggested name might make it clearer that in this context, we're truncating to bps. But `fromRay` does make more sense as a generic library name, which is suitable here (because it's a library function)

src/contracts/WadRayMath.sol
```diff
@@ -166,77 +127,27 @@ library WadRayMath {
...
-    // to avoid overflow, b/RAD == a
-    assembly {
-      b := mul(a, RAD)
-
-      if iszero(eq(div(b, RAD), a)) {
-        revert(0, 0)
-      }
-    }
+  function toBoundedRay(uint32 a) internal pure returns (uint256) {
+    require(a <= 1000_00, 'INVALID_BPS');
```
CheyenneAtapour: What about just allowing 1000%, and not anything higher `a < 1000_00`? Nice round number imo. Then also change the max constant in tests below if so

tests/unit/LiquidityHub/LiquidityHubAccrueInterestTest.t.sol
```diff
@@ -223,17 +223,17 @@ contract LiquidityHubAccrueInterestTest is BaseTest {
...
 
   function test_accrueInterest_fuzz_RPBorrowAndElapsed(
     uint256 borrowAmount,
     uint40 elapsed,
-    uint256 riskPremium
+    uint32 riskPremium
   ) public {
     borrowAmount = bound(borrowAmount, 1, 1e30);
-    riskPremium = bound(riskPremium, 0, MAX_BPS.bpsToRad());
+    riskPremium %= MAX_RISK_PREMIUM_BPS;
```
CheyenneAtapour, line 236: Why not just bound between 0 and `MAX_RISK_PREMIUM_BPS`? We duplicate values with this approach since we test e.g. 1 twice with 1 and the value `MAX_RISK_PREMIUM_BPS + 1`. Also I think the bound is easier to read

tests/unit/LiquidityHub/LiquidityHubInterestRateTest.t.sol
```diff
@@ -234,46 +225,55 @@ contract LiquidityHubInterestRateTest is BaseTest {
...
 
     uint256 borrowRate = _getBorrowRate(daiAssetId);
     uint256 baseBorrowRate = _getBaseBorrowRate(daiAssetId);
-    uint256 newRp = hub.getAsset(daiAssetId).riskPremiumRad;
-    assertEq(borrowRate, baseBorrowRate + (newRp.radToRay().rayMul(baseBorrowRate)));
+    uint256 newRp = hub.getAsset(daiAssetId).riskPremium.fromRay();
+    uint256 calcRp = (rpSpoke1 * drawSpoke1 + rpSpoke2 * drawSpoke2 + rpSpoke3 * drawSpoke3) /
+      (drawSpoke1 + drawSpoke2 + drawSpoke3);
+
+    console.log('let rpSpoke1 = %sn; let drawSpoke1 = %sn', rpSpoke1, drawSpoke1);
```
CheyenneAtapour: Do we want to keep the console log in here? I imagine it was for debugging

tests/unit/LiquidityHub/LiquidityHubAccrueInterestTest.t.sol
DhairyaSethi: it achieves almost the same effect as bound, we will have those duplications at a similar frequency with bound as well (if you try to take a book at how it's implemented, it does logarithmic domain to range mapping, the simplest is 1-1 mapping with the modulus here)

wrt to readability i have added a `bound` method in `BaseTest` which works with `uint32`s but for some reason solidity complains and doesn't accept that override. i think this works haha

tests/unit/LiquidityHub/LiquidityHubInterestRateTest.t.sol
DhairyaSethi: ah yeah will rm [8135465](https://github.com/aave/aave-v4/pull/183/commits/81354653aa06e699dfaa211fc744b88132db3080)

src/contracts/AssetLogic.sol
miguelmtzinf: I see, what about `rayify()`  and `derayify()` or more explicitly `extendToRay()` and `reduceFromRay()`

src/contracts/Spoke.sol
miguelmtzinf: ok, agreed

tests/unit/LiquidityHub/LiquidityHubAccrueInterestTest.t.sol
CheyenneAtapour: ok cool, this works for me then

src/dependencies/openzeppelin/EnumerableSet.sol
miguelmtzinf: whatever you prefer :saluting_face:

src/interfaces/ILiquidityHub.sol
miguelmtzinf: that's why we need it in the natspec lol

src/contracts/WadRayMath.sol
miguelmtzinf: not following, this enforces max amount is 1000%, wdym here?

tests/BaseTest.t.sol
```diff
@@ -518,4 +519,8 @@ abstract contract BaseTest is Test, Events {
   function randomizer(uint256 min, uint256 max, uint256) internal returns (uint256) {
     return vm.randomUint(min, max);
   }
+
+  function bound(uint32 x, uint32 min, uint32 max) internal pure returns (uint32) {
```
miguelmtzinf: but you still have this helper @DhairyaSethi

tests/BaseTest.t.sol
```diff
@@ -88,6 +88,7 @@ abstract contract BaseTest is Test, Events {
   using SharesMath for uint256;
 
   uint256 internal constant MAX_SUPPLY_AMOUNT = 1e30;
+  uint32 internal constant MAX_RISK_PREMIUM_BPS = 999_99;
```
miguelmtzinf: this shouldn't be 1000_00?

tests/BaseTest.t.sol
DhairyaSethi: yeah there seems to be an issue with this override to work, haven't been able to debug the reason (https://github.com/aave/aave-v4/pull/183#discussion_r1961586229) so just using a simple modulus to bound, will rm this. we agreed modulus is fine 

removed edde2e83bcf7fab90cf0b47f48078f06ad4564d4

src/contracts/AssetLogic.sol
DhairyaSethi: nice one, ty! added with `rayify` edde2e83bcf7fab90cf0b47f48078f06ad4564d4

src/contracts/WadRayMath.sol
DhairyaSethi: ty for the subtle catch! fixed here edde2e83bcf7fab90cf0b47f48078f06ad4564d4
@miguelmtzinf it was changing `<=` to `<`

tests/BaseTest.t.sol
DhairyaSethi: yep fixed here edde2e83bcf7fab90cf0b47f48078f06ad4564d4

tests/unit/WadRayMath.t.sol
```diff
@@ -115,4 +115,24 @@ contract WadRayMathTests is Test {
...
+      vm.expectRevert();
+      w.rayify(a);
+    } else {
+      assertEq(w.rayify(a), a * w.RAY());
+      assertEq(w.rayify(a), b);
+    }
+  }
+
+  function test_derayify_fuzz(uint256 a) public {
+    assertEq(w.derayify(a), a / w.RAY());
```
miguelmtzinf, line 136: we could also work with castings right, `uint10(a)` here

DhairyaSethi: rn the library only implements for an uint256 input to derayify, solc will truncate uint256(uint16(a)) if a > 2 ** 16 -1. i'm not totally following the goal?

miguelmtzinf:
Approving although I didn't review the MathUtilsTest fully. Happy to merge asap as the diff is huge.

DhairyaSethi:
@miguelmtzinf ty! yeah it **extends** the existing suite by adding more accurate ceil bounds, precision checks, better at retaining fuzzer heuristics, new cases (alt sub & add). implementation is slightly opinionated, we can discuss async for improvements. i will raising another PR for some more other test ideas i have, but it is exhaustive now

PR #192, opened 2025-02-20
[PoC] DO NOT MERGE debt precision 

yan-man:
To see this issue, run `forge test --mc BorrowIndex_Scenario3 -vvv`. In the scenario test:

```
  // Scenario:
  // t0	asset added, spoke1 added
  // t1	spoke1 supply, spoke1 draw
  // t2	spoke4 added
  // t3	spoke4 draw
  // t4	spoke4 supply
  // t5 spoke1 repays all
  // t6 spoke4 repay <------- issue arises
```

at time t6, when spoke4 attempts to repay, a bug surfaces where asset cumulatedBaseDebt (1.000822143011744987e18) is less than spoke cumulatedBaseDebt (1.000822143011744988e18), reverting bc this is not allowed in the update of [wAvg](https://github.com/aave/aave-v4/blob/tmp/precision-errors/src/contracts/MathUtils.sol#L155). 
- reverts regardless of the restore amount that spoke4 tries to repay

Screenshot 2025-02-19 at 19 32 50

### DIFF
--- a/tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario3.t.sol
+++ b/tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario3.t.sol
@@ -31,7 +31,7 @@ contract BorrowIndex_Scenario3Test is BorrowIndexBase {
 
   function test_borrowIndexScenario3() public {
     state.assetId = wethAssetId;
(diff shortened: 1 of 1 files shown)


yan-man:
See https://github.com/aave/aave-v4/tree/fix/precision-errors-remaining for updates

yan-man:
discussed on 2.20.25 call

PR #196, opened 2025-02-21
Fix User RP Getter

### LINKED ISSUE #197: User RP Issues
Currently, internal accounting is storing the latest "actual" user rp within the internal accounting in `_users[user_address][reserve_id]`. That is, the latest user rp value is held in storage at the last-used `reserveId` However, the exposed getter currently on main allows to pick the `reserveId` to query the user rp.

This getter MUST be called with the latest `reserve_id` the user interacted with in order to retrieve the correct value.

Maybe there is another way we can handle this, or we should definitely clarify in the code that this is the case (add a dev comment)

We decided in the end:
* keep user rp in a separate structure/mapping because it not related to the reserve
* good to have a getter to calc rp on the fly, so its the most up-to-date value
* We must also accrue interest upstream for all relevant assets (liquidity hub side)

Example showing our current strategy for user risk premium is gameable:
1. you deposit a shitton of ETH
2. you borrow DAI at low risk premium
3. you deposit shitcoin and withdraw ETH
4. time flies and premium does not acummulate until user interacts with DAI again
5. After 1 yr, if we calculate outstanding premium of the user, would be huge
6. User is smart, they lower the rp whenever they want to interact with DAI.
7. User lowers RP
8. User repays DAI,  drastically reducing the in-flight outstanding premium

CheyenneAtapour:
User RP getter was not exactly correct, because the only up-to-date user RP is stored in that user's last reserve in the internal accounting. 

This fix changes to have User RP calculated on the fly and returned.

EDIT: However, we don't want user rp to be calculated on the fly and returned, we in fact do want to fetch the latest one stored (because user RP should be unchanging between user actions)

Closes #197

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -347,11 +347,19 @@ contract LiquidityHub is ILiquidityHub {
     return cumulatedBaseDebt + cumulatedOutstandingPremium;
   }
 
-  function getSuppliedAmount(uint256 assetId, address spoke) external view returns (uint256) {
+  function getAssetSuppliedAmount(uint256 assetId) external view returns (uint256) {
+    return _assets[assetId].convertToAssetsDown(_assets[assetId].suppliedShares);
+  }
+
+  function getAssetSuppliedShares(uint256 assetId) external view returns (uint256) {
+    return _assets[assetId].suppliedShares;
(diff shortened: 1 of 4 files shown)


DhairyaSethi:
this was fixed with #207

Issue #198, opened 2025-02-21
Adjust suppliedShares getters


yan-man:
Separate specific getters for assets/spokes/reserves/users: 

LH:
- getSuppliedAssetShares/getSuppliedAssetAmount
- getSuppliedSpokeShares/getSuppliedSpokeAmount

Spoke:
- getSuppliedReserveShares/getSuppliedReserveAmount
- getSuppliedUserShares/getSuppliedUserAmount

yan-man:
done when RP fix is merged: https://github.com/aave/aave-v4/pull/196. Edit: Dhairya used a different branch, so that PR was closed.

PR #199, opened 2025-02-21
fix: specific supplied values for asset/spoke/reserves/users

yan-man:
- rename methods for consistency
- debug spoke details: return newReserveRiskPremium, keep newReserveRiskPremium in ray 

closes https://github.com/aave/aave-v4/issues/198

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -484,7 +484,7 @@ contract Spoke is ISpoke {
     }
 
     if (collateralValue == 0) return 0;
-    return newUserRiskPremium.rayDiv(collateralValue);
+    return (newUserRiskPremium / collateralValue).rayify();
   }
 
   /// @dev It's assumed interest has been accrued before this function call.


src/contracts/Spoke.sol
```diff
@@ -511,6 +523,8 @@ contract Spoke is ISpoke {
 
     reserve.riskPremium = newReserveRiskPremium;
     user.riskPremium = newUserRiskPremium;
+
+    return newReserveRiskPremium;
```
yan-man, line 527: @DhairyaSethi @CheyenneAtapour note that this is also missing

src/contracts/Spoke.sol
```diff
@@ -472,7 +484,7 @@ contract Spoke is ISpoke {
     }
 
     if (collateralValue == 0) return 0;
-    return newUserRiskPremium / collateralValue;
+    return newUserRiskPremium.rayDiv(collateralValue);
```
yan-man: also this so newUserRiskPremium is in ray units

DhairyaSethi: nice catch! 
this was correct when it was written since the liquidity premiums above are in bps terms, so we wanted to retain this to be in bps. 

i agree, there's a bug where we don't rayify this value before using as a value in the global spoke rp 
so rayDiv adds the ray multiplier here but it's not consistent with how other risk premiums are calculated (bc it adds precision before). 

not sure of the cleanest soln, but if we (rp/sumWieghts).rayify() instead & then derayify() this value on usage inside spoke rp calc, that's most consistent with the flow

src/contracts/Spoke.sol
CheyenneAtapour: nice catch! wth, I didn't even think that would compile without it lol

yan-man: I realized internal methods will just return the default value

src/contracts/Spoke.sol
DhairyaSethi: feel free to, i don't have my laptop till monday

yan-man: fixed: https://github.com/aave/aave-v4/pull/199/commits/e49489ca1d07e251c559c82d03c37b74f33cd866

yan-man:
@DhairyaSethi feel free to merge or make whatever edits you need once approved, you're taking over this branch anyway right?

Issue #178, opened 2025-02-17
Spoke RP - Missing Supply Unit Tests


miguelmtzinf:
https://github.com/aave/aave-v4/pull/140 added the main tests for Spoke supply, but missing some:

- test supply reverts with 0 amount
- test supply with increased index and no premium (where sharesAmount < amount)
- test supply with increased increased index and premium

(above with time passing as well)

reference: https://github.com/aave/aave-v4/pull/140#discussion_r1950035078

Issue #168, opened 2025-02-13
Asset to Share conversion rounding


DhairyaSethi:
![Image](https://github.com/user-attachments/assets/0276dfdb-c560-46dd-a214-0d1cda051a62)

Issue #118, opened 2025-01-24
Spoke - Supply & Withdraw


CanonicalJP:
- Implement supply and withdraw function on the spoke using the risk premium new algorithm, interest accrual, assets and shares on spoke and on liquidity hub
- update return signatures on LH supply/withdraw based on spoke implementation

PR #200, opened 2025-02-21
Spoke supply/withdraw unit tests

yan-man:
- Spoke supply & withdraw unit tests
- includes changes from: https://github.com/aave/aave-v4/pull/199
- [x] update with changes from https://github.com/aave/aave-v4/pull/201 after it's merged

closes https://github.com/aave/aave-v4/issues/178
closes https://github.com/aave/aave-v4/issues/120
closes https://github.com/aave/aave-v4/issues/198
closes https://github.com/aave/aave-v4/issues/118
closes https://github.com/aave/aave-v4/issues/168

### DIFF
--- a/tests/unit/Spoke/Spoke.Withdraw.t.sol
+++ b/tests/unit/Spoke/Spoke.Withdraw.t.sol
@@ -618,7 +618,7 @@ contract SpokeWithdrawTest is SpokeBase {
     state.rate = params.rate;
     state.timestamp = vm.getBlockTimestamp();
 
-    (, state.supplyShares) = _executeSupplyAndBorrow({
+    (, state.supplyShares) = _executeSpokeSupplyAndBorrow({
       spoke: spoke1,
       collateral: TestReserve({
         reserveId: state.collateralReserveId,
@@ -635,6 +635,7 @@ contract SpokeWithdrawTest is SpokeBase {
         borrower: alice
       }),
       rate: state.rate,
+      isMockRate: true,
       skipTime: params.skipTime
     });
 
@@ -875,7 +876,7 @@ contract SpokeWithdrawTest is SpokeBase {
     state.rate = params.rate;
     state.timestamp = vm.getBlockTimestamp();
 
-    (, state.supplyShares) = _executeSupplyAndBorrow({
+    (, state.supplyShares) = _executeSpokeSupplyAndBorrow({
       spoke: spoke1,
       collateral: TestReserve({
         reserveId: state.collateralReserveId,
@@ -892,6 +893,7 @@ contract SpokeWithdrawTest is SpokeBase {
         borrower: alice
       }),
       rate: state.rate,
+      isMockRate: true,
       skipTime: params.skipTime
     });
 
--- a/tests/unit/Spoke/SpokeBase.t.sol
+++ b/tests/unit/Spoke/SpokeBase.t.sol
@@ -50,64 +50,44 @@ contract SpokeBase is Base {
   // increase share conversion index on given reserve
   // bob supplies borrow asset
   // alice supply (weth) collateral asset, borrow asset, skip 1 year to increase index
-  /// @return supplyShares of collateral asset
-  /// @return supplyShares of borrowed asset
+  /// @return supply amount of collateral asset
+  /// @return supply shares of collateral asset
+  /// @return borrow amount of borrowed asset
+  /// @return supply shares of borrowed asset
+  /// @return supply amount of borrowed asset
   function _increaseReserveIndex(
     Spoke spoke,
     uint256 reserveId
   ) internal returns (uint256, uint256, uint256, uint256, uint256) {
     SupplyBorrowLocal memory state;
-    TestReserve memory collateral;
-    TestReserve memory borrow;
 
+    TestReserve memory collateral;
     collateral.reserveId = wethReserveId(spoke);
     collateral.supplyAmount = 1_000e18;
     collateral.supplier = alice;
+
+    TestReserve memory borrow;
     borrow.reserveId = reserveId;
     borrow.supplier = bob;
     borrow.borrower = alice;
     borrow.supplyAmount = 100e18;
     borrow.borrowAmount = borrow.supplyAmount / 2;
-    (state.borrowReserveAssetId, ) = getAssetInfo(spoke, reserveId);
-    state.borrowSupplyShares = hub.convertToShares(state.borrowReserveAssetId, borrow.supplyAmount);
 
-    // supply collateral asset
-    Utils.spokeSupply({
+    (state.borrowReserveAssetId, ) = getAssetInfo(spoke, borrow.reserveId);
+    (state.collateralSupplyShares, state.borrowSupplyShares) = _executeSpokeSupplyAndBorrow({
       spoke: spoke,
-      reserveId: collateral.reserveId,
-      user: collateral.supplier,
-      amount: collateral.supplyAmount,
-      onBehalfOf: collateral.supplier
-    });
-    setUsingAsCollateral({
-      spoke: spoke,
-      user: collateral.supplier,
-      reserveId: collateral.reserveId,
-      usingAsCollateral: true
+      collateral: collateral,
+      borrow: borrow,
+      rate: 0,
+      isMockRate: false,
+      skipTime: 365 days
     });
 
-    // other user supplies enough asset to be drawn
-    Utils.spokeSupply({
-      spoke: spoke,
-      reserveId: reserveId,
-      user: borrow.supplier,
-      amount: borrow.supplyAmount,
-      onBehalfOf: borrow.supplier
-    });
-
-    // borrower borrows asset
-    Utils.spokeBorrow({
-      spoke: spoke,
-      reserveId: reserveId,
-      user: borrow.borrower,
-      amount: borrow.borrowAmount,
-      onBehalfOf: borrow.borrower
-    });
-
-    // skip time to increase index
-    skip(365 days);
-    (uint256 assetId, ) = getAssetInfo(spoke, reserveId);
-    assertGt(borrow.borrowAmount, hub.convertToShares(assetId, borrow.borrowAmount));
+    // index has increased, ie now the shares are less than the amount
+    assertGt(
+      borrow.supplyAmount,
+      hub.convertToShares(state.borrowReserveAssetId, borrow.supplyAmount)
+    );
 
     return (
       collateral.supplyAmount,
@@ -121,20 +101,23 @@ contract SpokeBase is Base {
   // supply collateral asset, borrow asset, skip time to increase index on borrow asset
   /// @return supplyShares of collateral asset
   /// @return supplyShares of borrowed asset
-  function _executeSupplyAndBorrow(
+  function _executeSpokeSupplyAndBorrow(
     Spoke spoke,
     TestReserve memory collateral,
     TestReserve memory borrow,
     uint256 rate,
+    bool isMockRate,
     uint256 skipTime
   ) internal returns (uint256, uint256) {
     SupplyBorrowLocal memory state;
 
-    vm.mockCall(
-      address(irStrategy),
-      IReserveInterestRateStrategy.calculateInterestRates.selector,
-      abi.encode(rate)
-    );
+    if (isMockRate) {
+      vm.mockCall(
+        address(irStrategy),
+        IReserveInterestRateStrategy.calculateInterestRates.selector,
+        abi.encode(rate)
+      );
+    }
 
     (state.collateralReserveAssetId, ) = getAssetInfo(spoke, collateral.reserveId);
     (state.borrowReserveAssetId, ) = getAssetInfo(spoke, borrow.reserveId);
--- a/tests/unit/Spoke/Spoke.Supply.t.sol
+++ b/tests/unit/Spoke/Spoke.Supply.t.sol
@@ -280,7 +280,7 @@ contract SpokeSupplyTest is SpokeBase {
     });
 
     // increase index on reserveId
-    _executeSupplyAndBorrow({
+    _executeSpokeSupplyAndBorrow({
       spoke: spoke1,
       collateral: TestReserve({
         reserveId: wethReserveId(spoke1),
@@ -297,6 +297,7 @@ contract SpokeSupplyTest is SpokeBase {
         borrower: alice
       }),
       rate: rate,
+      isMockRate: true,
       skipTime: skipTime
     });
 
@@ -460,7 +461,7 @@ contract SpokeSupplyTest is SpokeBase {
     (uint256 assetId, IERC20 asset) = getAssetInfo(spoke1, reserveId);
 
     // alice supplies usdx as collateral, borrows dai
-    _executeSupplyAndBorrow({
+    _executeSpokeSupplyAndBorrow({
       spoke: spoke1,
       collateral: TestReserve({
         reserveId: wethReserveId(spoke1),
@@ -477,6 +478,7 @@ contract SpokeSupplyTest is SpokeBase {
         supplier: bob
       }),
       rate: rate,
+      isMockRate: true,
       skipTime: skipTime
     });
 


tests/unit/Spoke/SpokeBaseTest.t.sol
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  // increase share conversion index on borrow asset
+  /// @return supplyShares of collateral asset
+  /// @return supplyShares of borrowed asset
+  function _increaseShareConversionIndex(
+    TestReserve memory collateral,
+    TestReserve memory borrow,
+    uint256 rate
+  ) internal returns (uint256, uint256) {
```
DhairyaSethi: how about `_executeSupplyAndBorrow` or something similar?

tests/unit/Spoke/SpokeBaseTest.t.sol
```diff
@@ -0,0 +1,133 @@
...
+    Utils.borrow({
+      spoke: spoke1,
+      reserveId: borrow.reserveId,
+      user: borrow.borrower,
+      amount: borrow.borrowAmount,
+      onBehalfOf: borrow.borrower
+    });
+
+    // skip time to increase index
+    skip(365 days);
```
DhairyaSethi: prefer if we don't hardcode time skip, include part of params such that this can be fuzzed as well in this helper method

tests/unit/Spoke/SpokeBaseTest.t.sol
```diff
@@ -0,0 +1,133 @@
...
+      abi.encode(rate)
+    );
+
+    Spoke.Reserve memory collateralReserve = spoke1.getReserve(collateral.reserveId);
+    Spoke.Reserve memory borrowReserve = spoke1.getReserve(borrow.reserveId);
+    uint256 collateralSupplyShares = hub.convertToShares(
+      collateralReserve.assetId,
+      collateral.supplyAmount
+    );
+    uint256 borrowSupplyShares = hub.convertToShares(borrowReserve.assetId, borrow.supplyAmount);
```
DhairyaSethi: i think it's worth adding simple asserts here, checking supply shares diff before & after either through balance (once we add 4626 integration) or getSuppliedShares getter to see we actually received this exact amount during the action

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+    });
+
+    skip(365 days);
+
+    // carol repays all with interest
+    uint256 repayAmount = spoke1.getUserCumulativeDebt(params.reserveId, carol);
+    deal(asset, carol, repayAmount);
+    vm.startPrank(carol);
+    spoke1.repay(params.reserveId, repayAmount);
+    vm.stopPrank();
```
DhairyaSethi:
```suggestion
    vm.prank(carol);
    spoke1.repay(params.reserveId, repayAmount);
```

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+    uint256 stage = 0;
+    reserveData[stage] = _getReserveData(params.reserveId);
+    aliceData[stage] = _getUserData(params.reserveId, alice);
+    bobData[stage] = _getUserData(params.reserveId, bob);
+    tokenData[stage] = _getTokenBalances(asset, address(spoke1));
+
+    vm.assume(
+      aliceData[stage].suppliedAmount > params.aliceAmount &&
+        hub.convertToShares(assetId, aliceData[stage].suppliedAmount) > 0
+    );
```
DhairyaSethi: could we force it such that this condition passes in the bounds above itself?

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+    assertApproxEqAbs(bobData[stage].data.suppliedShares, 0, 1, 'bobData supplied shares'); // 1 share difference
+    assertEq(
+      bobData[stage].data.lastUpdateTimestamp,
+      reserveData[stage].data.lastUpdateTimestamp,
+      'bobData last update timestamp'
+    );
+
+    // token
+    assertEq(tokenData[stage].spokeBalance, 0, 'tokenData spoke balance');
+    assertApproxEqAbs(tokenData[stage].hubBalance, 0, 1e1, 'tokenData hub balance'); // 1 amoutn difference for each user
```
DhairyaSethi: 1e1 is a 10 wei difference, not 1 amount. that was also a typo, i think?

```suggestion
    assertApproxEqAbs(tokenData[stage].hubBalance, 0, 1, 'tokenData hub balance'); // 1 amount difference for each user
```

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+    });
+    Utils.borrow({
+      spoke: spoke1,
+      reserveId: params.reserveId,
+      user: carol,
+      amount: params.borrowAmount,
+      onBehalfOf: carol
+    });
+
+    skip(365 days);
```
DhairyaSethi:
```suggestion
    skip(params.skipTime);
```

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+
+    stage = 1;
+    state.withdrawnShares = hub.convertToShares(daiAssetId, state.withdrawAmount);
+    reserveData[stage] = _getReserveData(state.reserveId);
+    aliceData[stage] = _getUserData(state.reserveId, alice);
+    bobData[stage] = _getUserData(state.reserveId, bob);
+    tokenData[stage] = _getTokenBalances(address(tokenList.dai), address(spoke1));
+
+    vm.prank(bob);
+    spoke1.withdraw({reserveId: state.reserveId, amount: state.withdrawAmount, to: bob});
```
DhairyaSethi, line 515: can probably do without named args here, and a comment that `bob` is `to` parameter bc there's other places we do in this test but nbd, fine either way

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+    assertEq(
+      IERC20(asset).balanceOf(address(bob)),
+      MAX_SUPPLY_AMOUNT - state.borrowReserveSupplyAmount + state.withdrawAmount,
+      'bob balance'
+    );
+  }
+
+  function _getAsset(Spoke spoke, uint256 reserveId) internal returns (address) {
+    return spoke.getReserve(reserveId).asset;
+  }
```
DhairyaSethi:
```suggestion
  function _getAsset(Spoke spoke, uint256 reserveId) internal returns (IERC20) {
    return IERC20(spoke.getReserve(reserveId).asset);
  }
```
to save on all casts to ierc20, just need to cast to address on all `deal`s which are fewer (and ideally shouldn't be needed in tests as all actors should have sufficient balance (max supply amount) dealt in base

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+      tokenList.dai.balanceOf(bob),
+      MAX_SUPPLY_AMOUNT - amount,
+      'bob dai balance pre-withdraw'
+    );
+
+    vm.startPrank(bob);
+    vm.expectEmit(address(spoke1));
+    emit Withdrawn(_daiReserveId(spoke1), amount, bob);
+    spoke1.withdraw(_daiReserveId(spoke1), amount, bob);
+    vm.stopPrank();
```
DhairyaSethi:
```suggestion
    vm.expectEmit(address(spoke1));
    emit Withdrawn(_daiReserveId(spoke1), amount, bob);
    vm.prank(bob);
    spoke1.withdraw(_daiReserveId(spoke1), amount, bob);
```

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+      IReserveInterestRateStrategy.calculateInterestRates.selector,
+      abi.encode(uint256(10_00).bpsToRay())
+    );
+  }
+
+  function test_withdraw_revertsWith_supplied_amount_exceeded_zero_supplied() public {
+    uint256 reserveId = _daiReserveId(spoke1);
+    uint256 amount = 1;
+
+    vm.prank(address(alice));
```
DhairyaSethi:
```suggestion
    vm.prank(alice);
```

DhairyaSethi: `+ all other unneeded casts below

tests/unit/Spoke/SpokeBaseTest.t.sol
```diff
@@ -0,0 +1,133 @@
...
+    );
+    userData.data.suppliedShares = spoke1.getUserSuppliedShares(reserveId, user);
+    userData.data.lastUpdateTimestamp = userStorageData.lastUpdateTimestamp;
+    userData.data.riskPremium = spoke1.getUserRiskPremium(user);
+    userData.suppliedAmount = spoke1.getUserSuppliedAmount(reserveId, user);
+    return userData;
+  }
+
+  function _getTokenBalances(
+    address token,
```
DhairyaSethi: could pass as IERC20 only imo

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1040 @@
...
+    bobData[stage] = _getUserData(state.reserveId, bob);
+    tokenData[stage] = _getTokenBalances(address(tokenList.dai), address(spoke1));
+    daiData[stage].cumulatedBaseInterest = MathUtils.calculateLinearInterest(
+      state.rate,
+      uint40(timestamp)
+    );
+
+    state.withdrawAmount = hub.getAvailableLiquidity(daiAssetId); // withdraw all liquidity
+
+    assertTrue(
```
DhairyaSethi: these can be assertGt as well but nbd

tests/unit/Spoke/Spoke.Withdraw.t.sol
DhairyaSethi: the time skip in these tests is just to assert lastUpdateTimestamp updated properly? we're not asserting accrued interest etc?

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: prefer the named args generally also so that param order doesn't matter, so I'll leave this one

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: good catch. Actually this error is not merely 1 wei, but can be quite large bc it is constrained by the exchange rate. Because withdraw uses amount rather than shares, the amount -> share conversion can end up with a user having 1 share remaining even after a "full" withdrawal (bc of rounding in favor of protocol). So this abs err is actually: 

2 (bc 2 users in this fuzz test, each 1 share off) * 1 share (in amount) + 1

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: the time skip in these tests is just to accrue some interest to change the amount withdrawable to check the withdraw action. Figured not to assert on the accrued interest as it is outside the scope of the test (ie would be checked in accrual-specific tests) 

but def will add skipTime as a fuzz param: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: I think technically yes, but it's a pain bc essentially it needs to calculate the new exch rate based on all the actions, including the debt, premium accrual, etc. Would prefer not to need to do that.

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/SpokeBaseTest.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: ig other option is to withdraw with shares as input param on both LH/spoke. Worse for UX bc they need to calc shares amount, but better for their outcome to get full withdrawal. 

Or smt like amount as input param to spoke (with uint256 max input as "withdraw all") / shares as input param to LH. downside is mismatch between spoke/LH inputs

or leave as is, where dust collects 

cc @CheyenneAtapour @miguelmtzinf

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/SpokeBaseTest.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/SpokeBaseTest.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/03482d9db766bac80740c6a3d1bf39a43e4db112

tests/unit/Spoke/SpokeBaseTest.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/e24423c8829b9ff53817f2e0265a19f526bc1fc2

tests/Base.t.sol
```diff
@@ -548,4 +554,18 @@ abstract contract Base is Test {
...
+    return (reserve.assetId, IERC20(reserve.asset));
+  }
+
+  function _getWithdrawalLimit(
+    Spoke spoke,
+    uint256 reserveId,
+    address user
+  ) internal view returns (uint256) {
+    (uint256 baseDebt, ) = spoke1.getUserDebt(_daiReserveId(spoke1), user);
+    spoke.getUserSuppliedAmount(reserveId, user) - baseDebt;
```
```suggestion
    return spoke.getUserSuppliedAmount(reserveId, user) - baseDebt;
```
we can't withdraw more than outstanding premium? why is that also not subtracted?

tests/unit/Spoke/Spoke.Withdraw.t.sol
DhairyaSethi: yeah agree, since it's rejecting reasonable amount of fuzz runs it's fine. i was just curious if there's a clean workaround condition but haven't been able to come up with one

tests/unit/Spoke/Spoke.Withdraw.t.sol
DhairyaSethi: > a user having 1 share remaining even after a "full" withdrawal (bc of rounding in favor of protocol)

this means we're rounding **in favour of the user** tho, and is this not an exploit if i have one share remaining which i can keep redeeming w/o loosing that one share until the pool is empty?

miguelmtzinf: that should not happen. The user cannot end up with 1 share if they max withdrew.

src/contracts/Spoke.sol
```diff
@@ -361,7 +372,10 @@ contract Spoke is ISpoke {
     DataTypes.UserConfig storage user,
     uint256 amount
   ) internal view {
-    uint256 suppliedAmount = liquidityHub.convertToAssetsDown(reserve.assetId, user.suppliedShares);
+    uint256 suppliedAmount = liquidityHub.convertToAssetsDown(
+      reserve.assetId,
+      user.suppliedShares
+    ) - user.baseDebt;
```
miguelmtzinf: this is incorrect. We shouldn't take `debt` into account here. Only one validation is that it cannot withdraw more than supplied.

tests/Base.t.sol
```diff
@@ -11,21 +11,28 @@ import {WadRayMath} from 'src/contracts/WadRayMath.sol';
...
+import {WETH9} from 'src/dependencies/weth/WETH9.sol';
 
 abstract contract Base is Test {
   using WadRayMath for uint256;
   using SharesMath for uint256;
 
   uint256 internal constant MAX_SUPPLY_AMOUNT = 1e30;
   uint32 internal constant MAX_RISK_PREMIUM_BPS = 1000_00;
+  uint256 internal constant MAX_BORROW_RATE = 1000_00; // matches DefaultReserveInterestRateStrategy
+  uint256 internal constant MAX_SKIP_TIME = 10_000 days;
```
miguelmtzinf, line 35: should we make this `uint40`?

DhairyaSethi: imo fine w/o also bc this is used in 'skip's only, lastUpdateTime is used in calcLinearInterest

tests/Base.t.sol
```diff
@@ -485,11 +499,36 @@ abstract contract Base is Test {
     spoke.updateReserveConfig(reserveId, reserveData.config);
   }
 
+  function updateLiquidityPremium(
+    Spoke spoke,
+    uint256 reserveId,
+    uint256 newLiquidityPremium
+  ) internal {
+    DataTypes.Reserve memory reserveData = spoke.getReserve(reserveId);
```
miguelmtzinf: we can access config directly here (`DataTypes.ReserveConfig`)

tests/Base.t.sol
```diff
@@ -485,11 +499,36 @@ abstract contract Base is Test {
...
+    reserveData.config.liquidityPremium = newLiquidityPremium;
+    spoke.updateReserveConfig(reserveId, reserveData.config);
+  }
+
   /// @dev pseudo random randomizer
   function randomizer(uint256 min, uint256 max, uint256) internal returns (uint256) {
     return vm.randomUint(min, max);
   }
 
+  function _usdxReserveId(Spoke spoke) internal view returns (uint256) {
```
miguelmtzinf: nbd for now, but this means that every spoke always support these reserves. We ll probably need to refactor later and init 2-3 diff spokes and add a diff set of reserves to each one

tests/Base.t.sol
```diff
@@ -509,10 +548,24 @@ abstract contract Base is Test {
     DataTypes.UserConfig memory userConfig;
```
miguelmtzinf: this getter is a bit weird, because returns the same as `Spoke::getUser()` but with up-to-date values.
I think we must rename to `loadUserInfo` or just use spoke function directly.

tests/Base.t.sol
miguelmtzinf: as stated in the other comment, do not agree with this

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -0,0 +1,179 @@
...
+    uint256 reserveBaseDebtBefore;
+    uint256 borrowerBaseDebtAfter;
+    uint256 reserveBaseDebtAfter;
+  }
+
+  // increase share conversion index on borrow asset
+  // supply collateral asset, borrow asset, skip time to increase index
+  /// @return supplyShares of collateral asset
+  /// @return supplyShares of borrowed asset
+  function _executeSupplyAndBorrow(
```
miguelmtzinf: 1. do we just want to increase the exchange ratio? If yes, in favor of making this fn way more generic
just passing the asset whose index must inflate (debt asset)
2. Use `ISpoke` instead pls
3. why IR mocking is needed?

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -0,0 +1,179 @@
...
+    assertEq(state.reserveBaseDebtBefore + borrow.borrowAmount, state.reserveBaseDebtAfter);
+
+    // skip time to increase index
+    skip(skipTime);
+
+    return (state.collateralSupplyShares, state.borrowSupplyShares);
+  }
+
+  function _getReserveData(Spoke spoke, uint256 reserveId) internal view returns (TestData memory) {
+    // only to check lastUpdateTimestamp
```
miguelmtzinf: what does this mean?
here again loading up-to-date values into `DataTypes.ReserveData` which could be misleading

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -0,0 +1,179 @@
...
+    return reserveData;
+  }
+
+  function _getUserData(
+    Spoke spoke,
+    uint256 reserveId,
+    address user
+  ) internal view returns (TestData memory) {
+    // only to check lastUpdateTimestamp
+    DataTypes.UserConfig memory userStorageData = spoke.getUser(reserveId, user);
```
miguelmtzinf: same

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1106 @@
...
+
+contract SpokeWithdrawTest is SpokeBase {
+  using WadRayMath for uint256;
+  using PercentageMath for uint256;
+
+  function setUp() public override {
+    super.setUp();
+
+    // mock constant 10% IR by default
+    vm.mockCall(
```
miguelmtzinf: why we do need this and cannot use the already configured in `configureTokenList`

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1106 @@
...
+      address(irStrategy),
+      IReserveInterestRateStrategy.calculateInterestRates.selector,
+      abi.encode(uint256(10_00).bpsToRay())
+    );
+  }
+
+  function test_withdraw_revertsWith_supplied_amount_exceeded_zero_supplied() public {
+    uint256 reserveId = _daiReserveId(spoke1);
+    uint256 amount = 1;
+
```
miguelmtzinf, line 13: we can add an assert to validate `suppliedShares` or amount is 0

src/contracts/Spoke.sol
CheyenneAtapour: The remaining validation checks should come from Health Factor, right?

DhairyaSethi: to get the withdrawable amount, we'd need "diff" bw this & total borrowed value in base currency (they need to be normalised bc we can have more than one debt position open)?

tests/unit/Spoke/Spoke.Withdraw.t.sol
CheyenneAtapour: I think input could be shares amount to avoid this, and also can be calculated properly on UI side, so user does not experience worse UX

src/contracts/Spoke.sol
CheyenneAtapour: @DhairyaSethi I thought that calc also comes from HF

DhairyaSethi: withdrawable calc? how do you mean?

tests/unit/Spoke/SpokeBase.t.sol
CheyenneAtapour: We need to add back a getter for the timestamp then. It got removed with the interface cleanup

DhairyaSethi: no @CheyenneAtapour idt that's the concern here, getting lastUpdateTimestamp from `getReserve` is fine & expected. `getReserveRiskPremium` & `getReserveDebt` fetch up to date values which are being loaded in `DataTypes.Reserve`
tbh it's fine also imo, maybe we can just use a custom test struct for this to avoid any possible confusion

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -0,0 +1,1106 @@
...
+    assertEq(
+      bobData[stage].data.lastUpdateTimestamp,
+      vm.getBlockTimestamp(),
+      'bobData last update timestamp'
+    );
+
+    // token
+    assertEq(tokenData[stage].spokeBalance, 0, 'tokenData spoke balance');
+    assertEq(tokenData[stage].hubBalance, 0, 'tokenData hub balance');
+    // assertEq(asset.balanceOf(alice), 0, 'alice balance');
```
CheyenneAtapour: If Alice has some dust amounts for the reason commented above in the deal, can just ensure the difference is the repay amount

src/contracts/Spoke.sol
CheyenneAtapour: If you are able to supply assets, borrow against them, then withdraw all your supplied assets, that would be stealing from the protocol (because now you have no loss potential for not repaying your borrowed assets), so you should be only be able to withdraw up to HF = 1, right?

tests/unit/Spoke/SpokeBase.t.sol
yan-man: > what does this mean?

that it's calling `getReserve` rather than a specific getter to read the `lastUpdateTimestamp` directly, bc we had a convention to use getters to access other fields. Will update comments on this

> here again loading up-to-date values into DataTypes.ReserveData which could be misleading

will rename this for clarity, for user too 

edit - fixed: https://github.com/aave/aave-v4/commit/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/unit/Spoke/SpokeBase.t.sol
yan-man: > Use ISpoke instead pls 

Yes will update.

For the other 2 points, I wanted to give flexibility when calling this method so that it can be fuzzed with inputs as well. If it's more generic do you mean that the other params are hard coded rather than passed in?

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: ah good catch here, I commented out to test something. Need to add this back in 

fixed: https://github.com/aave/aave-v4/pull/200/commits/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: edit - fixed: https://github.com/aave/aave-v4/pull/200/commits/09b6be890ddafe7fc8250f68dc10a502a8a63e04

src/contracts/Spoke.sol
yan-man: > We shouldn't take debt into account here

will update this for spoke and LH (where it is also done) 

edit - fixed: https://github.com/aave/aave-v4/commit/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/unit/Spoke/SpokeBase.t.sol
miguelmtzinf: yeah, if the whole point is inflating the ex rate and those are not needed

yan-man: ok I'll create a separate method that's more generic for just inflating exch rate, and leave this one so that it can be fuzzed

edit - 
fixed: https://github.com/aave/aave-v4/commit/09b6be890ddafe7fc8250f68dc10a502a8a63e04
optimized: https://github.com/aave/aave-v4/pull/200/commits/ab62d24c7269a49fa4e69f474466bd557543c2a7

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: > is this not an exploit if i have one share remaining which i can keep redeeming

no not exactly, because after Alice redeems her "full" amount her supplied shares = 1 but supplied amount = 1, and she cannot retrieve that share bc it fails on validation

yan-man: discussed with Emilio on 2.28.25 call - decision is to leave rounding as is on supply, but on withdraw (in LH) round up from amount -> shares 

edit - fixed: https://github.com/aave/aave-v4/commit/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/Base.t.sol
yan-man: left this as is bc as Dhairya mentioned it is only used in skips

tests/Base.t.sol
yan-man: added some comments to clarify this assumption 

edit - fixed: https://github.com/aave/aave-v4/commit/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/unit/Spoke/SpokeBase.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/200/commits/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/Base.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/commit/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/Base.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/commit/09b6be890ddafe7fc8250f68dc10a502a8a63e04

tests/Base.t.sol
yan-man: edit - see fix below

CheyenneAtapour: Is it fixed? I thought we don't want to subtract out base debt. And it is also missing the return as @DhairyaSethi pointed out

yan-man: thanks @CheyenneAtapour good catch. Had only updated the spoke validation impl there. Fixed test helper properly here: https://github.com/aave/aave-v4/pull/200/commits/4e5b09885b9938524102ff01a62cd4b297b3c88b

PR #201, opened 2025-02-24
Spoke Borrow Tests + Custom Errors + Interface Consolidation 

### LINKED ISSUE #125: Spoke RP - Borrow unit testing

CheyenneAtapour:
Adds unit tests for `borrow` function on the spoke
Adds custom errors
Creates events for all relevant functions
Adds remaining external functions to ISpoke and ILiquidityHub interfaces
Moves events, errors, and remaining functions to interfaces
Moves all structs to `DataTypes`

closes 
#125 
#202 
#212 
#213

### DIFF
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -20,12 +20,7 @@ import {MockERC20} from './mocks/MockERC20.sol';
 import {MockPriceOracle, IPriceOracle} from './mocks/MockPriceOracle.sol';
 import {Utils} from './Utils.sol';
 
-contract Events {
-  // OpenZeppelin
-  event Transfer(address indexed from, address indexed to, uint256 value);
-}
-
-abstract contract Base is Test, Events {
+abstract contract Base is Test {
   using WadRayMath for uint256;
   using SharesMath for uint256;
 
--- a/tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
@@ -36,9 +36,10 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
 
   function test_supply_revertsWith_supply_cap_exceeded() public {
     uint256 amount = 100e18;
-    _updateSupplyCap(daiAssetId, address(spoke1), amount - 1);
+    uint256 newSupplyCap = amount - 1;
+    _updateSupplyCap(daiAssetId, address(spoke1), newSupplyCap);
 
-    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.SupplyCapExceeded.selector, amount - 1));
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.SupplyCapExceeded.selector, newSupplyCap));
     vm.prank(address(spoke1));
     hub.supply(daiAssetId, amount, 0, alice);
   }
@@ -63,7 +64,7 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
 
     vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.SupplyCapExceeded.selector, newSupplyCap));
     vm.prank(address(spoke2));
-    hub.supply(daiAssetId, 1e8, 0, alice);
+    hub.supply(daiAssetId, 1, 0, alice);
   }
 
   function test_supply() public {
@@ -171,7 +172,7 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     IERC20 asset = hub.assetsList(assetId);
 
     vm.expectEmit(address(asset));
-    emit Transfer(alice, address(hub), amount);
+    emit IERC20.Transfer(alice, address(hub), amount);
     vm.expectEmit(address(hub));
     emit ILiquidityHub.Supply(assetId, address(spoke1), amount);
 
@@ -247,7 +248,7 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     IERC20 asset2 = hub.assetsList(assetId2);
 
     vm.expectEmit(address(asset));
-    emit Transfer(alice, address(hub), amount);
+    emit IERC20.Transfer(alice, address(hub), amount);
     vm.expectEmit(address(hub));
     emit ILiquidityHub.Supply(assetId, address(spoke1), amount);
 
@@ -255,7 +256,7 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.supply(assetId, amount, 0, alice);
 
     vm.expectEmit(address(asset2));
-    emit Transfer(alice, address(hub), amount2);
+    emit IERC20.Transfer(alice, address(hub), amount2);
     vm.expectEmit(address(hub));
     emit ILiquidityHub.Supply(assetId2, address(spoke2), amount2);
 
--- a/tests/unit/LiquidityHub/LiquidityHub.Withdraw.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.Withdraw.t.sol
@@ -78,7 +78,7 @@ contract LiquidityHubWithdrawTest is LiquidityHubBase {
     );
 
     vm.expectEmit(address(tokenList.dai));
-    emit Transfer(address(hub), alice, amount);
+    emit IERC20.Transfer(address(hub), alice, amount);
     vm.expectEmit(address(hub));
     emit ILiquidityHub.Withdraw(daiAssetId, address(spoke1), alice, amount);
 
@@ -330,7 +330,7 @@ contract LiquidityHubWithdrawTest is LiquidityHubBase {
     );
 
     vm.expectEmit(address(asset));
-    emit Transfer(address(hub), alice, amount);
+    emit IERC20.Transfer(address(hub), alice, amount);
 
     vm.expectEmit(address(hub));
     emit ILiquidityHub.Withdraw(assetId, address(spoke1), alice, amount);


tests/unit/Spoke/SpokeTest.Borrow.t.sol
```diff
@@ -0,0 +1,377 @@
...
+
+    // Alice supply dai
+    Utils.spokeSupply(spoke1, daiReserveId, alice, daiAmount, alice);
+
+    Spoke.UserConfig memory bobData = _getUserSpokeInfo(spoke1, bob, wethReserveId);
+    Spoke.UserConfig memory aliceData = _getUserSpokeInfo(spoke1, alice, daiReserveId);
+
+    assertEq(
+      bobData.suppliedShares,
+      hub.convertToSharesDown(wethAssetId, wethAmount),
```
yan-man: think we should use the agnostic conversion here without rounding, ie `convertToShares`

tests/unit/Spoke/SpokeTest.Borrow.t.sol
```diff
@@ -0,0 +1,377 @@
...
+    assertEq(aliceData.baseDebt, 0, 'alice base debt pre-draw');
+    assertEq(tokenList.dai.balanceOf(address(spoke1)), 0, 'spoke1 dai balance pre-draw');
+    assertEq(tokenList.weth.balanceOf(address(spoke2)), 0, 'spoke2 weth balance pre-draw');
+    assertEq(tokenList.dai.balanceOf(bob), 0, 'bob dai balance pre-draw');
+    assertEq(tokenList.weth.balanceOf(alice), 0, 'alice weth balance pre-draw');
+
+    // Bob draw dai
+    vm.prank(bob);
+    vm.expectEmit(address(spoke1));
+    emit Borrowed(spokeInfo[spoke1].dai.reserveId, daiBorrowAmount, bob);
```
yan-man: re-use `daiReserveId` here and in remaining part of test

tests/unit/UserRiskPremiumTest.t.sol
yan-man, line 1: rename to UserRiskPremium.t.sol (remove Test), as per: https://www.notion.so/V4-Code-Conventions-17a9d63a22de80898fb0ee1f6df6038b?d=19e9d63a22de8055b310001c8a67b735&pvs=4#18a9d63a22de8001a3d2d542d8da828b

same on other test file names

tests/BaseTest.t.sol
```diff
@@ -517,4 +517,27 @@ abstract contract BaseTest is Test, Events {
...
+
+  function _getUserSpokeInfo(
+    Spoke spoke,
+    address user,
+    uint256 reserveId
+  ) internal view returns (DataTypes.UserConfig memory) {
+    DataTypes.UserConfig memory userConfig;
+    userConfig.suppliedShares = spoke.getSuppliedShares(reserveId, user);
+    (userConfig.baseDebt, ) = spoke.getUserDebt(reserveId, user);
+    return userConfig;
```
miguelmtzinf, line 516: why this only populates a couple of variables within the struct

tests/BaseTest.t.sol
```diff
@@ -517,4 +517,27 @@ abstract contract BaseTest is Test, Events {
...
+  }
+
+  function _updateActive(uint256 assetId, bool newActive) internal {
```
miguelmtzinf: I d rename to `_updateReserveActive` or `_setReserveActive`

tests/Utils.t.sol
```diff
@@ -3,7 +3,8 @@ pragma solidity ^0.8.0;
 
 import {Vm} from 'forge-std/Vm.sol';
 import {LiquidityHub, DataTypes} from 'src/contracts/LiquidityHub.sol';
-import {Spoke, ISpoke} from 'src/contracts/Spoke.sol';
+import {Spoke} from 'src/contracts/Spoke.sol';
```
miguelmtzinf: we should be able to use interfaces only

tests/Utils.t.sol
```diff
@@ -102,19 +102,19 @@ library Utils {
   }
 
   function updateLiquidationThreshold(Spoke spoke, uint256 reserveId, uint256 newLt) internal {
```
miguelmtzinf: why these helpers are here and the rest in `BaseTest`?

tests/unit/Spoke/SpokeTest.Borrow.t.sol
```diff
@@ -0,0 +1,377 @@
...
+    uint256 wethReserveId = spokeInfo[spoke1].weth.reserveId;
+
+    uint256 daiAmount = 100e18;
+    uint256 wethAmount = 10e18;
+
+    // Bob supply weth
+    Utils.spokeSupply(spoke1, wethReserveId, bob, wethAmount, bob);
+
+    // Alice supply dai
+    Utils.spokeSupply(spoke1, daiReserveId, alice, daiAmount, alice);
```
miguelmtzinf: i think these prep may not be needed

tests/unit/Spoke/SpokeTest.Borrow.t.sol
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  function test_borrow() public {
+    uint256 daiReserveId = spokeInfo[spoke1].dai.reserveId;
+    uint256 wethReserveId = spokeInfo[spoke1].weth.reserveId;
+
+    uint256 daiAmount = 100e18;
+    uint256 wethAmount = 10e18;
+
+    // Reset account balances
```
miguelmtzinf: you do this because you don't wanna calculate balances delta?

tests/unit/Spoke/SpokeTest.Borrow.t.sol
```diff
@@ -0,0 +1,377 @@
...
+  function test_borrow_fuzz_amounts(uint256 wethSupplyAmount, uint256 daiBorrowAmount) public {
+    uint256 daiReserveId = spokeInfo[spoke1].dai.reserveId;
+    uint256 wethReserveId = spokeInfo[spoke1].weth.reserveId;
+
+    wethSupplyAmount = bound(wethSupplyAmount, 1, MAX_SUPPLY_AMOUNT);
+    daiBorrowAmount = bound(daiBorrowAmount, 1, wethSupplyAmount / 2 + 1);
+
+    // Reset account balances
+    deal(address(tokenList.dai), bob, 0);
+    deal(address(tokenList.weth), alice, 0);
```
miguelmtzinf: removing this feels more generic and thus, preferable for a fuzz test

tests/unit/Spoke/SpokeTest.Borrow.t.sol
CheyenneAtapour: Dunno how this slipped. Thanks, good eye!

tests/Utils.t.sol
CheyenneAtapour: Some helpers in `LiquidityHubBaseTest` were useful in spoke tests, so I moved them to `BaseTest`. But I think better they are in Utils. Moving them

tests/BaseTest.t.sol
CheyenneAtapour: Done. Was actually supposed to be for `asset`, so changed to `updateAssetActive`

tests/unit/Spoke/SpokeTest.Borrow.t.sol
CheyenneAtapour: Was just trying to match structure of analogous tests. But I can calc delta

src/interfaces/ISpoke.sol
```diff
@@ -1,6 +1,13 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
-import {DataTypes} from '../libraries/types/DataTypes.sol';
+
+import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+
+/**
+ * @title ILiquidityHub
+ * @author Aave Labs
+ * @notice Basic interface for LiquidityHub
```
```suggestion
 * @title ISpoke
 * @author Aave Labs
 * @notice Basic interface for Spoke
```

src/libraries/types/DataTypes.sol
```diff
@@ -2,14 +2,28 @@
...
+    uint256 baseDebt; // asset
+    uint256 outstandingPremium; // asset
+    uint256 baseBorrowIndex; // in ray
+    uint256 baseBorrowRate; // in ray
+    uint256 riskPremium; // weighted average risk premium of all spokes with ray precision
+    uint256 lastUpdateTimestamp;
+    DataTypes.AssetConfig config;
   }
 
   // TODO: borrow cap per spoke
```
miguelmtzinf:
```suggestion

```

tests/Utils.t.sol
```diff
@@ -101,21 +107,52 @@ library Utils {
...
+    uint256 assetId,
+    address spoke,
+    uint256 newDrawCap
+  ) internal {
+    DataTypes.SpokeConfig memory spokeConfig = hub.getSpokeConfig(assetId, spoke);
+    spokeConfig.drawCap = newDrawCap;
+    hub.updateSpokeConfig(assetId, spoke, spokeConfig);
+  }
+
+  function getUserSpokeInfo(
```
miguelmtzinf: is this not just `getUserInfo`?

tests/mocks/IEventsMock.sol
miguelmtzinf, line 1: This can be removed now, as everything can be fetched from interfaces

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  function test_borrow() public {
+    uint256 daiReserveId = spokeInfo[spoke1].dai.reserveId;
+    uint256 wethReserveId = spokeInfo[spoke1].weth.reserveId;
+
+    uint256 daiAmount = 100e18;
+    uint256 wethAmount = 10e18;
+
+    uint256 bobStartingBalance = tokenList.dai.balanceOf(bob);
```
miguelmtzinf: no big deal, but afaik we usually use `bobBalanceBefore` and `bobBalanceAfter`
can you check if this is consistent with other tests? or all is mixed up

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -0,0 +1,339 @@
...
+    uint256 aliceStartingBalance = tokenList.weth.balanceOf(alice);
+
+    // Bob supply weth
+    Utils.spokeSupply(spoke1, wethReserveId, bob, wethAmount, bob);
+
+    // Alice supply dai
+    Utils.spokeSupply(spoke1, daiReserveId, alice, daiAmount, alice);
+
+    DataTypes.UserConfig memory bobData = Utils.getUserSpokeInfo(spoke1, bob, wethReserveId);
+    DataTypes.UserConfig memory aliceData = Utils.getUserSpokeInfo(spoke1, alice, daiReserveId);
```
miguelmtzinf: can we fetch data for both reserves pls
same in the fuzz test

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: ha think this one is all mixed up

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -0,0 +1,339 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/Base.t.sol';
+
+contract SpokeBorrowTest is Base {
+  function setUp() public override {
+    super.setUp();
+    super.initEnvironment();
```
yan-man: minor but shouldn't need the super:
```suggestion
    initEnvironment();
```

src/interfaces/ILiquidityHub.sol
```diff
@@ -109,3 +123,25 @@ interface ILiquidityHub {
   event Restore(uint256 indexed assetId, address indexed spoke, uint256 amount);
   event SpokeAdded(uint256 indexed assetId, address indexed spoke);
 }
+
+/**
+ * @title Liquidity Hub Errors Library
+ * @author Aave Labs
+ * @notice Defines the error messages emitted by the LiquidityHub
+ */
+library LiquidityHubErrors {
```
DhairyaSethi: this can just be a part of ILiquidityHub if we switch to custom errors here (along with events)

tests/Utils.t.sol
DhairyaSethi, line 1: Utils.sol
doesn't need to be `.t.sol` since it doesn't include any tests

tests/Utils.t.sol
```diff
@@ -101,21 +107,52 @@ library Utils {
...
     reserveData.config.borrowable = newBorrowable;
     spoke.updateReserveConfig(reserveId, reserveData.config);
   }
+
+  /// @dev pseudo random randomizer
+  function randomizer(uint256 min, uint256 max, uint256) internal returns (uint256) {
+    return vm.randomUint(min, max);
+  }
+
+  function updateDrawCap(
```
DhairyaSethi: it's ok & probably better is these are in BaseTest itself

src/interfaces/ILiquidityHub.sol
```diff
@@ -109,3 +123,25 @@ interface ILiquidityHub {
   event Restore(uint256 indexed assetId, address indexed spoke, uint256 amount);
   event SpokeAdded(uint256 indexed assetId, address indexed spoke);
 }
```
DhairyaSethi: should add all relevant events to `ILiquidityHub.sol` imo

(same for spoke)

DhairyaSethi:
probably should do the event/interface/error consolidation in a separate PR into main? bc this also includes Spoke tests which need to be refactored

tests/unit/Spoke/Spoke.Borrow.t.sol
CheyenneAtapour: Ok cool, I like this convention. I will change it here at least and also noted it on our V4 conventions

tests/unit/Spoke/Spoke.Borrow.t.sol
CheyenneAtapour: I added this for completeness, but Alice doesn't even touch weth in the fuzz test, so I omitted checking her balance (was getting stack too deep otherwise)

tests/Utils.t.sol
CheyenneAtapour: How do we distinguish which should go in BaseTest vs Utils? BaseTest also imports Utils

DhairyaSethi: till now the distinction i've made is any helper for general stuff (and those which involve mocking) be in baseTest, and Utils can have helpers which are multicall abstractions for external user facing stuff (and ideally can be used as a public helper lib after removing vm.pranks or can be used by integrators)

DhairyaSethi: cc @yan-man for input

yan-man: ha was going to ask this too. ig for Utils it can be more standalone and generic than BaseTest. BaseTest more specific for our testing. 

> can be used as a public helper lib

agree with this. But tbh lots of overlap. Maybe we default to putting as much as possible into BaseTest and later can extract into Utils whatever we feel is helpful to be externally used?

DhairyaSethi: we've never had consensus this fast lol. 
ok so thumb rule for now: let's keep supply/withdraw/draw/restore helpers (for hub, spoke) on Utils + addReserve, addAsset, updateReserve, and everything else in BaseTest. later on, if needed, we can move stuff to Utils

src/contracts/LiquidityHub.sol
```diff
@@ -102,7 +56,7 @@ contract LiquidityHub is ILiquidityHub {
     });
     assetCount++;
 
-    // TODO: emit event
+    emit AssetAdded(assetCount - 1, asset);
```
DhairyaSethi: let's either cache assetCount or remove the ++ above and do it here 

```suggestion
    emit AssetAdded(assetCount++, asset);
```

(this still emits with the correctCount)

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -0,0 +1,388 @@
...
+      'bob supply shares final balance'
+    );
+    assertEq(bobData.baseDebt, usdxBorrowAmount, 'bob base debt usdx final balance');
+    bobData = Utils.getUserInfo(spoke2, bob, wbtcReserveId);
+    assertEq(
+      bobData.suppliedShares,
+      hub.convertToShares(wbtcAssetId, MAX_SUPPLY_AMOUNT),
+      'bob supply shares final balance'
+    );
+    assertEq(bobData.baseDebt, wbtcBorrowAmount, 'bob base debt wbtc final balance');
```
yan-man, line 371: don't want to block merging this bc it is a large amount of changes. But some additional borrow tests to add after that might be good:
- extend the fuzz tests with some skip time and assertions on the debt/premium growth 
- tests on borrow amount effects on wAvgRP (but this might fit better as a scenario test)

src/contracts/LiquidityHub.sol
```diff
@@ -177,7 +131,7 @@ contract LiquidityHub is ILiquidityHub {
 
     // todo: Mitigate inflation attack (burn some amount if first supply)
     uint256 sharesAmount = asset.convertToSharesDown(amount);
-    require(sharesAmount > 0, 'INVALID_SHARES_AMOUNT');
+    require(sharesAmount > 0, InvalidSharesAmount(sharesAmount));
```
```suggestion
    require(sharesAmount > 0, InvalidSharesAmount());
```

according to the current, the shares amount param in the revert will always be zero (bc at non zero values we wouldn't revert here) 

so idt it makes sense to include the param here & unnecessarily increase revert return data size. let's just keep it w/o param. we can update this later if condition changes

src/contracts/LiquidityHub.sol
```diff
@@ -210,7 +164,7 @@ contract LiquidityHub is ILiquidityHub {
     _updateRiskPremiumAndBaseDebt(asset, spoke, _boundBps(riskPremium).rayify(), 0); // no base debt change
 
     uint256 sharesAmount = asset.convertToSharesDown(amount);
-    require(sharesAmount > 0, 'INVALID_SHARES_AMOUNT');
+    require(sharesAmount > 0, InvalidSharesAmount(sharesAmount));
```
DhairyaSethi: same as above

src/contracts/LiquidityHub.sol
```diff
@@ -302,6 +256,29 @@ contract LiquidityHub is ILiquidityHub {
...
+  function getSpokeConfig(
+    uint256 assetId,
+    address spoke
+  ) external view returns (DataTypes.SpokeConfig memory) {
+    return _spokes[assetId][spoke].config;
+  }
+
+  function getTotalAssets(uint256 assetId) external view returns (uint256) {
+    DataTypes.Asset storage asset = _assets[assetId];
+    return asset.getTotalAssets();
```
```suggestion
    return _assets[assetId].getTotalAssets();
```

tbh not sure if this getter is needed, maybe we can add when it's needed

src/contracts/LiquidityHub.sol
```diff
@@ -384,64 +361,69 @@ contract LiquidityHub is ILiquidityHub {
...
     // TODO: still allow withdrawal even if asset is not active, only prevent for frozen/paused?
-    require(asset.config.active, 'ASSET_NOT_ACTIVE');
-    require(amount > 0, 'INVALID_WITHDRAW_AMOUNT');
-    require(
-      amount <= asset.convertToAssetsDown(spoke.suppliedShares) - spoke.baseDebt,
-      'SUPPLIED_AMOUNT_EXCEEDED'
-    );
-    require(amount <= asset.availableLiquidity, 'NOT_AVAILABLE_LIQUIDITY');
+    require(asset.config.active, AssetNotActive(asset.id));
+    require(amount > 0, InvalidWithdrawAmount(amount));
```
DhairyaSethi: same

src/contracts/LiquidityHub.sol
```diff
@@ -384,64 +361,69 @@ contract LiquidityHub is ILiquidityHub {
...
+    uint256 withdrawable = asset.convertToAssetsDown(spoke.suppliedShares) - spoke.baseDebt;
+    require(amount <= withdrawable, SuppliedAmountExceeded(withdrawable));
```
DhairyaSethi, line 387: i think we would have precision issues causing a premature panic with this new revert. prolly best to add this after we fix that issue

src/contracts/LiquidityHub.sol
```diff
@@ -384,64 +361,69 @@ contract LiquidityHub is ILiquidityHub {
...
+  function _validateDraw(
+    DataTypes.Asset storage asset,
+    uint256 amount,
+    uint256 drawCap
+  ) internal view {
     // TODO: Other cases of status (frozen, paused)
-    require(asset.config.active, 'ASSET_NOT_ACTIVE');
-    require(amount > 0, 'INVALID_DRAW_AMOUNT');
+    require(asset.config.active, AssetNotActive(asset.id));
+    require(amount > 0, InvalidDrawAmount(amount));
```
DhairyaSethi: same

src/contracts/LiquidityHub.sol
```diff
@@ -384,64 +361,69 @@ contract LiquidityHub is ILiquidityHub {
...
     uint256 amountRestored
   ) internal view {
     // TODO: Other cases of status (frozen, paused)
-    require(asset.config.active, 'ASSET_NOT_ACTIVE');
+    require(asset.config.active, AssetNotActive(asset.id));
     // Ensure spoke is not restoring more than accrued drawn or equal 0
     require(
       amountRestored > 0 && amountRestored <= spoke.baseDebt + spoke.outstandingPremium,
-      'INVALID_RESTORE_AMOUNT'
+      InvalidRestoreAmount(amountRestored)
```
DhairyaSethi: either the param passed is the limit the function is exceeding (not this bc amount restored is the amount passed in spoke.repay only), or we can also not have the amount param (i prefer the latter tbh)

src/contracts/LiquidityHub.sol
```diff
@@ -488,8 +470,8 @@ contract LiquidityHub is ILiquidityHub {
   }
 
   function _addSpoke(uint256 assetId, DataTypes.SpokeConfig memory config, address spoke) internal {
-    require(spoke != address(0), 'INVALID_SPOKE');
-    _spokes[assetId][spoke] = SpokeData({
+    require(spoke != address(0), InvalidSpoke(spoke));
```
DhairyaSethi: this will revert with spoke address param as 0 always, so same, can rm the param

src/contracts/LiquidityHub.sol
```diff
@@ -526,7 +508,7 @@ contract LiquidityHub is ILiquidityHub {
   }
 
   function _boundBps(uint32 a) internal pure returns (uint256) {
-    require(a < 1000_00, 'INVALID_BPS');
+    require(a < 1000_00, InvalidBps(a));
```
DhairyaSethi: this can be longer and more expressive now, maybe related to the value being a risk premium. maybe invalidRiskPremiumInBps? not sure

src/contracts/Spoke.sol
```diff
@@ -92,63 +34,61 @@ contract Spoke is ISpoke {
...
+    uint256 assetId,
+    DataTypes.ReserveConfig memory params,
+    address asset
+  ) external returns (uint256) {
+    uint256 _reserveCount = reserveCount;
+    DataTypes.Reserve storage reserve = _reserves[_reserveCount];
+    // TODO: validate reserveId does not exist already, valid asset
+    require(
+      params.liquidityPremium <= PercentageMath.PERCENTAGE_FACTOR * 10,
+      InvalidLiquidityPremium(params.liquidityPremium)
```
DhairyaSethi: no need to include in param, it was passed as arg and stack trace is only 1

src/contracts/Spoke.sol
```diff
@@ -92,63 +34,61 @@ contract Spoke is ISpoke {
...
+        liquidityPremium: params.liquidityPremium,
+        borrowable: params.borrowable,
+        collateral: params.collateral
+      })
+    });
 
-  function getSuppliedShares(uint256 reserveId, address user) external view returns (uint256) {
-    return _users[user][reserveId].suppliedShares;
-  }
+    emit ReserveAdded(_reserveCount, assetId, asset);
```
DhairyaSethi: just assetId or asset is fine in params

src/contracts/Spoke.sol
```diff
@@ -92,63 +34,61 @@ contract Spoke is ISpoke {
...
   }
 
-  /// governance
-  function updateReserveConfig(uint256 reserveId, ReserveConfig calldata params) external {
+  function updateReserveConfig(
+    uint256 reserveId,
+    DataTypes.ReserveConfig calldata params
+  ) external {
+    // TODO: More sophisticated
+    require(_reserves[reserveId].asset != address(0), InvalidReserve(reserveId));
```
DhairyaSethi: no need to include reserveId in param

src/contracts/Spoke.sol
```diff
@@ -92,63 +34,61 @@ contract Spoke is ISpoke {
...
+    require(
+      params.liquidityPremium <= PercentageMath.PERCENTAGE_FACTOR * 10,
+      InvalidLiquidityPremium(params.liquidityPremium)
```
DhairyaSethi: same 
no need to add ceiling value as well since that's an input

src/contracts/Spoke.sol
```diff
@@ -166,14 +106,26 @@ contract Spoke is ISpoke {
     );
   }
 
+  // todo: access control, general setter like maker's dss, flag engine like v3
+  function updateLiquidityPremium(uint256 reserveId, uint256 liquidityPremium) external {
+    require(_reserves[reserveId].asset != address(0), InvalidReserve(reserveId));
+    require(
+      liquidityPremium <= PercentageMath.PERCENTAGE_FACTOR * 10,
+      InvalidLiquidityPremium(liquidityPremium)
+    );
```
DhairyaSethi: same as above for these two

src/contracts/Spoke.sol
```diff
@@ -286,148 +238,156 @@ contract Spoke is ISpoke {
...
+    DataTypes.Reserve storage reserve,
+    DataTypes.UserConfig storage user,
     uint256 amount
   ) internal view {
-    require(
-      liquidityHub.convertToAssetsDown(reserve.assetId, user.suppliedShares) >= amount,
-      'INSUFFICIENT_SUPPLY'
-    );
+    uint256 suppliedAmount = liquidityHub.convertToAssetsDown(reserve.assetId, user.suppliedShares);
+    require(suppliedAmount >= amount, InsufficientSupply(suppliedAmount));
```
DhairyaSethi: input amount not required in param

src/contracts/Spoke.sol
```diff
@@ -286,148 +238,156 @@ contract Spoke is ISpoke {
...
-    Reserve storage reserve,
-    UserConfig storage user,
+    DataTypes.Reserve storage reserve,
+    DataTypes.UserConfig storage user,
     uint256 amount
   ) internal view {
-    require(amount <= user.baseDebt + user.outstandingPremium, 'REPAY_EXCEEDS_DEBT');
+    require(
+      amount <= user.baseDebt + user.outstandingPremium,
+      RepayExceedsDebt(user.baseDebt + user.outstandingPremium)
```
DhairyaSethi: duplicate math, would prefer to cache

tests/Base.t.sol
```diff
@@ -512,9 +486,4 @@ abstract contract BaseTest is Test, Events {
       })
     );
   }
-
-  /// @dev pseudo random randomizer
-  function randomizer(uint256 min, uint256 max, uint256) internal returns (uint256) {
```
DhairyaSethi, line 517: let's bring back here as discussed above

tests/Utils.sol
```diff
@@ -101,21 +107,52 @@ library Utils {
...
+    uint256 newDrawCap
+  ) internal {
+    DataTypes.SpokeConfig memory spokeConfig = hub.getSpokeConfig(assetId, spoke);
+    spokeConfig.drawCap = newDrawCap;
+    hub.updateSpokeConfig(assetId, spoke, spokeConfig);
+  }
+
+  function getUserInfo(
+    ISpoke spoke,
+    address user,
```
DhairyaSethi: let's move both to base, wdyt?

src/contracts/LiquidityHub.sol
yan-man: in favor of keeping this one

src/contracts/Spoke.sol
yan-man: imo both assetId/asset are helpful here

DhairyaSethi: both are interchangeable constants, either can be derived from the other

DhairyaSethi: let's do only assetId imo, since reserveCount represents reserve id

src/contracts/LiquidityHub.sol
yan-man: here it could be 0 or > limit so tbh this specific case might be helpful to emit the amount restored so you can rule out the 0 case, but we should keep a consistent convention (ie always setting to the limit val). 

otherwise split into 2 errors, 1 for the 0 case. But that is more wordy

src/contracts/Spoke.sol
yan-man: > both are interchangeable constants

agree but then it requires reading from hub to get 1 from the other, so for some external party indexing on events might be helpful to just get both at once

DhairyaSethi: not really, imo external parties will already have assetId -> asset address cached/indexed, especially since that's very unlikely to change 
and for any reference, they'd likely need more asset data information, for that just the foreign key reference of assetId is enough

src/contracts/LiquidityHub.sol
DhairyaSethi: it can zeroOrSurplusAmountRestored 

not enough reason to add the input amount in params, if needed ceiling amount is fine (bc that's dynamic in the system, user/ui already knows its input amount)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -61,16 +59,19 @@ contract LiquidityHubSupplyTest is LiquidityHubBaseTest {
     });
     skip(365 days);
 
-    vm.expectRevert(TestErrors.SUPPLY_CAP_EXCEEDED);
-    hub.supply(daiAssetId, amount, 0, alice);
+    _updateSupplyCap(daiAssetId, address(spoke2), newSupplyCap);
```
yan-man: good catch on this fix. imo we should update the supply cap prior to the skip time and action. 

supply1:
- supplies up to supplyCap - 1. It passes, and then only the act of accrual increases the supplied amount above the cap. 

supply2: 
- after accrual increases supply > supplyCap, should try to supply amount=1 and fail (it would have originally passed prior to accrual)

yan-man:
I think would be helpful to update the description to include some of the other infra fixes like migrating to custom errors, data types lib, etc.

src/contracts/Spoke.sol
DhairyaSethi: also would prefer RepayAmountExceedsDebt

src/contracts/Spoke.sol
yan-man: don't feel too strongly either way

DhairyaSethi: we have to consider events from an indexing and database perspective, this is not a personal preference design choice.

tests/unit/Spoke/Spoke.Borrow.t.sol
CheyenneAtapour: Yeah, I will adjust Pr description and make a separate issue

src/contracts/LiquidityHub.sol
CheyenneAtapour: This is the same logic that we had previously. It currently doesn't break tests. We should note the new tests we write that break it and then address accordingly. All I did here was cache the math

src/contracts/LiquidityHub.sol
CheyenneAtapour: I can change it to that. There's also an argument to leave as is, since the function name itself is also generic.

src/contracts/Spoke.sol
CheyenneAtapour: This is a totally new reserve being added though and only from governance. If you are a third party and you just see a new reserveId and assetId from an on-chain event, you don't know if it's an old asset being re-listed like in E-Mode or a brand new asset, so imo helpful to have both asset address and Id

Edit: Ok, since this is on spoke only, you could just query the LH with assetId. But in the case of RWAs, would we have a reserve on the spoke that isn't on the LH? Or would it be in some other LH? If it will be on some LH, then perhaps we can just do reserveId and assetId

src/contracts/Spoke.sol
CheyenneAtapour: This one isn't input amount, but rather ceiling of what you can withdraw

src/contracts/Spoke.sol
DhairyaSethi: it'll be on the LH spoke is attached to, yes only reserveId & assetId is needed

src/contracts/Spoke.sol
DhairyaSethi: yeah it's your amount balance, which is known while doing the txn. imo not needed here regardless, but if you want to keep then pls reverse the condition, it's more helpful in comparisons if limit value is on the right

CheyenneAtapour: I agree with that convention of ordering. Doing it

src/interfaces/ISpoke.sol
```diff
@@ -1,12 +1,15 @@
...
+ * @author Aave Labs
+ * @notice Basic interface for Spoke
+ */
 interface ISpoke {
-  event Borrowed(uint256 indexed reserveId, uint256 amount, address indexed user);
-  event Repaid(uint256 indexed reserveId, uint256 amount, address indexed user);
-  event Supplied(uint256 indexed reserveId, uint256 amount, address indexed user);
-  event Withdrawn(uint256 indexed reserveId, uint256 amount, address indexed user);
-  event UsingAsCollateral(uint256 indexed reserveId, bool usingAsCollateral, address indexed user);
+  event ReserveAdded(uint256 indexed reserveId, uint256 assetId);
```
```suggestion
  event ReserveAdded(uint256 indexed reserveId, uint256 indexed assetId);
```

src/interfaces/ISpoke.sol
```diff
@@ -15,20 +18,57 @@ interface ISpoke {
     bool borrowable,
     bool collateral
   );
+  event LiquidityPremiumUpdated(uint256 indexed reserveId, uint256 liquidityPremium);
+
+  event Supplied(uint256 indexed reserveId, uint256 amount, address indexed user);
+  event Withdrawn(uint256 indexed reserveId, uint256 amount, address indexed user);
+  event Borrowed(uint256 indexed reserveId, uint256 amount, address indexed user);
+  event Repaid(uint256 indexed reserveId, uint256 amount, address indexed user);
```
DhairyaSethi: let's change order to match LH 
reserveId, user, amount

src/interfaces/ISpoke.sol
```diff
@@ -15,20 +18,57 @@ interface ISpoke {
...
+  function getUserRiskPremium(address user) external view returns (uint256);
+  function getLastUsedUserRiskPremium(address user) external view returns (uint256);
+  function getHealthFactor(address user) external view returns (uint256);
+  function getReservePrice(uint256 reserveId) external view returns (uint256);
+  function getLiquidityPremium(uint256 reserveId) external view returns (uint256);
+  function getReserve(uint256 reserveId) external view returns (DataTypes.Reserve memory);
+  function getUser(
+    uint256 reserveId,
+    address user
+  ) external view returns (DataTypes.UserConfig memory);
```
DhairyaSethi, line 73: we should rename this function to getUserPosition 
bc that's different per reserveId (there's a future todo as well to rename userConfig struct to userPosition in spoke - we can tackle this later or now)

src/contracts/LiquidityHub.sol
```diff
@@ -302,6 +255,29 @@ contract LiquidityHub is ILiquidityHub {
...
+  function getSpokeConfig(
+    uint256 assetId,
+    address spoke
+  ) external view returns (DataTypes.SpokeConfig memory) {
+    return _spokes[assetId][spoke].config;
+  }
+
+  function getTotalAssets(uint256 assetId) external view returns (uint256) {
+    DataTypes.Asset storage asset = _assets[assetId];
+    return asset.getTotalAssets();
```
```suggestion
    return _assets[assetId].getTotalAssets();
```

src/contracts/LiquidityHub.sol
```diff
@@ -384,64 +360,70 @@ contract LiquidityHub is ILiquidityHub {
...
   function _validateSupply(
-    Asset storage asset,
-    SpokeData storage spoke,
+    DataTypes.Asset storage asset,
+    DataTypes.SpokeData storage spoke,
     uint256 amount
   ) internal view {
-    require(amount > 0, 'INVALID_SUPPLY_AMOUNT');
-    require(assetsList[asset.id] != IERC20(address(0)), 'ASSET_NOT_LISTED');
+    require(amount > 0, InvalidSupplyAmount(amount));
```
```suggestion
    require(amount > 0, InvalidSupplyAmount());
```

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -0,0 +1,388 @@
...
+      aliceDaiData.suppliedShares,
+      hub.convertToShares(daiAssetId, daiSupplyAmount),
+      'alice dai supply shares pre-draw'
+    );
+    assertEq(aliceDaiData.baseDebt, 0, 'alice dai base debt pre-draw');
+    assertEq(aliceWethData.suppliedShares, 0, 'alice weth supply shares pre-draw');
+    assertEq(aliceWethData.baseDebt, 0, 'alice weth base debt pre-draw');
+
+    assertEq(tokenList.dai.balanceOf(address(spoke1)), 0, 'spoke1 dai balance pre-draw');
+    assertEq(tokenList.weth.balanceOf(address(spoke2)), 0, 'spoke2 weth balance pre-draw');
```
miguelmtzinf: aren't these checks a bit useless?

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -0,0 +1,388 @@
...
+    DataTypes.UserConfig memory aliceWethData = getUserInfo(spoke1, alice, wethReserveId);
+
+    uint256 bobDaiBalanceBefore = tokenList.dai.balanceOf(bob);
+    uint256 bobWethBalanceBefore = tokenList.weth.balanceOf(bob);
+    uint256 aliceDaiBalanceBefore = tokenList.dai.balanceOf(alice);
+
+    assertEq(
+      bobWethData.suppliedShares,
+      hub.convertToShares(wethAssetId, wethSupplyAmount),
+      'bob weth supply shares pre-draw'
```
miguelmtzinf: i d use `before` instead of `pre-draw`
easier to maintain and still useful for debugging

src/interfaces/ISpoke.sol
miguelmtzinf: in favor of doing it later, alongside the struct renaming

tests/Base.t.sol
```diff
@@ -18,7 +18,7 @@ import {DataTypes} from 'src/libraries/types/DataTypes.sol';
 import {TestnetERC20} from './mocks/TestnetERC20.sol';
 import {MockERC20} from './mocks/MockERC20.sol';
 import {MockPriceOracle, IPriceOracle} from './mocks/MockPriceOracle.sol';
-import {Utils} from './Utils.t.sol';
+import {Utils} from './Utils.sol';
 
 // library Constants {}
```
miguelmtzinf: this comment and Events interface seems unneeded? at least for Aave ones.

tests/Utils.sol
```diff
@@ -0,0 +1,77 @@
...
+  function withdraw(
+    ILiquidityHub hub,
+    uint256 assetId,
+    address spoke,
+    uint256 amount,
+    uint32 riskPremium,
+    address to
+  ) internal {
+    vm.prank(spoke);
+    hub.withdraw({assetId: assetId, amount: amount, riskPremium: riskPremium, to: to});
```
DhairyaSethi: these two don't need named params

tests/Base.t.sol
```diff
@@ -56,33 +56,7 @@ contract Events {
...
-  bytes constant INVALID_RESTORE_AMOUNT = 'INVALID_RESTORE_AMOUNT';
-  bytes constant INVALID_SHARES_AMOUNT = 'INVALID_SHARES_AMOUNT';
-  // Spoke
-  bytes constant NO_SUPPLY = 'NO_SUPPLY';
-  bytes constant REPAY_EXCEEDS_DEBT = 'REPAY_EXCEEDS_DEBT';
-  bytes constant RESERVE_NOT_LISTED = 'RESERVE_NOT_LISTED';
-}
-
-abstract contract BaseTest is Test, Events {
+abstract contract Base is Test, Events {
```
DhairyaSethi: we can rm Events contract above and inheritance here, events can be picked up from the interfaces

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -0,0 +1,388 @@
...
+    // Additional supply to accrue interest
+    Utils.spokeSupply(spoke1, daiReserveId, bob, 1e18, bob);
+
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.DrawCapExceeded.selector, drawCap));
+    Utils.spokeBorrow(spoke1, daiReserveId, bob, 1, bob);
+  }
+
+  function test_borrow_fuzz_multiple_reserves(
+    uint256 daiBorrowAmount,
+    uint256 wethBorrowamount,
```
miguelmtzinf: typo
```suggestion
    uint256 wethBorrowAmount,
```

tests/Base.t.sol
CheyenneAtapour: Miguel also mentioned this one. But we do check the OZ event. I am leaving for that and for any future stuff, it's probably a nice pattern to do it with

tests/unit/Spoke/Spoke.Borrow.t.sol
CheyenneAtapour: wow, nice eyes mate!

tests/Base.t.sol
DhairyaSethi: OZ events should be imported from OZ directly no? like we do with OZ errors (like IERC20Errors.InsufficientApproval)

DhairyaSethi: doesn't make sense for that one event really imo

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -61,16 +59,19 @@ contract LiquidityHubSupplyTest is LiquidityHubBaseTest {
...
+
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.SupplyCapExceeded.selector, newSupplyCap));
+    vm.prank(address(spoke2));
+    hub.supply(daiAssetId, 1e8, 0, alice);
```
yan-man: now can supply just 1 and test still works (checked locally)

```suggestion
    hub.supply(daiAssetId, 1, 0, alice);
```

PR #204, opened 2025-02-25
fix: risk premium

### LINKED ISSUE #193: Spoke - Add Back Dynamic Sort
While optimization is being done to determine best way to handle sorting assets by liquidity premium, we need to ensure our algorithm for computing user risk premium works.

We will temporarily add back the dynamic sort using open zeppelin's `Arrays.sol` library, and use their quicksort method to proceed with tests on user risk premium

### LINKED ISSUE #197: User RP Issues

DhairyaSethi:
closes #197 
closes #193

- [x] decimal normalisation, base for risk management (LT, HF etc)
- [x] run time sort
- [x] correct user risk premium calc
- [x] retrigger calc on hub
---
- [x] keyValueList tests
- [ ] cache asset.decimals on spoke to reuse on decimal normalisation
- [ ] user rp tests, correct rp applied when drawing multiple reserves 
- [ ] gas profiling

note: user risk premium doesn't need to be stored in rayified form (since it isn't maintained as a running avg, but rather always calculated on the fly) but is done so for uniformity with the other three risk premiums (can be optimised later)

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -59,6 +59,7 @@ contract Spoke is ISpoke {
      * then this value is 1550_000000000000000000000000000 (1550 * 1e27),
      * stored with high precision to be equivalent with other RPs (Asset, Spoke/Reserve)
      * since they have to maintain a running weighted average
+     * todo optimize: user RP doesn't need to be stored in full precision as described above
      */
     uint256 riskPremium;
     // todo supplied/borrowed (2d) bitmap


src/contracts/LiquidityHub.sol
```diff
@@ -283,6 +283,14 @@ contract LiquidityHub is ILiquidityHub {
     return amount;
   }
 
+  function accrueInterest(uint256 assetId, uint32 riskPremium) external {
```
miguelmtzinf, line 287:
```suggestion
  /// @inheritdoc ILiquidityHub
  function accrueInterest(uint256 assetId, uint32 riskPremium) external {
    // TODO: authorization - only spokes
```

src/contracts/LiquidityHub.sol
```diff
@@ -363,6 +371,10 @@ contract LiquidityHub is ILiquidityHub {
     return _spokes[assetId][spoke].riskPremium.derayify();
   }
 
+  function getAssetConfig(uint256 assetId) external view returns (DataTypes.AssetConfig memory) {
```
miguelmtzinf, line 378:
```suggestion
  /// @inheritdoc ILiquidityHub
  function getAssetConfig(uint256 assetId) external view returns (DataTypes.AssetConfig memory) {
```

src/contracts/LiquidityHub.sol
```diff
@@ -454,8 +466,8 @@ contract LiquidityHub is ILiquidityHub {
 
     uint256 newSpokeDebt = baseDebtChange > 0
       ? existingSpokeDebt + uint256(baseDebtChange) // debt added
-      : // force underflow: only possible when spoke takes repays amount more than net drawn
-      existingSpokeDebt - uint256(-baseDebtChange); // debt restored
+      // force underflow: only possible when spoke takes repays amount more than net drawn
+      : existingSpokeDebt - uint256(-baseDebtChange); // debt restored
```
miguelmtzinf, line 473: why the hell does this appear on every PR diff

DhairyaSethi: linter wars, i'll fix it once & for all [e251299](https://github.com/aave/aave-v4/pull/204/commits/e251299cc68f8ff8f8e9fafd23dce0bdc4f0d3f5)

src/contracts/Spoke.sol
```diff
@@ -14,9 +15,12 @@ contract Spoke is ISpoke {
   using WadRayMath for uint256;
   using PercentageMath for uint256;
   using SafeERC20 for IERC20;
+  using KeyValueListInMemory for KeyValueListInMemory.List;
 
   uint256 public constant DEFAULT_SPOKE_INDEX = 0;
-  ILiquidityHub public liquidityHub;
+  // todo capitalize, oracle should be mutable?
```
miguelmtzinf, line 21: no strong opinion for now. smt to decide later

src/contracts/Spoke.sol
```diff
@@ -38,46 +42,54 @@ contract Spoke is ISpoke {
...
+    uint256 assetPrice;
+    uint256 assetUnit;
     uint256 userCollateralInBaseCurrency;
     uint256 totalCollateralInBaseCurrency;
     uint256 totalDebtInBaseCurrency;
     uint256 avgLiquidationThreshold;
     uint256 userRiskPremium;
+    uint256 liquidityPremium;
     uint256 healthFactor;
+    uint256 suppliedReserveCount;
```
miguelmtzinf: is this the number of assets used as collateral? or supplied

src/contracts/Spoke.sol
```diff
@@ -278,10 +309,18 @@ contract Spoke is ISpoke {
   }
 
   // TODO: Needed?
-  function getInterestRate(uint256 assetId) public view returns (uint256) {
+  function getInterestRate(uint256 reserve) public view returns (uint256) {
     // read from state, convert to ray
     // TODO: should be final IR rather than base?
-    return ILiquidityHub(liquidityHub).getBaseInterestRate(assetId);
+    return ILiquidityHub(liquidityHub).getBaseInterestRate(reserve);
```
miguelmtzinf: shouldn't we use the associated `assetId` to the `reserveId`?

src/contracts/Spoke.sol
```diff
@@ -38,46 +42,54 @@ contract Spoke is ISpoke {
...
     uint256 baseDebt;
     uint256 outstandingPremium;
     uint256 suppliedShares;
     uint256 baseBorrowIndex;
-    uint256 riskPremium;
     uint256 lastUpdateTimestamp;
   }
 
+  struct UserData {
+    uint256 riskPremium;
```
miguelmtzinf, line 64: let's clarify this is a ray-extended bps (add an example too)

src/contracts/Spoke.sol
```diff
@@ -524,6 +519,7 @@ contract Spoke is ISpoke {
     return user.usingAsCollateral;
   }
 
+  // todo is user borrowing is baseDebt is 0, but outstandingPremium is non zero if we allow only repaying base debt?
```
miguelmtzinf: this won't never happen right? we don't allow repaying base debt only

src/contracts/Spoke.sol
DhairyaSethi: used as collateral, it's for the risk premium. will add a comment to clarify

src/contracts/Spoke.sol
DhairyaSethi: yep. this is a remnant not used to anywhere, was planning to rm this

src/contracts/Spoke.sol
DhairyaSethi: yes, essentially means the check below on whether an asset is can be baseDebt > 0 only, but regardless we'll switch to a bitmap for this

src/contracts/LiquidityHub.sol
DhairyaSethi: sure, not adding to other getters rn to not pollute diff

src/contracts/Spoke.sol
DhairyaSethi: how about 
```solidity
    /**
     * ray-extended risk premium bps of user
     * for example, if risk premium bps is 15_50 (15.5%),
     * then this value is 1550_000000000000000000000000000 (1550 * 1e27),
     * stored with high precision to be equivalent with other RPs
     * to maintain a running weighted average
     */
    uint256 riskPremium;
```

src/contracts/Spoke.sol
```diff
@@ -38,46 +42,55 @@ contract Spoke is ISpoke {
...
-  mapping(address => mapping(uint256 => UserConfig)) internal _users;
-  // reserve id => reserveData
-  mapping(uint256 => Reserve) internal _reserves;
+  // todo rename UserConfig => UserPosition
+  mapping(address user => mapping(uint256 reserveId => UserConfig position)) internal _users;
+  mapping(address user => UserData data) internal _userData;
+  mapping(uint256 reserveId => Reserve reserveData) internal _reserves;
 
-  uint256[] public reservesList; // reserveIds
+  uint256[] public reservesList; // todo: rm, not needed
```
yan-man, line 90: don't we need some way to get a list of reserveIds? or just assuming 0 -> reserveCount - 1

src/contracts/Spoke.sol
DhairyaSethi: note: user risk premium doesn't need to be stored in rayified form (since it isn't maintained as a running avg, but rather always calculated on the fly) but is done so for uniformity with the other three risk premiums (can be optimised later)

src/contracts/Spoke.sol
DhairyaSethi: if not storing pre-sorted then it can be just be 0 -> reserveCount - 1

src/contracts/Spoke.sol
```diff
@@ -172,6 +184,12 @@ contract Spoke is ISpoke {
       newAggregatedRiskPremium,
       msg.sender // supplier
     );
+    _notifyRiskPremiumUpdate(reserve.assetId, msg.sender);
+
+    // first supply, set as collateral by default
+    if (user.suppliedShares > 0) {
+      setUsingAsCollateral(reserveId, true);
```
yan-man: but a user isn't required to set supplied asset as collateral right? They can just earn supply yield for it without borrowing

src/contracts/Spoke.sol
```diff
@@ -524,49 +513,91 @@ contract Spoke is ISpoke {
     return user.usingAsCollateral;
   }
 
+  // todo opt: use bitmap
   function _borrowing(UserConfig storage user) internal view returns (bool) {
```
yan-man: can we rename this to `_isBorrowing`

src/contracts/Spoke.sol
DhairyaSethi: yes they're not

>They can just earn supply yield for it without borrowing

*without enabling as collateral, if so, yes, this was just to follow v3's design and make tests slightly simpler by not needing to setUsingAsCollateral for each supplied asset
it needs to be enabled as collateral to be included in the risk premium, at least that's how i've implemented it, debatable choice

src/contracts/Spoke.sol
DhairyaSethi: [e100982](https://github.com/aave/aave-v4/pull/204/commits/e1009826a433cfff8ffe8d545fb093594fbe5d1c)

src/contracts/Spoke.sol
yan-man: > it needs to be enabled as collateral to be included in the risk premium, at least that's how i've implemented it, debatable choice

agree, should only be collateral involved in RP bc only collateral is counted against debts

> make tests slightly simpler

imo don't think we should do this - bc Emilio was mentioning wanting to keep supply action (for supply yield) as cheap as possible, also don't think test considerations should lead impl

src/contracts/Spoke.sol
miguelmtzinf: I'd rename to `collateralReserveCount` too

src/contracts/LiquidityHub.sol
CheyenneAtapour: linter flips a coin lol

src/contracts/Spoke.sol
DhairyaSethi: fine with removing but the reasoning is not just for tests, like i mentioned, similar in v3 & it's better for UX to keep this (since it's only executed on first supply of an asset)

src/contracts/KeyValueListInMemory.sol
```diff
@@ -53,4 +52,8 @@ library KeyValueListInMemory {
   function keyComparator(uint256 a, uint256 b) internal pure returns (bool) {
```
miguelmtzinf: `keyComparator` and `valueComparator` are not longer needed
why removing `sortByValue`?

DhairyaSethi: keeping it clean, i can rm the other comparators as well

DhairyaSethi: we'll need these comparators if we want to optimise further (by packing more than one key,value pairs in each slot)

src/contracts/Spoke.sol
DhairyaSethi: [757cb71](https://github.com/aave/aave-v4/pull/204/commits/757cb71d0be9cbb90c9d552e3f29632172e6407f)

src/contracts/Spoke.sol
```diff
@@ -336,6 +364,13 @@ contract Spoke is ISpoke {
     });
   }
 
+  // todo: access control, general setter like maker's dss, flag engine like v3
+  function updateLiquidityPremium(uint256 reserveId, uint256 liquidityPremium) external {
+    require(_reserves[reserveId].asset != address(0), 'INVALID_RESERVE');
+    require(liquidityPremium <= PercentageMath.PERCENTAGE_FACTOR * 10, 'INVALID_LIQUIDITY_PREMIUM');
```
CheyenneAtapour, line 385: Ideally we also have this check on the params to `addReserve`

DhairyaSethi: yeah not adding to keep this focussed on the core issue, it can be taken up separately or i can add now. lmk
added [35f389e](https://github.com/aave/aave-v4/pull/204/commits/35f389e9bb66273d836d2930a32e8ade73c4d18e)

src/contracts/Spoke.sol
```diff
@@ -524,49 +508,92 @@ contract Spoke is ISpoke {
...
+    vars.i = 0;
+    vars.reserveId = 0;
+    while (vars.reserveId < reservesListLength) {
+      UserConfig storage user = _users[userAddress][vars.reserveId];
+      Reserve storage reserve = _reserves[vars.reserveId];
+      if (_usingAsCollateral(user)) {
+        vars.assetId = reserve.assetId;
+        vars.liquidityPremium = reserve.config.liquidityPremium;
+        vars.assetPrice = oracle.getAssetPrice(vars.assetId);
+        vars.assetUnit = 10 ** liquidityHub.getAssetConfig(vars.assetId).decimals; // unchecked
```
CheyenneAtapour: comment says unchecked, but this one is checked, right?

DhairyaSethi: yeah need to wrap in an unchecked block, [35f389e](https://github.com/aave/aave-v4/pull/204/commits/35f389e9bb66273d836d2930a32e8ade73c4d18e)

src/contracts/Spoke.sol
DhairyaSethi: idea is to keep seperate and multicall setUsingAsCollateral

src/contracts/Spoke.sol
```diff
@@ -682,4 +753,34 @@ contract Spoke is ISpoke {
...
+  function _notifyRiskPremiumUpdate(uint256 assetIdToAvoid, address userAddress) internal {
+    uint256 reserveCount_ = reserveCount;
+    uint256 i;
+    UserData storage userData = _userData[userAddress];
+    while (i < reserveCount_) {
+      UserConfig storage user = _users[userAddress][i];
+      Reserve storage reserve = _reserves[i];
+      uint256 assetId = reserve.assetId;
+      // todo keep borrowed assets in transient storage/pass through?
+      if (assetId != assetIdToAvoid && _isBorrowing(user)) {
```
yan-man:
```suggestion
      if (assetId != assetIdToAvoid && _usingAsCollateralOrBorrowing(user)) {
```
shouldn't we also update storage for collateral assets, as that could affect a user's RP?

DhairyaSethi: nope definitely not. 
at this point we're updating the other two RPs(asset, & spoke/reserve), which are defined per drawn asset only 
> as that could affect a user's RP

they don't affect the other RPs (asset, & spoke/reserve) since the user doesn't contribute to them since the user doesn't borrow this asset

src/contracts/Spoke.sol
```diff
@@ -651,6 +716,7 @@ contract Spoke is ISpoke {
 
   function _previewUserInterest(
     UserConfig storage user,
+    UserData storage userData, // todo opt: pass user rp only
```
CheyenneAtapour, line 727: If this would be an opt, can just do it now imo. low-hanging fruit

src/contracts/Spoke.sol
CheyenneAtapour: would be a (very) small opt to flip the order on these because the _isBorrowing would likely short-circuit more often than the `assetIdToAvoid` check

src/contracts/Spoke.sol
DhairyaSethi: actually requires a bigger refactor bc this method is used quite frequently 
the ideal opt is when we pass around a cached obj like v3

src/contracts/Spoke.sol
DhairyaSethi: [2eaf581](https://github.com/aave/aave-v4/pull/204/commits/2eaf58139ffe61cd3e58d4b76bd46c358625fd8a)

src/contracts/Spoke.sol
DhairyaSethi: [f2fa416](https://github.com/aave/aave-v4/pull/204/commits/f2fa4168502baff4ae9773865db8eacc9b0e2ab3)

Issue #205, opened 2025-02-25
User RP Storage Update Issue


yan-man:
Following from initial implementation fix to: https://github.com/aave/aave-v4/issues/197

Assume scenario (all assets have LP > 0):

T0: 
- Bob supplies dai
- Alice supplies usdx 
- Alice borrows dai

time flies to T1

At T1, the previewed outstanding premium for Alice on her dai should be > 0. When she borrows the dai, her user RP should be updated, but it is not.

PR #206, opened 2025-02-26
RP storage update fix

yan-man:
- naive fix to commit new user RP to storage at the end of accrual actions
- add getter for newly created userData 
- not meant to be fully exhaustive testing for user RP, just enough to target the specific failing scenario

closes https://github.com/aave/aave-v4/issues/205

### DIFF
--- a/tests/unit/Spoke/Spoke.UserRiskPremium.t.sol
+++ b/tests/unit/Spoke/Spoke.UserRiskPremium.t.sol
@@ -21,6 +21,7 @@ contract SpokeUserRiskPremiumTest is BaseTest {
     uint256 suppliedAmount;
     uint40 lastUpdateTimestamp;
     uint256 existingBaseDebt;
+    uint256 existingOutstandingPremium;
     uint256 cumulatedBaseInterest;
     uint256 cumulatedBaseDebt;
     uint256 cumulatedOutstandingPremium;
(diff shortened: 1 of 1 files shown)


tests/unit/Spoke/Spoke.UserRiskPremium.t.sol
```diff
@@ -0,0 +1,236 @@
...
+
+    // alice supplies usdx
+    vm.startPrank(alice);
+    spoke1.supply({reserveId: state.usdxReserveId, amount: 100e18});
+    spoke1.setUsingAsCollateral(state.usdxReserveId, true);
+
+    // alice borrows dai
+    spoke1.borrow({reserveId: state.daiReserveId, amount: 10e18, to: alice});
+
+    state.userRiskPremiumStorage = spoke1.getUserData(alice).riskPremium;
```
DhairyaSethi, line 62: let's never use internal data, we have exposed getLastUsedRiskPremium for the storage value

src/contracts/Spoke.sol
```diff
@@ -792,5 +796,7 @@ contract Spoke is ISpoke {
         ++i;
       }
     }
+    (uint256 newUserRiskPremium, , ) = _calculateUserAccountData(userAddress); // TODO: optimize
```
DhairyaSethi, line 799: for reference: this doesn't work bc the risk premium used in accounting is then incorrect as it doesn’t account for this borrow action change (consider the case for a second borrow, it’ll be apparent)

yan-man:
refer to https://github.com/aave/aave-v4/pull/207

PR #207, opened 2025-02-26
fix user rp calc

### LINKED ISSUE #205: User RP Storage Update Issue

### LINKED ISSUE #208: User RP Incorrect Calc
user risk premium currently is being updated twice, the second time in the notify method which also is not using the correct latest risk premium

DhairyaSethi:
fixes #205 
fixes #208

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -798,7 +798,8 @@ contract Spoke is ISpoke {
    * @dev It is assumed debt has already been accrued on this `reserve` & `user`Position, and newUserRiskPremium
    * is calculated with all accrued reserves.
    * @dev This is currently only used on `_notifyRiskPremiumUpdate`; since no debt is added/removed on this reserve,
-   * it doesn't change the new user risk premium. Optimize later to use this method in `supply/withdraw` as well.
+   * hence it doesn't change the new user risk premium.
+   * TODO: Optimize later to use this method in `supply` & `withdraw` as well.
    * @return New reserve risk premium (rayified)
    */
   function _refreshReserveRiskPremium(


src/contracts/Spoke.sol
```diff
@@ -141,6 +141,9 @@ contract Spoke is ISpoke {
     return _users[user][reserveId].suppliedShares;
   }
 
+  // todo by default returns only stored value, consider renaming to `getLast{Used,Stored}ReserveRiskPremium`
+  // to be inline with user's stored rp getter. we don't have an up to date rp concept here since that requires
+  // looping over all supplied users
```
CheyenneAtapour: I'm unsure about this comment. The idea behind the user risk premium side is that they can trigger an update of their own user rp if it becomes lower. I don't believe we will have the same concept here because we only ever use the stored value in the calcs, recalculating and storing on each action

DhairyaSethi: yeah exactly, that's what this comment says, we don't have a concept for updated risk premium for reserve, it's always stored
so we can consider renaming this method

src/contracts/Spoke.sol
```diff
@@ -497,19 +475,28 @@ contract Spoke is ISpoke {
       ? existingUserDebt + uint256(baseDebtChange) // debt added
       : existingUserDebt - uint256(-baseDebtChange); // debt restored
     // force underflow^: only possible when user takes repays amount more than net drawn
+    uint256 newReserveDebt = baseDebtChange > 0
```
CheyenneAtapour: Is it easier to just turn this logic into an if statement: `if baseDebtChange > 0` then add the base debt change to both user and reserve debt, else subtract it from both?

src/contracts/Spoke.sol
```diff
@@ -767,30 +753,76 @@ contract Spoke is ISpoke {
...
+   * @return New reserve risk premium (rayified)
+   */
+  function _refreshReserveRiskPremium(
+    Reserve storage reserve,
+    UserConfig storage user, // user position on this reserve
+    uint256 existingUserRiskPremium,
+    uint256 newUserRiskPremium
+  ) internal returns (uint256) {
+    uint256 reserveDebt = reserve.baseDebt;
+    uint256 userDebt = user.baseDebt;
```
CheyenneAtapour, line 811: You could just use these directly as `user.baseDebt` and `reserve.baseDebt` imo, following the same style as `reserve.riskPremium`

DhairyaSethi: `user.baseDebt` is used twice to better to cache it, ideally there should be another method in MathUtils to handle this. added a todo note for later, will create a ticket for this also [283a5da](https://github.com/aave/aave-v4/pull/207/commits/283a5da8feca3aca1938b9ff2e7b7353b7af6c70)

src/contracts/Spoke.sol
DhairyaSethi: sure good idea [5494b52](https://github.com/aave/aave-v4/pull/207/commits/5494b52ec3f45a736b104c4a8d9324351203ee91)

src/contracts/Spoke.sol
```diff
@@ -279,12 +278,10 @@ contract Spoke is ISpoke {
     liquidityHub.restore(
       reserve.assetId,
       amount,
-      newAggregatedRiskPremium,
+      uint32(newReserveRiskPremium.derayify()),
       msg.sender // repayer
     );
-    _notifyRiskPremiumUpdate(reserve.assetId, msg.sender);
-
-    emit Repaid(reserveId, amount, msg.sender);
```
CheyenneAtapour, line 287: Think you accidentally removed?

DhairyaSethi: ah yeah, brought back the event lol [6be816b](https://github.com/aave/aave-v4/pull/207/commits/6be816b7e93b76b363cab72df19037923c80fb00)

PR #214, opened 2025-02-27
Add Spoke repay tests

### LINKED ISSUE #127: Spoke RP - Repay unit testing

miguelmtzinf:
Closes #127

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -423,7 +423,7 @@ contract LiquidityHub is ILiquidityHub {
     // TODO: Other cases of status (frozen, paused)
     require(asset.config.active, AssetNotActive());
     require(amountRestored > 0, InvalidRestoreAmount());
-    // Ensure spoke is not restoring more than accrued drawn or equal 0
+    // Ensure spoke is not restoring more than accrued drawn
     uint256 maxAllowedRestore = spoke.baseDebt + spoke.outstandingPremium;
     require(amountRestored <= maxAllowedRestore, SurplusAmountRestored(maxAllowedRestore));
   }
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -361,10 +361,7 @@ contract Spoke is ISpoke {
     DataTypes.UserPosition storage user,
     uint256 amount
   ) internal view {
-    uint256 suppliedAmount = liquidityHub.convertToAssetsDown(
-      reserve.assetId,
-      user.suppliedShares
-    ) - user.baseDebt;
+    uint256 suppliedAmount = liquidityHub.convertToAssetsDown(reserve.assetId, user.suppliedShares);
     require(amount <= suppliedAmount, InsufficientSupply(suppliedAmount));
   }
 
--- a/tests/unit/Spoke/Spoke.Repay.t.sol
+++ b/tests/unit/Spoke/Spoke.Repay.t.sol
@@ -6,7 +6,7 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
 contract SpokeRepayTest is SpokeBase {
   using PercentageMath for uint256;
 
-  function test_repay_same_block2() public {
+  function test_repay_same_block() public {
     uint256 daiSupplyAmount = 100e18;
     uint256 wethSupplyAmount = 10e18;
     uint256 daiBorrowAmount = daiSupplyAmount / 2;
@@ -40,13 +40,11 @@ contract SpokeRepayTest is SpokeBase {
     assertEq(bobWethDataBefore.suppliedShares, hub.convertToShares(wethAssetId, wethSupplyAmount));
     assertEq(bobWethDataBefore.baseDebt, 0);
 
-    console.log('balance pre', tokenList.dai.balanceOf(bob));
     // Bob repays half of principal debt
     vm.expectEmit(address(spoke1));
     emit ISpoke.Repaid(daiReserveId(spoke1), bob, daiRepayAmount);
     vm.prank(bob);
     spoke1.repay(daiReserveId(spoke1), daiRepayAmount);
-    console.log('balance after', tokenList.dai.balanceOf(bob));
 
     DataTypes.UserPosition memory bobDaiDataAfter = getUserInfo(spoke1, bob, daiReserveId(spoke1));
     DataTypes.UserPosition memory bobWethDataAfter = getUserInfo(
@@ -209,6 +207,8 @@ contract SpokeRepayTest is SpokeBase {
     );
 
     assertEq(bobDaiDataAfter.suppliedShares, bobDaiDataBefore.suppliedShares);
+    assertEq(bobDaiDataAfter.outstandingPremium, 0, 'bob dai outstanding premium final balance');
+    assertEq(bobDaiDataAfter.baseDebt, daiBorrowAmount, 'bob dai base debt final balance');
     assertEq(
       bobDaiDataAfter.baseDebt + bobDaiDataAfter.outstandingPremium,
       daiBorrowAmount,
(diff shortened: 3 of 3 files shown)


src/interfaces/ISpoke.sol
```diff
@@ -97,4 +99,8 @@ interface ISpoke {
     uint256 reserveId,
     address user
   ) external view returns (DataTypes.UserPosition memory);
+  function liquidityHub() external view returns (ILiquidityHub);
+  function oracle() external view returns (IPriceOracle);
+  function reservesList(uint256) external view returns (uint256);
```
DhairyaSethi, line 104: this will always map 1-1 exactly, if the input is index in reserve list. 
unless the intention is to have input/out is assetId or index to sorted reserveId by liq premium? in which would appreciate implementing a new method with a different name (since this currently points to the reserveList array which is to rm-ed

tests/Base.t.sol
```diff
@@ -515,22 +516,22 @@ abstract contract Base is Test {
   }
 
   // assumes spoke has usdx supported
```
DhairyaSethi, line 518: we can have these helpers revert if spoke doesn't have the asset supported, to save accidental footguns & time debugging. wdyt? we would either need a reverse map (assetId=>reserveId) on the spoke (useful?) or this method just searches for it or we make it a convention in the code-base that ids start from 1 instead of 0

src/contracts/Spoke.sol
```diff
@@ -362,7 +361,10 @@ contract Spoke is ISpoke {
     DataTypes.UserPosition storage user,
     uint256 amount
   ) internal view {
-    uint256 suppliedAmount = liquidityHub.convertToAssetsDown(reserve.assetId, user.suppliedShares);
+    uint256 suppliedAmount = liquidityHub.convertToAssetsDown(
+      reserve.assetId,
+      user.suppliedShares
+    ) - user.baseDebt;
```
yan-man: @miguelmtzinf this is causing the failing tests - need to remove `user.baseDebt` from the withdraw validation as we agreed it shouldn't apply

src/contracts/LiquidityHub.sol
```diff
@@ -422,12 +422,10 @@ contract LiquidityHub is ILiquidityHub {
   ) internal view {
     // TODO: Other cases of status (frozen, paused)
     require(asset.config.active, AssetNotActive());
+    require(amountRestored > 0, InvalidRestoreAmount());
     // Ensure spoke is not restoring more than accrued drawn or equal 0
```
yan-man:
```suggestion
    // Ensure spoke is not restoring more than accrued drawn
```

tests/Base.t.sol
yan-man: > reverse map (assetId=>reserveId) on the spoke

imo this is very useful to be able to check the assets supported on spoke 

> or this method just searches for it

atm the only way is to loop over all reserves and check `assetId` right?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -0,0 +1,1253 @@
...
+    assertEq(bobDaiDataBefore.suppliedShares, 0);
+    assertEq(
+      bobDaiDataBefore.baseDebt + bobDaiDataBefore.outstandingPremium,
+      daiBorrowAmount,
+      'bob dai debt before'
+    );
+    assertEq(bobWethDataBefore.suppliedShares, hub.convertToShares(wethAssetId, wethSupplyAmount));
+    assertEq(bobWethDataBefore.baseDebt, 0);
+
+    console.log('balance pre', tokenList.dai.balanceOf(bob));
```
yan-man: is this intended to be left here? was just from debugging before right

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -0,0 +1,1253 @@
...
+    // Bob repays half of principal debt
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.Repaid(daiReserveId(spoke1), bob, daiRepayAmount);
+    vm.prank(bob);
+    spoke1.repay(daiReserveId(spoke1), daiRepayAmount);
+    console.log('balance after', tokenList.dai.balanceOf(bob));
```
yan-man: can remove

src/interfaces/ISpoke.sol
miguelmtzinf: this is just to complete the interface, so adding the getter for the array. We may want to rename/refactor later.

src/contracts/Spoke.sol
miguelmtzinf: ah ty, this went unoticed whem pulling main changes

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -0,0 +1,1253 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+contract SpokeRepayTest is SpokeBase {
+  using PercentageMath for uint256;
+
+  function test_repay_same_block2() public {
```
yan-man: can we rename this test to something more descriptive

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -0,0 +1,1253 @@
...
+    DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(spoke1, bob, daiReserveId(spoke1));
+    DataTypes.UserPosition memory bobWethDataBefore = getUserInfo(
+      spoke1,
+      bob,
+      wethReserveId(spoke1)
+    );
+    uint256 bobDaiBalanceBefore = tokenList.dai.balanceOf(bob);
+    uint256 bobWethBalanceBefore = tokenList.weth.balanceOf(bob);
+
+    assertEq(bobDaiDataBefore.suppliedShares, 0);
```
yan-man, line 172: I think also helpful to assert that `outstandingPremium` is now 0

tests/Base.t.sol
DhairyaSethi: >  only way is to loop over all reserves and check assetId right

yep that's the only way

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -0,0 +1,1253 @@
...
+
+    assertEq(
+      tokenList.dai.balanceOf(bob),
+      bobDaiBalanceBefore - daiRepayAmount,
+      'bob dai final balance'
+    );
+    assertEq(tokenList.weth.balanceOf(bob), bobWethBalanceBefore);
+  }
+
+  function test_repay_revertsWith_amount_exceeds_debt() public {
```
yan-man, line 466: can also do a version of this where user has a non-zero debt

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -0,0 +1,1253 @@
...
+      daiBorrowAmount,
+      'bob dai debt before'
+    );
+
+    // Bob repays
+    uint256 bobDaiInterest = bobDaiDataBefore.baseDebt +
+      bobDaiDataBefore.outstandingPremium -
+      daiBorrowAmount;
+    if (bobDaiInterest == 0) {
+      // no enough time travel for interest accrual
```
```suggestion
      // not enough time travel for interest accrual
```

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -0,0 +1,1253 @@
...
+  }
+
+  struct RepayMultipleLocal {
+    uint256 borrowAmount;
+    uint256 repayAmount;
+    DataTypes.UserPosition posBefore; // positionBefore
+    DataTypes.UserPosition posAfter; // positionAfter
+  }
+
+  /// borrow and replay multiple reserves
```
```suggestion
  /// borrow and repay multiple reserves
```

tests/Base.t.sol
CheyenneAtapour: I'm fine with a reverse map haha we had that before :)

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -0,0 +1,1281 @@
...
+    // Bob supply weth
+    Utils.spokeSupply(spoke1, wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    setUsingAsCollateral(spoke1, bob, wethReserveId(spoke1), true);
+
+    // Alice supply dai
+    Utils.spokeSupply(spoke1, daiReserveId(spoke1), alice, daiBorrowAmount, alice);
+
+    // Bob borrow dai
+    Utils.spokeBorrow(spoke1, daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+
```
yan-man, line 528: maybe overkill, but could also assert after the borrow that the actual collateral value in base units is >= debt in base units (ie that the _calcMinimumCollAmount method yielded correct output) 

supplied collateral * price / assetUnits  * LT >= debt amt * price / asset units

same for below where _calcMinimumCollAmount is used.

miguelmtzinf: i think that shouldn't be part of this test file and more of the healthfactor one for example. Here I could just supply a ton of assets as collateral, but wanted to do it a bit smart

tests/Base.t.sol
miguelmtzinf: is not enough just checking that the returned `reserveId` is not zero? or you want to check against the Spoke as well.
Also, a reverse map wouldn't work as we can have multiple reserves for one single asset.

DhairyaSethi: we currently allow 0 reserveId, & 0 assetId, that's why i mentioned the third option being we start the reserve/assetCounter from 1

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -223,4 +226,24 @@ contract SpokeBase is Base {
...
+    DataTypes.Reserve memory collData = spoke.getReserve(collReserveId);
+    uint256 collPrice = oracle.getAssetPrice(collData.assetId);
+    uint256 collAssetUnits = 10 ** hub.getAsset(collData.assetId).config.decimals;
+
+    DataTypes.Reserve memory debtData = spoke.getReserve(debtReserveId);
+    uint256 debtAssetUnits = 10 ** hub.getAsset(debtData.assetId).config.decimals;
+    uint256 debtPrice = oracle.getAssetPrice(debtData.assetId);
+
+    return
+      ((debtAmount * debtPrice * collAssetUnits) / (collPrice * debtAssetUnits)).percentDiv(
```
CheyenneAtapour, line 245: Do you have `collAssetUnits` and `debtAssetUnits` flipped?

DhairyaSethi: it's correct, normalised amount is amount/assetUnits

PR #216, opened 2025-02-28
interface cleanup

DhairyaSethi:


src/interfaces/ILiquidityHub.sol
```diff
@@ -101,7 +101,7 @@ interface ILiquidityHub {
    * @param assetId The asset id.
    * @param amount The amount of debt to draw.
    * @param riskPremium The new aggregated risk premium (in bps) of the calling spoke.
-   * @param to The address to draw debt to (user).
+   * @param to The address to credit the debt to.
```
miguelmtzinf: this is wrong no?, it's the address where assets are redirected to

DhairyaSethi: right yeah, i was writing just from the user's perspective, they're credited this debt for their prior supplied debits 
`to` is address where the debt is credited to, yes

how should we fix the wording?

src/interfaces/ISpoke.sol
```diff
@@ -43,9 +43,36 @@ interface ISpoke {
   function updateReserveConfig(uint256 reserveId, DataTypes.ReserveConfig calldata params) external;
   function updateLiquidityPremium(uint256 reserveId, uint256 liquidityPremium) external;
 
+  /**
+   * @notice Supply an `amount` of underlying asset of the specified reserve.
```
```suggestion
   * @notice Supply an amount of underlying asset of the specified reserve.
```
I d do it in this way instead, so amount does not replaced by the actual amount in the explorer. Same in withdraw and repay

src/interfaces/ISpoke.sol
```diff
@@ -43,9 +43,36 @@ interface ISpoke {
...
+  /**
+   * @notice Withdraw supplied `amount` of underlying asset from the specified reserve, sent at `to`.
+   * @param reserveId The reserveId of the underlying asset as registered on the spoke.
+   * @param amount The amount of asset to withdraw.
+   * @param to The address to transfer the assets to.
+   */
   function withdraw(uint256 reserveId, uint256 amount, address to) external;
+
+  /**
+   * @notice Allows the caller to draw debt from the specified reserve.
```
```suggestion
   * @notice Borrow an amount of underlying asset from the specified reserve.
```

src/interfaces/ISpoke.sol
```diff
@@ -43,9 +43,36 @@ interface ISpoke {
...
+   * @param reserveId The reserveId of the underlying asset as registered on the spoke.
+   * @param amount The amount of debt to draw.
```
```suggestion
   * @param amount The amount of underlying assets to borrow.
```

src/interfaces/ISpoke.sol
```diff
@@ -43,9 +43,36 @@ interface ISpoke {
...
+   * @param to The address to credit the debt to.
```
```suggestion
   * @param to The address to transfer the underlying assets to.
```

src/interfaces/ISpoke.sol
DhairyaSethi: > so amount does not replaced by the actual amount in the explorer

sorry, i don't follow? this is just md formatting which natspec supports, and etherscan etc format to highlight this is a param

miguelmtzinf: iirc, if you use `amount` parsers will replace the word with the value of the parameter
https://docs.soliditylang.org/en/latest/natspec-format.html#dynamic-expressions

src/interfaces/ISpoke.sol
DhairyaSethi: [0cd917c](https://github.com/aave/aave-v4/pull/216/commits/0cd917c49a2097d877a1ab3d667bafda53294e75)

src/interfaces/ILiquidityHub.sol
DhairyaSethi: [0cd917c](https://github.com/aave/aave-v4/pull/216/commits/0cd917c49a2097d877a1ab3d667bafda53294e75)

src/interfaces/ISpoke.sol
DhairyaSethi: ohh very interesting, didn't know this. yep rm, this is done this way in most codebases i've seen (including [v3](https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/interfaces/IPool.sol#L236) lol)

DhairyaSethi: it's a good thing tho, to have this, since it flows along with the sentance also? but removed [1ecdb29](https://github.com/aave/aave-v4/pull/216/commits/1ecdb29ec0e8fa503040525758f760a783dea21f)

src/interfaces/ILiquidityHub.sol
```diff
@@ -101,7 +102,7 @@ interface ILiquidityHub {
    * @param assetId The asset id.
    * @param amount The amount of debt to draw.
    * @param riskPremium The new aggregated risk premium (in bps) of the calling spoke.
-   * @param to The address to draw debt to (user).
+   * @param to The address to transfer the underlying assets to.
```
CheyenneAtapour, line 105: Not a big deal, but it's more formal to not end a sentence with a preposition (to). So can rewrite these to something like

`@param The address to which underlying assets are transferred.` or
`@param The address receiving the underlying assets`

src/interfaces/ISpoke.sol
```diff
@@ -43,9 +43,36 @@ interface ISpoke {
...
+  /**
+   * @notice Supply an amount of underlying asset of the specified reserve.
+   * @dev Liquidity Hub pulls underlying asset from caller, hence it needs prior approval.
+   * @param reserveId The reserveId of the underlying asset as registered on the spoke.
+   * @param amount The amount of asset to supply.
+   */
   function supply(uint256 reserveId, uint256 amount) external;
+
+  /**
+   * @notice Withdraw supplied amount of underlying asset from the specified reserve, sent at `to`.
```
CheyenneAtapour: the `sent at 'to'` part is a bit informal. I think it's even unnecessary (can just remove). Withdraw already indicates it must go somewhere else, and reader can see from later `param` comment where we specify the recipient address

src/interfaces/ISpoke.sol
miguelmtzinf: yeah, it's good but maybe not in a notice? nbd either way

src/interfaces/ISpoke.sol
DhairyaSethi: [3698a2b](https://github.com/aave/aave-v4/pull/216/commits/3698a2bec01112ab325b28abd89e3b7d75ae72e8)

PR #219, opened 2025-03-03
Health Factor Checks in Withdraw/Borrow

yan-man:
closes https://github.com/aave/aave-v4/issues/12

### DIFF
--- a/src/interfaces/ISpoke.sol
+++ b/src/interfaces/ISpoke.sol
@@ -91,7 +91,9 @@ interface ISpoke {
   function getReserveCumulativeDebt(uint256 reserveId) external view returns (uint256);
   function getReserveRiskPremium(uint256 reserveId) external view returns (uint256);
   function getUserRiskPremium(address user) external view returns (uint256);
-  function getUserAccountData(address user) external view returns (uint256, uint256, uint256);
+  function getUserAccountData(
+    address user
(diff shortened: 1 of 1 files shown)


src/contracts/Spoke.sol
```diff
@@ -586,7 +624,13 @@ contract Spoke is ISpoke {
       vars.userRiskPremium = (vars.userRiskPremium / vars.totalCollateralInBaseCurrency).rayify();
     }
 
-    return (vars.userRiskPremium, vars.avgLiquidationThreshold, vars.healthFactor);
+    return (
+      vars.userRiskPremium,
+      vars.avgLiquidationThreshold,
+      vars.healthFactor,
+      vars.totalCollateralInBaseCurrency,
+      vars.totalDebtInBaseCurrency
```
DhairyaSethi, line 659: fyi these are re-used to rp calc & not up to date at this point, let's only keep these two if necessary (& if so, then cache them before being re-used or something)

yan-man: yes they're cached, and seeing whether it's worthwhile to keep

DhairyaSethi: they're not cached in the current impl

yan-man:
see: https://github.com/aave/aave-v4/pull/233 with updates from user rp precision fix: https://github.com/aave/aave-v4/pull/227

PR #227, opened 2025-03-05
precise user rp

### LINKED ISSUE #228: User Risk Premium accounts for dust value in denominator
so the user for whom risk premium is being rounded down to zero (bc they supply 1 wei, and it gets rounded down to 0 while normalising by price precision) still contribute to the whole reserve risk premium weighted average because their debt is still accounted for in the running sum of weights (total debt)

the solution is that while calculating risk premium, we can not normalise asset value by the oracle precision (8) but rather need a different value which is > max asset units supported, so assuming max 18 decimal assets supported, normalising factor would be wad, hence calc becomes
`(assetAmount * assetPrice * 1e18) / assetUnits`
such that the smallest possible values, ie 1 wei, and 1e-8 usd price do not rounded off

DhairyaSethi:
closes #228 

todo - rename `{collateral,Debt}ValueInBaseCurrency` to `{collateral,Debt}ValueWadified` or something similar?

rayify() is not needed wadify() because wadify covers enough precision considering **max allowed token** is of 18 decimals

because we don't have bits left in the keyValueList slot (as per current implementation)
32 bits for key (ie liq premium) - (17 bits leads to max 1.3k bps is sufficient, we can switch to this)
leaves 224 bits for normalisedValue 
- with wad, max normalised value => 1e30 * 1e15 * 1e18 => 210 bits needed
- with ray, max normalised value => 1e30 * 1e15 * 1e27 => 239 bits needed

### DIFF
--- a/tests/unit/KeyListInMemory.t.sol
+++ b/tests/unit/KeyListInMemory.t.sol
@@ -10,7 +10,7 @@ contract KeyValueListInMemoryTest is Test {
   function test_fuzz_sortByKey(uint256[] memory seed) public pure {
     vm.assume(seed.length > 0);
     KeyValueListInMemory.List memory list = KeyValueListInMemory.init(seed.length);
-    for (uint256 i = 0; i < seed.length; ++i) {
+    for (uint256 i; i < seed.length; ++i) {
       list.add(i, _truncateKey(seed[i]), _truncateValue(seed[i]));
     }
     list.sortByKey();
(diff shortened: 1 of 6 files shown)


tests/unit/KeyListInMemory.t.sol
```diff
@@ -10,39 +10,50 @@ contract KeyValueListInMemoryTest is Test {
...
-      prevKey = key;
-    }
+    _assertSortedOrder(list);
   }
 
   function test_fuzz_sortByKey_length(uint256 length) public {
     length = bound(length, 1, 1e2);
     KeyValueListInMemory.List memory list = KeyValueListInMemory.init(length);
-    for (uint256 i = 0; i < length; ++i) {
+    for (uint256 i; i < length; ++i) {
```
yan-man, line 23: nbd either way but imo dont need to gas optimize in the tests, we can be explicit

DhairyaSethi: just being consistent

PR #229, opened 2025-03-05
Spoke Accrue Interest Tests

### LINKED ISSUE #144: Spoke RP - Accrue Interest Unit Tests

CheyenneAtapour:
Closes #144


tests/Base.t.sol
```diff
@@ -812,20 +812,15 @@ abstract contract Base is Test {
...
-    // asset.suppliedShares = hub.getAssetSuppliedShares(assetId);
-    // asset.availableLiquidity = hub.getAvailableLiquidity(assetId);
-    // (asset.baseDebt, asset.outstandingPremium) = hub.getAssetDebt(assetId);
-    // asset.baseBorrowIndex = hub.getAsset(assetId).baseBorrowIndex;
-    // asset.baseBorrowRate = hub.getBaseInterestRate(assetId);
-    // asset.riskPremium = hub.getAssetRiskPremium(assetId);
-    // asset.lastUpdateTimestamp = hub.getAsset(assetId).lastUpdateTimestamp;
-    // asset.config = hub.getAssetConfig(assetId);
-    // return asset;
+  function getSpokeInfo(
```
yan-man, line 873: i was thinking for these, like before we would call getters on each field rather than merely reading from storage

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1095 @@
...
+  }
+
+  /// Supply an asset only, and check no interest accrued.
+  function test_accrueInterest_OnlySupply(uint40 skipTime) public {
+    skipTime = uint40(bound(skipTime, 0, MAX_SKIP_TIME));
+    uint256 amount = 1000e18;
+    uint256 daiReserveId = _daiReserveId(spoke1);
+
+    // Bob supplies through spoke 1
+    Utils.supply(spoke1, daiReserveId, bob, amount, bob);
```
yan-man, line 62: hm seems skipTime isn't used? can add the skip here

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1095 @@
...
+
+    _checkDebts(spoke1, daiReserveId, bob, 0, 0, 'after supply, no interest accrued');
+  }
+
+  function test_accrueInterest_BorrowAndWait() public {
```
yan-man: i think this test can just call test_accrueInterest_fuzz_BorrowAndWait with skipTime of 1 year

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1095 @@
...
+    uint256 totalBase = _calcExpectedBaseDebt(amount, baseBorrowRate, startTime);
+    uint256 expectedPremiumDrawnShares = bobPosition.baseDrawnShares.percentMul(userRp);
+    uint256 expectedPremiumDebt = hub.convertToDrawnAssets(daiAssetId, expectedPremiumDrawnShares) -
+      bobPosition.premiumOffset +
+      bobPosition.realizedPremium;
+
+    _checkDebts(spoke1, daiReserveId, bob, totalBase, expectedPremiumDebt, 'after accrual');
+  }
+
+  function test_accrueInterest_fuzz_BorrowAndWait(uint40 skipTime) public {
```
yan-man: can just call test_accrueInterest_fuzz_BorrowAmountAndskipTime with inputs

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1095 @@
...
+    uint256 totalBase = _calcExpectedBaseDebt(amount, baseBorrowRate, startTime);
+    uint256 expectedPremiumDrawnShares = bobPosition.baseDrawnShares.percentMul(userRp);
+    uint256 expectedPremiumDebt = hub.convertToDrawnAssets(daiAssetId, expectedPremiumDrawnShares) -
+      bobPosition.premiumOffset +
+      bobPosition.realizedPremium;
+
+    _checkDebts(spoke1, daiReserveId, bob, totalBase, expectedPremiumDebt, 'after accrual');
+  }
+
+  function test_accrueInterest_fuzz_BorrowAmountAndskipTime(
```
yan-man:
```suggestion
  function test_accrueInterest_fuzz_BorrowAmountAndSkipTime(
```

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1095 @@
...
+      expectedPremiumDrawnShares
+    ) -
+      bobPosition.premiumOffset +
+      bobPosition.realizedPremium;
+
+    _checkDebts(spoke1, usdxReserveId, bob, expectedBaseDebt, expectedPremiumDebt, 'after accrual');
+  }
+
+  // Fuzz a mix of borrowed and supplied assets for bob, check his RP, ensure correct interest accrual
+  function test_accrueInterest_fuzz_RPBorrowAndskipTime(
```
yan-man:
```suggestion
  function test_accrueInterest_fuzz_RPBorrowAndSkipTime(
```

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1095 @@
...
+import {Spoke} from 'src/contracts/Spoke.sol';
+import {LiquidityHub} from 'src/contracts/LiquidityHub.sol';
+
+contract SpokeAccrueInterestTest is SpokeBase {
+  using SharesMath for uint256;
+  using WadRayMath for uint256;
+  using WadRayMathExtended for uint256;
+  using PercentageMath for uint256;
+
+  uint256 public constant MAX_BPS = 1000_00;
```
yan-man: not sure if this is used

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1095 @@
...
+      _wbtcReserveId(spoke1),
+      bob,
+      totalBase,
+      expectedPremiumDebt,
+      'wbtc after accrual'
+    );
+  }
+
+  // Fuzz a mix of borrowed and supplied assets for bob and rates, check his RP, ensure correct interest accrual
+  function test_accrueInterest_fuzz_RatesRPBorrowAndskipTime(
```
yan-man:
```suggestion
  function test_accrueInterest_fuzz_RatesRPBorrowAndSkipTime(
```

tests/Base.t.sol
CheyenneAtapour: I was under the same impression @DhairyaSethi had thoughts about this. I'm fine to go either way

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Ty, removed

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: I just removed this altogether in favor of the other fuzz tests

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Just removed this altogether in favor of the other fuzz tests

tests/Base.t.sol
yan-man: oh did @DhairyaSethi disagree with this? thought he'd be in favor of the getters

tests/Base.t.sol
```diff
@@ -982,6 +977,72 @@ abstract contract Base is Test {
     );
   }
 
+  function _checkDebts(
```
miguelmtzinf: so this only covers the case where there is only 1 user and 1 spoke?
I d create 3 functions so they can be used across tests:
```
assertUserDebt(spoke, reserveId, user, expectedBaseDebt, expectedPremiumDebt);
assertReserveDebt(spoke, reserveId, expectedBaseDebt, expectedPremiumDebt);
assertAssetDebt(spoke, assetId, expectedBaseDebt, expectedPremiumDebt);
```

miguelmtzinf: wondering if we should add a param for `expectedTotalDebt` or just assume it's the sum of both... but worth adding total debt check as well

tests/Base.t.sol
```diff
@@ -1132,6 +1193,14 @@ abstract contract Base is Test {
     newBaseDebt = initialDrawnShares.rayMulUp(newDebtIndex);
   }
 
+  function _calcExpectedBaseDebt(
+    uint256 startingDebt,
+    uint256 borrowRate,
+    uint40 startTime
```
```suggestion
  /// @dev Calculate expected base debt based on specified borrow rate
  function calculateExpectedBaseDebt(
    uint256 initialDebt,
    uint256 borrowRate,
    uint40 startTime
```
feels more aligned with the one above

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1043 @@
...
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+import {Spoke} from 'src/contracts/Spoke.sol';
+import {LiquidityHub} from 'src/contracts/LiquidityHub.sol';
+
+contract SpokeAccrueInterestTest is SpokeBase {
+  using SharesMath for uint256;
+  using WadRayMath for uint256;
+  using WadRayMathExtended for uint256;
+  using PercentageMath for uint256;
```
miguelmtzinf:
```suggestion

contract SpokeAccrueInterestTest is SpokeBase {
  using SharesMath for uint256;
  using WadRayMathExtended for uint256;
  using PercentageMath for uint256;
```

miguelmtzinf: we gotta move to extended as much as we can

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1043 @@
...
+    DataTypes.UserPosition memory bobPosition = spoke1.getUserPosition(usdxReserveId, bob);
+
+    uint256 expectedBaseDebt = _calcExpectedBaseDebt(borrowAmount, baseBorrowRate, startTime);
+    uint256 expectedPremiumDrawnShares = bobPosition.baseDrawnShares.percentMul(riskPremium);
+    uint256 expectedPremiumDebt = hub.convertToDrawnAssets(
+      usdxAssetId,
+      expectedPremiumDrawnShares
+    ) -
+      bobPosition.premiumOffset +
+      bobPosition.realizedPremium;
```
miguelmtzinf: i d be better if we compute this independently instead of using parts from local computation and onchain data.
smt like debtDelta*rp works

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1043 @@
...
+      _deployLiquidity(
+        spoke1,
+        _wbtcReserveId(spoke1),
+        MAX_SUPPLY_AMOUNT - amounts.wbtcSupplyAmount
+      );
+    }
+
+    // Bob borrows dai from spoke 1
+    if (amounts.daiBorrowAmount > 0) {
+      Utils.borrow(spoke1, _daiReserveId(spoke1), bob, amounts.daiBorrowAmount, bob);
```
miguelmtzinf, line 313: wondering if we should move this logic inside the helper. We are doing this across all tests though

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
miguelmtzinf: let's create a helper too pls

miguelmtzinf:
can you please get these tests aligned with [LiquidityHubAccrueInterest.t.sol](https://github.com/aave/aave-v4/blob/main/tests/unit/LiquidityHub/LiquidityHubAccrueInterest.t.sol) ? Missing some cases, and also some names would be worth to keep the same.
Apart from that, we are checking debt side but not supply. I think we should add it if not too much effort

tests/Base.t.sol
CheyenneAtapour: I created 4 for all of them. I changed the original to reference a single user and am using that still (which calls all the new functions). Lemme know if that's an alright approach. Also added total debt check, assuming it will be the sum of both expected debts

tests/Base.t.sol
CheyenneAtapour: shouldn't we be using `_` to prefix these anyway, because these are internal functions? I applied the rest of the suggestions

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: I'm not so sure this works when there are 2 accruals, because the new premium is only applied to part of the debt. Also it introduces more precision loss than the current approach

CheyenneAtapour:
> can you please get these tests aligned with [LiquidityHubAccrueInterest.t.sol](https://github.com/aave/aave-v4/blob/main/tests/unit/LiquidityHub/LiquidityHubAccrueInterest.t.sol) ? Missing some cases, and also some names would be worth to keep the same.

I saw 1 missing test case which I added. Some of the names are out of date because we use `skipTime` now instead of `elapsed`. Also some of the fuzz tests I felt were redundant, because other fuzz tests fuzz the same variables and more, so I removed the redundant ones. When we refactor `LiquidityHubAccrueInterest` tests, will align with these

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: I kinda liked keeping extra logic like this outside of utils, so we can also catch such edge cases in tests. I don't mind adding if you think it's better

tests/Base.t.sol
DhairyaSethi: yeah earlier the thinking was to reduce changes during code refactors which change the underlying accounting
now that's solidified, i was thinking we could have this getter return an adapted struct which also includes up to date debt and collateral balance values using the specific up to date getters

 it could clean up the tests slightly, but might be too much effort atm. wdyt?

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+
+    // Check bob's base debt and premium debt for all assets at user, reserve, spoke, and asset level
+    DataTypes.UserPosition memory bobPosition = spoke1.getUserPosition(_daiReserveId(spoke1), bob);
+    totalBase = _calculateExpectedBaseDebt(
+      amounts.daiBorrowAmount,
+      rates.daiBaseBorrowRate,
+      startTime
+    );
+    uint256 expectedPremiumDrawnShares = bobPosition.baseDrawnShares.percentMul(bobRp);
+    uint256 expectedPremiumDebt = hub.convertToDrawnAssets(daiAssetId, expectedPremiumDrawnShares) -
```
DhairyaSethi: here we're still using the contract's debt ex rate as expected and asserting against the same, like expectedBaseDebt above we can reuse the index to get expectedPremiumDrawnAssets; but nbd

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+          usingVirtualBalance: true
+        });
+      vm.mockCall(
+        address(irStrategy),
+        abi.encodeWithSelector(
+          IReserveInterestRateStrategy.calculateInterestRates.selector,
+          params
+        ),
+        abi.encode(rates.wethBaseBorrowRate)
+      );
```
DhairyaSethi: exact mocking feels overkill, can just partial match on calcInterestRate selector but nbd

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+          hub.convertToDrawnAssets(daiAssetId, 1),
+          bob
+        );
+      }
+      // Construct mock call so we can see the same user rp calc as within the borrow function
+      vm.mockCall(
+        address(spoke2),
+        abi.encodeWithSelector(Spoke.getUserTotalDebt.selector, _daiReserveId(spoke2), bob),
+        abi.encode(spoke2.getUserTotalDebt(_daiReserveId(spoke2), bob) + 1e18) // Debt amount seen in the borrow function when calculating user rp
+      );
```
DhairyaSethi: this is the workaround for #421, right? would appreciate if we link issue on the comment in case we have a fix
also wondering why it works bc wouldn't this be needed for all other drawn positions of bob? bc of the notify?

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+        bob,
+        amounts.wbtcBorrowAmount,
+        expectedPremiumDebt,
+        'wbtc before second accrual'
+      );
+      baseShares.wbtc = bobPosition.baseDrawnShares;
+
+      // Store index before accrual, and use this for calculating expected base debt
+      Indices memory indices;
+      indices.daiIndex = hub.getAsset(daiAssetId).baseDebtIndex;
```
DhairyaSethi: let's expose `hub.getNormalisedDebt(assetId)` like v3 so this is more future proof

DhairyaSethi: also all of these are not the up to date indexes here, that doesn't matter in the test below?

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+      );
+
+      indices.usdxIndex = calculateExpectedDebtIndex(
+        indices.usdxIndex,
+        rates.usdxBaseBorrowRate,
+        beginningTime
+      );
+      bobPosition = spoke2.getUserPosition(_usdxReserveId(spoke2), bob);
+      totalBase = baseShares.usdx.rayMulUp(indices.usdxIndex);
+      expectedPremiumDrawnShares = baseShares.usdx.percentMul(bobRp);
```
DhairyaSethi: can we rename this to baseTotal or something? more straightforward and consistent with the existing nomenclature of the test having the type as suffix. but nbd

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+
+      // Check bob's base debt and premium debt for all assets at user, reserve, spoke, and asset level
+      indices.daiIndex = calculateExpectedDebtIndex(
+        indices.daiIndex,
+        rates.daiBaseBorrowRate,
+        startTime
+      );
+      bobPosition = spoke2.getUserPosition(_daiReserveId(spoke2), bob);
+      totalBase = baseShares.dai.rayMulUp(indices.daiIndex);
+      expectedPremiumDrawnShares = baseShares.dai.percentMul(bobRp);
```
DhairyaSethi: do you think we should make this rounding direction explicit in this PR itself? would involve slight change on Spoke to set the exact rounding direction during premium calc (and potentially some other tests - if too many break we can avoid and tackle in a separate PR)

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+    amounts.daiBorrowAmount = bound(amounts.daiBorrowAmount, 0, MAX_SUPPLY_AMOUNT / 2);
+    amounts.wethBorrowAmount = bound(amounts.wethBorrowAmount, 0, MAX_SUPPLY_AMOUNT / 2);
+    amounts.usdxBorrowAmount = bound(amounts.usdxBorrowAmount, 0, MAX_SUPPLY_AMOUNT / 2);
+    amounts.wbtcBorrowAmount = bound(amounts.wbtcBorrowAmount, 0, MAX_SUPPLY_AMOUNT / 2);
+
+    return amounts;
+  }
+
+  function _bound(Rates memory rates) internal pure returns (Rates memory) {
+    rates.daiBaseBorrowRate = bound(rates.daiBaseBorrowRate, 1, 100_00);
```
DhairyaSethi: can use constant from PercentageMath

DhairyaSethi: also max borrow rate can be higher than 100%, no?

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+    rates.wethBaseBorrowRate = bound(rates.wethBaseBorrowRate, 1, 100_00);
+    rates.usdxBaseBorrowRate = bound(rates.usdxBaseBorrowRate, 1, 100_00);
+    rates.wbtcBaseBorrowRate = bound(rates.wbtcBaseBorrowRate, 1, 100_00);
+
+    // Put rates in ray
+    rates.daiBaseBorrowRate =
+      (rates.daiBaseBorrowRate * WadRayMathExtended.RAY) /
+      PercentageMath.PERCENTAGE_FACTOR;
```
DhairyaSethi: can use a test helper _bpsToRay for now, nbd

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1250 @@
...
+        _getValueInBaseCurrency(wethAssetId, amounts.wethSupplyAmount) +
+        _getValueInBaseCurrency(usdxAssetId, amounts.usdxSupplyAmount) +
+        _getValueInBaseCurrency(wbtcAssetId, amounts.wbtcSupplyAmount) >
+        2 *
+          (_getValueInBaseCurrency(daiAssetId, amounts.daiBorrowAmount) +
+            _getValueInBaseCurrency(wethAssetId, amounts.wethBorrowAmount) +
+            _getValueInBaseCurrency(usdxAssetId, amounts.usdxBorrowAmount) +
+            _getValueInBaseCurrency(wbtcAssetId, amounts.wbtcBorrowAmount))
+    );
+  }
```
DhairyaSethi: ah man i hope this doesn't make the test reject on higher fuzz runs 🫠

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
DhairyaSethi: we can use cached indexes maybe?

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Users only have 1 risk premium, and it's calculated using this mocked value, so the rest use this same calculated risk premium

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: We could easily handle by just returning out of the test in this case if we run into that issue. E.g. Switch this function to return a boolean and then return out of the function accordingly if this condition isn't met

DhairyaSethi: return exits the test, wastes a fuzz run and give a false sense of covered cases bc it's counted as a passing test. preference would be bound params instead

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
DhairyaSethi: yes but my point is on the debt positions. there will be mismatch will be for all others regardless, no?

CheyenneAtapour: All the debt positions are computed using this "mismatched" user RP, which we are accounting for here by calculating it how the protocol sees it when it calculates

DhairyaSethi: right so the other way around, makes sense ty! 
it's unfortunate we were seeing very high differences in premium debt bc of the token decimal otherwise approxEq could have worked here. 
sounds good 

(nit: pls encodeCall over encodeWithSelector)

tests/Base.t.sol
CheyenneAtapour: Maybe we can do that on a separate PR related to test cleanup

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Aren't they all up to date? Because an action just happened on the same block as using these getters

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: I'm not a fan of baseTotal, but it could also just be `baseDebt`

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Yeah, I now used the max from irStrategy

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
DhairyaSethi: we should note to do that for all existing tests

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: @miguelmtzinf Indeed, I see a precision delta of 12 wei with that approach, even with 1 accrual, but I've added this approach with the helper to all but 2 tests

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: I added the comment about it being a workaround, sorry missed that earlier

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Cool, I just did a quick look and didn't immediately see other tests that weren't doing this. So I made an issue: https://github.com/orgs/aave/projects/14?pane=issue&itemId=114073962&issue=aave%7Caave-v4%7C437

Currently, we define again the constant MAX_BORROW_RATE in our tests to match the value in the `irStrategy`. I mention in the above linked issue that we can instead pull directly from that contract, so it's more robust / future proof

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Ah, wait, you mean after the time skip. Indeed I calculate what the expected debt index should be, which is part of what I'm testing

CheyenneAtapour: On second thought, I'm fine to expose that function, but there's something nice about what I'm testing. Since the assets besides dai haven't been interacted with, I'm calculating their expected indices from the _beginning_ time t0 of this test. There are comments about this in the following calculations

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: This seems to cause some issues and even underflow, if I'm understanding correctly. Also feels best for another PR

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Alright, changed this to a bound

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
miguelmtzinf: which edge cases are you thinking about?

CheyenneAtapour: For example, reverting on trying to borrow 0. However, I suppose we can and should test such edge cases in the specific test suite where we don't use the Utils and instead directly call the function

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: This is now being done in https://github.com/aave/aave-v4/pull/440 and hopefully can improve the assertions on this PR

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1110 @@
...
+    );
+
+    bobPosition = spoke1.getUserPosition(_usdxReserveId(spoke1), bob);
+    baseDebt = _calculateExpectedBaseDebt(
+      amounts.usdxBorrowAmount,
+      rates.usdxBaseBorrowRate,
+      startTime
+    );
+    expectedPremiumDrawnShares = bobPosition.baseDrawnShares.percentMulUp(bobRp);
+    expectedPremiumDebt =
```
miguelmtzinf, line 940: we still cannot use calc helpers?

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -0,0 +1,1110 @@
...
+    );
+
+    bobPosition = spoke1.getUserPosition(_wbtcReserveId(spoke1), bob);
+    baseDebt = _calculateExpectedBaseDebt(
+      amounts.wbtcBorrowAmount,
+      rates.wbtcBaseBorrowRate,
+      startTime
+    );
+    expectedPremiumDrawnShares = bobPosition.baseDrawnShares.percentMulUp(bobRp);
+    expectedPremiumDebt =
```
miguelmtzinf, line 986: here too

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Created an issue here: https://github.com/orgs/aave/projects/14/views/1?pane=issue&itemId=116146155&issue=aave%7Caave-v4%7C466

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
CheyenneAtapour: Same issue as above

Issue #220, opened 2025-03-04
Additional tests for Spoke repay


miguelmtzinf:
- Multiple borrows and repays by the same user (user borrows, repays, borrow more, repay again)
- Multiple users borrow and repay the same reserve.
- Multiple users borrow and repay different reserves.
- User cannot repay 1 unit of assets because the exchange rate has increased a lot and shares become 0.

yan-man:
catalogued in notion

PR #233, opened 2025-03-06
Health Factor Checks in Withdraw/Borrow

yan-man:
- add HF checks to borrow/withdraw, based on V3 implementation
- borrow checks HF at the beginning and end of the action
- withdraw checks HF at the end of the action

additional task to examine further the precision/rounding: https://github.com/aave/aave-v4/issues/234

closes https://github.com/aave/aave-v4/issues/12

### DIFF
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -23,7 +23,6 @@ import {MockPriceOracle, IPriceOracle} from './mocks/MockPriceOracle.sol';
 // dependencies
 import {IERC20Errors} from 'src/dependencies/openzeppelin/IERC20Errors.sol';
 import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
-import {IERC20Metadata} from 'src/dependencies/openzeppelin/IERC20Metadata.sol';
 import {WETH9} from 'src/dependencies/weth/WETH9.sol';
 
 abstract contract Base is Test {
@@ -32,9 +31,9 @@ abstract contract Base is Test {
   using PercentageMath for uint256;
 
   uint256 internal constant MAX_SUPPLY_AMOUNT = 1e30;
-  uint256 internal constant MAX_ALLOWABLE_DECIMALS = 18;
+  uint256 internal constant MAX_TOKEN_DECIMALS_SUPPORTED = 18;
   uint256 internal constant MAX_SUPPLY_ASSET_UNITS =
-    MAX_SUPPLY_AMOUNT / 10 ** MAX_ALLOWABLE_DECIMALS;
+    MAX_SUPPLY_AMOUNT / 10 ** MAX_TOKEN_DECIMALS_SUPPORTED;
   uint256 internal MAX_SUPPLY_AMOUNT_USDX;
   uint256 internal MAX_SUPPLY_AMOUNT_DAI;
   uint256 internal MAX_SUPPLY_AMOUNT_WBTC;
@@ -153,18 +152,10 @@ abstract contract Base is Test {
     vm.label(address(tokenList.dai), 'DAI');
     vm.label(address(tokenList.wbtc), 'WBTC');
 
-    MAX_SUPPLY_AMOUNT_USDX =
-      MAX_SUPPLY_ASSET_UNITS *
-      10 ** IERC20Metadata(address(tokenList.usdx)).decimals();
-    MAX_SUPPLY_AMOUNT_WETH =
-      MAX_SUPPLY_ASSET_UNITS *
-      10 ** IERC20Metadata(address(tokenList.weth)).decimals();
-    MAX_SUPPLY_AMOUNT_DAI =
-      MAX_SUPPLY_ASSET_UNITS *
-      10 ** IERC20Metadata(address(tokenList.dai)).decimals();
-    MAX_SUPPLY_AMOUNT_WBTC =
-      MAX_SUPPLY_ASSET_UNITS *
-      10 ** IERC20Metadata(address(tokenList.wbtc)).decimals();
+    MAX_SUPPLY_AMOUNT_USDX = MAX_SUPPLY_ASSET_UNITS * 10 ** tokenList.usdx.decimals();
+    MAX_SUPPLY_AMOUNT_WETH = MAX_SUPPLY_ASSET_UNITS * 10 ** tokenList.weth.decimals();
+    MAX_SUPPLY_AMOUNT_DAI = MAX_SUPPLY_ASSET_UNITS * 10 ** tokenList.dai.decimals();
+    MAX_SUPPLY_AMOUNT_WBTC = MAX_SUPPLY_ASSET_UNITS * 10 ** tokenList.wbtc.decimals();
 
     address[3] memory users = [alice, bob, carol];
 


src/contracts/Spoke.sol
```diff
@@ -565,40 +599,42 @@ contract Spoke is ISpoke {
       }
     }
 
+    // (avgLiquidationThreshold / totalCollateral) * totalCollateral can be simplified to avgLiquidationThreshold
+    // then convert from BPS wad back into wad
+    vars.healthFactor = vars.totalDebtInBaseCurrency == 0
+      ? type(uint256).max
+      : vars.avgLiquidationThreshold.wadDiv(vars.totalDebtInBaseCurrency).percentMul(1); // HF of 1 -> 1e18
```
yan-man: this is the biggest change. Instead of first finding the avgLT, which reduces precision to BPS, it just simplifies the multiplication. 

avgLT = weighted LT / coll

Previous: avgLT * coll / debt
now: (weighted LT / coll) * coll / debt -> allow cancelling coll

src/contracts/Spoke.sol
```diff
@@ -198,6 +201,7 @@ contract Spoke is ISpoke {
     });
     liquidityHub.draw(reserve.assetId, amount, uint32(newReserveRiskPremium.derayify()), to);
     _notifyRiskPremiumUpdate(reserve.assetId, msg.sender, newUserRiskPremium);
+    _validateHealthFactor(msg.sender);
```
DhairyaSethi, line 202: let's fail early & do this before notify

src/contracts/Spoke.sol
DhairyaSethi: why percentMul(1) here?

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -370,4 +374,258 @@ contract SpokeBorrowTest is Base {
...
+
+    // cannot borrow more dai
+    vm.prank(bob);
+    vm.expectRevert(ISpoke.HealthFactorLowerThanLiquidationThreshold.selector);
+    spoke1.borrow(daiReserveId, 1e1, bob);
+
+    // cannot borrow more usdx
+    vm.prank(bob);
+    vm.expectRevert(ISpoke.HealthFactorLowerThanLiquidationThreshold.selector);
+    spoke1.borrow(usdxReserveId, 1e3, bob);
```
DhairyaSethi: should not be able to borrow 1 wei either, yes? let's try that as dust instead maybe

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -370,4 +374,258 @@ contract SpokeBorrowTest is Base {
...
+
+    // Bob draw max allowed debt amt of dai reserve liquidity
+    vm.prank(bob);
+    spoke1.borrow(daiReserveId, daiDebtAmount, bob);
+
+    vm.prank(bob);
+    spoke1.borrow(usdxReserveId, usdxDebtAmount, bob);
+
+    // valid HF
+    assertGe(spoke1.getHealthFactor(bob), spoke1.HEALTH_FACTOR_LIQUIDATION_THRESHOLD());
```
DhairyaSethi: can we do one which asserts the exact HF value instead of >,<

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -370,4 +374,258 @@ contract SpokeBorrowTest is Base {
...
+    vm.expectRevert(ISpoke.HealthFactorLowerThanLiquidationThreshold.selector);
+    spoke1.borrow(daiReserveId, 1e18, bob);
+
+    // cannot borrow more usdx
+    vm.prank(bob);
+    vm.expectRevert(ISpoke.HealthFactorLowerThanLiquidationThreshold.selector);
+    spoke1.borrow(usdxReserveId, 1e6, bob);
+  }
+
+  function test_borrow_revertsWith_HealthFactorLowerThanLiquidationThreshold_multiple_colls_with_interest()
```
DhairyaSethi: can we write a fuzzed variant which fuzzes amounts and delay, just to root out any obvious edge cases in this scenario

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -370,4 +374,258 @@ contract SpokeBorrowTest is Base {
...
+    Utils.spokeSupply(spoke1, daiReserveId, alice, maxDebtAmount * 2, alice); // supply enough buffer for multiple borrows
+
+    // Bob draw max allowed debt amt of dai reserve liquidity
+    vm.prank(bob);
+    spoke1.borrow(daiReserveId, maxDebtAmount, bob);
+
+    assertEq(spoke1.getHealthFactor(bob), spoke1.HEALTH_FACTOR_LIQUIDATION_THRESHOLD());
+
+    // collateral price drop by half so that bob is undercollateralized
+    oracle.setAssetPrice(wethAssetId, 1000e8);
```
DhairyaSethi: nbd 

```suggestion
    oracle.setAssetPrice(wethAssetId, oracle.getAssetPrice(wethAssetId) / 2);
```

src/contracts/Spoke.sol
```diff
@@ -794,4 +830,12 @@ contract Spoke is ISpoke {
 
     return newReserveRiskPremium;
   }
+
+  function _validateHealthFactor(address userAddress) internal view {
+    (, , uint256 healthFactor, , ) = _calculateUserAccountData(userAddress);
```
DhairyaSethi, line 769: i'm thinking that we always delay the HF check till after the debt action has been taken into account, to not have this function be dup called (since it's also in updateBaseDebtAndRiskPremium) 

but can think about during  refactor

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -281,22 +281,22 @@ contract SpokeBorrowTest is Base {
     uint256 wethReserveId = spokeInfo[spoke2].weth.reserveId;
```
DhairyaSethi, line 333: these var names are all shadowed with the helper method in base (func wethReserveId(spoke)), would appreciate a rename or something else for clarity

src/contracts/Spoke.sol
DhairyaSethi: so it looks like you're trying to remove the bps factor since `avgLiquidationThreshold` is in bps
> BPS wad back into wad

it's not super clear from this comment tbh, but i get it. would appreciate more info here (such has bps factor is leftover from avgLiqThreshold or something similar)

since `percentMul(1)` is essentially equal to `div(1e4)`, maybe we introduce another method called `.fromBps()` which does `/ 1e4` (current convention set for this is along the lines of `debpsify` lol, but the former has a better ring to it)
cc @miguelmtzinf

src/contracts/Spoke.sol
yan-man: yeah don't think it's very optimized atm but just wanted to get the logic in first

tests/unit/Spoke/Spoke.Borrow.t.sol
DhairyaSethi: @yan-man i like this helper from the v3 codebase, let's add this & we can re-use this in repay tests later
https://github.com/aave-dao/aave-v3-origin/blob/main/tests/utils/TestnetProcedures.sol#L356-L360

cc @miguelmtzinf

src/contracts/Spoke.sol
yan-man: going with `fromBps` for now, and will add more clarifying comments

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: renamed the internal helpers method instead to `_wethReserveId`

DhairyaSethi: yeah lol i wrote them with an _ originally ~~but @miguelmtzinf removed that in one his PRs~~. i'm fine with moving in either direction tbh, nbd

src/contracts/Spoke.sol
yan-man: added here: https://github.com/aave/aave-v4/issues/240

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: haha I think I was the one that removed it before

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: in some situations (ie besides price drops and interest accruals, where we KNOW the user has started with a HF < 1 already) where the action itself is supposed to drop the HF from >=1 to < 1, there are precision issues. Sometimes borrowing 1 unit more, or withdrawing 1 unit do not trigger a change in HF. I suspect this is due to precision loss relative to the total collateral/debt value in base units

fixed: https://github.com/aave/aave-v4/pull/233/commits/72e47a7a0bc17db4b39420a51da128d44f443594

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: did this where applicable / where it holds. Added comments for where precision issues arise, particularly happens due to multiple colls/debts 

fixed: https://github.com/aave/aave-v4/pull/233/commits/72e47a7a0bc17db4b39420a51da128d44f443594

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/233/commits/72e47a7a0bc17db4b39420a51da128d44f443594

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/233/commits/72e47a7a0bc17db4b39420a51da128d44f443594

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: added multiple tests here: https://github.com/aave/aave-v4/pull/233/commits/72e47a7a0bc17db4b39420a51da128d44f443594

and catalogued the precision issues in comments

src/contracts/Spoke.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/233/commits/72e47a7a0bc17db4b39420a51da128d44f443594

src/contracts/Spoke.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/233/commits/72e47a7a0bc17db4b39420a51da128d44f443594

src/contracts/WadRayMath.sol
```diff
@@ -187,4 +187,22 @@ library WadRayMath {
...
     return (a * 100_00) / RAY;
   }
+
+  /**
+   * @notice Removes BPS factor from a value.
+   * @param value The value of which to convert from BPS.
+   * @return result value (stripped of BPS factor).
+   */
+  function fromBps(uint256 value) internal pure returns (uint256) {
+    return value / 1e4;
```
DhairyaSethi: nit: for consistency can we make this below 100_00
or better yet, import percentage math & can use PercentageMath.PERCENTAGE_FACTOR
(above in rayToBps etc methods as well)

src/contracts/WadRayMath.sol
```diff
@@ -187,4 +187,22 @@ library WadRayMath {
...
+  }
+
+  /**
+   * @notice Adds BPS factor to a value.
+   * @param value The value of which to convert to BPS.
+   * @return result value (in BPS).
+   */
+  function toBps(uint256 value) internal pure returns (uint256) {
+    return value * 1e4;
```
DhairyaSethi: this needs overflow check, if we're not using toHps anywhere can also rm this it's fine imo

yan-man: ill just rm

https://github.com/aave/aave-v4/pull/233/commits/1660fafc1191f673a7e6e58ffb2240a357edc15f

src/contracts/WadRayMath.sol
yan-man: fixed: https://github.com/aave/aave-v4/pull/233/commits/1660fafc1191f673a7e6e58ffb2240a357edc15f

yan-man: moved to other lib: https://github.com/aave/aave-v4/pull/233/commits/dca32b117b3f1d416033dafd9d89ca92c7b1be98

src/contracts/Spoke.sol
```diff
@@ -578,40 +607,50 @@ contract Spoke is ISpoke {
       }
     }
 
+    // at this point avgCollateralFactor is a weighted sum of collateral scaled by collateralFactor
+    // (avgCollateralFactor / totalCollateral) * totalCollateral can be simplified to avgCollateralFactor
+    // strip BPS factor from result, because running avgCollateralFactor sum has been scaled by collateralFactor (in BPS) above
+    vars.healthFactor = vars.totalDebtInBaseCurrency == 0
+      ? type(uint256).max
+      : vars.avgCollateralFactor.wadDiv(vars.totalDebtInBaseCurrency).fromBps(); // HF of 1 -> 1e18
```
miguelmtzinf, line 615: the bps are not canceled out when dividing? (coll and debt are multiplied by 100_00)

yan-man: no, because right now avgCollateralFactor numerator is running total of `userCollateralInBaseCurrency * collateralFactor` from loop above, and CF is in BPS. `totalDebtInBaseCurrency` on denominator is lacking that BPS scaling that numerator has (merely the running total of `userDebtInBaseCurrency`

src/contracts/PercentageMath.sol
```diff
@@ -58,4 +58,13 @@ library PercentageMath {
       result := div(add(mul(value, PERCENTAGE_FACTOR), div(percentage, 2)), percentage)
     }
   }
+
+  /**
+   * @notice Removes BPS factor from a value.
+   * @param value The value of which to convert from BPS.
+   * @return result value (stripped of BPS factor).
```
DhairyaSethi:
```suggestion
   * @notice Truncates number from BPS precision. 
   * @param value The number in BPS precision. 
   * @return result (value / 1e4)
```

yan-man: https://github.com/aave/aave-v4/pull/233/commits/e3e2513f7d857a6176e99f58f706293aec4fd646

src/contracts/Spoke.sol
miguelmtzinf: calc HF twice then? it's also in `validateBorrow` fn

src/contracts/Spoke.sol
```diff
@@ -343,6 +341,32 @@ contract Spoke is ISpoke {
...
+      uint256 totalDebtInBaseCurrency
+    )
+  {
+    (
+      userRiskPremium,
+      avgCollateralFactor,
+      healthFactor,
+      totalCollateralInBaseCurrency,
+      totalDebtInBaseCurrency
+    ) = _calculateUserAccountData(user);
```
miguelmtzinf, line 373: cannot we just return? with not named returned variables

src/contracts/Spoke.sol
miguelmtzinf: better to make it work a la V3, where we calc HF at validation, and the amount of collateral needed for the new borrow. Then it's fine if we have enough coll.

src/contracts/Spoke.sol
yan-man: definitely - i just did it bc of the multiple outputs so purposefully was verbose here for clarity

src/contracts/Spoke.sol
miguelmtzinf: ah, true. Also I see it's because we do `*` instead of `percentMul` above (when coll * cf)
I guess it's more precise in current form

DhairyaSethi: cheaper (bc we avoid addition div(1e4) each iteration) and more precise (main div at the end) also this way like you said

src/interfaces/ISpoke.sol
```diff
@@ -35,6 +35,7 @@ interface ISpoke {
   error InvalidCollateralFactor();
   error InvalidLiquidationBonus();
   error InvalidReserveDecimals();
+  error HealthFactorLowerThanCollateralFactor();
```
miguelmtzinf: confusing. I d rename to smt like `HealthFactorBelowThreshold`

tests/Base.t.sol
```diff
@@ -672,4 +673,12 @@ abstract contract Base is Test {
   ) internal view returns (uint256) {
     return spoke.getUserSuppliedAmount(reserveId, user);
   }
+
+  // from AaveV3
+  // https://github.com/aave-dao/aave-v3-origin/blob/main/tests/utils/TestnetProcedures.sol#L356-L360
+  function _calcPrice(uint256 price, uint256 percent) public pure returns (uint256) {
```
miguelmtzinf: rename sig and params, change docs so explains what it does pls
no need to leave v3 ref

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -424,4 +432,952 @@ contract SpokeBorrowTest is Base {
...
+    vm.prank(bob);
+    spoke1.borrow(daiReserveId, maxDebtAmount, bob);
+
+    // valid HF after borrow
+    assertEq(spoke1.getHealthFactor(bob), spoke1.HEALTH_FACTOR_LIQUIDATION_THRESHOLD());
+
+    // cannot borrow a non trivial amount that brings HF below threshold
+    vm.prank(bob);
+    vm.expectRevert(ISpoke.HealthFactorLowerThanCollateralFactor.selector);
+    spoke1.borrow(daiReserveId, 1e4, bob); // TODO: update with exact amount, resolve precision
```
miguelmtzinf, line 460: just 1 wouldn't be enough to revert? maybe assets(1 share) right

tests/unit/Spoke/Spoke.Borrow.t.sol
```diff
@@ -424,4 +432,952 @@ contract SpokeBorrowTest is Base {
...
+    // invalid HF
+    assertLt(spoke1.getHealthFactor(bob), spoke1.HEALTH_FACTOR_LIQUIDATION_THRESHOLD());
+
+    // cannot borrow more usdx
+    vm.prank(bob);
+    vm.expectRevert(ISpoke.HealthFactorLowerThanCollateralFactor.selector);
+    spoke1.borrow(usdxReserveId, 1, bob);
+  }
+
+  // TODO: tests with other combos of collateral/debt, particularly with different units
```
miguelmtzinf, line 1353: also missing tests where HF increases because of collateral price increase

tests/unit/Spoke/Spoke.Repay.t.sol
miguelmtzinf, line 1: missing the following tests:
- repay debt increases HF
- HF < 1 goes up after debt repaid
- HF < 1, borrow reverts, repay, HF = 1, borrow reverts, repay, HF > 1, borrow ok

new ticket or to be added to the other issue for additional tests in repay #220

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -7,8 +7,12 @@ contract SpokeWithdrawTest is SpokeBase {
   using WadRayMath for uint256;
   using PercentageMath for uint256;
 
+  // WETH: $2000, 1e18 units
```
miguelmtzinf: remove?

tests/unit/Spoke/Spoke.Borrow.t.sol
DhairyaSethi: needs investigation imo 

https://github.com/aave/aave-v4/pull/233#discussion_r1984109905

cc @yan-man

tests/unit/Spoke/Spoke.Repay.t.sol
yan-man: had already created a ticket, added these additional ones here: https://github.com/aave/aave-v4/issues/249

tests/unit/Spoke/Spoke.Borrow.t.sol
yan-man: added to ticket: https://github.com/aave/aave-v4/issues/249

src/contracts/Spoke.sol
yan-man: running into a lot of precision issues in implementing this calculation and aligning with the tests back-calculation of min coll. I want to leave it as is here and tackle this after so that we can ship this first version for now. Added it here: https://github.com/aave/aave-v4/issues/240 

I'll work on it some more after merging this 

btw in v3 also it implicitly calculates HF twice, at the end during aToken transfer validation:
- https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/protocol/libraries/logic/ValidationLogic.sol#L241
- https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/protocol/libraries/logic/SupplyLogic.sol#L149

tests/unit/Spoke/Spoke.Withdraw.t.sol
yan-man: https://github.com/aave/aave-v4/pull/233/commits/54757d6bba3b24c70a51f00f1a2a2e2f0a6fdd35

tests/Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/233/commits/54757d6bba3b24c70a51f00f1a2a2e2f0a6fdd35

tests/Base.t.sol
```diff
@@ -701,11 +701,9 @@ abstract contract Base is Test {
...
   }
 
-  // from AaveV3
-  // https://github.com/aave-dao/aave-v3-origin/blob/main/tests/utils/TestnetProcedures.sol#L356-L360
-  function _calcPrice(uint256 price, uint256 percent) public pure returns (uint256) {
+  /// @dev Helper function to calculate a new price based on a percentage change
+  function calcNewPrice(uint256 price, uint256 percent) public pure returns (uint256) {
     if (percent == 0) return price;
-    uint256 result = price.percentMul(percent);
-    return result > price ? result - price : price - result;
```
yan-man: prev this was returning the price difference, but actually just want the new scaled price

src/contracts/Spoke.sol
```diff
@@ -578,40 +607,49 @@ contract Spoke is ISpoke {
...
     vars.i = 0;
-    // @dev from this point onwards, `totalCollateralInBaseCurrency` represents running collateral
-    // value used in risk premium, `totalDebtInBaseCurrency` represents running outstanding debt
-    vars.totalCollateralInBaseCurrency = 0;
-    while (vars.i < vars.collateralReserveCount && vars.totalDebtInBaseCurrency > 0) {
-      if (vars.totalDebtInBaseCurrency == 0) break;
+    // @dev from this point onwards, `collateralCounterInBaseCurrency` represents running collateral
+    // value used in risk premium, `debtCounterInBaseCurrency` represents running outstanding debt
+    while (vars.i < vars.collateralReserveCount && vars.debtCounterInBaseCurrency > 0) {
+      if (vars.debtCounterInBaseCurrency == 0) break;
```
CheyenneAtapour, line 629: If it's greater than 0 (which is required to enter the while loop), you would never hit this case here

src/contracts/WadRayMath.sol
```diff
@@ -181,10 +183,10 @@ library WadRayMath {
   }
 
   function bpsToRay(uint256 a) internal pure returns (uint256) {
-    return (a * RAY) / 100_00;
+    return (a * RAY) / PercentageMath.PERCENTAGE_FACTOR;
```
CheyenneAtapour, line 186: Why do we need to mix in the percentage math library here?

DhairyaSethi: it deals with BPS stuff. since this function deals with bps & ray, it's a subjective call on whether to put in percentageMath or wadRay. do you have a preference?

CheyenneAtapour: We can leave the function here, but I don't really like the import of another library here. Can we keep the `100_00` or redefine the constant and rename it? Or is that not clean enough?

src/contracts/Spoke.sol
yan-man: yeah you wouldnt hit this case on the first iteration of the loop but eventually you would bc debtCounterInBaseCurrency is getting reduced each time. Same as before right, with totalDebtInBaseCurrency as counter: https://github.com/aave/aave-v4/blob/main/src/contracts/Spoke.sol#L597C16-L597C39

src/contracts/WadRayMath.sol
DhairyaSethi: honestly no strong opinion, i was the one who initially suggested this but i'm equally ok with the prev or your approach
cc @yan-man @miguelmtzinf

src/contracts/Spoke.sol
CheyenneAtapour: when you hit 0, you would exit when you check the while loop. This is the same as before, but just realizing we can remove this check now

yan-man: ah gotcha

src/contracts/WadRayMath.sol
yan-man: I'm in favor of leaving this and reconsidering it, added here: https://github.com/aave/aave-v4/issues/240

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -243,10 +244,37 @@ contract SpokeBase is Base {
...
+    uint256 normalizedDebtAmount = (debtAmount * debtPrice).wadify() / debtAssetUnits;
+    uint256 normalizedCollPrice = collPrice.wadify() / collAssetUnits;
+
     return
-      ((debtAmount * debtPrice * collAssetUnits) / (collPrice * debtAssetUnits)).percentDiv(
-        collData.config.collateralFactor
-      ) + 1;
+      (normalizedDebtAmount.wadify() /
+        normalizedCollPrice.wadify().percentMul(collData.config.collateralFactor)) + 1;
+  }
```
DhairyaSethi, line 253: why are wadifying again on line 251-252? seems redundant, since we wadify den as well (instead of div by wad after division) we don't gain any precision as well

DhairyaSethi: @yan-man can we rm the extra wadify and also move around the algebra such that div(colAssetUnits) happens at the end, instead of in the middle, i _think_ that's the cause of our precision loss

yan-man: you mean this right: ((normalizedDebtAmount / normalizedCollPrice.wadify()).percentMul(collData.config.collateralFactor)) + 1

edit: this is breaking all the tests lol 

I did that to get more precision on the normalizedCollPrice.percentMul(CF)

src/contracts/WadRayMath.sol
miguelmtzinf: no strong opinion

tests/Base.t.sol
```diff
@@ -145,6 +153,19 @@ abstract contract Base is Test {
     vm.label(address(tokenList.dai), 'DAI');
     vm.label(address(tokenList.wbtc), 'WBTC');
 
+    MAX_SUPPLY_AMOUNT_USDX =
+      MAX_SUPPLY_ASSET_UNITS *
+      10 ** IERC20Metadata(address(tokenList.usdx)).decimals();
```
miguelmtzinf: `TestnetERC20` does not expose decimals directly?

tests/Base.t.sol
```diff
@@ -31,6 +32,13 @@ abstract contract Base is Test {
   using PercentageMath for uint256;
 
   uint256 internal constant MAX_SUPPLY_AMOUNT = 1e30;
+  uint256 internal constant MAX_ALLOWABLE_DECIMALS = 18;
```
miguelmtzinf: `MAX_TOKEN_DECIMALS_SUPPORTED` is clearer?

yan-man: https://github.com/aave/aave-v4/pull/233/commits/4f073de486ad0043f90558ab6002f3ba09bfe722

tests/Base.t.sol
yan-man: https://github.com/aave/aave-v4/commit/4f073de486ad0043f90558ab6002f3ba09bfe722

src/contracts/Spoke.sol
yan-man: didnt get around to this but added to ticket here: https://github.com/aave/aave-v4/issues/240

miguelmtzinf:
massive @yan-man :100:

tests/unit/Spoke/SpokeBase.t.sol
DhairyaSethi: nope, that's wrong imo, i meant 

((normalizedDebtAmount / normalizedCollPrice).percentMul(collData.config.collateralFactor)) + 1

tests/Base.t.sol
```diff
@@ -28,8 +28,16 @@ import {WETH9} from 'src/dependencies/weth/WETH9.sol';
...
+  using PercentageMath for uint256;
 
   uint256 internal constant MAX_SUPPLY_AMOUNT = 1e30;
+  uint256 internal constant MAX_TOKEN_DECIMALS_SUPPORTED = 18;
+  uint256 internal constant MAX_SUPPLY_ASSET_UNITS =
+    MAX_SUPPLY_AMOUNT / 10 ** MAX_TOKEN_DECIMALS_SUPPORTED;
+  uint256 internal MAX_SUPPLY_AMOUNT_USDX;
+  uint256 internal MAX_SUPPLY_AMOUNT_DAI;
+  uint256 internal MAX_SUPPLY_AMOUNT_WBTC;
+  uint256 internal MAX_SUPPLY_AMOUNT_WETH;
```
DhairyaSethi, line 40: don't like these non consts, since we know the decimals we're setting (or we make that a constant) these can be constants as well

PR #245, opened 2025-03-07
Refactor Spoke logic

### LINKED ISSUE #225: Refactor of LH and Spoke logic using stateless libraries
- Refactor logic by using stateless libraries that can be tested in isolation and segregating responsibilities (no calc + update, only calc)
- Refactor Spoke logic so it's similar to LH

reference conversation: https://github.com/aave/aave-v4/pull/140#discussion_r1949760548

miguelmtzinf:
This applies a minor refactor of the Spoke logic, so it's aligned with the logic flow on LiquidityHub.

We agreed to apply to refactor more once
1. Trust assumptions between hub<>spoke are reviewed again, which can lead to removal of duplicated data structures.
2. Attempt to replace index ratio mechanism to shares approach to simplify code complexity.
3. Clear approach for gas savings (cached object in memory or transient storage) and other topics(`accrueInterest` callable by spokes, etc)

Closes #225

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -25,7 +25,7 @@ contract Spoke is ISpoke {
   IPriceOracle public immutable oracle;
 
   mapping(address user => mapping(uint256 reserveId => DataTypes.UserPosition position))
-    internal _users;
+    internal _userPositions;
   mapping(address user => DataTypes.UserData data) internal _userData;
   mapping(uint256 reserveId => DataTypes.Reserve reserveData) internal _reserves;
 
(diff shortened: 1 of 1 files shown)


src/contracts/Spoke.sol
```diff
@@ -634,86 +629,8 @@ contract Spoke is ISpoke {
...
-      userData,
-      nextBaseBorrowIndex
-    );
-
-    user.baseDebt = cumulatedBaseDebt;
-    user.outstandingPremium = cumulatedOutstandingPremium;
-    user.baseBorrowIndex = nextBaseBorrowIndex;
-    user.lastUpdateTimestamp = block.timestamp;
+    reserve.accrueInterest(nextBaseBorrowIndex);
+    user.accrueInterest(userData, nextBaseBorrowIndex);
```
DhairyaSethi: wdyt about renaming all of these user objects to userPosition since it's more clear that these are a user position per reserveId

yan-man: agree but nice to have atm

DhairyaSethi: i would say imp to have for fresh eyes who aren't as used to the code as us

Issue #244, opened 2025-03-07
Missing User Risk Premium Tests


CheyenneAtapour:
Missing User Risk Premium Tests:

* user risk premium decreases after repayment
* user risk premium increases after withdrawal (or removal of collateral)
* user risk premium changes because of collateral accrual (no debt change)
* user risk premium changes because of debt accrual (no collateral change)
* user risk premium changes because of coll and debt accrual

PR #251, opened 2025-03-09
Add Missing RP Tests

### LINKED ISSUE #250: User Risk Premium In-Flight Test
We currently do in-flight checks of user risk premium in a couple of the user rp tetsts, but this value is not used, so the placement of the assertions is a bit awkward.

We want to move these to a dedicated test which rigorously checks user rp is calculated correctly in-flight. This is useful for when a user wants to check what their new user RP would be, and update it if it's lower

CheyenneAtapour:
Closes 

#250 
#244


CheyenneAtapour:
Closing in favor of #367

PR #274, opened 2025-03-20
fix: user unrealized premium impl

yan-man:


tests/ts/prototype/core.ts
```diff
@@ -295,7 +299,6 @@ export class Spoke {
 
   repay(amount: bigint, who: User) {
     const user = this.getUser(who);
-    const temp = this.hub.toDebtAssets(user.ghostDrawnShares, Rounding.CEIL) - user.offset;
```
yan-man, line 298: if this temp var is kept, should this be moved to after the hub accrue? to use the updated debt exch rate

tests/ts/prototype/core.ts
```diff
@@ -311,7 +314,8 @@ export class Spoke {
 
     user.ghostDrawnShares = percentMul(user.baseDrawnShares, user.riskPremium);
     user.offset = this.hub.toDebtAssets(user.ghostDrawnShares);
-    user.unrealisedPremium = oldUserUnrealisedPremium + temp - premiumDebtRestored;
+    user.unrealisedPremium +=
+      this.hub.toDebtAssets(oldUserGhostDrawnShares, Rounding.CEIL) - oldUserOffset;
```
yan-man: basically the same thing but without the temp var. but this uses the debt exch rate post-hub-accrual

but it does make sense that premiumDebtRestored should be factored in here in some way. Otherwise with my fixes here, the full repay case is broken: https://github.com/aave/aave-v4/blob/tmp/repay-bug/tests/ts/prototype/scenario.t.ts#L115

tests/ts/prototype/scenario.t.ts
```diff
@@ -79,20 +79,34 @@ assignSpokesToUsers();
...
 skip();
 
-const amount3 = p('1000');
-bob.supply(amount3);
-bob.borrow(amount3);
-bob.repay(amount3);
+// alice.log(true, true);
+
+alice.repay(amount1);
+alice.log(true, true);
```
yan-man, line 90: but **without** my changes, this simple case underflows under certain scenarios 

edit: it is resolved when incorporating premium debt restored

tests/ts/prototype/core.ts
DhairyaSethi: nope it needs the last exchange rate so has to be before hub.restore, gonna clean up & move to deduct from premium

tests/ts/prototype/core.ts
DhairyaSethi: this way you never realise the unrealised premium, add a skip in between and you'll see it's lost

tests/ts/prototype/core.ts
yan-man: but if you move to deduct from premium, that'd be after hub.accrue

I'm not saying moved to after hub.restore, but after **hub.accrue**

yan-man:
Resolved with latest changes to the rft/debt-accounting branch

PR #277, opened 2025-03-24
Refactor Protocol to New Accounting

DhairyaSethi:
- [x] structs 
- [x] asset data logic
- [x] hub methods
- [x] hub getters
- [x] spoke data logic
- [x] spoke external methods
- [x] spoke internal methods
- [x] spoke getters
- [x] basic flow tests
- [x] move tests back to `tests` and add `vm.skip()`

maybe's in this PR
- move logic libs to seperate dir
- naming
  -  `premiumDebt` to `premiumDrawnAssets`?
  - `suppliedShares` to `supplyShares`? better while referring in natspec for supply, withdraw; inconsistent with `drawnShares` naming
- refactor events to emit shares instead of assets 
- ~spoke.notify~
- ~fix all validations~

future PR(s)
- explicit rounding directions on getters
- remove reserve share data on spoke
- use in memory cache - ESSENTIAL -
- virtual offset for inflation attack - ESSENTIAL -
- custom types of debt/supply shares?
- refactor to stateless libs
- new actions accepting share inputs?
- fix rounding directions (if needed)

notes
- tokenize collateral assets so hub can pull/push with unified interface?
- user rp trailing?
- assets(premiumShares) - offset should never be < 0 ==> validate?
- allow hub to underflow on repay at spoke level only?
- make offset signed & add methods needed to facilitate ghost investment?

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -325,14 +325,13 @@ contract Spoke is ISpoke {
 
   function getReserveRiskPremium(uint256 reserveId) external view returns (uint256) {
     DataTypes.Reserve storage reserve = _reserves[reserveId];
-    return reserve.premiumDrawnShares.rayDiv(reserve.baseDrawnShares);
+    return reserve.premiumDrawnShares.rayDiv(reserve.baseDrawnShares); // trailing
   }
 
-  // todo user rp trailing?
-  //   function getUserRiskPremium(address user) external view returns (uint256) {
-  //     (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
-  //     return userRiskPremium.derayify();
-  //   }
+  function getUserRiskPremium(address user) external view returns (uint256) {
+    (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
+    return userRiskPremium;
+  }
 
   function getHealthFactor(address user) external view returns (uint256) {
     (, , uint256 healthFactor, , ) = _calculateUserAccountData(user);


CheyenneAtapour:
> premiumDebt to premiumDrawnAssets?

> suppliedShares to supplyShares? better while referring in natspec for supply, withdraw; inconsistent with drawnShares naming

At first glance, I think both these names are fine as is imo

src/contracts/LiquidityHub.sol
```diff
@@ -109,180 +106,153 @@ contract LiquidityHub is ILiquidityHub {
...
-      asset: asset,
-      spoke: spoke,
-      newSpokeRiskPremium: _boundBps(riskPremium).rayify(),
-      baseDebtAdded: 0,
-      baseDebtTaken: 0
-    });
 
-    uint256 sharesAmount = asset.convertToSharesUp(amount);
-    require(sharesAmount > 0, InvalidSharesAmount());
+    // todo name withdrawnShares, to align with spoke?
```
CheyenneAtapour: Yes, I think better to rename to `withdrawnShares`

src/contracts/LiquidityHub.sol
```diff
@@ -392,11 +380,11 @@ contract LiquidityHub is ILiquidityHub {
     require(!asset.config.paused, AssetPaused());
     require(!asset.config.frozen, AssetFrozen());
     require(assetsList[asset.id] != IERC20(address(0)), AssetNotListed());
-    require(
-      spoke.config.supplyCap == type(uint256).max ||
-        asset.convertToAssetsDown(spoke.suppliedShares) + amount <= spoke.config.supplyCap,
-      SupplyCapExceeded(spoke.config.supplyCap)
-    );
+    // require(
```
CheyenneAtapour: We'll need to add back the supply cap check for the spoke right?

src/contracts/LiquidityHub.sol
```diff
@@ -407,8 +395,8 @@ contract LiquidityHub is ILiquidityHub {
     require(amount > 0, InvalidWithdrawAmount());
     require(asset.config.active, AssetNotActive());
     require(!asset.config.paused, AssetPaused());
-    uint256 withdrawable = asset.convertToAssetsDown(spoke.suppliedShares);
-    require(amount <= withdrawable, SuppliedAmountExceeded(withdrawable));
+    // uint256 withdrawable = asset.convertToAssetsDown(spoke.suppliedShares);
```
CheyenneAtapour: We need to have validation that the spoke is not withdrawing more than they've supplied, right?

src/contracts/LiquidityHub.sol
```diff
@@ -433,113 +421,41 @@ contract LiquidityHub is ILiquidityHub {
...
-      newSpokeDebt
-    );
-
-    asset.baseDebt = newAssetDebt;
-    spoke.baseDebt = newSpokeDebt;
-
-    asset.riskPremium = newAssetRiskPremium;
-    spoke.riskPremium = newSpokeRiskPremium;
+    // uint256 maxAllowedRestore = spoke.baseDebt + spoke.outstandingPremium;
+    // require(amountRestored <= maxAllowedRestore, SurplusAmountRestored(maxAllowedRestore));
```
CheyenneAtapour: Did we decide if we want to allow a spoke to restore more? If so, we need the check, or to remove these comments otherwise

src/contracts/LiquidityHub.sol
DhairyaSethi: yes pls check description, will be adding, omitted intentionally rn

src/contracts/LiquidityHub.sol
DhairyaSethi: validations are todo rn mate
EDIT: validations are added. this is enforced rn but i think unneeded, we can rm to support excess repay, we will rm when needed

yan-man: hm i forget exactly where we landed on this. i think it was that we allow spoke to restore more if internal accounting for asset debt > sum of spoke debt, thus allowing DAO to call restore to true up the asset debt to 0 as needed (after all spokes repay). But not sure if this still applies to the new impl

src/contracts/AssetLogic.sol
```diff
@@ -15,128 +14,130 @@ library AssetLogic {
...
+  function toDrawnSharesDown(
     DataTypes.Asset storage asset,
     uint256 assets
-  ) external view returns (uint256) {
-    return assets.toSharesDown(asset.totalAssets(), asset.totalShares());
+  ) internal view returns (uint256) {
+    return assets.toSharesDown(asset.totalDrawnAssets(), asset.totalDrawnShares());
+  }
+
+  function premiumDebt(DataTypes.Asset storage asset) internal view returns (uint256) {
```
yan-man, line 56: i think kind of confusing that `asset` as param is used for type Asset but other places, `assets` is used to represent amount of assets as uint. but ig it's consistent with 4626 impl

tests/ts/prototype/core.ts
yan-man, line 1: perhaps rm these prototype files from here or close this PR: https://github.com/aave/aave-v4/pull/276/files, which version is the latest?

DhairyaSethi: i'd like to keep prototype in main for now, to experiment. we can rm later. i can merge that PR in rn, we can add your scenario tests (when fixed, if you want) on top or now

src/contracts/LiquidityHub.sol
```diff
@@ -301,79 +271,84 @@ contract LiquidityHub is ILiquidityHub {
...
+  }
+
+  function convertToDrawnShares(uint256 assetId, uint256 assets) external view returns (uint256) {
+    return _assets[assetId].toDrawnSharesUp(assets);
   }
 
-  function getInterestRate(uint256 assetId) public view returns (uint256) {
-    return _assets[assetId].getInterestRate();
+  function getBaseInterestRate(uint256 assetId) public view returns (uint256) {
+    return _assets[assetId].baseBorrowRate;
```
yan-man: should this be `asset.getBaseBorrowRate()`?

src/contracts/LiquidityHub.sol
```diff
@@ -431,115 +405,54 @@ contract LiquidityHub is ILiquidityHub {
...
-    require(config.irStrategy != address(0), InvalidIrStrategy());
+    require(address(config.irStrategy) != address(0), InvalidIrStrategy());
     require(config.decimals <= MAX_ALLOWED_ASSET_DECIMALS, InvalidAssetDecimals());
   }
+
+  function _getSpokeDebt(
+    DataTypes.Asset storage asset,
+    DataTypes.SpokeData storage spoke
+  ) internal view returns (uint256, uint256) {
+    uint256 premiumDebt = (asset.toDrawnAssetsUp(spoke.premiumDrawnShares) - spoke.premiumOffset) +
```
yan-man: is there a particular reason you wanted to specifically add brackets to the subtraction part prior to addition?

DhairyaSethi: yes, can you guess?

src/contracts/AssetLogic.sol
```diff
@@ -15,128 +14,130 @@ library AssetLogic {
...
+    return
+      asset.toDrawnAssetsUp(asset.premiumDrawnShares) -
+      asset.premiumOffset +
+      asset.unrealisedPremium;
```
yan-man: for consistency (see below) we should either add brackets here or remove from hub

DhairyaSethi: they were already added here [b9f3bb1](https://github.com/aave/aave-v4/pull/277/commits/b9f3bb1631f04dc0427d4c6cadddff4cfa2455d4), our git UI is lagging ig

src/contracts/LiquidityHub.sol
yan-man: some underflow protection / gas opt

DhairyaSethi: i hoped you'd be curious to discover what :(

src/libraries/types/DataTypes.sol
```diff
@@ -1,44 +1,46 @@
...
+    uint256 suppliedShares;
+    uint256 availableLiquidity;
+    uint256 baseDrawnShares;
+    uint256 premiumDrawnShares;
+    uint256 premiumOffset; // todo make signed
+    uint256 unrealisedPremium;
+    uint256 baseDrawnAssets;
+    uint256 baseBorrowRate;
     uint256 lastUpdateTimestamp;
+    uint256 id; // todo remove
```
yan-man, line 29: how come we can remove asset id?

src/libraries/types/DataTypes.sol
```diff
@@ -1,44 +1,46 @@
...
     DataTypes.AssetConfig config;
   }
 
   struct SpokeConfig {
-    uint256 drawCap; // asset denominated
-    uint256 supplyCap; // asset denominated
+    uint256 drawCap;
```
yan-man, line 34: are these no longer asset-based? can we instead add comment that it is shares-based otherwise

src/libraries/types/DataTypes.sol
```diff
@@ -56,12 +58,11 @@ library DataTypes {
...
-    uint256 baseDebt;
-    uint256 outstandingPremium;
     uint256 suppliedShares;
-    uint256 baseBorrowIndex;
-    uint256 lastUpdateTimestamp;
-    uint256 riskPremium; // weighted average risk premium of all users with ray precision
+    uint256 baseDrawnShares;
+    uint256 premiumDrawnShares;
+    uint256 premiumOffset;
+    uint256 unrealisedPremium;
```
yan-man: haha another minor issue - are we deciding to go with US english or UK/EU english? ie unrealized/unrealised. Just want to know for consistency sake, and can add to conventions doc

src/libraries/types/DataTypes.sol
DhairyaSethi: bc it's referenced with its id (see mapping)

src/libraries/types/DataTypes.sol
DhairyaSethi: they are still asset based, pls see impl, comment is cleanup

src/libraries/types/DataTypes.sol
DhairyaSethi: dw we're most likely renaming this var before this PR gets merged in 
let's discuss point 3 https://avaraverse.slack.com/archives/C04HCBMEEBZ/p1742843502330159 on slack

src/contracts/LiquidityHub.sol
yan-man: haha sorry, yes please enlighten me

src/contracts/LiquidityHub.sol
```diff
@@ -431,115 +405,54 @@ contract LiquidityHub is ILiquidityHub {
...
+      spoke.unrealisedPremium;
+    return (asset.toDrawnAssetsUp(spoke.baseDrawnShares), premiumDebt);
+  }
+
+  // handles underflow
+  function _add(uint256 a, int256 b) internal pure returns (uint256) {
```
yan-man, line 444: any reason you added this to both LH/spoke rather than putting this and _sub into math utils lib?

src/contracts/LiquidityHub.sol
CheyenneAtapour: Sounds good, but I don't see it on description. Can you comment the bullet point mentioning this topic, or double-check description?

DhairyaSethi: it's already been added mate

src/contracts/LiquidityHub.sol
DhairyaSethi: yeah these are fine here for now, want to make it cleaner eventually, and we can move to a math lib then

src/contracts/Spoke.sol
```diff
@@ -119,62 +111,36 @@ contract Spoke is ISpoke {
...
-      baseDebtTaken: 0
-    });
     uint256 suppliedShares = liquidityHub.supply(
       reserve.assetId,
       amount,
-      uint32(newReserveRiskPremium.derayify()),
       msg.sender // supplier
     );
-    _notifyRiskPremiumUpdate(reserve.assetId, msg.sender, newUserRiskPremium);
+    // todo _notifyRiskPremiumUpdate(reserve.assetId, msg.sender, newUserRiskPremium);
```
CheyenneAtapour: Will this function also recalculate user RP? Ofc we all know we need to do it, but also don't see a todo comment or the PR description mentioning this

DhairyaSethi: we'll use the calcAccountData method similar to how it's done on draw/repay. this doesn't need an additional comment imo

src/contracts/Spoke.sol
```diff
@@ -302,38 +294,26 @@ contract Spoke is ISpoke {
...
-    return cumulatedBaseDebt + cumulatedOutstandingPremium;
+  function getReserveTotalDebt(uint256 reserveId) external view returns (uint256) {
+    (uint256 baseDebt, uint256 premiumDebt) = _getReserveDebt(_reserves[reserveId]);
+    return baseDebt + premiumDebt;
   }
 
-  // todo by default returns only stored value, consider renaming to `getLast{Used,Stored}ReserveRiskPremium`
-  // to be inline with user's stored rp getter. we don't have an up to date rp concept here since that requires
-  // looping over all contributing users (ie one's drawing this reserve)
   function getReserveRiskPremium(uint256 reserveId) external view returns (uint256) {
```
CheyenneAtapour, line 326: I believe we can remove this now, right? Idt we have the concept of a risk premium for reserve anymore

DhairyaSethi: sure happy to rm now, but added as a temp stop gap for UI/debugging since we (kinda) do have the concept implicitly. we can rm later imo

src/contracts/Spoke.sol
```diff
@@ -210,36 +195,51 @@ contract Spoke is ISpoke {
     /// @dev TODO: onBehalfOf
```
CheyenneAtapour, line 217: Is the comment `Pass a value greater than the current debt to repay entire debt` meant to be a todo? Or is it referring to max uint256 case? If it's the latter, maybe we can clarify

DhairyaSethi: uint_max case is already implemented ser

src/contracts/Spoke.sol
```diff
@@ -422,75 +402,42 @@ contract Spoke is ISpoke {
...
-      reserve.outstandingPremium -= amount;
-    }
+  ) internal view returns (uint256, uint256) {
+    (uint256 baseDebt, uint256 premiumDebt) = _getUserDebt(userPosition, assetId);
 
-    return baseDebtRestored;
+    if (amount == type(uint256).max) {
+      return (baseDebt, premiumDebt);
+    }
+    if (amount < premiumDebt) {
```
CheyenneAtapour: Very small nit, but can change `<` to `<=` to save a subtraction in the case that `amount == premiumDebt`

src/contracts/Spoke.sol
```diff
@@ -671,27 +614,37 @@ contract Spoke is ISpoke {
...
+      reserve.premiumOffset) + reserve.unrealisedPremium;
+    return (liquidityHub.convertToDrawnAssets(assetId, reserve.baseDrawnShares), premiumDebt);
+  }
+
+  // todo optimize, merge logic duped borrow/repay, rename
   /**
    * @dev Trigger risk premium update on all drawn reserves of `user` except the reserve's corresponding
    * to `assetIdToAvoid` as those are expected to be updated outside of this method.
    * We only update risk premium for drawn assets and not supplied bc user RP does not contribute to
    * the other two RPs (Asset, Spoke/Reserve) as by definition they're based on drawn assets only.
```
CheyenneAtapour: Comment is a bit out of date because we don't have risk premiums anymore for asset and reserve, but rather just premium debt amounts. Don't have to adjust in this PR, but noting for future

src/contracts/Spoke.sol
```diff
@@ -210,36 +195,51 @@ contract Spoke is ISpoke {
...
+    int256 oldUserPremiumDrawnShares = int256(userPosition.premiumDrawnShares);
+    int256 oldUserPremiumOffset = int256(userPosition.premiumOffset);
+    int256 oldUserUnrealisedPremium = int256(userPosition.unrealisedPremium);
+
+    userPosition.premiumDrawnShares = userPosition.baseDrawnShares.percentMul(newUserRiskPremium);
+    userPosition.premiumOffset = liquidityHub.convertToDrawnAssets(
+      assetId,
+      userPosition.premiumDrawnShares
+    );
+    userPosition.unrealisedPremium += _sub(
```
CheyenneAtapour: Is this logic correct? It doesn't match the implementation in JS where this is assigned rather than `+=`

DhairyaSethi: it does, JS impl has a +oldUnrealsedPremium at the end which is exactly equivalent

src/contracts/LiquidityHub.sol
```diff
@@ -109,180 +106,153 @@ contract LiquidityHub is ILiquidityHub {
...
   }
 
   /// @inheritdoc ILiquidityHub
-  function withdraw(
-    uint256 assetId,
-    uint256 amount,
-    uint32 riskPremium,
-    address to
-  ) external returns (uint256) {
+  function withdraw(uint256 assetId, uint256 amount, address to) external returns (uint256) {
```
yan-man, line 139: maybe a question we can answer later, but do we want to do shares input here so that we can more easily do the withdraw max directly on LH?

DhairyaSethi: worth it only if we have shares input on spoke (imo we should support both on spoke like 4626)

it's less practical rn to have shares input here when spoke has asset bc all conversions happen on hub only

re max withdraw, we won't add max_uint stuff on hub. it'll always be spoke's responsibility bc hub actions are atomic (unlike spoke)

src/contracts/Spoke.sol
CheyenneAtapour: Yes, can we clarify the comment then to something like `Pass Max uint256 to repay entire debt`?

src/contracts/Spoke.sol
CheyenneAtapour: Just to make sure, we are talking about this line, right?

```user.unrealisedPremium = premiumDebt - premiumDebtRestored;```

File: src/contracts/Spoke.sol (hunk 15 above)
Comment (DhairyaSethi) [reply src/contracts/Spoke.sol:217]: that's understood & needs to be part of the broader natspec. this comment is whether we should round to current debt if currentDebt + x is passed. there is a use case for this w advanced intent fillers overestimating repayment

File: src/contracts/Spoke.sol (hunk 18 above)
Comment (DhairyaSethi) [reply src/contracts/Spoke.sol]: yeah it's equivalent if you expand the terms 
(premiumDebt includes oldUnrealisedPremium)

pls try to work it out & lmk if you disagree

File: src/contracts/Spoke.sol (hunk 15 above)
Comment (CheyenneAtapour) [reply src/contracts/Spoke.sol:217]: Ok, understood now. Would be nice to add a todo to the comment then to reduce the confusion here

File: src/contracts/LiquidityHub.sol (hunk 2 above)
Comment (CheyenneAtapour) [reply src/contracts/LiquidityHub.sol]: Amazing, thanks mate!

File: src/contracts/AssetLogic.sol [hunk 20]
```diff
@@ -15,128 +14,130 @@ library AssetLogic {
...
 
-  // @dev So solc doesn't inline
-  function getTotalAssets(DataTypes.Asset storage asset) external view returns (uint256) {
-    return asset.totalAssets();
+  // total drawn assets does not incl totalOutstandingPremium to accrue base rate separately
+  function toDrawnAssetsUp(
+    DataTypes.Asset storage asset,
+    uint256 shares
+  ) internal view returns (uint256) {
+    return shares.toAssetsUp(asset.totalDrawnAssets(), asset.totalDrawnShares());
```
miguelmtzinf, line 34: you prefer not to use `asset.baseDrawnShares` directly?

src/contracts/AssetLogic.sol
miguelmtzinf: we can move to `assetData` if confusing later
also, i d like not to use `assets` term as param in public functions

src/contracts/AssetLogic.sol
DhairyaSethi: it won't be accurate & up to date for the preview getters which re-use this method

miguelmtzinf: disagree, shares are static so do not see how they won't be accurate.
anyway, this approach is used across the file as a way to avoid footguns ig

src/contracts/AssetLogic.sol
```diff
@@ -15,128 +14,130 @@ library AssetLogic {
...
-      uint40(lastUpdateTimestamp)
+  ) internal {
+    uint256 baseBorrowRate = asset.config.irStrategy.calculateInterestRates(
+      DataTypes.CalculateInterestRatesParams({
+        liquidityAdded: liquidityAdded,
+        liquidityTaken: liquidityTaken,
+        totalDebt: asset.baseDrawnAssets,
+        reserveFactor: 0, // TODO
+        assetId: asset.id,
+        // todo + signedRealizedPremium
```
miguelmtzinf: do not think this is needed

src/contracts/AssetLogic.sol
```diff
@@ -15,128 +14,130 @@ library AssetLogic {
...
+  function toSuppliedSharesDown(
+    DataTypes.Asset storage asset,
+    uint256 assets
+  ) internal view returns (uint256) {
+    return assets.toSharesDown(asset.totalSuppliedAssets(), asset.totalSuppliedShares());
   }
 
+  // risk premium interest rate is calculated offchain
+  function baseInterestRate(DataTypes.Asset storage asset) internal view returns (uint256) {
+    return asset.baseBorrowRate;
```
miguelmtzinf, line 102: renaming to `borrowRate` would be fine, no? unless we wanna keep using `base` term so it's clear it's separate from `premium`

src/contracts/AssetLogic.sol
DhairyaSethi: ah i read that as baseDrawnAssets lol 

yeah it's consistency. and compiler will inline the internal method & optimise anyway. i prefer this

src/contracts/LiquidityHub.sol
```diff
@@ -109,180 +106,153 @@ contract LiquidityHub is ILiquidityHub {
...
-      spoke: spoke,
-      newSpokeRiskPremium: _boundBps(riskPremium).rayify(),
-      baseDebtAdded: 0,
-      baseDebtTaken: 0
-    });
 
     // todo: Mitigate inflation attack (burn some amount if first supply)
-    uint256 sharesAmount = asset.convertToSharesDown(amount);
-    require(sharesAmount > 0, InvalidSharesAmount());
+    uint256 suppliedShares = asset.toSuppliedSharesDown(amount);
```
miguelmtzinf, line 122: i d use `sharesAmount` name in all functions, to refer the converted amount in shares.

DhairyaSethi: it already has a sharesSuffix rn so it's clear it's a share not asset imo. do you mean supplied and drawn share instead? bc i had that before but chose this to be more context aware

src/contracts/AssetLogic.sol
DhairyaSethi: prefer verbosity rn tbh

src/contracts/AssetLogic.sol
DhairyaSethi: it will be needed if we add -ve offset idea, we can rm this comment if scrap that

src/contracts/LiquidityHub.sol
```diff
@@ -301,79 +271,74 @@ contract LiquidityHub is ILiquidityHub {
     return _spokes[assetId][spoke].config;
   }
 
-  function getTotalAssets(uint256 assetId) external view returns (uint256) {
-    return _assets[assetId].getTotalAssets();
+  // todo 4626 getter naming
```
miguelmtzinf, line 274: what do you have in mind?

src/contracts/LiquidityHub.sol
```diff
@@ -387,14 +352,13 @@ contract LiquidityHub is ILiquidityHub {
     DataTypes.SpokeData storage spoke,
     uint256 amount
   ) internal view {
-    require(amount > 0, InvalidSupplyAmount());
```
miguelmtzinf, line 390: you removed this in favor of `sharesAmount > 0`? not worth to fail earlier?

src/contracts/LiquidityHub.sol
```diff
@@ -431,115 +395,54 @@ contract LiquidityHub is ILiquidityHub {
...
-      newSpokeRiskPremium, // use new spoke risk premium
-      newSpokeDebt
-    );
-
-    asset.baseDebt = newAssetDebt;
-    spoke.baseDebt = newSpokeDebt;
-
-    asset.riskPremium = newAssetRiskPremium;
-    spoke.riskPremium = newSpokeRiskPremium;
+    // todo: rm this check?
```
miguelmtzinf: why?

src/contracts/LiquidityHub.sol
DhairyaSethi: we still fail early bc we check suppliedShares amount is > 0, so this is redundant here imo

src/contracts/LiquidityHub.sol
miguelmtzinf: @DhairyaSethi troll

src/contracts/LiquidityHub.sol
DhairyaSethi: whether want the hub to accept over repayment. will discuss more in the sync today

src/contracts/LiquidityHub.sol
miguelmtzinf: i d suggest to flip it actually. Better readability imho
`existingStuff + newStuff`

src/contracts/Spoke.sol
miguelmtzinf: worth to move it down after the accounting update imo

DhairyaSethi: we'll likely not have this here and will instead expect UI to multicall refresh & notify

src/contracts/AssetLogic.sol
CheyenneAtapour: We only have 1 borrow rate now, so yes would make sense to rename eventually at least

src/contracts/LiquidityHub.sol
CheyenneAtapour: I also agree with Dhairya that `suppliedShares` (and similarly `withdrawnShares`, `drawnShares`, etc.) more accurately conveys the context and is clear that it is shares. I like this name as is personally

src/contracts/LiquidityHub.sol
CheyenneAtapour: We basically decided against the hub accepting over repayment, right?

DhairyaSethi: yep looks like it. will rm all the comments on this

src/contracts/LiquidityHub.sol
```diff
@@ -404,10 +368,10 @@ contract LiquidityHub is ILiquidityHub {
     DataTypes.SpokeData storage spoke,
     uint256 amount
   ) internal view {
-    require(amount > 0, InvalidWithdrawAmount());
+    require(amount != 0, InvalidWithdrawAmount());
     require(asset.config.active, AssetNotActive());
     require(!asset.config.paused, AssetPaused());
-    uint256 withdrawable = asset.convertToAssetsDown(spoke.suppliedShares);
+    uint256 withdrawable = asset.toSuppliedAssetsUp(spoke.suppliedShares);
```
CheyenneAtapour: ERC4626 for max withdraw rounds down. Is there a reason we changed here to round up? https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol#L142

src/contracts/LiquidityHub.sol
```diff
@@ -109,180 +106,153 @@ contract LiquidityHub is ILiquidityHub {
...
-    uint256 baseDebtRestored = _deductFromOutstandingPremium(asset, spoke, amount);
-    _updateRiskPremiumAndBaseDebt({
-      asset: asset,
-      spoke: spoke,
-      newSpokeRiskPremium: _boundBps(riskPremium).rayify(),
-      baseDebtAdded: 0,
-      baseDebtTaken: baseDebtRestored
-    });
+    _validateRestore(asset, spoke, baseAmount, premiumAmount);
+    asset.updateBorrowRate({liquidityAdded: baseAmount, liquidityTaken: 0}); // both can be zero
```
CheyenneAtapour, line 207: Why is `liquidityAdded` just the `baseAmount`?

src/contracts/AssetLogic.sol
CheyenneAtapour: Also, should we have a getter that returns (perhaps by previewing) the most up-to-date rate? E.g. would be after accrual

src/contracts/Spoke.sol
```diff
@@ -700,74 +671,45 @@ contract Spoke is ISpoke {
...
-
-    // @dev no need to update `reserve.baseDebt` & `userPosition.baseDebt` as there is no debt change
-    reserve.riskPremium = newReserveRiskPremium;
-
-    return newReserveRiskPremium;
-  }
-
-  function _validateHealthFactor(address userAddress) internal view {
-    (, , uint256 healthFactor, , ) = _calculateUserAccountData(userAddress);
+  function _validateUserPosition(address userAddress) internal view returns (uint256) {
```
CheyenneAtapour, line 708: I wonder if there is a better name for this because it does return `userRiskPremium`, and doesn't just validate like the other validation functions. But ofc we can decide to rename this later on

src/contracts/AssetLogic.sol
DhairyaSethi: new rate depends on utilisation: the debt, available liq & liq added + taken. getter is not needed, one can use the IR strategy for new rate if curious

src/contracts/LiquidityHub.sol
DhairyaSethi: bc only that it relevant for the base borrow rate, similar to how debt passed is also baseDrawn. otherwise we compound on base rate inherently too

src/contracts/AssetLogic.sol
miguelmtzinf: imo better to remove and create an issue to track that idea

src/contracts/LiquidityHub.sol
miguelmtzinf: @DhairyaSethi

DhairyaSethi: basically 4626 compatibility such that we can run existing vault test suites like https://github.com/a16z/erc4626-tests
so accurate getter names (can be overloaded, here or in tests)

src/contracts/Spoke.sol
DhairyaSethi: yeah it's intentional and i'm fine with this for now. validateAndReturn is too long, thought it could return full userAccountData instead of just rp but it won't used to it's fine. we're bound to validate more stuff here, can come up with a name then

src/contracts/LiquidityHub.sol
```diff
@@ -109,180 +106,153 @@ contract LiquidityHub is ILiquidityHub {
...
 
-    asset.availableLiquidity += amount;
+    uint256 totalRestoredAmount = baseAmount + premiumAmount;
+    uint256 baseDrawnSharesRestored = asset.toDrawnSharesUp(baseAmount);
```
miguelmtzinf, line 210: this should round down, not up

src/contracts/LiquidityHub.sol
DhairyaSethi: yeah bc we round up in withdraw normally while converting to shares, i _think_ OZ rounds down here just to prevent overflow? but could be wrong

but will round down here and we can fix later if needed

src/contracts/LiquidityHub.sol
miguelmtzinf: feels a bit opinionated and given that we have the same check in the rest of validation functions, it makes sense to me to keep it

src/contracts/LiquidityHub.sol
DhairyaSethi: that leads to dust shares remaining (not just amounts) when repaying full bc we round up on draw

src/contracts/LiquidityHub.sol
DhairyaSethi: sure will bring back, we can opt later when needed. but it's duplicated bc when supply amount is zero, suppliedShares will be zero, and if dust it's again caught if supplied shares is zero

it's needed on withdraw & everywhere else bc there we round up, so shares wouldn't round to zero unless amount is zero and we don't check for zero shares there

src/contracts/LiquidityHub.sol
miguelmtzinf: oh yeah, we should take as much shares as we can indeed

Issue #296, opened 2025-03-26
LH Config tests refactor


yan-man:
LiquidityHub.Config.t.sol

PR #298, opened 2025-03-26
LH supply/config test migration

yan-man:
- migrate LH supply/config tests

closes https://github.com/aave/aave-v4/issues/278
closes https://github.com/aave/aave-v4/issues/296

### DIFF
--- a/tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
@@ -193,7 +193,7 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
     assertEq(hub.getAssetConfig(daiAssetId).decimals, newDecimals, 'asset decimals');
   }
 
-  function test_updateAssetConfig_decimals_revertsWith_InvalidAssetDecimals(
+  function test_updateAssetConfig_fuzz_decimals_revertsWith_InvalidAssetDecimals(
     uint256 assetId
   ) public {
     assetId = bound(assetId, 0, hub.assetCount() - 1);
@@ -209,7 +209,7 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
     hub.updateAssetConfig(assetId, config);
   }
 
-  function test_updateAssetConfig_fuzz_decimals_revertsWith_InvalidAssetDecimals() public {
+  function test_updateAssetConfig_decimals_revertsWith_InvalidAssetDecimals() public {
     DataTypes.AssetConfig memory config = hub.getAssetConfig(daiAssetId);
     uint256 newDecimals = hub.MAX_ALLOWED_ASSET_DECIMALS() + 1; // invalid decimals
     assertNotEq(config.decimals, newDecimals);
@@ -430,7 +430,7 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
   }
 
   function test_addAsset_revertsWith_InvalidAssetAddress() public {
-    uint256 decimals = hub.MAX_ALLOWED_ASSET_DECIMALS() - 1;
+    uint256 decimals = hub.MAX_ALLOWED_ASSET_DECIMALS();
 
     vm.expectRevert(ILiquidityHub.InvalidAssetAddress.selector);
     vm.prank(HUB_ADMIN);
@@ -470,7 +470,7 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
   }
 
   function test_addAsset_revertsWith_InvalidIrStrategy() public {
-    uint256 decimals = hub.MAX_ALLOWED_ASSET_DECIMALS() - 1;
+    uint256 decimals = hub.MAX_ALLOWED_ASSET_DECIMALS();
 
     vm.expectRevert(ILiquidityHub.InvalidIrStrategy.selector);
     vm.prank(HUB_ADMIN);


tests/unit/LiquidityHub/LiquidityHubBase.t.sol
```diff
@@ -113,4 +145,29 @@ contract LiquidityHubBase is Base {
...
+  // triggers refresh to cache premium debt
+  function _createPremiumDebt(ISpoke spoke, uint256 daiAmount) internal returns (uint256) {
+    uint256 daiReserveId = _daiReserveId(spoke);
+    Utils.spokeSupply({
+      spoke: spoke,
+      reserveId: daiReserveId,
+      user: alice,
+      amount: daiAmount,
+      onBehalfOf: alice
+    });
```
DhairyaSethi, line 161: the hub test helpers using spoke actions to accrue premium feels weird tbh. i get why though, figuring out the correct risk premium debt values it itself prone to errors
we could just do hub tests w/o risk premium and test risk premium stuff on spoke, and have a seperate hub.refresh test file. this also works for me tho

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -284,596 +257,430 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+  }
+
+  function test_supply_fuzz_single_spoke_multi_supply(
+    uint256 amount,
+    uint256 skipTime,
+    uint256 rate
+  ) public {
+    uint256 assetId = daiAssetId;
+    uint256 numSupplies = 5;
+    amount = bound(amount, hub.convertToDrawnAssets(assetId, 1), MAX_SUPPLY_AMOUNT / numSupplies);
```
DhairyaSethi: hub.convertToDrawnAssets(assetId, 1) should be a static value since all these tests have the same init state
basically min asset value such shares doesn't round down (will be round up for withdraw) to 0, what you've written
https://github.com/a16z/erc4626-tests/blob/main/ERC4626.test.sol#L111 we don't support 0 value unfortunately

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -284,596 +257,430 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+    params.spoke2SuppliedAmount = hub.convertToSuppliedAssets(assetId, params.spoke2SuppliedShares);
+    params.aliceBalance = MAX_SUPPLY_AMOUNT + params.drawnAmount;
+    params.bobBalance = MAX_SUPPLY_AMOUNT - amount;
+
+    for (uint256 i = 0; i < numSupplies; i++) {
+      uint256 supplyShares = 1; // minimum for 1 share
+      uint256 supplyAmount = hub.convertToSuppliedAssets(assetId, supplyShares);
+      supplyAmount = hub.convertToSuppliedShares(assetId, supplyAmount) < 1
+        ? supplyAmount + 1
+        : supplyAmount; // account for rounding down on assets
```
DhairyaSethi, line 605: `supplyAmount` is storing the return value of `shares`? typo most likely?

yan-man: no supplyAmount is just amount, first I am finding the min asset amount corresponding to 1 share in the line above. But due to rounding down, at times that asset amount doesn't convert back into 1 share, so I am checking this case and adding 1 to make sure that asset amount at least converts into 1 share.

tests/unit/LiquidityHub/LiquidityHubBase.t.sol
yan-man: > we could just do hub tests w/o risk premium and test risk premium stuff on spoke

yeah was considering this but feels like more complete to have it here too. I think we can also do hub.refresh test file and in that case perhaps we can rm this. Prefer to leave it for now.

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
yan-man: gotcha, will just bound to 1

edit: https://github.com/aave/aave-v4/pull/298/commits/2d814abc5e18b9afbd0259e272ac43718fe17980

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
DhairyaSethi: ah yeah makes sense, i didn't take the ternary into account lol

yan-man: np thanks for the deep review

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
```diff
@@ -1,369 +1,512 @@
...
-  
-}
+    vm.expectRevert(ILiquidityHub.InvalidAssetDecimals.selector);
+    vm.prank(HUB_ADMIN);
+    hub.updateAssetConfig(assetId, config);
+  }
 
-  function test_updateAssetConfig_decimals() public {
-    vm.skip(true, 'pending refactor');
+  function test_updateAssetConfig_fuzz_decimals_revertsWith_InvalidAssetDecimals() public {
```
CheyenneAtapour: I think you meant to name the one above as the `fuzz` test, because this one doesn't fuzz anything, but the one above fuzzes `assetId`

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
```diff
@@ -1,369 +1,512 @@
...
-//         active: true,
-//         frozen: false,
-//         paused: false,
-//         irStrategy: address(irStrategy)
-//       }),
-//       address(0)
-//     );
-  
-}
+    uint256 decimals = hub.MAX_ALLOWED_ASSET_DECIMALS() - 1;
```
CheyenneAtapour: The smallest nit ever, but you don't need to do this `- 1` here and similar places, because `MAX_ALLOWED_DECIMALS()` itself is a valid value. I think it would be cleaner personally and help the reader focus more so on the point of the test

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
yan-man: yeah flipped the names, good catch! 

https://github.com/aave/aave-v4/pull/298/commits/f913ca906176da273923390f2bc45314d840eb16

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
yan-man: https://github.com/aave/aave-v4/pull/298/commits/f913ca906176da273923390f2bc45314d840eb16

PR #299, opened 2025-03-26
feat: share events

### LINKED ISSUE #268: Rename Functions
Spoke (user facing functions):
- `borrow` and `repay` to be kept
- `supply` and `withdraw` to be kept

LH (spoke facing functions):
- `supply` -> `add`
- `withdraw` -> `remove`
- `borrow` -> `draw`
- `repay` -> `settle`

Risk Premium:
- for liquidity premium -> `collateralRisk`
- for risk premium -> `riskPremium`

Algo, we are a bit more flexible:
- for rp base debt shares -> `baseDebtShares` or `debtShares`
- for rp debt shares -> `premiumDebtShares`
- for offset -> `premiumDebtOffset`

DhairyaSethi:
closes #268 
changes
- all events in present tense
- new event with premium deltas for the refresh method
- liquidityHub->hub on spoke
- hub.supply/withdraw -> hub.add/remove

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -157,7 +157,7 @@ contract Spoke is ISpoke {
     );
     _notifyRiskPremiumUpdate(assetId, msg.sender, newUserRiskPremium);
 
-    emit Withdraw(reserveId, msg.sender, to, withdrawnShares);
+    emit Withdraw(reserveId, msg.sender, withdrawnShares, to);
   }
 
   /// @inheritdoc ISpoke
@@ -195,7 +195,7 @@ contract Spoke is ISpoke {
     );
     _notifyRiskPremiumUpdate(assetId, msg.sender, newUserRiskPremium);
 
-    emit Borrow(reserveId, msg.sender, to, baseDrawnShares);
+    emit Borrow(reserveId, msg.sender, baseDrawnShares, to);
   }
 
   /// @inheritdoc ISpoke
--- a/src/interfaces/ISpoke.sol
+++ b/src/interfaces/ISpoke.sol
@@ -19,14 +19,14 @@ interface ISpoke {
   event Withdraw(
     uint256 indexed reserveId,
     address indexed user,
-    address indexed to,
-    uint256 suppliedShares
+    uint256 suppliedShares,
+    address indexed to
   );
   event Borrow(
     uint256 indexed reserveId,
     address indexed user,
-    address indexed to,
-    uint256 drawnShares
+    uint256 drawnShares,
+    address indexed to
   );
   event Repay(uint256 indexed reserveId, address indexed user, uint256 drawnShares);
   event UsingAsCollateral(uint256 indexed reserveId, address indexed user, bool usingAsCollateral);


yan-man:
pls update description and link an issue

src/contracts/Spoke.sol
```diff
@@ -166,7 +162,7 @@ contract Spoke is ISpoke {
     );
     _notifyRiskPremiumUpdate(assetId, msg.sender, newUserRiskPremium);
 
-    emit Withdrawn(reserveId, msg.sender, amount);
+    emit Withdraw(reserveId, to, withdrawnShares);
```
yan-man: now the event emission is with the to instead of msg.sender, you intended to change this right

DhairyaSethi: ~yes yes, it was incorrect earlier. `to` is the address to whom the funds are pushed~
no actually msg.sender is the user position, sorry will revert

src/interfaces/ILiquidityHub.sol
```diff
@@ -20,15 +20,17 @@ interface ILiquidityHub {
     uint256 supplyCap
   );
 
-  event Supply(uint256 indexed assetId, address indexed spoke, uint256 amount);
-  event Withdraw(
+  event Add(uint256 indexed assetId, address indexed spoke, uint256 suppliedShares);
+  event Remove(uint256 indexed assetId, address indexed spoke, uint256 suppliedShares);
+  event Draw(uint256 indexed assetId, address indexed spoke, uint256 drawnShares);
+  event Restore(uint256 indexed assetId, address indexed spoke, uint256 drawnShares);
```
CheyenneAtapour, line 26: This should be `Settle` now :)

src/contracts/LiquidityHub.sol
```diff
@@ -128,15 +128,15 @@ contract LiquidityHub is ILiquidityHub {
     spoke.suppliedShares += suppliedShares;
 
     // TODO: fee-on-transfer
-    assetsList[assetId].safeTransferFrom(supplier, address(this), amount);
+    assetsList[assetId].safeTransferFrom(from, address(this), amount);
```
CheyenneAtapour, line 131: The rationale to rename `supplier` to `from` is because this function is now `add` instead of `supply`, right? But we still have `suppliedShares` in the event definition, and also within this function, and thus the event emission. 

Therefore, I'd personally be in favor of keeping `supplier` over changing to `from`

I'm open to discussion on this. There's an argument for either way

src/interfaces/ILiquidityHub.sol
```diff
@@ -66,27 +68,27 @@ interface ILiquidityHub {
   ) external;
 
   /**
-   * @notice Supply asset on behalf of user.
+   * @notice Add/Supply asset on behalf of user.
    * @dev Only callable by spokes.
    * @param assetId The asset id.
-   * @param amount The amount of asset to supply.
+   * @param amount The amount of asset liquidity to add or supply.
```
CheyenneAtapour: I don't like that we have to clarify this in the nat spec like this. But if we must, I think it's better to write `add/supply` like you do above in the `@notice` instead of `add or supply` because the way this is written makes it seem like there are 2 separate cases that can happen

src/contracts/LiquidityHub.sol
DhairyaSethi: no the argument is @miguelmtzinf didn't like the `supplier` term regardless, even before, hence `from` also makes sense

src/interfaces/ILiquidityHub.sol
DhairyaSethi: image

src/interfaces/ILiquidityHub.sol
DhairyaSethi: [f5d5eb2](https://github.com/aave/aave-v4/pull/299/commits/f5d5eb23a1ac4f5440938d2a182f0b3c9819a986)

src/contracts/LiquidityHub.sol
miguelmtzinf: open for alternatives @CheyenneAtapour , but supplier/repayer do not sound good right

src/contracts/Spoke.sol
```diff
@@ -166,7 +157,7 @@ contract Spoke is ISpoke {
     );
     _notifyRiskPremiumUpdate(assetId, msg.sender, newUserRiskPremium);
 
-    emit Withdrawn(reserveId, msg.sender, amount);
+    emit Withdraw(reserveId, msg.sender, to, withdrawnShares);
```
miguelmtzinf: i think better to have amount as 3rd value always

src/contracts/Spoke.sol
```diff
@@ -209,7 +195,7 @@ contract Spoke is ISpoke {
     );
     _notifyRiskPremiumUpdate(assetId, msg.sender, newUserRiskPremium);
 
-    emit Borrowed(reserveId, to, amount);
+    emit Borrow(reserveId, msg.sender, to, baseDrawnShares);
```
miguelmtzinf: same, for consistency

src/interfaces/ILiquidityHub.sol
```diff
@@ -66,27 +68,27 @@ interface ILiquidityHub {
   ) external;
 
   /**
-   * @notice Supply asset on behalf of user.
+   * @notice Add/Supply asset on behalf of user.
```
miguelmtzinf, line 71: i d get rid of supply completely, but nbd now

src/contracts/Spoke.sol
DhairyaSethi: [df1d3bc](https://github.com/aave/aave-v4/pull/299/commits/df1d3bc65730bd2bf658c179a32ae4d535340757)

src/contracts/Spoke.sol
DhairyaSethi: [df1d3bc](https://github.com/aave/aave-v4/pull/299/commits/df1d3bc65730bd2bf658c179a32ae4d535340757)

src/interfaces/ILiquidityHub.sol
DhairyaSethi: imo it's useful inference that this is usually called within a spoke's supply operation

CheyenneAtapour: Yeah, I agree with @DhairyaSethi's point because e.g. an integrating dev will examine the flow from Spoke calling supply, then calling into this function, so it's helpful to maintain a notion that it's the LH handling the supply call

src/contracts/LiquidityHub.sol
CheyenneAtapour: It's tricky here, yeah because now the function is `add` instead of `supply`, but a e.g. an integrating dev will trace the flow of `supply` being called on the spoke, and it is helpful to have a notion that this is the hub handling the `supply` call, but on the other hand, indeed it looks weird in a vacuum that this is an `add` function with a notion of `supply`.

We can just proceed and revisit these names later on

DhairyaSethi: yeah mate i disagree slightly but let's please focus on shipping the critical flow for now 🚀

PR #303, opened 2025-03-27
fix: correctly settle premium debt on repay

DhairyaSethi:
fixes repay flow on spoke by fully realising the premium debt using an additional `refresh` before moving forward

### DIFF
--- a/tests/unit/Spoke/Spoke.Repay.t.sol
+++ b/tests/unit/Spoke/Spoke.Repay.t.sol
@@ -58,9 +58,7 @@ contract SpokeRepayTest is SpokeBase {
     _assertUserRpUnchanged(_daiReserveId(spoke1), spoke1, bob);
     _assertUserRpUnchanged(_usdxReserveId(spoke1), spoke1, bob);
 
-    skip(123 days);
-
-    hub.getSpokeDebt(daiAssetId, address(spoke1));
+    skip(232 days);
 
     spoke1.repay(_daiReserveId(spoke1), 25e18);
     _assertUserRpUnchanged(_daiReserveId(spoke1), spoke1, bob);


src/contracts/Spoke.sol
```diff
@@ -212,39 +212,45 @@ contract Spoke is ISpoke {
...
+      _signedDiff(userPosition.premiumOffset, userPremiumOffset),
+      _signedDiff(userPosition.realizedPremium, userRealizedPremium)
+    ); // we settle premium debt here
+    uint256 restoredShares = hub.restore(
       reserve.assetId,
       baseDebtRestored,
       premiumDebtRestored,
       msg.sender
-    );
+    ); // we settle base debt here
```
DhairyaSethi, line 234: pls note we have to pass `premiumDebtRestored` here even though it was taken into account in the premium accounting with the refresh below bc we have to account for the addition in available liquidity for the supply ex change

src/contracts/Spoke.sol
```diff
@@ -212,39 +212,45 @@ contract Spoke is ISpoke {
...
 
-    reserve.baseDrawnShares -= baseRestoredShares;
-    userPosition.baseDrawnShares -= baseRestoredShares;
+    reserve.baseDrawnShares -= restoredShares;
+    userPosition.baseDrawnShares -= restoredShares;
 
-    // calc needs new user position, just updating base debt is fine so far
     (uint256 newUserRiskPremium, , , , ) = _calculateUserAccountData(msg.sender);
```
CheyenneAtapour, line 239: I thought we would only do this recalc of user risk premium in the multi-call no? Ig it's needed here to update the `premiumDrawnShares`?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -6,804 +6,861 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
 contract SpokeRepayTest is SpokeBase {
   using PercentageMath for uint256;
 
+  function test_repay_all_with_accruals() public {
```
CheyenneAtapour, line 9: Thank you for adding this test case in :)

src/contracts/Spoke.sol
DhairyaSethi: yeah after we have multicall we'll remove rp update and the second refresh from here 
plan to start writing exhaustive repay/borrow tests meanwhile

CheyenneAtapour: Amazing, you're quicker than me haha - see approval comment

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -6,804 +6,861 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
...
+    _assertUserRpUnchanged(_daiReserveId(spoke1), spoke1, bob);
+    _assertUserRpUnchanged(_usdxReserveId(spoke1), spoke1, bob);
+
+    spoke1.withdraw(_usdxReserveId(spoke1), 0.01e6, bob);
+    _assertUserRpUnchanged(_daiReserveId(spoke1), spoke1, bob);
+    _assertUserRpUnchanged(_usdxReserveId(spoke1), spoke1, bob);
+
+    skip(123 days);
+
+    hub.getSpokeDebt(daiAssetId, address(spoke1));
```
yan-man: can rm this

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -6,804 +6,861 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
...
+    spoke1.repay(_daiReserveId(spoke1), 25e18);
+    _assertUserRpUnchanged(_daiReserveId(spoke1), spoke1, bob);
+    _assertUserRpUnchanged(_usdxReserveId(spoke1), spoke1, bob);
+
+    vm.stopPrank();
+  }
+
+  function _assertUserRpUnchanged(uint256 reserveId, ISpoke spoke, address user) internal {
+    DataTypes.UserPosition memory pos = spoke.getUserPosition(reserveId, user);
+    uint256 riskPremiumStored = pos.premiumDrawnShares.percentDiv(pos.baseDrawnShares);
```
yan-man, line 72: prob doesnt need to settled in this PR, but should we then also create a getter for stored user RP now that we're not putting this explicitly in storage, as getUserAccountData would return accrued user RP which wouldn't really be accurately used anywhere unless an action happens. 

or replace the current getUserRiskPremium getter to return this rather than the _calculateUserAccountData output

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -6,804 +6,859 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
...
+
+    uint256 supplyAmount = 5000e18;
+    spoke1.supply(_daiReserveId(spoke1), supplyAmount);
+    spoke1.setUsingAsCollateral(_daiReserveId(spoke1), true);
+
+    uint256 borrowAmount = 1000e18;
+    spoke1.borrow(_daiReserveId(spoke1), borrowAmount, bob);
+
+    skip(365 days);
+    spoke1.getUserDebt(_daiReserveId(spoke1), bob);
```
yan-man, line 20: can rm

tests/unit/Spoke/Spoke.Repay.t.sol
DhairyaSethi: > getUserRiskPremium getter to return this

yeah for which position tho? we'd need to loop over all to find one drawn position and then just return that or for all? it's not as straightforward that's why i didn't include in my refactor PR

> getUserAccountData would return accrued user RP which wouldn't really be accurately used anywhere

no it works here bc no time has passed and we immediately check for rp

yan-man: > works here bc no time has passed

yeah def agreed here bc no time has passed, it only diverges after time passes w accrual

PR #308, opened 2025-03-29
Add Max Withdraw

### LINKED ISSUE #309: Max Withdraw
User should be able to pass in max value of uint256 to spoke's `withdraw` function, and this should allow the user to withdraw all of their supplied shares for the corresponding asset

CheyenneAtapour:
Allows for a user to pass max value of uint256 to withdraw all their supplied assets

closes #309

### DIFF
--- a/tests/unit/Spoke/Spoke.Withdraw.t.sol
+++ b/tests/unit/Spoke/Spoke.Withdraw.t.sol
@@ -279,13 +279,8 @@ contract SpokeWithdrawTest is SpokeBase {
     );
 
     // Withdraw all supplied assets
-    Utils.spokeWithdraw({
-      spoke: spoke1,
-      reserveId: _daiReserveId(spoke1),
-      user: bob,
-      amount: type(uint256).max,
-      onBehalfOf: bob
-    });
+    vm.prank(bob);
+    spoke1.withdraw(_daiReserveId(spoke1), type(uint256).max, bob);
 
     _checkSuppliedAmounts(daiAssetId, _daiReserveId(spoke1), spoke1, bob, 0, 'after withdraw');
   }
@@ -310,13 +305,8 @@ contract SpokeWithdrawTest is SpokeBase {
     );
 
     // Withdraw all supplied assets
-    Utils.spokeWithdraw({
-      spoke: spoke1,
-      reserveId: _daiReserveId(spoke1),
-      user: bob,
-      amount: type(uint256).max,
-      onBehalfOf: bob
-    });
+    vm.prank(bob);
+    spoke1.withdraw(_daiReserveId(spoke1), type(uint256).max, bob);
 
     _checkSuppliedAmounts(daiAssetId, _daiReserveId(spoke1), spoke1, bob, 0, 'after withdraw');
   }
@@ -369,13 +359,8 @@ contract SpokeWithdrawTest is SpokeBase {
       amount: type(uint256).max
     });
 
-    Utils.spokeWithdraw({
-      spoke: spoke1,
-      reserveId: _daiReserveId(spoke1),
-      user: bob,
-      amount: type(uint256).max,
-      onBehalfOf: bob
-    });
+    vm.prank(bob);
+    spoke1.withdraw(_daiReserveId(spoke1), type(uint256).max, bob);
 
     _checkSuppliedAmounts(daiAssetId, _daiReserveId(spoke1), spoke1, bob, 0, 'after withdraw');
   }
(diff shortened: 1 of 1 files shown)


tests/Base.t.sol
```diff
@@ -702,4 +702,56 @@ abstract contract Base is Test {
...
+    ISpoke spoke,
+    address user,
+    uint256 expectedSuppliedAmount,
+    string memory when
+  ) internal {
+    uint256 expectedSuppliedShares = hub.convertToSuppliedShares(assetId, expectedSuppliedAmount);
+    assertEq(
+      hub.getAssetSuppliedShares(assetId),
+      expectedSuppliedShares,
+      string(abi.encodePacked('asset supplied shares ', when))
```
yan-man, line 719: nice, I like this structure

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -259,6 +259,126 @@ contract SpokeWithdrawTest is SpokeBase {
...
+    Utils.spokeBorrow({
+      spoke: spoke1,
+      reserveId: _daiReserveId(spoke1),
+      user: bob,
+      amount: borrowAmount,
+      onBehalfOf: bob
+    });
+
+    // Wait a year to accrue interest
+    skip(365 days);
```
yan-man: can you fuzz the skip time as well here

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -259,6 +259,126 @@ contract SpokeWithdrawTest is SpokeBase {
...
+      amount: type(uint256).max,
+      onBehalfOf: bob
+    });
+
+    _checkSuppliedAmounts(daiAssetId, _daiReserveId(spoke1), spoke1, bob, 0, 'after withdraw');
+  }
+
+  function test_withdraw_fuzz_all_with_interest(uint256 supplyAmount, uint256 borrowAmount) public {
+    supplyAmount = bound(supplyAmount, 1000, MAX_SUPPLY_AMOUNT);
+    borrowAmount = bound(borrowAmount, 500, supplyAmount / 2);
```
yan-man: are these bounded this way due to inflation offset / rounding?

tests/unit/Spoke/Spoke.Withdraw.t.sol
CheyenneAtapour: I added a separate test which fuzzes time skip

tests/unit/Spoke/Spoke.Withdraw.t.sol
CheyenneAtapour: It was to ensure interest grows, but now I just added a `vm.assume` statement to cover it

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -355,6 +355,70 @@ contract SpokeWithdrawTest is SpokeBase {
     // Wait a year to accrue interest
     skip(365 days);
 
+    vm.assume(hub.getAssetSuppliedAmount(daiAssetId) > supplyAmount);
+    assertGt(hub.getAssetSuppliedAmount(daiAssetId), supplyAmount, 'asset grows with interest');
```
yan-man: minor, don't think this assert is needed given the vm assume. Same below

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -259,6 +259,190 @@ contract SpokeWithdrawTest is SpokeBase {
...
+    Utils.spokeWithdraw({
+      spoke: spoke1,
+      reserveId: _daiReserveId(spoke1),
+      user: bob,
+      amount: type(uint256).max,
+      onBehalfOf: bob
+    });
+
+    _checkSuppliedAmounts(daiAssetId, _daiReserveId(spoke1), spoke1, bob, 0, 'after withdraw');
+  }
```
yan-man, line 503: sorry 1 other test, bc I was seeing issues with this in the prototype: withdraw partial, then withdraw max

CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -443,6 +443,95 @@ contract SpokeWithdrawTest is SpokeBase {
...
+    });
+
+    uint256 expectedSupplied = totalSupplied - partialWithdrawAmount;
+    assertEq(
+      expectedSupplied,
+      spoke1.getUserSuppliedAmount(_daiReserveId(spoke1), bob),
+      'expected supplied'
+    );
+
+    Utils.spokeWithdraw({
```
yan-man: nit: on the thing we're specifically isolating to test, prefer to explicitly call the method rather than using the abstracted helper.

tests/Base.t.sol
DhairyaSethi, line 719: string(abi.encodePacked -> string.concat

tests/Base.t.sol
```diff
@@ -702,4 +702,56 @@ abstract contract Base is Test {
     if (percent == 0) return price;
     return price.percentMul(percent);
   }
+
+  /// @dev Helper function to check consistent supplied amounts within accounting
+  function _checkSuppliedAmounts(
```
DhairyaSethi, line 707: don't like this method, it's only for the case of one user supplying to one spoke on a hub, would have preferred a more generalised method

PR #239, opened 2025-03-06
add ability to repay max

foodaka:
closes https://github.com/aave/aave-v4/issues/223

PR #311, opened 2025-04-01
Migrate Spoke Repay Tests

### LINKED ISSUE #290: Spoke Tests: Restore Refactor

CheyenneAtapour:
Migrate Spoke Repay Tests as per new accounting

Closes #290

### DIFF
--- a/tests/unit/Spoke/Spoke.Repay.t.sol
+++ b/tests/unit/Spoke/Spoke.Repay.t.sol
@@ -128,14 +128,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiRepayAmount = daiSupplyAmount / 4;
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -207,14 +207,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiRepayAmount = daiSupplyAmount / 4;
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -337,14 +337,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiBorrowAmount = daiSupplyAmount / 2;
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -451,14 +451,14 @@ contract SpokeRepayTest is SpokeBase {
     skipTime = uint40(bound(skipTime, 1, MAX_SKIP_TIME));
 
     // Bob supply weth as collateral
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai for Bob to borrow
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -540,14 +540,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiBorrowAmount = daiSupplyAmount / 2;
 
     // Bob supplies WETH as collateral
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supplies DAI
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrows DAI
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -630,14 +630,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiBorrowAmount = daiSupplyAmount / 2;
 
     // Bob supplies WETH as collateral
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supplies DAI
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrows DAI
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -775,14 +775,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiBorrowAmount = 100e18;
 
     // Bob supplies WETH as collateral
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supplies DAI
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrows DAI
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -860,14 +860,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiBorrowAmount = 100;
 
     // Bob supplies WETH as collateral
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supplies DAI
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrows DAI
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -981,14 +981,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiBorrowAmount = daiSupplyAmount / 2;
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -1097,14 +1097,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiBorrowAmount = daiSupplyAmount / 2;
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -1207,14 +1207,14 @@ contract SpokeRepayTest is SpokeBase {
     );
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -1312,14 +1312,14 @@ contract SpokeRepayTest is SpokeBase {
     );
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
@@ -1435,14 +1435,14 @@ contract SpokeRepayTest is SpokeBase {
     );
 
     // Bob supply weth
-    Utils.spokeSupply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
     setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
 
     // Alice supply dai
-    Utils.spokeSupply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
 
     // Bob borrow dai
-    Utils.spokeBorrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
 
     DataTypes.UserPosition memory bobDaiDataBefore = getUserInfo(
       spoke1,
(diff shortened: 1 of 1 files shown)


tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -75,869 +81,978 @@ contract SpokeRepayTest is SpokeBase {
...
+
+    assertEq(
+      tokenList.dai.balanceOf(bob),
+      bobDaiBalanceBefore - daiRepayAmount,
+      'bob dai final balance'
+    );
+    assertEq(tokenList.weth.balanceOf(bob), bobWethBalanceBefore);
   }
 
   function test_repay_revertsWith_amount_exceeds_debt() public {
```
```suggestion
  function test_repay_revertsWith_SurplusAmountRestored() public {
```

src/contracts/LiquidityHub.sol
```diff
@@ -207,7 +207,8 @@ contract LiquidityHub is ILiquidityHub {
     asset.updateBorrowRate({liquidityAdded: baseAmount, liquidityTaken: 0}); // both can be zero
 
     uint256 totalRestoredAmount = baseAmount + premiumAmount;
-    uint256 baseDrawnSharesRestored = asset.toDrawnSharesUp(baseAmount);
+    uint256 baseDrawnSharesRestored = asset.toDrawnSharesDown(baseAmount);
+    require(totalRestoredAmount > 0 || baseDrawnSharesRestored != 0, InvalidSharesAmount());
```
yan-man: don't think we need the first part, bc in _validateRestore we are already checking that total amount > 0

also, can you add a test for InvalidSharesAmount case? ie totalAmount > 0, but shares = 0

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -75,869 +81,978 @@ contract SpokeRepayTest is SpokeBase {
...
+
+    assertEq(
+      tokenList.dai.balanceOf(bob),
+      bobDaiBalanceBefore - daiRepayAmount,
+      'bob dai final balance'
+    );
+    assertEq(tokenList.weth.balanceOf(bob), bobWethBalanceBefore);
   }
 
   function test_repay_max() public {
```
yan-man, line 537: can you also add a version with repay partial then repay max? (found some underflow issue w this case on prototype)

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -75,869 +81,978 @@ contract SpokeRepayTest is SpokeBase {
...
+
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.Repay(
+      _daiReserveId(spoke1),
+      bob,
+      hub.convertToDrawnShares(daiAssetId, bobDaiBefore.baseDebt)
+    );
+    // Bob repays using repay Amount > full debt
+    vm.prank(bob);
+    spoke1.repay(_daiReserveId(spoke1), repayAmount);
```
yan-man: this test is supposed to be that - you don't just need to input uint max in order to repay full, you can input any value > your debt, and it'll repay max. That's why it is bounded as is previously, between `full debt + 1...uint max`

It also means that the method `_calculateRestoreAmount` should be updated to: 

`if (amount > baseDebt + premiumDebt) {
      return (baseDebt, premiumDebt);
    }`

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -75,869 +81,978 @@ contract SpokeRepayTest is SpokeBase {
...
+
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.Repay(
+      _daiReserveId(spoke1),
+      bob,
+      hub.convertToDrawnShares(daiAssetId, bobDaiBefore.baseDebt)
+    );
+    // Bob repays using max uint256 to signal full repayment
+    vm.prank(bob);
+    spoke1.repay(_daiReserveId(spoke1), type(uint256).max);
```
yan-man: see above, pending those changes to _calculateRestoreAmount: 

```suggestion
    spoke1.repay(_daiReserveId(spoke1), fullDebt+1);
```

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1384,113 +1582,127 @@ contract SpokeRepayTest is SpokeBase {
...
+
+      assertEq(tokenList.dai.balanceOf(bob), 0, 'bob dai final balance');
+      assertEq(tokenList.weth.balanceOf(bob), bobWethBalanceBefore);
+
+      // repays only base debt
+      assertGe(bobDaiAfter.baseDebt, daiBorrowAmount);
+    }
   }
 
   struct RepayMultipleLocal {
```
yan-man: general q - should we move structs to the top of the file or is it easier in the tests to keep it closer to the test where it's being utilized? I'm agnostic but just want to know so I can be consistent

CheyenneAtapour: I personally like them at the top of the file, unless it's more of a specific struct for that particular test case. But even then, I lean more towards just having structs as generic as possible, and put them at top.

Curious what @miguelmtzinf and @DhairyaSethi think

src/contracts/LiquidityHub.sol
CheyenneAtapour: Should we just get rid of this error then? Because if I remove the first part, this is just going to fail whenever shares are 0 but the amount is not 0. The idea here is that we shouldn't care if the user is trying to restore less than a share.

I can add a case, yes

yan-man: I think I misunderstood the intention then, I thought we were going to force the user to restore at least 1 share's worth of asset. 

but it makes sense to remove this require if we do want to allow users to restore for an asset amount > 0 even if shares rounds to 0.

CheyenneAtapour: That was also the plan, but then I found that it works fine if we relax this condition to have to pay at minimum one share amount. We can get feedback too from the others @miguelmtzinf @DhairyaSethi

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: If we roll with the current approach, where in `_calculateRestoreAmount`, we check if the user supplied a value greater than total debt, and return the current total debt, then we actually will not have the case of surplus debt restored. So I just removed this test

src/contracts/LiquidityHub.sol
miguelmtzinf: I d revert to what I initially proposed. (premiumAmount > 0 or shares > 0)
I don't think it makes sense to allow repaying non-0 assets and 0 shares. Actually, can it happen?
Not sure what would be the consequences of a user ending up with 0 shares and non-0 base debt, but system uses shares as source of truth for accounting and this non-0 base debt assets wouldn't be accruing interest?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -75,869 +81,1234 @@ contract SpokeRepayTest is SpokeBase {
...
+    bobDaiBefore.totalDebt = spoke1.getUserTotalDebt(_daiReserveId(spoke1), bob);
+    (bobDaiBefore.baseDebt, bobDaiBefore.premiumDebt) = spoke1.getUserDebt(
+      _daiReserveId(spoke1),
+      bob
+    );
+    assertGt(bobDaiBefore.totalDebt, daiBorrowAmount, 'Accrued interest increased bob dai debt');
+
+    uint256 repayAmount = 1;
+
+    // Ensure that the repay amount is less than 1 share
```
miguelmtzinf, line 838: don't think we should allow this

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -75,869 +81,1234 @@ contract SpokeRepayTest is SpokeBase {
...
+    assertEq(bobWethBefore.totalDebt, spoke1.getUserTotalDebt(_wethReserveId(spoke1), bob));
+    assertEq(
+      tokenList.dai.balanceOf(bob),
+      bobDaiBalanceBefore - repayAmount,
+      'bob dai final balance'
+    );
+    assertEq(tokenList.weth.balanceOf(bob), bobWethBalanceBefore);
   }
 
   function test_repay_fuzz_max_amount_gt_current_debt(uint256 repayAmount) public {
```
miguelmtzinf: should we fuzz initial borrow amount and skip time too?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -75,869 +81,1234 @@ contract SpokeRepayTest is SpokeBase {
...
+    assertEq(bobWethAfter.totalDebt, bobWethBefore.totalDebt);
+    assertEq(
+      tokenList.dai.balanceOf(bob),
+      bobDaiBalanceBefore - daiRepayAmount,
+      'bob dai final balance'
+    );
+    assertEq(tokenList.weth.balanceOf(bob), bobWethBalanceBefore);
   }
 
   /// repay all accrued base debt when outstanding premium is zero
```
miguelmtzinf:
```suggestion
  /// repay all accrued base debt interest when premium debt is zero
```

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -75,869 +81,1234 @@ contract SpokeRepayTest is SpokeBase {
...
+    assertEq(baseDaiDebt, 0);
+    assertEq(premiumDaiDebt, 0);
+
+    // verify LH asset debt is 0
+    uint256 lhAssetDebt = hub.getAssetTotalDebt(_daiReserveId(spoke1));
+    assertEq(lhAssetDebt, 0);
   }
 
-  /// repay all accrued base debt when outstanding premium is already repaid
+  /// repay all accrued base debt when premium debt is already repaid
```
miguelmtzinf:
```suggestion
  /// repay all accrued base debt interest when premium debt is already repaid
```
it's just the interest right?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1037,106 +1434,124 @@ contract SpokeRepayTest is SpokeBase {
...
+      assertEq(tokenList.dai.balanceOf(bob), 0, 'bob dai final balance');
+      assertEq(tokenList.weth.balanceOf(bob), bobWethBalanceBefore);
+
+      // repays only interest
+      // it can be equal because of 1 wei rounding issue when repaying
+      assertGe(bobDaiAfter.totalDebt, daiBorrowAmount);
+    }
   }
 
   /// repay all or a portion of outstanding premium debt
```
miguelmtzinf: do we wanna get rid of `outstanding` across the file?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1145,113 +1560,133 @@ contract SpokeRepayTest is SpokeBase {
...
+      daiRepayAmount = 0;
+      vm.expectRevert(ILiquidityHub.InvalidRestoreAmount.selector);
+      vm.prank(bob);
+      spoke1.repay(_daiReserveId(spoke1), daiRepayAmount);
+    } else {
+      // interest is at least 1
+      daiRepayAmount = bound(daiRepayAmount, 1, bobDaiPremium);
+      deal(address(tokenList.dai), bob, daiRepayAmount);
+
+      (uint256 baseRepaid, uint256 premiumRepaid) = _calculateRestoreAmount(
```
miguelmtzinf: we can simplify a bit as `baseRepaid` should be always 0?

tests/unit/Spoke/Spoke.Repay.t.sol
miguelmtzinf: no strong opinion, let's move it up

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1509,173 +1977,237 @@ contract SpokeRepayTest is SpokeBase {
...
+    if (amount >= baseDebt + premiumDebt) {
+      return (baseDebt, premiumDebt);
+    }
+    if (amount <= premiumDebt) {
+      return (0, amount);
+    }
+    return (amount - premiumDebt, premiumDebt);
   }
 
   /// todo: borrow, repay, borrow more, repay
```
miguelmtzinf: can we add these 2 or any other we may be missing?
also, a good test case to add is borrow X amount and get Y base debt shares, repay X and burn Y base debt shares, always.

src/contracts/LiquidityHub.sol
CheyenneAtapour: I've reverted and fixed the tests :)

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: I split into 2 test cases: One where this works because we are repaying premium debt, and another case where this fails because it's trying to repay only base debt, less than 1 share

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Yes, done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Yes, updated

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Done

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
```diff
@@ -389,11 +396,18 @@ contract LiquidityHubDrawTest is LiquidityHubBase {
     vm.assume(baseDebt > drawCap);
 
     // restore to provide liquidity
+    // Must repay at least one full share
+    uint256 singleShareAmount = hub.convertToDrawnAssets(daiAssetId, 1);
```
miguelmtzinf: does it make sense to create a helper like `minimumDebtAssetsPerShare`? same for supply

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1434,7 +1434,16 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiRepayAmount,
     uint40 skipTime
   ) public {
-    daiBorrowAmount = bound(daiBorrowAmount, 1, MAX_SUPPLY_AMOUNT / 2);
+    daiBorrowAmount = bound(
+      daiBorrowAmount,
+      hub.convertToDrawnAssets(daiAssetId, 1),
```
miguelmtzinf: this won't work well because ex rate will change during the test right? so better to recalculate and adjust right before the action?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1500,11 +1509,13 @@ contract SpokeRepayTest is SpokeBase {
       vm.prank(bob);
       spoke1.repay(_daiReserveId(spoke1), daiRepayAmount);
     } else {
-      // interest is at least 1
-      daiRepayAmount = bound(daiRepayAmount, 1, bobDaiInterest);
+      // Assume interest is at least 1 share
```
miguelmtzinf: do we need this? we could repay 1 amount of premium for example, so no friction here

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1780,7 +1800,10 @@ contract SpokeRepayTest is SpokeBase {
       spoke1.repay(_daiReserveId(spoke1), daiRepayAmount);
     } else {
       // interest is at least 1
-      daiRepayAmount = bound(daiRepayAmount, 1, bobDaiBaseDebt);
+      uint256 singleShareAmount = hub.convertToDrawnAssets(daiAssetId, 1);
+      daiRepayAmount = bobDaiBaseDebt >= singleShareAmount
```
miguelmtzinf: cannot we just recalc? if daiRepayAmount is < minimAssetsPerDebtShare, daiRepayAmount = minAssetsPerDebtShare
actually, is it possible that getUserDebt returns less than minAssetsPerDebtShare? i dont think so

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1909,7 +1945,10 @@ contract SpokeRepayTest is SpokeBase {
       spoke1.repay(_daiReserveId(spoke1), daiRepayAmount);
     } else {
       // interest is at least 1
-      daiRepayAmount = bound(daiRepayAmount, 1, bobDaiBaseDebt);
+      uint256 singleShareAmount = hub.convertToDrawnAssets(daiAssetId, 1);
```
miguelmtzinf: same

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -702,7 +709,90 @@ contract SpokeRepayTest is SpokeBase {
     assertEq(hub.getAssetTotalDebt(_daiReserveId(spoke1)), 0);
   }
 
-  function test_repay_less_than_share() public {
+  function test_repay_revertsWith_InvalidSharesAmount_repay_less_than_share() public {
```
miguelmtzinf: `test_repay_less_than_share_revertsWith_InvalidSharesAmount`

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Yes, done

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
CheyenneAtapour: Yes, done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: But this is not exactly the case, right?

It'd be borrow X amount, get Y base debt shares, repay _everything (max uint256)_, and burn Y base debt shares

Reason is that if exchange ratio is, say 1:20 (shares:assets), and I borrow 10, I really have 20 debt and 1 share that I need to repay. Repaying just the 10 reverts because it's less than 1 share

src/contracts/LiquidityHub.sol
```diff
@@ -196,7 +196,7 @@ contract LiquidityHub is ILiquidityHub {
     address from
   ) external returns (uint256) {
     // TODO: authorization - only spokes
-    // global & spoke premiumDebt (ghost, offset, unrealised) is *expected* to be updated on the `refreshPremiumDebt` callback
+    // global & spoke premiumDebt (ghost, offset, unrealized) is *expected* to be updated on the `refreshPremiumDebt` callback
```
```suggestion
    // global & spoke premiumDebt (ghost, offset, realized) is *expected* to be updated on the `refreshPremiumDebt` callback
```

src/contracts/Spoke.sol
```diff
@@ -420,13 +420,12 @@ contract Spoke is ISpoke {
     uint256 premiumDebt,
     uint256 amount
   ) internal view returns (uint256, uint256) {
-    if (amount == type(uint256).max) {
+    if (amount >= baseDebt + premiumDebt) {
```
DhairyaSethi: don't like the discrepancy w withdraw where it reverts trying withdraw more than balance until uint.max, we should follow that here as well imo

DhairyaSethi: or note and update in withdraw later

tests/Base.t.sol
```diff
@@ -64,6 +64,7 @@ abstract contract Base is Test {
   address internal alice = makeAddr('alice');
   address internal bob = makeAddr('bob');
   address internal carol = makeAddr('carol');
+  address internal dave = makeAddr('dave');
```
```suggestion
  address internal derl = makeAddr('derl');
```
pls ty

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
```diff
@@ -356,8 +356,14 @@ contract LiquidityHubDrawTest is LiquidityHubBase {
     assertGt(baseDebt, drawCap);
 
     // restore to provide liquidity
+    // Must repay at least one full share
     vm.startPrank(address(spoke1));
-    hub.restore({assetId: daiAssetId, baseAmount: 1, premiumAmount: 0, from: alice});
+    hub.restore({
+      assetId: daiAssetId,
+      baseAmount: minimumAssetsPerDrawnShare(daiAssetId),
+      premiumAmount: 0,
```
DhairyaSethi, line 368: not sure if this is totally accurate bc we always deduct from premium debt first (expected to be settled in a refresh before) and correct amount passed. with existing premium debt the accounting is not accurate

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1507,177 +2085,1098 @@ contract SpokeRepayTest is SpokeBase {
...
+      // Check hub debts match sum of user debts
+      assertEq(hub.getAssetTotalDebt(daiAssetId), totalDaiDebtActual, 'dai asset debt');
+      assertEq(hub.getAssetTotalDebt(wethAssetId), totalWethDebtActual, 'weth asset debt');
+      assertEq(hub.getAssetTotalDebt(usdxAssetId), totalUsdxDebtActual, 'usdx asset debt');
+      assertEq(hub.getAssetTotalDebt(wbtcAssetId), totalWbtcDebtActual, 'wbtc asset debt');
+    }
+  }
+  */
+
+  function _calculateRestoreAmount(
```
DhairyaSethi: move to base?

src/contracts/Spoke.sol
```diff
@@ -639,9 +639,12 @@ contract Spoke is ISpoke {
     DataTypes.UserPosition storage userPosition,
     uint256 assetId
   ) internal view returns (uint256, uint256) {
-    uint256 premiumDebt = userPosition.realizedPremium +
-      (hub.convertToDrawnAssets(assetId, userPosition.premiumDrawnShares) -
-        userPosition.premiumOffset);
+    // underflow protection: conversion of `premiumDrawnShares` into assets is accurate to 1 wei precision
```
DhairyaSethi: we're sure it's only ever off by 1 wei? (idts)

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -2323,40 +2326,46 @@ contract SpokeRepayTest is SpokeBase {
...
+    action1.skipTime = uint40(bound(action1.skipTime, 1, MAX_SKIP_TIME / 2));
+    action2.skipTime = uint40(bound(action2.skipTime, 1, MAX_SKIP_TIME / 2));
+    action1.borrowAmount = bound(action1.borrowAmount, 10e18, MAX_SUPPLY_AMOUNT / 4);
+    action2.borrowAmount = bound(action2.borrowAmount, 10e18, MAX_SUPPLY_AMOUNT / 4);
+    action1.repayAmount = bound(action1.repayAmount, 10e18, action1.borrowAmount);
+    action2.repayAmount = bound(action2.repayAmount, 10e18, action2.borrowAmount);
+    console.log('action1.borrowAmount', action1.borrowAmount);
+    console.log('action1.repayAmount', action1.repayAmount);
+    console.log('action2.borrowAmount', action2.borrowAmount);
+    console.log('action2.repayAmount', action2.repayAmount);
```
DhairyaSethi: logs

miguelmtzinf:
Added underflow protection in a separate branch, so it's easier to test with and without it.
https://github.com/aave/aave-v4/pull/326

src/contracts/Spoke.sol
miguelmtzinf: based on fuzzing results, confirmed it's only 1 wei

src/contracts/Spoke.sol
miguelmtzinf: yeah, we should follow same approach in withdraw. Let's fix in a separate PR

CheyenneAtapour: cool, sounds good to me!

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Done

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
CheyenneAtapour: For sure, this was just to provide some available liquidity. But I went ahead and set the liquidity premium to 0, so now this should be fine

tests/Base.t.sol
CheyenneAtapour: Done :)

src/contracts/LiquidityHub.sol

src/contracts/Spoke.sol
DhairyaSethi: no before we do that, i still don't agree it should be this way in repay. @CheyenneAtapour could you write the original rationale again? (something that's not a patch to make the tests work)

src/contracts/Spoke.sol
DhairyaSethi: we discovered this is not the case

src/contracts/Spoke.sol
CheyenneAtapour: There's no rationale, and I'm fine with going back to how it was. There was a previous comment in this PR suggesting it, but we can revert if we all agree

yan-man: here is previous discussion on repay max: https://github.com/aave/aave-v4/pull/239#discussion_r1986169048, so think we should update withdraw to match this

CheyenneAtapour: Oops, I reverted this change, sounds like majority is saying to put it back. @DhairyaSethi do you agree with it?

src/contracts/Spoke.sol
```diff
@@ -445,7 +445,6 @@ contract Spoke is ISpoke {
     if (amount <= premiumDebt) {
       return (0, amount);
     }
-    // todo ensure `amount` is not greater than total debt?
```
DhairyaSethi: lets retain this comment for now?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -67,877 +112,1165 @@ contract SpokeRepayTest is SpokeBase {
...
-    //     (uint256 baseDaiDebt, uint256 outstandingDaiPremium) = spoke1.getReserveDebt(
-    //       _daiReserveId(spoke1)
-    //     );
-    //     assertEq(baseDaiDebt, 0);
-    //     assertEq(outstandingDaiPremium, 0);
-
-    //     // verify LH asset debt is 0
-    //     uint256 lhAssetDebt = hub.getAssetCumulativeDebt(_daiReserveId(spoke1));
-    //     assertEq(lhAssetDebt, 0);
+  function test_repay_partial_then_max() public {
```
DhairyaSethi, line 627: can we just have more granularity here? tests which repay:
- partial premium debt (also assert base debt untouched)
- complete premium debt (also assert base debt untouched)
- partial base debt (also assert premium debt was fully repaid)

DhairyaSethi: i think most of these (and more are covered, just comments explaining which covers the case would be great

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1507,177 +1904,1573 @@ contract SpokeRepayTest is SpokeBase {
...
+    Utils.spokeBorrow(spoke1, _wethReserveId(spoke1), bob, wethInfo.borrowAmount, bob);
+    Utils.spokeBorrow(spoke1, _usdxReserveId(spoke1), bob, usdxInfo.borrowAmount, bob);
+    Utils.spokeBorrow(spoke1, _wbtcReserveId(spoke1), bob, wbtcInfo.borrowAmount, bob);
+
+    daiInfo.posBefore = getUserInfo(spoke1, bob, _daiReserveId(spoke1));
+    wethInfo.posBefore = getUserInfo(spoke1, bob, _wethReserveId(spoke1));
+    usdxInfo.posBefore = getUserInfo(spoke1, bob, _usdxReserveId(spoke1));
+    wbtcInfo.posBefore = getUserInfo(spoke1, bob, _wbtcReserveId(spoke1));
+
+    Debts memory bobDaiBefore;
```
DhairyaSethi, line 2009: this can be cleaned up slightly with the 2d array of users x reserves you have added elsewhere but nbd

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1507,177 +1904,1573 @@ contract SpokeRepayTest is SpokeBase {
...
+    assertEq(spoke1.getUserTotalDebt(_daiReserveId(spoke1), bob), 0, 'bob total debt after repay');
+  }
+
+  function test_repay_multiple_users_multiple_assets(
+    UserAssetInfo memory bobInfo,
+    UserAssetInfo memory aliceInfo,
+    UserAssetInfo memory carolInfo,
+    uint40 skipTime
+  ) public {
+    vm.skip(true, 'Pending fix for this test');
```
DhairyaSethi:
```suggestion
    vm.skip(true, 'Pending fix for underflow issue');
```
or link issue number, just anything slightly more descriptive

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
```diff
@@ -343,6 +343,9 @@ contract LiquidityHubDrawTest is LiquidityHubBase {
   }
 
   function test_draw_revertsWith_DrawCapExceeded_due_to_interest() public {
+    // Set liquidity premium of dai to 0
+    updateLiquidityPremium(spoke1, _daiReserveId(spoke1), 0);
+
```
yan-man, line 349: can assertTrue here that LP is now 0

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
```diff
@@ -386,11 +395,19 @@ contract LiquidityHubDrawTest is LiquidityHubBase {
     });
 
     (uint256 baseDebt, ) = hub.getAssetDebt(daiAssetId);
-    vm.assume(baseDebt > drawCap);
+    uint256 singleShareAmount = minimumAssetsPerDrawnShare(daiAssetId);
```
yan-man:
```suggestion
    uint256 singleShareInAssets = minimumAssetsPerDrawnShare(daiAssetId);
```

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -6,6 +6,50 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
...
+
+  struct UserAssetInfo {
+    AssetInfo daiInfo;
+    AssetInfo wethInfo;
+    AssetInfo usdxInfo;
+    AssetInfo wbtcInfo;
+    address user;
+  }
+
   function test_repay_all_with_accruals() public {
```
yan-man, line 53: can you add some assertions here after repay all or is the point just that it doesn't revert on repay uint max? if so pls just add some /// comments to describe point of the test

src/contracts/Spoke.sol
CheyenneAtapour: I just made a separate PR for this if/when we agree on these changes: https://github.com/aave/aave-v4/pull/338

src/contracts/Spoke.sol

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: I added the comments, and changed 1 test to cover these

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: Only 1 user here, so skipping for now

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: done

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
CheyenneAtapour: done

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour: done

PR #313, opened 2025-04-01
fix: Fix hub restore rounding

### LINKED ISSUE #312: LiquidityHub.restore should round assets down, as users must over-repay
We want users to repay as less shares as possible based on their assets amount input.

https://github.com/aave/aave-v4/blob/3d56bc69abc73f9e684b922608c441c768aacb91/src/contracts/LiquidityHub.sol#L210

reference: https://github.com/aave/aave-v4/pull/277#discussion_r2013879392

miguelmtzinf:
Closes #312

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -196,7 +196,7 @@ contract LiquidityHub is ILiquidityHub {
     address from
   ) external returns (uint256) {
     // TODO: authorization - only spokes
-    // global & spoke premiumDebt (ghost, offset, unrealised) is *expected* to be updated on the `refreshPremiumDebt` callback
+    // global & spoke premiumDebt (ghost, offset, unrealized) is *expected* to be updated on the `refreshPremiumDebt` callback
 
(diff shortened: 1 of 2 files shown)


src/contracts/LiquidityHub.sol
```diff
@@ -207,7 +207,8 @@ contract LiquidityHub is ILiquidityHub {
     asset.updateBorrowRate({liquidityAdded: baseAmount, liquidityTaken: 0}); // both can be zero
 
     uint256 totalRestoredAmount = baseAmount + premiumAmount;
-    uint256 baseDrawnSharesRestored = asset.toDrawnSharesUp(baseAmount);
+    uint256 baseDrawnSharesRestored = asset.toDrawnSharesDown(baseAmount);
+    require(premiumAmount > 0 || baseDrawnSharesRestored != 0, InvalidSharesAmount());
```
CheyenneAtapour, line 211: What if only base debt is restored, and it's less than a share? Are we concerned about this case?

miguelmtzinf: user needs to repay at least 1 share

PR #318, opened 2025-04-02
test: Add tests for Hub refreshPremiumDebt

### LINKED ISSUE #314: Spoke Tests: Refresh

miguelmtzinf:
Closes #314

### DIFF
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -1,13 +1,13 @@
 {
   "borrow": "486714",
-  "repay: full": "218997",
-  "repay: partial": "288228",
-  "supply: 0 debt, collateralDisabled": "169591",
-  "supply: 0 debt, collateralEnabled": "169567",
-  "supply: 1 debt": "169615",
-  "supply: 2 debt": "169626",
-  "supply: 3 debt": "169579",
+  "repay: full": "218882",
+  "repay: partial": "288236",
+  "supply: 0 debt, collateralDisabled": "169579",
+  "supply: 0 debt, collateralEnabled": "169555",
+  "supply: 1 debt": "169603",
+  "supply: 2 debt": "169614",
+  "supply: 3 debt": "169567",
   "usingAsCollateral": "50661",
-  "withdraw: full": "199829",
-  "withdraw: partial": "211692"
+  "withdraw: full": "199855",
+  "withdraw: partial": "211718"
 }


tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+      );
+  }
+
+  function _assertPremiumData(PremiumDataLocal memory a, PremiumDataLocal memory b) internal pure {
+    assertEq(a.premiumDebtShares, b.premiumDebtShares, 'premium debt shares do not match');
+    assertEq(a.premiumOffset, b.premiumOffset, 'premium offset do not match');
+    assertEq(a.realizedPremium, b.realizedPremium, 'realized premium do not match');
+  }
+
+  // only assets listed can be refreshed
```
yan-man, line 48: saw before you were using /// rather than // for test descriptions, do you care either way? just want to keep my stuff consistent with whatever we're doing 

https://github.com/aave/aave-v4/commit/767d002973982a1613c52d2a8533277e7505b5c2#diff-c3b728636ae9e535fdd38fa2792f03669a2ac8dfa2213cf1fb7d16d233141da7

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+        uint256(newPremiumOffset + int256(initPremiumOffset)),
+        uint256(newRealizedPremium + int256(initRealizedPremium))
+      );
+      _assertPremiumData(_loadAssetPremiumData(assetId), newData);
+      _assertPremiumData(_loadSpokePremiumData(assetId, address(spoke1)), newData);
+    }
+  }
+
+  // refresh premium data of DAI by 3 different spokes
+  /// forge-config: default.fuzz.runs = 100
```
yan-man, line 225: why restrict fuzz runs?

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+  // premium data values cannot go negative
+  function test_refresh_from_init_revertsWith_underflow() public {
+    uint256 assetId = 0;
+    _assertPremiumData(_loadAssetPremiumData(assetId), PremiumDataLocal(0, 0, 0));
+    _assertPremiumData(_loadSpokePremiumData(assetId, address(spoke1)), PremiumDataLocal(0, 0, 0));
+
+    vm.expectRevert(stdError.arithmeticError);
+    vm.prank(address(spoke1));
+    hub.refreshPremiumDebt(assetId, -1, -1, -1);
+  }
```
yan-man, line 95: worth it to add another test to fuzz each of the refreshPremiumDebt params negative input value separately?

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+    uint256 targetToChange,
+    uint256 newValue
+  ) public {
+    uint256 assetId = daiAssetId;
+    address[3] memory spokes = [address(spoke1), address(spoke2), address(spoke3)];
+    spokeIndex = bound(spokeIndex, 0, spokes.length - 1);
+    initPremiumShares = bound(initPremiumShares, 0, MAX_SUPPLY_AMOUNT / 2);
+    initPremiumOffset = bound(initPremiumOffset, 0, MAX_SUPPLY_AMOUNT / 2);
+    initRealizedPremium = bound(initRealizedPremium, 0, MAX_SUPPLY_AMOUNT / 2);
+    targetToChange = bound(targetToChange, 0, 3); // [premiumDebtShares, premiumOffset, realizedPremium]
```
yan-man, line 240: I think can just bound to 2?
```suggestion
    targetToChange = bound(targetToChange, 0, 2); // [premiumDebtShares, premiumOffset, realizedPremium]
```

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+    uint256 notListedAsset = hub.assetCount() + 1;
+
+    vm.expectRevert(ILiquidityHub.AssetNotListed.selector);
+    vm.prank(address(spoke1));
+    hub.refreshPremiumDebt(notListedAsset, 1, 1, 1);
+  }
+
+  // only authorized spokes can refresh
+  function test_refresh_revertsWith_AuthorizedSpoke() public {
+    // TODO
```
DhairyaSethi, line 59:
```suggestion
    vm.skip(true, 'todo after spoke auth');
```

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+    DataTypes.SpokeData memory spokeData = hub.getSpoke(assetId, spoke);
+    return
+      PremiumDataLocal(
+        spokeData.premiumDrawnShares,
+        spokeData.premiumOffset,
+        spokeData.realizedPremium
+      );
+  }
+
+  function _assertPremiumData(PremiumDataLocal memory a, PremiumDataLocal memory b) internal pure {
```
DhairyaSethi, line 42: i'd just name this as an `assertEq` tbh

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+    uint256 assetId = 0;
+    _assertPremiumData(_loadAssetPremiumData(assetId), PremiumDataLocal(0, 0, 0));
+    _assertPremiumData(_loadSpokePremiumData(assetId, address(spoke1)), PremiumDataLocal(0, 0, 0));
+
+    vm.expectRevert(stdError.arithmeticError);
+    vm.prank(address(spoke1));
+    hub.refreshPremiumDebt(assetId, -1, -1, -1);
+  }
+
+  // refresh premium data with positive values from initial state
```
DhairyaSethi, line 97: currrently it's not positive values when the latter three fuzzed values are zero but nbd

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+  function test_refresh_from_init_fuzz_positive_amounts(
```
DhairyaSethi, line 98: we'd be adding more assertions in the refresh method that total debt has not changed 
ideally it can only decrease on a refresh and remain the same on all other cases

refresh can **never** increase total debt, i would change this method to negative amounts for now tbh

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+    assetId = bound(assetId, 0, hub.assetCount() - 1);
+    initPremiumShares = bound(initPremiumShares, 0, MAX_SUPPLY_AMOUNT / 2);
+    initPremiumOffset = bound(initPremiumOffset, 0, MAX_SUPPLY_AMOUNT / 2);
+    initRealizedPremium = bound(initRealizedPremium, 0, MAX_SUPPLY_AMOUNT / 2);
+    premiumDebtShares = bound(premiumDebtShares, 0, MAX_SUPPLY_AMOUNT / 2);
+    premiumOffset = bound(premiumOffset, 0, MAX_SUPPLY_AMOUNT / 2);
+    realizedPremium = bound(realizedPremium, 0, MAX_SUPPLY_AMOUNT / 2);
+
+    // random sign: negative if even
+    int256 newPremiumShares = (initPremiumShares + premiumDebtShares) % 2 == 0
```
DhairyaSethi, line 157:
```suggestion
    // random sign: negative if even
    int256 newPremiumShares = (vm.randomUint()) % 2 == 0
```

or we add a `_randomChance(5)` helper in baseTest

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
```diff
@@ -0,0 +1,316 @@
...
+      int256(15)
+    );
+    vm.prank(address(spoke1));
+    hub.refreshPremiumDebt(assetId, 5, 10, 15);
+
+    _assertPremiumData(_loadAssetPremiumData(assetId), PremiumDataLocal(5, 10, 15));
+    _assertPremiumData(
+      _loadSpokePremiumData(assetId, address(spoke1)),
+      PremiumDataLocal(5, 10, 15)
+    );
```
DhairyaSethi, line 83: we can also assert the correct total debt increase (and that ONLY premium debt has increased)

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
miguelmtzinf: takes a long time otherwise. I can remove and decrease iterations maybe, or just leave it as it is and create a new profile for the CI where we used reduced number of fuzz runs?

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
miguelmtzinf: added no-fuzz test `test_refresh_from_init_revertsWith_underflow_separately`. lmk if works or you think fuzz adds more value here

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
miguelmtzinf: good catch! fixed

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
DhairyaSethi: i'm ok w overriding fuzz runs on the default profile (local) to 256 or 1000 runs at the very least. but ci should be running the 10k runs 
it's parallelised on the ci, if it takes over 20 mins we can reconsider for particular tests

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
miguelmtzinf: i mean going from init (zero) and using positive inputs/delta. Starting from random positive state in other tests (e.g. `test_refresh_fuzz_amounts`)

tests/unit/LiquidityHub/LiquidityHub.RefreshPremiumDebt.t.sol
miguelmtzinf: this test goes from 0 to >0, so cannot add negative delta here. 
`test_refresh_fuzz_amounts` does the work you suggest I think. Is it fine to you or should I add something else?

DhairyaSethi: yeah my point was we'd need to rm this test once we add the assertions on refresh that total debt cannot increase but yeah since the other test covers both (including -ve) cases it's fine

DhairyaSethi:
superseded by #597

PR #321, opened 2025-04-03
Migrate Spoke User Risk Premium Tests

### LINKED ISSUE #315: Spoke Tests: User Risk Premium
Migrate Spoke's user risk premium tests, also removing any unnecessary tests

CheyenneAtapour:
Closes #315


src/contracts/Spoke.sol
```diff
@@ -349,6 +349,20 @@ contract Spoke is ISpoke {
...
   }
 
+  /// @dev Reverts if user is not borrowing reserveId
+  function getUserPreviousRiskPremium(
+    uint256 reserveId,
+    address user
+  ) external view returns (uint256) {
+    if (!_isBorrowing(_userPositions[user][reserveId])) {
+      revert UserNotBorrowingReserve(reserveId);
+    }
```
```suggestion
    require(_isBorrowing(_userPositions[user][reserveId])), UserNotBorrowingReserve(reserveId));
```

src/contracts/Spoke.sol
```diff
@@ -349,6 +349,20 @@ contract Spoke is ISpoke {
...
+    return
+      _userPositions[user][reserveId].premiumDrawnShares.rayDiv(
+        _userPositions[user][reserveId].baseDrawnShares
+      ) / 1e23;
+  }
```
```suggestion
    return
      _userPositions[user][reserveId].premiumDrawnShares.percentDiv(
        _userPositions[user][reserveId].baseDrawnShares
      );
  }
```

snapshots/Hub.Operations.json
```diff
@@ -1,8 +1,8 @@
 {
-  "add": "184887",
-  "draw": "124244",
-  "remove: full": "81367",
-  "remove: partial": "94523",
-  "restore: full": "92109",
-  "restore: partial": "101709"
+  "add": "163151",
+  "draw": "63060",
```
DhairyaSethi: pls regen the snapshots w isolate mode

src/contracts/Spoke.sol

snapshots/Hub.Operations.json
CheyenneAtapour: Added

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,814 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+
+    daiInfo.reserveId = _daiReserveId(spoke1);
+    usdxInfo.reserveId = _usdxReserveId(spoke1);
+
+    daiInfo.lp = spoke1.getLiquidityPremium(daiInfo.reserveId);
+    usdxInfo.lp = spoke1.getLiquidityPremium(usdxInfo.reserveId);
+
+    // Bob supply dai into spoke1
+    Utils.supply(spoke1, daiInfo.reserveId, bob, bobDaiInfo.supplyAmount, bob);
+    setUsingAsCollateral(spoke1, bob, daiInfo.reserveId, true);
```
DhairyaSethi: Utils.supplyCollateral abstracts both calls

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,814 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+    assertEq(spoke1.getLiquidityPremium(_usdxReserveId(spoke1)), 50_00, 'usdx lp');
+    assertEq(spoke1.getLiquidityPremium(_wethReserveId(spoke1)), 15_00, 'weth lp');
+    assertEq(spoke1.getLiquidityPremium(_daiReserveId(spoke1)), 20_00, 'dai lp');
+
+    // Set LT to 1 for Alice collateral
+    updateCollateralFactor(spoke1, _wethReserveId(spoke1), 10_000);
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 10_000);
+
+    vm.prank(bob);
+    spoke1.supply(_daiReserveId(spoke1), vars.daiBorrowAmount);
```
DhairyaSethi: maybe makes sense to refactor this test to use utils lib? nbd

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,814 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+    assertEq(baseDaiDebt, 0);
+    assertEq(daiPremiumDebt, 0);
+    (baseDaiDebt, daiPremiumDebt) = spoke1.getReserveDebt(_daiReserveId(spoke1));
+    assertEq(baseDaiDebt, 0);
+    assertEq(daiPremiumDebt, 0);
+    (baseDaiDebt, daiPremiumDebt) = hub.getSpokeDebt(daiAssetId, address(spoke1));
+    assertEq(baseDaiDebt, 0);
+    assertEq(daiPremiumDebt, 0);
+
+    assertEq(spoke1.getUserRiskPremium(alice), 0);
```
DhairyaSethi: can we add this assert to the _repayAll helper as well? also we could use that helper here

snapshots/Spoke.Operations.json
```diff
@@ -7,7 +7,7 @@
   "supply: 1 debt": "170225",
   "supply: 2 debt": "170236",
   "supply: 3 debt": "170189",
-  "usingAsCollateral": "50661",
+  "usingAsCollateral": "50683",
```
DhairyaSethi, line 10: very curious about this gas increase. since we don't touch this function (or any associated code) & just added one function my guess is that it's become slightly more expensive to search for this selector now. we have enough for the compiler to resort to adding the binary search routine and newly added selector is less than this 🤔

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,814 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+    Utils.borrow(spoke1, usdxInfo.reserveId, bob, bobUsdxInfo.borrowAmount, bob);
+
+    // Alice draw dai
+    Utils.borrow(spoke1, daiInfo.reserveId, alice, aliceDaiInfo.borrowAmount, alice);
+
+    // Alice draw usdx
+    Utils.borrow(spoke1, usdxInfo.reserveId, alice, aliceUsdxInfo.borrowAmount, alice);
+
+    ExpectedUserRp memory expectedUserRp;
+    expectedUserRp.bobExpectedRiskPremium = _calculateExpectedUserRP(bob, spoke1);
```
DhairyaSethi: i would rename the struct keys to avoid `expected` twice but nbd

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,814 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+        'alice risk premium after borrow'
+      );
+
+      // Verify Bob's risk premium remains the same
+      assertApproxEqAbs(
+        spoke1.getUserRiskPremium(bob),
+        bobExpectedRiskPremium,
+        2,
+        'bob risk premium after alice borrow'
+      );
```
DhairyaSethi: ci fails here with a delta of 120, as we were discussing, possibly because of supply ex rate increasing due to donations

CheyenneAtapour: yep, changed this assumption

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: Done

snapshots/Spoke.Operations.json
CheyenneAtapour: The gas also just increased quite a bit here with my latest commit, can look into this

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: Done, I check that hf is max(uint256) because if we only repay 1 asset and have another debt, user risk premium can be nonzero. But if we have no debts, both hf will be max, and user rp will be 0

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: Nice, done

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1121 @@
...
+    for (uint256 reserveId = 0; reserveId < spoke1.reserveCount(); reserveId++) {
+      DataTypes.UserPosition memory bobInfo = getUserInfo(spoke1, bob, reserveId);
+      assertEq(bobInfo.suppliedShares, 0, 'bob supplied collateral');
+    }
+    assertEq(spoke1.getUserRiskPremium(bob), 0, 'user risk premium');
+  }
+
+  /// Without a collateral set, user risk premium is 0.
+  function test_getUserRiskPremium_no_collateral_set() public {
+    Utils.supply(spoke1, _daiReserveId(spoke1), bob, 100e18, bob);
```
yan-man, line 45: maybe worthwhile to assert here that the dai supply is indeed not set usingAsCollateral

better yet, maybe the loop above to go over all reserves and assert that no reserve is set as collateral

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1121 @@
...
+
+    // With single collateral, user rp will match liquidity premium of collateral
+    assertEq(userRiskPremium, daiInfo.config.liquidityPremium, 'user risk premium');
+  }
+
+  /// When supplying and borrowing one reserve (fuzzed amounts), user risk premium matches the liquidity premium of that reserve.
+  function test_getUserRiskPremium_fuzz_single_reserve_collateral_borrowed_amount(
+    uint256 borrowAmount
+  ) public {
+    borrowAmount = bound(borrowAmount, 1, MAX_SUPPLY_AMOUNT / 2);
```
yan-man, line 86: can you add a todo to add another test here for borrow amounts >= supplied amount too, ie undercollateralized case?

perhaps can go into your next PR, using something like this helper: https://github.com/aave/aave-v4/pull/331/files#diff-115b914a0e1cfa085c4ee7ab9d3d3cc0191508b4201224757f0f31db4a719fa3R93

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1121 @@
...
+
+    // With single collateral, user rp will match liquidity premium of collateral
+    assertEq(userRiskPremium, daiInfo.lp, 'user risk premium');
+
+    // Supplying more risky reserve (usdx) should not impact user risk premium
+    Utils.supplyCollateral(spoke1, usdxInfo.reserveId, bob, additionalSupplyAmount, bob);
+    assertEq(spoke1.getUserRiskPremium(bob), userRiskPremium, 'user risk premium after supply');
+  }
+
+  function test_riskPremium_postActions() public {
```
yan-man, line 200: can you add some comments describing this test

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+      hub.convertToDrawnAssets(daiAssetId, alicePosition.premiumDrawnShares) -
+      alicePosition.premiumOffset;
+
+    (baseDaiDebt, daiPremiumDebt) = spoke1.getUserDebt(_daiReserveId(spoke1), alice);
+
+    assertEq(baseDaiDebt, vars.daiBorrowAmount + accruedDaiDebt, 'dai base debt after weth supply');
+    assertEq(daiPremiumDebt, expectedPremiumDebt, 'dai premium debt after weth supply');
+
+    // Alice repays everything
+    _repayAll(spoke1, _daiReserveId);
```
yan-man, line 199: can you rename this to something that also implies that rp is checked after repay too

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1121 @@
...
+
+    Utils.withdraw(spoke1, _daiReserveId(spoke1), bob, 0.01e18, bob);
+    _assertUserRpUnchanged(_daiReserveId(spoke1), spoke1, bob);
+    _assertUserRpUnchanged(_usdxReserveId(spoke1), spoke1, bob);
+
+    Utils.withdraw(spoke1, _usdxReserveId(spoke1), bob, 0.01e6, bob);
+    _assertUserRpUnchanged(_daiReserveId(spoke1), spoke1, bob);
+    _assertUserRpUnchanged(_usdxReserveId(spoke1), spoke1, bob);
+
+    skip(232 days);
```
yan-man, line 221: maybe for next PR (and maybe you already will do this), but can you also test that after time has passed (but without action), that these asserts would be false (assuming supply/debt grow at diff rates)?

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1121 @@
...
+    // Bob supply usdx into spoke3
+    if (usdxInfo.supplyAmount > 0) {
+      Utils.supplyCollateral(spoke3, usdxInfo.reserveId, bob, usdxInfo.supplyAmount, bob);
+    }
+
+    // Bob supply wbtc into spoke3
+    Utils.supplyCollateral(spoke3, wbtcInfo.reserveId, bob, wbtcInfo.supplyAmount, bob);
+
+    // Alice supply remaining weth into spoke3
+    if (MAX_SUPPLY_AMOUNT - wethInfo.supplyAmount > 0) {
```
yan-man, line 1045: why does alice only supply this specific remaining amount into spoke3, is it just to have borrowable liquidity available?

if so, can also use _deployLiquidity helper

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: I'll add the todo. Can you perma-link me the above? It doesn't seem to direct properly

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: @DhairyaSethi do you have a good idea for a name here?

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: Yes, I am planning to add such tests

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: Yes, done

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
DhairyaSethi: not really, imo this works, it's an obvious side effect that when repay all debt for asset will be zero and rp can potentially be zero, idt we need to make it obvious here since the rp check is conditional

src/contracts/Spoke.sol
```diff
@@ -370,6 +370,18 @@ contract Spoke is ISpoke {
     return reserve.premiumDrawnShares.rayDiv(reserve.baseDrawnShares); // trailing
   }
 
+  /// @dev Reverts if user is not borrowing reserveId
+  function getUserPreviousRiskPremium(
```
avniculae: previous seems a bit ambiguous. maybe `getUserRiskPremiumAtLastUpdate`, wdyt?

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -107,14 +107,20 @@ contract SpokeBase is Base {
...
+    uint256 supplyAmount;
     uint256 borrowAmount;
+    uint256 suppliedShares;
     uint256 repayAmount;
     uint256 baseRestored;
     uint256 premiumRestored;
     address user;
   }
 
+  struct UserBorrowAction {
```
avniculae, line 119: can you pls explain the difference between UserAction and UserBorrowAction?

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -653,65 +663,65 @@ contract SpokeBase is Base {
...
+      uint256 supplyAmount = _getReserveValueInBaseCurrency(
+        assetId,
+        hub.convertToSuppliedAssets(assetId, userPosition.suppliedShares)
+      );
+
+      if (supplyAmount >= totalDebt) {
+        userRP += totalDebt * lp;
+        utilizedSupply += totalDebt;
+        totalDebt = 0;
+        break;
```
avniculae, line 715: no need to break here, totalDebt is 0 so while will break irrespectively

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -653,65 +663,65 @@ contract SpokeBase is Base {
...
+    for (uint256 reserveId; reserveId < spoke.reserveCount(); reserveId++) {
+      if (spoke.getUsingAsCollateral(reserveId, user)) {
+        reserveLP.add(idx, spoke.getLiquidityPremium(reserveId), reserveId);
+        ++idx;
+      }
+    }
+
+    // Sort supplied reserves by LP
+    reserveLP.sortByKey();
+
```
avniculae, line 698: shall we assert here that keys are sorted correctly? (in linear time)

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1122 @@
...
+    uint256 daiReserveId = _daiReserveId(spoke1);
+    uint256 daiAmount = 100e18;
+
+    // Bob supply dai into spoke1
+    Utils.supplyCollateral(spoke1, daiReserveId, bob, daiAmount, bob);
+
+    assertEq(spoke1.getUserRiskPremium(bob), 0, 'user risk premium');
+  }
+
+  /// When supplying and borrowing one reserve, user risk premium matches the liquidity premium of that reserve.
```
avniculae, line 65: why do we want to allow borrows of supplied assets?

src/contracts/Spoke.sol
avniculae: also, do we want to add this getter? we can already compute this using `getUserPosition`, as in `_assertUserRpUnchanged`

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1122 @@
...
+    Utils.borrow(spoke1, usdxInfo.reserveId, bob, usdxInfo.borrowAmount, bob);
+
+    // Weth is enough to cover the total debt
+    assertGe(
+      _getReserveValueInBaseCurrency(wethAssetId, wethInfo.supplyAmount),
+      _getReserveValueInBaseCurrency(daiAssetId, daiInfo.borrowAmount) +
+        _getReserveValueInBaseCurrency(usdxAssetId, usdxInfo.borrowAmount),
+      'weth supply covers debt'
+    );
+    uint256 expectedUserRiskPremium = wethInfo.lp;
```
avniculae, line 268: this is only true when weth has the lowest liquidity premium. can we add one more test where one supplied asset covers the entire debt, but there are other supplied assets with lower liquidity premium?

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1122 @@
...
+    // Alice supply weth into spoke1
+    Utils.supplyCollateral(spoke1, wethInfo.reserveId, alice, wethInfo.supplyAmount, alice);
+
+    // Bob draw weth
+    Utils.borrow(spoke1, wethInfo.reserveId, bob, wethInfo.borrowAmount, bob);
+
+    // Dai and usdx will each cover half the debt, because dai has lower lp than usdx
+    assertEq(
+      spoke1.getUserRiskPremium(bob),
+      _calculateExpectedUserRP(bob, spoke1),
```
avniculae: I think this function is good, but it would have been good to assert here against (dai.lp + usdx.lp)/2, wdyt?

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1122 @@
...
+    uint256 wethSupplyAmount,
+    uint256 wbtcBorrowAmount
+  ) public {
+    uint256 totalBorrowAmount = MAX_SUPPLY_AMOUNT / 2;
+    daiSupplyAmount = bound(daiSupplyAmount, 0, MAX_SUPPLY_AMOUNT);
+    wethSupplyAmount = bound(wethSupplyAmount, 0, MAX_SUPPLY_AMOUNT);
+    usdxSupplyAmount = bound(usdxSupplyAmount, 0, MAX_SUPPLY_AMOUNT);
+    wbtcBorrowAmount = bound(wbtcBorrowAmount, 0, totalBorrowAmount);
+
+    ReserveInfoLocal memory daiInfo;
```
avniculae, line 473: can we not define these as storage in the test contract?

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1122 @@
...
+      Utils.supplyCollateral(spoke2, wethInfo.reserveId, bob, wethInfo.supplyAmount, bob);
+    }
+
+    // Bob supply dai into spoke2
+    if (daiInfo.supplyAmount > 0) {
+      Utils.supplyCollateral(spoke2, daiInfo.reserveId, bob, daiInfo.supplyAmount, bob);
+    }
+
+    // Bob supply usdx into spoke2
+    if (usdxInfo.supplyAmount > 0) {
```
avniculae, line 768: can we do this if inside the util instead?

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+
+    // Bob's debts on usdx should remain unchanged
+    (debtChecks.actualBaseDebt, debtChecks.actualPremium) = spoke1.getUserDebt(
+      usdxInfo.reserveId,
+      bob
+    );
+    assertEq(debtChecks.actualBaseDebt, bobUsdxInfo.baseDebt, 'bob usdx base debt after');
+    assertEq(debtChecks.actualPremium, bobUsdxInfo.premiumDebt, 'bob usdx premium debt after');
+
+    _verifyProtocolDebtAmounts(
```
avniculae, line 505: shall we perform this multiple times in the test? (at the different points in time)

snapshots/Spoke.Operations.json
miguelmtzinf: could you try removing the `gas_snapshot_check` config from the toml? Just confirmed it's disabled by default but who knows. Also, what about the foundry version?

src/contracts/Spoke.sol
```diff
@@ -370,6 +370,18 @@ contract Spoke is ISpoke {
...
+    uint256 reserveId,
+    address user
+  ) external view returns (uint256) {
+    require(_isBorrowing(_userPositions[user][reserveId]), UserNotBorrowingReserve(reserveId));
```
miguelmtzinf: not sure this makes sense. It must be zero if user is not borrowing right. This sounds like a condition that must be enforced at logic level rather than in a getter.I d remove it

src/contracts/Spoke.sol
miguelmtzinf: good points @avniculae 
- `getLastUserRiskPremium` or `getUserLastRiskPremium` sounds nicer to me yeah
- i'm good with keep it for better devX

src/contracts/Spoke.sol
```diff
@@ -576,6 +588,7 @@ contract Spoke is ISpoke {
     return _usingAsCollateral(userPosition) || _isBorrowing(userPosition);
   }
 
+  /// @dev If user collateral value is less than debt value, user rp calc terminates early
```
miguelmtzinf: is this correct? it's actually the opposite, goes through all the collateral reserves right

avniculae: I agree, it's the opposite

DhairyaSethi: this is trying to highlight the underwater position case, it terminates early in both cases till the min of collateral and debt value

tests/unit/Spoke/SpokeBase.t.sol
CheyenneAtapour: It's fine either way imo. The `break` skips the `idx` update at least

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: No reason to disallow that.

Also, just for example use case: If I supply usdc to borrow some usdc, I could buy another asset, e.g. BTC with the USDC, and I could be expecting BTC to rise higher than my debt grows in USDC terms, so it's netting me a profit

src/contracts/Spoke.sol
CheyenneAtapour: This is a tricky one because the idea is to use the fact that the user RP set the premium drawn shares by multiplying by base drawn shares. If the user is not borrowing a particular reserve, then this function would incorrectly return 0, even though they may be borrowing other reserves, and have a nonzero user risk premium.

The idea behind the revert was to ensure the function is called appropriately with a reserve the user is actually borrowing, and not to falsely return 0. But I'm open to ideas here. We could also loop through all reserves and find one that the user is borrowing, but that seemed potentially too tedious

src/contracts/Spoke.sol
CheyenneAtapour: "terminates early" was from the POV of the algorithm not stopping until it covered all of a user's debt. Maybe I can further clarify

tests/unit/Spoke/SpokeBase.t.sol
CheyenneAtapour: The `UserAction` more generic and can include a restore action. The `UserBorrowAction` I specifically made just for a borrow, and also just to reduce the fuzzed parameters in one of the tests where I was previously passing in `UserAction`

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: I was also thinking the name is fine as is honestly

snapshots/Spoke.Operations.json
CheyenneAtapour: We learned that running forge test is different than running with isolate, which are the gas snapshots we want to compare against. We also found that nightly build of foundry seemed to all of a sudden care about gas snapshot check, so I've explicitly marked it false now, and our ci is running good now

tests/unit/Spoke/SpokeBase.t.sol
CheyenneAtapour: I think it's fine not to do this because we've already tested this for this structure in a dedicated test inside `KeyListInMemory.sol` here: https://github.com/aave/aave-v4/blob/48e3fd5d6693d4c353e16dde93e95e3fd8a6cfcd/tests/unit/KeyListInMemory.t.sol#L17

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: I think it's fine to leave utils as they are so we can test the case of supplying, borrowing, restoring, etc. 0 using the utils and expecting a revert

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: In this test I already do it after the accrual, do you want it before the skip time as well?

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: Imo I think it's fine as is, and to not have to assume particular storage structs exist in the test contracts. But yeah that is an alternative approach

src/contracts/Spoke.sol

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: Done

src/contracts/Spoke.sol
CheyenneAtapour: I changed the comment, lmk your thoughts, or feel free to resolve

src/contracts/Spoke.sol
miguelmtzinf: I see your point. Imo, the user rp is only one for the entire user position, and the reserve does not matter.
I d go with returning what is stored for this reserve and that's all, and provide a clear @dev comment.
We could also add a `getLastUserRiskPremium(address user)`, and loop to find the first non zero rp.

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+    vars.wethSupplyAmount = 5e18; // 10_000 usd, 15 lp
+    vars.daiBorrowAmount = 10_000e18; // 10_000 usd, 20 lp
+    vars.delay = 365 days;
+
+    // Validate liquidity premiums
+    assertEq(spoke1.getLiquidityPremium(_usdxReserveId(spoke1)), 50_00, 'usdx lp');
+    assertEq(spoke1.getLiquidityPremium(_wethReserveId(spoke1)), 15_00, 'weth lp');
+    assertEq(spoke1.getLiquidityPremium(_daiReserveId(spoke1)), 20_00, 'dai lp');
+
+    // Set LT to 1 for Alice collateral
```
miguelmtzinf:
```suggestion
    // Set CF to 100.00% for Alice collateral
```

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+    updateCollateralFactor(spoke1, _wethReserveId(spoke1), 10_000);
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 10_000);
```
miguelmtzinf:
```suggestion
    updateCollateralFactor(spoke1, _wethReserveId(spoke1), 100_00);
    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 100_00);
```

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
    * +-----------+------------+------------------+--------+----------+
```
miguelmtzinf, line 75: should this be moved to the above the setUp?

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+    );
+    assertEq(
+      spoke1.getUserSuppliedAmount(_wethReserveId(spoke1), alice),
+      vars.wethSupplyAmount,
+      'supplied weth'
+    );
+
+    uint256 accruedDaiDebt = vars.daiBorrowAmount.rayMul(
+      MathUtils.calculateLinearInterest(
+        hub.getBaseInterestRate(daiAssetId), // note: IR strategy has a pending fix
```
miguelmtzinf: what's this? change to `todo` so it's easy to find

miguelmtzinf: also, we gotta start using helpers for calcExpectedDebtIndex or calcExpectedDebt.
do we have them already?

tests/unit/Spoke/Spoke.RiskPremium.t.sol
DhairyaSethi: wouldn't you be better off in that case just buying from your usdc before supplying, or am i missing something? 

one use case is leverage farming points by looping that asset

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+      aliceDaiInfo,
+      bobUsdxInfo,
+      aliceUsdxInfo,
+      'after alice repay'
+    );
+  }
+
+  /// Bob and Alice each supply and borrow varying fuzzed amounts of usdx and dai,
+  /// with different risk premiums. We check interest accrues correctly and values percolate to hub.
+  /// @dev We don't store user risk premium directly, so compare calculated premiumDrawnShares as proxy for expected risk premium
```
miguelmtzinf: this is no longer the case as we compare rp directly?

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+    (bobDaiInfo.baseDebt, ) = spoke1.getUserDebt(_daiReserveId(spoke1), bob);
+    (bobUsdxInfo.baseDebt, ) = spoke1.getUserDebt(_usdxReserveId(spoke1), bob);
+    (aliceDaiInfo.baseDebt, ) = spoke1.getUserDebt(_daiReserveId(spoke1), alice);
+    (aliceUsdxInfo.baseDebt, ) = spoke1.getUserDebt(_usdxReserveId(spoke1), alice);
+
+    // Verify final reserve states and hub propagation for both assets
+    _verifyProtocolDebtShares(bobDaiInfo, aliceDaiInfo, bobUsdxInfo, aliceUsdxInfo, 'final');
+  }
 
   function _getValueInBaseCurrency(
```
miguelmtzinf: is this duplicated? wondering if it's part of base helpers already, or can be moved there

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -176,4 +875,241 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+    action.borrowAmount = bound(action.borrowAmount, 1, action.supplyAmount / 2);
+    return action;
+  }
+
+  function _verifyProtocolDebtAmounts(
+    UserInfoLocal memory bobDaiInfo,
+    UserInfoLocal memory aliceDaiInfo,
+    UserInfoLocal memory bobUsdxInfo,
+    UserInfoLocal memory aliceUsdxInfo,
+    string memory when
```
miguelmtzinf:
```suggestion
    string memory label
```

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -176,4 +875,241 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+      string.concat('hub asset premium debt ', when)
+    );
+  }
+
+  function _verifyProtocolDebtShares(
+    UserInfoLocal memory bobDaiInfo,
+    UserInfoLocal memory aliceDaiInfo,
+    UserInfoLocal memory bobUsdxInfo,
+    UserInfoLocal memory aliceUsdxInfo,
+    string memory when
```
miguelmtzinf:
```suggestion
    string memory label
```

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -176,4 +875,241 @@ contract SpokeRiskPremiumScenarioTest is Base {
...
+
+  function _boundUserBorrowAction(
+    UserBorrowAction memory action
+  ) internal pure returns (UserBorrowAction memory) {
+    action.supplyAmount = bound(action.supplyAmount, 2, MAX_SUPPLY_AMOUNT / 2);
+    action.borrowAmount = bound(action.borrowAmount, 1, action.supplyAmount / 2);
+    return action;
+  }
+
+  function _verifyProtocolDebtAmounts(
```
miguelmtzinf, line 879: wondering if we can move these 2 to any of the base files, so can be used across test files. They d need to be more generic I guess

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -32,141 +75,797 @@ contract SpokeRiskPremiumScenarioTest is Base {
    * +-----------+------------+------------------+--------+----------+
    */
   function test_riskPremiumPropagatesCorrectly_singleBorrow() public {
```
miguelmtzinf, line 78: add description above pls

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -0,0 +1,1174 @@
...
+  }
+
+  /// With no collateral supplied, user risk premium is 0.
+  function test_getUserRiskPremium_no_collateral() public {
+    // Assert Bob has no collateral
+    for (uint256 reserveId = 0; reserveId < spoke1.reserveCount(); reserveId++) {
+      DataTypes.UserPosition memory bobInfo = getUserInfo(spoke1, bob, reserveId);
+      assertEq(bobInfo.suppliedShares, 0, 'bob supplied collateral');
+    }
+    assertEq(spoke1.getUserRiskPremium(bob), 0, 'user risk premium');
```
miguelmtzinf, line 40: do we wanna assert also that stored user rp for all reserves is 0? we can write a helper for this

miguelmtzinf: it can be used in multiple tests, to validate stored rp is consistent across all reserves or only for those that are being borrowed for example (depends on the approach we follow on`getLastUserRiskPremium`)

tests/unit/Spoke/Spoke.RiskPremium.t.sol
miguelmtzinf: yeah, historically the main use case s been position leverage.

tests/unit/Spoke/Spoke.RiskPremium.t.sol
yan-man: my bad: https://github.com/aave/aave-v4/blob/9b59a43c60b44c6c387e6e0a7d327b4c5dae667a/tests/Base.t.sol#L1016

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: There's no setUp function in this test file. I think it's good here because this is before the first test

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: should we create those helpers in this PR, or a future one?

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: This was meant for previous user risk premiums. Indeed I do this within this test case. I updated the comment to clarify

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: Good catch, duplicated indeed

tests/unit/Spoke/Spoke.RiskPremium.t.sol
CheyenneAtapour: I'm a bit hesitant to rely on the `getLastUserRiskPremium`, because as mentioned above, I prefer to use `premiumDrawnShares` as a proxy check for stored user risk premium. The reason being that the division of premium drawn shares into base drawn shares sometimes introduces precision loss, so it's not as accurate as calculating user rp and then calculating expected premium drawn shares

CheyenneAtapour:
> Thanks @CheyenneAtapour ! are there tests for the following cases?
> 
> * User rp gets back to zero after repaying all debt.
> * Collateral does not cover debt, user rp is weighted avg of all collateral assets' lp
> 
> also, refreshRp() is out of the scope right? i think it can be easily added here (we can do it in a separate pr)

The user rp going back to 0 check happens everywhere you see the function `_repayAll()`

I've now added the collateral insufficient to cover debt case

After thinking about refreshRP(), I think perhaps it should go in a different PR because it needs to handle accounting, accrual, and updating premium drawn shares for all assets appropriately. We also should have dedicated tests to ensure it works properly. Additionally, my concern with this function is that the user may end up pressing that button on ui, but their resulting user rp might actually be different than what is shown on UI due to prices changing etc before the transaction actually lands. But that is just a ui concern that we can't address on the smart contract level. Except we could maybe enforce that it must go down, or it reverts?

Also please note the additional test cases in this PR: https://github.com/aave/aave-v4/pull/367

src/contracts/Spoke.sol
CheyenneAtapour: So I have concerns about this getter's accuracy anyway, because there can be precision loss in the division of `premiumDrawnShares` by `baseDrawnShares`. I am currently only using this getter sparingly in tests when there are no actions in between or when I am confirming that user risk premium has changed. We could actually just remove it altogether

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
CheyenneAtapour: Yeah, they are not super generic because they use the structs. I'm open to ideas though

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
yan-man: can we use full "collateral factor", bc close factor for liquidation is also CF

tests/unit/Spoke/Spoke.RiskPremium.t.sol
yan-man: also update comment above

PR #326, opened 2025-04-04
fix: Add underflow protection on premium accrual calc

### LINKED ISSUE #325: Make premium accrual more conservative by adding underflow protection
We can potentially have an underflow when calculating accrued premium ( `assetsUp(premiumDrawnShares) - premiumOffset` ), due to the classic +-1 wei precision when doing conversions via ex rate.

We can make the calculation more safe if we are less strict:
```
    uint256 premiumDrawnAssets = asset.toDrawnAssetsUp(asset.premiumDrawnShares);
    uint256 accrued = premiumDrawnAssets < asset.premiumOffset
      ? 0
      : premiumDrawnAssets - asset.premiumOffset;
```

miguelmtzinf:
Closes #325

### DIFF
--- a/src/contracts/AssetLogic.sol
+++ b/src/contracts/AssetLogic.sol
@@ -7,6 +7,7 @@ import {MathUtils} from 'src/contracts/MathUtils.sol';
 import {SharesMath} from 'src/contracts/SharesMath.sol';
 import {PercentageMath} from 'src/contracts/PercentageMath.sol';
 import {WadRayMath} from 'src/contracts/WadRayMath.sol';
+import {PremiumHelper} from 'src/libraries/PremiumHelper.sol';
 
 library AssetLogic {
   using AssetLogic for DataTypes.Asset;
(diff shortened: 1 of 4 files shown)


src/contracts/AssetLogic.sol
```diff
@@ -54,9 +54,12 @@ library AssetLogic {
...
   function premiumDebt(DataTypes.Asset storage asset) internal view returns (uint256) {
-    return
-      asset.realizedPremium +
-      (asset.toDrawnAssetsUp(asset.premiumDrawnShares) - asset.premiumOffset);
+    // underflow protection: conversion of `premiumDrawnShares` into assets is accurate to 1 wei precision
+    uint256 premiumDrawnAssets = asset.toDrawnAssetsUp(asset.premiumDrawnShares);
+    uint256 accrued = premiumDrawnAssets < asset.premiumOffset
+      ? 0
+      : premiumDrawnAssets - asset.premiumOffset;
+    return asset.realizedPremium + accrued;
```
DhairyaSethi: how about moving this logic to a standard static library wherever this is used
basically:
```solidity
library PremiumHelper {
  function calcPremiumDebt(IHub hub, uint premiumDrawnShares, uint premiumOffset, uint realizedPremium) internal view {
    ...<logic with flooring only at 1 wei diff>
    }
}
```

DhairyaSethi:
dup #567

Issue #328, opened 2025-04-05
Spoke constructor validation


yan-man:
- validation of constructor params

Issue #327, opened 2025-04-05
deprecate Errors.sol


yan-man:
- fully migrate to custom errors
- only currently used by IR strategy contract

Issue #322, opened 2025-04-04
spoke liquidation config params


yan-man:
admin can define spoke-wide config params:

- define variable liquidation bonus (linearly decreasing from HF = 1 to some lower HF_f value, with a LB that maxes at HF_f and is min at HF=1)
- variable close factor

Issue #253, opened 2025-03-10
Make Oracle in Spoke reserve-specific


miguelmtzinf:
LH should not care about the oracle being used for the asset, as it only cares about accountability and IR.
It's Spoke's responsibility the oracle to use and how, so the oracle address should be linked to the `DataTypes.ReserveConfig`  struct.

yan-man:
oracle is in the spoke already but it is universal across all reserves, so we still need this to be reserve-specific

edit: nvm, will not be doing this. discussed on 4.11.25

yan-man:
discussion around keeping oracle immutable on spoke-level: https://github.com/aave/aave-v4/pull/329#discussion_r2039856901

PR #329, opened 2025-04-05
Implement Liquidation Config Params

yan-man:
- clean up library locations
- implement initial liquidation logic 
- implement liquidation config params
- validate spoke constructor params

closes https://github.com/aave/aave-v4/issues/328
closes https://github.com/aave/aave-v4/issues/330
closes https://github.com/aave/aave-v4/issues/322
closes https://github.com/aave/aave-v4/issues/327

### DIFF
--- a/src/libraries/math/PercentageMath.sol
+++ b/src/libraries/math/PercentageMath.sol
@@ -19,7 +19,7 @@ library PercentageMath {
    * @notice Executes a percentage multiplication
    * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
    * @param value The value of which the percentage needs to be calculated
-   * @param percentage The percentage of the value to be calculated
+   * @param percentage The percentage of the value to be calculated (in BPS)
    * @return result value percentmul percentage
    */
   function percentMul(uint256 value, uint256 percentage) internal pure returns (uint256 result) {
@@ -42,7 +42,7 @@ library PercentageMath {
    * @notice Executes a percentage division
    * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
    * @param value The value of which the percentage needs to be calculated
-   * @param percentage The percentage of the value to be calculated
+   * @param percentage The percentage of the value to be calculated (in BPS)
    * @return result value percentdiv percentage
    */
   function percentDiv(uint256 value, uint256 percentage) internal pure returns (uint256 result) {
--- a/src/libraries/math/PercentageMathExtended.sol
+++ b/src/libraries/math/PercentageMathExtended.sol
@@ -1,0 +1,93 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+/**
+ * @title PercentageMathExtended library
+ * @author Aave
+ * @notice Provides functions to perform percentage calculations
+ * @dev Percentages are defined by default with 2 decimals of precision (100.00). The precision is indicated by PERCENTAGE_FACTOR
+ */
+library PercentageMathExtended {
+  // Maximum percentage factor (100.00%)
+  uint256 internal constant PERCENTAGE_FACTOR = 1e4;
+
+  /**
+   * @notice Executes a percentage multiplication, rounded down
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param value The value of which the percentage needs to be calculated
+   * @param percentage The percentage of the value to be calculated
+   * @return result value percentMul percentage
+   */
+  function percentMulDown(
+    uint256 value,
+    uint256 percentage
+  ) internal pure returns (uint256 result) {
+    // to avoid overflow, value <= type(uint256).max / percentage
+    assembly ('memory-safe') {
+      if iszero(or(iszero(percentage), iszero(gt(value, div(not(0), percentage))))) {
+        revert(0, 0)
+      }
+
+      result := div(mul(value, percentage), PERCENTAGE_FACTOR)
+    }
+  }
+
+  /**
+   * @notice Executes a percentage multiplication, rounded up
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param value The value of which the percentage needs to be calculated
+   * @param percentage The percentage of the value to be calculated (in BPS)
+   * @return result value percentMul percentage
+   */
+  function percentMulUp(uint256 value, uint256 percentage) internal pure returns (uint256 result) {
+    // to avoid overflow, value <= type(uint256).max / percentage
+    assembly ('memory-safe') {
+      if iszero(or(iszero(percentage), iszero(gt(value, div(not(0), percentage))))) {
+        revert(0, 0)
+      }
+      result := mul(value, percentage)
+      // Add 1 if (value * percentage) % PERCENTAGE_FACTOR > 0 to round up the division of (value * percentage) by PERCENTAGE_FACTOR
+      result := add(div(result, PERCENTAGE_FACTOR), gt(mod(result, PERCENTAGE_FACTOR), 0))
+    }
+  }
+
+  /**
+   * @notice Executes a percentage division, rounded down
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param value The value of which the percentage needs to be calculated
+   * @param percentage The percentage of the value to be calculated (in BPS)
+   * @return result value percentDiv percentage
+   */
+  function percentDivDown(
+    uint256 value,
+    uint256 percentage
+  ) internal pure returns (uint256 result) {
+    // to avoid overflow, value <= type(uint256).max / PERCENTAGE_FACTOR
+    assembly ('memory-safe') {
+      if or(iszero(percentage), iszero(iszero(gt(value, div(not(0), PERCENTAGE_FACTOR))))) {
+        revert(0, 0)
+      }
+
+      result := div(mul(value, PERCENTAGE_FACTOR), percentage)
+    }
+  }
+
+  /**
+   * @notice Executes a percentage division, rounded up
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param value The value of which the percentage needs to be calculated
+   * @param percentage The percentage of the value to be calculated (in BPS)
+   * @return result value percentDiv percentage
+   */
+  function percentDivUp(uint256 value, uint256 percentage) internal pure returns (uint256 result) {
+    // to avoid overflow, value <= type(uint256).max / PERCENTAGE_FACTOR
+    assembly ('memory-safe') {
+      if or(iszero(percentage), iszero(iszero(gt(value, div(not(0), PERCENTAGE_FACTOR))))) {
+        revert(0, 0)
+      }
+      result := mul(value, PERCENTAGE_FACTOR)
+      // Add 1 if (value * PERCENTAGE_FACTOR) % percentage > 0 to round up the division of (value * PERCENTAGE_FACTOR) by percentage
+      result := add(div(result, percentage), gt(mod(result, percentage), 0))
+    }
+  }
+}
--- a/src/libraries/math/WadRayMathExtended.sol
+++ b/src/libraries/math/WadRayMathExtended.sol
@@ -1,0 +1,163 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+/**
+ * @title WadRayMathExtended library, with explicit rounding
+ * @author Aave
+ * @notice Provides functions to perform calculations with Wad and Ray units
+ * @dev Provides mul and div function for wads (decimal numbers with 18 digits of precision) and rays (decimal numbers
+ * with 27 digits of precision)
+ */
+
+library WadRayMathExtended {
+  uint256 internal constant WAD = 1e18;
+  uint256 internal constant RAY = 1e27;
+
+  /**
+   * @dev Multiplies two wad, rounding down
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param a Wad
+   * @param b Wad
+   * @return c = floor(a*b), in wad
+   */
+  function wadMulDown(uint256 a, uint256 b) internal pure returns (uint256 c) {
+    // to avoid overflow, a <= type(uint256).max / b
+    assembly ('memory-safe') {
+      if iszero(or(iszero(b), iszero(gt(a, div(not(0), b))))) {
+        revert(0, 0)
+      }
+
+      c := div(mul(a, b), WAD)
+    }
+  }
+
+  /**
+   * @dev Multiplies two wad, rounding up
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param a Wad
+   * @param b Wad
+   * @return c = ceil(a*b), in wad
+   */
+  function wadMulUp(uint256 a, uint256 b) internal pure returns (uint256 c) {
+    // to avoid overflow, a <= type(uint256).max / b
+    assembly ('memory-safe') {
+      if iszero(or(iszero(b), iszero(gt(a, div(not(0), b))))) {
+        revert(0, 0)
+      }
+      c := mul(a, b)
+      // Add 1 if (a * b) % WAD > 0 to round up the division of (a * b) by WAD
+      c := add(div(c, WAD), gt(mod(c, WAD), 0))
+    }
+  }
+
+  /**
+   * @dev Divides two wad, rounding down
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param a Wad
+   * @param b Wad
+   * @return c = floor(a/b), in wad
+   */
+  function wadDivDown(uint256 a, uint256 b) internal pure returns (uint256 c) {
+    // to avoid overflow, a <= type(uint256).max / WAD
+    assembly ('memory-safe') {
+      if or(iszero(b), iszero(iszero(gt(a, div(not(0), WAD))))) {
+        revert(0, 0)
+      }
+
+      c := div(mul(a, WAD), b)
+    }
+  }
+
+  /**
+   * @dev Divides two wad, rounding up
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param a Wad
+   * @param b Wad
+   * @return c = ceil(a/b), in wad
+   */
+  function wadDivUp(uint256 a, uint256 b) internal pure returns (uint256 c) {
+    // to avoid overflow, a <= type(uint256).max / WAD
+    assembly ('memory-safe') {
+      if or(iszero(b), iszero(iszero(gt(a, div(not(0), WAD))))) {
+        revert(0, 0)
+      }
+      c := mul(a, WAD)
+      // Add 1 if (a * WAD) % b > 0 to round up the division of (a * WAD) by b
+      c := add(div(c, b), gt(mod(c, b), 0))
+    }
+  }
+
+  /**
+   * @dev Multiplies two ray, rounding down
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param a Ray
+   * @param b Ray
+   * @return c = floor(a*b), in ray
+   */
+  function rayMulDown(uint256 a, uint256 b) internal pure returns (uint256 c) {
+    // to avoid overflow, a <= type(uint256).max / b
+    assembly ('memory-safe') {
+      if iszero(or(iszero(b), iszero(gt(a, div(not(0), b))))) {
+        revert(0, 0)
+      }
+
+      c := div(mul(a, b), RAY)
+    }
+  }
+
+  /**
+   * @dev Multiplies two ray, rounding up
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param a Ray
+   * @param b Ray
+   * @return c = ceil(a*b), in ray
+   */
+  function rayMulUp(uint256 a, uint256 b) internal pure returns (uint256 c) {
+    // to avoid overflow, a <= type(uint256).max / b
+    assembly ('memory-safe') {
+      if iszero(or(iszero(b), iszero(gt(a, div(not(0), b))))) {
+        revert(0, 0)
+      }
+      c := mul(a, b)
+      // Add 1 if (a * b) % RAY > 0 to round up the division of (a * b) by RAY
+      c := add(div(c, RAY), gt(mod(c, RAY), 0))
+    }
+  }
+
+  /**
+   * @dev Divides two ray, rounding down
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param a Ray
+   * @param b Ray
+   * @return c = floor(a/b), in ray
+   */
+  function rayDivDown(uint256 a, uint256 b) internal pure returns (uint256 c) {
+    // to avoid overflow, a <= type(uint256).max / RAY
+    assembly ('memory-safe') {
+      if or(iszero(b), iszero(iszero(gt(a, div(not(0), RAY))))) {
+        revert(0, 0)
+      }
+
+      c := div(mul(a, RAY), b)
+    }
+  }
+
+  /**
+   * @dev Divides two ray, rounding up
+   * @dev assembly optimized for improved gas savings, see https://twitter.com/transmissions11/status/1451131036377571328
+   * @param a Ray
+   * @param b Ray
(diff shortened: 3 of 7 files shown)


tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -7,6 +7,17 @@ contract SpokeConfigTest is SpokeBase {
   using SharesMath for uint256;
   using WadRayMath for uint256;
 
+  function test_spoke_deploy_revertsWith_InvalidHubAddress() public {
+    vm.expectRevert(ISpoke.InvalidHubAddress.selector);
+    ISpoke(new Spoke(address(0), HEALTH_FACTOR_LIQUIDATION_THRESHOLD));
```
DhairyaSethi:
```suggestion
    new Spoke(address(0), HEALTH_FACTOR_LIQUIDATION_THRESHOLD);
```

src/contracts/Spoke.sol
```diff
@@ -1,40 +1,63 @@
...
 
   mapping(address user => mapping(uint256 reserveId => DataTypes.UserPosition position))
     internal _userPositions;
   mapping(uint256 reserveId => DataTypes.Reserve reserveData) internal _reserves;
 
+  DataTypes.VariableLiquidationBonusConfig internal _variableLiquidationBonusConfig;
   uint256[] public reservesList; // todo: rm, not needed
   uint256 public reserveCount;
-  uint256 public constant HEALTH_FACTOR_LIQUIDATION_THRESHOLD = WadRayMath.WAD; // todo configurable
+  uint256 public closeFactor;
```
DhairyaSethi: can we move this and liq bonus config to an independent spokeConfig struct

src/contracts/Spoke.sol
```diff
@@ -1,40 +1,63 @@
...
 
   // /////
   // Governance
   // /////
 
+  function updateCloseFactor(uint256 newCloseFactor) external {
+    // TODO: AccessControl
+    _validateCloseFactor(newCloseFactor);
+    closeFactor = newCloseFactor;
+    emit CloseFactorUpdated(newCloseFactor);
```
DhairyaSethi: let's also emit old value here

src/contracts/Spoke.sol
```diff
@@ -1,40 +1,63 @@
...
+  }
+
+  function updateVariableLiquidationBonusConfig(
+    DataTypes.VariableLiquidationBonusConfig calldata variableLiquidationBonusConfig
+  ) external {
+    _validateVariableLiquidationBonusConfig(variableLiquidationBonusConfig);
+    _variableLiquidationBonusConfig = variableLiquidationBonusConfig;
+    emit VariableLiquidationBonusConfigUpdated(variableLiquidationBonusConfig);
```
DhairyaSethi: same

src/contracts/Spoke.sol
```diff
@@ -508,9 +549,10 @@ contract Spoke is ISpoke {
         }
         continue;
       }
-      vars.assetId = _reserves[vars.reserveId].assetId;
+      DataTypes.Reserve memory reserve = _reserves[vars.reserveId];
+      vars.assetId = reserve.assetId;
 
-      vars.assetPrice = oracle.getAssetPrice(vars.assetId);
+      vars.assetPrice = reserve.config.oracle.getAssetPrice(vars.assetId);
```
DhairyaSethi: still using asset id for oracle, it's not truly reserve specific this way
if reserve specific then no asset param would be required

src/libraries/helpers/Errors.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.0;
 
+// TODO: delete file, use custom errors
```
DhairyaSethi: fine to delete now itself, since all custom errors are defined on interfaces now

src/contracts/Spoke.sol
yan-man: cc @miguelmtzinf that's the intention then right? completely reserve-specific oracle, ie 1 oracle per reserve/asset, no asset param

src/contracts/Spoke.sol
yan-man: but to confirm, on other "update" event emissions we don't emit old values, like spoke config, asset config, etc. We should be consistent, or do you want to also emit old values there too? or do you see these 2 somehow different?

src/contracts/Spoke.sol
yan-man: will consolidate these 2 update helpers under a single `updateLiquidationConfig` given your comment above to have a separate struct to house both of these

src/libraries/helpers/Errors.sol
yan-man: no, bc this is still used under DefaultReserveInterestRateStrategy, which we didn't update yet to custom errors bc we'll move to V2 first. Will clarify in comments

DhairyaSethi: it's fine just convert those to custom errors as well, not worth leaving this file around just for a contract which is regardless going to be updated

src/contracts/Spoke.sol
DhairyaSethi: moving to 1 struct as described [here](https://github.com/aave/aave-v4/pull/329#discussion_r2030289050) should be fine

yan-man: but are you saying you still want to emit old values too, or not? I did not in the latest changes: https://github.com/aave/aave-v4/pull/329/commits/119787e5284335b1b7fd0ba355a1e714dfac0dd9

DhairyaSethi: if having individual setters for stuff (against this inherently bad) then yes, if one setter for the one struct w all config values then dealers choice (bc emitting both old & new struct is very expensive w duplication) 

the idea is indexers can better track changes if they don't have to maintain config values but it's ok if we make them do it

src/contracts/Spoke.sol
DhairyaSethi: imo don't see the need for reserve specific in the first place if oracle is an immutable on a spoke. if spoke needs a different feed for an asset, it'd likely be on a custom spoke anyway. seems like unnecessary gas expense. will wait for @miguelmtzinf's thoughts

src/libraries/helpers/Errors.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/119787e5284335b1b7fd0ba355a1e714dfac0dd9

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/119787e5284335b1b7fd0ba355a1e714dfac0dd9

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/119787e5284335b1b7fd0ba355a1e714dfac0dd9

tests/unit/Spoke/Spoke.Config.t.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/119787e5284335b1b7fd0ba355a1e714dfac0dd9

src/libraries/types/DataTypes.sol
```diff
@@ -106,4 +107,14 @@ library DataTypes {
     uint256 userRiskPremium;
     uint256 healthFactor;
   }
+
+  struct LiquidationConfig {
+    uint256 closeFactor; // BPS, HF value to restore to during a liquidation
+    VariableLiquidationBonusConfig variableLiquidationBonusConfig;
```
DhairyaSethi:
```suggestion
    VariableLiquidationBonusConfig variableLiquidationBonus;
```
also, let's think of a shorter name haha

src/contracts/Spoke.sol
```diff
@@ -344,6 +353,27 @@ contract Spoke is ISpoke {
...
+    uint256 reserveId,
+    uint256 healthFactor
+  ) public view returns (uint256) {
+    uint256 liquidationBonus = _reserves[reserveId].config.liquidationBonus;
+    // if healthFactorBonusThreshold == 0, always return base liquidationBonus
+    if (_liquidationConfig.variableLiquidationBonusConfig.healthFactorBonusThreshold == 0) {
+      return liquidationBonus;
+    }
+    return
+      _liquidationConfig.variableLiquidationBonusConfig.calculateVariableLiquidationBonus(
```
```suggestion
      _liquidationConfig.variableLiquidationBonus.get(
```
or .calculate(

src/interfaces/ISpoke.sol
```diff
@@ -34,6 +34,8 @@ interface ISpoke {
     int256 premiumOffsetDelta,
     int256 realizedPremiumDelta
   );
+  event OracleUpdated(uint256 indexed reserveId, address oracle);
```
```suggestion
  event OracleUpdated(uint256 indexed reserveId, address indexed oracle);
```
second one can be indexed too

src/libraries/types/DataTypes.sol
```diff
@@ -106,4 +107,14 @@ library DataTypes {
     uint256 userRiskPremium;
     uint256 healthFactor;
   }
+
+  struct LiquidationConfig {
+    uint256 closeFactor; // BPS, HF value to restore to during a liquidation
```
miguelmtzinf: is not this per asset? Depending on asset's volatility we can make this more or less demanding

src/libraries/types/DataTypes.sol
```diff
@@ -106,4 +107,14 @@ library DataTypes {
...
+    VariableLiquidationBonusConfig variableLiquidationBonusConfig;
+  }
+
+  struct VariableLiquidationBonusConfig {
```
miguelmtzinf: why do we need a sub-struct here?

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -0,0 +1,35 @@
...
+library LiquidationLogic {
+  using PercentageMath for uint256;
+
+  function calculateVariableLiquidationBonus(
+    DataTypes.VariableLiquidationBonusConfig storage config,
+    uint256 healthFactor,
+    uint256 healthFactorLiquidationThreshold,
+    uint256 liquidationBonus
+  ) internal view returns (uint256) {
+    // if HF <= healthFactorBonusThreshold, liquidation bonus is max
```
miguelmtzinf: is there a reason for having `liquidationBonus` as max not the min?

miguelmtzinf: downside of this approach is that if `liquidationBonusFactor` is wrong, liqudation won't be okay.
it could happen that `liquidationBonus` itself is not correctly set as well, but less likely as we are more used to that term

src/libraries/types/DataTypes.sol
```diff
@@ -76,6 +76,7 @@ library DataTypes {
     uint256 collateralFactor; // BPS
     uint256 liquidationBonus; // TODO: liquidationProtocolFee
     uint256 liquidityPremium; // BPS
+    IPriceOracle oracle;
```
miguelmtzinf: if we go with a single contract for all addresses a la V3 AaveOracle we don't need this here and can be same for all assets

src/contracts/Spoke.sol
```diff
@@ -1,46 +1,62 @@
...
 contract Spoke is ISpoke {
   using WadRayMath for uint256;
   using PercentageMath for uint256;
   using KeyValueListInMemory for KeyValueListInMemory.List;
+  using LiquidationLogic for DataTypes.VariableLiquidationBonusConfig;
 
-  // todo capitalize, oracle should be mutable?
-  ILiquidityHub public immutable hub;
-  IPriceOracle public immutable oracle;
+  uint256 public constant HEALTH_FACTOR_LIQUIDATION_THRESHOLD = WadRayMath.WAD; // todo configurable?
```
miguelmtzinf: don't we need this configurable for variable close factor? per asset?

src/libraries/types/DataTypes.sol
yan-man: when I spoke to Emilio he mentioned this should be protocol-wide instead of being more granular per asset. Emilio confirmed today this is the case, and we should update collateralFactor instead if we want to adjust per asset. We can update later if we want to change it

src/libraries/types/DataTypes.sol
yan-man: edit: collapsed into single struct

fixed: https://github.com/aave/aave-v4/pull/329/commits/b03b7730f5b742fd08b466c0be3d34a1df65c7af

src/contracts/Spoke.sol
yan-man: this is just the HF at which the position becomes liquidatable, whereas close factor is the HF that the liquidation restores to. 

Emilio confirmed on tg that this can be kept non-configurable and set to 1

src/libraries/logic/LiquidationLogic.sol
yan-man: > is there a reason for having liquidationBonus as max not the min?

prefer to have `liqBonus` define max instead of min bc we do `percentMul(liquidationBonusFactor)` to calc min, so it is easier to bound minLiqBonus to be <= 100% of `liqBonus`. But if `liqBonus` defines the min, then max liqBonus is `liqBonus.percentDiv(liquidationBonusFactor)` which is a bit less intuitive to reason about, and harder to bound max via `liquidationBonusFactor`

but you make a good point about risk of setting `liquidationBonusFactor` wrong.

and another thing to consider is the default value if variable bonus is disabled, which would be the liquidationBonus, also might be safer as min instead of max

src/libraries/types/DataTypes.sol
yan-man: purposefully made this change according to this ticket: https://github.com/aave/aave-v4/issues/253#issue-2907705452

should I revert back to original (which was a single contract for all reserves)?

src/interfaces/ISpoke.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/b1d07dc3e6dc02f06aeb69713f2f2f20fea1b1fe

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/b1d07dc3e6dc02f06aeb69713f2f2f20fea1b1fe

src/libraries/types/DataTypes.sol
yan-man: went with liqBonusConfig

https://github.com/aave/aave-v4/pull/329/commits/b1d07dc3e6dc02f06aeb69713f2f2f20fea1b1fe

src/contracts/Spoke.sol
```diff
@@ -1,46 +1,62 @@
...
   uint256[] public reservesList; // todo: rm, not needed
   uint256 public reserveCount;
-  uint256 public constant HEALTH_FACTOR_LIQUIDATION_THRESHOLD = WadRayMath.WAD; // todo configurable
 
-  constructor(address hubAddress, address oracleAddress) {
-    hub = ILiquidityHub(hubAddress);
-    oracle = IPriceOracle(oracleAddress);
+  constructor(address hubAddress, uint256 closeFactorValue) {
+    require(hubAddress != address(0), InvalidHubAddress());
+    // close factor is required, but variable liquidation bonus config is not
```
yan-man, line 36: another point is this - want to be able to allow variable liquidation bonus to be switched on/off, I did it by not enforcing the liq bonus config being set. And here: https://github.com/aave/aave-v4/pull/329/files#diff-15083d6ab99ac85a56f41ffa1c43ffcdfe7f2ba0265ce7d47a68a02d30d10978R378, if the value is 0, just return the liq bonus 

Open to other thoughts, ie to explicitly turn it on/off, etc

cc @miguelmtzinf @DhairyaSethi @CheyenneAtapour

src/contracts/Spoke.sol
yan-man: btw I did it based on our London discussions that Emilio brought up: https://github.com/aave/aave-v4/issues/253

src/libraries/types/DataTypes.sol
```diff
@@ -76,6 +76,7 @@ library DataTypes {
     uint256 collateralFactor; // BPS
     uint256 liquidationBonus; // TODO: liquidationProtocolFee
```
yan-man: another thing @miguelmtzinf - afaik the liquidation bonus in v3 is a BPS on top of 100%, ie `105_00` represents a 5% LB. It's a bit weird bc it seems like the only config represented this way. Is there a particular reason for this approach and would you like to keep it here (in that case I need to make changes)?

cc @DhairyaSethi @CheyenneAtapour

miguelmtzinf: let's confirm with Emilio but afaik it was done just as a way to save computation (instead of doing 10000 + bonus, we just do bonus).
better to be consistent across config params in my opinion and don't follow that road imho.

yan-man: made changes to add the 100_00 after the fact: https://github.com/aave/aave-v4/pull/329/commits/f5b46008224885edd952bb49b0377bd9697f8b86

yan-man: edit: Emilio mentioned this on 4.10.25 call and wants to keep it as is, as a 105_00 in this case 

fixed: https://github.com/aave/aave-v4/pull/329/commits/b03b7730f5b742fd08b466c0be3d34a1df65c7af

src/contracts/DefaultReserveInterestRateStrategy.sol
```diff
@@ -1,12 +1,12 @@
...
-import {Errors} from '../libraries/helpers/Errors.sol';
-import {IDefaultInterestRateStrategy} from '../interfaces/IDefaultInterestRateStrategy.sol';
-import {IReserveInterestRateStrategy} from '../interfaces/IReserveInterestRateStrategy.sol';
-import {WadRayMath} from './WadRayMath.sol';
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
+import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+import {Errors} from 'src/libraries/helpers/Errors.sol';
+import {IDefaultInterestRateStrategy} from 'src/interfaces/IDefaultInterestRateStrategy.sol';
+import {IReserveInterestRateStrategy} from 'src/interfaces/IReserveInterestRateStrategy.sol';
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
```
CheyenneAtapour: `WadRayMath` library listed twice it seems

src/libraries/types/DataTypes.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.10;
 
 import {IReserveInterestRateStrategy} from 'src/interfaces/IReserveInterestRateStrategy.sol';
-
+import {IPriceOracle} from 'src/interfaces/IPriceOracle.sol';
```
CheyenneAtapour: Very small nit: White space line here after import statements and before library declaration

tests/Base.t.sol
```diff
@@ -6,10 +6,10 @@ import {console2 as console} from 'forge-std/console2.sol';
...
-import {PercentageMath} from 'src/contracts/PercentageMath.sol';
-import {WadRayMath} from 'src/contracts/WadRayMath.sol';
-import {SharesMath} from 'src/contracts/SharesMath.sol';
-import {MathUtils} from 'src/contracts/MathUtils.sol';
+import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
+import {SharesMath} from 'src/libraries/math/SharesMath.sol';
+import {MathUtils} from 'src/libraries/math/MathUtils.sol';
 import {DefaultReserveInterestRateStrategy, IDefaultInterestRateStrategy, IReserveInterestRateStrategy} from 'src/contracts/DefaultReserveInterestRateStrategy.sol';
 import {ISpoke} from 'src/interfaces/ISpoke.sol';
```
CheyenneAtapour: Btw, we don't need this import, it gets imported already on line 8, if you could please remove (ik you didn't touch it originally)

tests/Base.t.sol
```diff
@@ -268,43 +272,47 @@ abstract contract Base is Test {
       frozen: false,
       paused: false,
       collateralFactor: 80_00,
-      liquidationBonus: 0,
+      liquidationBonus: 10_000,
```
CheyenneAtapour: I think this was a typo, and should be `10_00`, right?

tests/unit/Spoke/Spoke.Getters.t.sol
```diff
@@ -0,0 +1,81 @@
...
+        _config,
+        healthFactor,
+        spoke1.getReserve(reserveId).config.liquidationBonus,
+        spoke1.HEALTH_FACTOR_LIQUIDATION_THRESHOLD()
+      ),
+      'calc should match'
+    );
+  }
+
+  function test_getVariableLiquidationBonus_configured() public {
```
CheyenneAtapour, line 39: I guess it's fine to show this exact case, but it should get covered in the fuzz test it calls anyway

tests/unit/libraries/LiquidationLogic.t.sol
```diff
@@ -0,0 +1,247 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {LiquidationLogic} from 'src/libraries/logic/LiquidationLogic.sol';
+import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
+import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
+import {Base} from 'tests/Base.t.sol';
+import {console2 as console} from 'forge-std/console2.sol';
```
CheyenneAtapour: Can remove this console import now cause unused

src/contracts/Spoke.sol
CheyenneAtapour: You can remove the `todo configurable?` comment because of this

src/contracts/Spoke.sol
```diff
@@ -746,4 +773,22 @@ contract Spoke is ISpoke {
...
+      InvalidLiquidationBonusFactor()
+    );
+    // if healthFactorBonusThreshold == HEALTH_FACTOR_LIQUIDATION_THRESHOLD, then calculate will be undefined
+    require(
+      config.healthFactorBonusThreshold < HEALTH_FACTOR_LIQUIDATION_THRESHOLD,
+      InvalidHealthFactorBonusThreshold()
+    );
+  }
+
+  function _validateCloseFactor(uint256 closeFactor) internal view {
```
CheyenneAtapour, line 791: imo don't need this function and can just do `require` statement within the `_validateLiquidationConfig` function, but nbd

tests/Base.t.sol
yan-man: actually just wrote it wrong but should be 100_00, now liq bonus has to start at 100_00 to be 0% LB, so that's the min value allowed. Good catch will fix

src/contracts/Spoke.sol
yan-man: so the reason I need this one is that in the constructor it is requiring a close factor but not the variable liquidation config, so that's why I split it out so that only close factor can be specifically validated. Thoughts?

tests/unit/Spoke/Spoke.Getters.t.sol
yan-man: yeah kind of like just having a specific case as well as the fuzz

tests/unit/libraries/LiquidationLogic.t.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/ab22070c68fa10cf7695d81610549470ce6d4771

tests/Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/ab22070c68fa10cf7695d81610549470ce6d4771

tests/Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/ab22070c68fa10cf7695d81610549470ce6d4771

src/libraries/types/DataTypes.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/ab22070c68fa10cf7695d81610549470ce6d4771

src/contracts/DefaultReserveInterestRateStrategy.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/ab22070c68fa10cf7695d81610549470ce6d4771

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/329/commits/498f23061848c46b2803ba2f3868b71991b9b148

src/contracts/Spoke.sol
yan-man: decided we will not do reserve-specific oracle, reverting changes and will close ticket

https://github.com/aave/aave-v4/pull/329/commits/c01f96549cc81dd362c37255679b685cec5e6e50

src/libraries/types/DataTypes.sol
miguelmtzinf: yes, ty

PR #334, opened 2025-04-08
fix: Fix rounding issue on repay

### LINKED ISSUE #333: Hub::repay should round down input assets to shares to ensure that borrowers do not underpay their obligations.
https://github.com/aave/aave-v4/blob/d7812d541152d82f13f00f0b8f72f9e01fe2c873/src/contracts/LiquidityHub.sol#L210

Current code allows borrowers to repay 1 share of debt with less of 1 share in asset terms.
Example:
- ex rate is 2 assets : 1 share
- borrow repay 1 wei of asset
- 1 wei of share is repaid (when 1 share actually worth 2 wei of assets)

miguelmtzinf:
Closes #333

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -32,7 +32,7 @@ contract LiquidityHub is ILiquidityHub {
   // /////
 
   function addAsset(DataTypes.AssetConfig calldata config, address asset) external {
-    // TODO: AccessControl
+    // TODO: AccessControl, prevent dup entry
     _validateAssetConfig(config, asset);
     assetsList.push(IERC20(asset));
     uint256 assetId = assetCount++;


src/contracts/Spoke.sol
```diff
@@ -135,11 +135,20 @@ contract Spoke is ISpoke {
...
+    userPosition.premiumDrawnShares = 0;
+    userPosition.premiumOffset = 0;
+    userPosition.realizedPremium += accruedPremium;
 
+    _refreshPremiumDebt(
+      reserve,
+      -int256(userPremiumDrawnShares),
+      -int256(userPremiumOffset),
+      int256(accruedPremium)
+    ); // unnecessary but we settle premium debt here
```
miguelmtzinf, line 151: do we wanna extract an internal fn for this now? or later
can be used in repay too if we add a param

DhairyaSethi: let's do that later imo, i would like to remove this from  borrow/withdraw once we have a fundamental fix

PR #338, opened 2025-04-08
Allow Repay Greater Than Max Debt

### LINKED ISSUE #339: Spoke Repay Greater Than Current Debt
The consensus was mixed (https://github.com/aave/aave-v4/pull/311#discussion_r2033855099) as to what should happen when the Spoke's repay function is called with a value greater than the user's total debt.

We can either

Revert
Just repay the max debt amount
This PR implements option 2.

Whichever we choose, we should also make the withdraw function behave similarly (and create a corresponding issue)

CheyenneAtapour:
The consensus was mixed (https://github.com/aave/aave-v4/pull/311#discussion_r2033855099) as to what should happen when the Spoke's `repay` function is called with a value greater than the user's total debt. 

We can either

1. Revert
2. Just repay the max debt amount

This PR implements option 2. 

Whichever we choose, we should also make the withdraw function behave similarly (and create a corresponding issue)

Closes #339

### DIFF
--- a/tests/unit/Spoke/Spoke.Repay.t.sol
+++ b/tests/unit/Spoke/Spoke.Repay.t.sol
@@ -119,14 +119,14 @@ contract SpokeRepayTest is SpokeBase {
     uint256 daiRepayAmount = daiSupplyAmount / 4;
 
     // Bob supply weth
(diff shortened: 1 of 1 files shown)


DhairyaSethi:
this was added in #343

PR #341, opened 2025-04-09
Migrate Spoke Withdraw Tests

### LINKED ISSUE #288: Spoke Tests: Withdraw Refactor
i've already begun refactoring

CheyenneAtapour:
Closes #288


tests/unit/Spoke/Spoke.Withdraw.Validation.t.sol
```diff
@@ -0,0 +1,1543 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+contract SpokeWithdrawValidationTest is SpokeBase {
+  using WadRayMath for uint256;
+  using PercentageMath for uint256;
```
yan-man: can remove this lib

tests/unit/Spoke/Spoke.Withdraw.Validation.t.sol
```diff
@@ -0,0 +1,1543 @@
...
+    // newWithdrawalLimit with accrued interest should be greater than supplyAmount
+    vm.assume(newWithdrawalLimit > supplyAmount);
+
+    vm.expectRevert(abi.encodeWithSelector(ISpoke.InsufficientSupply.selector, newWithdrawalLimit));
+    vm.prank(alice);
+    spoke1.withdraw({reserveId: reserveId, amount: newWithdrawalLimit + 1, to: alice});
+  }
+
+  /// @dev cannot withdraw an amount if resulting withdrawal would result in HF < threshold
+  function test_withdraw_revertsWith_HealthFactorBelowThreshold_singleBorrow() public {
```
yan-man: i think we can move all these HF-related ones to another `Spoke.Withdraw.HealthFactor.t.sol` file

tests/unit/Spoke/Spoke.Withdraw.t.sol
```diff
@@ -6,184 +6,48 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
 contract SpokeWithdrawTest is SpokeBase {
   using WadRayMath for uint256;
   using PercentageMath for uint256;
```
yan-man: can also rm this lib

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
```diff
@@ -0,0 +1,261 @@
...
+    assertEq(
+      expectedSupplied,
+      spoke1.getUserSuppliedAmount(_daiReserveId(spoke1), bob),
+      'expected supplied'
+    );
+
+    // Withdraw all supplied assets
+    spoke1.withdraw(_daiReserveId(spoke1), type(uint256).max, bob);
+
+    _checkSuppliedAmounts(daiAssetId, _daiReserveId(spoke1), spoke1, bob, 0, 'after withdraw');
```
DhairyaSethi, line 116: can we also assert in these tests that supply ex rate is monotonically increasing only?

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
```diff
@@ -0,0 +1,261 @@
...
+      state.asset.balanceOf(alice),
+      MAX_SUPPLY_AMOUNT - params.aliceAmount + aliceData[0].suppliedAmount,
+      'alice balance'
+    );
+    assertEq(
+      state.asset.balanceOf(bob),
+      MAX_SUPPLY_AMOUNT - params.bobAmount + bobData[1].suppliedAmount,
+      'bob balance'
+    );
+  }
```
DhairyaSethi, line 298: can we also assert the two round trip properties mentioned here for deposit & withdraw https://github.com/a16z/erc4626-tests/blob/main/ERC4626.prop.sol

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
CheyenneAtapour: I don't believe we have this property yet without inflation protection because we can have the case that everyone has withdrawn, and thus the exchange rate goes back to 0.

So perhaps we wait on this until we have inflation protection implemented

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
CheyenneAtapour: Added

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
DhairyaSethi: imo let's still add it in the tests and comment it the internal function being called for now?

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
```diff
@@ -0,0 +1,372 @@
...
+    // Withdraw and confirm share amount from event emission
+    uint256 shares2 = hub.convertToSuppliedShares(reserve.assetId, assets);
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.Withdraw(reserveId, caller, shares2, caller);
+    vm.prank(caller);
+    spoke1.withdraw(reserveId, assets, caller);
+
+    // Check for delta of 0
+    uint256 delta = 0;
+    assertApproxGeAbs(shares2, shares1, delta);
```
DhairyaSethi:
```suggestion
    assertGe(shares2, shares1);
```
looks like we don't need that helper, let's rm it

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
CheyenneAtapour: Done, I also just went with `assertEq` since delta is 0

PR #343, opened 2025-04-13
rft: Debt in Scaled Balances

### LINKED ISSUE #319: Fix Repay Underflow
We want to always draw and restore in share amounts. This has a few consequences:

-We need to force a user to repay a full debt share if their debt amount is less than 1 share. We need to handle subtracting their debt appropriately and not cause an underflow
-We need to adjust our `_calculateRestore` function to allow for this

-Similarly, we need to force a user to only be able to draw in debt share amounts, so whenever they draw an amount, we must draw an amount of shares that sufficiently covers this debt (rounding up)

DhairyaSethi:
- refactor debt accounting to use scaled balances, an effort to mathematically ensure debt exchange ratio never decreases & close #319 underflow bug due to premium accounting
- introduce new `convertToPremiumDrawnShares` conversion getter with a different rounding for premium debt
- allow repay full by passing any amount greater than current debt (TODO: same behaviour on withdraw)
- refactor prototype to new accounting
- build scenario engine & dependency-free test framework for prototype
- refactor & clean repay tests

### DIFF
--- a/tests/unit/Spoke/Spoke.Repay.t.sol
+++ b/tests/unit/Spoke/Spoke.Repay.t.sol
@@ -3334,6 +3334,7 @@ contract SpokeRepayTest is SpokeBase {
 
   /// repay partial premium, base & full debt, with no interest accrual (no time pass)
   /// supply ex rate can increase while debt ex rate should remain the same
+  /// this is due to donation on available liquidity
   function test_fuzz_repay_effect_on_ex_rates(uint256 daiBorrowAmount) public {
     daiBorrowAmount = bound(daiBorrowAmount, 1, MAX_SUPPLY_AMOUNT / 2);
     uint256 wethSupplyAmount = _calcMinimumCollAmount(
@@ -3351,11 +3352,10 @@ contract SpokeRepayTest is SpokeBase {
     skip(365 days); // initial increase in index, no time passes for subsequent checks
 
     Debts memory bobDebt = getUserDebt(spoke1, bob, _daiReserveId(spoke1));
-    uint256 supplyExchangeRatioBefore = hub.convertToSuppliedAssets(daiAssetId, 1e30);
-    uint256 debtExchangeRatio = hub.convertToDrawnAssets(daiAssetId, 1e30);
+    uint256 supplyExchangeRatioBefore = hub.convertToSuppliedAssets(daiAssetId, MAX_SUPPLY_AMOUNT);
+    uint256 debtExchangeRatio = hub.convertToDrawnAssets(daiAssetId, MAX_SUPPLY_AMOUNT);
 
     // repay partial premium debt
-    spoke1.getUserPosition(_daiReserveId(spoke1), bob);
     vm.assume(bobDebt.premiumDebt > 1);
     uint256 daiRepayAmount = bound(vm.randomUint(), 1, bobDebt.premiumDebt - 1);
 
@@ -3364,9 +3364,9 @@ contract SpokeRepayTest is SpokeBase {
     vm.prank(bob);
     spoke1.repay(_daiReserveId(spoke1), daiRepayAmount);
 
-    uint256 supplyExchangeRatioAfter = hub.convertToSuppliedAssets(daiAssetId, 1e30);
+    uint256 supplyExchangeRatioAfter = hub.convertToSuppliedAssets(daiAssetId, MAX_SUPPLY_AMOUNT);
     assertGe(supplyExchangeRatioAfter, supplyExchangeRatioBefore);
-    assertEq(hub.convertToDrawnAssets(daiAssetId, 1e30), debtExchangeRatio);
+    assertEq(hub.convertToDrawnAssets(daiAssetId, MAX_SUPPLY_AMOUNT), debtExchangeRatio);
 
     bobDebt = getUserDebt(spoke1, bob, _daiReserveId(spoke1));
     // repay partial base debt
@@ -3376,19 +3376,18 @@ contract SpokeRepayTest is SpokeBase {
     vm.prank(bob);
     spoke1.repay(_daiReserveId(spoke1), daiRepayAmount);
 
-    supplyExchangeRatioAfter = hub.convertToSuppliedAssets(daiAssetId, 1e30);
+    supplyExchangeRatioAfter = hub.convertToSuppliedAssets(daiAssetId, MAX_SUPPLY_AMOUNT);
     assertGe(supplyExchangeRatioAfter, supplyExchangeRatioBefore);
-    assertEq(hub.convertToDrawnAssets(daiAssetId, 1e30), debtExchangeRatio);
+    assertEq(hub.convertToDrawnAssets(daiAssetId, MAX_SUPPLY_AMOUNT), debtExchangeRatio);
 
     supplyExchangeRatioBefore = supplyExchangeRatioAfter;
 
     vm.prank(bob);
     spoke1.repay(_daiReserveId(spoke1), type(uint256).max);
 
-    supplyExchangeRatioAfter = hub.convertToSuppliedAssets(daiAssetId, 1e30);
+    supplyExchangeRatioAfter = hub.convertToSuppliedAssets(daiAssetId, MAX_SUPPLY_AMOUNT);
     assertGe(supplyExchangeRatioAfter, supplyExchangeRatioBefore);
-    supplyExchangeRatioBefore = supplyExchangeRatioAfter;
-    assertEq(hub.convertToDrawnAssets(daiAssetId, 1e30), debtExchangeRatio);
+    assertEq(hub.convertToDrawnAssets(daiAssetId, MAX_SUPPLY_AMOUNT), debtExchangeRatio);
   }
 
   function _assertUserRpUnchanged(uint256 reserveId, ISpoke spoke, address user) internal view {
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -118,7 +118,7 @@ contract LiquidityHub is ILiquidityHub {
 
     asset.updateBorrowRate({liquidityAdded: amount, liquidityTaken: 0});
 
-    // todo: Mitigate inflation attack (burn some amount if first supply)
+    // todo: Mitigate inflation attack
     uint256 suppliedShares = asset.toSuppliedSharesDown(amount);
     require(suppliedShares != 0, InvalidSharesAmount());
 
--- a/tests/unit/Spoke/Spoke.Supply.t.sol
+++ b/tests/unit/Spoke/Spoke.Supply.t.sol
@@ -583,4 +583,44 @@ contract SpokeSupplyTest is SpokeBase {
       'user suppliedShares after-supply'
     );
   }
+
+  /// supply an asset with existing debt, with no interest accrual the two ex rates
+  /// can increase due to rounding, with interest accrual should strictly increase
+  function test_fuzz_supply_effect_on_ex_rates(uint256 amount, uint256 delay) public {
+    delay = bound(delay, 1, MAX_SKIP_TIME);
+    amount = bound(amount, 1, MAX_SUPPLY_AMOUNT / 2);
+    uint256 wethSupplyAmount = _calcMinimumCollAmount(
+      spoke1,
+      _wethReserveId(spoke1),
+      _daiReserveId(spoke1),
+      amount
+    );
+    Utils.supply(spoke1, _daiReserveId(spoke1), bob, amount, bob);
+    Utils.supply(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob); // bob collateral
+    setUsingAsCollateral(spoke1, bob, _wethReserveId(spoke1), true);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, amount, bob); // introduce debt
+
+    uint256 supplyExchangeRatio = hub.convertToSuppliedAssets(daiAssetId, MAX_SUPPLY_AMOUNT);
+    uint256 debtExchangeRatio = hub.convertToDrawnAssets(daiAssetId, MAX_SUPPLY_AMOUNT);
+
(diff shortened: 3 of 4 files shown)


foundry.toml
```diff
@@ -10,7 +10,7 @@ evm_version = "cancun"
 optimizer = true
 optimizer_runs = 200
 bytecode_hash = "none"
-isolate = true
+# isolate = true
```
CheyenneAtapour: You didn't mean to push this, right? If not, just pointing out to revert

src/contracts/Spoke.sol
```diff
@@ -686,20 +687,27 @@ contract Spoke is ISpoke {
...
 
   // todo rm reserve accounting here & fetch from hub
   function _getReserveDebt(
     DataTypes.Reserve storage reserve
   ) internal view returns (uint256, uint256) {
     uint256 assetId = reserve.assetId;
-    uint256 premiumDebt = reserve.realizedPremium +
-      (HUB.convertToDrawnAssets(assetId, reserve.premiumDrawnShares) - reserve.premiumOffset);
-    return (HUB.convertToDrawnAssets(assetId, reserve.baseDrawnShares), premiumDebt);
+    uint256 accruedPremium = (HUB.convertToPremiumDrawnAssets(assetId, reserve.premiumDrawnShares) -
```
CheyenneAtapour: Extra set of parens here unneeded

foundry.toml
DhairyaSethi: i did will be removing isolate by default for now

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -441,7 +433,7 @@ contract SpokeRepayTest is SpokeBase {
 
   /// repay partial or full premium debt, but no base debt
   function test_repay_only_premium(uint256 daiBorrowAmount, uint40 skipTime) public {
```
```suggestion
  function test_repay_fuzz_only_premium(uint256 daiBorrowAmount, uint40 skipTime) public {
```

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -674,10 +668,11 @@ contract SpokeRepayTest is SpokeBase {
...
-    (uint256 baseRestored, ) = _calculateRestoreAmount(
+    (uint256 baseRestored, uint256 premiumRestored) = _calculateExactRestoreAmount(
       bobDaiBefore.baseDebt,
       bobDaiBefore.premiumDebt,
-      partialRepayAmount
+      partialRepayAmount,
+      daiAssetId
     );
 
     // Partial repay
```
yan-man, line 669: L686 - can change from startPrank to prank

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -700,29 +695,41 @@ contract SpokeRepayTest is SpokeBase {
...
-    (baseRestored, ) = _calculateRestoreAmount(
+    (baseRestored, ) = _calculateExactRestoreAmount(
       bobDaiAfter.baseDebt,
       bobDaiAfter.premiumDebt,
-      bobDaiAfter.totalDebt
+      bobDaiAfter.totalDebt,
+      daiAssetId
     );
 
     // Full repay
```
yan-man, line 725: L744 can change startPrank to prank

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -794,66 +802,55 @@ contract SpokeRepayTest is SpokeBase {
...
-
-    vm.startPrank(bob);
+    vm.expectEmit(address(tokenList.dai));
+    emit IERC20.Transfer(bob, address(hub), repayAmount);
+    vm.prank(bob);
     spoke1.repay(_daiReserveId(spoke1), repayAmount);
-    vm.stopPrank();
+
+    // debt remains unchanged & is donated (premium was already 0)
+    assertEq(getUserDebt(spoke1, bob, _daiReserveId(spoke1)), bobDaiDebtBefore);
```
yan-man, line 835: maybe can confirm donation, ie that hub balance increases, and index has not changed?

tests/unit/Spoke/Spoke.Repay.t.sol
DhairyaSethi: cleaned up all

tests/unit/Spoke/Spoke.Repay.t.sol
DhairyaSethi: yep that's what the expect Transfer event does above (on dai token)

yan-man: also on index change

DhairyaSethi: wdym? index only changes with time

tests/Base.t.sol
```diff
@@ -519,28 +520,40 @@ abstract contract Base is Test {
     vm.stopPrank();
   }
 
-  function updateAssetActive(ILiquidityHub hub, uint256 assetId, bool newActiveFlag) internal {
-    DataTypes.AssetConfig memory assetConfig = hub.getAsset(assetId).config;
+  function updateAssetActive(
+    ILiquidityHub liquidityHub,
```
CheyenneAtapour, line 524: why rename to `liquidityHub` instead of keeping as `hub`?

DhairyaSethi: shadowed var (we have hub state var) to silence compiler warning

tests/unit/Spoke/Spoke.Repay.t.sol
yan-man: just having an assertion that index can't ever be manipulated due to donation; skip time, repay action with donated amount, index remains constant

tests/unit/Spoke/Spoke.Repay.t.sol
CheyenneAtapour, line 1: Did you want to add a test explicitly showing you cannot manipulate index from donating on restore?

We can make a separate ticket for it if we wanna do it outside this PR

DhairyaSethi: discussed it's not possible by definition (unlike supply ex ratio which def needs these checks), but will be adding more sanity checks on #345, added here [f5b4305](https://github.com/aave/aave-v4/pull/343/commits/f5b430558287a11d1b18a84dcad68b4a540c237f)

tests/unit/Spoke/Spoke.Repay.t.sol
DhairyaSethi: [f5b4305](https://github.com/aave/aave-v4/pull/343/commits/f5b430558287a11d1b18a84dcad68b4a540c237f)

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1177,7 +1177,7 @@ contract SpokeRepayTest is SpokeBase {
   }
 
   /// repay all or a portion of total debt in same block
-  function test_repay_same_block_fuzz_amounts(
+  function test_fuzz_repay_same_block_fuzz_amounts(
```
yan-man, line 1180: for consistency like below, let's name like `test_repay_fuzz...`

https://www.notion.so/avara/V4-Code-Conventions-17a9d63a22de80898fb0ee1f6df6038b?pvs=4#17b9d63a22de8067a975e52104fcce03

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1399,7 +1399,7 @@ contract SpokeRepayTest is SpokeBase {
   }
 
   /// repay all or a portion of debt interest
-  function test_repay_fuzz_amounts_only_interest(
+  function test_fuzz_repay_amounts_only_interest(
```
yan-man, line 1402: also here

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -1514,7 +1514,7 @@ contract SpokeRepayTest is SpokeBase {
   }
 
   /// repay all or a portion of premium debt
-  function test_repay_fuzz_amounts_only_premium(
+  function test_fuzz_amounts_repay_only_premium(
```
yan-man, line 1517: also here

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -3332,6 +3332,65 @@ contract SpokeRepayTest is SpokeBase {
...
+    skip(365 days); // initial increase in index, no time passes for subsequent checks
+
+    Debts memory bobDebt = getUserDebt(spoke1, bob, _daiReserveId(spoke1));
+    uint256 supplyExchangeRatioBefore = hub.convertToSuppliedAssets(daiAssetId, 1e30);
+    uint256 debtExchangeRatio = hub.convertToDrawnAssets(daiAssetId, 1e30);
+
+    // repay partial premium debt
+    spoke1.getUserPosition(_daiReserveId(spoke1), bob);
+    vm.assume(bobDebt.premiumDebt > 1);
+    uint256 daiRepayAmount = bound(vm.randomUint(), 1, bobDebt.premiumDebt - 1);
```
yan-man, line 3360: any particular reason to do this over adding daiRepayAmount as a fuzzing input?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -3332,6 +3332,65 @@ contract SpokeRepayTest is SpokeBase {
...
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiBorrowAmount, alice);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+    skip(365 days); // initial increase in index, no time passes for subsequent checks
+
+    Debts memory bobDebt = getUserDebt(spoke1, bob, _daiReserveId(spoke1));
+    uint256 supplyExchangeRatioBefore = hub.convertToSuppliedAssets(daiAssetId, 1e30);
+    uint256 debtExchangeRatio = hub.convertToDrawnAssets(daiAssetId, 1e30);
+
+    // repay partial premium debt
+    spoke1.getUserPosition(_daiReserveId(spoke1), bob);
```
yan-man: can rm

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -3332,6 +3332,65 @@ contract SpokeRepayTest is SpokeBase {
...
+    assertGe(supplyExchangeRatioAfter, supplyExchangeRatioBefore);
+    assertEq(hub.convertToDrawnAssets(daiAssetId, 1e30), debtExchangeRatio);
+
+    supplyExchangeRatioBefore = supplyExchangeRatioAfter;
+
+    vm.prank(bob);
+    spoke1.repay(_daiReserveId(spoke1), type(uint256).max);
+
+    supplyExchangeRatioAfter = hub.convertToSuppliedAssets(daiAssetId, 1e30);
+    assertGe(supplyExchangeRatioAfter, supplyExchangeRatioBefore);
```
yan-man, line 3389: nice, I think in addition to this we should have similar behavior in a supply action where supply ex rate increases (due to rounding down on shares), yes? We can add a test for this too

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -3459,7 +3332,66 @@ contract SpokeRepayTest is SpokeBase {
...
+    uint256 daiRepayAmount = bound(vm.randomUint(), 1, bobDebt.premiumDebt - 1);
+
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.Repay(_daiReserveId(spoke1), bob, 0);
+    vm.prank(bob);
+    spoke1.repay(_daiReserveId(spoke1), daiRepayAmount);
+
+    uint256 supplyExchangeRatioAfter = hub.convertToSuppliedAssets(daiAssetId, 1e30);
+    assertGe(supplyExchangeRatioAfter, supplyExchangeRatioBefore);
+    assertEq(hub.convertToDrawnAssets(daiAssetId, 1e30), debtExchangeRatio);
```
yan-man: the 1e30 is just some arbitrary large amount to demonstrate the ex rate right? perhaps can just re-use MAX_SUPPLY_AMOUNT

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -526,6 +534,22 @@ contract SpokeBase is Base {
     );
   }
 
+  function getUserDebt(
```
yan-man, line 537: there is function name collision with the spoke method that makes this a bit confusing, perhaps we can rename this or just `_getUserDebt`

yan-man:
also closes this issue: https://github.com/aave/aave-v4/issues/339

ticket for withdraw: https://github.com/aave/aave-v4/issues/346

tests/unit/Spoke/Spoke.Repay.t.sol
DhairyaSethi: this makes more sense logically so will prefer to stick with it

tests/unit/Spoke/Spoke.Repay.t.sol
DhairyaSethi: yeah bc it's used twice, didn't want to re-use the same seed

tests/unit/Spoke/SpokeBase.t.sol
DhairyaSethi: that's fine imo, this is just consistency with the other helpers

tests/unit/Spoke/Spoke.Repay.t.sol
DhairyaSethi: yep that's right, added

DhairyaSethi:
good call yep, added for supply & withdraw (we already have for borrow & this PR also adds for repay)
we can expand more on these tests for supply, withdraw, repay like the borrow cases alongside the invariant tests, we should also add these cases for updateRiskPremium (when added) & refreshPremiumDebt (ticket #348)

invariant and formal verification tests would be more exhaustive for this but regardless good to have these unit tests

PR #345, opened 2025-04-15
Add Donation Checks

CheyenneAtapour:
Add sanity tests which ensure that donations (either direct or via repay) do not change the debt exchange ratio. They should not because the debt exchange ratio is calculated via an index.


CheyenneAtapour:
These tests are irrelevant since debt is computed via index, and so ofc donations don't affect it

PR #350, opened 2025-04-17
fix: preview offset

DhairyaSethi:
- utilizes soln 2: round down when calculating offset only to always undercount it such that we don't underflow and supply rate does not decrease
- adds bounds checks to `refreshPremiumDebt` and introduces `settlePremiumDebt` for repay
- adds `_repayAll` to tests to validate no dust is left on full repay in different scenarios

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -232,6 +232,7 @@ contract LiquidityHub is ILiquidityHub {
     _refresh(assetId, msg.sender, premiumDrawnShareDelta, premiumOffsetDelta, realizedPremiumDelta);
     (uint256 baseDebtAfter, uint256 premiumDebtAfter) = _assets[assetId].debt();
     // can increase due to precision loss on premium debt (base unchanged)
+    // todo mathematically find premium diff ceiling and replace the `2`
     require(baseDebtAfter == baseDebt && premiumDebtAfter - premiumDebt <= 2, InvalidDebtChange());
   }
 


src/contracts/LiquidityHub.sol
```diff
@@ -220,40 +220,59 @@ contract LiquidityHub is ILiquidityHub {
...
-     * - check that total debt can only:
-     *   - reduce until `premiumDebt` if called after a restore (tstore premiumDebt?)
-     *   - remains unchanged on all other calls
-     * `refreshPremiumDebt` is game-able only for premium stuff
-     */
+    // todo only spoke
+    uint256 totalDebt = _assets[assetId].totalDebt();
+    _refresh(assetId, msg.sender, premiumDrawnShareDelta, premiumOffsetDelta, realizedPremiumDelta);
+    uint256 totalDebtAfter = _assets[assetId].totalDebt();
+    require(totalDebtAfter >= totalDebt && totalDebtAfter - totalDebt <= 2, InvalidDebtChange()); // can increase due to precision
```
DhairyaSethi: the two here is seen from our fuzz tests, i believe we can also mathematically figure out the max precision loss possible and set that here since hub will be immutable but i would leave it for the future

src/contracts/LiquidityHub.sol
```diff
@@ -220,40 +220,61 @@ contract LiquidityHub is ILiquidityHub {
     return baseDrawnSharesRestored;
   }
 
+  /// @inheritdoc ILiquidityHub
   function refreshPremiumDebt(
     uint256 assetId,
-    int256 premiumDrawnSharesDelta,
+    int256 premiumDrawnShareDelta,
```
CheyenneAtapour, line 226: not really a big deal, but I think it was fine the way it was

DhairyaSethi: intentional such that the _refresh below lints in one line

src/interfaces/ILiquidityHub.sol
```diff
@@ -114,20 +115,44 @@ interface ILiquidityHub {
     address from
   ) external returns (uint256);
 
+  /**
+   * @notice Refresh premium debt to be called when moving accrued premium to realized premium.
```
CheyenneAtapour: I think the `@notice` comment could be made simpler, along the lines of "Realizes premium debt", and then another `@dev` can be added to describe when/where it should be called. Same applies to the below natspec for `settlePremiumDebt`

Currently the `@notice` comment isn't a grammatically correct sentence.

But also this is nbd, can fix later

src/contracts/LiquidityHub.sol
CheyenneAtapour: ok, interesting. I think it's fine here, but in general, Idt we should worry about changing our variable names wrt linting, instead just try to pick the best variable names as top priority. But the cost/benefit here is super marginal, so I think this is fine as you made it

PR #353, opened 2025-04-18
Fix supply cap rounding direction

### LINKED ISSUE #335: Wrong rounding direction when checking supply cap
**Issue**
When validating the supply in the `_validateSupply` function, we check that the supply cap is not exceeded. To do this, we convert the supplied shares into assets, and we round down. However, this is not in the advantage of the protocol, which should protect itself by rounding up.
https://github.com/aave/aave-v4/blob/d7812d541152d82f13f00f0b8f72f9e01fe2c873/src/contracts/LiquidityHub.sol#L368-L372

**Proposed Fix**
Line 370 should say:
`asset.toSuppliedAssetsUp(spoke.suppliedShares) + amount <= spoke.config.supplyCap,`

avniculae:
Closes #335

### DIFF
--- a/tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
@@ -121,6 +121,8 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     uint256 totalSuppliedAssets = hub.getTotalSuppliedAssets(daiAssetId);
     uint256 totalSuppliedShares = hub.getAssetSuppliedShares(daiAssetId);
 
+    // Depending on the borrow rate, this may not be true
+    // It can be adjusted by changing the amount of assets passed to _increaseExchangeRate
     assertNotEq(
       totalSuppliedAssets % totalSuppliedShares,
       0,
@@ -229,6 +231,8 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     uint256 totalSuppliedAssets = hub.getTotalSuppliedAssets(daiAssetId);
     uint256 totalSuppliedShares = hub.getAssetSuppliedShares(daiAssetId);
 
+    // Depending on the borrow rate, this may not be true
+    // It can be adjusted by changing the amount of assets passed to _increaseExchangeRate
     assertNotEq(
       totalSuppliedAssets % totalSuppliedShares,
       0,


tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,29 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.add(daiAssetId, amount, alice);
   }
 
+  function test_supply_revertsWith_SupplyCapExceeded_due_to_rounding() public {
+    uint256 amount = 100e18;
+    _increaseExchangeRate(amount);
+
+    uint256 sharePrice = hub.convertToSuppliedAssets(daiAssetId, 1e18);
+    assertLt(1e18, sharePrice);
+    assertLt(sharePrice, 2e18);
```
CheyenneAtapour: This is what I think @DhairyaSethi was mentioning we don't need to assume in standup today

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,29 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+    // set cap of spoke1 to 3 assets
+    uint256 newSupplyCap = 3;
+    _updateSupplyCap(daiAssetId, address(spoke1), newSupplyCap);
+
+    // spoke1 already supplied 1 share, which rounded up is 2 assets
+    // supplying 2 more assets will exceed the cap
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.SupplyCapExceeded.selector, newSupplyCap));
+    vm.prank(address(spoke1));
+    hub.add(daiAssetId, 2, alice);
+  }
```
CheyenneAtapour, line 160: Even without rounding, this test would fail because 4 supplied assets is more than 3 supplied assets (the cap). Maybe there is another example that can better show what you want to show here? Or can rename the test to better reflect

avniculae: This is not true, the test succeeds if share->asset conversion performs rounding down. This is because the supplied share will convert to 1 asset, which together with the 2 assets that are being supplied in the last call, will sum to 3 assets, which is equal (but not higher) to the cap.

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: yeah, but I am still unsure what the proposed solution entails

yan-man: in general here, share price can inflate such that it is larger than 2e18 right? so why do we need this assertion

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
yan-man: can you utilize sharesMath lib and perhaps show that the rounding down would be less than the cap (ie theoretically not reaching the cap limit)?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -164,6 +187,35 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+    hub.add(daiAssetId, 2, alice);
+
+    // set cap of spoke1 to 5 assets
+    uint256 newSupplyCap = 5;
+    _updateSupplyCap(daiAssetId, address(spoke1), newSupplyCap);
+
+    // spoke1 already supplied 2 shares, which rounded up is 3 assets
+    // supplying 2 more assets will reach the cap, but not exceed it
+    // note that actual supplied assets is 6, but 1 share was lost
+    // due to rounding errors on suppliedShares
```
yan-man: is there an error in the rounding?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
CheyenneAtapour: @avniculae basically you can just remove these assertions

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
CheyenneAtapour: @avniculae ah ok, I see what you're saying. As @yan-man says, might be nice to show a test case confirming that the test would succeed in the case of rounding down

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: if price is >= 2, then rounding up and down will result in the same outcome (take price=2 as an example)

avniculae: curious to understand what you would approach this @yan-man @CheyenneAtapour

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: yes

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: Addressed: [1e86ae3](https://github.com/aave/aave-v4/pull/353/commits/1e86ae312f542dd0fae3ed6a78fc631d2a70d17d)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: Addressed: [1e86ae3](https://github.com/aave/aave-v4/pull/353/commits/1e86ae312f542dd0fae3ed6a78fc631d2a70d17d)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
yan-man: even after this supply cap rounding fix in this pr?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,47 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.add(daiAssetId, amount, alice);
   }
 
+  function test_supply_revertsWith_SupplyCapExceeded_due_to_rounding() public {
```
miguelmtzinf, line 118: can you add a comment above, with a description of the test please?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,47 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+      totalSuppliedAssets % totalSuppliedShares,
+      0,
+      'totalSuppliedAssets % totalSuppliedShares is zero'
+    );
+
+    // spoke1 supplies 1 share worth of assets (rounded down) + 1
+    // the supplied share is 1, which rounded up is equal to the
+    // amount of assets supplied
+    vm.prank(address(spoke1));
+    hub.add(daiAssetId, totalSuppliedAssets / totalSuppliedShares + 1, alice);
```
miguelmtzinf: better to create a variable for this amount

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,47 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.add(daiAssetId, amount, alice);
   }
 
+  function test_supply_revertsWith_SupplyCapExceeded_due_to_rounding() public {
+    uint256 amount = 100e18;
```
miguelmtzinf: this amount is never used again

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -164,6 +205,35 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+  function test_supply_SupplyCapReachedButNotExceeded() public {
+    uint256 amount = 100e18;
+    _increaseExchangeRate(amount);
+
+    uint256 sharePrice = hub.convertToSuppliedAssets(daiAssetId, 1e18);
+    assertLt(1e18, sharePrice);
+    assertLt(sharePrice, 2e18);
+
+    // spoke 1 supplies 2 dai (1 share)
+    vm.prank(address(spoke1));
```
miguelmtzinf: use utils functions pls

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -164,6 +205,35 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+    hub.add(daiAssetId, 2, alice);
+
+    // set cap of spoke1 to 5 assets
+    uint256 newSupplyCap = 5;
+    _updateSupplyCap(daiAssetId, address(spoke1), newSupplyCap);
+
+    // spoke1 already supplied 2 shares, which rounded up is 3 assets
+    // supplying 2 more assets will reach the cap, but not exceed it
+    // note that actual supplied assets is 6, but 1 share was lost
```
miguelmtzinf: add getters and asserts for these assumptions

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: what do you mean?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: Addressed: [2c45b63](https://github.com/aave/aave-v4/pull/353/commits/2c45b63bcaac80a922d318718e71ee4a62d1b63f)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: Addressed: [2c45b63](https://github.com/aave/aave-v4/pull/353/commits/2c45b63bcaac80a922d318718e71ee4a62d1b63f)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: Addressed: [2c45b63](https://github.com/aave/aave-v4/pull/353/commits/2c45b63bcaac80a922d318718e71ee4a62d1b63f)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: Addressed: [2c45b63](https://github.com/aave/aave-v4/pull/353/commits/2c45b63bcaac80a922d318718e71ee4a62d1b63f)

I also refactored this test following the same approach as the other one: computing the values on the go, relative to the share price. It complicated things a bit, but let me know what you think

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
DhairyaSethi: https://github.com/aave/aave-v4/blob/fix/335-supply-cap-rounding/tests/Utils.sol
usage: https://github.com/aave/aave-v4/blob/fix/335-supply-cap-rounding/tests/unit/Spoke/Spoke.Repay.t.sol#L131

avniculae: thanks, addressed: [789ec9d](https://github.com/aave/aave-v4/pull/353/commits/789ec9d24223b9266311e74b2b6a5cf361ed8111)

note that I couldn't use the utils when I expect the add to revert, given that the util approves first, and hence the next call does not revert (but rather the one after that one)

DhairyaSethi: where's the expect revert you're talking about? was going to suggest we make approval optional by having an allowance check but annoyingly the expectRevert would be consumed by this check, we can just use the naked method for expecting a revert for now (unless you had some thoughts @yan-man?)

yan-man: yeah that's annoying, let's keep it simple and just use the naked method for the reverts

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,9 +114,11 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.add(daiAssetId, amount, alice);
   }
 
+  // Tests that the supply cap is enforced by rounding up the conversion
+  // of supplied shares into assets. Also verifies that rounding down
+  // would fail to enforce the cap.
   function test_supply_revertsWith_SupplyCapExceeded_due_to_rounding() public {
-    uint256 amount = 100e18;
-    _increaseExchangeRate(amount);
+    _increaseExchangeRate(100e18);
```
yan-man: i know I wrote this originally but I think i did a poor job bc it only works with dai, can you generalize this helper method to accept assetId

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,54 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+  // Tests that the supply cap is enforced by rounding up the conversion
+  // of supplied shares into assets. Also verifies that rounding down
+  // would fail to enforce the cap.
+  function test_supply_revertsWith_SupplyCapExceeded_due_to_rounding() public {
+    _increaseExchangeRate(100e18);
+
+    uint256 availableLiquidity = hub.getAvailableLiquidity(daiAssetId);
+    uint256 totalDebt = hub.getAssetTotalDebt(daiAssetId);
+
+    uint256 totalSuppliedAssets = availableLiquidity + totalDebt;
```
yan-man: nit: i think you can remove availableLiquidity and totalDebt vars and just add them together here as they aren't used again

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: [1c81532](https://github.com/aave/aave-v4/pull/353/commits/1c815323fc7b706808ca36ca8b8512de4f9a8139)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: [1c81532](https://github.com/aave/aave-v4/pull/353/commits/1c815323fc7b706808ca36ca8b8512de4f9a8139)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: yes, this does not refer to the supply cap issue this PR is fixing, but rather by the small rounding errors that take place at each supply, where the actual sum of assets provided can be higher than the total asset amount as perceived by the hub (which is converting total shares to asset). see comments in the test after latest fixes

tests/unit/LiquidityHub/LiquidityHubBase.t.sol
```diff
@@ -58,45 +58,46 @@ contract LiquidityHubBase is Base {
     hub.updateSpokeConfig(assetId, spoke, spokeConfig);
   }
 
-  /// @dev spoke1 (alice) supplies dai, spoke1 (alice) draws dai, skip 1 year
+  /// @dev spoke1 (alice) supplies asset, spoke1 (alice) draws asset, skip 1 year
```
miguelmtzinf:
```suggestion
  /// @dev spoke2 (bob) supplies asset, spoke1 (alice) draws asset, skip 1 year
```

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,51 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.add(daiAssetId, amount, alice);
   }
 
+  // Tests that the supply cap is enforced by rounding up the conversion
+  // of supplied shares into assets. Also verifies that rounding down
+  // would fail to enforce the cap.
+  function test_supply_revertsWith_SupplyCapExceeded_due_to_rounding() public {
+    _increaseExchangeRate(daiAssetId, 100e18);
+
+    uint256 totalSuppliedAssets = hub.getAvailableLiquidity(daiAssetId) + hub.getAssetTotalDebt(daiAssetId);
```
miguelmtzinf: why not use `hub.getAssetSuppliedAssets`

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,51 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+    uint256 totalSuppliedShares = hub.getAssetSuppliedShares(daiAssetId);
+
+    assertNotEq(
+      totalSuppliedAssets % totalSuppliedShares,
```
miguelmtzinf, line 127: but this is not always the case right? depending on the borrow rate we could have suppliedAssets being divisible by suppliedShares no?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,51 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+      0,
+      'totalSuppliedAssets % totalSuppliedShares is zero'
+    );
+
+    // The asset amount is 1 share worth of assets (rounded down) + 1
+    // The supplied share is 1, which rounded up is equal to the
+    // amount of assets supplied
+    uint256 assetAmount = totalSuppliedAssets / totalSuppliedShares + 1;
```
miguelmtzinf: better to rename to `supplyAmount`

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -114,6 +114,51 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.add(daiAssetId, amount, alice);
   }
 
+  // Tests that the supply cap is enforced by rounding up the conversion
+  // of supplied shares into assets. Also verifies that rounding down
+  // would fail to enforce the cap.
```
miguelmtzinf:
```suggestion
  /// supply reverts if the cap is exceeded due to interest accrual, with proper rounding (up) applied to shares into assets conversion
```
more concise is fine i think

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: you mean `getAssetSuppliedAmount` ? I want to compute the exchange rate accurately, so I replicate `AssetLogic.totalSuppliedAssets` given that we don't have an exposed getter for it. using `hub.getAssetSuppliedAmount` will result in a slightly different number (due to donations)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: [2b285e2](https://github.com/aave/aave-v4/pull/353/commits/2b285e26d9105bd93b5103712aa6dca7f929ca3c)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
miguelmtzinf: wondering if we should add `getAssetSuppliedAssets` for better dx

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
miguelmtzinf: leave a comment here @avniculae

avniculae: [481f876](https://github.com/aave/aave-v4/pull/353/commits/481f8765ac6349bac94902ca15f1580cdf23a104)

PR #355, opened 2025-04-21
certora setup LiquidityHub

nd-certora:


certora/spec/LiquidityHub.spec
```diff
@@ -0,0 +1,310 @@
...
+{
+    uint256 assetId;
+
+    requireAllInvariants(assetId, e);
+    // use ghost to avoid repeating complex computation
+    mathint assetsBefore = supplyAmountBefore; 
+    mathint sharesBefore = supplyShareBefore;
+
+    // todo filter out when no time pass
+    require liquidityHub._assets[assetId].lastUpdateTimestamp == e.block.timestamp;
```
DhairyaSethi: this comment was actually just meant to isolate the accrue interest (supply rate should strictly increase if there is debt in the system) and general case (it's greater than or equal to)
maybe we rm this assert from here for now?

nd-certora: we must do modular verification, there is an assumption that  accure does nothing in  all the rules in this files.
 ``` 
 function AssetLogic.accrue(DataTypes.Asset storage asset) internal => NONDET;
 ```
 So this require is just to make it clearer and we will add a comment to explain this. 
 
 We have a different spec file for checking accure related properties and we will check that accure also preserves this monotonic property

certora/spec/Hub.spec
```diff
@@ -474,5 +474,15 @@ rule frontRunOnRemove(uint256 assetId, method f) {
     remove@withrevert(e,assetId, amount, from);
     assert !lastReverted;
     // it is possible for everyone to remove and than zero shares left
-    satisfy !lastReverted && supplyAmountBefore!=0 && getAssetSuppliedAmount(e,assetId) == 0;
+    satisfy !lastReverted && supplyAmountBefore!=0 && getAssetAddedAmount(e,assetId) == 0;
 }
+
+invariant premiumOffset_Integrity(uint256 assetId) 
+    hub._assets[assetId].premiumOffset <= hub._assets[assetId].premiumShares * hub._assets[assetId].drawnRate / wadRayMath.RAY()
```
DhairyaSethi: it should be multiplied by `drawnIndex` (as that's the drawn share price, `drawnRate` is v3 equivalent of borrow rate, we can rename if it's confusing), the division should round up?

PR #365, opened 2025-04-24
fix: run ci on stable forge 

DhairyaSethi:
parameterise bgd workflow: https://github.com/bgd-labs/github-workflows/pull/52

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -21,6 +21,11 @@ runs = 5000
 [profile.ci.fuzz]
 runs = 100000
 
+[profile.gas]
+isolate = true
+gas_snapshot_check = true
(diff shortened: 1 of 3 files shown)


foundry.toml
```diff
@@ -11,9 +11,16 @@ optimizer = true
 optimizer_runs = 200
 bytecode_hash = "none"
 
-[fuzz]
+[profile.default.fuzz]
+runs = 5000
+seed = "0x640"
```
yan-man, line 16: this one is optional, do you feel like it helps here?

DhairyaSethi: yes it makes the runs deterministic

PR #366, opened 2025-04-25
feat: optimise risk premium calc implementation

### LINKED ISSUE #141: Keeping Assets Sorted by Liquidity Premium

avniculae:
Closes #141 

The main idea is to duplicate the Reserve struct fiels whch are used in the rp algorithm into a new array, which is always sorted by the liquidityPremium

- it results in less gas generally
- it allows us to further optimise storage (by packing things into the same slot)
- it passes all tests from #321

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.0;
 // libraries
 import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
-import {KeyValueListInMemory} from 'src/libraries/helpers/KeyValueListInMemory.sol';
+import {OrderedReserveRiskArray} from 'src/libraries/helpers/OrderedReserveRiskArray.sol';
 import {DataTypes} from 'src/libraries/types/DataTypes.sol';
 import {LiquidationLogic} from 'src/libraries/logic/LiquidationLogic.sol';
 
@@ -16,7 +16,7 @@ import {IPriceOracle} from 'src/interfaces/IPriceOracle.sol';
 contract Spoke is ISpoke {
   using WadRayMath for uint256;
   using PercentageMath for uint256;
-  using KeyValueListInMemory for KeyValueListInMemory.List;
+  using OrderedReserveRiskArray for DataTypes.ReserveRiskConfig[];
   using LiquidationLogic for DataTypes.LiquidationConfig;
 
   uint256 public constant HEALTH_FACTOR_LIQUIDATION_THRESHOLD = WadRayMath.WAD;
@@ -27,7 +27,7 @@ contract Spoke is ISpoke {
     internal _userPositions;
   mapping(uint256 reserveId => DataTypes.Reserve reserveData) internal _reserves;
   DataTypes.LiquidationConfig internal _liquidationConfig;
-  uint256[] public reservesList; // todo: rm, not needed
+  DataTypes.ReserveRiskConfig[] public reserveRiskConfigs; // sorted in ascending order by liquidityPremium
   uint256 public reserveCount;
 
   constructor(address hubAddress, address oracleAddress, uint256 closeFactorValue) {
@@ -62,7 +62,14 @@ contract Spoke is ISpoke {
     address asset = address(HUB.assetsList(assetId)); // will revert on invalid assetId
     uint256 reserveId = reserveCount++;
     // TODO: AccessControl
-    reservesList.push(reserveId);
+    reserveRiskConfigs.insert(
+      DataTypes.ReserveRiskConfig({
+        reserveId: reserveId,
+        assetId: assetId,
+        collateralFactor: config.collateralFactor,
+        liquidityPremium: config.liquidityPremium
+      })
+    );
     _reserves[reserveId] = DataTypes.Reserve({
       reserveId: reserveId,
       assetId: assetId,
@@ -99,6 +106,14 @@ contract Spoke is ISpoke {
     DataTypes.Reserve storage reserve = _reserves[reserveId];
     require(reserve.asset != address(0), InvalidReserve());
     // TODO: AccessControl
+    reserveRiskConfigs.update(
+      DataTypes.ReserveRiskConfig({
+        reserveId: reserveId,
+        assetId: reserve.assetId,
+        collateralFactor: config.collateralFactor,
+        liquidityPremium: config.liquidityPremium
+      })
+    );
     reserve.config = DataTypes.ReserveConfig({
       decimals: reserve.config.decimals, // decimals remains existing value
       active: config.active,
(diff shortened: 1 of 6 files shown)


src/libraries/helpers/OrderedReserveRiskArray.sol
```diff
@@ -0,0 +1,70 @@
...
+    _remove(array, _find(array, value));
+    _insert(array, value, _upper_bound(array, value));
+  }
+
+  function _upper_bound(
+    DataTypes.ReserveRiskConfig[] storage array,
+    DataTypes.ReserveRiskConfig memory value
+  ) private view returns (uint256) {
+    uint256 index = array.length;
+    while (index > 0 && array[index - 1].liquidityPremium > value.liquidityPremium) {
```
DhairyaSethi, line 29: this is sorted so we can binary search for this & same for `_find`

avniculae: yes, I know. I wanted to keep it simple for the first iteration

src/contracts/Spoke.sol
```diff
@@ -584,126 +600,95 @@ contract Spoke is ISpoke {
   function _calculateUserAccountData(
     address userAddress
   ) internal view returns (uint256, uint256, uint256, uint256, uint256) {
-    DataTypes.CalculateUserAccountDataVars memory vars;
-    uint256 reservesListLength = reservesList.length;
+    uint256 reserveRiskConfigsLength = reserveRiskConfigs.length;
+
+    uint256 totalDebtInBaseCurrency;
+    uint256[] memory userCollateralInBaseCurrency = new uint256[](reserveRiskConfigsLength);
```
DhairyaSethi, line 606: not needed, we can iteratively keep a running sum like before

avniculae: I need to first calculate the debt, and I am storing these in the array to not load the common variables again

avniculae: the running sum is still there, is called `collateralCoveredInBaseCurrency`

src/contracts/Spoke.sol
```diff
@@ -584,126 +600,95 @@ contract Spoke is ISpoke {
...
 
-    while (vars.reserveId < reservesListLength) {
-      DataTypes.UserPosition storage userPosition = _userPositions[userAddress][vars.reserveId];
+    uint256 totalCollateralInBaseCurrency;
+    uint256 avgCollateralFactor;
+
+    for (uint256 i = 0; i < reserveRiskConfigsLength; i += 1) {
+      DataTypes.ReserveRiskConfig storage reserveRiskConfig = reserveRiskConfigs[i];
+      DataTypes.UserPosition storage userPosition = _userPositions[userAddress][
+        reserveRiskConfig.reserveId
```
DhairyaSethi, line 614: so we can't escape this sload for reserveId on each iteration (unlike before)
because of this, i'm not sure if this scales with a higher global reserve count
can you compare the gas usage of this and previous iteration (which still has a countBits opt pending to reduce loops from 3 to 2) for the simplest case where there's 128 reserves listed, and the user borrows, 2, 3, 5, 10?
i fear bc of this sload this will be more expensive

src/contracts/Spoke.sol
```diff
@@ -584,126 +600,95 @@ contract Spoke is ISpoke {
...
-        ++vars.i;
+      : avgCollateralFactor.wadDiv(totalCollateralInBaseCurrency).fromBps();
+
+    uint256 collateralCoveredInBaseCurrency;
+    uint256 userRiskPremium;
+    for (
+      uint256 i = 0;
+      i < reserveRiskConfigsLength && collateralCoveredInBaseCurrency < totalDebtInBaseCurrency;
+      i += 1
+    ) {
```
DhairyaSethi, line 669: we can think about optimising further & removing this loop, since the list is presorted, we can build and numerator (col value * lp) and denominator (total col value until debt = total debt value) separately and at the end divide
the problem is slicing the numerator sum until the total debt value is reached

avniculae: > the problem is slicing the numerator sum until the total debt value is reached

exactly, this is why I think we need 2 loops

src/libraries/helpers/OrderedReserveRiskArray.sol
DhairyaSethi: pls add a comment for now

src/contracts/Spoke.sol
avniculae: you are right, for 100 reserves listed:
- borrow, this algorithm: 1691202
- borrow, previously: 1238292

however, one of the benefits for this algorithm is that we can easily pack the fields into less slots, whereas for the other one I don't think that was possible due to the on-demand sorting, right?

DhairyaSethi: we can definitely pack there as well, and it'd be in memory there whereas here we'd be packing the reserve list in storage here but these would different values so can we compare. i don't expect meaningful differences tbh bc the cost with this appraoch expands with total reserve count whereas the latter expands with supplied reserve count

src/contracts/Spoke.sol
DhairyaSethi: yeah we can get the current impl down to 2 loops as well w [this](https://github.com/aave/aave-v4/blob/main/src/contracts/Spoke.sol#L608) opt, wondering if we can select a probabilistic anchor point to slice the array here or something

src/contracts/Spoke.sol
DhairyaSethi: ah yeah! the problem is this array in memory of size totalReserveCount whereas the in memory on demand sorting array kept this [before](https://github.com/aave/aave-v4/blob/main/src/contracts/Spoke.sol#L650) which was of length totalUserSuppliedReserveCount 
this will be also show significant differences with high totalReserveCounts since memory expansion costs increase quadratically  
regardless we can somewhat mitigate this by keeping this array packed as well and at the cost of another loop, match it's size with the current impl

src/libraries/helpers/OrderedReserveRiskArray.sol
miguelmtzinf, line 1: is there any battle-tested lib we can use for this? If not, I d abstract it a bit, so it's not "hidden" that we sort by liquidityPremium internally

src/contracts/Spoke.sol
miguelmtzinf: we cannot use transient because it does not support arrays yet right
but yeah, let's opt later

DhairyaSethi: transient storage will just be used for a lookup for reserveId => userCollateralValueInBaseCurrency which has been computed already such that the size of the in-memory in both solutions can be packed

avniculae:
this is resulting in more gas once you have multiple reserves registered

PR #367, opened 2025-04-26
Additional User RP Tests

### LINKED ISSUE #244: Missing User Risk Premium Tests

CheyenneAtapour:
closes #244

### DIFF
--- a/tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
+++ b/tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
@@ -922,6 +922,85 @@ contract SpokeRiskPremiumScenarioTest is SpokeBase {
     _verifyProtocolDebtShares(bobDaiInfo, aliceDaiInfo, bobUsdxInfo, aliceUsdxInfo, 'final');
   }
 
+  /// Bob supplies and borrows varying amounts of 4 reserves. We fuzz prices and liquidity premiums, and wait arbitrary time.
+  /// We ensure risk premium is calculated correctly before and after the time passing
+  function test_getUserRiskPremium_fuzz_inflight_calcs(
+    UserBorrowAction memory daiAmounts,
+    UserBorrowAction memory wethAmounts,
+    UserBorrowAction memory usdxAmounts,
+    UserBorrowAction memory wbtcAmounts,
+    uint40 skipTime
+  ) public {
+    skipTime = uint40(bound(skipTime, 1, MAX_SKIP_TIME));
+
+    daiAmounts.supplyAmount = bound(daiAmounts.supplyAmount, 0, MAX_SUPPLY_AMOUNT);
+    wethAmounts.supplyAmount = bound(wethAmounts.supplyAmount, 0, MAX_SUPPLY_AMOUNT);
+    usdxAmounts.supplyAmount = bound(usdxAmounts.supplyAmount, 0, MAX_SUPPLY_AMOUNT);
+    wbtcAmounts.supplyAmount = bound(wbtcAmounts.supplyAmount, 0, MAX_SUPPLY_AMOUNT);
+
+    daiAmounts.borrowAmount = bound(daiAmounts.borrowAmount, 0, daiAmounts.supplyAmount / 2);
+    wethAmounts.borrowAmount = bound(wethAmounts.borrowAmount, 0, wethAmounts.supplyAmount / 2);
+    usdxAmounts.borrowAmount = bound(usdxAmounts.borrowAmount, 0, usdxAmounts.supplyAmount / 2);
+    wbtcAmounts.borrowAmount = bound(wbtcAmounts.borrowAmount, 0, wbtcAmounts.supplyAmount / 2);
+
+    // Ensure supplied value is at least double borrowed value to pass hf checks
+    vm.assume(
+      _getValueInBaseCurrency(daiAssetId, daiAmounts.supplyAmount) +
+        _getValueInBaseCurrency(wethAssetId, wethAmounts.supplyAmount) +
+        _getValueInBaseCurrency(usdxAssetId, usdxAmounts.supplyAmount) +
+        _getValueInBaseCurrency(wbtcAssetId, wbtcAmounts.supplyAmount) >=
+        2 *
+          (_getValueInBaseCurrency(daiAssetId, daiAmounts.borrowAmount) +
+            _getValueInBaseCurrency(wethAssetId, wethAmounts.borrowAmount) +
+            _getValueInBaseCurrency(usdxAssetId, usdxAmounts.borrowAmount) +
+            _getValueInBaseCurrency(wbtcAssetId, wbtcAmounts.borrowAmount))
+    );
+
+    // Bob supplies and draws all assets on spoke1
+    if (daiAmounts.supplyAmount > 0) {
+      Utils.supplyCollateral(spoke1, _daiReserveId(spoke1), bob, daiAmounts.supplyAmount, bob);
+    }
+    if (wethAmounts.supplyAmount > 0) {
+      Utils.supplyCollateral(spoke1, _wethReserveId(spoke1), bob, wethAmounts.supplyAmount, bob);
+    }
+    if (usdxAmounts.supplyAmount > 0) {
+      Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), bob, usdxAmounts.supplyAmount, bob);
+    }
+    if (wbtcAmounts.supplyAmount > 0) {
+      Utils.supplyCollateral(spoke1, _wbtcReserveId(spoke1), bob, wbtcAmounts.supplyAmount, bob);
+    }
+
+    if (daiAmounts.borrowAmount > 0) {
+      Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiAmounts.borrowAmount, bob);
+    }
+    if (wethAmounts.borrowAmount > 0) {
+      Utils.borrow(spoke1, _wethReserveId(spoke1), bob, wethAmounts.borrowAmount, bob);
+    }
+    if (usdxAmounts.borrowAmount > 0) {
+      Utils.borrow(spoke1, _usdxReserveId(spoke1), bob, usdxAmounts.borrowAmount, bob);
+    }
+    if (wbtcAmounts.borrowAmount > 0) {
+      Utils.borrow(spoke1, _wbtcReserveId(spoke1), bob, wbtcAmounts.borrowAmount, bob);
+    }
+
+    // Check bob's user risk premium
+    assertEq(
+      spoke1.getUserRiskPremium(bob),
+      _calculateExpectedUserRP(bob, spoke1),
+      'user risk premium'
+    );
+
+    // Now skip some time
+    skip(skipTime);
+
+    // Recheck bob's user risk premium
+    assertEq(
+      spoke1.getUserRiskPremium(bob),
+      _calculateExpectedUserRP(bob, spoke1),
+      'user risk premium after time skip'
+    );
+  }
+
   function _boundUserBorrowAction(
     UserBorrowAction memory action
   ) internal pure returns (UserBorrowAction memory) {


src/contracts/Spoke.sol
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
+import {console2 as console} from 'forge-std/console2.sol';
```
yan-man: rm

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,444 @@
...
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+contract SpokeRiskPremiumEdgeCasesTest is SpokeBase {
+  using SharesMath for uint256;
+  using WadRayMath for uint256;
+  using PercentageMath for uint256;
+
+  /// Bob supplies 2 collateral assets, borrows an amount such that both of them cover it, and then repays any nonzero amount of the higher lp one
+  /// Bob's user risk premium should decrease or remain same after repay
+  /// @dev due to rounding within risk premium calc, repaying doesn't guarantee user rp decrease
```
yan-man, line 12: is it only the repay donation scenario that ends up with the rp not decreasing?

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,444 @@
...
+  function test_riskPremium_nonIncreasingAfterRepay(
+    uint256 usdxSupplyAmount,
+    uint256 daiSupplyAmount,
+    uint256 borrowAmount,
+    uint256 repayAmount
+  ) public {
+    // Make usdx liquidity premium 10 so it's the lower lp asset
```
yan-man:
```suggestion
    // Make usdx liquidity premium 10% so it's the lower lp asset
```

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,444 @@
...
+    updateLiquidityPremium(spoke2, _usdxReserveId(spoke2), 10_00);
```
yan-man, line 20: can perhaps add an assert that usdx is indeed the lower lp asset

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,444 @@
...
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+contract SpokeRiskPremiumEdgeCasesTest is SpokeBase {
+  using SharesMath for uint256;
+  using WadRayMath for uint256;
+  using PercentageMath for uint256;
+
+  /// Bob supplies 2 collateral assets, borrows an amount such that both of them cover it, and then repays any nonzero amount of the higher lp one
```
yan-man: is this meant to be 3 coll assets then? curious why you used dai/dai2 and 3 coll assets instead of 2

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,444 @@
...
+      spoke: spoke2,
+      user: bob,
+      reserveId: _daiReserveId(spoke2),
+      usingAsCollateral: false
+    });
+
+    assertGt(
+      spoke2.getUserRiskPremium(bob),
+      riskPremium,
+      'Risk premium should increase after removing collateral'
```
yan-man:
```suggestion
      'Risk premium should increase after disabling lower LP asset as collateral'
```

or something like this

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,444 @@
...
+      reserveId: _daiReserveId(spoke2),
+      user: bob,
+      amount: daiSupplyAmount,
+      onBehalfOf: bob
+    });
+
+    assertGt(
+      spoke2.getUserRiskPremium(bob),
+      riskPremium,
+      'Risk premium should increase after withdrawing collateral'
```
yan-man:
```suggestion
      'Risk premium should increase after withdrawing lower LP collateral'
```

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: dai2 is just used for borrowing power, and it's higher LP than the other 2 collaterals being tested in the user RP here, so it doesn't affect the user rp calc, intentionally

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: No, it can be due to rounding in general because we are using bps values. 1 wei of asset difference won't generally change the user rp calc

src/contracts/Spoke.sol
CheyenneAtapour: done, ty

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: done

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,449 @@
...
+
+    daiSupplyAmount = bound(daiSupplyAmount, 4, MAX_SUPPLY_AMOUNT);
+    borrowAmount = bound(borrowAmount, 3, MAX_SUPPLY_AMOUNT / 2);
+    // Force least lp asset supply amount to be less than borrow amount, so borrow covered by 2 collaterals at least
+    // Here it will be due to decimals of usdx vs dai
+    usdxSupplyAmount = bound(usdxSupplyAmount, 1, borrowAmount);
+    repayAmount = bound(repayAmount, 2, borrowAmount);
+
+    // Deal bob dai to cover dai and dai2 supply
+    deal(address(tokenList.dai), bob, MAX_SUPPLY_AMOUNT * 2);
```
yan-man, line 38: maybe can add a setup method and include this deal in there, ig only used 2 times nbd

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,449 @@
...
+    // Now bob withdraws dai
+    Utils.withdraw({
+      spoke: spoke2,
+      reserveId: _daiReserveId(spoke2),
+      user: bob,
+      amount: daiSupplyAmount,
+      onBehalfOf: bob
+    });
+
+    assertGt(
```
yan-man: can it also be Ge for the same rationale, ie if some dust amount is withdrawn? maybe can add a fuzz test where not full `daiSupplyAmount`, but just some fuzzed amount <= `daiSupplyAmount` is withdrawn

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
yan-man, line 1: nice test scenarios. Maybe also:
- some fuzzed versions of these
- initially 1 coll covers debt, then due to borrowing more of another asset (or same asset), it changes RP to be covered by 2 colls
- initially 1 coll covers debt, then adding lower lp coll changes (or completely replaces) rp 

apologies if this was already incl in the previous pr, i forget

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: Yeah, it's just a niche use-case here, I think fine to leave as is

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: Yes, that's right. I'll write a fuzz test with that

CheyenneAtapour: Added

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,1048 @@
...
+    });
+
+    // Bob's current risk premium should be greater than liquidity premium of wbtc, since debt is not fully covered by it
+    assertGt(
+      spoke1.getUserRiskPremium(bob),
+      spoke1.getLiquidityPremium(_wbtcReserveId(spoke1)),
+      'Bob user rp after borrow'
+    );
+
+    // Mock calls to ensure dai debt does not does not grow due to interest
```
DhairyaSethi: kinda funky workaround, how about instead we make wbtc's index so high it grows faster than the debt and then eventually is enough to cover entire debt?

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,1048 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+contract SpokeRiskPremiumEdgeCasesTest is SpokeBase {
+  using SharesMath for uint256;
+  using WadRayMath for uint256;
+  using PercentageMath for uint256;
```
miguelmtzinf: probably unneeded

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,1048 @@
...
+    updateLiquidityPremium(spoke2, _usdxReserveId(spoke2), 10_00);
+    assertLt(
+      spoke2.getLiquidityPremium(_usdxReserveId(spoke2)),
+      spoke2.getLiquidityPremium(_daiReserveId(spoke2)),
+      'Usdx lower lp than dai'
+    );
+
+    daiSupplyAmount = bound(daiSupplyAmount, 4, MAX_SUPPLY_AMOUNT);
+    borrowAmount = bound(borrowAmount, 3, MAX_SUPPLY_AMOUNT / 2);
+    // Force least lp asset supply amount to be less than borrow amount, so borrow covered by 2 collaterals at least
```
miguelmtzinf, line 29: does it make sense to dynamically calculate supply amounts based on borrow amount to make sure any of the two collateral assets can cover the debt?
if we change the price of this assets in the setup this won't work

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,1048 @@
...
+    // Here it will be due to decimals of usdx vs dai
+    usdxSupplyAmount = bound(usdxSupplyAmount, 1, borrowAmount);
+    repayAmount = bound(repayAmount, 2, borrowAmount);
+
+    // Deal bob dai to cover dai and dai2 supply
+    deal(address(tokenList.dai), bob, MAX_SUPPLY_AMOUNT * 2);
+
+    // Supply max dai2, the highest lp asset, to allow borrowing without affecting RP
```
miguelmtzinf, line 40: why do we need this if the other collateral assets are enough to cover the debt? it's because they can be greater than the debt but not enough for the HF?

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
miguelmtzinf: cannot we merge the 2 of them into 1? with gte

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,1048 @@
...
+    });
+
+    assertGt(
+      spoke2.getUserRiskPremium(bob),
+      riskPremium,
+      'Risk premium should increase after disabling lower LP reserve as collateral'
+    );
+  }
+
+  /// Supply dai2 and dai as collateral, borrow dai2, then withdraw dai as collateral and risk premium should increase
```
miguelmtzinf: borrowing dai2 does not matter right? we can say just 1 debt and user any other asset (so it's more generic)

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
miguelmtzinf: yeah, better to move wbtc utilization up and debt utilization down (or just mock borrow rates)

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,1048 @@
...
+      'Bob user risk premium after collateral accrual'
+    );
+  }
+
+  /// Debt is initially covered by 2 collaterals (dai + dai2), then 1 collateral becomes enough to cover the debt due to interest accrual
+  function test_riskPremium_fuzz_nonIncreasesAfterCollateralAccrual(
+    uint256 daiSupplyAmount,
+    uint40 skipTime
+  ) public {
+    daiSupplyAmount = bound(daiSupplyAmount, 1e18, MAX_SUPPLY_AMOUNT / 2 - 2); // Leave some room to allow borrowing more than this supply, and for Alice borrow
```
miguelmtzinf: wdym with this comment?

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: Do you prefer that @miguelmtzinf ? I can do, but I kinda like showing the strictly increasing case

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: That's the motivation, yeah. There are other ways to do it, but I have been using this pattern elsewhere in user rp tests, so stuck with it

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: For sure, I can lean towards these comments being more generic. Just wanted to explicitly describe this test

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: I need to borrow more dai than this supplied amount as per the test condition to have debt covered by 2 collaterals. I am assuming a borrow cap of `MAX_SUPPLY_AMOUNT / 2`, and I also need Alice to borrow 1 dai to accrue debt interest. That's why there is a `-2`, to not go over the borrow cap

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: Added these

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: Changed

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: ty, removed

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: Done, added a helper function `_calcEquivalentAssetAmount` for this in base

tests/Base.t.sol
```diff
@@ -747,6 +748,27 @@ abstract contract Base is Test {
...
+
+  /// @dev Helper function to calculate the equivalent asset amount for a given asset
+  function _calcEquivalentAssetAmount(
+    uint256 inputAssetId,
+    uint256 inputAssetAmount,
+    uint256 outputAssetId
+  ) internal view returns (uint256) {
+    uint256 valueOfInputAsset = _getValueInBaseCurrency(inputAssetId, inputAssetAmount);
+    uint256 valueOfWeiOutput = _getValueInBaseCurrency(outputAssetId, 1);
+    return valueOfInputAsset / valueOfWeiOutput;
```
miguelmtzinf, line 775: this won't work if `valueOfWeiOutput` is > `valueOfInputAsset`
it's an extreme case but worth handling

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
miguelmtzinf: shouldn't we use it here to make sure dai or usdx amounts cover the  debt?

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
miguelmtzinf: probably missing something. how do you make the first being gt and second ge?

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
miguelmtzinf: it's just that supply and borrow the same asset looks like a very specific case. I d borrow a different asset to make the test more generic

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: In gt case, I withdraw **all** of the lower lp collateral. In the ge case, I fuzz the withdraw amount, so withdrawing e.g. 1 wei of the asset may not be enough to actually change user rp (due to rounding)

tests/Base.t.sol
yan-man: i had to do something similar in liquidations - created the opposite method to convert base currency into amount: https://github.com/aave/aave-v4/blob/eeb781c61a3fe9ba9d6b4972ed007e2c7d50afa1/tests/Base.t.sol#L959

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -229,14 +290,39 @@ contract SpokeRiskPremiumEdgeCasesTest is SpokeBase {
...
+      totalDebt: baseDebt,
+      reserveFactor: 0,
+      assetId: wbtcAssetId,
+      virtualUnderlyingBalance: wbtcAsset.availableLiquidity,
+      usingVirtualBalance: true
+    });
+    vm.mockCall(
+      address(irStrategy),
+      abi.encodeWithSelector(IReserveInterestRateStrategy.calculateInterestRates.selector, params),
+      abi.encode(1000e27)
```
yan-man: would you mind updating to conventions from other places
```suggestion
      abi.encode(uint256(10_00).bpsToRay())
```

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -0,0 +1,1140 @@
...
+      amount: wbtcSupplyAmount,
+      onBehalfOf: bob
+    });
+
+    // Now risk premium should be less than or equal to LP of weth
+    assertLe(
+      spoke1.getUserRiskPremium(bob),
+      spoke1.getLiquidityPremium(_wethReserveId(spoke1)),
+      'Bob user risk premium after supplying lower LP collateral'
+    );
```
yan-man, line 759: wondering if on all these edge case tests whether it's worthwhile also assert the user rp is also eq to `_calculateExpectedUserRP` output, given you already have that helper available, rather than merely directional (Le, Ge, etc)

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
yan-man, line 1: yeah this test file got quite large, maybe can split into `EdgeCases.unit.t.sol` and `EdgeCases.fuzz.t.sol`?

CheyenneAtapour: I just pushed the pattern where I call into the fuzz tests with particular edge case params. It's down to 764 lines. I think it's alright?

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: imho here it's really just meant to be the math that's tested, and it's clean that dai and dai2 have the same decimals. The conversions kinda make the tests messy. And in fact, this test suite is meant to be the edge cases, so I'd also posit that it's appropriate to test this kind of thing here

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
yan-man: but on the flip side, if we had 2 colls exactly covering a debt
- collA: 1e5, LP = 10%
- collB: 1 wei, LP = 5%

the user rp would be skewed such that collB should only contribute an amount that rounds to 0 bc it is less than 1 BPS, so that user rp should be 10_00 BPS. But due to precision it'd instead be 99_99 BPS. Is that correct? I ask bc I saw this also in dealing with wAvg collateral factor on HF calcs

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: Added

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
DhairyaSethi: all coll values are normalised to base currency (18 + 8 decimals) tho

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
DhairyaSethi: agree think it's fine now

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
CheyenneAtapour: @yan-man I just tested this case. It's correct though that user risk premium should be 9_99 BPS. And it is indeed the case

PR #374, opened 2025-04-28
feat: index diff accounting

DhairyaSethi:


tests/ts/prototype/core.ts
```diff
@@ -236,8 +236,11 @@ export class LiquidityHub {
     return this.toDrawnShares(assets);
   }
 
-  previewOffset(premiumDrawnShares: bigint) {
-    return this.toDrawnAssets(premiumDrawnShares);
+  previewAccruedPremium(shares: bigint, trailingIndex: bigint) {
+    this.accrue();
+    const nextIndex = this.assetNormalizedDebt();
+    const currentIndex = trailingIndex;
+    return rayMul(shares, nextIndex - currentIndex, Rounding.CEIL);
```
yan-man, line 258: my fear is sum across user/spoke premium not matching up with total asset outstanding premium due to precision losses on each spoke's currentIndex. We can either round up or down 

(guessing)
- up: then sum across spoke premium > asset premium (last spoke repays underflows)
- down: then sum across spoke premium < asset premium (last spoke repays will have remaining asset premium dust) 

(in a vacuum, we preferred the latter right?) do you agree with these potential outcomes?

Issue #190, opened 2025-02-19
custom reserve factor per spoke


DhairyaSethi:

miguelmtzinf:
@DhairyaSethi per spoke? reserve factor is linked to the Hub.
I guess this is quite old and can be closed now?

Issue #346, opened 2025-04-16
Withdraw greater than current amount


yan-man:
when withdrawing, putting in an amount param >= current supplied amount will withdraw all. Not just uint256 max. Similar behavior as for repay 

ref:
- https://github.com/aave/aave-v4/issues/339
- https://github.com/aave/aave-v4/pull/343

PR #375, opened 2025-04-29
Add treasury fee accrual via reserve factor

### LINKED ISSUE #39: Add reserve factor to LH
- The DAO takes a cut from the yield payed by borrowers
- This cut can be taken fin form of shares.

(we can check how Morpho does it, it must be pretty similar)

### LINKED ISSUE #438: Updates of reserve factor or IR of an asset in the Hub require interest accrual
Updating immediately without accruing interest first would apply the new RF or IR retroactively until the next action, which is unfair. DAO-approved changes are intended to take effect immediately.

miguelmtzinf:
Closes #39 
Partially close #438 

## Design Overview
- Treasury is a new spoke, added per asset in LH, with max supply/draw caps.
  - DAO can use it to supply more or even borrow, with an isolated entrypoint/spoke.
  - It's per asset. Every time an asset is added to the hub, must come with oracle, IR and treasury config.
  - A change of the treasury involves the accrual of fees and restriction of activity of the old one.
- Note that debt exchange rate only depends on debt index, whereas supply exchange rate depends on supply shares / assets.
- Debt index reflects the increase of the drawn assets. This is used by the debt exchange rate.
- Fee accrual occurs by applying the reserve factor to the interest accrued and storing the result in a separate variable, which is later converted into `feeShares`.
- Typically, shares are static, but this no longer applies to the treasury, as its shares grow over time with interest accrual. All shares and assets calculations account for interest accrual and the reserve factor cut, and this is also exposed for retrieving the shares and assets of the spoke treasury.
- Configuration of fees is done in a single update function that makes sure that accrued fees do not go to zero address. The property is that if liquidity fee is non-zero, fee receiver shouldn't be zero.

## Other Notes
- Treasury spoke per asset or across assets. If we want to enforce 1 single treasury for all assets:
  - `updateTreasury` does not accept `assetId`, but it iterates over all
  - Every time an `addAsset`is executed, the `treasurySpoke` must be updated. This can be added to the `addAsset` fn itself.   
  - Hub more rigid, but we still have flexibility at the treasury spoke level.
  - Still a bit counter-intuitive the fact that treasury is for all assets, and spoke is per asset.

## TODO
- Remove any notion of reserve factor from the IR calculation, since it's not needed -> #443 
- Add a new function to accumulate liquidity fees from authorized spokes -> #449
- Evaluate the need of event emission on liquidity fees accrual
- Enrich TreasurySpoke functionality -> #450
- https://github.com/aave/aave-v4/issues/451
- Better tests in `SpokeAccrueLiquidityFeeTest`, kinda blocked by #229


src/contracts/TreasurySpoke.sol
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    HUB.remove(assetId, amount, to);
+  }
+
+  function getSuppliedAmount(uint256 assetId) external view returns (uint256) {
+    return HUB.getSpokeSuppliedAmount(assetId, address(this));
+  }
+
+  function gerSuppliedShares(uint256 assetId) external view returns (uint256) {
```
yan-man:
```suggestion
  function getSuppliedShares(uint256 assetId) external view returns (uint256) {
```

src/libraries/logic/AssetLogic.sol
```diff
@@ -121,19 +129,28 @@ library AssetLogic {
...
-      return baseDebtIndex;
+      return (previousIndex, 0);
+    }
+    uint256 newIndex = previousIndex.rayMulUp(
+      MathUtils.calculateLinearInterest(asset.baseBorrowRate, uint40(lastUpdateTimestamp))
+    );
+    uint256 feesIndex = newIndex;
+    uint256 reserveFactor = asset.config.reserveFactor;
+    if (reserveFactor > 0) {
+      feesIndex = newIndex.percentMulDown(PercentageMath.PERCENTAGE_FACTOR);
```
yan-man: is this meant to be percentMulDown `reserveFactor`

if we do += above on `feesIndex` in `accrue` to accumulate, shouldn't this be the reserveFactor % of the accumulated interest only, ie index delta (newIndex - previousIndex)?

tests/Base.t.sol
```diff
@@ -140,10 +142,17 @@ abstract contract Base is Test {
   }
 
   function initEnvironment() internal {
+    configureTreasury();
     deployMintAndApproveTokenList();
     configureTokenList();
   }
 
+  function configureTreasury() internal {
+    // alternative is using standard `LiquidityHub.addSpoke` function
```
yan-man: this would be a bit diff right, bc `updateTreasury` is basically across all assets but `addSpoke` would need to be per-asset?

and would need to add additional logic into `addSpoke` to update treasury address?

src/contracts/LiquidityHub.sol
```diff
@@ -27,6 +27,8 @@ contract LiquidityHub is ILiquidityHub {
   IERC20[] public assetsList; // TODO: Check if Enumerable or Set makes more sense
   uint256 public assetCount;
 
+  address public treasury;
```
CheyenneAtapour: Do we want to set a treasury address upon deployment in a constructor, or better to keep as is and set within the deploying proposal payload?

Order would be something like: Deploy hub, deploy treasury spoke, set treasury spoke on hub

src/contracts/TreasurySpoke.sol
```diff
@@ -0,0 +1,37 @@
...
+contract TreasurySpoke is Ownable, ITreasurySpoke {
+  ILiquidityHub public immutable HUB;
+
+  constructor(address owner, address hubAddress) Ownable(owner) {
+    require(hubAddress != address(0), InvalidHubAddress());
+
+    HUB = ILiquidityHub(hubAddress);
+  }
+
+  function supply(uint256 assetId, uint256 amount) external onlyOwner {
```
CheyenneAtapour: Maybe I'm missing something, but what is the purpose of the treasury spoke supplying?

src/libraries/logic/AssetLogic.sol
miguelmtzinf: yes, this in an oversight.

src/contracts/LiquidityHub.sol
yan-man: yeah agree, otherwise we'll need to add some more logic to check for zero-address treasury, etc. which I don't think we have atm

src/contracts/LiquidityHub.sol
```diff
@@ -106,14 +110,25 @@ contract LiquidityHub is ILiquidityHub {
...
+
   /// @inheritdoc ILiquidityHub
   function add(uint256 assetId, uint256 amount, address from) external returns (uint256) {
     // TODO: authorization - only spokes
 
     DataTypes.Asset storage asset = _assets[assetId];
     DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
 
-    asset.accrue();
+    asset.accrue(_spokes[assetId][treasury]);
```
yan-man: need some validation for treasury that is not yet set and still zero addr?

src/contracts/LiquidityHub.sol
```diff
@@ -462,6 +477,7 @@ contract LiquidityHub is ILiquidityHub {
     require(asset != address(0), InvalidAssetAddress());
     require(address(config.irStrategy) != address(0), InvalidIrStrategy());
     require(config.decimals <= MAX_ALLOWED_ASSET_DECIMALS, InvalidAssetDecimals());
+    require(config.reserveFactor <= PercentageMath.PERCENTAGE_FACTOR, InvalidReserveFactor());
```
yan-man:
```suggestion
    require(config.reserveFactor <= PercentageMathExtended.PERCENTAGE_FACTOR, InvalidReserveFactor());
```

src/contracts/TreasurySpoke.sol
```diff
@@ -0,0 +1,37 @@
...
+    HUB.add(assetId, amount, msg.sender);
+  }
+
+  function withdraw(uint256 assetId, uint256 amount, address to) external onlyOwner {
+    // If uint256.max is passed, withdraw all supplied assets
+    if (amount == type(uint256).max) {
```
yan-man, line 38: should this be `amount > HUB.getSpokeSuppliedAmount(assetId, address(this))` to align with repay all: https://github.com/aave/aave-v4/pull/239#discussion_r1984829110

we were intending to do this anyway on normal spoke withdraw later anyway

src/libraries/logic/AssetLogic.sol
```diff
@@ -119,20 +125,39 @@ library AssetLogic {
   }
 
   // @dev Utilizes existing `asset.baseBorrowRate`
-  function accrue(DataTypes.Asset storage asset) internal {
-    asset.baseDebtIndex = asset.previewIndex();
+  function accrue(DataTypes.Asset storage asset, DataTypes.SpokeData storage treasury) internal {
+    uint256 drawnIndex = asset.previewDrawnIndex();
```
yan-man, line 136: should we escape here if `drawnIndex == asset.baseDebtIndex`?

atm for txs in the same block, i think we dupe all storage updates/calcs below this point

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
```diff
@@ -357,37 +357,41 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
...
     assertEq(config.frozen, false, 'asset frozen');
     assertEq(config.paused, false, 'asset paused');
+    assertEq(config.decimals, 18, 'asset decimals');
+    assertEq(config.reserveFactor, 5_00, 'reserve factor');
     assertEq(address(config.irStrategy), address(irStrategy), 'asset irStrategy');
   }
 
   function test_addAsset_fuzz(DataTypes.AssetConfig memory newConfig, address asset) public {
     newConfig.decimals = bound(newConfig.decimals, 0, hub.MAX_ALLOWED_ASSET_DECIMALS());
+    newConfig.reserveFactor = bound(newConfig.reserveFactor, 0, PercentageMath.PERCENTAGE_FACTOR);
```
yan-man:
```suggestion
    newConfig.reserveFactor = bound(newConfig.reserveFactor, 0, PercentageMathExtended.PERCENTAGE_FACTOR);
```

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
```diff
@@ -411,18 +416,65 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
     bool active,
     bool frozen,
     bool paused,
+    uint256 reserveFactor,
     IReserveInterestRateStrategy irStrategy
   ) public {
-    uint256 invalidDecimals = hub.MAX_ALLOWED_ASSET_DECIMALS() + 1;
+    reserveFactor = bound(reserveFactor, 0, PercentageMath.PERCENTAGE_FACTOR);
```
yan-man:
```suggestion
    reserveFactor = bound(reserveFactor, 0, PercentageMathExtended.PERCENTAGE_FACTOR);
```

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
```diff
@@ -491,19 +549,22 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
     bool frozen,
     bool paused,
     address token,
-    uint256 decimals
+    uint256 decimals,
+    uint256 reserveFactor
   ) public {
     decimals = bound(decimals, 0, hub.MAX_ALLOWED_ASSET_DECIMALS());
+    reserveFactor = bound(reserveFactor, 0, PercentageMath.PERCENTAGE_FACTOR);
```
yan-man:
```suggestion
    reserveFactor = bound(reserveFactor, 0, PercentageMathExtended.PERCENTAGE_FACTOR);
```

src/libraries/logic/AssetLogic.sol
```diff
@@ -119,20 +125,39 @@ library AssetLogic {
...
+      return previousIndex;
     }
     return
-      baseDebtIndex.rayMulUp(
+      previousIndex.rayMulUp(
         MathUtils.calculateLinearInterest(asset.baseBorrowRate, uint40(lastUpdateTimestamp))
       );
   }
+
+  function previewFeesShares(
```
CheyenneAtapour: small nit / suggestion, but what about naming it just `previewFeeShares` to not have the double "s"?

src/contracts/LiquidityHub.sol
```diff
@@ -64,11 +67,13 @@ contract LiquidityHub is ILiquidityHub {
     _validateAssetConfig(config, address(assetsList[assetId]));
     DataTypes.Asset storage asset = _assets[assetId];
     // TODO: AccessControl
+    // todo: if reserveFactor or irStrategy update, accrue interest
```
miguelmtzinf: created a ticket for this: https://github.com/aave/aave-v4/issues/438

src/contracts/LiquidityHub.sol
miguelmtzinf: I think it's fine to keep it separate, since it's not essential for the pool's core functionality.
Although we will for sure receive a security finding because of this :dagger:

src/contracts/TreasurySpoke.sol
miguelmtzinf: DAO can use it to supply into V4 through this spoke.

tests/unit/Spoke/Spoke.ReserveFactor.t.sol
```diff
@@ -0,0 +1,266 @@
...
+    expectedBaseDebt += 750e18; // 50% of 1500 (base debt accrual)
+    expectedPremiumDebt += 0;
+    expectedTreasuryFees += 37.5e18; // 5% of 750 (reserve factor on base debt)
+
+    skip(365 days);
+
+    assertDebtEq(
+      spoke1,
+      reserveId,
+      expectedBaseDebt,
```
miguelmtzinf: failing here

tests/unit/Spoke/Spoke.ReserveFactor.t.sol
```diff
@@ -0,0 +1,266 @@
...
+      hub.getSpokeSuppliedShares(assetId, address(treasurySpoke)),
+      hub.convertToSuppliedShares(assetId, expectedTreasuryFees),
+      'treasury fees'
+    );
+
+    // 0% premium
+    updateLiquidityPremium(spoke1, reserveId, 0);
+    spoke1.updateUserRiskPremium(alice);
+
+    // Bob supplies 1 share to trigger interest accrual
```
CheyenneAtapour: nit: Alice is supplying btw, not Bob

tests/Base.t.sol
miguelmtzinf: I made treasury a real spoke with proper config. Think it makes more sense and if we do access control by using the mapping in some way, we d need it.

src/contracts/LiquidityHub.sol
miguelmtzinf: Added conditional in `AssetLogic.accrue()` fn so it only adds fees as `suppliedShares` to the treasurySpoke and asset if those are non-zero. Fee shares are not zero if reserve factor is not zero.
And treasury must be initialized if the reserve factor is not zero.

New invariants:
- if treasury is 0, reserve factor is 0
- if reserve factor is 0, fee shares in accrue each time is 0

src/contracts/TreasurySpoke.sol
miguelmtzinf: to be tackled in https://github.com/aave/aave-v4/issues/346

src/libraries/types/DataTypes.sol
```diff
@@ -40,6 +40,7 @@ library DataTypes {
     bool frozen;
     bool paused;
     uint256 decimals;
+    uint256 reserveFactor;
```
avniculae: maybe call it `assetFactor`?

avniculae: Created #446

src/libraries/logic/AssetLogic.sol
miguelmtzinf: escaping in `previewFeeShares`

src/libraries/types/DataTypes.sol
miguelmtzinf: Renamed to `liquidityFee`

src/contracts/LiquidityHub.sol
miguelmtzinf: This is fixed thanks to the new `updateAssetFee` update fn

src/contracts/LiquidityHub.sol
```diff
@@ -27,6 +27,9 @@ contract LiquidityHub is ILiquidityHub {
   IERC20[] public assetsList; // TODO: Check if Enumerable or Set makes more sense
   uint256 public assetCount;
 
+  // Mapping of fee recipients by asset identifier
+  mapping(uint256 => address) internal feeReceivers;
```
```suggestion
  mapping(uint256 address => address feeReceiver) internal feeReceivers;
```
also wondering whether this should just live in DataTypes.Asset
also also, not a fan of `feeReceiver` tbh, prefer treasury even if we call reserve factor as liquidityFee, but nbd

src/contracts/LiquidityHub.sol
```diff
@@ -107,14 +121,70 @@ contract LiquidityHub is ILiquidityHub {
...
+          _updateSpokeConfig(
+            assetId,
+            feeReceiver,
+            DataTypes.SpokeConfig({supplyCap: type(uint256).max, drawCap: type(uint256).max})
+          );
+        }
+      }
+      feeReceivers[assetId] = feeReceiver;
+      emit FeeReceiverUpdated(assetId, oldFeeReceiver, feeReceiver);
+    }
```
DhairyaSethi, line 174: with these nested if-else's and separate events, makes sense to just write two methods ie updateFeeRec & updateLiqFee instead, with the only overlap bw the two being if you're unsetting feeReceiver then you should've already set liq fee to 0 (bc the other way around is not safe but this is)
both would accrue and that's fine. reduced complexity. wdyt?

src/contracts/LiquidityHub.sol
```diff
@@ -226,56 +296,55 @@ contract LiquidityHub is ILiquidityHub {
...
   }
 
   function _refresh(
     uint256 assetId,
     address spokeAddress,
     int256 premiumDrawnShareDelta,
     int256 premiumOffsetDelta,
-    int256 realizedPremiumDelta
+    uint256 realizedPremiumAdded,
+    uint256 realizedPremiumTaken
```
DhairyaSethi, line 326: a comment somewhere which specifies the need for this discrepancy (added + taken, vs delta) would be appreciated - it's such that we can do debt decreased by premium restored amt check iirc, right?
wondering if this limits the spoke usecases

src/contracts/LiquidityHub.sol
```diff
@@ -309,13 +378,27 @@ contract LiquidityHub is ILiquidityHub {
     return _assets[assetId].toSuppliedAssetsDown(shares);
   }
 
+  function convertToSuppliedAssetsUp(
```
DhairyaSethi, line 382: leftover from testing i assume? pls rm

src/contracts/LiquidityHub.sol
```diff
@@ -309,13 +378,27 @@ contract LiquidityHub is ILiquidityHub {
...
+  }
+
   function convertToSuppliedShares(
     uint256 assetId,
     uint256 assets
   ) external view returns (uint256) {
     return _assets[assetId].toSuppliedSharesDown(assets);
   }
 
+  function convertToSuppliedSharesUp(
```
DhairyaSethi, line 396: pls rm

src/contracts/LiquidityHub.sol
```diff
@@ -448,18 +553,18 @@ contract LiquidityHub is ILiquidityHub {
...
       premiumDrawnShares: 0,
       premiumOffset: 0,
       realizedPremium: 0,
-      lastUpdateTimestamp: 0,
+      lastUpdateTimestamp: block.timestamp,
       config: config
     });
 
-    emit SpokeAdded(assetId, spoke);
+    emit SpokeAdded(assetId, spoke); // todo: emit config
```
DhairyaSethi, line 565: maybe we call _updateSpokeConfig in this method, which would also simplify one layer if your if else branches in `updateAssetFees` and omit needing config on SpokeAdded

src/dependencies/openzeppelin/Ownable.sol
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.0.0) (access/Ownable.sol)
```
DhairyaSethi, line 2: all OZ deps are v5.3

src/interfaces/ILiquidityHub.sol
```diff
@@ -94,10 +103,21 @@ interface ILiquidityHub {
     DataTypes.SpokeConfig memory config
   ) external;
 
+  /**
+   * @notice Updates the fee configuration for a specified asset.
+   * @dev Accrues asset fees to the current receiver before applying any updates.
+   * @dev Disables the old fee receiver as spoke by setting its caps to zero.
+   * @dev The new fee receiver cannot be zero if the liquidity fee is non-zero.
+   * @param assetId The identifier of the asset.
```
DhairyaSethi, line 104:
```suggestion
   * @param assetId The asset identifer.
```
reads better imo (same for below) but nbd

src/interfaces/ILiquidityHub.sol
```diff
@@ -145,47 +165,37 @@ interface ILiquidityHub {
    * @notice Refreshes premium debt accounting.
    * @dev To be called when moving accrued premium to realized premium.
    * @dev Only callable by spokes.
-   * @dev Total debt should not change, reverts with `InvalidDebtChange` when violated.
-   * @param assetId The asset id.
+   * @dev Premium debt can only decrease by at most the amount of realized premium taken.
```
DhairyaSethi, line 161: been rooting for this from day 1 of the latest accounting, ty ❤️

src/contracts/TreasurySpoke.sol
```diff
@@ -0,0 +1,57 @@
...
+   */
+  constructor(address owner_, address hub_) Ownable(owner_) {
+    require(hub_ != address(0), InvalidHubAddress());
+
+    HUB = ILiquidityHub(hub_);
+  }
+
+  /// @inheritdoc ITreasurySpoke
+  function supply(uint256 reserveId, uint256 amount) external onlyOwner {
+    HUB.add(reserveId, amount, msg.sender);
```
DhairyaSethi, line 32: is it fair to call this param reserveId? i get natspec talk about this needing to be equal to assetId on hub bc there's no reserve registrations but i would still add comment here

src/libraries/logic/AssetLogic.sol
```diff
@@ -116,30 +119,80 @@ library AssetLogic {
...
+    DataTypes.Asset storage asset,
+    uint256 indexDelta
+  ) internal view returns (uint256) {
+    uint256 liquidityFee = asset.config.liquidityFee;
+    if (indexDelta == 0 || liquidityFee == 0) {
+      return 0;
+    }
+    uint256 feesAmount = indexDelta
+      .rayMulDown(asset.baseDrawnShares + asset.premiumDrawnShares)
+      .percentMulDown(liquidityFee);
```
DhairyaSethi, line 185:
```suggestion
    uint256 feesAmount = indexDelta
      .rayMulUp(asset.baseDrawnShares + asset.premiumDrawnShares)
      .percentMulDown(liquidityFee);
```
this should mul up for consistency, otherwise we end up with donated amount which is also fine tbh 
```
  function test_maxLiquidationFee() public {
    hub.updateAssetFees(daiAssetId, address(treasurySpoke), 100_00);
    Utils.supplyCollateral(spoke1, _daiReserveId(spoke1), alice, 100e18, alice);
    Utils.borrow(spoke1, _daiReserveId(spoke1), alice, 50e18, alice);
    skip(342 days);
    console.log('treasury amount', hub.getSpokeSuppliedAmount(daiAssetId, address(treasurySpoke)));
    console.log('alice accrued  ', spoke1.getUserTotalDebt(_daiReserveId(spoke1), alice) - 50e18);
  }
Logs:
  treasury amount 2810958904109589040
  alice accrued   2810958904109589042
```

src/contracts/LiquidityHub.sol
```diff
@@ -107,14 +121,70 @@ contract LiquidityHub is ILiquidityHub {
     emit SpokeConfigUpdated(assetId, spoke, config.drawCap, config.supplyCap);
   }
 
+  /// @inheritdoc ILiquidityHub
+  function updateAssetFees(uint256 assetId, address feeReceiver, uint256 liquidityFee) public {
```
DhairyaSethi, line 124: naming is a little misleading, would also add bps suffix to liquidityFee or something

tests/Base.t.sol
```diff
@@ -862,11 +889,29 @@ abstract contract Base is Test {
 
   /// @dev Helper function to calculate asset amount corresponding to single supplied share
   function minimumAssetsPerSuppliedShare(uint256 assetId) internal view returns (uint256) {
-    return hub.convertToSuppliedAssets(assetId, 1);
+    return hub.convertToSuppliedAssetsUp(assetId, 1);
+  }
+
+  /// @dev Helper function to calculate expected supplied assets based on amount to supply and current exchange rate
```
DhairyaSethi, line 901:
```suggestion
  /// @dev Helper function to calculate expected supplied assets based on amount to supply and current exchange rate taking potential donation into account
```

tests/unit/LiquidityHub/LiquidityHub.UpdateTreasury.t.sol
```diff
@@ -0,0 +1,257 @@
...
+import 'tests/unit/LiquidityHub/LiquidityHubBase.t.sol';
+
+contract LiquidityHubUpdateAssetFeesTest is LiquidityHubBase {
+  /// Triggers accrual always, based on old config
+  function test_updateAssetFees_fuzz_always_accrue(
+    address feeReceiver,
+    uint256 liquidityFee
+  ) public {
+    liquidityFee = bound(liquidityFee, 0, PercentageMathExtended.PERCENTAGE_FACTOR);
+    if (liquidityFee > 0) vm.assume(feeReceiver != address(0));
```
DhairyaSethi: i would run this fuzz test again inside a unit test once more to ensure these cases are covered (bc they may be missed by the fuzzer); in general covering more than one branch utilising a fuzzed params should be avoided

tests/unit/LiquidityHub/LiquidityHub.UpdateTreasury.t.sol
```diff
@@ -0,0 +1,257 @@
...
+    // reverts if zero receiver and non-zero current fee
+    uint256 currentLiquidityFee = _getLiquidityFee(assetId);
+    vm.expectRevert(ILiquidityHub.InvalidFeeReceiver.selector);
+    hub.updateAssetFees(assetId, address(0), currentLiquidityFee);
+  }
+
+  /// Liquidity fee cannot be above maximum
+  function test_updateAssetFees_revertsWith_InvalidLiquidityFee() public {
+    uint256 assetId = daiAssetId;
+    address validReceiver = address(1);
```
DhairyaSethi: address(1) is ecrecover, would do vm.randomAddress() for sanity but nbd here

tests/unit/LiquidityHub/LiquidityHub.UpdateTreasury.t.sol
```diff
@@ -0,0 +1,257 @@
...
+      assetId,
+      currentFeeReceiver,
+      type(uint256).max,
+      type(uint256).max
+    );
+
+    vm.expectEmit(address(hub));
+    emit ILiquidityHub.FeeReceiverUpdated(assetId, newFeeReceiver, currentFeeReceiver);
+
+    // todo: LiquidityFeeUpdated not emitted
```
DhairyaSethi: can use `vm.expectEmit({emitter: address(hub), count: 0});`

tests/unit/Spoke/Spoke.LiqudityFee.t.sol
```diff
@@ -0,0 +1,454 @@
...
+    skip(365 days);
+
+    _assertSpokeDebt(
+      spoke1,
+      reserveId,
+      expectedBaseDebt,
+      expectedPremiumDebt,
+      'after base debt accrual'
+    );
+    console.log(hub.getSpokeSuppliedAmount(assetId, address(treasurySpoke)));
```
DhairyaSethi: pls rm

tests/unit/Spoke/Spoke.LiqudityFee.t.sol
```diff
@@ -0,0 +1,454 @@
...
```
DhairyaSethi: pls rm

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -0,0 +1,123 @@
...
+      amount + interestAndFees,
+      address(treasurySpoke)
+    );
+  }
+
+  function _treasurySpoke() internal view returns (ISpoke) {
+    return ISpoke(address(treasurySpoke));
+  }
+
+  // todo: test that supplying from treasury does not create any issue. existing fees are added to the supply amount
```
DhairyaSethi, line 122: also when treasury fee is 100%

src/contracts/LiquidityHub.sol
miguelmtzinf: This is a good point. I d move it to DataTypes.Asset after #447 . It d be way simpler.
Created https://github.com/aave/aave-v4/issues/451

miguelmtzinf: any other suggestion for `feeReceiver`? do not like treasury at core level tbh

src/contracts/LiquidityHub.sol
miguelmtzinf: Couple of notes:
1. it all started with 2 separate fns and thought merging them was a good idea to make sure config is consistent/correct. Happy to split again if we think it would be better.
2. Great part of the complexity comes from caps changes of the treasury spoke. We can simplify but just creating a new spoke with no caps if new `feeReceiver`. (stopping activity of the old receiver would need to be called separately)

src/contracts/LiquidityHub.sol
miguelmtzinf: The RF functionality does not need this tbf, but added this change during the dev process.
Imo it's good to do because allow us to merge settlePremium and refresh into one fn.
I don't see this as a limiting factor because it's equivalent to passing a delta

src/contracts/LiquidityHub.sol
miguelmtzinf: no, it's used for the `minimumAssetsPerSuppliedShare` helper. This function returns how much assets user needs to supply for at least 1 share.
I wanna add converter functions with both rounding directions because they are useful. (needs renaming)

src/contracts/LiquidityHub.sol
miguelmtzinf: nope, this is needed to preview how much shares would be removed when withdrawing...

src/contracts/LiquidityHub.sol
miguelmtzinf: hmm yeah, good call
In any way, we should be consistent with `addAsset` as much as we can. There we are doing:
```
    emit AssetAdded(assetId, asset);
    emit AssetConfigUpdated(assetId, config);

```

src/dependencies/openzeppelin/Ownable.sol
miguelmtzinf: I don't think so:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.1.0/contracts/utils/Address.sol#L2
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.3.0/contracts/utils/Address.sol#L2

src/contracts/TreasurySpoke.sol
miguelmtzinf: It is said on every function natspec, and at contract definition too. I think adding it inside every fn is overkill

tests/unit/LiquidityHub/LiquidityHub.UpdateTreasury.t.sol
miguelmtzinf: created a new case for this.

tests/unit/LiquidityHub/LiquidityHub.UpdateTreasury.t.sol
miguelmtzinf: changed to `makeAddr` just to avoid collisions

tests/unit/LiquidityHub/LiquidityHub.UpdateTreasury.t.sol
miguelmtzinf: we need new cheatcodes for this

tests/unit/Spoke/TreasurySpoke.t.sol
miguelmtzinf: created a ticket for this: https://github.com/aave/aave-v4/issues/452

src/contracts/LiquidityHub.sol
miguelmtzinf: I see little benefit from splitting: https://gist.github.com/miguelmtzinf/4b7054269c0a1dd85f4fa4a0068326e8

src/contracts/LiquidityHub.sol
miguelmtzinf: i did not add it for consistency. There are other config params that use `bps` or `ray` and don't clarify it in the name...

tests/unit/LiquidityHub/LiquidityHub.UpdateTreasury.t.sol
DhairyaSethi: we want to assert it's not been emitted, right? expect emit with count 0 achieves that, otherwise can also recordLogs and assert non existence there

LE: ah sorry i forgot you're on the prev foundry version here, as discussed will raise a separate PR for this

src/dependencies/openzeppelin/Ownable.sol
DhairyaSethi: i meant our local deps should be last updated for 5.3, for which looks like ownable was last updated on 5.0 which is cool. i think we should use Ownable2Step instead but can be done later

src/libraries/logic/AssetLogic.sol
miguelmtzinf: added a test for 100% fee, but need more cases (https://github.com/aave/aave-v4/issues/452)

miguelmtzinf: Re rounding direction,
- I decided to use `rayMulDown` because it's safer: leaving some donation/dust instead of bad debt
- `percentMulDown` because if the `feeAmount` is v small we don't care

src/contracts/LiquidityHub.sol
DhairyaSethi: i think updateLiquidationFee checks should just live in updateAssetConfig, and for consistency we do not emit old values (ie oldFeeReceiver, oldLiquidityFee) and we don't emit specific config events (LiquidityFeeUpdated) since all config are expected to be part of AssetConfigUpdated
also `feeReceivers` should be prefixed with an underscore

i would do something like: https://github.com/aave/aave-v4/compare/feat/39-reserve-factor...opt/upd-fee-rec?expand=1
note imo we don't if (key != 0) state[key] = newState; we can just directly write and retain the noop behaviour when current state is already in desired state

DhairyaSethi: note that we do no need timestamp on SpokeData, and that's the only difference between `_updateSpokeConfig` and `_addSpoke` (apart from input zero addr check) hence it's best to merge them to make this cleaner

src/contracts/LiquidityHub.sol
DhairyaSethi: idts? the helper uses global exchange rate to enforce rounding 
image
i agree they're useful but tbh i don't think we should merge with this naming, we can when required when we refactor the getter names

src/contracts/LiquidityHub.sol
DhairyaSethi: same rationale as https://github.com/aave/aave-v4/pull/375#discussion_r2149481218

src/libraries/logic/AssetLogic.sol
DhairyaSethi: yeah agree it does seem safer, we can analyse if roundUp can over result in an overflow (of owed to treasury - owed by borrowers) but i don't like another donation vector in the code haha, we'll definitely get a finding for this, i would atleast like to document with a comment so we can revisit
this is only re rayMulDown, totally aligned on percentMul it's irrelevant tbh

miguelmtzinf: added a comment. My doubt is if the 2 interest calc approaches are equivalent...
 In fees accrual we are using the following:
```
indexDelta = nextIndex - currentIndex
indexDelta * (debtShares + premiumShares)
```
while in the rest of the code we are using index accrual for baseDebt and individual premium accrual based on users' rp...

src/contracts/LiquidityHub.sol
miguelmtzinf: i referred to a different helper @DhairyaSethi

DhairyaSethi: oops apologies! but what do you think about using this logic for that helper?

src/libraries/logic/AssetLogic.sol
DhairyaSethi: yeah wondering if it should rather be 
```
baseAccrual = indexDelta * baseShares 
premiumAccrual = nextIndex * premiumShares - offset 
```
bc i think technically `offset != oldIndex * premiumShares` since it's maintained incrementally unlike user position where premiumShares is always overwritten

miguelmtzinf: will create a separate PR for this.

src/contracts/LiquidityHub.sol
```diff
@@ -107,14 +120,70 @@ contract LiquidityHub is ILiquidityHub {
...
+
+    // accrue always, so fees until now are generated based on old config (fee and receiver)
+    address oldFeeReceiver = _assets[assetId].config.feeReceiver;
+    _assets[assetId].accrue(_spokes[assetId][oldFeeReceiver]);
+
+    // Update liquidity fee
+    uint256 oldLiquidityFee = _assets[assetId].config.liquidityFee;
+    if (liquidityFee != oldLiquidityFee) {
+      _assets[assetId].config.liquidityFee = liquidityFee;
+      emit AssetConfigUpdated(assetId, _assets[assetId].config);
```
DhairyaSethi: this can now be emitted twice (see line 171), would recc just emitting once at the end even if nothing changed it's fine tbh

src/contracts/LiquidityHub.sol
```diff
@@ -43,15 +43,17 @@ contract LiquidityHub is ILiquidityHub {
...
+      lastUpdateTimestamp: block.timestamp,
       id: assetId, // todo rm
       config: DataTypes.AssetConfig({
-        decimals: config.decimals, // todo fetch decimals from token
+        feeReceiver: config.feeReceiver, // todo: add validations
         active: config.active,
         frozen: config.frozen,
         paused: config.paused,
+        decimals: config.decimals, // todo fetch decimals from token
+        liquidityFee: 0,
```
avniculae: why is liquidityFee 0 here? shouldn't we set it to `config.liquidityFee`?

src/contracts/LiquidityHub.sol
```diff
@@ -448,18 +549,18 @@ contract LiquidityHub is ILiquidityHub {
...
-    require(spoke != address(0), InvalidSpoke());
+    require(spoke != address(0), InvalidSpoke()); // todo: how to remove spoke
     _spokes[assetId][spoke] = DataTypes.SpokeData({
       suppliedShares: 0,
       baseDrawnShares: 0,
       premiumDrawnShares: 0,
       premiumOffset: 0,
       realizedPremium: 0,
-      lastUpdateTimestamp: 0,
+      lastUpdateTimestamp: block.timestamp,
```
DhairyaSethi, line 561: we can rm this now itself, lastUpdateTimestamp is not needed on SpokeDara
also i thought we agreed to call _updateSpokeConfig here instead?

src/contracts/LiquidityHub.sol
```diff
@@ -107,14 +120,70 @@ contract LiquidityHub is ILiquidityHub {
...
+    // Update fee receiver
+    if (feeReceiver != oldFeeReceiver) {
+      // restrict activity of old receiver, if any
+      if (oldFeeReceiver != address(0)) {
+        _updateSpokeConfig(
+          assetId,
+          oldFeeReceiver,
+          DataTypes.SpokeConfig({supplyCap: 0, drawCap: 0})
+        );
+      }
```
avniculae, line 150: if we do this, then how will the old treasury spoke be able to withdraw the accrued fees that were not withdrawn?

src/contracts/LiquidityHub.sol
```diff
@@ -107,14 +120,70 @@ contract LiquidityHub is ILiquidityHub {
...
+            feeReceiver
+          );
+        } else {
+          _updateSpokeConfig(
+            assetId,
+            feeReceiver,
+            DataTypes.SpokeConfig({supplyCap: type(uint256).max, drawCap: type(uint256).max})
+          );
+        }
+      }
```
avniculae, line 168: I feel like this logic should be present in the Configurator instead, and the hub should only verify that the old spoke has caps set to 0 and that the new spoke was reigstered, wdyt?

avniculae:
should we close #190 ?

src/contracts/LiquidityHub.sol
```diff
@@ -365,11 +451,26 @@ contract LiquidityHub is ILiquidityHub {
     return _assets[assetId].totalSuppliedAssets();
   }
 
+  function getTotalSuppliedShares(uint256 assetId) external view override returns (uint256) {
+    return _assets[assetId].totalSuppliedShares();
+  }
+
   function getSpokeSuppliedAmount(uint256 assetId, address spoke) external view returns (uint256) {
```
avniculae, line 460: this function can just do `_assets[assetId].toSuppliedAssetsDown(getSpokeSuppliedShares(assetId, spoke))`

src/contracts/Spoke.sol
```diff
@@ -272,18 +276,20 @@ contract Spoke is ISpoke, Multicall {
     uint256 userPremiumDrawnShares = userPosition.premiumDrawnShares;
     uint256 userPremiumOffset = userPosition.premiumOffset;
     uint256 userRealizedPremium = userPosition.realizedPremium;
```
avniculae, line 278: I think we can remove this now?

src/contracts/LiquidityHub.sol
miguelmtzinf: let's tackle this on Alex PR #447 please

src/contracts/LiquidityHub.sol
miguelmtzinf: this does not prevent the old spoke to withdraw. (only new supplies and borrows)

src/contracts/LiquidityHub.sol
miguelmtzinf: left a comment related to this: https://github.com/aave/aave-v4/pull/375/files#r2148683241
we an tackle in #447  imo

src/contracts/LiquidityHub.sol
miguelmtzinf: hmm yeah, i d need to make it `public` instead of `external`

src/contracts/Spoke.sol
miguelmtzinf: why?

avniculae: I think it's unused

miguelmtzinf: yeah, we can remove in a separate pr

src/libraries/types/DataTypes.sol
CheyenneAtapour: What aout `protocolRate`? or `protocolShare`?

src/contracts/TreasurySpoke.sol
```diff
@@ -0,0 +1,57 @@
...
+    return HUB.getSpokeSuppliedAmount(reserveId, address(this));
+  }
+
+  /// @inheritdoc ITreasurySpoke
+  function getSuppliedShares(uint256 reserveId) external view returns (uint256) {
+    return HUB.getSpokeSuppliedShares(reserveId, address(this));
+  }
+
+  // todo: add functions to transfer ERC20 out, assuming this can hold assets
+  // todo: add functions to rescue
```
CheyenneAtapour, line 56: These are for a future PR I imagine?

DhairyaSethi: correct, #450

src/contracts/LiquidityHub.sol
DhairyaSethi: @avniculae pls note^

tests/unit/Spoke/Spoke.AccrueLiqudityFee.t.sol
```diff
@@ -0,0 +1,507 @@
...
+    // Time passes
+    skip(skipTime);
+
+    DataTypes.UserPosition memory bobPosition = spoke1.getUserPosition(reserveId, bob);
+    {
+      uint256 baseDebt = _calculateExpectedBaseDebt(borrowAmount, baseBorrowRate, startTime);
+      uint256 expectedPremiumDrawnShares = bobPosition.baseDrawnShares.percentMulUp(userRp);
+      uint256 expectedPremiumDebt = hub.convertToDrawnAssets(assetId, expectedPremiumDrawnShares) -
+        bobPosition.premiumOffset +
+        bobPosition.realizedPremium;
```
CheyenneAtapour, line 79: I'm personally fine with this approach, but do you wanna change to the approach of using the `_getPremiumDebt` helper like the spoke accrue interest tests?

tests/unit/Spoke/Spoke.AccrueLiqudityFee.t.sol
```diff
@@ -0,0 +1,507 @@
...
+    updateLiquidityPremium(spoke1, reserveId, expectedRp);
+    uint256 liquidityFee = 5_00;
+    updateLiquidityFee(hub, assetId, liquidityFee);
+
+    uint256 borrowAmount = 1000e18;
+    uint256 supplyAmount = _calcMinimumCollAmount(spoke1, reserveId, reserveId, borrowAmount);
+    uint256 rate = 50_00; // 50.00% base borrow rate
+    uint256 expectedBaseDebtAccrual = borrowAmount.percentMul(rate);
+    uint256 expectedBaseDebt = borrowAmount + expectedBaseDebtAccrual;
+    uint256 expectedPremiumDebt = expectedBaseDebtAccrual.percentMul(expectedRp);
```
CheyenneAtapour: should we change these `percentMul` to `percentMulUp`?

miguelmtzinf: This does not follow the same calculation approach as src. I changed it because tests are passing, but the whole test file needs to be rework, as it needs more tests and can re-use some tests from Spoke.AccrueInterestTests (#229)

tests/unit/Spoke/Spoke.AccrueLiqudityFee.t.sol
miguelmtzinf: same as below. Essentially we need better tests here

src/libraries/logic/AssetLogic.sol
miguelmtzinf: https://github.com/aave/aave-v4/issues/457

Issue #378, opened 2025-05-01
Helpers for expected debt in test suite


CheyenneAtapour:
Add helpers for expected debt to our test suite

From Miguel comment on PR: 

"We gotta start using helpers for calcExpectedDebtIndex or calcExpectedDebt"

yan-man:
catalogued in notion

PR #379, opened 2025-05-05
Fixes debt index grow while no debt

### LINKED ISSUE #376: Hub:accrue increases the debt index even if there is no debt in the system
Currently, the debt index grows even if there is no debt, meaning that a user can supply 100 USDC with a base rate of 5% and get 105 USDC at the end of the year even if their liquidity hasn't been used at all.
 
https://github.com/aave/aave-v4/blob/a67383c5dd5b5ce3c8bde0fbee33fe08b4a2c89e/src/libraries/logic/AssetLogic.sol#L127-L137

L130 must become
```

    if (lastUpdateTimestamp == block.timestamp || asset.baseDrawnShares == 0) {
      return (previousIndex, 0);
    }
```

The test file `LiquidityHubAccrueInterestTest` must cover this issue.

miguelmtzinf:
- Closes #376 
- Migrates `LiquidityHubAccrueInterest` tests leaving premium part out of the scope. Tests for this part belongs to a premium data unit tests. Premium data is also tested in restore and settle unit tests separately. We can tests this altogether in Hub or/and Spoke side.
- Partially tackles #378 but will be more complete once RF accrual is added

### DIFF
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -829,7 +829,7 @@ abstract contract Base is Test {
     uint256 initialDebtIndex,
     uint256 borrowRate,
     uint40 startTime
-  ) internal returns (uint256) {
+  ) internal view returns (uint256) {
     return initialDebtIndex.rayMulUp(MathUtils.calculateLinearInterest(borrowRate, startTime));
   }
 
@@ -839,7 +839,7 @@ abstract contract Base is Test {
     uint256 initialDebtIndex,
     uint256 borrowRate,
     uint40 startTime
-  ) internal returns (uint256 newDebtIndex, uint256 newBaseDebt) {
+  ) internal view returns (uint256 newDebtIndex, uint256 newBaseDebt) {
     newDebtIndex = calculateExpectedDebtIndex(initialDebtIndex, borrowRate, startTime);
     newBaseDebt = initialDrawnShares.rayMulUp(newDebtIndex);
   }
--- a/tests/unit/LiquidityHub/LiquidityHubAccrueInterest.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHubAccrueInterest.t.sol
@@ -78,7 +78,7 @@ contract LiquidityHubAccrueInterestTest is Base {
     assertEq(getAssetBaseDebt(daiAssetId), 0);
   }
 
-  /// no interest accrued when no debt
+  /// no interest accrued when no debt after repay
   function test_accrueInterest_NoInterest_NoDebt(uint40 elapsed) public {
     elapsed = uint40(bound(elapsed, 1, type(uint40).max / 3));
 


tests/Base.t.sol
```diff
@@ -27,6 +27,7 @@ import {WETH9} from 'src/dependencies/weth/WETH9.sol';
 
 abstract contract Base is Test {
   using WadRayMath for uint256;
```
DhairyaSethi: can rm this, it's unused (except one invocation of WAD constant which can be taken from extended lib. imo let's start cleaning this up as we go along in PRs

tests/Base.t.sol
```diff
@@ -823,4 +824,32 @@ abstract contract Base is Test {
...
+  /// @dev Calculate expected debt index and base debt based on input params
+  function calculateExpectedDebt(
+    uint256 initialDrawnShares,
+    uint256 initialDebtIndex,
+    uint256 borrowRate,
+    uint40 startTime
+  ) internal returns (uint256 newDebtIndex, uint256 newBaseDebt) {
+    newDebtIndex = initialDebtIndex.rayMulUp(
+      MathUtils.calculateLinearInterest(borrowRate, startTime)
+    );
```
DhairyaSethi: can re-use above helper

tests/unit/LiquidityHub/LiquidityHubAccrueInterest.t.sol
```diff
@@ -265,98 +227,97 @@ contract LiquidityHubAccrueInterestTest is Base {
...
+    Timestamps memory timestamps;
+    AssetDataLocal memory assetData;
+    Spoke1Amounts memory spoke1Amounts;
+    CumulatedInterest memory cumulated;
+
+    spoke1Amounts.supply0 = borrowAmount * 2;
+    timestamps.t0 = uint40(vm.getBlockTimestamp());
+
+    vm.startPrank(address(spoke1));
+    hub.add(daiAssetId, spoke1Amounts.supply0, address(spoke1));
```
DhairyaSethi: utils can be used to avoid needing a prank but nbd

miguelmtzinf:
Thanks @DhairyaSethi . I think `test_accrueInterest_NoInterest_OnlySupply` and `test_accrueInterest_NoInterest_NoDebt` do decent work. We can also add more cases to other unit test files

tests/Base.t.sol
```diff
@@ -823,4 +823,30 @@ abstract contract Base is Test {
...
+  function calculateExpectedDebtIndex(
+    uint256 initialDebtIndex,
+    uint256 borrowRate,
+    uint40 startTime
+  ) internal returns (uint256) {
+    return initialDebtIndex.rayMulUp(MathUtils.calculateLinearInterest(borrowRate, startTime));
+  }
+
+  /// @dev Calculate expected debt index and base debt based on input params
+  function calculateExpectedDebt(
```
CheyenneAtapour, line 837: These 2 expected debt functions can be declared as `view`

tests/unit/LiquidityHub/LiquidityHubAccrueInterest.t.sol
```diff
@@ -83,161 +40,257 @@ contract LiquidityHubAccrueInterestTest is Base {
...
-
-    //     // Spoke 2 does a supply to accrue interest
-    //     Utils.supply(hub, daiAssetId, address(spoke2), 1000e18, 0, address(spoke2), address(spoke2));
-
-    //     DataTypes.Asset memory daiInfo = hub.getAsset(daiAssetId);
-
-    //     uint256 totalBase = MathUtils.calculateLinearInterest(baseBorrowRate, uint40(startTime)).rayMul(
-    //       initialDebt
-    //     );
+  /// no interest accrued when no debt
```
CheyenneAtapour: This description is slightly misleading because there is debt and we test the interest accrual as normal. It's only after the repay of debt that we test the no interest accrued.

So how about `no interest accrued when no debt after repay`? Or similar to describe the full behavior of the test

PR #386, opened 2025-05-07
chore: emit amounts on hub

DhairyaSethi:
i am doing this under duress

### DIFF
--- a/.github/workflows/tests-merge.yml
+++ b/.github/workflows/tests-merge.yml
@@ -1,27 +1,27 @@
 name: Tests
- 
- on:
-   push:
-     branches:
-       - main
- 
- jobs:
-   test:
-     name: Foundry build n test
-     runs-on: ubuntu-latest
-     steps:
-       - uses: actions/checkout@v4
-         with:
-           submodules: recursive
-           token: ${{ secrets.READ_ONLY_PAT || github.token }}
- 
-       - name: Run Foundry setup
-         uses: bgd-labs/github-workflows/.github/actions/foundry-setup@main
-         with:
-           FOUNDRY_VERSION: stable
- 
-       - name: Run Forge tests
-         id: test
-         uses: bgd-labs/github-workflows/.github/actions/foundry-test@main
-         with:
-           FOUNDRY_PROFILE: ci
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  test:
+    name: Foundry build n test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+          token: ${{ secrets.READ_ONLY_PAT || github.token }}
+
+      - name: Run Foundry setup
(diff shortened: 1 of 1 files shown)


src/interfaces/ILiquidityHub.sol
```diff
@@ -20,10 +20,31 @@ interface ILiquidityHub {
...
+    uint256 indexed assetId,
+    address indexed spoke,
+    uint256 drawnShares,
+    uint256 drawnAmount
+  );
+  event Restore(
+    uint256 indexed assetId,
+    address indexed spoke,
+    uint256 baseRestoredShares,
+    uint256 totalRestoredAmount
```
miguelmtzinf, line 46: this won't help to infer the debt index. Are you thinking about using any other value to calc `baseDebtAmount`?

src/libraries/logic/AssetLogic.sol
```diff
@@ -120,8 +126,9 @@ library AssetLogic {
 
   // @dev Utilizes existing `asset.baseBorrowRate`
   function accrue(DataTypes.Asset storage asset) internal {
-    asset.baseDebtIndex = asset.previewIndex();
+    uint256 baseDebtIndex = asset.baseDebtIndex = asset.previewIndex();
     asset.lastUpdateTimestamp = block.timestamp;
+    emit ILiquidityHub.DrawnIndexUpdate(asset.id, baseDebtIndex);
```
miguelmtzinf, line 131: this needing the interface sucks, but okay

DhairyaSethi: i really hope this is temp and backend just does the historic block queries

src/interfaces/ILiquidityHub.sol
DhairyaSethi: `totalRestoredAmount` is needed to accurately calc `availableLiquidity`
these amounts are not added for the share->asset conversion (they're inaccurate anyway bc of donation); backend is expected to get calculate the value using the ex ratios

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -179,7 +179,7 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     assertEq(tokenList.dai.balanceOf(address(hub)), 0);
 
     vm.expectEmit(address(hub));
-    emit ILiquidityHub.Add(daiAssetId, address(spoke1), amount);
+    emit ILiquidityHub.Add(daiAssetId, address(spoke1), amount, amount);
```
yan-man, line 182: for later - but these have always been technically wrong prev, but just working bc ex rate was not inflated right, bc it should be converting to shares. But just so happens shares == amounts

DhairyaSethi: yeah exactly, all hub tests are not migrated so it's ok but we should have had an event emission assert on hub.add since that is migrated. it'll be a todo for later

PR #389, opened 2025-05-08
Deficit Accounting - Integrated into LH `Restore` method

### LINKED ISSUE #387: Deficit accounting
unblocks https://github.com/aave/aave-v4/issues/262

deficit accounting discussed with Emilio/Dhairya on 5.5.25 - after liquidation, check whether bad debt remains and if so, clear it by calling a function on Hub (either merged into hub restore method or a standalone separate method). clear away debt from the user/spoke position and add it to the deficit storage var on hub.

also include deficit into supply ex rate
- Miguel prefers 2 separate methods rather than merging into hub restore

yan-man:
- implement deficit accounting into spoke liquidations by integrating logic into LH restore method
- rm unsetting as collateral in all liquidation cases

closes #387 
closes https://github.com/aave/aave-v4/issues/435

### DIFF
--- a/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.MultiReserve.BadPremiumDebt.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.MultiReserve.BadPremiumDebt.t.sol
@@ -304,8 +304,6 @@ contract LiquidationCallMultiReserveBadPremiumDebtTest is SpokeLiquidationBase {
 
     ) = _calculateAvailableCollateralToLiquidate(state.spoke, state, UINT256_MAX);
 
-    console.log('collToLiq: %e, debtToLiq: %e', state.collToLiq, state.debtToLiq);
-
     // logs to read protocol fee from tmp emitted event
     // TODO: update when treasury accounting is done
     vm.recordLogs();


src/contracts/LiquidityHub.sol
```diff
@@ -202,18 +206,20 @@ contract LiquidityHub is ILiquidityHub {
...
-    _validateRestore(asset, spoke, baseAmount, premiumAmount);
+    _validateRestore(asset, spoke, baseAmount, premiumAmount, deficitAmount);
     asset.updateBorrowRate({liquidityAdded: baseAmount, liquidityTaken: 0}); // both can be zero
 
     uint256 totalRestoredAmount = baseAmount + premiumAmount;
     uint256 baseDrawnSharesRestored = asset.toDrawnSharesDown(baseAmount);
+    uint256 actualRestoredAmount = totalRestoredAmount - deficitAmount;
 
-    asset.availableLiquidity += totalRestoredAmount;
+    asset.availableLiquidity += actualRestoredAmount;
```
DhairyaSethi, line 289: we could also have available liquidity goes up by total restored amount & subtract deficit (instead of add) in supply rate

src/contracts/Spoke.sol
```diff
@@ -431,12 +592,25 @@ contract Spoke is ISpoke {
...
 
       // repay debt
-      vars.restoredShares = HUB.restore(
-        vars.debtAssetId,
-        vars.baseDebtToLiquidate,
-        vars.premiumDebtToLiquidate,
-        msg.sender
-      );
+      if (vars.deficit == 0) {
+        // if no bad debt remains, restore debt normally
```
DhairyaSethi: two branches can be merged in one?

src/contracts/Spoke.sol
```diff
@@ -532,19 +713,45 @@ contract Spoke is ISpoke {
     );
   }
 
+  /** @dev Settle deficit accounting for user's remaining debt assets except for
+   * `reserveIdToAvoid` as that is expected to be handled in liquidation
+   */
+  function _settleRemainingDeficit(uint256 reserveIdToAvoid, address userAddress) internal {
```
DhairyaSethi: this is SUPER expensive rn 😭 the snapshots aren't generated correctly, could you push the correct ones so we can see?

src/contracts/Spoke.sol
```diff
@@ -532,19 +713,45 @@ contract Spoke is ISpoke {
...
+    // get all user's debt assets except assetToAvoid
+    uint256 reserveCount_ = reserveCount;
+    uint256 reserveId;
+    while (reserveId < reserveCount_) {
+      DataTypes.UserPosition storage userPosition = _userPositions[userAddress][reserveId];
+      if (_isBorrowing(userPosition) && reserveId != reserveIdToAvoid) {
+        (uint256 baseDebt, uint256 premiumDebt) = _getUserDebt(userPosition, reserveId);
+        _executeRepay(
```
DhairyaSethi: dup work being done with the notify in case of an update so maybe we club this function there

src/contracts/LiquidityHub.sol
yan-man: what benefits do you think from that

src/contracts/Spoke.sol
yan-man: this i am not sure - bc we need to restore, settle internal accounting, before we calc the new user rp, so I think they need to happen at diff parts of the flow

src/contracts/Spoke.sol
yan-man: pushed. wouldn't show up properly though I think, need to add bad debt liq scenario into the gas report for that max gas action, also with multi-debts

src/contracts/Spoke.sol
yan-man: yep had the same thoughts, but didn't want to overly complicate yet

src/contracts/LiquidityHub.sol
```diff
@@ -202,20 +204,25 @@ contract LiquidityHub is ILiquidityHub {
 
     asset.accrue();
 
-    _validateRestore(asset, spoke, baseAmount, premiumAmount);
+    _validateRestore(asset, spoke, baseAmount, premiumAmount, deficitAmount);
     asset.updateBorrowRate({liquidityAdded: baseAmount, liquidityTaken: 0}); // both can be zero
```
miguelmtzinf: should we account for deficit here?

DhairyaSethi: good catch, yes i think so bc baseAmount is not necessarily the full amount which will be added to available liquidity later

src/contracts/Spoke.sol
```diff
@@ -973,6 +918,81 @@ contract Spoke is ISpoke, Multicall {
...
+      (newUserRiskPremium, , , , ) = _calculateUserAccountData(user);
+      userPremiumDrawnShares = userPosition.premiumDrawnShares = userPosition
+        .baseDrawnShares
+        .percentMul(newUserRiskPremium);
+      userPremiumOffset = userPosition.premiumOffset = HUB.previewOffset(
+        vars.assetId,
+        userPosition.premiumDrawnShares
+      );
+    } else {
+      userPremiumDrawnShares = userPosition.premiumDrawnShares = 0;
```
miguelmtzinf: we repay DAI debt with non-zero deficit, why rp is 0 now?

src/contracts/LiquidityHub.sol
```diff
@@ -202,20 +204,25 @@ contract LiquidityHub is ILiquidityHub {
...
     asset.baseDrawnShares -= baseDrawnSharesRestored;
 
     spoke.baseDrawnShares -= baseDrawnSharesRestored;
 
-    assetsList[assetId].safeTransferFrom(from, address(this), totalRestoredAmount);
+    assetsList[assetId].safeTransferFrom(from, address(this), actualRestoredAmount);
 
     emit Restore(assetId, msg.sender, baseDrawnSharesRestored, totalRestoredAmount);
+    if (deficitAmount > 0) {
+      emit DeficitCreated(msg.sender, assetId, deficitAmount);
```
DhairyaSethi: nit: let's follow assetId, caller order here for consistency

src/contracts/Spoke.sol
```diff
@@ -1175,18 +1215,48 @@ contract Spoke is ISpoke, Multicall {
     );
   }
 
+  // TODO: opt by merging with
```
miguelmtzinf: with what?

src/contracts/LiquidityHub.sol
yan-man: wdyt is the best way to do that here? bc deficit can be a combo of both base/premium, but the liquidityAdded should just be based on base debt. Is this correct?

assume premium = 20, base = 80
- if deficit is 10, liquidityAdded = 80
- if deficit is 30, liquidityAdded = 70

src/contracts/Spoke.sol
yan-man: if there is a non-zero deficit, it means bad debt of some amount (although partially repaid). Therefore it means collateral remaining is 0, and remaining debt goes into deficit. By the end, the user has effectively 0 coll / 0 debt, so 0 rp. 

therefore any deficit should result in 0 rp

src/contracts/LiquidityHub.sol
yan-man: good call 

https://github.com/aave/aave-v4/pull/389/commits/793e45796d9a395878b57869c171805f64a4b20b

src/contracts/Spoke.sol
yan-man: ha good catch, meant `_notifyRiskPremiumUpdate`

https://github.com/aave/aave-v4/pull/389/commits/793e45796d9a395878b57869c171805f64a4b20b

src/contracts/LiquidityHub.sol
DhairyaSethi: yeah something like `liqAdded = base - zeroFloorSub(deficit, premium)`

DhairyaSethi: > deficit can be a combo of both base/premium, but the liquidityAdded should just be based on base debt. Is this correct?

yes iirc that's bc we only want the base rate to be concerned with the base debt, premium restored amount is introduced to available liquidity independently - although this means that base rate will be different right after an operation, wondering if this opens up mev ops? @miguelmtzinf

yan-man: updated here: https://github.com/aave/aave-v4/pull/389/commits/ed38aaf8c65502a613dda610e4befa08efbfa417

src/interfaces/ILiquidityHub.sol
```diff
@@ -222,7 +226,7 @@ interface ILiquidityHub {
   function getSpokeSuppliedShares(uint256 assetId, address spoke) external view returns (uint256);
 
   function getSpokeTotalDebt(uint256 assetId, address spoke) external view returns (uint256);
-
+  function getDeficit(uint256 assetId) external view returns (uint256);
```
CheyenneAtapour, line 240: Nit to maintain consistent whitespace

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -27,48 +27,60 @@ contract SpokeLiquidationBase is SpokeBase {
...
+    Balance[] debts;
+    uint256 hfBadDebtThreshold;
   }
 
   uint256 internal constant MIN_AMOUNT_IN_BASE_CURRENCY = 1e26;
 
   function setUp() public virtual override {
     super.setUp();
-    _addBorrowableLiquidity(MAX_SUPPLY_AMOUNT);
+    _deployBorrowableLiquidities(MAX_SUPPLY_AMOUNT * 10); // additional liquidity buffer for if collateral reserve == debt reserve
```
CheyenneAtapour: In theory, we should be enforcing supply caps, which means, when we do, this would fail, right?

yan-man: yeah if the supply cap were less than this value, it would fail

CheyenneAtapour: I think it'd be better to keep the condition that we are not supplying more than `MAX_SUPPLY_AMOUNT` in the tests, otherwise it defeats the purpose of naming it as such

src/contracts/Spoke.sol
CheyenneAtapour: Yes, since we only realize deficit when 0 collateral, then non-zero deficit implies 0 rp

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -294,24 +331,75 @@ contract SpokeLiquidationBase is SpokeBase {
...
+      state.finalTotalCollateralInBaseCurrency,
+      0,
+      string.concat('totalCollateralInBaseCurrency should be > 0 ', label)
+    );
+    assertGt(
+      state.finalTotalDebtInBaseCurrency,
+      0,
+      string.concat('totalDebtInBaseCurrency should be > 0 ', label)
+    );
+    // with collateral/debt remaining, user rp is not 0
```
CheyenneAtapour: you can have 0 rp with collateral / debt remaining, if the collateral has 0 liquidity premium

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -294,24 +331,75 @@ contract SpokeLiquidationBase is SpokeBase {
...
+      string.concat('deficit added to hub ', label)
+    );
+  }
+
+  /// generic assertions in non bad debt scenarios
+  function _assertNoBadDebt(
+    LiquidationTestLocalParams memory state,
+    string memory label
+  ) internal pure {
+    // total debt/collateral in user's position should be > 0
```
CheyenneAtapour, line 406: I'm unsure about these conditions, because you can have no bad debt with both 0 debt and 0 collateral, as well as nonzero collateral and zero debt

yan-man: > nonzero collateral and zero debt

this one idt is possible bc you liq (at most, with 0 liq bonus) equal amts of collateral/debt, so this scenario means you started with debt < coll, ie HF should be at worst still >= 1 (assuming collFactor = 1) and position was not liquidatable to begin with

> no bad debt with both 0 debt and 0 collateral

this is the "perfect" liquidation case, but i also don't think it's possible due to the rounding enforced (coll to liq up, debt to liq down), meaning some dust should remain on debt, which would get turned into deficit

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: added a check on all collaterals having 0 LP: https://github.com/aave/aave-v4/pull/389/commits/a17091fd8f185dc1b2bfaf82d33632972deba8ae

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/389/commits/a17091fd8f185dc1b2bfaf82d33632972deba8ae

src/interfaces/ILiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/389/commits/a17091fd8f185dc1b2bfaf82d33632972deba8ae

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -400,22 +523,20 @@ contract SpokeLiquidationBase is SpokeBase {
...
-    uint256 collateralReserveId,
+    address user,
     uint256 liquidationBonus
   ) internal view returns (uint256) {
-    return
-      _calcLowestHfForCloseFactorFromCollateralFactor(
-        _getCollateralFactor(spoke, collateralReserveId),
-        liquidationBonus
-      );
+    (, uint256 avgCollateralFactor, , , ) = spoke.getUserAccountData(user);
```
DhairyaSethi, line 544: we can re-use calcUserAccountData test helper from Spoke.repay tests instead here

src/contracts/Spoke.sol
```diff
@@ -930,6 +876,110 @@ contract Spoke is ISpoke, Multicall {
     return premiumIncrease;
   }
 
+  function _executeRepay(
+    uint256 reserveId,
+    uint256 amount,
+    address user,
+    DataTypes.UserPosition storage userPosition,
```
DhairyaSethi, line 921: this can just instantiate userPos storage pointer again, will be cleaner on repay tbh, nbd tho

src/contracts/Spoke.sol
DhairyaSethi: agree, we should note this somewhere for certora to prove mathematically with fv

src/contracts/Spoke.sol
yan-man: just that in `_settleRemainingDeficit` i already instantiate it to check `isBorrowing` on userPos so wanted to pass it in vs instantiating within this method

src/contracts/Spoke.sol
yan-man: @DhairyaSethi have you already started a doc anywhere on certora FV rules or should I spin that up

src/contracts/Spoke.sol
DhairyaSethi: yeah i saw, it's zero cost instantiate again but also nbd can resolve this

src/contracts/Spoke.sol
DhairyaSethi: let's just add it in the protocol invariants doc you'd spinned up i think for now

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: i can't find this helper, can you point me to it pls

DhairyaSethi: https://github.com/aave/aave-v4/blob/main/tests/unit/Spoke/SpokeBase.t.sol#L687

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
CheyenneAtapour: ah ok, I think I wasn't treating these conditions as being post-liquidation. Was just treating them as general conditions, which is the cause of my confusion

src/interfaces/ILiquidityHub.sol
```diff
@@ -52,6 +52,7 @@ interface ILiquidityHub {
     int256 premiumOffsetDelta,
     int256 realizedPremiumDelta
   );
+  event DeficitCreated(uint256 indexed debtAssetId, address indexed spoke, uint256 amountCreated);
```
miguelmtzinf:
```suggestion
  event DeficitCreated(uint256 indexed assetId, address indexed spoke, uint256 amount);
```

src/contracts/LiquidityHub.sol
```diff
@@ -203,20 +205,28 @@ contract LiquidityHub is ILiquidityHub {
 
     asset.accrue();
 
-    _validateRestore(asset, spoke, baseAmount, premiumAmount);
-    asset.updateBorrowRate({liquidityAdded: baseAmount, liquidityTaken: 0}); // both can be zero
+    _validateRestore(asset, spoke, baseAmount, premiumAmount, deficitAmount);
+    asset.updateBorrowRate({
+      liquidityAdded: _calculateLiquidityAdded(baseAmount, premiumAmount, deficitAmount),
+      liquidityTaken: 0
```
miguelmtzinf, line 282: why not this:
```suggestion
      liquidityAdded: baseAmount,
      liquidityTaken: deficit
```
why are we substracting the deficit from premium? do we wanna cover any deficit with the premium?

src/contracts/LiquidityHub.sol
```diff
@@ -437,8 +451,10 @@ contract LiquidityHub is ILiquidityHub {
     DataTypes.Asset storage asset,
     DataTypes.SpokeData storage spoke,
     uint256 baseAmountRestored,
-    uint256 premiumAmountRestored
+    uint256 premiumAmountRestored,
+    uint256 deficitAmount
   ) internal view {
+    require(deficitAmount <= baseAmountRestored + premiumAmountRestored, InvalidDeficitAmount());
```
miguelmtzinf, line 560: this sounds reasonable, we don't wanna allow to report a deficit bigger than the debt being repaid. Is it possible that we want to report a deficit that is higher that the debt being repaid, but lower than the total debt of the user?

src/contracts/Spoke.sol
```diff
@@ -928,11 +874,117 @@ contract Spoke is ISpoke, Multicall {
...
+      (newUserRiskPremium, , , , ) = _calculateUserAccountData(user);
+      userPremiumDrawnShares = userPosition.premiumDrawnShares = userPosition
+        .baseDrawnShares
+        .percentMulUp(newUserRiskPremium);
+      userPremiumOffset = userPosition.premiumOffset = HUB.previewOffset(
+        vars.assetId,
+        userPosition.premiumDrawnShares
+      );
+    } else {
+      userPremiumDrawnShares = userPosition.premiumDrawnShares = 0;
```
```suggestion
      // non-zero deficit means user ends up with zero debt
      userPremiumDrawnShares = userPosition.premiumDrawnShares = 0;
```

src/contracts/Spoke.sol
```diff
@@ -1036,14 +1060,58 @@ contract Spoke is ISpoke, Multicall {
...
+      // deficit accounting
       if (vars.newUserSuppliedShares == 0) {
-        userCollateralPosition.usingAsCollateral = false;
-        emit UsingAsCollateral(collateralReserveId, users[vars.i], false);
+        uint256 outstandingDebt = vars.baseDebt +
+          vars.premiumDebt -
+          vars.baseDebtToLiquidate -
+          vars.premiumDebtToLiquidate;
+        /// @notice user supplied shares only applies to single coll reserve
+        /// but hasNoCollateralLeft includes all coll reserves
```
miguelmtzinf, line 1139:
```suggestion
        // collateral asset in liquidation is the sole collateral asset of the user
```
I think it's clearer, not sure if I am missing anything else about this comment

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/389/files/1764839775ad9babb12f4992ff896f7c76350e91#r2110430906

this was my operating assumption, was that incorrect?

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: hm but I need avgCollateralFactor and that helper just gives stored user rp right? wdym by re-using it here

src/contracts/Spoke.sol
yan-man: this comment I added specifically bc there's 2 levels of checks, both related to collateral to determine if deficit remains:
- `if (vars.newUserSuppliedShares == 0)`
- `vars.hasNoCollateralLeft`

and both these checks are necessary because `vars.newUserSuppliedShares` only gives info about the current coll asset involved in the liq, but `hasNoCollateralLeft` is across ALL collaterals a user might have.

src/interfaces/ILiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/389/commits/201e976b33f74b301ecfa8e2e928ed0ed3e345e8

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/389/commits/201e976b33f74b301ecfa8e2e928ed0ed3e345e8

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
DhairyaSethi: adding extra stuff so essentially we calc user data in the test (dup code unfortunately) rather than through getter

yan-man:
decision to implement deficit accounting as a separate dedicated method, refer here instead: https://github.com/aave/aave-v4/pull/472

PR #391, opened 2025-05-12
Implement Multicall

### LINKED ISSUE #226: RefreshUserRiskPremium method
permissionless method which can be called for a user when their risk premium has changed (due to price fluctuations, collateralised asset supply interest, liquidity premium change)

should refresh user risk premium & propagate this update across all of their debt assets

### LINKED ISSUE #292: Spoke: Refresh/UpdateUserRiskPremium Method

### LINKED ISSUE #390: Implement Multicall
We want to use OpenZeppelin's Multicall to combine actions such as supply and set as collateral to provide a gas efficient option for the user

### LINKED ISSUE #393: Consistent OpenZeppelin Contract Versions
Ensure we are using a consistent version across our OpenZeppelin dependencies

CheyenneAtapour:
* Implements Multicall for spoke, with some basic tests
* Updates all OpenZeppelin dependencies to version 5.1
* Implements a function to update user risk premium

closes #390 
closes #393 
closes #226 
closes #292 (duplicate)

### DIFF
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -1,15 +1,15 @@
 {
-  "borrow": "480229",
-  "repay: full": "220864",
-  "repay: partial": "288768",
-  "supply: 0 debt, collateralDisabled": "171540",
-  "supply: 0 debt, collateralEnabled": "171516",
-  "supply: 1 debt": "171564",
-  "supply: 2 debt": "171575",
-  "supply: 3 debt": "171528",
-  "updateUserRiskPremium": "185386",
-  "updateUserRiskPremium: 2 debts": "225911",
+  "borrow": "481358",
+  "repay: full": "221740",
+  "repay: partial": "289863",
+  "supply: 0 debt, collateralDisabled": "171824",
+  "supply: 0 debt, collateralEnabled": "171800",
+  "supply: 1 debt": "171848",
+  "supply: 2 debt": "171859",
+  "supply: 3 debt": "171812",
+  "updateUserRiskPremium": "186234",
+  "updateUserRiskPremium: 2 debts": "227183",
   "usingAsCollateral": "50661",
-  "withdraw: full": "216963",
-  "withdraw: partial": "227914"
+  "withdraw: full": "218092",
+  "withdraw: partial": "229043"
 }
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -157,6 +157,7 @@ contract Spoke is ISpoke, Multicall {
 
     _refreshPremiumDebt(
       reserve,
+      msg.sender,
       assetId,
       -int256(userPremiumDrawnShares),
       -int256(userPremiumOffset),
@@ -180,6 +181,7 @@ contract Spoke is ISpoke, Multicall {
 
     _refreshPremiumDebt(
       reserve,
+      msg.sender,
       assetId,
       int256(userPremiumDrawnShares),
       int256(userPremiumOffset),
@@ -210,6 +212,7 @@ contract Spoke is ISpoke, Multicall {
 
     _refreshPremiumDebt(
       reserve,
+      msg.sender,
       assetId,
       -int256(userPremiumDrawnShares),
       -int256(userPremiumOffset),
@@ -233,6 +236,7 @@ contract Spoke is ISpoke, Multicall {
 
     _refreshPremiumDebt(
       reserve,
+      msg.sender,
       assetId,
       int256(userPremiumDrawnShares),
       int256(userPremiumOffset),
@@ -250,7 +254,7 @@ contract Spoke is ISpoke, Multicall {
     DataTypes.Reserve storage reserve = _reserves[reserveId];
     uint256 assetId = reserve.assetId;
 
-    (uint256 baseDebt, uint256 premiumDebt) = _getUserDebt(userPosition, reserve.assetId);
+    (uint256 baseDebt, uint256 premiumDebt) = _getUserDebt(userPosition, assetId);
     (uint256 baseDebtRestored, uint256 premiumDebtRestored) = _calculateRestoreAmount(
       baseDebt,
       premiumDebt,
@@ -268,10 +272,11 @@ contract Spoke is ISpoke, Multicall {
 
     _settlePremiumDebt(
       reserve,
+      msg.sender,
       assetId,
       -int256(userPremiumDrawnShares),
       -int256(userPremiumOffset),
-      _signedDiff(userPosition.realizedPremium, userRealizedPremium)
+      _signedDiff(premiumDebt - premiumDebtRestored, userRealizedPremium)
     ); // we settle premium debt here
     uint256 restoredShares = HUB.restore(
       assetId,
@@ -295,6 +300,7 @@ contract Spoke is ISpoke, Multicall {
 
     _refreshPremiumDebt(
       reserve,
+      msg.sender,
       assetId,
       int256(userPremiumDrawnShares),
       int256(userPremiumOffset),
@@ -335,6 +341,7 @@ contract Spoke is ISpoke, Multicall {
 
     _refreshPremiumDebt(
       reserve,
+      user,
       assetId,
       -int256(userPremiumDrawnShares),
       -int256(userPremiumOffset),
@@ -359,6 +366,7 @@ contract Spoke is ISpoke, Multicall {
 
     _refreshPremiumDebt(
       reserve,
+      user,
       assetId,
       int256(newUserPremiumDrawnShares),
       int256(userPremiumOffset),
@@ -552,12 +560,19 @@ contract Spoke is ISpoke, Multicall {
 
   function _refreshPremiumDebt(
     DataTypes.Reserve storage reserve,
+    address userAddress,
     uint256 assetId,
     int256 premiumDrawnSharesDelta,
     int256 premiumOffsetDelta,
     int256 realizedPremiumDelta
   ) internal {
-    _refresh(reserve, premiumDrawnSharesDelta, premiumOffsetDelta, realizedPremiumDelta);
+    _refresh(
+      reserve,
+      userAddress,
+      premiumDrawnSharesDelta,
+      premiumOffsetDelta,
+      realizedPremiumDelta
+    );
     HUB.refreshPremiumDebt(
       assetId,
       premiumDrawnSharesDelta,
@@ -568,12 +583,19 @@ contract Spoke is ISpoke, Multicall {
 
   function _settlePremiumDebt(
     DataTypes.Reserve storage reserve,
+    address userAddress,
     uint256 assetId,
     int256 premiumDrawnSharesDelta,
     int256 premiumOffsetDelta,
     int256 realizedPremiumDelta
   ) internal {
-    _refresh(reserve, premiumDrawnSharesDelta, premiumOffsetDelta, realizedPremiumDelta);
+    _refresh(
+      reserve,
+      userAddress,
+      premiumDrawnSharesDelta,
+      premiumOffsetDelta,
+      realizedPremiumDelta
+    );
     HUB.settlePremiumDebt(
       assetId,
       premiumDrawnSharesDelta,
@@ -584,6 +606,7 @@ contract Spoke is ISpoke, Multicall {
 
   function _refresh(
     DataTypes.Reserve storage reserve,
+    address userAddress,
     int256 premiumDrawnSharesDelta,
     int256 premiumOffsetDelta,
     int256 realizedPremiumDelta
@@ -594,6 +617,7 @@ contract Spoke is ISpoke, Multicall {
 
     emit RefreshPremiumDebt(
       reserve.reserveId,
+      userAddress,
       premiumDrawnSharesDelta,
       premiumOffsetDelta,
       realizedPremiumDelta
@@ -839,6 +863,7 @@ contract Spoke is ISpoke, Multicall {
 
         _refreshPremiumDebt(
           reserve,
+          userAddress,
           assetId,
           _signedDiff(userPosition.premiumDrawnShares, oldUserPremiumDrawnShares),
           _signedDiff(userPosition.premiumOffset, oldUserPremiumOffset),
--- a/snapshots/Hub.Operations.json
+++ b/snapshots/Hub.Operations.json
@@ -1,8 +1,8 @@
 {
-  "add": "186154",
-  "draw": "122344",
-  "remove: full": "79723",
-  "remove: partial": "96009",
-  "restore: full": "93510",
-  "restore: partial": "103110"
+  "add": "186437",
+  "draw": "122627",
+  "remove: full": "80006",
+  "remove: partial": "96292",
+  "restore: full": "93793",
+  "restore: partial": "103393"
 }
--- a/src/interfaces/ILiquidityHub.sol
+++ b/src/interfaces/ILiquidityHub.sol
@@ -20,7 +20,11 @@ interface ILiquidityHub {
     uint256 supplyCap
   );
 
-  event DrawnIndexUpdate(uint256 indexed assetId, uint256 drawnIndex);
+  event DrawnIndexUpdate(
+    uint256 indexed assetId, 
+    uint256 drawnIndex,
+    uint256 lastUpdateTimestamp
+  );
   event Add(
     uint256 indexed assetId,
     address indexed spoke,
--- a/src/interfaces/ISpoke.sol
+++ b/src/interfaces/ISpoke.sol
@@ -30,6 +30,7 @@ interface ISpoke {
   event UsingAsCollateral(uint256 indexed reserveId, address indexed user, bool usingAsCollateral);
   event RefreshPremiumDebt(
     uint256 indexed reserveId,
+    address indexed user,
     int256 premiumDrawnSharesDelta,
     int256 premiumOffsetDelta,
     int256 realizedPremiumDelta
--- a/src/libraries/logic/AssetLogic.sol
+++ b/src/libraries/logic/AssetLogic.sol
@@ -128,7 +128,7 @@ library AssetLogic {
   function accrue(DataTypes.Asset storage asset) internal {
     uint256 baseDebtIndex = asset.baseDebtIndex = asset.previewIndex();
     asset.lastUpdateTimestamp = block.timestamp;
-    emit ILiquidityHub.DrawnIndexUpdate(asset.id, baseDebtIndex);
+    emit ILiquidityHub.DrawnIndexUpdate(asset.id, baseDebtIndex, block.timestamp);
   }
 
   function previewIndex(DataTypes.Asset storage asset) internal view returns (uint256) {


src/dependencies/openzeppelin/Context.sol
```diff
@@ -21,4 +21,8 @@ abstract contract Context {
   function _msgData() internal view virtual returns (bytes calldata) {
     return msg.data;
   }
+
+  function _contextSuffixLength() internal view virtual returns (uint256) {
```
yan-man, line 25: curious about this, is this an updated version of OZ? why do we need this change

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
```diff
@@ -0,0 +1,252 @@
...
+
+    // Check the assets
+    DataTypes.Asset memory daiAsset = hub.getAsset(daiAssetId);
+    DataTypes.Asset memory usdxAsset = hub.getAsset(usdxAssetId);
+    _checkAssets(newDaiAsset, daiAsset);
+    _checkAssets(newUsdxAsset, usdxAsset);
+  }
+
+  function test_multicall_addMultipleSpokes() public {
+    ISpoke spoke4 = ISpoke(
```
yan-man: nit: if we might add more spokes for the tests so it's > 3, maybe better to have this as newSpoke1 and 2 or something like that instead of continuation of 4/5

src/contracts/Spoke.sol
yan-man, line 1: maybe good to also rm the addSpokes method / refactor the test setup with multicall but could also go into a future pr instead if you add a ticket

tests/unit/Spoke/Spoke.Multicall.t.sol
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  /// Supply and update user risk premium using multicall
+  function test_multicall_supply_updateUserRp() public {
+    vm.skip(true, 'pending user rp update function');
+  }
+
+  /// Withdraw and update user risk premium using multicall
+  function test_multicall_withdraw_updateUserRp() public {
+    vm.skip(true, 'pending user rp update function');
```
yan-man: maybe also add to a ticket so we can track (if you havent already)

src/dependencies/openzeppelin/Context.sol
DhairyaSethi: it's needed essentially to support meta txns which are a thing of the past https://github.com/aave/aave-v4/pull/391/files#diff-5f56460a947debfff0465f78aacba8c7b83e9627347750bc3b356c6ff7399c8bR16-R19

src/dependencies/openzeppelin/Multicall.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.3.0) (utils/Multicall.sol)
```
DhairyaSethi: don't like that that these OZ deps are of different versions, can we make sure to pick up consistent versioning across all OZ deps?

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
DhairyaSethi: personally i think continuation is better & less mental overhead, what's the reason for this pref? maybe i'm missing something

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
```diff
@@ -0,0 +1,252 @@
...
+    uint256 newDrawCap = 20e18;
+    uint256 newSupplyCap = 20e18;
+    DataTypes.SpokeConfig memory newSpokeConfig = DataTypes.SpokeConfig({
+      drawCap: newDrawCap,
+      supplyCap: newSupplyCap
+    });
+
+    // Set up the multicall
+    bytes[] memory calls = new bytes[](2);
+    calls[0] = abi.encodeWithSelector(
```
DhairyaSethi: prefer using `abi.encodeCall`

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
```diff
@@ -0,0 +1,252 @@
...
+      baseDrawnShares: 0,
+      premiumDrawnShares: 0,
+      premiumOffset: 0,
+      realizedPremium: 0,
+      baseDebtIndex: WadRayMath.RAY,
+      baseBorrowRate: 0,
+      lastUpdateTimestamp: vm.getBlockTimestamp(),
+      id: aaveAssetId,
+      config: assetConfig
+    });
```
DhairyaSethi: prob cleaner if written as the following but nbd
```suggestion
    DataTypes.Asset memory expectedAaveAsset;
    expectedAaveAsset.baseDebtIndex = WadRayMath.RAY;
    expectedAaveAsset.lastUpdateTimestamp = vm.getBlockTimestamp();
    expectedAaveAsset.id = aaveAssetId;
    expectedAaveAsset/config = assetConfig;
```

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
```diff
@@ -0,0 +1,252 @@
...
+      newSpokeConfig
+    );
+
+    vm.expectEmit(address(hub));
+    emit ILiquidityHub.SpokeConfigUpdated(daiAssetId, address(spoke1), newDrawCap, newSupplyCap);
+    vm.expectEmit(address(hub));
+    emit ILiquidityHub.SpokeConfigUpdated(usdxAssetId, address(spoke1), newDrawCap, newSupplyCap);
+
+    // Execute the multicall
+    Multicall(address(hub)).multicall(calls);
```
DhairyaSethi: i think `ILiquidityHub` can just inherit from `IMulticall`; same for spoke

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
```diff
@@ -0,0 +1,252 @@
...
+    assertEq(expected.premiumOffset, actual.premiumOffset, 'premiumOffset mismatch');
+    assertEq(expected.realizedPremium, actual.realizedPremium, 'realizedPremium mismatch');
+    assertEq(
+      expected.lastUpdateTimestamp,
+      actual.lastUpdateTimestamp,
+      'lastUpdateTimestamp mismatch'
+    );
+    assertEq(expected.config.drawCap, actual.config.drawCap, 'drawCap mismatch');
+    assertEq(expected.config.supplyCap, actual.config.supplyCap, 'supplyCap mismatch');
+  }
```
DhairyaSethi: another sanity assert which is:
```suggestion
    assertEq(expected.config.supplyCap, actual.config.supplyCap, 'supplyCap mismatch');
    assertEq(abi.encode(expected), abi.encode(actual)); // sanity
  }
```
is nice to have in case we overlook validating some keys

src/contracts/LiquidityHub.sol
```diff
@@ -11,7 +12,7 @@ import {SharesMath} from 'src/libraries/math/SharesMath.sol';
 import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
 
 // @dev Amounts are `asset` denominated by default unless specified otherwise with `share` suffix
-contract LiquidityHub is ILiquidityHub {
+contract LiquidityHub is ILiquidityHub, Multicall {
```
DhairyaSethi: i don't think hub needs this; since all actions are always expected to be initiated from the spoke level & they can already impl multicall. better to rm from here imo to keep hub simpler

tests/unit/Spoke/Spoke.Multicall.t.sol
```diff
@@ -0,0 +1,202 @@
...
+    );
+    assertEq(
+      expected.config.liquidationBonus,
+      actual.config.liquidationBonus,
+      'Liquidation bonus config should match'
+    );
+    assertEq(
+      expected.config.liquidityPremium,
+      actual.config.liquidityPremium,
+      'Liquidity premium config should match'
```
DhairyaSethi:
```suggestion
      'liquidity premium mismatch'
```
to be consistent w the above

tests/unit/Spoke/Spoke.Multicall.t.sol
DhairyaSethi: #226

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
yan-man: if we expand to more spokes, ie at 4, then this would no longer be 4 and 5. Thought we might be adding more spokes in testing

DhairyaSethi: ahh right gotcha ok, aligned

src/dependencies/openzeppelin/Multicall.sol
```diff
@@ -0,0 +1,37 @@
...
+   * @custom:oz-upgrades-unsafe-allow-reachable delegatecall
+   */
+  function multicall(bytes[] calldata data) external virtual returns (bytes[] memory results) {
+    bytes memory context = msg.sender == _msgSender()
+      ? new bytes(0)
+      : msg.data[msg.data.length - _contextSuffixLength():];
+
+    results = new bytes[](data.length);
+    for (uint256 i = 0; i < data.length; i++) {
+      results[i] = Address.functionDelegateCall(address(this), bytes.concat(data[i], context));
```
DhairyaSethi, line 33: imo we can fork this & rm the extra context stuff to save gas by omitting this bytes.concat 
wdyt

src/contracts/Spoke.sol
CheyenneAtapour: Ig we will no longer be adding multicall to hub, so will leave this as is

snapshots/Spoke.Operations.json
```diff
@@ -1,13 +1,13 @@
...
+  "repay: full": "221378",
+  "repay: partial": "289411",
+  "supply: 0 debt, collateralDisabled": "172161",
+  "supply: 0 debt, collateralEnabled": "172137",
+  "supply: 1 debt": "172185",
+  "supply: 2 debt": "172196",
+  "supply: 3 debt": "172149",
+  "usingAsCollateral": "50661",
+  "withdraw: full": "217384",
+  "withdraw: partial": "228335"
```
DhairyaSethi: maybe worth adding a snapshot for multicall gas overhead but nbd

DhairyaSethi: also curious, these increased just bc of an additional inheritance that only adds one more external method or are the snapshots on main not up to date?

src/dependencies/openzeppelin/Multicall.sol
CheyenneAtapour: Have we been using 5.1.0 for everything? I can double-check our contracts all match that release

CheyenneAtapour: Ah, looks like we are not. We can decide on a proper version and I can tackle here or we can do it separately. I made an issue so we are aware: https://github.com/orgs/aave/projects/14/views/1?filterQuery=&pane=issue&itemId=110531006&issue=aave%7Caave-v4%7C393

DhairyaSethi: let's just do it here mate, issues are pilling on lmao. let's stick with 5.1 for now

tests/unit/Spoke/Spoke.Multicall.t.sol
CheyenneAtapour: The LH test suite is gone now, so this pattern is consistent with the test suite now. But if you prefer mismatch for conciseness, I can change

src/dependencies/openzeppelin/Multicall.sol
CheyenneAtapour: We can wait to see what BGD did on 3.4, maybe a different implementation of multicall entirely. As a general preference, I don't like forking OZ, and just prefer to use out of the box. But what you're saying is possible too to save gas. Curious what @yan-man and @miguelmtzinf think

snapshots/Spoke.Operations.json
CheyenneAtapour: It's possible the snapshots on main were not up to date, but yeah, I was thinking they increased because of the additional inheritance

DhairyaSethi: but why would it increase due to inheritance when it doesn't change the function code? my best guess it takes more operations to match the correct selector during dispatch

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
CheyenneAtapour: I'm not totally clear on how you're suggesting this to be done, because we want to inherit the multicall definition from OZ, but if we have a definition of multicall on 2 different parents (e.g Multicall and ISpoke), then we need to define a definition on the spoke

src/dependencies/openzeppelin/Multicall.sol
CheyenneAtapour: we blew up the diff, but alright :)

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
CheyenneAtapour: Done. This is because it's more type-safe, right? Also shorter

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
CheyenneAtapour: I followed this pattern for the relevant parts of existing test suite

tests/unit/Spoke/Spoke.Multicall.t.sol
CheyenneAtapour: I went ahead with "mismatch" for brevity

src/dependencies/openzeppelin/Multicall.sol
yan-man: imo agree with @CheyenneAtapour on the preferring not to fork off OZ, perhaps we can wait on 3.4 first to fully decide

rationale is just adding diff to an audited version of OZ contract, which to me maybe isn't worth the cost-benefit for gas savings.

src/dependencies/openzeppelin/Multicall.sol
yan-man: nice now we have another ticket to close :)

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
CheyenneAtapour: I added this check to the spoke reserve check

src/contracts/LiquidityHub.sol
CheyenneAtapour: Removed

src/contracts/Spoke.sol
```diff
@@ -315,6 +317,55 @@ contract Spoke is ISpoke {
...
+      int256(accruedPremium)
+    );
+
+    uint256 newUserRiskPremium = _validateUserPosition(user); // validates HF
+
+    uint256 newUserPremiumDrawnShares = userPosition.premiumDrawnShares = userPosition
+      .baseDrawnShares
+      .percentMul(newUserRiskPremium);
+    // Check new premium drawn shares as proxy for user risk premium
+    require(newUserPremiumDrawnShares < userPremiumDrawnShares, NoUserRiskPremiumDecrease());
```
DhairyaSethi: i think a better condition would be if it increased then msg sender is either dao or user - that's all
imo don't need to restrict user "accidentally" increasing their rp

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
DhairyaSethi: pls add it to all struct checks

src/contracts/Spoke.sol
```diff
@@ -315,6 +317,55 @@ contract Spoke is ISpoke {
     emit UsingAsCollateral(reserveId, msg.sender, usingAsCollateral);
   }
 
+  /// @dev Must be called on a reserve user is already borrowing
+  /// @dev If called by user, reverts if user risk premium increases
+  function updateUserRiskPremium(uint256 reserveId, address user) external {
```
DhairyaSethi, line 328: since we're anyway doing the notify i think this doesn't need to take in reserve id as input, user rp right now is reserve agnostic

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
DhairyaSethi: ISpoke defines & multicall implements; wouldn't need to define on spoke again

src/contracts/Spoke.sol
CheyenneAtapour: Imo, it's easy to just restrict that. I'm gonna make this check just check if increased, then sender must be DAO (see the todo above at beginning of this function)

src/contracts/Spoke.sol
CheyenneAtapour: But the notify updates everything except one asset. So the one as input is basically the one to avoid. Do you suggest re-writing all of this functionality specifically for update user rp?

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
CheyenneAtapour: In all of the test suites? There is only 1 struct check here: the reserve

src/contracts/Spoke.sol
DhairyaSethi: no that input can be uint.max for ex

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
DhairyaSethi: the other custom assert eqs which validate structs

src/contracts/Spoke.sol
DhairyaSethi: it's overly restrictive

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
CheyenneAtapour: The compiler disagrees with you :/

DhairyaSethi: yeah looks like i was wrong sadge, let's keep as is 🫡

tests/unit/LiquidityHub/LiquidityHub.Multicall.t.sol
CheyenneAtapour: This is the only one I see atm in our test suite

src/contracts/Spoke.sol
CheyenneAtapour: I need to fetch a valid user position though so that I can enforce the call is not decreasing user risk premium. Without the input, I'd have to loop through, which is more gas intensive

DhairyaSethi: notify can return the previous user rp

src/contracts/Spoke.sol
CheyenneAtapour: Ultimately I still want to avoid the (admittedly very niche) edge case where a user wants to update their rp to decrease, but somehow price changes etc happen before the tx lands on chain and actually their rp goes up. Since this behavior (rp must decrease) is enforced when called by not the user or dao, maybe there is a way to achieve this using position manger. 

Also curious to hear @yan-man and @miguelmtzinf takes here

DhairyaSethi: imo should be taken care of at the position manager level if we must guarantee, keeping modular here makes sense. 
tldr 
only user/dao can potentially increase rp 
anyone can decrease anyone's rp

CheyenneAtapour: After more conversations, I think we can get this guarantee outside of the protocol

src/contracts/Spoke.sol
CheyenneAtapour: planning to leave implementation as is for now, and can revisit later with opts if neeeded

src/contracts/Spoke.sol
EthWarrior: I noticed comment about msg sender, would this implementation allow a user to delegate to reduce rp to someone else, to call it?

Use-case would be that we enable a service to do so on the Aave backend that the user can subscribe to

DhairyaSethi: yes right now reducing rp is permissionless, only an increase is limited to dao/user. that's a perfect usecase, third party automation like revert, defi saver should be seemless to integrate

CheyenneAtapour: yup, and third-party can only decrease original user's rp, so that works out great

src/contracts/Spoke.sol
```diff
@@ -315,6 +317,58 @@ contract Spoke is ISpoke {
...
+    _refreshPremiumDebt(
+      reserve,
+      assetId,
+      int256(newUserPremiumDrawnShares),
+      int256(userPremiumOffset),
+      0
+    );
+    _notifyRiskPremiumUpdate(assetId, user, newUserRiskPremium);
+
+    emit UserRiskPremiumUpdated(user, newUserRiskPremium);
```
DhairyaSethi: nit: pls make it present tense to align with the convention

snapshots/Spoke.Operations.json
```diff
@@ -1,13 +1,13 @@
...
+  "repay: full": "220864",
+  "repay: partial": "288768",
+  "supply: 0 debt, collateralDisabled": "171540",
+  "supply: 0 debt, collateralEnabled": "171516",
+  "supply: 1 debt": "171564",
+  "supply: 2 debt": "171575",
+  "supply: 3 debt": "171528",
+  "usingAsCollateral": "50661",
+  "withdraw: full": "216963",
+  "withdraw: partial": "227914"
```
DhairyaSethi: could you pls add a gas snapshot for this

src/contracts/Spoke.sol
CheyenneAtapour: Done. Are we going with user actions as present tense? Because I do see some of the config events are past tense, e.g. `ReserveAdded` `LiquidityPremiumUpdated`

snapshots/Spoke.Operations.json
CheyenneAtapour: Done

CheyenneAtapour:
> optimistic approval pending on a gas snapshot test & event name fix. would have preferred if we reused just the notify method for this & have the method be reserveId agnostic but can opt later; pls add a ticket for this

Issue added here: https://github.com/orgs/aave/projects/14/views/1?pane=issue&itemId=110957047&issue=aave%7Caave-v4%7C394

src/contracts/Spoke.sol
DhairyaSethi: all actions will be present tense, that's what we'd settled on iirc

src/contracts/Spoke.sol
EthWarrior: Okay this make sense, and when you say DAO can has the AC, would it mean the DAO can delegate it to a steward?

CheyenneAtapour: Yeah, access control is not yet implemented. But likely we will make a role for this, which DAO will have and can grant to others, e.g. a steward contract

PR #401, opened 2025-05-21
fix: gas snapshots

### LINKED ISSUE #398: Update Gas Snapshots
- Update forge to `1.1.0` (latest stable version)
- Update gas snapshots (latest stable version results in different gas values for some spoke getters -- need to understand why)
- Inline `isolate` flag in test files by adding `/// forge-config: default.isolate = true` at the top of the test contract: https://book.getfoundry.sh/forge/gas-section-snapshots?highlight=isolate#description. Update or remove `make gas-report` command (if update, just delete `--isolate`)

avniculae:
Closes #398

### DIFF
--- a/.github/workflows/tests-pr.yml
+++ b/.github/workflows/tests-pr.yml
@@ -30,9 +30,6 @@ jobs:
         uses: bgd-labs/github-workflows/.github/actions/foundry-test@main
         with:
           FOUNDRY_PROFILE: gas
-          # even though the gas profile sets FORGE_SNAPSHOT_CHECK to true,
-          # this is overwritten by the foundry-test gh action to false, which
-          # takes priority. hence, we need to overwrite it back to true. 
           FORGE_SNAPSHOT_CHECK: true
 
       - name: Run Forge tests


foundry.toml
```diff
@@ -23,8 +23,6 @@ runs = 5000
 runs = 10000
 
 [profile.gas]
```
DhairyaSethi, line 25: don't need this profile and the extra step in ci anymore if snapshot check is true by default, let's rm

avniculae: I think this is actually useful: it makes ci fail quickly in case we forgot to run gas snapshots. otherwise we would need to wait for all tests to finish in the ci, which will take a while. wdyt?

avniculae: after chatting with @miguelmtzinf : will keep the check for now, and will address it later by updating gas values in the ci/cd #403

DhairyaSethi: you mean ci committing the snapshots? that's overkill, and we should know the numbers anyway before merging the pr for evaluation

avniculae: yes, this is what I meant

DhairyaSethi: which one? im disagreeing, i dont think ci should commit & according to the issue it says move to tests-merge.yml which is incorrect bc we need to know before

avniculae: this is what I meant --> ci commiting the snapshots (Miguel suggested this)

I suggested moving the check in tests-merge.yml as part of the tests, to account for cases where someone pushed on main directly

avniculae: so --> commit snapshots by ci in the PR, check snapshots by ci on main

DhairyaSethi: ci to commit is overkill, it's better if the dev commits it, gives a better idea of the change impact 
let's tackle this later & keep the check on pr like now

.github/workflows/tests-pr.yml
```diff
@@ -30,6 +30,10 @@ jobs:
         uses: bgd-labs/github-workflows/.github/actions/foundry-test@main
         with:
           FOUNDRY_PROFILE: gas
+          # even though the gas profile sets FORGE_SNAPSHOT_CHECK to true,
+          # this is overwritten by the foundry-test gh action to false, which
+          # takes priority. hence, we need to overwrite it back to true.
```
DhairyaSethi: inaccurate, pls rm

PR #402, opened 2025-05-21
Check `from` in `hub.add`

### LINKED ISSUE #400: checks on LH add method
in case of compromised spoke which passes in the `from` as the hub address itself, add validation on `from` such that it cannot equal the hub address

avniculae:
Closes #400

### DIFF
--- a/tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
+++ b/tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
@@ -1308,7 +1308,7 @@ contract SpokeRepayScenarioTest is SpokeBase {
     address caller,
     uint256 assets
   ) public {
-    vm.assume(caller != address(0));
+    vm.assume(caller != address(0) && caller != address(hub));
     reserveId = bound(reserveId, 0, spoke1.reserveCount() - 1);
     userBorrowing = bound(userBorrowing, 0, MAX_SUPPLY_AMOUNT / 2 - 1); // Allow some buffer from borrow cap
     skipTime = uint40(bound(skipTime, 0, MAX_SKIP_TIME));
@@ -1361,7 +1361,7 @@ contract SpokeRepayScenarioTest is SpokeBase {
     uint256 assets
   ) public {
     uint256 MAX_BORROW_AMOUNT = MAX_SUPPLY_AMOUNT / 2;
-    vm.assume(caller != address(0));
+    vm.assume(caller != address(0) && caller != address(hub));
     reserveId = bound(reserveId, 0, spoke1.reserveCount() - 1);
     userBorrowing = bound(userBorrowing, 0, MAX_BORROW_AMOUNT - 2); // Allow some buffer from borrow cap
     skipTime = uint40(bound(skipTime, 0, MAX_SKIP_TIME));


tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
```diff
@@ -353,7 +353,7 @@ contract SpokeWithdrawScenarioTest is SpokeBase {
     address caller,
     uint256 assets
   ) public {
-    vm.assume(caller != address(0));
+    vm.assume(caller != address(0) && caller != address(hub));
```
yan-man, line 356: dont we need to add these bounds also to hub supply tests?

avniculae: have you identified any -- if yes pls let me know? I didn't find any other fuzz tests that needed adjustments in this regard, but I did miss one spoke test: [c3004fc](https://github.com/aave/aave-v4/pull/402/commits/c3004fc22068e588f99e9abed23ca71ae6c8eee0)

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -399,6 +399,19 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.add(daiAssetId, supplyAmount, alice);
   }
 
+  function test_supply_fuzz_revertsWith_InvalidSelfAdd(uint256 daiAmount) public {
```
DhairyaSethi: this doesn't need to be a fuzzed, just the unit test is enough

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
yan-man: ah maybe I was mistaken thought we had fuzzed some `from` args, wonder if that's worthwhile fuzz test to add, wdyt

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
yan-man: yeah on amount dont think it's needed. wdyt about fuzzing on from address?

src/contracts/LiquidityHub.sol
```diff
@@ -383,9 +383,11 @@ contract LiquidityHub is ILiquidityHub {
   function _validateSupply(
     DataTypes.Asset storage asset,
     DataTypes.SpokeData storage spoke,
-    uint256 amount
+    uint256 amount,
+    address from
   ) internal view {
     require(amount != 0, InvalidSupplyAmount());
+    require(from != address(this), InvalidSelfAdd());
```
yan-man: not a fan of this error name, I'd prefer something like `InvalidAddFromHubAddress` but open to others' thoughts too if it's just nit

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: hmm in this case I want to fix `from` to the hub, to assert that the error is thrown

src/contracts/LiquidityHub.sol
avniculae: [718bb5c](https://github.com/aave/aave-v4/pull/402/commits/718bb5c6fc7c279ac2e70a16d48085223c548719)

avniculae: is this better?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: [718bb5c](https://github.com/aave/aave-v4/pull/402/commits/718bb5c6fc7c279ac2e70a16d48085223c548719)

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
avniculae: smth like this [40cb68f](https://github.com/aave/aave-v4/pull/402/commits/40cb68ff06d88011fe3924b890c93dcf0e0ccd1d) ?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -212,19 +212,26 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
   }
 
   /// @dev User makes a first supply, shares and assets amounts are correct, no precision loss
-  function test_supply_fuzz_single_asset(uint256 assetId, uint256 amount) public {
+  function test_supply_fuzz_single_asset(address user, uint256 assetId, uint256 amount) public {
```
miguelmtzinf:
```suggestion
  function test_supply_fuzz_single_asset(uint256 assetId, address user, uint256 amount) public {
```
feels more consistent

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -399,6 +406,13 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     hub.add(daiAssetId, supplyAmount, alice);
   }
 
+  function test_supply_revertsWith_InvalidAddFromHub() public {
+    vm.expectRevert(ILiquidityHub.InvalidAddFromHub.selector, address(hub));
```
DhairyaSethi, line 519: hmm dislike the syntax choice from forge here, emitter should be the first param to align with expectEmit & avoid confusion with emitter potentially being a revert arg

avniculae: I agree

avniculae: it is inconsistent with expectCall/expectCallRevert also

PR #404, opened 2025-05-22
feat: make update user rp reserveId agnostic

### LINKED ISSUE #394: Optimize update user rp function
We want to make the user rp function agnostic to reserveId. The way it's implemented, it currently accepts a reserve id that the user is borrowing. We want to remove this and use the `_notifyRiskPremiumUpdate` for this

DhairyaSethi:
closes #394 

note:
- ~is HF check necessary?~ HF check is removed
- updateRp method specific tests are pending, will be taken up separately after access control is added

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -88,6 +88,7 @@ contract Spoke is ISpoke, Multicall {
     });
 
     emit ReserveAdded(reserveId, assetId);
+    emit ReserveConfigUpdated(reserveId, config);
 
     return reserveId;
   }
--- a/tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
+++ b/tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
@@ -4,7 +4,7 @@ pragma solidity ^0.8.0;
 import 'tests/unit/Spoke/SpokeBase.t.sol';
 
 contract SpokeUpdateUserRiskPremium is SpokeBase {
-  function test_updateUserRiskPremium_success_on_rpIncrease(address caller) public {
+  function test_updateUserRiskPremium_on_rpIncrease(address caller) public {
     _deployLiquidity(spoke1, _daiReserveId(spoke1), 2500e18);
     Utils.supplyCollateral(spoke1, _wethReserveId(spoke1), alice, 1e18, alice); // 2k usd
     Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice); // 1k usd


src/contracts/Spoke.sol
```diff
@@ -323,57 +324,12 @@ contract Spoke is ISpoke, Multicall {
...
-      reserve,
-      user,
-      assetId,
-      int256(newUserPremiumDrawnShares),
-      int256(userPremiumOffset),
-      0
-    );
-    _notifyRiskPremiumUpdate(assetId, user, newUserRiskPremium);
-
+    bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, newUserRiskPremium);
```
DhairyaSethi: assumes we'll never use uint.max as `assetId`, alternatively we can begin count from 1 & use 0 as this sentinel

CheyenneAtapour: no 0 please :) I think we are good to have only uint.max -1 assets :)

src/contracts/Multicall.sol
CheyenneAtapour, line 1: Should we put this straight into `src`, or another folder, something like `dependencies`?

src/contracts/Spoke.sol
```diff
@@ -323,58 +324,13 @@ contract Spoke is ISpoke, Multicall {
...
-    );
-    _notifyRiskPremiumUpdate(assetId, user, newUserRiskPremium);
-
-    emit UserRiskPremiumUpdate(user, newUserRiskPremium);
+  /// @inheritdoc ISpoke
+  function updateUserRiskPremium(address user) external {
+    (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
+    bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
+    // todo allow authorized caller to increase as well
+    require(msg.sender == user || premiumIncrease, UnAuthorized());
```
CheyenneAtapour: `UnAuthorized` seems a bit weird to me as the error, because it doesn't indicate the person is unauthorized because the user rp would go up. What about like `UnauthorizedRPIncrease`? In any case, "Unauthorized" is itself one word, so we don't need to capitalize the "Authorized" part separately imo

src/contracts/Spoke.sol
```diff
@@ -861,11 +819,17 @@ contract Spoke is ISpoke, Multicall {
         userPosition.premiumOffset = HUB.previewOffset(assetId, userPosition.premiumDrawnShares);
         userPosition.realizedPremium += accruedUserPremium;
 
+        int256 premiumDrawnSharesDelta = _signedDiff(
+          userPosition.premiumDrawnShares,
+          oldUserPremiumDrawnShares
+        );
+        if (!premiumIncrease) premiumIncrease = premiumDrawnSharesDelta > 0;
```
CheyenneAtapour, line 827: Why is there an if clause here? `premiumIncrease` is always false currently. Are there future plans to alter this behavior?

src/interfaces/ISpoke.sol
```diff
@@ -100,19 +103,41 @@ interface ISpoke {
...
+  /**
+   * @notice Allows suppliers to enable/disable a specific supplied reserve as collateral.
+   * @param reserveId The reserveId of the underlying asset as registered on the spoke.
+   * @param usingAsCollateral True if the user wants to use the supply as collateral, false otherwise.
+   */
   function setUsingAsCollateral(uint256 reserveId, bool usingAsCollateral) external;
-  function updateUserRiskPremium(uint256 reserveId, address user) external;
+
+  /**
+   * @notice Allows committing an updated risk premium on all user specified positions.
```
CheyenneAtapour: user does not get to _specify_ the positions on which their user risk premium is updated, so imo omit the word "specified"

src/contracts/Spoke.sol
yan-man: yeah think we had discussed starting at 1 before, forgot what the context was. Is it purely just design choice or did it affect things elsewhere

src/contracts/Multicall.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: MIT
```
yan-man: and do we need attribution of any sort?

src/interfaces/ISpoke.sol
yan-man:
```suggestion
   * @notice Allows updating risk premium on all user positions.
```
could also rm the committing part

src/interfaces/ISpoke.sol
```diff
@@ -100,19 +103,41 @@ interface ISpoke {
...
+   * @dev If the risk premium has increased, the sender must be authorized or the owner of the positions,
```
yan-man:
```suggestion
   * @dev If the risk premium has increased, the sender must be authorized or the owner of the positions.
```

maybe caller instead of sender?

could also start a new line on `Reverts with...`

src/interfaces/IMulticall.sol
```diff
@@ -0,0 +1,14 @@
...
+pragma solidity ^0.8.0;
+
+/**
+ * @title IMulticall
+ */
+interface IMulticall {
+  /**
+   * @notice Call multiple functions in the current contract and return the data from all of them if they all succeed
+   * @param data The encoded function data for each of the calls to make to this contract
+   * @return results The results from each of the calls passed in via data
```
yan-man: periods on natspec

src/interfaces/IMulticall.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/**
+ * @title IMulticall
```
yan-man: do we need author?

src/contracts/Spoke.sol
DhairyaSethi: it's a sanity, since all rp's should match atm, but it's better to not override the boolean bc we're just looking for the first truthy val

src/contracts/Multicall.sol
DhairyaSethi: already have dependencies & thats for external deps. put it in `misc`

src/contracts/Multicall.sol
DhairyaSethi: it's consistent, will add more context

miguelmtzinf: we can do smt like [this](https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/misc/aave-upgradeability/VersionedInitializable.sol#L6)

src/interfaces/ISpoke.sol
miguelmtzinf: there is only 1 user position imo, 1 single user rp

src/contracts/Spoke.sol
miguelmtzinf: I think `Unauthorized` is fine

src/contracts/Spoke.sol
miguelmtzinf:
```suggestion
        if (!premiumIncrease && premiumDrawnSharesDelta > 0;) premiumIncrease = true
```
I though this was cheaper, but remix does not say the same :thinking:

DhairyaSethi: yeah i think bc the former reuses the result from gt(premiumShares, 0) and the latter adds an extra push1 0x1; regardless something compiler should opt

PR #405, opened 2025-05-22
feat: dynamic risk config

### LINKED ISSUE #266: Dynamic Risk Configuration
- store the risk config associated with the asset at the moment the user borrows
- change the liq threshold only affecting the new users
- risk manager should be change the risk config of new borrowers
    - update on withdraw
    - we still need to give the risk managers the ability to update old positions
- 2 func `addCollateralFactor` and `updateCollateralFactor`

DhairyaSethi:
closes #266 

## Design
Introduces `dynamic` configuration for reserve, adding on to the current implicitly `static` (ie same for all active users) configuration. Moves `collateralFactor` from `static` to `dynamic` configuration.
This is done by introducing new mapping for dynamic configurations uniquely referenced by a `configKey`. 
The most up to date key is maintained on each reserve and each user has a snapshot of their current dynamic configuration through this key. 
For each user position, this dynamic configuration is only updated from the trailing values when a user performs an action to worsen their position ie through disableUsingAsCollateral, withdraw, & borrow. We also refresh when a user initially sets an asset as collateral.

## Feature Notes
- ConfigKey is kept as a uint16 (to be decided is uint8 is fine as well, ie 255 max active configs) ie ~65k max active configs. Keys wrap around on exhaustion and invalidate the most stale keys in order.
- ConfigKey is updated to the latest state for all reserves for a userPosition on:
  - enable/disable usingAsCollateral
  - borrow
  - withdraw
- ConfigKey is not updated and trailing configs for all reservers for a userPosition on:
  - supply
  - repay
  - liquidationCall
  - updateRiskPremium
### Edge Cases
- It's expected that the config can be out of date/inconsistent for non debt actions, for example a user an activate as collateral when dynamic config is in his favour and supply later.

## Future Todos 
- Allow governance to overwrite any previous `configKey` #448 
- Add getter for users so they can preview their position after the refresh.
- Move liquidation bonus to dynamic config. #459


src/contracts/Spoke.sol
```diff
@@ -64,11 +66,13 @@ contract Spoke is ISpoke {
...
   function addReserve(
     uint256 assetId,
-    DataTypes.ReserveConfig calldata config
+    DataTypes.ReserveConfig calldata config,
+    DataTypes.DynamicReserveConfig calldata dynamicConfig
   ) external returns (uint256) {
     _validateReserveConfig(config);
     address asset = address(HUB.assetsList(assetId)); // will revert on invalid assetId
     uint256 reserveId = reserveCount++;
+    uint16 dynamicConfigKey = 1;
```
CheyenneAtapour: why not instead start from 0?

src/contracts/Spoke.sol
```diff
@@ -103,27 +98,34 @@ contract Spoke is ISpoke {
...
+    DataTypes.DynamicReserveConfig calldata dynamicConfig
+  ) external {
+    // TODO: AccessControl, More sophisticated
+    DataTypes.Reserve storage reserve = _reserves[reserveId];
+    uint16 dynamicConfigKey;
+    // @dev overflow is desired, we implicitly invalidate & override stale config
+    unchecked {
+      dynamicConfigKey = ++reserve.dynamicConfigKey;
+    }
+    // todo opt: cancat key to use single lookup
```
CheyenneAtapour: nit, but it's just a todo comment anyway: `concat` spelling

src/contracts/Spoke.sol
```diff
@@ -336,7 +341,7 @@ contract Spoke is ISpoke {
       address debtAsset,
       uint256 debtToLiquidate,
       uint256 collateralToLiquidate,
-      uint256 liquidationProtocolFeeShares // TODO: emit in event
+      uint256 liquidationProtocolFeeShares // TODO: emit in event */
```
CheyenneAtapour: extraneous `*/`

src/contracts/Spoke.sol
```diff
@@ -465,18 +474,38 @@ contract Spoke is ISpoke {
...
+    return _reserves[reserveId].config;
+  }
+
+  function getDynamicReserveConfig(
+    uint256 reserveId
+  ) external view returns (DataTypes.DynamicReserveConfig memory) {
+    return _dynamicConfig[reserveId][_reserves[reserveId].dynamicConfigKey];
+  }
+
+  function getDynamicReserveConfig(
```
CheyenneAtapour, line 503: Is there a possible consideration to call this one `getDynamicReserveConfigByKey`? Or is that just unnecessary in your opinion?

src/libraries/types/DataTypes.sol
```diff
@@ -57,13 +57,14 @@ library DataTypes {
...
     uint256 assetId;
-    address asset;
+    address asset; // todo rm not needed
     uint256 suppliedShares;
     uint256 baseDrawnShares;
     uint256 premiumDrawnShares;
     uint256 premiumOffset;
     uint256 realizedPremium;
     ReserveConfig config;
+    uint16 dynamicConfigKey;
```
miguelmtzinf:
```suggestion
    uint16 dynamicConfigKey; // key of the last reserve config 
```

src/libraries/types/DataTypes.sol
```diff
@@ -73,19 +74,23 @@ library DataTypes {
...
+  }
+
   struct UserPosition {
     bool usingAsCollateral;
     uint256 suppliedShares;
     uint256 baseDrawnShares;
     uint256 premiumDrawnShares;
     uint256 premiumOffset;
     uint256 realizedPremium;
+    uint16 configKey;
```
miguelmtzinf:
```suggestion
    uint16 configKey; // key of the reserve config for the user
```

src/contracts/Spoke.sol
```diff
@@ -103,27 +98,34 @@ contract Spoke is ISpoke {
...
   }
 
+  function updateDynamicReserveConfig(
+    uint256 reserveId,
+    DataTypes.DynamicReserveConfig calldata dynamicConfig
+  ) external {
+    // TODO: AccessControl, More sophisticated
+    DataTypes.Reserve storage reserve = _reserves[reserveId];
+    uint16 dynamicConfigKey;
+    // @dev overflow is desired, we implicitly invalidate & override stale config
```
miguelmtzinf, line 120: so we can basically update the config 65536 (2**16) times "only". After that, we won't be able to update the CF again which sounds v bad (although unlikely). Options:
1. We override the last one so we loose dynamic config feature at some point
2. We override the oldest one. This is nice (iirc uniswap v3 does something similar) and allow us to use a smaller type (256 with uint8 is not bad)

DhairyaSethi: this implements option 2 only, we loop back the first key after 2**16 are exhausted and so on bc of the unchecked add, will clarify more with comments

src/contracts/Spoke.sol
DhairyaSethi: think it's fine to overload here imo, `getDynamicReserveConfig` returns value with the latest config key unless specified

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
```diff
@@ -248,51 +249,5 @@ contract LiquidationCallValidationTest is SpokeLiquidationBase {
     vm.expectRevert(ISpoke.CollateralCannotBeLiquidated.selector);
     spoke1.liquidationCall(wethReserveId, daiReserveId, alice, debtToCover);
   }
-
-  function test_liquidationCall_revertsWith_CollateralCannotBeLiquidated_collateralFactor() public {
```
DhairyaSethi, line 252: the following two tests are not relevant anymore since the idea of this feature is opt-in updates and existing positions cannot be made underwater by lower cf anymore, unless the user intentionally potentially worsens their position by borrowing more/withdrawing/disabling coll

src/contracts/Spoke.sol
```diff
@@ -104,27 +100,35 @@ contract Spoke is ISpoke, Multicall {
...
+  ) external {
+    _validateDynamicReserveConfig(dynamicConfig);
+    // TODO: AccessControl, More sophisticated
+    DataTypes.Reserve storage reserve = _reserves[reserveId];
+    uint16 dynamicConfigKey;
+    // @dev overflow is desired, we implicitly invalidate & override stale config
+    unchecked {
+      dynamicConfigKey = ++reserve.dynamicConfigKey;
+    }
+    // todo opt: concat key to use single lookup
```
yan-man, line 124: what key to concat for this opt?

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
```diff
@@ -0,0 +1,66 @@
...
+      emit ISpoke.DynamicReserveConfigUpdated(reserveId, dynamicConfigKey, dynConf);
+      vm.prank(SPOKE_ADMIN);
+      spoke1.updateDynamicReserveConfig(reserveId, dynConf);
+
+      configs[reserveId].key = dynamicConfigKey;
+      assertEq(_getSpokeDynConfigKeys(spoke1), configs);
+    }
+  }
+
+  // todo test key overwrites stale slot, dynamically determine struct size & overwrite dynamicConfigKey or use mock spoke
```
yan-man, line 117: are these in a ticket too

src/contracts/Spoke.sol
DhairyaSethi: keys of the map: reserveId & configKey; will evaluate if it's cheaper

src/contracts/Spoke.sol
```diff
@@ -526,17 +551,15 @@ contract Spoke is ISpoke, Multicall {
     // todo validate user not trying to repay more
   }
 
-  function _validateUserPosition(address userAddress) internal view returns (uint256) {
-    (uint256 userRiskPremium, , uint256 healthFactor, , ) = _calculateUserAccountData(userAddress);
+  function _validateAndRefreshUserPosition(address user) internal returns (uint256) {
```
miguelmtzinf: i d revert the name change, because i) refresh actually happens first and ii) will be done inside the calculate hf fn

DhairyaSethi: re (ii) it's now outside the calculate hf function, that takes user's trailing config key and expects it to be updated before it needed, i'd prefer this method is clear about it's purpose ie refreshing & validating so wdyt about refreshAndValidate instead?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
miguelmtzinf: untrue, DAO can decide to change an old CF no? Affecting users right away

tests/Base.t.sol
```diff
@@ -692,6 +704,21 @@ abstract contract Base is Test {
     assertEq(spoke.getReserve(reserveId).config.liquidationProtocolFee, newLiquidationProtocolFee);
   }
 
+  function updateCollateralFactor(
+    ISpoke spoke,
+    function(ISpoke) pure returns (uint256) reserveIdFn,
```
miguelmtzinf, line 732: why do we need it?

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -745,4 +729,101 @@ contract SpokeBase is Base {
...
+  }
+
+  // deref and return current UserDynamicReserveConfig for a specific reserveId on user position.
+  function _getUserDynConfigKeys(
+    ISpoke spoke,
+    address user,
+    uint256 reserveId
+  ) internal view returns (DynamicConfig memory) {
+    DataTypes.UserPosition memory pos = spoke.getUserPosition(reserveId, user);
+    return DynamicConfig(pos.configKey, pos.usingAsCollateral);
```
miguelmtzinf, line 827: i wonder if 
i) using 0 value as `configKey` (valid pointer), or
ii) leaving a valid `configKey` even if the user has not activated the asset as collateral
could be problematic or bring any edge case... I do think it's fine

tests/Base.t.sol
DhairyaSethi: syntax sugar for our _daiReserveId(spoke1) etc helpers such that we can do `update(spoke1, _daiReserveId)` instead of `update(spoke1, _daiReserveId(spoke1))`

tests/unit/Spoke/SpokeBase.t.sol
DhairyaSethi: there's no uninitialised configKey, it's only a matter of enabledAsCollateral, and having this helper struct retain this info by using enabled config helpers keep the assertion helpers on this (comparing user & spoke state) clean

src/contracts/Spoke.sol
```diff
@@ -64,11 +67,13 @@ contract Spoke is ISpoke, Multicall {
...
   function addReserve(
     uint256 assetId,
-    DataTypes.ReserveConfig calldata config
+    DataTypes.ReserveConfig calldata config,
+    DataTypes.DynamicReserveConfig calldata dynamicConfig
   ) external returns (uint256) {
     _validateReserveConfig(config);
     address asset = address(HUB.assetsList(assetId)); // will revert on invalid assetId
     uint256 reserveId = reserveCount++;
+    uint16 dynamicConfigKey;
```
```suggestion
    uint16 dynamicConfigKey = 0; // 0 as first key to use
```

src/contracts/Spoke.sol
```diff
@@ -104,27 +100,35 @@ contract Spoke is ISpoke, Multicall {
...
   }
 
+  function updateDynamicReserveConfig(
+    uint256 reserveId,
+    DataTypes.DynamicReserveConfig calldata dynamicConfig
+  ) external {
+    _validateDynamicReserveConfig(dynamicConfig);
+    // TODO: AccessControl, More sophisticated
+    DataTypes.Reserve storage reserve = _reserves[reserveId];
+    uint16 dynamicConfigKey;
```
miguelmtzinf: i d rename to `nextConfigKey`

src/contracts/Spoke.sol
```diff
@@ -465,6 +469,7 @@ contract Spoke is ISpoke, Multicall {
       uint256 totalDebtInBaseCurrency
     )
   {
+    // todo separate getter with refreshed config for users trying to incrementally build hf?
```
miguelmtzinf, line 477: yeah, this is needed. Let's create a ticket

src/contracts/Spoke.sol
```diff
@@ -104,27 +100,35 @@ contract Spoke is ISpoke, Multicall {
...
+    DataTypes.Reserve storage reserve = _reserves[reserveId];
+    uint16 dynamicConfigKey;
+    // @dev overflow is desired, we implicitly invalidate & override stale config
+    unchecked {
+      dynamicConfigKey = ++reserve.dynamicConfigKey;
+    }
+    // todo opt: concat key to use single lookup
+    _dynamicConfig[reserveId][dynamicConfigKey] = dynamicConfig;
+    emit DynamicReserveConfigUpdated(reserveId, dynamicConfigKey, dynamicConfig);
+    // todo emit if stale config overwritten?
```
miguelmtzinf, line 127: related to updating previous dynamic configs.. can you create a ticket?

src/contracts/Spoke.sol
```diff
@@ -474,15 +479,35 @@ contract Spoke is ISpoke, Multicall {
     ) = _calculateUserAccountData(user);
   }
 
-  // public
+  // todo: make `getReserve` return dynamic config as well
```
miguelmtzinf: why not doing it now? or it ruins our tests
anyway, I think it's fine the way it is. You get the key from `getReserve` and call `getDynamicReserveConfig`

src/contracts/Spoke.sol
```diff
@@ -755,9 +791,13 @@ contract Spoke is ISpoke, Multicall {
     vars.i = 0;
     vars.reserveId = 0;
     while (vars.reserveId < reservesListLength) {
-      DataTypes.UserPosition storage userPosition = _userPositions[userAddress][vars.reserveId];
+      DataTypes.UserPosition storage userPosition = _userPositions[user][vars.reserveId];
       DataTypes.Reserve storage reserve = _reserves[vars.reserveId];
       if (_usingAsCollateral(userPosition)) {
+        DataTypes.DynamicReserveConfig storage dynConfig = _dynamicConfig[vars.reserveId][
```
miguelmtzinf, line 783: why not using it directly, without this pointer declaration

src/contracts/Spoke.sol
```diff
@@ -930,6 +968,21 @@ contract Spoke is ISpoke, Multicall {
...
+    while (reserveId < reservesListLength) {
+      DataTypes.UserPosition storage userPosition = _userPositions[user][reserveId];
+      if (_usingAsCollateral(userPosition)) {
+        userPosition.configKey = _reserves[reserveId].dynamicConfigKey;
+      }
+      unchecked {
+        ++reserveId;
+      }
+    }
+    // todo emit event UserConfigRefreshed(user)?
```
miguelmtzinf: Please add. Suggestion: `UserDynamicConfigRefreshed(address indexed user)`

src/contracts/Spoke.sol
```diff
@@ -930,6 +968,21 @@ contract Spoke is ISpoke, Multicall {
     return premiumIncrease;
   }
 
+  function _refreshConfig(address user) internal {
```
```suggestion
  function _refreshDynamicConfig(address user) internal {
```

src/contracts/Spoke.sol
```diff
@@ -1225,8 +1281,10 @@ contract Spoke is ISpoke, Multicall {
 
     // If unsetting, check HF and update user rp
     if (!usingAsCollateral) {
-      uint256 newUserRiskPremium = _validateUserPosition(user); // validates HF
+      uint256 newUserRiskPremium = _validateAndRefreshUserPosition(user); // validates HF
       _notifyRiskPremiumUpdate(type(uint256).max, user, newUserRiskPremium);
+    } else {
```
miguelmtzinf, line 1281: for readability, i d flip the order, so positive case first

src/interfaces/ISpoke.sol
```diff
@@ -13,6 +13,11 @@ interface ISpoke is IMulticall {
   event ReserveAdded(uint256 indexed reserveId, uint256 indexed assetId);
   event ReserveConfigUpdated(uint256 indexed reserveId, DataTypes.ReserveConfig config);
   event LiquidityPremiumUpdated(uint256 indexed reserveId, uint256 liquidityPremium);
+  event DynamicReserveConfigUpdated(
+    uint256 indexed reserveId,
+    uint16 configKey,
```
miguelmtzinf: should we index too?

src/libraries/types/DataTypes.sol
```diff
@@ -73,19 +74,23 @@ library DataTypes {
...
+  }
+
   struct UserPosition {
     bool usingAsCollateral;
     uint256 suppliedShares;
     uint256 baseDrawnShares;
     uint256 premiumDrawnShares;
     uint256 premiumOffset;
     uint256 realizedPremium;
+    uint16 configKey; // key of the last reserve config
```
miguelmtzinf:
```suggestion
    uint16 configKey; // key of the last user config
```

src/contracts/Spoke.sol
```diff
@@ -36,6 +36,8 @@ contract Spoke is ISpoke, Multicall {
   mapping(address user => mapping(uint256 reserveId => DataTypes.UserPosition position))
     internal _userPositions;
   mapping(uint256 reserveId => DataTypes.Reserve reserveData) internal _reserves;
+  mapping(uint256 reserveId => mapping(uint16 configKey => DataTypes.DynamicReserveConfig config))
```
miguelmtzinf, line 37: this is actually a dictionary of dynamic configurations. Worth adding a quick comment

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -0,0 +1,128 @@
...
+  function test_borrow_triggers_dynamicConfigUpdate() public {
+    DynamicConfig[] memory configs = _getUserDynConfigKeys(spoke1, alice);
+
+    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
+    _deployLiquidity(spoke1, _daiReserveId(spoke1), 600e18);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), alice, 500e18, alice);
+    configs = _getUserDynConfigKeys(spoke1, alice);
+    skip(322 days);
+
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 0);
```
miguelmtzinf, line 90: ping @DhairyaSethi

tests/Base.t.sol
miguelmtzinf: i saw a mix of things below. Pinged you there

> 
    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 0);

miguelmtzinf: ah, it's duplicated... can you fix so we only have 1 helper for this please?

tests/unit/LiquidityHub/LiquidityHubRiskPremium.t.sol
miguelmtzinf, line 1: why are we getting all of these diffs now?

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -607,35 +614,12 @@ contract SpokeBase is Base {
...
-      actual.config.liquidationBonus,
-      'Liquidation bonus config mismatch'
-    );
-    assertEq(
-      expected.config.liquidityPremium,
-      actual.config.liquidityPremium,
-      'Liquidity premium config mismatch'
-    );
-    assertEq(abi.encode(expected), abi.encode(actual), 'Encoded reserve mismatch'); // sanity check
+  function assertEq(DataTypes.Reserve memory a, DataTypes.Reserve memory b) internal pure {
```
miguelmtzinf, line 672: THANKS for all these helpers :1st_place_medal:

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
```diff
@@ -0,0 +1,66 @@
...
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+contract SpokeDynamicConfigTest is SpokeBase {
+  using SafeCast for uint256;
+
+  function test_updateDynamicReserveConfig_fuzz_revertsWith_InvalidCollateralFactor() public {
+    uint16 collateralFactor = bound(
+      vm.randomUint(),
```
miguelmtzinf: preferring this to save on fuzzing runs? i d remove the fuzz preffix from test name though

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
```diff
@@ -0,0 +1,66 @@
...
+      emit ISpoke.DynamicReserveConfigUpdated(reserveId, dynamicConfigKey, dynConf);
+      vm.prank(SPOKE_ADMIN);
+      spoke1.updateDynamicReserveConfig(reserveId, dynConf);
+
+      configs[reserveId].key = dynamicConfigKey;
+      assertEq(_getSpokeDynConfigKeys(spoke1), configs);
+    }
+  }
+
+  function test_fuzz_updateDynamicReserveConfig_trailing_order(bytes32) public {
```
miguelmtzinf, line 43:
```suggestion
  function test_updateDynamicReserveConfig_trailing_order() public {
```

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -745,4 +729,101 @@ contract SpokeBase is Base {
...
+  function assertNotEq(DynamicConfig[] memory a, DynamicConfig[] memory b) internal pure {
+    require(a.length == b.length);
+    for (uint256 i; i < a.length; ++i) {
+      if (a[i].enabled && b[i].enabled) {
+        assertNotEq(a[i].key, b[i].key, string.concat('reserve ', vm.toString(i)));
+      }
+    }
+  }
+
+  function _nextConfigKey(ISpoke spoke, uint256 reserveId) internal view returns (uint16) {
```
miguelmtzinf:
```suggestion
  function _nextDynamicConfigKey(ISpoke spoke, uint256 reserveId) internal view returns (uint16) {
```

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
miguelmtzinf: also, can you clarify what's the difference between this one and the previous test?
just updating multiple times, so we end up overriding? we are not overriding with 100 iterations only though

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -0,0 +1,128 @@
...
+  function test_supply_does_not_trigger_dynamicConfigUpdate() public {
+    DynamicConfig[] memory configs = _getUserDynConfigKeys(spoke1, alice);
+
+    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), _randomBps());
+
+    assertEq(_getUserDynConfigKeys(spoke1, alice), configs);
+
+    _deployLiquidity(spoke1, _daiReserveId(spoke1), 500e18);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), alice, 500e18, alice);
```
miguelmtzinf, line 18: we need to borrow because until now (first time the refresh happens), the user was pointing to an incorrect dynamic config (0 by default can be not the correct one).
This is a bit counter intuitive though

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -0,0 +1,128 @@
...
+    configs = _getUserDynConfigKeys(spoke1, alice);
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), _randomBps());
+
+    assertEq(_getUserDynConfigKeys(spoke1, alice), configs);
+
+    Utils.supply(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
+
+    // user config should not change
+    assertEq(_getUserDynConfigKeys(spoke1, alice), configs);
+    assertNotEq(_getSpokeDynConfigKeys(spoke1), configs);
```
miguelmtzinf, line 28: should we rename to `_getReserveDynConfigKeys`?
we ve been using a mix of naming convention, but majority of functions use `spoke` for hub side, `reserve` for spoke side

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -0,0 +1,128 @@
...
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 90_10);
+    skip(322 days);
+    Utils.repay(spoke1, _daiReserveId(spoke1), alice, UINT256_MAX);
+
+    // user config should not change
+    assertEq(_getUserDynConfigKeys(spoke1, alice), configs);
+    assertNotEq(_getSpokeDynConfigKeys(spoke1), configs);
+  }
+
+  function test_liquidate_does_not_trigger_dynamicConfigUpdate() public {
```
miguelmtzinf, line 47: even in the case where a liquidation makes the position less healthy, it's better not to trigger the update imo.

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
miguelmtzinf, line 1: can we add a test case to offboard an asset? updateCollFactor to 0, no new borrows can happen, old users unaffected
feels like a must to have, since it's one of the end goals of this feature

tests/Base.t.sol
DhairyaSethi: yeah the dup was intentional bc i wanted to override that method to accept both kinds of args for the syntax sugar tbh

src/contracts/Spoke.sol
DhairyaSethi: yeah not doing now bc larger test refactor required. tbh i think it all of these should be overloaded and we can tackle this part on the DATA_PROVIDER contract we should eventually write

src/contracts/Spoke.sol
DhairyaSethi: for verbosity on which key was used, can rm tho

tests/unit/LiquidityHub/LiquidityHubRiskPremium.t.sol
DhairyaSethi: linter accidentally ran on these files 😭 
no change on these in this PR

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
DhairyaSethi: spoke makes sense to me bc this returns an array of **all reserves** in a spoke, unlike others where we're dealing with a single reserve. lmk your thoughts

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
DhairyaSethi: liq can never make the position less healthy, bc that's bad debt which is absorbed by the protocol rather than letting the posn remain, no? regardless agree on liq should not trigger update, did you mean to also validate a specific liq case?

src/contracts/Spoke.sol
DhairyaSethi: #448

src/contracts/Spoke.sol
DhairyaSethi: #448

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
DhairyaSethi: correct, but that outside the scope of this PR. noted in #448

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
DhairyaSethi: nice yep added [0c051be](https://github.com/aave/aave-v4/pull/405/commits/0c051bea8b97681872b334f61409beca9cae8780)

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
DhairyaSethi: as discussed on sync & for visibility: we use a unnamed fuzz param to get multiple iterations on this and we choose to generate fuzzed input (coll factor) randomly rather than through the fuzzer bc that value is never acted on but rather only stored (after bounding) & read

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
DhairyaSethi: yes it's easier to run this test when overriding col factor is added so leaving it for #448

src/contracts/Spoke.sol
miguelmtzinf: no big deal. We may need it later if we add more stuff to the dynamic config

tests/Base.t.sol
miguelmtzinf: please add this suggestion to #430

src/contracts/Spoke.sol
miguelmtzinf: imo we can use a data provider and this is fine as is. i d remove the comment

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
miguelmtzinf: understood. there is no tests that overwrites (going above 2**16) old keys... would be nice to have that, any idea?

DhairyaSethi: yep that's a todo, more info https://github.com/aave/aave-v4/pull/405#discussion_r2148014343

src/contracts/Spoke.sol
miguelmtzinf: can you rm now @DhairyaSethi ?

src/contracts/Spoke.sol
```diff
@@ -472,15 +477,35 @@ contract Spoke is ISpoke, Multicall {
     ) = _calculateUserAccountData(user);
   }
 
-  // public
+  // todo: make `getReserve` return dynamic config as well
   function getReserve(uint256 reserveId) external view returns (DataTypes.Reserve memory) {
     return _reserves[reserveId];
   }
 
+  function getReserveConfig(
```
miguelmtzinf, line 491: not completely sure why we need this, since can be accessed by `getReserve(reserveId).config` no?

src/contracts/Spoke.sol
DhairyaSethi: yep will do

src/contracts/Spoke.sol
DhairyaSethi: yes but it seems like a nice to have distinction since we have getDynamicReserveConfig
and it's a cheaper method for external contracts looking to pull in the just the config

PR #411, opened 2025-05-23
Enrich assetConfigUpdated event

### LINKED ISSUE #410: Enrich AssetConfigUpdates
- add `config` to AssetConfigUpdated event
- emit AssetConfigUpdated on LiquidityHub.addAsset

CheyenneAtapour:
closes #410

### DIFF
--- a/tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
@@ -93,12 +93,16 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
 
     config.paused = true;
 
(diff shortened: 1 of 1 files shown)


tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
```diff
@@ -279,6 +279,16 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
     );
   }
 
+  function test_updateAssetConfig_event_emission() public {
```
DhairyaSethi: don't need this test, i would rather just add an expectEmit on all other updateAssetConfig tests but ok

PR #412, opened 2025-05-23
Collateral Unset

### LINKED ISSUE #408: Unset As Collateral HF Check and RP Update
When unsetting a reserve as collateral, there should be an HF check, as well as an RP update

CheyenneAtapour:
Unsetting a reserve as collateral should trigger a health factor check and an RP update

Closes #408

### DIFF
--- a/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.EdgeCases.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.EdgeCases.t.sol
@@ -38,8 +38,13 @@ contract LiquidationCallEdgeCasesTest is SpokeLiquidationBase {
     // position is liquidatable
     assertLt(spoke1.getHealthFactor(alice), HEALTH_FACTOR_LIQUIDATION_THRESHOLD);
 
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.UsingAsCollateral(_usdxReserveId(spoke1), alice, false);
     vm.prank(bob);
     spoke1.liquidationCall(_usdxReserveId(spoke1), _usdyReserveId(spoke1), alice, UINT256_MAX);
+
+    // Alice's usdx collateral unset
+    assertFalse(spoke1.getUsingAsCollateral(_usdxReserveId(spoke1), alice));
 
     // all collateral liquidated without overflowing
     assertEq(
@@ -86,8 +91,13 @@ contract LiquidationCallEdgeCasesTest is SpokeLiquidationBase {
     // position is liquidatable
     assertLt(spoke1.getHealthFactor(alice), HEALTH_FACTOR_LIQUIDATION_THRESHOLD);
 
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.UsingAsCollateral(_usdxReserveId(spoke1), alice, false);
     vm.prank(bob);
     spoke1.liquidationCall(_usdxReserveId(spoke1), _usdyReserveId(spoke1), alice, UINT256_MAX);
+
+    // Alice's usdx collateral unset
+    assertFalse(spoke1.getUsingAsCollateral(_usdxReserveId(spoke1), alice));
 
     // all collateral liquidated without overflowing
     assertEq(


src/contracts/Spoke.sol
```diff
@@ -320,18 +320,21 @@ contract Spoke is ISpoke, Multicall {
     _validateSetUsingAsCollateral(reserve, userPosition, usingAsCollateral);
     userPosition.usingAsCollateral = usingAsCollateral;
 
-    // consider updating user rp & notify here especially when deactivating collateral
+    // If unsetting, check HF and update user rp
+    if (!usingAsCollateral) {
+      require(
+        this.getHealthFactor(msg.sender) >= HEALTH_FACTOR_LIQUIDATION_THRESHOLD,
+        HealthFactorBelowThreshold()
+      );
```
DhairyaSethi: we can reuse _validateUserPosition method here which does this

src/contracts/Spoke.sol
```diff
@@ -320,18 +320,21 @@ contract Spoke is ISpoke, Multicall {
...
+      _executeUserRiskPremiumUpdate(msg.sender);
```
DhairyaSethi: very expensive bc we recalc risk premium in this when we've already done in the above check, should figure out a way to avoid dup work

src/contracts/Spoke.sol
```diff
@@ -516,6 +519,14 @@ contract Spoke is ISpoke, Multicall {
     return (amount - premiumDebt, premiumDebt);
   }
 
+  function _executeUserRiskPremiumUpdate(address user) internal {
+    (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
+    bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
+    // todo allow authorized caller to increase as well
+    require(msg.sender == user || !premiumIncrease, Unauthorized());
```
DhairyaSethi: if this method is being made internal then we should not use msg.sender inside but rather extract out use as param

yan-man: i remember we had a discussion about this somewhere, but added here for ref: https://docs.craft.do/editor/d/55fd6b0d-5d65-45f2-9cda-2522b781c9a1/4b68a791-0589-4fe0-9c1e-e64520b174bc/b/ebdf48c7-5ee9-4013-a7ca-411b52b0fe3c?s=T7LCSjBq3spRpSZP6D98nE7T8aT1aVg8rRqFS54EannN#BF1EF09D-D524-49C3-BACB-0626341F3313

src/contracts/Spoke.sol

yan-man: separate side q - in general should we avoid the this.method calls for external methods and instead make the method public?

src/contracts/Spoke.sol
```diff
@@ -320,7 +320,11 @@ contract Spoke is ISpoke, Multicall {
     _validateSetUsingAsCollateral(reserve, userPosition, usingAsCollateral);
     userPosition.usingAsCollateral = usingAsCollateral;
 
-    // consider updating user rp & notify here especially when deactivating collateral
+    // If unsetting, check HF and update user rp
+    if (!usingAsCollateral) {
```
yan-man: what if the `usingAsCollateral` is already false and user calls setting to false again, should we short circuit, or still do this?

CheyenneAtapour: I like short-circuiting when it's unchanged. Pushed these changes. Check happens in validation

src/contracts/Spoke.sol
DhairyaSethi: yes, this.method does an eth_call whereas calling public methods is a jump, so much cheaper. this.method is required in very specific cases where we'd want a new context or something

src/contracts/Spoke.sol
```diff
@@ -1041,6 +1041,7 @@ contract Spoke is ISpoke, Multicall {
       // TODO: not compulsory, decide whether to rm
       if (vars.newUserSuppliedShares == 0) {
         userCollateralPosition.usingAsCollateral = false;
+        emit UsingAsCollateral(collateralReserveId, users[vars.i], false);
```
yan-man, line 1044: might be good to add this emit event to a test, it could get annoying on each liq test scenario where this happens but Spoke.LiquidationCall.EdgeCases.t.sol should unset as collateral always. Can you also add an assertion there for `assertFalse(spoke1.getUsingAsCollateral(_usdxReserveId(spoke1), alice))` too

CheyenneAtapour: Done and pushed

PR #427, opened 2025-05-30
feat: healthFactorForFullLiquidation config field

### LINKED ISSUE #380: liquidation config - health factor threshold to liquidate all debt
discussed with Dhairya/Emilio on 5.5.25 - currently liquidatable debt is capped by the fact there is a close factor, ie a final health factor that user should have post-liq. This means that due to definition of HF:

`HF = avg collateral factor * collateral in base / debt in base`

debt in base must be > 0. Only case where debt can be fully liquidated currently is if remaining collateral exactly == debt, allowing a "perfect liquidation". this is a rare edge case. 

in order to allow debt to fully be liquidated and clear away potential dust debt positions remaining, allow admin to configure a health factor threshold such that if a user has a HF under this value, all debt can be liquidated and close factor does not apply.

yan-man:
merge after https://github.com/aave/aave-v4/pull/389

closes #380

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -505,6 +505,10 @@ contract Spoke is ISpoke, Multicall {
       config.healthFactorForMaxBonus < HEALTH_FACTOR_LIQUIDATION_THRESHOLD,
       InvalidHealthFactorForMaxBonus()
     );
+    require(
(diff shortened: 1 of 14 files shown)


yan-man:
instead, enforce liquidation dust thresholds on debt and collateral: https://github.com/aave/aave-v4/pull/796; https://github.com/aave/aave-v4/pull/605

Issue #430, opened 2025-06-02
Refactor tests


miguelmtzinf:
- [ ] Remove all Solidity compiler warnings from tests files.
- [ ] Use `supplyCollateral` helper instead of `supply` and `setUsingAsCollateral` separate call. 
- [ ] Create more sophisticated helpers to assert supply and debt exchange rate remains always equal or increases overtime between actions.
- [ ] Use `_deployLiquidity` helper instead of alice supply's.

DhairyaSethi:
- add more sophisticated helpers to bound fuzzed collateral amount used to cover debts (currently we heavily rely on vm.assume's) => addressed for hf tests [here](https://github.com/aave/aave-v4/pull/431/files#diff-e8b2ce6c280565449a53f150645df5edbb2c5ddbc11ac8fa6fc6675fbbd5af3dR404-R418)

yan-man:
> add more sophisticated helpers to bound fuzzed collateral amount used to cover debts

given lots of the tests involve liq, added this to a separate subissue under liquidation ticket #434

CheyenneAtapour:
Can consider refactoring `setNewPrice` function in `Base.t.sol`, reposting Miguel's comments (some I tackled already):

such a weird helper:
* expecting a percent instead of the final price
* unclear which direction the percent goes
* can we moved to SpokeBase no?

I also think it's fine to keep in `Base.t.sol` because Liquidity Hub tests might want to use it

CheyenneAtapour:
In `LiquidityHub.Draw.t.sol` tests, there are some tests which fuzz an `assetId` across all the assets on the hub, but only utilize those with a spoke that may not have all those assets listed as reserves, thus we are doing things like ```assetId = bound(assetId, 0, hub.assetCount() - 3); // Exclude duplicated DAI and usdy```.

We should clean this up. Miguel has the following suggestions:
1. do not fuzz assetId and used a fixed asset
2. create an array with proper assets, and use the fuzzed uint256 input param as index of the array
3. create a helper that gives the list of assets of the Hub that are connected to a given Spoke. It may need to distinguish between collateral and borrowable assets... Kind of tricky/ugly but I think we may end up needing smt like this

DhairyaSethi:
syntax sugar for our _reserveId(ISpoke) etc helpers such that we can do update(spoke1, _reserveId) instead of update(spoke1, _reserveId(spoke1))

yan-man:
[...]

PR #431, opened 2025-06-02
refactor hf tests

DhairyaSethi:
in an effort for #430 

- resolve precision issue todos on hf tests
- refactor to reduce vm.assume usage such that we can run the test suite on higher than 10k fuzz runs

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -17,7 +17,7 @@ runs = 1000
 seed = "0x640"
 
 [profile.pr.fuzz]
-runs = 50000
+runs = 20000
 
 [profile.ci.fuzz]
 runs = 10000


tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -401,6 +401,22 @@ contract SpokeBase is Base {
     return maxDebt > 1 ? maxDebt - 1 : maxDebt;
   }
 
+  function _boundCollAmt(
```
yan-man, line 404: it's good you did this but annoying bc it'd only work for single-collateral asset cases

DhairyaSethi: true yeah, we can add another method for multiple coll separately, i think an additive logic on this should work

Issue #414, opened 2025-05-27
Multiple Hubs Design


CanonicalJP:
Need to research 2 hubs <> 1 spoke implementation

- Miguel has an idea about listing an asset multiple times with different reserveids
    - have an address of the hub for every reserve
- Emilio in favor of tracking the address of the LH at the user position lvl
    - have a mapping of hubs in the reserve

The-3D:
Detailing my proposed implementation:

1. Introduce either a pair, or a mapping, of hub addresses instead of a single one in Spoke.sol https://github.com/aave/aave-v4/blob/874a1b179ba90a700373ec3123138938c9c539b1/src/contracts/Spoke.sol#L25
2. Receive as input parameter in Borrow() the address of the hub where the user would borrow from, and validate it against the values in 1. 
3. Introduce a "drawSource" variable in the UserPosition struct https://github.com/aave/aave-v4/blob/874a1b179ba90a700373ec3123138938c9c539b1/src/libraries/types/DataTypes.sol#L82C10-L82C22
4. store the source hub where liquidity is drawn on borrow and clear up on repay. In case of multiple partial borrow/repay, input Hub must match with the hub the user was already using

This avoids the following:

1. no multiple assets listing needed
2. very limited impact on the codebase 

to evaluate:

Impact on the AssetID (as we store the source assetId of every reserve into the Spoke data struct, what happens if the assetID of the destination hub is different than the one stored in the spokeconfig struct) Potential solution, don't store the assetID but fetch from the hub (more gas intensive)

CheyenneAtapour:
> Detailing my proposed implementation:

This solution does not allow for a user to simultaneously borrow the same asset from 2 hubs, correct? With the other solution of re-listing, we can achieve this

CheyenneAtapour:
@The-3D Prototyping your proposed implementation here: https://github.com/aave/aave-v4/tree/feat/2-hub

CheyenneAtapour:
We are proceeding with Miguel's implementation, in review here: https://github.com/aave/aave-v4/pull/432

CheyenneAtapour:
This issue was completed with this PR: https://github.com/aave/aave-v4/pull/432

PR #432, opened 2025-06-03
Multiple Hub Relist Implementation

CheyenneAtapour:
Miguel's approach on https://github.com/aave/aave-v4/issues/414


src/contracts/Spoke.sol
```diff
@@ -53,6 +55,11 @@ contract Spoke is ISpoke, Multicall {
   // Governance
   // /////
 
+  function addHub(address hub) external {
```
miguelmtzinf: we don't need this right? we weren't doing it before, so now nothing changes imo

src/contracts/Spoke.sol
```diff
@@ -67,7 +74,7 @@ contract Spoke is ISpoke, Multicall {
     DataTypes.ReserveConfig calldata config
   ) external returns (uint256) {
     _validateReserveConfig(config);
-    address asset = address(HUB.assetsList(assetId)); // will revert on invalid assetId
+    address asset = address(ILiquidityHub(_hubs[config.hubId]).assetsList(assetId)); // will revert on invalid assetId
```
miguelmtzinf: maybe we can use an internal helper to return the ILiquidityHub object directly so we don't need to cast everywhere.

snapshots/Spoke.Operations.json
```diff
@@ -1,17 +1,17 @@
...
+  "repay: partial": "311132",
+  "supply: 0 debt, collateralDisabled": "176158",
+  "supply: 0 debt, collateralEnabled": "176134",
+  "supply: 1 debt": "176182",
+  "supply: 2 debt": "176193",
+  "supply: 3 debt": "176146",
+  "updateUserRiskPremium": "192828",
+  "updateUserRiskPremium: 2 debts": "235186",
+  "usingAsCollateral": "50839",
+  "withdraw: full": "234636",
```
miguelmtzinf: the rest are cheaper, but this is the one getting more expensive, compared to https://github.com/aave/aave-v4/pull/433

DhairyaSethi: tbh i think we'll see the scales tip another way (this becomes more expensive) when we have a practical amount of reserves listed (current tests have 4, ideally we should run numbers on 100) bc calcUserAccountData is cheaper in #433

CheyenneAtapour: Keep in mind that the implementation was not complete in #433 and it was unclear how to proceed. The storage requirements change as well

src/contracts/Spoke.sol
CheyenneAtapour: I added this where it makes sense, but there are still a couple places where we cast from the config struct. I can make another helper to cover this as well, if you think it's better to have

src/contracts/Spoke.sol
CheyenneAtapour: Removed

src/libraries/types/DataTypes.sol
```diff
@@ -77,6 +78,7 @@ library DataTypes {
     uint256 liquidationBonus; // BPS, 100_00 represent a 0% bonus TODO: use smaller uint
     uint256 liquidityPremium; // BPS TODO: use smaller uint
     uint256 liquidationProtocolFee; // BPS TODO: use smaller uint
+    ILiquidityHub hub;
```
miguelmtzinf: we may want to put it at the beginning of the struct, but no big deal for now ig

src/contracts/Spoke.sol
```diff
@@ -150,14 +149,16 @@ contract Spoke is ISpoke, Multicall {
 
     // If uint256.max is passed, withdraw all user's supplied assets
     if (amount == type(uint256).max) {
-      amount = HUB.convertToSuppliedAssets(assetId, userPosition.suppliedShares);
+      amount = _getHub(reserve).convertToSuppliedAssets(assetId, userPosition.suppliedShares);
```
miguelmtzinf: the internal fns feel unnecessary now. I d delete an used the already available storage pointer when possible

tests/Base.t.sol
```diff
@@ -850,9 +878,10 @@ abstract contract Base is Test {
     return price.percentMul(percent);
   }
 
-  function setNewPrice(uint256 assetId, uint256 percent) public {
-    uint256 newPrice = calcNewPrice(oracle.getAssetPrice(assetId), percent);
-    oracle.setAssetPrice(assetId, newPrice);
+  function setNewPrice(IPriceOracle oracle, uint256 reserveId, uint256 percent) public {
```
miguelmtzinf: such a weird helper:
- expecting a percent instead of the final price
- unclear which direction the percent goes
- can we moved to SpokeBase no?
- we could use spoke as first input param (instead of oracle)

if you agree please add it as item to the tests refactor issue #430

tests/Base.t.sol
```diff
@@ -1024,13 +1073,13 @@ abstract contract Base is Test {
     uint256 reserveId,
     uint256 desiredHf
   ) internal view returns (uint256 requiredDebtAmount) {
+    IPriceOracle oracle = spoke.oracle();
```
miguelmtzinf: remove

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
```diff
@@ -39,6 +39,7 @@ abstract contract LiquidityHubScenarioBaseTest is Base {
   TestState internal state;
   DataTypes.SpokeConfig internal spokeConfig;
   Spoke internal spoke4; // init to be added during scenario tests
+  IPriceOracle internal oracle4;
```
miguelmtzinf, line 42: feels unnecessary

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
```diff
@@ -92,7 +92,9 @@ contract LiquidityHubDrawTest is LiquidityHubBase {
     vm.expectEmit(address(hub));
     emit ILiquidityHub.Draw(assetId, address(spoke1), drawAmount, drawAmount);
     vm.prank(address(spoke1));
+    console.log('right before draw');
     hub.draw({assetId: assetId, amount: drawAmount, to: alice});
+    console.log('after draw');
```
miguelmtzinf:
```suggestion
    hub.draw({assetId: assetId, amount: drawAmount, to: alice});
```

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
```diff
@@ -72,7 +72,7 @@ contract LiquidityHubDrawTest is LiquidityHubBase {
   }
 
   function test_draw_fuzz_amounts_same_block(uint256 assetId, uint256 daiAmount) public {
-    assetId = bound(assetId, 0, hub.assetCount() - 2); // Exclude duplicated DAI
+    assetId = bound(assetId, 0, hub.assetCount() - 3); // Exclude duplicated DAI and usdy
```
miguelmtzinf, line 74: why is this? cannot recall now
what's the problem of them being duplicated?

tests/unit/LiquidityHub/LiquidityHubBorrowIndex.t.sol
```diff
@@ -230,14 +230,15 @@ contract LiquidityHubBorrowIndex is Base {
...
+  function _deployAndAddSpoke(uint256 assetId) internal returns (address, address) {
+    IPriceOracle oracle = new MockPriceOracle();
+    Spoke spoke = new Spoke(address(oracle));
     hub.addSpoke(
       assetId,
       DataTypes.SpokeConfig({supplyCap: type(uint256).max, drawCap: type(uint256).max}),
       address(spoke)
     );
-    return address(spoke);
+    return (address(spoke), address(oracle));
```
miguelmtzinf: why we need to return the oracle now?

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -428,11 +434,17 @@ contract SpokeLiquidationBase is SpokeBase {
   /// @notice Will contain precision loss due to conversion split into two steps.
   /// @return Converted amount of toAsset.
   function _convertAssetAmount(
```
miguelmtzinf: wondering why we have this intermediate fn helper. We can use the internal one directly or just build a proper one and move it to base file

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -28,7 +23,8 @@ contract SpokeConfigTest is SpokeBase {
       liquidityPremium: reserveData.config.liquidityPremium + 1,
       liquidationProtocolFee: reserveData.config.liquidationProtocolFee + 1,
       borrowable: !reserveData.config.borrowable,
-      collateral: !reserveData.config.collateral
+      collateral: !reserveData.config.collateral,
+      hub: reserveData.config.hub // hub won't get updated
```
miguelmtzinf: technically doable no? i d remove this comment and create a ticket to evaluate if we wanna allow updating the hub or not

tests/unit/Spoke/Spoke.TwoHub.t.sol
miguelmtzinf, line 1: i d rename to MultipleHub

tests/unit/Spoke/Spoke.TwoHub.t.sol
```diff
@@ -0,0 +1,258 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+import {Spoke} from 'src/contracts/Spoke.sol';
+import {LiquidityHub} from 'src/contracts/LiquidityHub.sol';
```
miguelmtzinf: are these not part of the Base already?

tests/unit/Spoke/Spoke.TwoHub.t.sol
```diff
@@ -0,0 +1,258 @@
...
+  uint256 daiHub3ReserveId;
+  uint256 hub3DaiAssetId;
+  uint256 hub3UsdxAssetId;
+  uint256 hub3WbtcAssetId;
+  uint256 hub3WethAssetId;
+
+  function setUp() public virtual override {
+    super.setUp();
+
+    // Create a second hub
```
miguelmtzinf: i d move the following to Base as fixtures

miguelmtzinf: I mean, the deployment and config of hubs at least. The configuration of  a new reserve from another Hub would do it here

tests/unit/Spoke/Spoke.TwoHub.t.sol
```diff
@@ -0,0 +1,258 @@
...
+      liquidationProtocolFee: 0,
+      borrowable: true,
+      collateral: true,
+      hub: hub3
+    });
+    daiHub3ReserveId = spoke1.addReserve(hub3DaiAssetId, daiHub3Config);
+
+    vm.stopPrank();
+  }
+
```
miguelmtzinf: I d like to see a clear config of the setup. We can use a fixture as I said before, execute it in the setUp alongside the configuration, and leave good comments.
The unit tests below are good, but I want to to have an explicit test for the following usecases (at least the first one):
1. isolation mode / bootrapping liquidity of new hub-spoke:
- new hub and spoke with A collateral, B borrowable
- users cannot borrow B because have 0 liquidity
- list B asset again from the canonical Hub, with 0 supplyCap and X credit limit as borrowCap
- users cannot borrow B from new hub, same as before
- users can borrow B from canonical Hub, up until X
- users cannot supply B into canonical hub through new spoke (they can still do it through the canonical spoke)
- users supply B  into new hub
- DAO offboard credit line from canonical hub: borrowCap set to 0
- users repay B in canonical hub

2. siloed mode
- new hub and spoke with only B borrowable asset
- list A collateral from canonical Hub, X supply cap and 0 borrowCap
- users can borrow up to B borrow cap and A collateral supply cap

lmk wdyt

miguelmtzinf: we can do it in a separate PR, but I believe the first one is top priority

src/contracts/Spoke.sol
CheyenneAtapour: Ok, resolving in favor of removing the helper now as per more recent comment below

tests/Base.t.sol
CheyenneAtapour: I agree it's a bit strange. I think it's possible liquidity hub tests might want a helper function to change the price of a reserve, so we can consider keeping in `Base.t.sol`. I made it take `ISpoke` as first input. I also made a note in that issue to consider it when refactoring

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
CheyenneAtapour: The tests are commented out, but that spoke4 is deployed and used in the inheriting tests. We need a new oracle for spoke4 since oracles are spoke specific

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
CheyenneAtapour: spoke1 is not properly configured to draw Dai2, instead it is a reserve unique to spoke2

tests/unit/LiquidityHub/LiquidityHubBorrowIndex.t.sol
CheyenneAtapour: I thought it would be convenient because the new oracle is unique to the newly deployed spoke, and so whatever configurations are made to add reserves to the spoke must also be accounted for on the new oracle. 

Keep in mind that this is not really used atm because we haven't migrated liquidity hub tests, but if and when we use this helper, I think it'd be convenient for the above reason to be able to configure the new oracle

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
CheyenneAtapour: This is basically trying to achieve the same thing as `_calcEquivalentAssetAmount`. I think it's nice to have this as a dedicated helper function. Which approach do you like better / what kind of approach do you think is best to tackle what this is doing: Converting from one asset to another?

tests/unit/Spoke/Spoke.Config.t.sol
CheyenneAtapour: Yes, technically doable. Added a ticket: https://github.com/aave/aave-v4/issues/444

Personally, I do think it should be allowed

tests/unit/Spoke/Spoke.TwoHub.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.TwoHub.t.sol
CheyenneAtapour: Yeah, thanks for flagging, removed

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
CheyenneAtapour: Done

tests/Base.t.sol

src/contracts/Spoke.sol

src/libraries/types/DataTypes.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.TwoHub.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.TwoHub.t.sol
CheyenneAtapour: It's fine, I can add both on this PR.

Question: If users cannot supply B into the canonical hub, how does it get liquidity so that users can borrow B from it?

> * users can borrow B from canonical Hub, up until X
> * users cannot supply B into canonical hub


Regarding 2, If I understand the goals correctly, we could also just set it so collateral A from the canonical hub is not borrowable

tests/Base.t.sol
miguelmtzinf: Idk why the Hub would need to care about prices, since there is no collateral enforcement on that end. But all good, we can tackle later!

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
miguelmtzinf: something to fix/change. Can you add to the refactor task? #430 
Suggestions:
1. do not fuzz `assetId` and used a fixed asset
2. create an array with proper assets, and use the fuzzed uint256 input param as index of the array
3. create a helper that gives the list of assets of the Hub that are connected to a given Spoke. It may need to distinguish between collateral and borrowable assets... Kind of tricky/ugly but I think we may end up needing smt like this

tests/unit/LiquidityHub/LiquidityHubBorrowIndex.t.sol
miguelmtzinf: I d need to take a closer look to the tests, but it seems to me that we could abstract the oracle away for us. And if we need it, we can always access from the spoke's getter

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
miguelmtzinf: why do we have 2 functions that do the same? can we have only one an re-use?

tests/unit/Spoke/Spoke.Config.t.sol
miguelmtzinf: ty, added to #265

tests/unit/Spoke/Spoke.TwoHub.t.sol
miguelmtzinf: Re 1:
- I was meaning that users cannot supply B into canonical hub through the new spoke (because of zero supply cap). They can still supply B into canonical Hub through canonical Spoke, but the credit line from canonical hub to new spoke remains unchanged, defining the risk exposure of the DAO to that strategy. (I fixed the initial comment now).
- To clarify, the idea is that the credit limit to the new spoke is temporary, and ultimately, we want to stop relying on the canonical Hub. That’s why the goal is to have users start supplying B directly into the new Hub.
- btw, this example is building a new Hub for the isolated collateral. We could do everything in the same hub, relisting A collateral and B collateral, and creating a new spoke I think?

Re 2:
- Yes, but it's good to have controls on both ends. We could see it as Spoke being deployed by an external party and canonical hub belonging to the DAO. The DAO may want to partner up to a certain amount for this particular usecase.

miguelmtzinf: Would be nice if you can document a bit how both use-cases would work in terms of configuration and what are other available options.

tests/Base.t.sol
```diff
@@ -604,6 +643,169 @@ abstract contract Base is Test {
     vm.stopPrank();
   }
 
+  function deployAndConfigureAdditionalHubs() internal {
```
miguelmtzinf: I d create multiple functions. We can add `fixture` to the name so it's clearer and docs to each function to provide info about the setup.
I d only do the deployment of the Hub with a set of assets in the fixture, and leave the config between hubs and spokes out of it.

tests/unit/Spoke/Spoke.TwoHub.t.sol
miguelmtzinf: Siloed borrowing is the usecase where a user can only borrow an asset.
In the example:

- new spoke for this specific usecase
- B must be the only borrowable asset in this new spoke
- B can be listed again in the canonical hub, or use a new hub to list B
- To bootstrap liquidity, we list B from canonical Hub with 0 supplyCap and X borrowCap
- We are a bit agnostic about collateral assets I think. The example above is saying we list A from canonical hub and make it Y supplyCap and 0 borrowCap.

src/contracts/Spoke.sol
```diff
@@ -181,7 +182,7 @@ contract Spoke is ISpoke, Multicall {
     userPremiumDrawnShares = userPosition.premiumDrawnShares = userPosition
       .baseDrawnShares
       .percentMulUp(newUserRiskPremium);
-    userPremiumOffset = userPosition.premiumOffset = HUB.previewOffset(
+    userPremiumOffset = userPosition.premiumOffset = reserve.config.hub.previewOffset(
```
DhairyaSethi: cache `reserve.config.hub` for this, borrow and restore cases pls

src/contracts/Spoke.sol
```diff
@@ -261,13 +264,17 @@ contract Spoke is ISpoke, Multicall {
     DataTypes.Reserve storage reserve = _reserves[reserveId];
     uint256 assetId = reserve.assetId;
 
-    (uint256 baseDebt, uint256 premiumDebt) = _getUserDebt(userPosition, assetId);
+    _validateRepay(reserve);
+    (uint256 baseDebt, uint256 premiumDebt) = _getUserDebt(
+      userPosition,
+      assetId,
+      reserve.config.hub
```
DhairyaSethi: use cached hub. also param order hub, assetId, userPos makes more to me but nbd

tests/Base.t.sol
CheyenneAtapour: kk sounds good!

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
CheyenneAtapour: Added: https://github.com/aave/aave-v4/issues/430#issuecomment-2974685650

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
CheyenneAtapour: We don't need both, I was trying to ask which one you think does it better. I'm opting for keeping this one over `_calcEquivalentAssetAmount`

tests/unit/Spoke/Spoke.TwoHub.t.sol
CheyenneAtapour: Cool, we are on the same page about 1 and 2. RE 1: It feels clearer to me to keep it with the new hub, might make the test more confusing to follow if it's the same hub

tests/unit/LiquidityHub/LiquidityHubBorrowIndex.t.sol
CheyenneAtapour: Yes, that's a good point about the new spoke's getter for oracle. I reverted this change

tests/Base.t.sol

src/contracts/Spoke.sol

src/contracts/Spoke.sol

tests/Base.t.sol
```diff
@@ -70,13 +70,20 @@ abstract contract Base is Test {
   IERC20 internal eth;
   IERC20 internal wbtc;
 
-  MockPriceOracle internal oracle;
+  MockPriceOracle internal oracle1;
+  MockPriceOracle internal oracle2;
+  MockPriceOracle internal oracle3;
   ILiquidityHub internal hub;
+  ILiquidityHub internal hub2;
```
miguelmtzinf: do we need this as global? I was thinking the fixture can return the hub itself. Same for `hub3`

tests/unit/Spoke/Spoke.MultipleHub.Scenario.Base.t.sol
```diff
@@ -0,0 +1,307 @@
...
+
+import {LiquidityHub, ILiquidityHub} from 'src/contracts/LiquidityHub.sol';
+import {Spoke, ISpoke} from 'src/contracts/Spoke.sol';
+import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+import {DefaultReserveInterestRateStrategy, IDefaultInterestRateStrategy} from 'src/contracts/DefaultReserveInterestRateStrategy.sol';
+
+import {TestnetERC20} from 'tests/mocks/TestnetERC20.sol';
+import {MockPriceOracle, IPriceOracle} from 'tests/mocks/MockPriceOracle.sol';
+
+contract SpokeMultipleHubScenarioBase is Test {
```
miguelmtzinf:
```suggestion
contract SpokeMultipleHubBase is Test {
```
i think better? filename too

just because it's hard to distinguish from the other test file.

tests/unit/Spoke/Spoke.MultipleHub.t.sol
```diff
@@ -0,0 +1,146 @@
...
+    // Check Bob's total debt on each hub
+    assertEq(spoke1.getUserTotalDebt(_daiReserveId(spoke1), bob), hub1DaiBorrowAmount);
+    assertEq(spoke1.getUserTotalDebt(daiHub2ReserveId, bob), hub2DaiBorrowAmount);
+
+    assertEq(hub.getAssetTotalDebt(daiAssetId), hub1DaiBorrowAmount);
+    assertEq(hub2.getAssetTotalDebt(daiAssetId), hub2DaiBorrowAmount);
+  }
+
+  /// @dev Test showcasing dai may be borrowed from hub 3 via spoke 1, with collateral supplied to hub 1
+  function test_borrow_thirdHub() public {
```
miguelmtzinf, line 139: is not essentially the same scenario as the one above? plus a repay + withdraw at the end

tests/unit/Spoke/Spoke.MultipleHub.Scenario.Base.t.sol
```diff
@@ -0,0 +1,307 @@
...
+    // Configure interest rate strategy for assets A and B
+    newIrStrategy.setInterestRateParams(isolationVars.assetAId, irData);
+    newIrStrategy.setInterestRateParams(isolationVars.assetBId, irData);
+  }
+
+  /* @dev Adds asset B to the new hub and new spoke with 100k draw cap.
+   * Adds Asset A to the canonical hub and canonical spoke with no restrictions.
+   * Relists Asset A from the canonical hub on the new spoke, with supply cap 500k, 0 borrow cap.
+   * SUMMARY:
+   * New Spoke: AssetA, canonical hub supplyable up to 500k; Asset B, new hub borrowable up to 100k.
```
miguelmtzinf: let's discuss today if asset B would need to be completely isolated from the canonical hub (as it is now), it could be just using B from canonical hub up to certain draw cap

tests/unit/Spoke/Spoke.MultipleHub.Scenario.Base.t.sol
miguelmtzinf: feeling that i d be better to have 1 file for isolation mode, another file for siloed...
@DhairyaSethi you were suggesting this setup, what are your thoughts?

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
```diff
@@ -0,0 +1,248 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/Spoke.MultipleHub.Scenario.Base.t.sol';
+
+contract SpokeMultipleHubScenarioTest is SpokeMultipleHubScenarioBase {
+  /* @dev Test showcasing a possible configuration for isolation mode
+   * A new hub and spoke are deployed with new assets A and B.
+   * There is no liquidity for asset B on the new hub, so instead
+   * Asset B is listed on the canonical hub and linked to the new spoke with a draw cap.
```
miguelmtzinf:
```suggestion
   * Asset B is listed from the canonical hub and linked to the new spoke with a draw cap.
```
?

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
```diff
@@ -0,0 +1,248 @@
...
+   */
+  function test_isolation_mode() public {
+    setUpIsolationMode();
+
+    // Bob can supply asset A to the new spoke and set it as collateral
+    vm.startPrank(bob);
+    assetA.approve(address(newHub), type(uint256).max);
+    deal(address(assetA), bob, MAX_SUPPLY_AMOUNT);
+    newSpoke.supply(isolationVars.reserveAId, MAX_SUPPLY_AMOUNT);
+    newSpoke.setUsingAsCollateral(isolationVars.reserveAId, true);
```
miguelmtzinf: due to new setUp, we cannot use `Utils` (cc @DhairyaSethi )

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
```diff
@@ -0,0 +1,248 @@
...
+
+    // Configure interest rate strategy for asset B on the main hub
+    irStrategy.setInterestRateParams(isolationVars.assetBIdMainHub, irData);
+
+    // Bob still cannot borrow asset B from the new hub because there is no liquidity
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.NotAvailableLiquidity.selector, 0));
+    newSpoke.borrow(isolationVars.reserveBId, 100e18, bob);
+    vm.stopPrank();
+
+    // List reserve B on spoke 1 for the main hub, allowing supplying and borrowing
```
miguelmtzinf: do not understand this. why this is needed?

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
```diff
@@ -0,0 +1,248 @@
...
+    oracle1.setReservePrice(isolationVars.spoke1ReserveBId, 50_000e8);
+
+    // Link main hub and spoke 1 for asset B
+    hub.addSpoke(
+      isolationVars.assetBIdMainHub,
+      DataTypes.SpokeConfig({drawCap: type(uint256).max, supplyCap: type(uint256).max}),
+      address(spoke1)
+    );
+
+    // Alice can supply asset B to the main hub via spoke 1
```
miguelmtzinf: this is regular business right? is the config above what allows this?

If so, i was confused because I thought this was previously configured. (and assumed)

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
```diff
@@ -0,0 +1,248 @@
...
+    // Bob cannot borrow asset B from main hub via new spoke past draw cap
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.DrawCapExceeded.selector, 100_000e18));
+    newSpoke.borrow(isolationVars.reserveBIdMainHub, 1e18, bob);
+
+    // Bob cannot supply B to main hub via new spoke because supply cap is 0
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.SupplyCapExceeded.selector, 0));
+    newSpoke.supply(isolationVars.reserveBIdMainHub, 1e18);
+    vm.stopPrank();
+
+    // DAO offboards credit line to new spoke from the canonical hub by setting Asset B draw cap to 0
```
miguelmtzinf: before this, the DAO would wait until liquidity is deployed in the new spoke, new hub. Can add these steps?
- suppliers come in
- bob migrates, so repays on canonical hub and borrows from new hub
- dao offboards

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
```diff
@@ -0,0 +1,248 @@
...
+  /* @dev Test showcasing a possible configuration for siloed mode
+   * A new hub and spoke are deployed with only Asset B as borrowable.
+   * Users can use usdx as collateral on the new spoke, which supplies to the canonical hub.
+   * Users may not borrow usdx from the new spoke, but can use it as collateral to borrow
+   * the only available asset: Asset B.
+   */
+  function test_siloed_mode() public {
+    setUpSiloedMode();
+
+    // Bob can supply Asset A to the new spoke, canonical hub, up to 500k and set it as collateral
```
miguelmtzinf: based on docs above, only B exists on new spoke

tests/unit/Spoke/Spoke.MultipleHub.Scenario.Base.t.sol
CheyenneAtapour: The reason I opted for this is that it's only inherited by the scenario test, not `Spoke.MultipleHub.t.sol`, so I didn't want that to be confusing, but yeah I can change to this name

tests/unit/Spoke/Spoke.MultipleHub.t.sol
CheyenneAtapour: The first test also borrows from both hubs, but yeah these are just simple tests

tests/unit/Spoke/Spoke.MultipleHub.Scenario.Base.t.sol
CheyenneAtapour: It can indeed just be using B from canonical hub. I had thought it looked clearer to just use the new hub, but now we've separated the setup, I think can go either way

tests/unit/Spoke/Spoke.MultipleHub.Scenario.Base.t.sol
DhairyaSethi: yes my suggestion is to go for separate files for each case, both can inherit from Base (instead of Test such that usual helpers are available) and override setUp

tests/Base.t.sol

tests/unit/Spoke/Spoke.MultipleHub.Scenario.Base.t.sol
CheyenneAtapour: Moved to separate files

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
CheyenneAtapour: Now using `Utils` with new setup as proposed by @DhairyaSethi

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
CheyenneAtapour: Moved to setup to reduce confusion

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
CheyenneAtapour: It's regular business, yes moved to setup to reduce confusion

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.MultipleHub.Scenario.t.sol
CheyenneAtapour: Fixed

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -427,17 +433,6 @@ contract SpokeLiquidationBase is SpokeBase {
...
-  /// @notice Convert 1 asset amount to equivalent amount in another asset.
-  /// @notice Will contain precision loss due to conversion split into two steps.
-  /// @return Converted amount of toAsset.
-  function _convertAssetAmount(
-    uint256 assetId,
-    uint256 amount,
-    uint256 toAssetId
-  ) internal view returns (uint256) {
-    return _convertBaseCurrencyToAmount(toAssetId, _convertAmountToBaseCurrency(assetId, amount));
-  }
```
yan-man, line 439: was this function unused? why rm

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
```diff
@@ -0,0 +1,297 @@
...
+
+contract SpokeMultipleHubIsolationModeTest is SpokeMultipleHubBase {
+  struct IsolationLocalVars {
+    uint256 assetAId;
+    uint256 assetBId;
+    uint256 reserveAId;
+    uint256 reserveBId;
+    uint256 assetBIdMainHub;
+    uint256 reserveBIdMainHub;
+    uint256 spoke1ReserveBId;
```
yan-man, line 14: i think maybe you can have substructs to be more extensible or cleaner

```
Asset {
uint assetId;
uint256 reserveId;
uint256 assetIdMainHub;
...
}
```

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
```diff
@@ -0,0 +1,297 @@
...
+    // Configure interest rate strategy for asset B on the main hub
+    irStrategy.setInterestRateParams(isolationVars.assetBIdMainHub, irData);
+  }
+
+  /* @dev Test showcasing a possible configuration for isolation mode
+   * A new hub and spoke are deployed with new assets A and B.
+   * There is no liquidity for asset B on the new hub, so instead
+   * Asset B is listed from the canonical hub and linked to the new spoke with a draw cap.
+   * Thus users can borrow asset B from the canonical hub via the new spoke,
+   * without being able to supply it from the new spoke.
```
yan-man, line 176: why do we need new hub at all? can't we achieve this just through re-listing assetA and assetB on canonical hub?

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
CheyenneAtapour: Nah, just moved to base. We had 2 functions that did the same thing. Opted to keep this one and reuse everywhere else we were using the other one

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
```diff
@@ -0,0 +1,297 @@
...
+   */
+  function test_isolation_mode() public {
+    // Bob can supply asset A to the new spoke and set it as collateral
+    vm.prank(bob);
+    assetA.approve(address(newHub), type(uint256).max);
```
yan-man: i think can move all approvals and deals for each user to each asset/hub into setup, nbd

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
CheyenneAtapour: The idea is the new hub and spoke are isolated from each other, after bootstrapping the liquidity. See the diagram I made on craft, which I think clears it up

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
```diff
@@ -0,0 +1,297 @@
...
+    );
+    assertEq(
+      newHub.getAssetSuppliedAmount(isolationVars.assetAId),
+      MAX_SUPPLY_AMOUNT,
+      'total supplied amount of assetA on new hub'
+    );
+
+    // Bob cannot borrow asset B because there is no liquidity
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.NotAvailableLiquidity.selector, 0));
+    Utils.borrow(newSpoke, isolationVars.reserveBId, bob, 100e18, bob);
```
yan-man:
```suggestion
    Utils.borrow(newSpoke, isolationVars.reserveBId, bob, 1, bob);
```
?

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
```diff
@@ -0,0 +1,297 @@
...
+    // Bob CAN borrow asset B from the main hub via new spoke up until the draw cap of 100k
+    Utils.borrow(newSpoke, isolationVars.reserveBIdMainHub, bob, 100_000e18, bob);
+
+    // Check Bob's total debt of asset B on the new spoke
+    assertEq(newSpoke.getUserTotalDebt(isolationVars.reserveBIdMainHub, bob), 100_000e18);
+    assertEq(hub.getAssetTotalDebt(isolationVars.assetBIdMainHub), 100_000e18);
+
+    // Bob cannot borrow asset B from main hub via new spoke past draw cap
+    vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.DrawCapExceeded.selector, 100_000e18));
+    Utils.borrow(newSpoke, isolationVars.reserveBIdMainHub, bob, 1e18, bob);
```
yan-man:
```suggestion
    Utils.borrow(newSpoke, isolationVars.reserveBIdMainHub, bob, 1, bob);
```

tests/unit/Spoke/Spoke.MultipleHub.SiloedBorrowing.t.sol
```diff
@@ -0,0 +1,223 @@
...
+      dynReserveConfig
+    );
+
+    // Set the price of B reserve for the new oracle
+    newOracle.setReservePrice(siloedVars.reserveBId, 2000e8);
+
+    // Link new hub and new spoke for asset B, 100k draw cap
+    newHub.addSpoke(
+      siloedVars.assetBId,
+      DataTypes.SpokeConfig({drawCap: siloedVars.assetBDrawCap, supplyCap: type(uint256).max}),
```
yan-man:
```suggestion
      DataTypes.SpokeConfig({drawCap: siloedVars.assetBDrawCap, supplyCap: UINT256_MAX}),
```

generally we can always use `UINT256_MAX` in tests which is defined as a constant in forge `CommonBase` but nbd

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
yan-man, line 1: does this iso mode support exactly as v3? https://aave.com/help/supplying/isolation-mode

> if you already have other assets enabled as collateral, the Isolated Asset can still be supplied, but it won’t be eligible as collateral. This means you can earn interest on it, but you can’t use it to borrow unless it’s your only collateralised asset.

is this scenario possible

CheyenneAtapour: @miguelmtzinf wdyt? Is this necessary to enforce? Or was this a limitation on V3?

Btw, it's set currently that we cannot supply the asset back to the hub that liquidity is being bootstrapped from

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
CheyenneAtapour: I feel like this might be over-engineering because it's just a one-time use struct specific for this test. But this is a great idea for later on when we add more exhaustive tests

tests/unit/Spoke/Spoke.MultipleHub.SiloedBorrowing.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
CheyenneAtapour: Done, also moved dealings of tokens

tests/Base.t.sol
```diff
@@ -102,6 +105,11 @@ abstract contract Base is Test {
   uint256 internal usdyAssetId = 4;
   uint256 internal dai2AssetId = 5;
 
+  uint256 internal hub3DaiAssetId;
+  uint256 internal hub3UsdxAssetId;
+  uint256 internal hub3WbtcAssetId;
+  uint256 internal hub3WethAssetId;
```
miguelmtzinf: if `hub2` and `hub3` are not global, these neither

tests/Base.t.sol
```diff
@@ -70,13 +70,16 @@ abstract contract Base is Test {
...
+  MockPriceOracle internal oracle1;
+  MockPriceOracle internal oracle2;
+  MockPriceOracle internal oracle3;
   ILiquidityHub internal hub;
   ITreasurySpoke internal treasurySpoke;
   ISpoke internal spoke1;
   ISpoke internal spoke2;
   ISpoke internal spoke3;
   DefaultReserveInterestRateStrategy internal irStrategy;
+  DefaultReserveInterestRateStrategy internal creditLineIRStrategy;
```
miguelmtzinf: rm

tests/Base.t.sol
```diff
@@ -480,11 +501,15 @@ abstract contract Base is Test {
...
+    oracle2.setReservePrice(spokeInfo[spoke2].wbtc.reserveId, 50_000e8);
+    oracle2.setReservePrice(spokeInfo[spoke2].weth.reserveId, 2000e8);
+    oracle2.setReservePrice(spokeInfo[spoke2].dai.reserveId, 1e8);
+    oracle2.setReservePrice(spokeInfo[spoke2].usdx.reserveId, 1e8);
+
     hub.addSpoke(wbtcAssetId, spokeConfig, address(spoke2));
     hub.addSpoke(wethAssetId, spokeConfig, address(spoke2));
     hub.addSpoke(daiAssetId, spokeConfig, address(spoke2));
     hub.addSpoke(usdxAssetId, spokeConfig, address(spoke2));
-    hub.addSpoke(usdyAssetId, spokeConfig, address(spoke2));
```
miguelmtzinf: why this was removed?

tests/unit/Spoke/Spoke.MultipleHub.t.sol
miguelmtzinf: not v useful... let's do:
1. test_borrow_secondHub() -> collateral and debt in hub1, we borrow from second hub, we repay a bit from hub1 debt and hub2 debt
2. test_collateral_thirdHub() -> collateral and debt in hub1, we add collateral to third hub, we withdraw from hub1

tests/unit/Spoke/Spoke.MultipleHub.t.sol
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /// @dev Test showcasing dai may be borrowed from hub 2 and hub 1 via spoke 1
+  function test_borrow_secondHub() public {
+    uint256 hub1DaiBorrowAmount = 5e18;
+    uint256 hub2DaiBorrowAmount = 1e18;
+
+    // Bob supply to spoke 1 on hub 1
+    vm.startPrank(bob);
+    spoke1.supply(_daiReserveId(spoke1), 100000e18);
```
miguelmtzinf: let's use utils now no?

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
miguelmtzinf: this works in v4 too, two options:
- user can still supply that asset via canonical spoke to earn yield
- if the asset is listed as reserve in the new spoke as suppliable and borrowable but zero collateral factor. This is a bit messy/counter-intuitive, because in order to earn yield on that asset other users need to borrow and it's not the purpose of this new spoke.

From product design pov, better to do the first option imo. @CheyenneAtapour can you add to the doc pls? Also, it would be great to add a deposit to the canonical hub via canonical spoke to earn yield (first option), to showcase this fact.
hope it's clear

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
miguelmtzinf: doable @yan-man , just going with new hub/spoke approach. This is an example of isolation mode + bootstrapping liquidity actually.

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
yan-man: > you can’t use it to borrow unless it’s your only collateralised asset

dont think this part is possible but that's fine right

tests/unit/Spoke/Spoke.MultipleHub.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
miguelmtzinf: I think it's possible by just changing the config

tests/Base.t.sol
CheyenneAtapour: I saw it was never actually properly configured, but I'm configuring it now

PR #433, opened 2025-06-03
2 Hubs 1 Spoke

CheyenneAtapour:
Emilio's approach on https://github.com/aave/aave-v4/issues/414

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -71,6 +71,7 @@ contract Spoke is ISpoke, Multicall {
 
   function addReserve(
     uint256 assetId,
+    uint256 hubId,
     DataTypes.ReserveConfig calldata config
   ) external returns (uint256) {
     _validateReserveConfig(config);


src/contracts/Spoke.sol
```diff
@@ -198,23 +218,39 @@ contract Spoke is ISpoke, Multicall {
     );
     _notifyRiskPremiumUpdate(assetId, msg.sender, newUserRiskPremium);
 
+    if (
+      userPosition.baseDrawnShares == 0 &&
+      userPosition.realizedPremium == 0 &&
+      userPosition.suppliedShares == 0
```
DhairyaSethi, line 225: can also do sum == 0 since all are uint's

src/contracts/Spoke.sol
```diff
@@ -198,23 +218,39 @@ contract Spoke is ISpoke, Multicall {
...
+    ) {
+      // Reset supplied hub
```
DhairyaSethi, line 227: don't necessarily have to tho, no? i'm not sure if the zero checks below are necessary but nbd

CheyenneAtapour: This is to allow user to in the future use a different hub for the asset

CheyenneAtapour:
Closing this in favor of #432

PR #443, opened 2025-06-12
fix: Interest Rate Strategy

### LINKED ISSUE #95: Fix Default Reserve Rate Strategy
the contract currently returns `baseRate` in bps on `calculateInterestRate` whereas it should be in ray (according to natspec as well) 

the rate is expected to be ray in the math interest calc as well. pick updated strategy from v3-origin

avniculae:
Closes #95

### DIFF
--- a/src/libraries/math/WadRayMathExtended.sol
+++ b/src/libraries/math/WadRayMathExtended.sol
@@ -189,6 +189,11 @@ library WadRayMathExtended {
     }
   }
 
+  /**
+   * @notice Converts value from basis points to ray
+   * @param a The value in basis points
+   * @return The value in ray (= a * 1e27 / 1e4)
+   */
   function bpsToRay(uint256 a) internal pure returns (uint256) {
     return (a * RAY) / PERCENTAGE_FACTOR;
   }
--- a/tests/unit/AssetInterestRateStrategy.t.sol
+++ b/tests/unit/AssetInterestRateStrategy.t.sol
@@ -161,8 +161,8 @@ contract AssetInterestRateStrategyTest is Base {
     uint256 liquidityAdded,
     uint256 liquidityTaken
   ) public {
-    totalDebt = bound(totalDebt, 1, type(uint64).max);
     availableLiquidity = bound(availableLiquidity, 0, type(uint64).max);
+    totalDebt = bound(totalDebt, 1, type(uint64).max);
     liquidityAdded = bound(liquidityAdded, 0, type(uint64).max);
     liquidityTaken = bound(
       liquidityTaken,
@@ -182,8 +182,8 @@ contract AssetInterestRateStrategyTest is Base {
 
   function test_calculateInterestRate_revertsWith_ArithmeticUnderflow() public {
     test_calculateInterestRate_fuzz_revertsWith_ArithmeticUnderflow({
-      totalDebt: 100e6,
       availableLiquidity: 100e6,
+      totalDebt: 100e6,
       liquidityAdded: 10e6,
       liquidityTaken: 120e6
     });


src/contracts/DefaultReserveInterestRateStrategy.sol
```diff
@@ -24,50 +23,55 @@ contract DefaultReserveInterestRateStrategy is IDefaultInterestRateStrategy {
...
+  uint16 public constant MIN_OPTIMAL_POINT = 1_00; // 1.00% in BPS
 
   /// @inheritdoc IDefaultInterestRateStrategy
-  uint256 public constant MAX_OPTIMAL_POINT = 99_00; // 99.00% in BPS
+  uint16 public constant MAX_OPTIMAL_POINT = 99_00; // 99.00% in BPS
 
   /// @dev Map of assetId and their interest rate data (reserveAddress => interestRateData)
   mapping(uint256 => InterestRateData) internal _interestRateData;
 
+  error INVALID_ADDRESSES_PROVIDER();
```
DhairyaSethi: move errors to interface pls

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
avniculae, line 1: @miguelmtzinf FYI this is the test file where I had to change the test to accept 1 wei differences

src/contracts/DefaultReserveInterestRateStrategy.sol
avniculae: this has been addressed in the meantime, please re-review

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,145 @@
...
+ *   due to the caching of the PoolAddressesProvider and the usage of underlying addresses as
+ *   index of the _interestRateData
+ */
+contract DefaultAssetInterestRateStrategy is IDefaultInterestRateStrategy {
+  using WadRayMath for uint16;
+  using WadRayMath for uint32;
+  using WadRayMath for uint256;
+
+  /// @inheritdoc IDefaultInterestRateStrategy
+  address public immutable ADDRESSES_PROVIDER;
```
miguelmtzinf: we can rm completely, or point to the HUB instead (I guess it d be needed for the permissioning part)

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,145 @@
...
+    );
+  }
+
+  /// @inheritdoc IDefaultInterestRateStrategy
+  function getInterestRateData(uint256 assetId) external view returns (InterestRateData memory) {
+    return _interestRateData[assetId];
+  }
+
+  /// @inheritdoc IDefaultInterestRateStrategy
+  function getOptimalUsageRatio(uint256 assetId) external view returns (uint16) {
```
miguelmtzinf: i think it's probably better to return all of them as `uint256`

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,145 @@
...
+  ) external view virtual override returns (uint256) {
+    InterestRateData memory rateData = _interestRateData[assetId];
+    require(rateData.optimalUsageRatio != 0, INTEREST_RATE_DATA_NOT_SET(assetId));
+
+    uint256 currentVariableBorrowRateRay = rateData.baseVariableBorrowRate.bpsToRay();
+    if (totalDebt == 0) {
+      return currentVariableBorrowRateRay;
+    }
+
+    uint256 borrowUsageRatioRay = totalDebt.rayDiv(availableLiquidity + totalDebt);
```
miguelmtzinf: we can rename to `usageRatio` or `usageRatioRay` since there is only 1 usage (there were supply and borrow earlier)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: should we append `Bps` to function names in this case?

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: yes, it will be needed for the permissioning part, not sure which contract will push params though

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,145 @@
...
+import {IDefaultInterestRateStrategy} from 'src/interfaces/IDefaultInterestRateStrategy.sol';
+import {IAssetInterestRateStrategy} from 'src/interfaces/IAssetInterestRateStrategy.sol';
+
+/**
+ * @title DefaultAssetInterestRateStrategy contract
+ * @author Aave Labs
+ * @notice Default interest rate strategy used by the Aave protocol
+ * @dev Strategies are pool-specific: each contract CAN'T be used across different Aave pools
+ *   due to the caching of the PoolAddressesProvider and the usage of underlying addresses as
+ *   index of the _interestRateData
```
DhairyaSethi: pls fix this natspec. also good to mention we do not consider reserve factor here anymore

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: [5c57ae4](https://github.com/aave/aave-v4/pull/443/commits/5c57ae42c93416306cfd45718a46897598b9f170)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: [5c57ae4](https://github.com/aave/aave-v4/pull/443/commits/5c57ae42c93416306cfd45718a46897598b9f170)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: [5c57ae4](https://github.com/aave/aave-v4/pull/443/commits/5c57ae42c93416306cfd45718a46897598b9f170)

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,140 @@
...
+ */
+contract DefaultAssetInterestRateStrategy is IDefaultInterestRateStrategy {
+  using WadRayMath for uint16;
+  using WadRayMath for uint32;
+  using WadRayMath for uint256;
```
DhairyaSethi:
```suggestion
  using WadRayMath for *;
```

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,140 @@
...
+  uint256 public constant MAX_BORROW_RATE = 1000_00; // 1000.00% in BPS
+
+  /// @inheritdoc IDefaultInterestRateStrategy
+  uint256 public constant MIN_OPTIMAL_RATIO = 1_00; // 1.00% in BPS
+
+  /// @inheritdoc IDefaultInterestRateStrategy
+  uint256 public constant MAX_OPTIMAL_RATIO = 99_00; // 99.00% in BPS
+
+  /// @dev Map of assetId and their interest rate data (assetId => interestRateData)
+  mapping(uint256 => InterestRateData) internal _interestRateData;
```
DhairyaSethi:
```suggestion
  mapping(uint256 assetId => InterestRateData data) internal _interestRateData;
```

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,140 @@
...
+    if (usageRatioRay <= optimalUsageRatioRay) {
+      currentVariableBorrowRateRay += rateData
+        .variableRateSlope1
+        .bpsToRay()
+        .rayMul(usageRatioRay)
+        .rayDiv(optimalUsageRatioRay);
+    } else {
+      currentVariableBorrowRateRay += rateData.variableRateSlope1.bpsToRay();
+
+      currentVariableBorrowRateRay += rateData
```
DhairyaSethi: slightly cheaper to combine these twp but optimizer most likely does it, nbd

src/libraries/logic/AssetLogic.sol
```diff
@@ -111,17 +111,11 @@ library AssetLogic {
...
-        assetId: asset.id,
-        virtualUnderlyingBalance: asset.availableLiquidity, // without current liquidity change
-        usingVirtualBalance: true
-      })
-    );
+    asset.baseBorrowRate = asset.config.irStrategy.calculateInterestRate({
+      assetId: asset.id,
+      totalDebt: asset.baseDebt(),
+      availableLiquidity: asset.availableLiquidity + liquidityAdded - liquidityTaken
+    });
```
DhairyaSethi: wondering if it may be safer to pass `liquidityAdded` and `liquidityTaken` independently to allow ir strat to remain extensible in case any strat in the future may require the two, tho we haven't ever found the need so far afaik, wdyt @miguelmtzinf @avniculae?

tests/Base.t.sol
```diff
@@ -45,7 +45,7 @@ abstract contract Base is Test {
   uint256 internal MAX_SUPPLY_AMOUNT_USDY;
   uint256 internal constant MAX_SUPPLY_IN_BASE_CURRENCY = 1e39;
   uint32 internal constant MAX_RISK_PREMIUM_BPS = 1000_00;
-  uint256 internal constant MAX_BORROW_RATE = 1000_00; // matches DefaultReserveInterestRateStrategy
+  uint256 internal constant MAX_BORROW_RATE = 1000_00; // matches DefaultAssetInterestRateStrategy
```
DhairyaSethi: yeah wondering if we should add a quick `_validateConstants` in the base setUp for all of these but most likely overkill

src/interfaces/IDefaultInterestRateStrategy.sol
```diff
@@ -56,92 +45,97 @@ interface IDefaultInterestRateStrategy is IReserveInterestRateStrategy {
...
 
   /**
-   * @notice Sets interest rate data for an Aave rate strategy
-   * @param assetId The assetId to update
-   * @param rateData The reserve interest rate data to apply to the given reserve
-   *   Being specific to this custom implementation, with custom struct type,
-   *   overloading the function on the generic interface
+   * @notice Thrown whe the addresses provided is address zero
+   */
+  error INVALID_ADDRESSES_PROVIDER();
```
DhairyaSethi: rm

src/interfaces/IDefaultInterestRateStrategy.sol
```diff
@@ -56,92 +45,97 @@ interface IDefaultInterestRateStrategy is IReserveInterestRateStrategy {
...
+
+  /**
+   * @notice Thrown when the max possible rate is greater than `MAX_BORROW_RATE`
+   */
+  error INVALID_MAX_RATE();
```
DhairyaSethi:
```suggestion
  error InvalidMaxRate();
```
same for all below

tests/unit/DefaultAssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,240 @@
...
+    test_calculateInterestRate_LeftToKinkPoint(100_00);
+  }
+
+  function test_calculateInterestRate_RightToKinkPoint(uint256 percentageFromKinkPointBps) public {
+    uint256 percentageFromKinkPointRay = bound(percentageFromKinkPointBps, 1, 100_00).bpsToRay();
+
+    (uint256 totalDebt, uint256 availableLiquidity) = _computeDebtAndAvailableLiquidity(
+      rateData.optimalUsageRatio.bpsToRay() +
+        percentageFromKinkPointRay.rayMul(WadRayMath.RAY - rateData.optimalUsageRatio.bpsToRay())
+    );
```
DhairyaSethi: can also just bound the input from kink pt to 100, more readable

tests/unit/DefaultAssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,240 @@
...
+
+  function test_calculateInterestRate_AtMaxUtilization() public {
+    test_calculateInterestRate_RightToKinkPoint(100_00);
+  }
+
+  function _computeDebtAndAvailableLiquidity(
+    uint256 targetUtilizationRatioRay
+  ) internal pure returns (uint256 totalDebt, uint256 availableLiquidity) {
+    /// @dev using 27 decimals to avoid precision loss, even though assets in the hub can have up to 18 decimals
+    totalDebt = 100e27;
```
DhairyaSethi: is there significant loss if we fix this to MAX_SUPPLY_AMOUNT or something?

tests/unit/DefaultAssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,240 @@
...
+      assetId: mockAssetId,
+      totalDebt: totalDebt,
+      availableLiquidity: availableLiquidity
+    });
+
+    uint256 expectedVariableRate = rateData.baseVariableBorrowRate.bpsToRay() +
+      rateData.variableRateSlope1.bpsToRay() +
+      rateData.variableRateSlope2.bpsToRay().rayMul(percentageFromKinkPointRay);
+    assertEq(variableBorrowRate, expectedVariableRate);
+  }
```
DhairyaSethi: very nice! left & right to kink point cover all cases but i don't like is that totalDebt is fixed on all of these, we can do another set of these by where we fix available liq but i don't see that being very useful tbh, let's note in your improvement issue and decide if we'll ever want to do it later

tests/unit/DefaultAssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,240 @@
...
+
+import {Test} from 'forge-std/Test.sol';
+
+/// TODO: Access Control; Check that only authorized address can set interest rate data
+contract DefaultAssetInterestRateStrategyTest is Test {
+  using WadRayMath for uint16;
+  using WadRayMath for uint32;
+  using WadRayMath for uint256;
+
+  event RateDataUpdate(
```
DhairyaSethi: use the event from from the interface w/o defining again, like you're doing w errors

tests/unit/DefaultAssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,240 @@
...
+
+    rateStrategy.setInterestRateData(mockAssetId, rateData);
+
+    test_getInterestRateData();
+    test_getOptimalUsageRatio();
+    test_getBaseVariableBorrowRate();
+    test_getVariableRateSlope1();
+    test_getVariableRateSlope2();
+    test_getMaxVariableBorrowRate();
+  }
```
DhairyaSethi: this is fine but what i'd do move this all of the above in a separate test suffixed with .setters.t.sol and have a separate test contract inheriting from a getter test contract which runs all of these and have one fuzz test with fuzzed input and all the failure cases there

src/libraries/logic/AssetLogic.sol
avniculae: hmm yeah, I think we can revert this one to have more generalisability, wdyt @miguelmtzinf ?

tests/Base.t.sol
avniculae: we can have it immutable and fetched in the setup, wdyt?

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: good catch, this is some old code I missed [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: I fuzzed totalDebt as well: [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: MAX_SUPPLY_AMOUNT is not avail in this test file bc it's not inheriting Base, see [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

src/interfaces/IDefaultInterestRateStrategy.sol
avniculae: [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

src/interfaces/IDefaultInterestRateStrategy.sol
avniculae: good catch, thx [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: yeah I think it's a bit more readable this way, and compile should optimise

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: [8a3fd9f](https://github.com/aave/aave-v4/pull/443/commits/8a3fd9fa1fdf8b5985fba0e0d7861f05a4b823ec)

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: I think it's fine to keep this in a single file, as the IR Strategy contract is not too big, wdyt?

src/libraries/logic/AssetLogic.sol
miguelmtzinf: yes, let's do the split please

tests/Base.t.sol
DhairyaSethi: like having all constant assumptions explicitly mentioned in one place, with immutables will have to track the value down, even if this means some duplication

tests/unit/DefaultAssetInterestRateStrategy.t.sol
DhairyaSethi: we can simply inherit from base tho it's fine

tests/unit/DefaultAssetInterestRateStrategy.t.sol
DhairyaSethi: this doesn't completely address the issue tbh, bc totalDebt is fuzzed bw 1e27 & 1e30, ideally we'd have full range on both combinations

tests/unit/DefaultAssetInterestRateStrategy.t.sol
DhairyaSethi: nope even if IR strat is small, it's impactful and can have multiple test files. it's more readable w/o duplication with the suggestion

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,134 @@
...
+  uint256 public constant MAX_OPTIMAL_RATIO = 99_00; // 99.00% in BPS
+
+  /// @dev Map of assetId and their interest rate data (assetId => interestRateData)
+  mapping(uint256 assetId => InterestRateData data) internal _interestRateData;
+
+  /**
+   * @dev Constructor.
+   */
+  constructor() {
+    /// TODO: Access Control; Store authorized address to set interest rate data
```
yan-man: I think can still implement this now to save this authorized addr, even if the other methods don't validate against it yet

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,134 @@
...
+    );
+
+    require(rateData.variableRateSlope1 <= rateData.variableRateSlope2, Slope2MustBeGteSlope2());
+
+    // The maximum rate should not be above certain threshold
+    require(
+      rateData.baseVariableBorrowRate + rateData.variableRateSlope1 + rateData.variableRateSlope2 <=
+        MAX_BORROW_RATE,
+      InvalidMaxRate()
+    );
```
yan-man: can also put these requires into a separate validate method, might be a bit cleaner

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
+import {IDefaultInterestRateStrategy} from 'src/interfaces/IDefaultInterestRateStrategy.sol';
+import {IAssetInterestRateStrategy} from 'src/interfaces/IAssetInterestRateStrategy.sol';
+
+/**
+ * @title DefaultAssetInterestRateStrategy contract
```
yan-man: nit: natspec periods

src/interfaces/IDefaultInterestRateStrategy.sol
```diff
@@ -26,27 +25,17 @@ interface IDefaultInterestRateStrategy is IReserveInterestRateStrategy {
...
    * @dev All values are in basis points (bps), where 1 bps = 0.01%.
    * This means that 10000 bps = 100%.
    * The maximum supported interest rate is 4294967295 bps (2**32-1) or 42949672.95%.
    *
-   * @param optimalUsageRatio The optimal usage ratio, in bps (0-10000)
+   * @param optimalUsageRatio The optimal borrow usage ratio, in bps (1-9900)
    * @param baseVariableBorrowRate The base variable borrow rate, in bps
-   * @param variableRateSlope1 The slope of the variable interest curve, before hitting the optimal ratio, in bps
-   * @param variableRateSlope2 The slope of the variable interest curve, after hitting the optimal ratio, in bps
+   * @param variableRateSlope1 The slope of the variable interest curve, before hitting the optimal borrow usageratio, in bps
```
yan-man:
```suggestion
   * @param variableRateSlope1 The slope of the variable interest curve, before hitting the optimal borrow usage ratio, in bps
```

src/interfaces/IDefaultInterestRateStrategy.sol
```diff
@@ -26,27 +25,17 @@ interface IDefaultInterestRateStrategy is IReserveInterestRateStrategy {
...
   /**
-   * @notice Holds the interest rate data for a given reserve
+   * @notice Holds the interest rate data for a given asset
    *
    * @dev All values are in basis points (bps), where 1 bps = 0.01%.
    * This means that 10000 bps = 100%.
    * The maximum supported interest rate is 4294967295 bps (2**32-1) or 42949672.95%.
    *
-   * @param optimalUsageRatio The optimal usage ratio, in bps (0-10000)
+   * @param optimalUsageRatio The optimal borrow usage ratio, in bps (1-9900)
```
yan-man: why is this changed -> 9900?

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
+import {IDefaultInterestRateStrategy} from 'src/interfaces/IDefaultInterestRateStrategy.sol';
+import {IAssetInterestRateStrategy} from 'src/interfaces/IAssetInterestRateStrategy.sol';
```
yan-man: can we rm this import

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
```diff
@@ -70,7 +70,7 @@ contract BorrowIndex_Scenario1Test is BorrowIndexScenarioBaseTest {
 
   function precondition(Stage stage) internal override {
     super.precondition(stage);
-    mockBaseBorrowRate(state.baseBorrowRate[t]);
+    _mockInterestRate(state.baseBorrowRate[t]);
```
yan-man: separate from this specific PR, but more generally I think we're all over the place on internal test helpers using `_method` (I know I am). Do we want to maintain the _ convention for test helpers too

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
yan-man, line 1: I think you can delete all these borrowIndex tests (1-4), no longer relevant

tests/unit/DefaultAssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,247 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
```
yan-man: let's always use the extended ones now?

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
```
yan-man: let's use all the extended ones now

tests/unit/DefaultAssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,247 @@
...
+  function _computeDebtAndAvailableLiquidity(
+    uint256 targetUtilizationRatioRay,
+    uint256 totalDebtUnbounded
+  ) internal pure returns (uint256 totalDebt, uint256 availableLiquidity) {
+    /// @dev using high value to avoid precision loss
+    totalDebt = bound(totalDebtUnbounded, 10e27, 1e30);
+
+    // utilizationRatio = totalDebt / (totalDebt + availableLiquidity)
+    // utilizationRatio * totalDebt + utilizationRatio * availableLiquidity = totalDebt
+    // availableLiquidity = totalDebt * (1 - utilizationRatio) / utilizationRatio
```
yan-man: rm comments

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
```diff
@@ -530,7 +530,12 @@ contract SpokeRiskPremiumScenarioTest is SpokeBase {
       alice
     );
     assertEq(debtChecks.actualBaseDebt, aliceUsdxInfo.baseDebt, 'alice usdx base debt after');
-    assertEq(debtChecks.actualPremium, aliceUsdxInfo.premiumDebt, 'alice usdx premium debt after');
+    assertApproxEqAbs(
```
yan-man, line 533: hm any idea why this is changing, is it due to not using the extended libs with explicit rounding?

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: if we fuzz totalDebt below 1e27 then we would encounter precision loss

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: for titles as well?

src/interfaces/IDefaultInterestRateStrategy.sol
avniculae: min is 1% and max is 99%

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: we can't, it's used below
```
  function calculateInterestRate(
```

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
avniculae: we can address this later, as part of clean up, wdyt? can you pls create an issue for this?

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: why? aren't they useful?

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
avniculae: no, it's bc 1 wei of premium is created instantly due to the different rounding directions of the premium offset and the premium assets

DhairyaSethi: i think he meant why the ir strategy change leads to needing an update here unlike before

tests/unit/DefaultAssetInterestRateStrategy.t.sol
DhairyaSethi: imo useful, would like to keep

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
avniculae: I think this was not surfaced before, but the increased precision of the interest rate strategy might have helped with surfacing this easier

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: [9a4ac24](https://github.com/aave/aave-v4/pull/443/commits/9a4ac24e36952521bb695e9c435d81cbc51afb9b)

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: [9a4ac24](https://github.com/aave/aave-v4/pull/443/commits/9a4ac24e36952521bb695e9c435d81cbc51afb9b)

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
avniculae: [9a4ac24](https://github.com/aave/aave-v4/pull/443/commits/9a4ac24e36952521bb695e9c435d81cbc51afb9b)

src/interfaces/IDefaultInterestRateStrategy.sol
avniculae: [9a4ac24](https://github.com/aave/aave-v4/pull/443/commits/9a4ac24e36952521bb695e9c435d81cbc51afb9b)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: I think it's fine to keep it like this, as function is not complex at all -- just a setter, wdyt?

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: I would like to wait until we implement AccessControl in the entire repo, as it's still undecided which approach we want to take there

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: I am not sure I understand the motivation/benefits of this tbh, can you pls elaborate?

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: [63a0564](https://github.com/aave/aave-v4/pull/443/commits/63a056471ee684021006793e64f1c418a12ae1f0)

tests/Base.t.sol
avniculae: in this case I would keep it like this, wdyt?

src/libraries/logic/AssetLogic.sol
avniculae: [9a4ac24](https://github.com/aave/aave-v4/pull/443/commits/9a4ac24e36952521bb695e9c435d81cbc51afb9b)

tests/unit/DefaultAssetInterestRateStrategy.t.sol
DhairyaSethi: avoids having to duplicate test runs like this, they're implicitly run on a new setup 
separation of concerns bc this test file contains everything

src/interfaces/IAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+/**
+ * @title IAssetInterestRateStrategy
+ * @author Aave Labs
+ * @notice Basic interface for any rate strategy used by the Aave protocol
+ */
+interface IAssetInterestRateStrategy {
```
miguelmtzinf: Thinking about dropping `DefaultAssetInterestRateStrategy` name and using `AssetInterestRateStrategy` instead.
Then, we can merge both interfaces or keeping this one because it could be useful for integrations that only need to call the calc function. In this case, we can call this one smt like `IBasicInterestRateStrategy` or `IMinimalInterestRateStrategy`.

thoughts?

src/interfaces/IAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,26 @@
...
+   * @param totalDebt The total debt of the asset
+   * @param availableLiquidity The available liquidity of the asset
+   * @param liquidityAdded The amount of liquidity added to the asset
+   * @param liquidityTaken The amount of liquidity taken from the asset
+   * @return variableBorrowRate The variable borrow rate expressed in ray
+   */
+  function calculateInterestRate(
+    uint256 assetId,
+    uint256 totalDebt,
+    uint256 availableLiquidity,
```
miguelmtzinf:
```suggestion
    uint256 availableLiquidity,
    uint256 totalDebt,
```
feels better to flip them

src/interfaces/IAssetInterestRateStrategy.sol
avniculae: i agree, i don't like the default terminology either

avniculae: i think keeping interfaces separate is still beneficial tbh

src/interfaces/IDefaultInterestRateStrategy.sol
```diff
@@ -56,92 +45,92 @@ interface IDefaultInterestRateStrategy is IReserveInterestRateStrategy {
...
-   *   Being specific to this custom implementation, with custom struct type,
-   *   overloading the function on the generic interface
+   * @notice Thrown when the max possible rate is greater than `MAX_BORROW_RATE`
+   */
+  error InvalidMaxRate();
+
+  /**
+   * @notice Thrown when slope 2 (after kink point) is less than slope 1 (before kink point)
+   */
+  error Slope2MustBeGteSlope2();
```
miguelmtzinf:
```suggestion
  error Slope2MustBeGteSlope1();
```
?

src/contracts/DefaultAssetInterestRateStrategy.sol
miguelmtzinf: imo it's fine, but would remove linebreaks
```
checks

effects

event
```

src/contracts/DefaultAssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,138 @@
...
+    uint256 assetId,
+    uint256 totalDebt,
+    uint256 availableLiquidity,
+    uint256 liquidityAdded,
+    uint256 liquidityTaken
+  ) external view virtual override returns (uint256) {
+    InterestRateData memory rateData = _interestRateData[assetId];
+    require(rateData.optimalUsageRatio != 0, InterestRateDataNotSet(assetId));
+
+    uint256 currentVariableBorrowRateRay = rateData.baseVariableBorrowRate.bpsToRay();
```
miguelmtzinf: i think we can do `.bpsToRay()` just at the end, before returning? It's fine to sum variables in ray later (base + slope1 + slope2).bpsToRay()

src/contracts/DefaultAssetInterestRateStrategy.sol
miguelmtzinf: i think it reads better combining actually

src/libraries/logic/AssetLogic.sol
```diff
@@ -114,17 +114,13 @@ library AssetLogic {
     uint256 liquidityAdded,
     uint256 liquidityTaken
   ) internal {
-    asset.baseBorrowRate = asset.config.irStrategy.calculateInterestRates(
```
miguelmtzinf, line 117: can you check if we can remove `baseInterestRate` getter, right before this fn

src/libraries/math/WadRayMathExtended.sol
```diff
@@ -182,9 +184,13 @@ library WadRayMathExtended {
...
    */
-  function dewadify(uint256 a) internal pure returns (uint256 b) {
+  function dewadifyDown(uint256 a) internal pure returns (uint256 b) {
     assembly {
       b := div(a, WAD)
     }
   }
+
+  function bpsToRay(uint256 a) internal pure returns (uint256) {
+    return (a * RAY) / PercentageMathExtended.PERCENTAGE_FACTOR;
```
miguelmtzinf: wondering if we should just add `PERCENTAGE_FACTOR` as constant here.

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: wdym by remove linebreaks? this is following the prettier rules

tests/unit/DefaultAssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,319 @@
...
+    } else {
+      assertApproxEqAbs(variableBorrowRate, expectedVariableRate, 0.0001e27);
+    }
+  }
+
+  function test_calculateInterestRate_AtMaxUtilization() public {
+    test_calculateInterestRate_RightToKinkPoint(100_00, 100e18);
+  }
+
+  function _computeCalculateInterestRateParams(
```
miguelmtzinf: i d rename to `bound`

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
```diff
@@ -142,15 +148,6 @@ abstract contract LiquidityHubScenarioBaseTest is Base {
     return uint40(timestamps[uint256(Stage.unwrap(stage))]);
   }
 
-  /// @param baseBorrowRate base borrow rate in bps
-  function mockBaseBorrowRate(uint256 baseBorrowRate) internal {
```
miguelmtzinf, line 146: is this not used across tests?
there may be something similar, smt like `mockRate`

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: I think converting to ray first gives us higher precision when computing interest rate according to slopes, wdyt?

src/libraries/math/WadRayMathExtended.sol
avniculae: I don't have strong opinions at the moment, we need to revisit this in any case when we'll add the missing fns #460

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: why? the function takes in an argument and computes 4 values and returns them

tests/scenario/liquidityHub/LiquidityHub.ScenarioBase.t.sol
avniculae: I've deleted all helper fns doing the same thing (mocking rate), and added 2 _mockRate fns in Base.t.sol

snapshots/Hub.Operations.json
miguelmtzinf, line 1: wondering if we wanna add a gas profile for `calculateBorrowRate`

tests/unit/DefaultAssetInterestRateStrategy.t.sol
miguelmtzinf: because it generates arbitrary values by fuzzing the usageRatio. We can find another name, but the current one it's not good imho

avniculae: ah I see, I agree that `compute` does not work well with the random mechanism. shall we call it `_generateCalculateInterestRateParams`?

snapshots/Hub.Operations.json
avniculae: hmm we definitely can, but not sure if we want since no future changes are foreseen for the interest rate strategy wdyt?

src/libraries/logic/AssetLogic.sol
avniculae: [eeb99e9](https://github.com/aave/aave-v4/pull/443/commits/eeb99e99ad4f91b1019f82bc3a2d607ce4fd2c07)

src/interfaces/IDefaultInterestRateStrategy.sol
avniculae: [eeb99e9](https://github.com/aave/aave-v4/pull/443/commits/eeb99e99ad4f91b1019f82bc3a2d607ce4fd2c07)

src/interfaces/IAssetInterestRateStrategy.sol
avniculae: [eeb99e9](https://github.com/aave/aave-v4/pull/443/commits/eeb99e99ad4f91b1019f82bc3a2d607ce4fd2c07)

src/interfaces/IAssetInterestRateStrategy.sol
avniculae: [eeb99e9](https://github.com/aave/aave-v4/pull/443/commits/eeb99e99ad4f91b1019f82bc3a2d607ce4fd2c07)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: I think separate is more readable bc one can think of the 2 different sections defined by the kink point, but nbd 

[eeb99e9](https://github.com/aave/aave-v4/pull/443/commits/eeb99e99ad4f91b1019f82bc3a2d607ce4fd2c07)

tests/unit/DefaultAssetInterestRateStrategy.t.sol
avniculae: [131427e](https://github.com/aave/aave-v4/pull/443/commits/131427eb88be2dc6ea429c36911bcdc6b8154b5f)

src/contracts/DefaultAssetInterestRateStrategy.sol
yan-man: eh not even sure about this nbd specifically for titles, but applies everywhere else anyway

src/interfaces/IDefaultInterestRateStrategy.sol
yan-man: jw why not 0.01% and 99.99%?

tests/scenario/liquidityHub/borrowIndex/BorrowIndex_Scenario1.t.sol
yan-man: yeah let's leave for later

https://github.com/aave/aave-v4/issues/430#issuecomment-2986371872

tests/unit/DefaultAssetInterestRateStrategy.t.sol
yan-man: ah sorry thought these were commented out old code

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
yan-man: can you create an issue for this for later and then feel free to resolve so we can get this merged

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * @title AssetInterestRateStrategy contract
+ * @author Aave Labs
+ * @notice Asset interest rate strategy used by the Aave protocol
+ * @dev Strategies are hub-specific: one strategy CAN'T be used across different Aave hubs
+ *   due to the usage of asset id as index of the _interestRateData
+ */
+contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
+  using WadRayMathExtended for *;
```
yan-man, line 15: does this need to apply to * instead of just uint256

avniculae: yes we need bc the interest rate data fields are uint16 and uint32, and we apply bpsToRay to those. we also apply rayMul, rayDiv for uint255.

src/interfaces/IDefaultInterestRateStrategy.sol
avniculae: hmm this is what `MIN_OPTIMAL_RATIO` and `MAX_OPTIMAL_RATIO` are set to

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
DhairyaSethi: @yan-man i think i'm pretty convinced of the above reasoning, since instant premium debt creation happens bc of precision loss while using converting through debt ex rate (ie index which depends on borrow rate) and we had a static borrow rate before it's possible we didn't see it here 

did you have any thoughts or particular concerns with that not being the reason?

src/contracts/DefaultAssetInterestRateStrategy.sol
miguelmtzinf: ah, it's only worth when above the kink point. all good

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,319 @@
...
+    test_calculateInterestRate_RightToKinkPoint(100_00, 100e18);
+  }
+
+  function _generateCalculateInterestRateParams(
+    uint256 targetUtilizationRatioRay
+  )
+    internal
+    returns (
+      uint256 totalDebt,
+      uint256 availableLiquidity,
```
miguelmtzinf: flip
```suggestion
      uint256 availableLiquidity,
      uint256 totalDebt,
```

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,319 @@
...
+    test_calculateInterestRate_LeftToKinkPoint(100_00, 100e18);
+  }
+
+  function test_calculateInterestRate_RightToKinkPoint(uint256 utilizationRatio, uint256) public {
+    uint256 utilizationRatioRay = bound(utilizationRatio, rateData.optimalUsageRatio + 1, 100_00)
+      .bpsToRay();
+
+    (
+      uint256 totalDebt,
+      uint256 availableLiquidity,
```
miguelmtzinf:
```suggestion
      uint256 availableLiquidity,
      uint256 totalDebt,
```

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,319 @@
...
+  function test_calculateInterestRate_ZeroDebtZeroLiquidity() public {
+    test_calculateInterestRate_fuzz_ZeroDebt(0, 0, 0);
+  }
+
+  function test_calculateInterestRate_LeftToKinkPoint(uint256 utilizationRatio, uint256) public {
+    uint256 utilizationRatioRay = bound(utilizationRatio, 1, rateData.optimalUsageRatio).bpsToRay();
+
+    (
+      uint256 totalDebt,
+      uint256 availableLiquidity,
```
miguelmtzinf:
```suggestion
      uint256 availableLiquidity,
      uint256 totalDebt,
```

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,319 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+import {WadRayMathExtended} from 'src/libraries/math/WadRayMathExtended.sol';
+import {AssetInterestRateStrategy, IAssetInterestRateStrategy} from 'src/contracts/AssetInterestRateStrategy.sol';
+
+import {Test, stdError} from 'forge-std/Test.sol';
+
+/// TODO: Access Control; Check that only authorized address can set interest rate data
+contract AssetInterestRateStrategyTest is Test {
```
miguelmtzinf: why do we want to inherit from `Test` and not `BaseTest`

snapshots/Hub.Operations.json
miguelmtzinf: just curious about how much it costs, nbd

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: this stayed the same from old tests. These are standalone tests and we don't need anything from the Base, I'm fine with both options tbh, wdyt?

DhairyaSethi: let's inherit from Base and completely override the `setUp` imo, it's more consistent, cleaner imports, and we get Base helpers if they could be potentially needed in the future

src/contracts/DefaultAssetInterestRateStrategy.sol
DhairyaSethi: groups all checks w/o line breaks and so on, prettier only enforces at max one line break but that one line break is at the user's discretion

src/libraries/math/WadRayMathExtended.sol
DhairyaSethi: let's add imo, and add to the unit test on this lib which validates constants

tests/unit/DefaultAssetInterestRateStrategy.t.sol
DhairyaSethi: generate is also not quite accurate bc it's more _generateWith(utilisationRatio)_, i agree something along the lines of bound makes more sense but nbd imo

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: when I run prettier it splits a single line on 5 lines, as follows
```
require(
      MIN_OPTIMAL_RATIO <= rateData.optimalUsageRatio &&
        rateData.optimalUsageRatio <= MAX_OPTIMAL_RATIO,
      InvalidOptimalUsageRatio()
    );
```

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: [fbd223c](https://github.com/aave/aave-v4/pull/443/commits/fbd223c5486b166d3f1473595b588586ddf41f39)

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: [fbd223c](https://github.com/aave/aave-v4/pull/443/commits/fbd223c5486b166d3f1473595b588586ddf41f39)

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: [fbd223c](https://github.com/aave/aave-v4/pull/443/commits/fbd223c5486b166d3f1473595b588586ddf41f39)

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: [fbd223c](https://github.com/aave/aave-v4/pull/443/commits/fbd223c5486b166d3f1473595b588586ddf41f39)

src/libraries/math/WadRayMathExtended.sol
avniculae: [fbd223c](https://github.com/aave/aave-v4/pull/443/commits/fbd223c5486b166d3f1473595b588586ddf41f39)

src/contracts/DefaultAssetInterestRateStrategy.sol
avniculae: [fbd223c](https://github.com/aave/aave-v4/pull/443/commits/fbd223c5486b166d3f1473595b588586ddf41f39)

src/libraries/math/WadRayMathExtended.sol
```diff
@@ -182,9 +183,13 @@ library WadRayMathExtended {
...
    * @return b (= a / 1e18)
    */
-  function dewadify(uint256 a) internal pure returns (uint256 b) {
+  function dewadifyDown(uint256 a) internal pure returns (uint256 b) {
     assembly {
       b := div(a, WAD)
     }
   }
+
+  function bpsToRay(uint256 a) internal pure returns (uint256) {
```
miguelmtzinf, line 197: add docs pls

src/libraries/types/DataTypes.sol
```diff
@@ -42,20 +42,10 @@ library DataTypes {
     bool paused;
     uint256 decimals;
     uint256 liquidityFee;
-    IReserveInterestRateStrategy irStrategy;
+    IBasicInterestRateStrategy irStrategy;
```
miguelmtzinf, line 45: can we change to address so we don't need the import here? cc @DhairyaSethi

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -0,0 +1,316 @@
...
+      totalDebt: totalDebt,
+      liquidityAdded: liquidityAdded,
+      liquidityTaken: liquidityTaken
+    });
+  }
+
+  function test_calculateInterestRate_revertsWith_ArithmeticUnderflow() public {
+    test_calculateInterestRate_fuzz_revertsWith_ArithmeticUnderflow({
+      totalDebt: 100e6,
+      availableLiquidity: 100e6,
```
miguelmtzinf:
```suggestion
      availableLiquidity: 100e6,
      totalDebt: 100e6,
```

src/libraries/types/DataTypes.sol
avniculae: I think it's better to do this in #454, since many tests will need to be adjusted, and they were already touched in the Configurator PR, wdyt?

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: [0218143](https://github.com/aave/aave-v4/pull/443/commits/0218143e58d15bdbd48fb6ceea71799ea2c7aae2)

src/libraries/math/WadRayMathExtended.sol
avniculae: [0218143](https://github.com/aave/aave-v4/pull/443/commits/0218143e58d15bdbd48fb6ceea71799ea2c7aae2)

tests/unit/WadRayMath.t.sol
```diff
@@ -379,4 +385,20 @@ contract WadRayMathExtendedDifferentialTests is Test {
...
+    unchecked {
+      b = a * wExtended.RAY();
+      safetyCheck = b / wExtended.RAY() == a;
+    }
+    if (!safetyCheck) {
+      vm.expectRevert();
+      wExtended.bpsToRay(a);
+    } else {
+      assertEq(wExtended.bpsToRay(a), (a * w.RAY()) / 100_00);
+      assertEq(wExtended.bpsToRay(a), b / 100_00);
```
DhairyaSethi, line 401:
```suggestion
      assertEq(wExtended.bpsToRay(a), (a * w.RAY()) / w.PERCENTAGE_FACTOR());
      assertEq(wExtended.bpsToRay(a), b / w.PERCENTAGE_FACTOR());
```

tests/unit/Spoke/Spoke.RiskPremium.Scenario.t.sol
yan-man: ok if you guys feel no need to investigate this more, fine 

> instant premium debt creation happens bc of precision loss

is this true, i thought due to explicitly opposite rounding directions

DhairyaSethi: yes it's due to the precision loss from opposite rounding directions

Issue #397, opened 2025-05-20
Hub & Spoke Clean Up


avniculae:
Hub
  - replace `_validateSupply` with `_validateAdd` in Hub
  - replace `_validateWithdraw` with `_validateRemove` in Hub
  - replace `LiquidityHub.Supply.t.sol` with `LiquidityHub.Add.t.sol`, and replace `test_supply_...` tests with `test_add_...`
  - replace supplyCap with addCap, 
  - replace drawCap with removeCap
  - rename LiquidityHub to Hub

Spoke
  - replace `userAddress` with `user`
  - replace `liquidityPremium` with `collateralRisk`
  - replace Invalid... errors with ZeroAddress / ZeroAmount for constructor and gov-only fns (https://github.com/aave/aave-v4/pull/478#discussion_r2184640534) --> maybe applicable to hub as well

yan-man:
only remaining ones:
- replace supplyCap with addCap 
- replace Invalid... errors with ZeroAddress / ZeroAmount for constructor and gov-only fns (https://github.com/aave/aave-v4/pull/478#discussion_r2184640534)

PR #447, opened 2025-06-14
feat: Add Configurator

### LINKED ISSUE #445: Configurator functions for Hub and Spoke risk params
The LiquidityHub has very open functions to update config params. This is cumbersome, risky and error-prone.
https://github.com/aave/aave-v4/blob/e91cb44338f5a105ae21fae92af1508f7685ec4a/src/contracts/LiquidityHub.sol#L34-L75

## Context
For context, in V3 we have a separate contract (`PoolConfigurator`) which calls the core contract (`Pool`) to perform configurations. This contract has granular functions to apply updates on the pool reserves.
Apart from that, there is a periphery contract called [`ConfigEngine`](https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/extensions/v3-config-engine/AaveV3ConfigEngine.sol) that is used for configuration purposes as a library via delegatecall. X entity has permission to call ABC functions on the pool, and uses `ConfigEngine` via delegatecall to perform any update.

## Proposed implementation
Having granular functions to update each parameter individually would be overkill. I d propose the following idea:
- Add enough functions to update safely the params of the Hub and Spoke.
  -  Some params (e.g. reserveFactor) need to execute some actions afterwards (e.g. accrue interest, see #438), and thus, they cannot be put together other params update (theoretically doable, but code would become very messy).
  - Rest of params can be put into groups that make sense: config flags, caps, etc
- Create an external contract (`Configurator`) with granular fns to update each parameter. This one will have proper rights in the Hub/Spoke and perform all sanity checks needed.
- We can build something similar to `ConfigEngine` later.

avniculae:
Closes #445 

TODO:
- https://github.com/aave/aave-v4/pull/375#discussion_r2149877037
- https://github.com/aave/aave-v4/pull/375#discussion_r2149929698


src/contracts/LiquidityHub.sol
```diff
@@ -25,18 +28,34 @@ contract LiquidityHub is ILiquidityHub {
...
+
+    require(asset != address(0), InvalidAssetAddress());
+    require(irStrategyAddress != address(0), InvalidIrStrategy());
+
+    uint256 id = assetsList.length;
     assetsList.push(IERC20(asset));
-    uint256 assetId = assetCount++;
-    _assets[assetId] = DataTypes.Asset({
+
+    uint256 decimals = IERC20Metadata(asset).decimals();
```
miguelmtzinf, line 45: we used to pass the decimals as input because some assets do not respect the interface (maker dai)

src/contracts/LiquidityHub.sol
```diff
@@ -45,34 +64,47 @@ contract LiquidityHub is ILiquidityHub {
...
-      id: assetId, // todo rm
-      config: DataTypes.AssetConfig({
-        decimals: config.decimals, // todo fetch decimals from token
-        active: config.active,
-        frozen: config.frozen,
-        paused: config.paused,
-        irStrategy: config.irStrategy
-      })
+      baseBorrowRate: 0,
+      config: config
```
miguelmtzinf, line 67: let's put here the config right away, without intermediate variable

src/contracts/LiquidityHub.sol
```diff
@@ -45,34 +64,47 @@ contract LiquidityHub is ILiquidityHub {
...
-    emit AssetAdded(assetId, asset);
-    emit AssetConfigUpdated(assetId, config);
+    emit AssetAdded(id, asset);
+    emit AssetConfigUpdated(id, config);
   }
 
-  function updateAssetConfig(uint256 assetId, DataTypes.AssetConfig calldata config) external {
-    _validateAssetConfig(config, address(assetsList[assetId]));
-    DataTypes.Asset storage asset = _assets[assetId];
+  function updateAssetFlags(uint256 assetId, bool active, bool paused, bool frozen) external {
```
miguelmtzinf, line 74: please check v3.3 to see if there is any logic to exec when applying updates to these flags.
I think they set ltv to 0 when deactivating or smt like that, but it may not be interesting for us.
please research and provide insights in the pr desc

src/libraries/logic/AssetLogic.sol
```diff
@@ -15,9 +15,19 @@ library AssetLogic {
...
+
   // todo: option for cached object
   // todo: add virtual offset for inflation attack
 
+  function get(
+    mapping(uint256 assetId => DataTypes.Asset assetData) storage assets,
+    uint256 assetId
+  ) internal view returns (DataTypes.Asset storage asset) {
+    asset = assets[assetId];
+    require(address(asset.config.irStrategy) != address(0), AssetNotListed());
```
miguelmtzinf, line 28: this is a weird way to check existence. We have a similar issue with existence of spokes...

src/contracts/LiquidityHub.sol
```diff
@@ -45,34 +64,47 @@ contract LiquidityHub is ILiquidityHub {
...
+    require(reserveFactor <= PercentageMath.PERCENTAGE_FACTOR, InvalidReserveFactor());
+
+    DataTypes.Asset storage asset = _assets.get(assetId);
+    asset.accrue();
+    asset.config.reserveFactor = reserveFactor;
+
+    emit AssetConfigUpdated(assetId, asset.config);
+  }
+
+  function updateAssetIrStrategy(uint256 assetId, address irStrategyAddress) external {
```
```suggestion
  function updateInterestRateStrategy(uint256 assetId, address irStrategy) external {
```

src/contracts/LiquidityHub.sol
```diff
@@ -45,34 +64,47 @@ contract LiquidityHub is ILiquidityHub {
...
-    emit AssetConfigUpdated(assetId, config);
+    DataTypes.Asset storage asset = _assets.get(assetId);
+    asset.config.active = active;
+    asset.config.paused = paused;
+    asset.config.frozen = frozen;
+
+    emit AssetConfigUpdated(assetId, asset.config);
+  }
+
+  function updateAssetReserveFactor(uint256 assetId, uint256 reserveFactor) external {
```
```suggestion
  function updateReserveFactor(uint256 assetId, uint256 reserveFactor) external {
```

src/contracts/LiquidityHub.sol
```diff
@@ -100,8 +132,8 @@ contract LiquidityHub is ILiquidityHub {
   ) external {
     // TODO: AccessControl
     _spokes[assetId][spoke].config = DataTypes.SpokeConfig({
-      drawCap: config.drawCap,
-      supplyCap: config.supplyCap
+      supplyCap: config.supplyCap,
```
miguelmtzinf, line 134: can you create a separate ticket for a refactor/clean up on the Hub?
Suggestions for renaming:
- supplyCap -> addCap
- drawCap -> removeCap
- LiquidityHub -> Hub

src/contracts/Configurator.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.10;
+
+import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+
+import {IConfigurator, ILiquidityHub} from 'src/interfaces/IConfigurator.sol';
+
+contract Configurator is IConfigurator {
+  ILiquidityHub public immutable override HUB;
```
miguelmtzinf: we are not adding `override` if not needed usually

src/contracts/LiquidityHub.sol
avniculae: Updated existing #397 issue

src/libraries/logic/AssetLogic.sol
avniculae: we could also compare assetId to the length of the _assetsList array, but it would require one more arg to this fn. alternatively, we could move this function to the hub, but we would not be able to do `_assets.get(..)` anymore, wdyt?

src/contracts/LiquidityHub.sol
avniculae: did this bc config is emitted in the event below, wdyt?

src/contracts/LiquidityHub.sol
avniculae: I see that DAI has `decimals()` [exposed](https://etherscan.io/token/0x6b175474e89094c44da98b954eedeac495271d0f#readContract#F5). I did this change bc we also had an inline TODO to do this. wdyt?

src/contracts/Configurator.sol
avniculae: curious to understand why, isn't override considered good practice?

src/libraries/types/DataTypes.sol
```diff
@@ -17,6 +17,7 @@ library DataTypes {
   }
 
   struct Asset {
+    uint256 id;
```
DhairyaSethi: rm, we don't want this here since this struct is uniquely referenced by its id

src/libraries/logic/AssetLogic.sol
DhairyaSethi: we already have the length check for invalid asset id when we do _assetList[id].safeTransfer in all spoke methods except refresh thx to solc
i'm good with avoiding this entirely tbh or length check in the hub and loosing the syntax sugar in which case do the array lookup for transfer in assembly w/o bound check bc we would've failed early

src/libraries/types/DataTypes.sol
avniculae: [6b56f4a](https://github.com/aave/aave-v4/pull/447/commits/6b56f4a96ef6b1ed581eee76151468638301bbc7)

src/contracts/LiquidityHub.sol
avniculae: [24a37f7](https://github.com/aave/aave-v4/pull/447/commits/24a37f73c7db872955c12d8e364506c3e9d567fe)

src/contracts/LiquidityHub.sol
avniculae: [24a37f7](https://github.com/aave/aave-v4/pull/447/commits/24a37f73c7db872955c12d8e364506c3e9d567fe)

PR #455, opened 2025-06-17
Add Access Control management

### LINKED ISSUE #4: Access Control (RBAC system)

CheyenneAtapour:
This is a proof of concept implementation showcasing using OpenZeppelin's `AccessManaged` and `AccessManager` contracts for access control.

Closes #4

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -396,8 +396,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   function updateUserRiskPremium(address user) external {
     (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
     bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
-    // check permissions if premium increases or not called by user
-    if (premiumIncrease || user != msg.sender) {
+    // check permissions if premium increases and not called by user
+    if (premiumIncrease && user != msg.sender) {
       _checkCanCall(_msgSender(), _msgData());
     }
     emit UserRiskPremiumUpdate(user, userRiskPremium);
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -9,8 +9,8 @@
   "supply: 1 debt": "174178",
   "supply: 2 debt": "174189",
   "supply: 3 debt": "174142",
-  "updateUserRiskPremium": "176435",
-  "updateUserRiskPremium: 2 debts": "226122",
+  "updateUserRiskPremium": "176407",
+  "updateUserRiskPremium: 2 debts": "226094",
   "usingAsCollateral": "62112",
   "withdraw: full": "233457",
   "withdraw: partial": "244529"


src/dependencies/openzeppelin/AccessManaged.sol
```diff
@@ -0,0 +1,113 @@
...
+      authority(),
+      caller,
+      address(this),
+      bytes4(data[0:4])
+    );
+    if (!immediate) {
+      if (delay > 0) {
+        _consumingSchedule = true;
+        IAccessManager(authority()).consumeScheduledOp(caller, data);
+        _consumingSchedule = false;
```
DhairyaSethi, line 107: i'd like to see a modern tstore variant of this tbh, this approach seems very gas expensive over access control

CheyenneAtapour: I don't imagine we would need the delay for granting access, so we can omit all this logic for delay

DhairyaSethi: agree 💯

src/contracts/LiquidityHub.sol
```diff
@@ -234,9 +245,7 @@ contract LiquidityHub is ILiquidityHub {
   }
 
   /// @inheritdoc ILiquidityHub
-  function draw(uint256 assetId, uint256 amount, address to) external returns (uint256) {
-    // TODO: authorization - only spokes
-
+  function draw(uint256 assetId, uint256 amount, address to) external restricted returns (uint256) {
```
DhairyaSethi: on Hub do we need the spoke setup on accessManager? isn't it cheaper and better to just validate spoke exists from the stored mapping if required?

src/contracts/Spoke.sol
```diff
@@ -374,8 +379,7 @@ contract Spoke is ISpoke, Multicall {
   function updateUserRiskPremium(address user) external {
     (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
     bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
-    // todo allow authorized caller to increase as well
-    require(msg.sender == user || !premiumIncrease, Unauthorized());
+    require(msg.sender == user || _isGovernor(msg.sender) || !premiumIncrease, Unauthorized());
```
```suggestion
    require(!premiumIncrease || msg.sender == user || _isGovernor(msg.sender), Unauthorized());
```
cheaper in all cases

src/contracts/Spoke.sol
```diff
@@ -712,6 +716,12 @@ contract Spoke is ISpoke, Multicall {
     return userPosition.baseDrawnShares > 0;
   }
 
+  function _isGovernor(address user) internal view returns (bool) {
```
DhairyaSethi: i like to call this `who` or `caller`, not really an user but nbd

src/dependencies/openzeppelin/AuthorityUtils.sol
```diff
@@ -0,0 +1,32 @@
...
+  ) internal view returns (bool immediate, uint32 delay) {
+    (bool success, bytes memory data) = authority.staticcall(
+      abi.encodeCall(IAuthority.canCall, (caller, target, selector))
+    );
+    if (success) {
+      if (data.length >= 0x40) {
+        (immediate, delay) = abi.decode(data, (bool, uint32));
+      } else if (data.length >= 0x20) {
+        immediate = abi.decode(data, (bool));
+      }
```
DhairyaSethi, line 28: let's add a test to be sure our `authority` is of the latter type here (w/o delay) bc we want to avoid paying for that functionality here
also can we confirm it's a zero cost abstraction like _msgSender?

src/dependencies/openzeppelin/AccessManaged.sol
```diff
@@ -0,0 +1,113 @@
...
+  function _setAuthority(address newAuthority) internal virtual {
+    _authority = newAuthority;
+    emit AuthorityUpdated(newAuthority);
+  }
+
+  /**
+   * @dev Reverts if the caller is not allowed to call the function identified by a selector. Panics if the calldata
+   * is less than 4 bytes long.
+   */
+  function _checkCanCall(address caller, bytes calldata data) internal virtual {
```
miguelmtzinf, line 96: damn, this does not check the target contract right?
i thought it was by caller, target and fn
if there are 2 fns with same name we d be screwed

src/libraries/types/Roles.sol
miguelmtzinf, line 1: i thought this is just configuration to be made, not part of the code

src/contracts/LiquidityHub.sol
miguelmtzinf, line 1: IR needs access control too

tests/Base.t.sol
```diff
@@ -151,29 +155,82 @@ abstract contract Base is Test {
...
     vm.label(address(spoke3), 'spoke3');
   }
 
+  function setUpRoles() internal {
+    // Hub Admin has default admin role
+    vm.startPrank(HUB_ADMIN);
+    // Grant roles
+    accessManager.grantRole(Roles.HUB_ADMIN_ROLE, HUB_ADMIN, 0);
+    accessManager.grantRole(Roles.SPOKE_ADMIN_ROLE, HUB_ADMIN, 0);
+    accessManager.grantRole(Roles.TREASURY_ADMIN_ROLE, HUB_ADMIN, 0);
```
yan-man: is this role needed?

src/contracts/LiquidityHub.sol
yan-man: discussed today, but if done thru stored mapping, prefer to base it on an explicit flag rather than supply/draw cap

tests/Base.t.sol
```diff
@@ -151,29 +155,82 @@ abstract contract Base is Test {
...
+
+    accessManager.setTargetFunctionRole(address(hub), hubSelectors, Roles.HUB_ADMIN_ROLE);
+
+    // Spoke functionalities
+    bytes4[] memory spokeSelectors = new bytes4[](5);
+    spokeSelectors[0] = ILiquidityHub.add.selector;
+    spokeSelectors[1] = ILiquidityHub.remove.selector;
+    spokeSelectors[2] = ILiquidityHub.draw.selector;
+    spokeSelectors[3] = ILiquidityHub.restore.selector;
+    spokeSelectors[4] = ILiquidityHub.refreshPremiumDebt.selector;
```
yan-man: assuming we don't read directly from `_spokes` were you thinking these would happen as a separate tx, rather than incorporated into `addSpoke`?

tests/Base.t.sol
```diff
@@ -151,29 +155,82 @@ abstract contract Base is Test {
...
 
   function deployFixtures() internal {
+    vm.startPrank(HUB_ADMIN);
     oracle = new MockPriceOracle();
     irStrategy = new DefaultReserveInterestRateStrategy(mockAddressesProvider);
-    hub = new LiquidityHub();
-    spoke1 = ISpoke(new Spoke(address(hub), address(oracle)));
-    spoke2 = ISpoke(new Spoke(address(hub), address(oracle)));
-    spoke3 = ISpoke(new Spoke(address(hub), address(oracle)));
+    accessManager = new AccessManager(HUB_ADMIN);
```
yan-man: think this is a bit confusing bc `HUB_ADMIN` is both `AccessManager` initialAdmin and then admin on the hub itself too. Maybe this can be renamed. 
 
bc `HUB_ADMIN` is the one that grants admin roles on the spoke too which is a bit weird

src/contracts/Spoke.sol
```diff
@@ -719,6 +719,12 @@ contract Spoke is ISpoke, Multicall {
...
+    AccessManager accessManager = AccessManager(authority());
+    (bool result, ) = accessManager.hasRole(Roles.DEFAULT_ADMIN_ROLE, user);
```
DhairyaSethi: pls use interface instead of the contract here
also i think another role over default admin makes more sense?

src/libraries/types/Roles.sol
CheyenneAtapour: The roles themselves are uint64 which get passed into the Open Zeppelin functions. I think it's cleaner to have them as constants like this. But I could also just have these defined in only the Base setup. The one part of concern now is when checking if a governor can increase a user's RP. But I could add that to a constant within just the spoke

src/contracts/Spoke.sol

src/contracts/Spoke.sol

tests/Base.t.sol
CheyenneAtapour: No, now we'll go with `owner`, so removed

src/contracts/Spoke.sol

tests/Base.t.sol
CheyenneAtapour: Yeah, I made a separate `ADMIN` role. I'm adopting the convention that for a full setup, I'll use admin, because that feels more realistic, and also annoying to keep switching pranks during a full setup. But then for individual config changes, I'll prank the least privileged account that can call that function

src/contracts/LiquidityHub.sol
CheyenneAtapour: Added

src/dependencies/openzeppelin/AccessManaged.sol
CheyenneAtapour: Each target inherits from `AccessManaged` itself, so it's including itself as the target in this call. Check how it's passing `address(this)` on line 100.

The `AccessManager` can indeed handle this with checking different targets. Please check out this function: https://github.com/aave/aave-v4/blob/df194db3044d5afa9cd91d1250c6bfe0f4d3c5d6/src/dependencies/openzeppelin/AccessManager.sol#L139-L155

tests/Base.t.sol
CheyenneAtapour: Added flag during `addSpoke` and now we check the flag from `_spokes` storage instead of a role

src/contracts/AssetInterestRateStrategy.sol
yan-man, line 1: just to clarify, this is merely a renaming of `DefaultReserveIRStrategy` contract right?

src/contracts/Spoke.sol
```diff
@@ -43,15 +43,9 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-    address hubAddress,
-    address oracleAddress,
-    address accessManager
-  ) AccessManaged(accessManager) {
-    require(hubAddress != address(0), InvalidHubAddress());
+  constructor(address oracleAddress, address accessManager) AccessManaged(accessManager) {
     require(oracleAddress != address(0), InvalidOracleAddress());
 
-    HUB = ILiquidityHub(hubAddress);
     oracle = IPriceOracle(oracleAddress);
```
yan-man, line 56: nbd here and doesnt need to be settled now, but I remember Emilio commented on this and I'd like to have something in place too to distinguish function args, I prefer `variable_` but wondering if you guys had thoughts on preference. So that we don't have to use `oracleAddress` cc @DhairyaSethi

tests/Base.t.sol
```diff
@@ -183,20 +185,15 @@ abstract contract Base is Test {
   }
 
   function setUpRoles() internal virtual {
-    // Hub Admin has default admin role
-    vm.startPrank(HUB_ADMIN);
+    vm.startPrank(ADMIN);
     // Grant roles
+    accessManager.grantRole(Roles.HUB_ADMIN_ROLE, ADMIN, 0);
+    accessManager.grantRole(Roles.SPOKE_ADMIN_ROLE, ADMIN, 0);
+    accessManager.grantRole(Roles.INTEREST_RATE_CONTROLLER_ROLE, ADMIN, 0);
```
yan-man: nit: can just have `INTEREST_RATE_ADMIN_ROLE` for consistency until we figure out naming but nbd

src/contracts/LiquidityHub.sol
```diff
@@ -119,10 +119,11 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   ) internal {
     _spokes[assetId][spoke].config = DataTypes.SpokeConfig({
       drawCap: config.drawCap,
-      supplyCap: config.supplyCap
+      supplyCap: config.supplyCap,
+      active: config.active
     });
 
-    emit SpokeConfigUpdated(assetId, spoke, config.drawCap, config.supplyCap);
+    emit SpokeConfigUpdated(assetId, spoke, config.drawCap, config.supplyCap, config.active);
```
yan-man: hm should we enforce anything on turning spoke inactive? ie having 0 outstanding supplies or something

tests/unit/Spoke/Spoke.Access.t.sol
```diff
@@ -1,31 +1,76 @@
...
+      address(HUB_ADMIN),
+      address(hub),
+      ILiquidityHub.addAsset.selector
+    );
+    assertTrue(immediate, 'Authority should allow immediate call');
+    assertEq(delay, 0, 'Authority should have no delay');
+  }
+
+  /*
+  // TODO: How? and Needed?
```
yan-man: here I think Dhairya was just saying to check that using access managed without time delay doesnt add gas overhead (ie by escaping early or whatnot) vs if the access managed code didn't have any time component-related code at all. don't think we need it as a separate gas snapshot, imo can just run some local tests or check the OZ contract. don't think it's a big deal either way

tests/unit/Spoke/Spoke.Access.t.sol
```diff
@@ -87,5 +132,231 @@ contract SpokeAccessTest is SpokeBase {
...
+        active: true,
+        frozen: false,
+        paused: false,
+        decimals: 18,
+        liquidityFee: 0,
+        irStrategy: irStrategy
+      }),
+      address(tokenList.dai)
+    );
+    uint256 assetAId = hub.assetCount() - 1; // Asset A Id
```
yan-man: hm if dai was the last asset added shouldn't this id be of DAI instead of assetA

tests/unit/Spoke/Spoke.Access.t.sol
```diff
@@ -87,5 +132,231 @@ contract SpokeAccessTest is SpokeBase {
...
+      abi.encodeWithSelector(IAccessManaged.AccessManagedUnauthorized.selector, address(this))
+    );
+    hub.updateAssetFees(daiAssetId, address(0), 0);
+
+    // Hub Admin can update asset fees
+    vm.prank(HUB_ADMIN);
+    hub.updateAssetFees(daiAssetId, address(0), 0);
+  }
+
+  function test_interest_rate_access() public {
```
yan-man:
```suggestion
  function test_ setInterestRateData_access() public {
```
?

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
```diff
@@ -34,4 +36,10 @@ contract SpokeUpdateUserRiskPremium is SpokeBase {
     vm.prank(caller);
     spoke1.updateUserRiskPremium(alice);
   }
+
+  function _isGovernor(address caller, ISpoke spoke) internal view returns (bool) {
+    IAccessManager accessManager = IAccessManager(IAccessManaged(address(spoke)).authority());
+    (bool result, ) = accessManager.hasRole(Roles.GOVERNOR_ROLE, caller);
```
yan-man:
```suggestion
    (bool result, ) = IAccessManager(IAccessManaged(address(spoke)).authority()).hasRole(Roles.GOVERNOR_ROLE, caller);
```

given it's just used once but nbd

src/contracts/Spoke.sol
```diff
@@ -719,9 +720,9 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     return userPosition.baseDrawnShares > 0;
   }
 
-  function _isGovernor(address user) internal view returns (bool) {
-    AccessManager accessManager = AccessManager(authority());
-    (bool result, ) = accessManager.hasRole(Roles.DEFAULT_ADMIN_ROLE, user);
+  function _isGovernor(address caller) internal view returns (bool) {
+    IAccessManager accessManager = IAccessManager(authority());
+    (bool result, ) = accessManager.hasRole(Roles.GOVERNOR_ROLE, caller);
```
yan-man:
```suggestion
    (bool result, ) = IAccessManager(authority()).hasRole(Roles.GOVERNOR_ROLE, caller);
```

given it's only used once but nbd

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Yeah, also this happened in a different PR, #443

src/contracts/Spoke.sol
CheyenneAtapour: You mean for constructor arguments, right? I think it's fine as is, I also vaguely remember talking to @miguelmtzinf about the use of `_` or similar, but the current way seems to be how V3 does it

src/contracts/LiquidityHub.sol
CheyenneAtapour: Yeah, I think we should do that. Actually we had a task to remove spokes that still hasn't been tackled. We could save for there, because probably there's other logic we'd like to handle properly: https://github.com/orgs/aave/projects/14/views/1?filterQuery=remove+spoke&pane=issue&itemId=107269398&issue=aave%7Caave-v4%7C358

tests/unit/Spoke/Spoke.Access.t.sol
CheyenneAtapour: Yeah, this was my attempt at it, but didn't seem to be working well. Talking with @DhairyaSethi to determine best way to proceed here

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
CheyenneAtapour: esp since this is just in a test and we probably care more about readability than efficiency, I prefer to keep as is

src/contracts/Spoke.sol
yan-man: yeah constructor args or general function args too

tests/unit/Spoke/Spoke.Access.t.sol
yan-man: maybe can just create an issue for it as a lower priority opt

src/contracts/LiquidityHub.sol
```diff
@@ -145,7 +153,11 @@ contract LiquidityHub is ILiquidityHub {
         _updateSpokeConfig(
           assetId,
           oldFeeReceiver,
-          DataTypes.SpokeConfig({supplyCap: 0, drawCap: 0})
+          DataTypes.SpokeConfig({
+            supplyCap: 0,
+            drawCap: 0,
+            active: _spokes[assetId][oldFeeReceiver].config.active
```
```suggestion
            active: false
```

src/contracts/LiquidityHub.sol
```diff
@@ -114,16 +119,19 @@ contract LiquidityHub is ILiquidityHub {
   ) internal {
     _spokes[assetId][spoke].config = DataTypes.SpokeConfig({
       drawCap: config.drawCap,
-      supplyCap: config.supplyCap
+      supplyCap: config.supplyCap,
+      active: config.active
     });
 
-    emit SpokeConfigUpdated(assetId, spoke, config.drawCap, config.supplyCap);
+    emit SpokeConfigUpdated(assetId, spoke, config.drawCap, config.supplyCap, config.active);
```
DhairyaSethi: let's just emit config at this point

src/libraries/types/DataTypes.sol
```diff
@@ -34,6 +34,7 @@ library DataTypes {
   struct SpokeConfig {
     uint256 drawCap;
     uint256 supplyCap;
+    bool active; // TODO: Opt to remove this and check that spoke is calling hub functions another way
```
miguelmtzinf: we need to stop adding todo comments in the code, and creating new gh tickets intead. Can we start now pls?

src/contracts/Spoke.sol
```diff
@@ -719,6 +720,11 @@ contract Spoke is ISpoke, Multicall {
     return userPosition.baseDrawnShares > 0;
   }
 
+  function _isGovernor(address caller) internal view returns (bool) {
+    (bool result, ) = IAccessManager(authority()).hasRole(Roles.GOVERNOR_ROLE, caller);
```
miguelmtzinf: I d prefer to use the simple `restricted` modifier instead of custom ones. This allow us to be completely decouple from Roles library, at the cost of extra functions (we need to split `updateUserRiskPremium` into two now).
The great benefit of this approach is being able to change `restricted` conditions dynamically. We loose this benefit If we use an static role I think.

src/contracts/Spoke.sol
```diff
@@ -5,19 +5,23 @@ import {Multicall} from 'src/misc/Multicall.sol';
 
 import {SafeERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
 import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
+import {AccessManaged} from 'src/dependencies/openzeppelin/AccessManaged.sol';
+import {AccessManager} from 'src/dependencies/openzeppelin/AccessManager.sol';
+import {IAccessManager} from 'src/dependencies/openzeppelin/IAccessManager.sol';
```
miguelmtzinf: do we need the 3 of them?

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -29,14 +30,13 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
   /**
    * @dev Constructor.
    */
-  constructor() {
-    /// TODO: Access Control; Store authorized address to set interest rate data
-  }
+  constructor(address accessManager) AccessManaged(accessManager) {}
```
miguelmtzinf: probably better to depend on Hub here? `onlyHub`

src/contracts/Spoke.sol
miguelmtzinf: I prefer to use `variable_`. I was thinking about doing it in a refactor/clean PR.
(fyi, the `TreasurySpoke` contract is already following this practice).

src/contracts/Spoke.sol
CheyenneAtapour: Just two, removed the other

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Done

src/contracts/Spoke.sol
CheyenneAtapour: As discussed today, changed the role name

src/libraries/types/DataTypes.sol
CheyenneAtapour: As discussed, we need the flag instead of relying on caps, so removed this comment

src/contracts/LiquidityHub.sol

src/contracts/LiquidityHub.sol
CheyenneAtapour: Done - was potentially concerned about there still being supplied shares on this spoke, but that must be handled before changing fee receiver

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -26,17 +26,18 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
...
-    /// TODO: Access Control; Store authorized address to set interest rate data
+  constructor(address liquidityHub) {
+    _liquidityHub = liquidityHub;
   }
 
   /// @inheritdoc IAssetInterestRateStrategy
   function setInterestRateData(uint256 assetId, InterestRateData calldata rateData) external {
-    /// TODO: Access Control; Only authorized address can set interest rate data
-
+    require(msg.sender == _liquidityHub, OnlyLiquidityHub());
```
miguelmtzinf:
```suggestion
    require(_liquidityHub == msg.sender, OnlyLiquidityHub());
```

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -26,17 +26,18 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
   /// @dev Map of assetId and their interest rate data (assetId => interestRateData)
   mapping(uint256 assetId => InterestRateData data) internal _interestRateData;
 
+  address internal _liquidityHub;
```
miguelmtzinf:
```suggestion
  address immutable public LIQUIDITY_HUB;
```
wdyt

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -26,17 +26,18 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
...
+
   /**
    * @dev Constructor.
    */
-  constructor() {
-    /// TODO: Access Control; Store authorized address to set interest rate data
+  constructor(address liquidityHub) {
```
miguelmtzinf:
```suggestion
  constructor(address liquidityHub_) {
```

src/contracts/LiquidityHub.sol
```diff
@@ -27,12 +28,14 @@ contract LiquidityHub is ILiquidityHub {
   IERC20[] public assetsList; // TODO: Check if Enumerable or Set makes more sense
   uint256 public assetCount;
 
+  constructor(address accessManager) AccessManaged(accessManager) {}
```
miguelmtzinf: let's add docs pls

miguelmtzinf: do we wanna use `authority` name instead of `accessManager` ? just wondering

src/contracts/LiquidityHub.sol
```diff
@@ -180,8 +196,6 @@ contract LiquidityHub is ILiquidityHub {
 
   /// @inheritdoc ILiquidityHub
   function add(uint256 assetId, uint256 amount, address from) external returns (uint256) {
-    // TODO: authorization - only spokes
```
miguelmtzinf, line 131: i wonder if we should fail earlier and do the active check here instead than in validate fn

src/contracts/Spoke.sol
```diff
@@ -719,6 +719,11 @@ contract Spoke is ISpoke, Multicall {
     return userPosition.baseDrawnShares > 0;
   }
 
+  function _isUserRpUpdater(address caller) internal view returns (bool) {
```
```suggestion
  function _isUserRiskPremiumUpdater(address caller) internal view returns (bool) {
```
however, i think smt like `hasFunctionRole` would be more generic and extendable... super ugly though: `hasUpdateUserRiskPremiumRole`

I still prefer the split into 2 function lol. Feels way cleaner to me

miguelmtzinf: fyi: https://forum.openzeppelin.com/t/complex-conditional-logic-for-function-access-with-accessmanaged/44414

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -26,17 +26,18 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
...
+    _liquidityHub = liquidityHub;
   }
 
   /// @inheritdoc IAssetInterestRateStrategy
   function setInterestRateData(uint256 assetId, InterestRateData calldata rateData) external {
```
miguelmtzinf: should we add the function in the Hub that calls this?

tests/unit/Spoke/Spoke.MultipleHub.t.sol
```diff
@@ -88,6 +92,40 @@ contract SpokeMultipleHubTest is SpokeBase {
     tokenList.dai.approve(address(hub2), type(uint256).max);
     tokenList.dai.approve(address(hub3), type(uint256).max);
     vm.stopPrank();
+
+    configureRoles();
+  }
+
+  function configureRoles() internal {
```
miguelmtzinf: i think this should be part of the fixture?

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
```diff
@@ -34,4 +36,10 @@ contract SpokeUpdateUserRiskPremium is SpokeBase {
     vm.prank(caller);
     spoke1.updateUserRiskPremium(alice);
   }
+
+  function _isUserRpUpdater(address caller, ISpoke spoke) internal view returns (bool) {
+    IAccessManager accessManager = IAccessManager(IAccessManaged(address(spoke)).authority());
```
miguelmtzinf: this chain cast sucks

src/interfaces/ILiquidityHub.sol
miguelmtzinf, line 1: should we inherit from `IAccessManaged` ? or create specific functions to query `authority`

tests/unit/Spoke/Spoke.Access.t.sol
```diff
@@ -0,0 +1,343 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {AuthorityUtils} from 'src/dependencies/openzeppelin/AuthorityUtils.sol';
+import {IAccessManaged} from 'src/dependencies/openzeppelin/IAccessManaged.sol';
+import {Context} from 'src/dependencies/openzeppelin/Context.sol';
+import {Roles} from 'src/libraries/types/Roles.sol';
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+contract SpokeAccessTest is SpokeBase, Context {
```
miguelmtzinf: move all imports to Base. Why we need `Context` as base?

tests/unit/Spoke/Spoke.Access.t.sol
```diff
@@ -0,0 +1,343 @@
...
+    // Spoke admin can call updateDynamicReserveConfig
+    vm.prank(SPOKE_ADMIN);
+    spoke1.updateDynamicReserveConfig(
+      _daiReserveId(spoke1),
+      DataTypes.DynamicReserveConfig({collateralFactor: 75_00})
+    );
+  }
+
+  /// @dev Test showing that restricted functions on hub can only be called by hub admin.
+  function test_liquidity_hub_admin_access() public {
```
miguelmtzinf: i d move this and the next one to an specific file: `LiquidityHub.Access.t.sol`

src/dependencies/openzeppelin/AuthorityUtils.sol
miguelmtzinf: yes please

src/libraries/types/Roles.sol
miguelmtzinf: I'm good keeping it as a library, but as said in other comments, I d prefer not couple to any other function (updateUserRp role)

src/libraries/types/Roles.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+library Roles {
+  uint64 public constant DEFAULT_ADMIN_ROLE = 0;
+  uint64 public constant HUB_ADMIN_ROLE = 1;
+  uint64 public constant SPOKE_ADMIN_ROLE = 2;
+  uint64 public constant USER_RP_UPDATER_ROLE = 3;
+  uint64 public constant INTEREST_RATE_ADMIN_ROLE = 4;
```
miguelmtzinf: this can be the same as HUB_ADMIN_ROLE imo

miguelmtzinf: question, now all restricted functions are under HUB_ADMIN_ROLE. What would be the process for more complex config? e.g. half of fns controlled by A entity, another half by B
A test showcasing this would be great as well

src/contracts/LiquidityHub.sol
miguelmtzinf: chiming in to say that we may do not want to enforce, if we want to use it as a temporarily measure? e.g. we stop operations from Spoke 1 because there is a potential vulnerability

src/contracts/AssetInterestRateStrategy.sol
avniculae: or just HUB? TreasurySpoke named it HUB, and I think we want to rename LiquidityHub to Hub in the future anyways right?

src/contracts/AssetInterestRateStrategy.sol
avniculae: or just `hub`?

src/contracts/AssetInterestRateStrategy.sol
avniculae: I actually think `msg.sender` first makes more sense, as it represents the value against which we enforce the condition (of being the hub), nbd though

src/dependencies/openzeppelin/AccessManager.sol
```diff
@@ -0,0 +1,775 @@
...
+
+  /// @inheritdoc IAccessManager
+  function revokeRole(uint64 roleId, address account) public virtual onlyAuthorized {
+    _revokeRole(roleId, account);
+  }
+
+  /// @inheritdoc IAccessManager
+  function renounceRole(uint64 roleId, address callerConfirmation) public virtual {
+    if (callerConfirmation != _msgSender()) {
+      revert AccessManagerBadConfirmation();
```
avniculae, line 257: why do you think they added this here?

src/dependencies/openzeppelin/AuthorityUtils.sol
avniculae: you mean the former type (with delay), right? 

from AccessManager.sol
```
* NOTE: This contract implements a form of the {IAuthority} interface, but {canCall} has additional return data so it
 * doesn't inherit `IAuthority`. It is however compatible with the `IAuthority` interface since the first 32 bytes of
 * the return data are a boolean as expected by that interface.
```

src/dependencies/openzeppelin/AccessManager.sol
```diff
@@ -0,0 +1,775 @@
...
+
+    bool newMember = _roles[roleId].members[account].since == 0;
+    uint48 since;
+
+    if (newMember) {
+      since = Time.timestamp() + grantDelay;
+      _roles[roleId].members[account] = Access({since: since, delay: executionDelay.toDelay()});
+    } else {
+      // No setback here. Value can be reset by doing revoke + grant, effectively allowing the admin to perform
+      // any change to the execution delay within the duration of the role admin delay.
```
avniculae, line 300: I am not sure I understand this comment here. This `else` flow does not modify `_roles[roleId].members[account].since`, which means that the account remains a member but has its delay modified. However, this seems potentially more advantageous for the `account` than doing revoke + grant, bc when it has to grant again the account will become a member at `Time.timestamp() + grantDelay`. Also, the admin cannot change the grant delay instantaneously

avniculae: Or does the comment not suggest that these 2 approaches are equivalent?

src/dependencies/openzeppelin/AuthorityUtils.sol
CheyenneAtapour: We don't want to use the delay at all @avniculae 

I am unclear still how to best do the tests @DhairyaSethi is referencing

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Yeah, I agree @avniculae. What is the reasoning for this preference @miguelmtzinf ?

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: I'm fine with either hub or liquidity hub, just making it `LIQUIDITY_HUB` for now

src/contracts/Spoke.sol
CheyenneAtapour: ah, nice forum post!

It's strange to me still to duplicate the code and esp when the user can already increase their own risk premium, so makes more sense to me to just also allow a governor in that same condition rather than duplicating code

I implemented your suggested commit

tests/unit/Spoke/Spoke.MultipleHub.t.sol
CheyenneAtapour: Hmm, but I kinda like it as is, because it gets everything done after all deployments. Currently we have 2 separate fixtures for hub 2 and hub 3, so with that, I think this is cleaner as is

src/libraries/types/Roles.sol
CheyenneAtapour: I'm unsure what you mean by all restricted functions are _under_ `HUB_ADMIN_ROLE`. To clarify, in the current setup, the `HUB_ADMIN_ROLE` cannot call, for example, the functions which require `SPOKE_ADMIN_ROLE`, like `Spoke.updateLiquidationConfig`

I can clarify this is the case in some dedicated tests

Removed that role

src/contracts/LiquidityHub.sol
CheyenneAtapour: Yeah, that's fine because it's more generic. I was thinking `accessManager` would be nice because it does need to implement the `accessManager` interface, but I went with `authority`, and mentioned the implementation requirement in the doc

src/contracts/LiquidityHub.sol
CheyenneAtapour: Do you think that's worth it? I feel nicer to keep the functions code smaller and abstract it to validate function. Doesn't fail much earlier and to me makes sense in a `validate` function

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Done, but instead of making basically a wrapper function on the hub to call this, why don't we just add a role directly to this function on the `AssetInterestRateStrategy` contract itself, like it was in this PR before?

Seems like this is just adding extra code to liquidity hub

tests/unit/Spoke/Spoke.Access.t.sol
CheyenneAtapour: Apologies, that was leftover from trying to make a test to see if `_authority()` was a 0 gas cost abstraction compared to the likes of `_msgSender()`

Done

tests/unit/Spoke/Spoke.Access.t.sol
CheyenneAtapour: Done

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
CheyenneAtapour: Fixed

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Done, went with `LIQUIDITY_HUB` for now

src/contracts/LiquidityHub.sol
```diff
@@ -174,14 +196,21 @@ contract LiquidityHub is ILiquidityHub {
     }
   }
 
+  /// @inheritdoc ILiquidityHub
+  function setInterestRateData(
+    uint256 assetId,
+    IAssetInterestRateStrategy.InterestRateData calldata rateData,
+    IAssetInterestRateStrategy irStrategy
+  ) external restricted {
+    irStrategy.setInterestRateData(assetId, rateData);
```
avniculae: we need to first accrue before setting the new interest rate data

src/contracts/AssetInterestRateStrategy.sol
avniculae: this is the reason https://github.com/aave/aave-v4/pull/455/files#r2168413416

src/contracts/AssetInterestRateStrategy.sol
miguelmtzinf: I believe we are using this pattern across the code: stored vs input, existing vs new

src/contracts/AssetInterestRateStrategy.sol
miguelmtzinf: i d like to change all to HUB honestly. We can leave as it is and rename later all at once

src/contracts/LiquidityHub.sol
```diff
@@ -27,12 +29,19 @@ contract LiquidityHub is ILiquidityHub {
   IERC20[] public assetsList; // TODO: Check if Enumerable or Set makes more sense
   uint256 public assetCount;
 
+  /*
+   * @notice Initializes the contract with access control.
+   * @param authority The address of the authority contract which manages permissions.
+   * @dev The authority should implement the AccessManaged interface to control access.
```
```suggestion
   * @dev The authority should implement the AccessManaged interface to control access.
   * @param authority The address of the authority contract which manages permissions.
```

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -23,20 +23,21 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
   /// @inheritdoc IAssetInterestRateStrategy
   uint256 public constant MAX_OPTIMAL_RATIO = 99_00; // 99.00% in BPS
 
+  address public immutable LIQUIDITY_HUB;
```
miguelmtzinf, line 27: add to the interface pls, and inherit docs

src/contracts/LiquidityHub.sol
```diff
@@ -174,14 +196,21 @@ contract LiquidityHub is ILiquidityHub {
     }
   }
 
+  /// @inheritdoc ILiquidityHub
+  function setInterestRateData(
+    uint256 assetId,
+    IAssetInterestRateStrategy.InterestRateData calldata rateData,
+    IAssetInterestRateStrategy irStrategy
```
miguelmtzinf: why passing this?

src/contracts/Spoke.sol
```diff
@@ -39,7 +42,7 @@ contract Spoke is ISpoke, Multicall {
   uint256[] public reservesList; // todo: rm, not needed
   uint256 public reserveCount;
 
-  constructor(address oracleAddress) {
+  constructor(address oracleAddress, address accessManager) AccessManaged(accessManager) {
```
```suggestion
  constructor(address oracleAddress, address authority) AccessManaged(authority) {
```
add docs pls

src/interfaces/ILiquidityHub.sol
```diff
@@ -96,6 +98,19 @@ interface ILiquidityHub {
     DataTypes.SpokeConfig memory config
   ) external;
 
+  /**
+   * @notice Updates the interest rate strategy for a specified asset.
+   * @dev Only callable by the liquidity hub admin.
```
miguelmtzinf: none of the restricted fns have this. I d either remove it or make it more general since roles/permissions can change dynamically.

tests/Base.t.sol
```diff
@@ -88,11 +93,13 @@ abstract contract Base is Test {
   address internal carol = makeAddr('carol');
   address internal derl = makeAddr('derl');
 
+  address internal ADMIN = makeAddr('ADMIN');
   address internal HUB_ADMIN = makeAddr('HUB_ADMIN');
   address internal SPOKE_ADMIN = makeAddr('SPOKE_ADMIN');
   address internal TREASURY_ADMIN = makeAddr('TREASURY_ADMIN');
   address internal TREASURY = makeAddr('TREASURY');
   address internal LIQUIDATOR = makeAddr('LIQUIDATOR');
+  address internal GOVERNOR = makeAddr('GOVERNOR');
```
miguelmtzinf: i d suggest to use `SPOKE_ADMIN` instead

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
```diff
@@ -0,0 +1,133 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/Base.t.sol';
+
+contract LiquidityHubAccessTest is Base {
+  function setUp() public override {
```
miguelmtzinf: why do we need this override? cannot we use the default setup?

miguelmtzinf: why not inheriting from `LiquidityHubBase`

src/libraries/types/Roles.sol
miguelmtzinf:
```
    hubSelectors[0] = ILiquidityHub.addAsset.selector;
    hubSelectors[1] = ILiquidityHub.updateAssetConfig.selector;
    hubSelectors[2] = ILiquidityHub.addSpoke.selector;
    hubSelectors[3] = ILiquidityHub.addSpokes.selector;
    hubSelectors[4] = ILiquidityHub.updateSpokeConfig.selector;
    hubSelectors[5] = ILiquidityHub.updateAssetFees.selector;
    hubSelectors[6] = ILiquidityHub.setInterestRateData.selector;
```
These functions can be called by `HUB_ADMIN_ROLE`. What happens if we decide that `setInterestRateData` must be called by a different entity/role. Can you write a test for that please?

src/dependencies/openzeppelin/AccessManager.sol
miguelmtzinf: you mean `callerConfirmation`? I think it's just a way to confirm they really want to renounce, to avoid dumb errors

src/contracts/LiquidityHub.sol
```diff
@@ -114,16 +125,19 @@ contract LiquidityHub is ILiquidityHub {
...
 
   /// @inheritdoc ILiquidityHub
-  function updateAssetFees(uint256 assetId, address feeReceiver, uint256 liquidityFee) public {
-    // TODO: AccessControl
-
+  function updateAssetFees(
+    uint256 assetId,
+    address feeReceiver,
+    uint256 liquidityFee
+  ) public restricted {
```
```suggestion
  ) external restricted {
```

tests/unit/Spoke/Spoke.MultipleHub.t.sol
miguelmtzinf: I do think each hub/spoke should have its own accessManager, and the configuration is part of the deployment process (eg fixture)

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
```diff
@@ -34,4 +36,10 @@ contract SpokeUpdateUserRiskPremium is SpokeBase {
     vm.prank(caller);
     spoke1.updateUserRiskPremium(alice);
   }
+
+  function _isUserRiskPremiumUpdater(address caller, ISpoke spoke) internal view returns (bool) {
+    IAccessManager accessManager = IAccessManager(spoke.authority());
```
miguelmtzinf: now that we re inheriting `IAccessManager` in `ILiquidityHub`, we can use `ILiquidityHub` here

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
```diff
@@ -2,6 +2,8 @@
 pragma solidity ^0.8.0;
 
 import 'tests/unit/Spoke/SpokeBase.t.sol';
+import {IAccessManager} from 'src/dependencies/openzeppelin/IAccessManager.sol';
+import {IAccessManaged} from 'src/dependencies/openzeppelin/IAccessManaged.sol';
```
miguelmtzinf: if needed, move to `Base`

src/contracts/LiquidityHub.sol
```diff
@@ -174,14 +196,21 @@ contract LiquidityHub is ILiquidityHub {
     }
   }
 
+  /// @inheritdoc ILiquidityHub
+  function setInterestRateData(
+    uint256 assetId,
+    IAssetInterestRateStrategy.InterestRateData calldata rateData,
```
DhairyaSethi: can we not use enforce the struct but instead pass rateData as bytes which is decoded on the IR strategy into the struct such that we're not limited to IR strats in the future which use different data

src/contracts/Spoke.sol
```diff
@@ -719,6 +722,11 @@ contract Spoke is ISpoke, Multicall {
     return userPosition.baseDrawnShares > 0;
   }
 
+  function _isUserRiskPremiumUpdater(address caller) internal view returns (bool) {
```
```suggestion
  function _canUpdateRiskPremium(address caller) internal view returns (bool) {
```

DhairyaSethi: or canForceUpdateRp or something

tests/unit/Spoke/Spoke.MultipleHub.t.sol
CheyenneAtapour: What about for the canonical hub and the 3 spokes, should we leave those as is, or also use a different access manager contract for each? And should we also split that into different fixtures too?

CheyenneAtapour: To further clarify, should each hub itself have it's own unique accessManager contract, and same with each spoke?

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
CheyenneAtapour: we are not inheriting `IAccessManager` in `ILiquidityHub`, rather `IAccessManaged`. Also, I think this is clearer to understand by using the more explicit interface

src/contracts/Spoke.sol
CheyenneAtapour: @miguelmtzinf chose the current name, but yeah I like this suggestion. I also like keeping `user` there to make it clear what risk premium it is. But since we don't really use it outside of this context, maybe we can omit `user`

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -23,20 +23,23 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
...
+    LIQUIDITY_HUB = liquidityHub_;
   }
 
   /// @inheritdoc IAssetInterestRateStrategy
-  function setInterestRateData(uint256 assetId, InterestRateData calldata rateData) external {
-    /// TODO: Access Control; Only authorized address can set interest rate data
-
+  function setInterestRateData(uint256 assetId, bytes calldata data) external {
+    require(LIQUIDITY_HUB == msg.sender, OnlyLiquidityHub());
+    InterestRateData memory rateData = _decodeInterestRateData(data);
```
DhairyaSethi:
```suggestion
    InterestRateData memory rateData = abi.decode(data, (InterestRateData));
```
this doesn't need a separate method

tests/Base.t.sol
```diff
@@ -624,60 +674,74 @@ abstract contract Base is Test {
...
-        variableRateSlope1: 5_00, // 5.00%
-        variableRateSlope2: 5_00 // 5.00%
-      })
+      _encodeInterestRateData(
+        IAssetInterestRateStrategy.InterestRateData({
+          optimalUsageRatio: 90_00, // 90.00%
+          baseVariableBorrowRate: 5_00, // 5.00%
+          variableRateSlope1: 5_00, // 5.00%
+          variableRateSlope2: 5_00 // 5.00%
+        })
```
```suggestion
      abi.encode(
        IAssetInterestRateStrategy.InterestRateData({
          optimalUsageRatio: 90_00, // 90.00%
          baseVariableBorrowRate: 5_00, // 5.00%
          variableRateSlope1: 5_00, // 5.00%
          variableRateSlope2: 5_00 // 5.00%
        })
```
similarly these can all just be abi.encodes

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Done

src/libraries/types/Roles.sol
CheyenneAtapour: Done

src/contracts/LiquidityHub.sol

src/contracts/LiquidityHub.sol

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Done

src/contracts/LiquidityHub.sol
CheyenneAtapour: Originally was to make it cleaner than casting from asset config, but it's fixed now and this argument removed

src/contracts/Spoke.sol

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: Done - removed

tests/Base.t.sol

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
CheyenneAtapour: Fixed

src/contracts/LiquidityHub.sol

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
CheyenneAtapour: Done

src/contracts/LiquidityHub.sol

src/contracts/LiquidityHub.sol
```diff
@@ -174,14 +196,19 @@ contract LiquidityHub is ILiquidityHub {
     }
   }
 
+  /// @inheritdoc ILiquidityHub
+  function setInterestRateData(uint256 assetId, bytes calldata data) external restricted {
```
miguelmtzinf, line 134: We still have coupling in `updateBorrowRate` via `AssetLogic`...
https://github.com/aave/aave-v4/issues/481

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
```diff
@@ -130,4 +125,36 @@ contract LiquidityHubAccessTest is Base {
...
+        baseVariableBorrowRate: 100_00, // 100.00% in BPS
+        variableRateSlope1: 200_00, // 200.00% in BPS
+        variableRateSlope2: 300_00 // 300.00% in BPS
+      });
+
+    // Change the role responsible for setting interest rate data on the hub
+    bytes4[] memory hubSelectors = new bytes4[](1);
+    hubSelectors[0] = ILiquidityHub.setInterestRateData.selector;
+    vm.prank(ADMIN);
+    accessManager.setTargetFunctionRole(address(hub), hubSelectors, Roles.DEFAULT_ADMIN_ROLE);
```
miguelmtzinf, line 111: so this means a function cannot have multiple roles at the same time, right? So if we wanna be completely dynamic we need to have 1 role per fn? or just 1 role for all and change it later if needed (would require migrating all addresses granted with old role to the new role right).
Hope I am making sense, can you think about these scenarios please?

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
miguelmtzinf: coolio, i d showcase we can access `authority` via `hub` but strong opinion

src/contracts/Spoke.sol
miguelmtzinf: you know i do not like this, but if i were to keep it, i d make it general: `hasUpdateUserRiskPremiumRole`

src/contracts/Spoke.sol
```diff
@@ -39,7 +42,13 @@ contract Spoke is ISpoke, Multicall {
   uint256[] public reservesList; // todo: rm, not needed
   uint256 public reserveCount;
 
-  constructor(address oracleAddress) {
+  /*
+   * @notice Initializes the contract with access control.
```
```suggestion
   * @dev Constructor.
```

pls, we are using dev across code

DhairyaSethi: we **are** using notice on natspec (?)

tests/Base.t.sol
```diff
@@ -151,29 +158,71 @@ abstract contract Base is Test {
...
+  }
+
+  function setUpRoles(
+    ILiquidityHub hub,
+    ISpoke spoke,
+    IAccessManager accessManager
+  ) internal virtual {
+    vm.startPrank(ADMIN);
+    // Grant roles
+    accessManager.grantRole(Roles.HUB_ADMIN_ROLE, ADMIN, 0);
```
DhairyaSethi, line 199: just a quick comment or named args here to immediately indicate 0 is the delay would be very appreciated

src/contracts/LiquidityHub.sol
CheyenneAtapour: Should we tackle later, or determine the fix and implement for this PR?

src/contracts/Spoke.sol
CheyenneAtapour: I think @miguelmtzinf means for `constructor` specifically, and is indeed the case for `TreasurySpoke` and `AssetInterestRateStrategy`

Done

tests/Base.t.sol
CheyenneAtapour: Done

tests/Base.t.sol

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Done

src/contracts/Spoke.sol

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
CheyenneAtapour: Done - Added a quick test in liquidity hub access tests

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
CheyenneAtapour: Yes, each selector can be called by exactly one role, but addresses can have multiple roles. I can add some tests to cover the scenario you are describing and perhaps related ones.

I just added a test showcasing migrating the responsibility of one role to another one, that also showcases users can have multiple roles at once, and we can also revoke the roles.

tests/unit/Spoke/Spoke.MultipleHub.t.sol
miguelmtzinf: For tests I was thinking we can use 1 authority contract for each Hub/spoke pair. I think it's the way it's implemented now right?

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
miguelmtzinf, line 1: Can you add a simple test case in Hub and Spoke where you update the authority please? the new one can have a very basic config, just to try that after authority update we can still call restricted fns

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
```diff
@@ -160,6 +160,108 @@ contract LiquidityHubAccessTest is LiquidityHubBase {
...
+    vm.stopPrank();
+
+    vm.prank(alice);
+    hub.setInterestRateData(daiAssetId, encodedIrData);
+    vm.prank(bob);
+    hub.setInterestRateData(daiAssetId, encodedIrData);
+    vm.prank(carol);
+    hub.setInterestRateData(daiAssetId, encodedIrData);
+
+    // Now, we change the role responsible for setting interest rate data to SPOKE_ADMIN role.
```
miguelmtzinf: I d suggest to create a new role specifically for this, but nbd.
just because I think it's cleaner. (we can add it here, internally, not in Roles lib)

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
```diff
@@ -160,6 +160,108 @@ contract LiquidityHubAccessTest is LiquidityHubBase {
...
+    vm.prank(alice);
+    hub.setInterestRateData(daiAssetId, encodedIrData);
+    vm.prank(bob);
+    hub.setInterestRateData(daiAssetId, encodedIrData);
+    vm.prank(carol);
+    hub.setInterestRateData(daiAssetId, encodedIrData);
+
+    // Alice, Bob, and Carol currently have both HUB_ADMIN and SPOKE_ADMIN roles.
+    IAccessManager accessManager = IAccessManager(hub.authority());
+    (bool hasHubAdminRoleAlice, ) = accessManager.hasRole(Roles.HUB_ADMIN_ROLE, alice);
```
miguelmtzinf: do you wanna build a helper real quick? smt like `hasRole(address authority, bytes32 role, address user) returns bool`

tests/unit/Spoke/Spoke.MultipleHub.t.sol
CheyenneAtapour: Currently just each unique _hub_ has it's own unique authority contract. The spokes 1, 2 and 3 use the same one as the canonical hub. There's not really a notion of making a unique one per pair of hub/spoke because each is their own separate contract. 

E.g. if spoke2 has authority 2 and hub 1 has authority 1, it doesn't matter that they are interacting with each other, and doesn't make sense to have an authority contract just for this specific pair, and then for example a different one with spoke3 with authority 3 and hub 1 with authority 1. 

Hope it's clear that it doesn't make sense per pair, just per contract

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
CheyenneAtapour: Done

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
CheyenneAtapour: Done

tests/unit/LiquidityHub/LiquidityHub.Access.t.sol
CheyenneAtapour: Done

src/contracts/Spoke.sol
```diff
@@ -392,8 +396,10 @@ contract Spoke is ISpoke, Multicall {
   function updateUserRiskPremium(address user) external {
     (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
     bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
-    // todo allow authorized caller to increase as well
-    require(msg.sender == user || !premiumIncrease, Unauthorized());
+    // check permissions if premium increases or not called by user
+    if (premiumIncrease || user != msg.sender) {
```
```suggestion
    // check permissions if premium increases and not called by user
    if (premiumIncrease && user != msg.sender) {
```
if premium has increased & caller is user then we don't need to check

src/contracts/LiquidityHub.sol
miguelmtzinf: later is fine

Issue #452, opened 2025-06-16
Add tests for 100% liquidity fee


miguelmtzinf:
Add tests covering the case of 100% liquidity fee, which means to all interest goes to treasury (lps don't see their assets growing).
This can be added to `SpokeaccrueLiquidityFeeTest` and `TreasurySpokeTest`

PR #458, opened 2025-06-17
feat: Add more equivalent fees calc in previewFeeShares

### LINKED ISSUE #457: Fees amount could potentially exceed interest growth if 100% fee and wrong rounding
Reference: https://github.com/aave/aave-v4/pull/375#discussion_r2148797956

https://github.com/aave/aave-v4/blob/d118321e0dd211bf77e2ffa1ee8bc564a39e6aa9/src/libraries/logic/AssetLogic.sol#L173-L187

```
  // this test shows that with 100% fee, we leave some dust in the pool
  function test_maxLiquidityFee() public {
    hub.updateAssetFees(daiAssetId, address(treasurySpoke), 100_00);
    Utils.supplyCollateral(spoke1, _daiReserveId(spoke1), alice, 100e18, alice);
    Utils.borrow(spoke1, _daiReserveId(spoke1), alice, 50e18, alice);
    skip(342 days);
    console.log('treasury amount', hub.getSpokeSuppliedAmount(daiAssetId, address(treasurySpoke)));
    console.log('alice accrued  ', spoke1.getUserTotalDebt(_daiReserveId(spoke1), alice) - 50e18);
  }
Logs:
  treasury amount 2810958904109589040
  alice accrued   2810958904109589042
```

Rationale of current code:
- `rayMulDown` because it's safer: leaving some donation/dust instead of bad debt
- `percentMulDown` because if the feeAmount is v small we don't care

---

Apart from that, it's true that the calculation of interest and premium accrual is different from the rest of the code. In fees we are using the following:
```
indexDelta = nextIndex - currentIndex
indexDelta * (debtShares + premiumShares)
```
while in the rest of the code we are using index accrual for baseDebt and individual premium accrual based on users' rp...

An alternative implementation for `previewFeeShares` is:
```
baseAccrual = indexDelta * baseShares 
premiumAccrual = nextIndex * premiumShares - offset 
```

miguelmtzinf:
Closes #457 
Closes https://github.com/aave/aave-v4/issues/452

### DIFF
--- a/tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
+++ b/tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
@@ -7,7 +7,7 @@ contract SpokeAccrueLiquidityFeeEdgeCasesTest is SpokeBase {
   uint256 public constant MAX_LIQUIDITY_FEE = 100_00;
 
   /// @dev Max liquidity fee with premium debt accrual
-  function test_accrueLiquidityFee_fuzz_maxLiquidityFee_with_premium() public {
+  function test_accrueLiquidityFee_maxLiquidityFee_with_premium() public {
     test_accrueLiquidityFee_fuzz_maxLiquidityFee_with_premium({
       reserveId: _daiReserveId(spoke1),
       borrowAmount: 500e18,


src/libraries/logic/AssetLogic.sol
```diff
@@ -168,32 +167,50 @@ library AssetLogic {
...
+    uint256 nextDrawnIndex
   ) internal view returns (uint256) {
     uint256 liquidityFee = asset.config.liquidityFee;
-    if (indexDelta == 0 || liquidityFee == 0) {
+    if (nextDrawnIndex == currentDrawnIndex || liquidityFee == 0) {
       return 0;
     }
     // liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee
-    uint256 feesAmount = indexDelta
+    uint256 feesAmount = (nextDrawnIndex - currentDrawnIndex)
```
miguelmtzinf: this and sanity check to be removed before merging

src/libraries/logic/AssetLogic.sol
```diff
@@ -168,32 +167,50 @@ library AssetLogic {
...
+
+    return
+      feesAmountActual.toSharesDown(
+        asset.totalSuppliedAssets() - feesAmountActual,
+        asset.suppliedShares
+      );
   }
 
+  error FeeDiff(uint256 feesAmount, uint256 feesAmountActual);
+  function absDiff(uint256 a, uint256 b) internal pure returns (uint256 c) {
```
yan-man: should we move this to Math lib?

DhairyaSethi: it's to be removed before merging

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
```diff
@@ -0,0 +1,110 @@
...
+      spoke1.getUserSuppliedAmount(reserveId, alice),
+      supplyAmount,
+      'alice does not earn anything'
+    );
+    assertEq(
+      spoke1.getUserSuppliedAmount(reserveId, bob),
+      supplyAmount2,
+      'bob does not earn anything'
+    );
+    assertLe(
```
yan-man, line 109: in single user case, finding that the accrued exactly matches the treasury accrued fees, but in multi-user case here, we still have same condition as before, where treasury accrued <= user accrued

src/libraries/logic/AssetLogic.sol
```diff
@@ -72,7 +72,7 @@ library AssetLogic {
   }
 
   function totalSuppliedShares(DataTypes.Asset storage asset) internal view returns (uint256) {
-    return asset.suppliedShares + asset.getFeeShares(asset.getDrawnIndex() - asset.baseDebtIndex);
+    return asset.suppliedShares + asset.getFeeShares(asset.baseDebtIndex, asset.getDrawnIndex());
```
DhairyaSethi: think the opposite order; ie nextIndex, currentIndex is more intuitive but nbd

src/libraries/logic/AssetLogic.sol
```diff
@@ -166,21 +166,23 @@ library AssetLogic {
...
       return 0;
     }
-    // liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee
-    uint256 feesAmount = indexDelta
-      .rayMulDown(asset.baseDrawnShares + asset.premiumDrawnShares)
-      .percentMulDown(liquidityFee);
+
+    uint256 feesAmount = ((nextDrawnIndex - currentDrawnIndex).rayMulUp(asset.baseDrawnShares) +
+      nextDrawnIndex.rayMulUp(asset.premiumDrawnShares) -
+      asset.premiumOffset).percentMulDown(liquidityFee);
```
DhairyaSethi: @yan-man why did you specifically decide to roundUp here? thought we agreed round down is safer. in any case, if we're certain through tests this doesn't lead to treasury being owed more than what's owed by users it's fine

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -24,22 +24,28 @@ contract TreasurySpokeTest is SpokeBase {
 
     vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, caller));
     vm.prank(caller);
-    treasurySpoke.supply(daiAssetId, 1, caller);
+    treasurySpoke.supply(_daiReserveId(_treasurySpoke()), 1, caller);
```
DhairyaSethi: these were fine before and are incorrect now, treasury spoke inputs are assetId, pls see natspec. tests should have failed, i think we've registered spokes 1-1 to assets luckily but that's a very dangerous assumption 

haha @miguelmtzinf i did say this was confusing

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
DhairyaSethi: yes as long as it isn't greater than. still a bit unsure of rounding up there 
can we also check the same test when all users repay

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
```diff
@@ -0,0 +1,110 @@
...
+      hub.getSpokeSuppliedAmount(spoke1.getReserve(reserveId).assetId, address(treasurySpoke)),
+      spoke1.getUserTotalDebt(reserveId, alice) -
+        borrowAmount +
+        spoke1.getUserTotalDebt(reserveId, bob) -
+        borrowAmount2,
```
DhairyaSethi, line 114: can we check this against spoke total debt

src/libraries/logic/AssetLogic.sol
miguelmtzinf: yeah, good point
cannot recall the specific reason, but agree generally better to round down here

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
```diff
@@ -0,0 +1,197 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+contract SpokeAccrueLiquidityFeeEdgeCasesTest is SpokeBase {
+  uint256 public constant MAX_LIQUIDITY_FEE = 100_00;
+
+  /// @dev Max liquidity fee with premium debt accrual
+  function test_accrueLiquidityFee_fuzz_maxLiquidityFee_with_premium() public {
```
avniculae: we don't fuzz anything here, I think we can remove `fuzz` from the name

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
```diff
@@ -0,0 +1,197 @@
...
+    assertApproxEqAbs(
+      hub.getSpokeSuppliedAmount(assetId, address(treasurySpoke)),
+      spoke1.getUserTotalDebt(reserveId, alice) - borrowAmount,
+      3,
+      'fees == total user accrued'
+    );
+    assertApproxEqAbs(
+      hub.getSpokeSuppliedAmount(assetId, address(treasurySpoke)),
+      hub.getSpokeTotalDebt(assetId, address(spoke1)) - borrowAmount,
+      3,
```
avniculae, line 61: what's the theoretical delta here? 1 share worth of assets? or maybe 2?

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
yan-man: decided to round down to be safe

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
yan-man: https://github.com/aave/aave-v4/pull/458/commits/8ec3a5c67d8f2466f3bef8699b060944ed2cbd5b

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
DhairyaSethi: unsure bc we underestimate twice by rounding down the debt, taking % of interest, i think 1 share worth of assets + 1 asset from both respectively

PR #462, opened 2025-06-18
feat: position manager

### LINKED ISSUE #360: Position Manager
Analyze implementation of a `Position Manager` to execute other functions on top (e.g. GHO soft liquidations)
- this contract could do anything with the user's collateral and debt
- need to add an address in DataTypes.UserPosition and authorize it
- however dangerous, it can be limited through governance (e.g. whitelist of position managers)

DhairyaSethi:
closes #360 

## Design
- Introduces `PositionManager` which can, with approval, modify user positions through the all user facing actions, ie: supply, withdraw, borrow, restore, useAsCollateral, & updateRiskPremium. **Note:** Only active position managers can act on user's behalf.
- Governance can toggle the active status of a position manager through `setPositionManager`.
- Users can grant or provoke explicit approval to modify any position in account through `setApprovalForPositionManager`. Users can only grant approval to active position managers for improved UX, but can revoke approval regardless of active status.
- All actions are introduced with an `onBehalfOf` parameter, indicating the address of the user whose position will be modified.
- Position manager can renounce approval on user position through `renouncePositionManagerRole`.
- WithdrawTo & BorrowTo functionality is removed to accommodate and streamline current interfaces, it's expected this functionality can be achieved through account abstraction.

## Notes
- The approved position manager check currently costs two lookups for each operation done on behalf of user due to the whitelist. 
- `InactivePositionManager` check on `setApprovalForPositionManager` check is only for UX and can be removed to save gas.
- `renouncePositionManagerRole` method can be removed since we only allow whitelisted position managers.
- `NoExplicitAmountToRepayOnBehalf` check is omitted since we only allow whitelisted position managers but can be introduced for sanity.

### Todo
- Gas snapshots for actions performed through position manager.

### DIFF
--- a/tests/unit/Spoke/Spoke.PositionManager.t.sol
+++ b/tests/unit/Spoke/Spoke.PositionManager.t.sol
@@ -29,7 +29,7 @@ contract SpokePositionManagerTest is SpokeBase {
     spoke1.setUserPositionManager(POSITION_MANAGER, true);
   }
 
-  function test_disableApproval_on_inactivePositionManager() public {
+  function test_disableApproval_on_InactivePositionManager() public {
     _approvePositionManager(alice);
     assertTrue(spoke1.isPositionManager(alice, POSITION_MANAGER));
     assertTrue(spoke1.isPositionManagerActive(POSITION_MANAGER));


src/contracts/Spoke.sol
```diff
@@ -132,24 +141,28 @@ contract Spoke is ISpoke, Multicall {
   // /////
 
   /// @inheritdoc ISpoke
-  function supply(uint256 reserveId, uint256 amount) external {
+  function supply(uint256 reserveId, uint256 amount, address onBehalfOf) external {
```
miguelmtzinf: we may want to put restrictions into donations as well... tbd

src/contracts/Spoke.sol
```diff
@@ -366,19 +384,34 @@ contract Spoke is ISpoke, Multicall {
...
     bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
-    // todo allow authorized caller to increase as well
-    require(msg.sender == user || !premiumIncrease, Unauthorized());
+    require(!premiumIncrease || _isPositionManager(user, msg.sender), Unauthorized());
     emit UserRiskPremiumUpdate(user, userRiskPremium);
   }
 
+  /// @inheritdoc ISpoke
+  function setPositionManager(address positionManager, bool approve) external {
+    _positionManager[msg.sender][positionManager] = approve;
```
miguelmtzinf: add validation so only DAO-allowlisted addresses can be used as position managers.

src/contracts/Spoke.sol
```diff
@@ -34,12 +34,20 @@ contract Spoke is ISpoke, Multicall {
   mapping(address user => mapping(uint256 reserveId => DataTypes.UserPosition position))
     internal _userPositions;
   mapping(uint256 reserveId => DataTypes.Reserve reserveData) internal _reserves;
+  mapping(address user => mapping(address positionManager => bool authorized))
```
miguelmtzinf: we could also move it to `DataTypes.UserPosition` no?

miguelmtzinf: that would mean it's per reserve as well, which is a big downside i guess

DhairyaSethi: yeah it should be a global user property, i was thinking for gho soft liquidations we may require different posm per reserve but we can abstract this with another posm lol

src/contracts/Spoke.sol
DhairyaSethi: very expensive bc of an unavoidable additional lookup on every _isPositionManager

src/contracts/Spoke.sol
DhairyaSethi: as discussed today, proceeding on the side of caution and adding

yan-man:
@DhairyaSethi can you instead copy/move the detailed description you have here into the ticket as well?

src/contracts/TreasurySpoke.sol
```diff
@@ -28,18 +28,18 @@ contract TreasurySpoke is Ownable, ITreasurySpoke {
   }
 
   /// @inheritdoc ITreasurySpoke
-  function supply(uint256 reserveId, uint256 amount) external onlyOwner {
+  function supply(uint256 reserveId, uint256 amount, address) external onlyOwner {
```
yan-man, line 35: is this just for consistency w other spokes

src/interfaces/ILiquidityHub.sol
```diff
@@ -112,7 +112,7 @@ interface ILiquidityHub {
    * @dev Only callable by spokes.
    * @param assetId The identifier of the asset.
    * @param amount The amount of asset liquidity to add/supply.
-   * @param from The address which we pull assets from (user).
+   * @param from The address to pull supplied assets from.
```
```suggestion
   * @param from The address from which supplied assets are pulled.
```

src/interfaces/ISpoke.sol
```diff
@@ -20,21 +20,96 @@ interface ISpoke is IMulticall {
   );
   event UserDynamicConfigRefreshed(address indexed user);
 
-  event Supply(uint256 indexed reserveId, address indexed user, uint256 suppliedShares);
+  /**
+   * @notice Emitted on the supply action.
+   * @param reserveId The reserve identifier of the underlying asset as registered on the spoke.
+   * @param caller The transaction initiator, & supplier of the underlying asset.
+   * @param onBehalfOf The owner of the modified position.
```
yan-man: why modified? this just means modified bc more has been supplied right

src/interfaces/ISpoke.sol
```diff
@@ -111,40 +189,53 @@ interface ISpoke is IMulticall {
...
 
   /**
-   * @notice Borrows a specified amount of underlying asset from the given reserve.
+   * @notice Borrows a specified amount of underlying asset from the given reserve
+   * @dev Caller must be `onBehalfOf` or an authorized position manager for `onBehalfOf`.
+   * @dev Caller receives the underlying asset borrowed.
    * @param reserveId The identifier of the reserve.
    * @param amount The amount of asset to borrow.
-   * @param to The address receiving the borrowed assets.
+   * @param onBehalfOf The owner of the position against which debt is generated.
```
yan-man, line 231:
```suggestion
   * @param onBehalfOf The owner of the debt position.
```

src/interfaces/ITreasurySpoke.sol
```diff
@@ -2,34 +2,37 @@
 pragma solidity ^0.8.0;
 
 import {ILiquidityHub} from 'src/interfaces/ILiquidityHub.sol';
+
 /**
  * @title ITreasurySpoke
  */
 interface ITreasurySpoke {
+  // todo inherit from ISpokeBase
```
yan-man, line 10: so this is intending to create ISpokeBase that ISpoke and ITreasurySpoke both inherit from right? If so can you create issue

src/interfaces/ITreasurySpoke.sol
```diff
@@ -2,34 +2,37 @@
...
   error InvalidHubAddress(); // todo: this is duplicated
 
   /**
    * @notice Supplies a specified amount of the underlying asset to a given reserve.
    * @dev The Liquidity Hub pulls the underlying asset from the caller, so prior approval is required.
-   * @dev The reserve identifier matches the corresponding asset in the Liquidity Hub.
+   * @dev The reserve identifier **should match** corresponding asset identifier in the Liquidity Hub.
```
yan-man, line 16: happy to use ID instead of identifier everywhere for brevity, any particular reason for long form?

src/libraries/types/DataTypes.sol
```diff
@@ -172,4 +177,17 @@ library DataTypes {
     uint256 usersLength;
     uint256 newUserSuppliedShares;
   }
+
+  struct ExecuteRepayLocalVars {
```
yan-man, line 185: is this leftover from my PR?

src/contracts/TreasurySpoke.sol
DhairyaSethi: not just consistency, we have to maintain a standard bc treasury spokes use the Util methods so they need to follow ISpoke standard for now

src/interfaces/ISpoke.sol
DhairyaSethi: yeah exactly

src/interfaces/ITreasurySpoke.sol
DhairyaSethi: yesser

src/interfaces/ITreasurySpoke.sol
DhairyaSethi: miguel introduced & i flipped the wording around. i like it bc reserveId and assetId natspec are more verbose now instead of reserveId The reserve Id.

src/libraries/types/DataTypes.sol
DhairyaSethi: no haha was introduced to tackle stack too deep bc repay has an additional arg (onBehalfOf) and it was right on the edge before. and i think you do this in the deficit pr no?

src/interfaces/ISpoke.sol
DhairyaSethi: i would atleast add debtPosition modified, but then it is the exact same to morpho's natspec on this

src/contracts/Spoke.sol
```diff
@@ -34,12 +34,18 @@ contract Spoke is ISpoke, Multicall {
...
     internal _userPositions;
   mapping(uint256 reserveId => DataTypes.Reserve reserveData) internal _reserves;
+  mapping(address positionManager => DataTypes.PositionManagerConfig) internal _positionManager;
   mapping(uint256 reserveId => mapping(uint16 configKey => DataTypes.DynamicReserveConfig config))
     internal _dynamicConfig; // dictionary of dynamic configs per reserve
   DataTypes.LiquidationConfig internal _liquidationConfig;
   uint256[] public reservesList; // todo: rm, not needed
   uint256 public reserveCount;
 
+  modifier onlyPositionManager(address onBehalfOf) {
```
avniculae, line 49: I understand that implementing this as a modifier is useful here to avoid code duplication, but we do not use modifiers in the rest of the repo.  Any thoughts on this? I am fine to duplicate the require statement tbh, wdyt?

src/contracts/Spoke.sol
```diff
@@ -250,83 +274,87 @@ contract Spoke is ISpoke, Multicall {
...
-    DataTypes.UserPosition storage userPosition = _userPositions[msg.sender][reserveId];
+  function repay(
+    uint256 reserveId,
+    uint256 amount,
+    address onBehalfOf
+  ) external onlyPositionManager(onBehalfOf) {
+    DataTypes.UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
     DataTypes.Reserve storage reserve = _reserves[reserveId];
-    uint256 assetId = reserve.assetId;
+    DataTypes.ExecuteRepayLocalVars memory vars;
```
avniculae: :(((

src/contracts/Spoke.sol
```diff
@@ -250,83 +274,87 @@ contract Spoke is ISpoke, Multicall {
...
 
-    (uint256 baseDebt, uint256 premiumDebt) = _getUserDebt(userPosition, assetId);
-    (uint256 baseDebtRestored, uint256 premiumDebtRestored) = _calculateRestoreAmount(
-      baseDebt,
-      premiumDebt,
+    vars.assetId = reserve.assetId;
+    (vars.baseDebt, vars.premiumDebt) = _getUserDebt(userPosition, vars.assetId);
+    _validateRepay(reserve);
```
avniculae: shall we do it before `DataTypes.ExecuteRepayLocalVars memory vars;`?

DhairyaSethi: there are todo's on validateRepay which need base & premium debt, hence this will eventually end up somewhere here

src/contracts/Spoke.sol
DhairyaSethi: yeah it's intentional, this is a prime use case, there's very few things implemented rn which could've benefitted from & i'm not good with duplicated requires hidden in nested validate fn's bc it's easy miss

src/contracts/Spoke.sol
DhairyaSethi: (for a future PR) i'm debating whether it's a good idea passing a ref to the partially filled localVars memory struct here or just go individual args ie debts & amounts, wdyt?

src/contracts/Spoke.sol
```diff
@@ -366,20 +394,52 @@ contract Spoke is ISpoke, Multicall {
...
+    DataTypes.PositionManagerConfig storage config = _positionManager[positionManager];
+    // @dev only allow approval when position manager is active for improved UX
+    require(!approve || config.active, InactivePositionManager());
+    config.approval[msg.sender] = approve;
+    emit ApprovalForPositionManager(msg.sender, positionManager, approve);
+  }
+
+  /// @inheritdoc ISpoke
+  function renouncePositionManagerRole(address user) external {
+    _positionManager[msg.sender].approval[user] = false;
```
avniculae: should a position manager be able to ronounce its DAO approval as well? I think it would be useful

ie config.approval[msg.sender] = false;

DhairyaSethi: that would actually be _config.active = false 

re feat, i'm not even sure renouncing for user is really needed bc these are vetted whitelisted contracts

src/contracts/Spoke.sol
```diff
@@ -553,6 +613,7 @@ contract Spoke is ISpoke, Multicall {
     require(reserve.config.active, ReserveNotActive());
     require(!reserve.config.paused, ReservePaused());
     // todo validate user not trying to repay more
+    // todo NoExplicitAmountToRepayOnBehalf
```
avniculae, line 599: can you pls explain what this is? I saw PR description that is omitted intentionally, but not sure what it is

DhairyaSethi: disallow a front run borrow when you know someone is repaying on your behalf, so only allow full repayment when done through credit delegation 
this is in v3 

as PR description says, it's up for debate whether this will really be a problem since posm's are permissioned

src/contracts/Spoke.sol
```diff
@@ -1286,4 +1347,10 @@ contract Spoke is ISpoke, Multicall {
 
     emit UsingAsCollateral(reserveId, user, usingAsCollateral);
   }
+
+  function _isPositionManager(address user, address manager) private view returns (bool) {
+    if (user == manager) return true;
```
avniculae: I am wondering if there's any concern when `onBehalfOf` is the position manager (eg if the position manager would allow users to specify onBehalfOf). I don't see any issues at the moment, but just wanted to make a note in case we think of concerning scenarios here

src/interfaces/ISpoke.sol
```diff
@@ -97,6 +171,10 @@ interface ISpoke is IMulticall {
   error UsersAndDebtLengthMismatch();
   error Unauthorized();
   error CollateralStatusUnchanged();
+  error InactivePositionManager();
+  error NoExplicitAmountToRepayOnBehalf();
```
avniculae: shall we remove for now, given it's not used?

src/contracts/Spoke.sol
avniculae: we could also convert the modifier into a function and call it at the beginning of every function

src/libraries/types/DataTypes.sol
yan-man: yes same in my deficit pr

src/interfaces/ISpoke.sol
yan-man: eh dont mind either way

src/contracts/Spoke.sol
yan-man: at least for liqs Miguel was mentioning it being unclear when i was passing in structs as args (dont have a solution for that yet) but maybe just individual args unless it becomes too many?

src/contracts/Spoke.sol
DhairyaSethi: mate.. that's an ugly modifier with extra steps haha

src/contracts/Spoke.sol
```diff
@@ -127,14 +134,25 @@ contract Spoke is ISpoke, Multicall {
     // todo emit if stale config overwritten?
   }
 
+  /// @inheritdoc ISpoke
+  function setPositionManager(address positionManager, bool active) external {
+    // todo access control
+    _positionManager[positionManager].active = active;
+    emit PositionManagerSet(positionManager, active);
```
yan-man: doesnt need to be handled here but in general - seem to remember @miguelmtzinf wanting events to be active rather than past tense, was that right? ie Supply vs Supplied

I know we don't do that on all the other set methods atm either

DhairyaSethi: yes Set methods are the exception

src/contracts/Spoke.sol
```diff
@@ -168,20 +190,20 @@ contract Spoke is ISpoke, Multicall {
...
-      msg.sender,
+      onBehalfOf,
       assetId,
       -int256(userPremiumDrawnShares),
       -int256(userPremiumOffset),
       accruedPremium,
       0
     ); // unnecessary but we realize premium debt here
-    uint256 withdrawnShares = HUB.remove(assetId, amount, to);
+    uint256 withdrawnShares = HUB.remove(assetId, amount, msg.sender);
```
yan-man: so idea is not to send directly to `onBehalfOf` but posm will need to separately send. It it to give more flexibility to posm? only downside is gas ig

src/contracts/Spoke.sol
DhairyaSethi: yeah fair call, not an issue here, should be checked on the posm level

tests/Base.t.sol
```diff
@@ -198,7 +199,15 @@ abstract contract Base is Test {
...
 
-    address[6] memory users = [alice, bob, carol, derl, LIQUIDATOR, TREASURY_ADMIN];
+    address[7] memory users = [
+      alice,
+      bob,
+      carol,
+      derl,
+      LIQUIDATOR,
+      TREASURY_ADMIN,
+      POSITION_MANAGER
```
yan-man, line 303: if we end up at 25 users I would like to add `yan`

src/contracts/Spoke.sol
DhairyaSethi: yeah posm will likely call this as part of a strategy will multiple txns so it's much better if it takes ownership. 
also this is to simplify the interface and make it consistent, we could add have three fields: caller, onBehalfOf, and to on each action (morpho does this) but it seems excessive

src/contracts/Spoke.sol
yan-man: worth it to short circuit if it's already false? ig only posm bearing cost

src/contracts/Spoke.sol
```diff
@@ -366,20 +394,52 @@ contract Spoke is ISpoke, Multicall {
...
+    );
     emit UserRiskPremiumUpdate(user, userRiskPremium);
   }
 
-  function getUsingAsCollateral(uint256 reserveId, address user) external view returns (bool) {
+  /// @inheritdoc ISpoke
+  function setApprovalForPositionManager(address positionManager, bool approve) external {
+    DataTypes.PositionManagerConfig storage config = _positionManager[positionManager];
+    // @dev only allow approval when position manager is active for improved UX
+    require(!approve || config.active, InactivePositionManager());
```
yan-man, line 389: worth it to short circuit if `config.approval[msg.sender] == approve` already?

src/contracts/Spoke.sol
yan-man: i dont feel too strongly either way as long as we're consistent, but why exception on set methods?

src/contracts/Spoke.sol
yan-man: minor bytecode savings in internal function vs modifier i think but yeah seems more ugly than modifier and perhaps not worth it

src/contracts/Spoke.sol
DhairyaSethi: no particular reason, we can refactor all set methods in one go though

CheyenneAtapour: `set` is the same active tense and past tense

src/contracts/Spoke.sol
```diff
@@ -166,20 +188,20 @@ contract Spoke is ISpoke, Multicall {
...
-      msg.sender,
+      onBehalfOf,
       assetId,
       -int256(userPremiumDrawnShares),
       -int256(userPremiumOffset),
       accruedPremium,
       0
     ); // unnecessary but we realize premium debt here
-    uint256 withdrawnShares = hub.remove(assetId, amount, to);
+    uint256 withdrawnShares = hub.remove(assetId, amount, msg.sender);
```
CheyenneAtapour, line 208: The idea behind `to` was to specify where the funds get withdrawn to, which could be different than `msg.sender`. Do we not need that anymore? 

Also, is there a consideration to only allow the funds to be sent to the original position holder in case of a withdraw?

src/contracts/Spoke.sol
```diff
@@ -224,20 +248,20 @@ contract Spoke is ISpoke, Multicall {
...
-      msg.sender,
+      onBehalfOf,
       assetId,
       -int256(userPremiumDrawnShares),
       -int256(userPremiumOffset),
       accruedPremium,
       0
     ); // unnecessary but we realize premium debt here
-    uint256 baseDrawnShares = hub.draw(assetId, amount, to);
+    uint256 baseDrawnShares = hub.draw(assetId, amount, msg.sender);
```
CheyenneAtapour, line 240: Same thing with this, we are losing the ability to draw shares and receive the funds to a different address. We don't need / want that functionality?

src/contracts/Spoke.sol
```diff
@@ -249,83 +273,88 @@ contract Spoke is ISpoke, Multicall {
...
-    uint256 restoredShares = hub.restore(
-      assetId,
-      baseDebtRestored,
-      premiumDebtRestored,
+
+    vars.restoredShares = vars.hub.restore(
+      vars.assetId,
+      vars.baseDebtRestored,
+      vars.premiumDebtRestored,
       msg.sender
```
CheyenneAtapour, line 286: This also should be `onBehalfOf`, right?

src/contracts/Spoke.sol
DhairyaSethi: yeah that's intentionally removed to keep the interface simpler and more consistent, it's expected withdrawTo & borrowTo happens with a multicall at the posm manager (or user using 7702) level

src/contracts/Spoke.sol
DhairyaSethi: no this is the argument to pull the repay funds from (hub has no concept for user position, only spoke position), it's paid by posm or user (which is the caller)

src/contracts/Spoke.sol
DhairyaSethi: yep it's intentional, see above. lmk your thoughts

CheyenneAtapour: sounds good!

tests/unit/Spoke/Spoke.PositionManager.t.sol
```diff
@@ -0,0 +1,198 @@
...
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.ApprovalForPositionManager(who, POSITION_MANAGER, true);
+    vm.prank(who);
+    spoke1.setApprovalForPositionManager(POSITION_MANAGER, true);
+
+    assertTrue(spoke1.isPositionManager(who, POSITION_MANAGER));
+    assertTrue(spoke1.isPositionManagerActive(POSITION_MANAGER));
+  }
+
+  function _resetAllowance(address user) internal {
```
CheyenneAtapour: Maybe this function can be more generalized by taking in input a token, and then can move to spoke base, just in case we'd like to reuse

tests/unit/Spoke/Spoke.Repay.EdgeCases.t.sol
```diff
@@ -80,8 +78,7 @@ contract SpokeRepayEdgeCaseTest is SpokeBase {
     supplyExRateBefore = getSupplyExRate(daiAssetId);
     debtExRateBefore = getDebtExRate(daiAssetId);
 
-    vm.prank(bob);
-    spoke1.repay(_daiReserveId(spoke1), type(uint256).max);
+    Utils.repay(spoke1, _daiReserveId(spoke1), bob, type(uint256).max, bob);
```
CheyenneAtapour, line 84: @yan-man had suggested previously that we use the defined constant `UINT256_MAX` now, but nbd can address in a later refactor PR

src/interfaces/ISpoke.sol
```diff
@@ -23,21 +23,96 @@ interface ISpoke is IMulticall, IAccessManaged {
   event UserDynamicConfigRefreshedAll(address indexed user);
   event UserDynamicConfigRefreshedSingle(address indexed user, uint256 reserveId);
 
-  event Supply(uint256 indexed reserveId, address indexed user, uint256 suppliedShares);
+  /**
+   * @notice Emitted on the supply action.
+   * @param reserveId The reserve identifier of the underlying asset as registered on the spoke.
+   * @param caller The transaction initiator, & supplier of the underlying asset.
```
miguelmtzinf: sorry for this nit, but let's use `and` because we ve never used & before and feels inconsistent

src/interfaces/ISpoke.sol
```diff
@@ -23,21 +23,96 @@ interface ISpoke is IMulticall, IAccessManaged {
...
+    address indexed onBehalfOf,
+    uint256 suppliedShares
+  );
+
+  /**
+   * @notice Emitted on the withdraw action.
+   * @param reserveId The reserve identifier of the underlying asset as registered on the spoke.
+   * @param caller The transaction initiator, & recipient of the underlying asset being withdrawn.
+   * @param onBehalfOf The owner of the modified position.
+   * @param suppliedShares The amount of supply shares burned.
```
miguelmtzinf, line 45: should this be `supplyShares`? so we have `supplyShares` and `debtShares`

src/interfaces/ISpoke.sol
```diff
@@ -23,21 +23,96 @@ interface ISpoke is IMulticall, IAccessManaged {
...
   );
-  event Repay(uint256 indexed reserveId, address indexed user, uint256 drawnShares);
+
+  /**
+   * @notice Emitted on setUsingAsCollateral action.
+   * @param reserveId The reserve identifier of the underlying asset as registered on the spoke.
+   * @param user The owner of the position being modified.
+   * @param usingAsCollateral Boolean whether the reserve is enabled or disabled as collateral.
+   */
   event UsingAsCollateral(uint256 indexed reserveId, address indexed user, bool usingAsCollateral);
```
miguelmtzinf: what about the caller here

src/interfaces/ISpoke.sol
```diff
@@ -23,21 +23,96 @@ interface ISpoke is IMulticall, IAccessManaged {
...
+
+  /**
+   * @notice Emitted on updateUserRiskPremium action.
+   * @param user The owner of the position being modified.
+   * @param riskPremium The new risk premium (BPS) value of user.
+   */
+  event UserRiskPremiumUpdate(address indexed user, uint256 riskPremium);
```
miguelmtzinf, line 101: same, caller?

src/interfaces/ISpoke.sol
```diff
@@ -23,21 +23,96 @@ interface ISpoke is IMulticall, IAccessManaged {
...
+   * @notice Emitted on the supply action.
+   * @param reserveId The reserve identifier of the underlying asset as registered on the spoke.
+   * @param caller The transaction initiator, & supplier of the underlying asset.
+   * @param onBehalfOf The owner of the modified position.
+   * @param suppliedShares The amount of supply shares minted.
+   */
+  event Supply(
+    uint256 indexed reserveId,
+    address indexed caller,
+    address indexed onBehalfOf,
```
miguelmtzinf: kinda prefer `user` here, specifically because there are other events using `user` as the position modified

src/interfaces/ISpoke.sol
```diff
@@ -115,40 +192,53 @@ interface ISpoke is IMulticall, IAccessManaged {
...
-   * @param reserveId The identifier of the reserve
+   * @notice Supply an amount of underlying asset of the specified reserve.
+   * @dev The Liquidity Hub pulls the underlying asset from the caller, so prior token approval is required.
+   * @dev Caller must be `onBehalfOf` or an authorized position manager for `onBehalfOf`.
+   * @param reserveId The reserve identifier.
    * @param amount The amount of asset to supply.
+   * @param onBehalfOf The owner of position to add supply shares to.
    */
-  function supply(uint256 reserveId, uint256 amount) external;
+  function supply(uint256 reserveId, uint256 amount, address onBehalfOf) external;
```
miguelmtzinf, line 212: wondering if we should flip it, so it's `reserveId - user - amount`

src/contracts/Spoke.sol
```diff
@@ -214,59 +235,75 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
     ); // we settle base debt here
 
-    reserve.baseDrawnShares -= restoredShares;
-    userPosition.baseDrawnShares -= restoredShares;
+    reserve.baseDrawnShares -= vars.restoredShares;
+    userPosition.baseDrawnShares -= vars.restoredShares;
 
     if (userPosition.baseDrawnShares == 0) {
       _positionStatus[msg.sender].setBorrowing(reserveId, false);
```
miguelmtzinf: we must use `onBehalfOf` instead.
We are doing smt wrong if tests are passing...

src/contracts/Spoke.sol
```diff
@@ -320,26 +361,52 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-    (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
-    bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
+  function updateUserRiskPremium(address onBehalfOf) external {
+    (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(onBehalfOf);
+    bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, onBehalfOf, userRiskPremium);
+
     // check permissions if premium increases and not called by user
-    if (premiumIncrease && user != msg.sender) {
+    if (premiumIncrease && !_isPositionManager({user: onBehalfOf, manager: msg.sender})) {
       _checkCanCall(_msgSender(), _msgData());
```
miguelmtzinf: do we wanna change to `msg.sender` and `msg.data` now?

src/contracts/Spoke.sol
DhairyaSethi: ah great catch ty! was missed in one of the rebase. yes we are, will add regression test & fix existing assertions

src/interfaces/ISpoke.sol
DhairyaSethi: chose suppliedShares bc that's our vars (& withdrawnShares w remove) but agree will rename

src/contracts/Spoke.sol
```diff
@@ -320,26 +361,52 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
     }
   }
 
-  function getUsingAsCollateral(uint256 reserveId, address user) external view returns (bool) {
+  /// @inheritdoc ISpoke
+  function setApprovalForPositionManager(address positionManager, bool approve) external {
```
miguelmtzinf: nbd but a bit long. would `setPositionManager` or `approvePositionManager` be fine as well?

src/contracts/Spoke.sol
```diff
@@ -320,26 +361,52 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+    emit ApprovalForPositionManager(onBehalfOf, msg.sender, false);
+  }
+
+  /// @inheritdoc ISpoke
+  function isPositionManager(address user, address positionManager) external view returns (bool) {
+    return _isPositionManager(user, positionManager);
+  }
+
+  /// @inheritdoc ISpoke
+  function isPositionManagerActive(address positionManager) external view returns (bool) {
```
miguelmtzinf, line 406: annoying thing is that we cannot check existence, meaning that we cannot know if it was active sometime ago

src/contracts/Spoke.sol
miguelmtzinf: can you create a new ticket for this please, so we can discuss

src/contracts/TreasurySpoke.sol
miguelmtzinf, line 35: can you create a ticket about having an interface for ISpokeBase where we have main functions as standard?

src/contracts/Spoke.sol
DhairyaSethi: setPositionManager was my pref thought was as well but using that for the governance method with whitelists a posm rn, can rename that to something else? (needs a general name it accepts a bool as input)
in general went with set* convention for both user & gov methods bc of the using as collateral setter convention 

approvePosm - same argument as above, setter looks better bc bool arg, and set* convention but i'm agnostic tbh

src/interfaces/ISpoke.sol
DhairyaSethi: image
let's make this update in a separate PR? 😅

DhairyaSethi: we have `drawnShares` also (not `debtShares`) imo let's tackle both of these together

src/interfaces/ISpoke.sol
DhairyaSethi: kept user from below

src/interfaces/ISpoke.sol
DhairyaSethi: same, maybe the other two should have been onBehalfOf bc that's more technically accurate, but have aligned all to have this param `user`
we can switch all to `onBehalfOf` as well if needed

src/interfaces/ISpoke.sol
DhairyaSethi: unsure, since it's the standard across the industry, v3, morpho, euler, fluid to have amount, user

but i'm mostly agnostic

src/contracts/Spoke.sol
DhairyaSethi: so it was bc we don't expose whether a reserve is being borrowed at all! once it got removed from UserPosition struct looks like all assertions were removed for this
brought them back and exposed a dedicated getter like isUsingAsCollateral: [bc2b36d](https://github.com/aave/aave-v4/pull/462/commits/bc2b36ddb3dddcc5c920cb0b03f0880c66547855)

src/contracts/Spoke.sol
DhairyaSethi: as in it was previously ever active? lmk what you think are the benefits - i can do some gas golfing with the user approval mapping lookup such that adding a timestamp or smthing here won't impact user checks in terms of gas a lot
personally i think its fine

src/contracts/Spoke.sol
miguelmtzinf: worth to keep renounce fn we can support only once position managers. Example:
1. User activates X posm for one time repay with collateral when ETH hits 10k USDC (:pray:)
2. ETH hits 10k USDC, repayment happens
3. posm is deactivated automatically

src/interfaces/ISpoke.sol
miguelmtzinf: yes... added to #479 

```
## Liquidity Hub
8. `supplyShares` -> `addShares`
9. `debtShares` -> `drawnShares`

## Spoke
3. `suppliedShares` and `drawnShares` -> `supplyShares` and `debtShares`
```
good?

src/interfaces/ISpoke.sol
miguelmtzinf: @DhairyaSethi caller and user may be different and we are not emitting this info

src/interfaces/ISpoke.sol
miguelmtzinf: same, do we need `caller` info?

src/interfaces/ISpoke.sol
miguelmtzinf: true, let's keep it as is
was thinking about ERC20 where we have `token - user - amount`... but not exactly the same usecase :D

src/contracts/Spoke.sol
miguelmtzinf: thanks for adding some checks in the tests, but let's create a dedicated ticket for checking flags please... we need to make sure this is well covered

src/contracts/Spoke.sol
miguelmtzinf: yeah.. seems we are in need of `set`. `setPositionManager`/`setReservePositionManager` and `setUserPositionManager` d be confusing?

src/interfaces/ISpoke.sol
DhairyaSethi: unsure, would like to avoid bc it needs passing down caller in notify & the possible callers are user, posm (whitelisted), and liquidator

src/contracts/Spoke.sol
DhairyaSethi: yeah i think that'd be confusing haha

src/contracts/Spoke.sol
miguelmtzinf: just more informative, but i think is fine

src/interfaces/ISpoke.sol
DhairyaSethi: so here i would to avoid caller tbh bc wanted to move emitting the event in the UserPosition library and will need to pass down caller there, possible callers are user, posm which is whitelisted (unlike v3 where delegated addr is arbitrary) & liquidator - same as update user rp 
but we can add if you think the benefits are worth it

src/contracts/Spoke.sol
DhairyaSethi: #544

src/contracts/Spoke.sol
miguelmtzinf: I d be fine with `setReservePositionManager` or `setGlobalPositionManager` and `setUserPositionManager` but no strong opinion

src/contracts/TreasurySpoke.sol
DhairyaSethi: https://github.com/aave/aave-v4/issues/545

src/contracts/Spoke.sol
miguelmtzinf: @DhairyaSethi to remove and create ticket

src/interfaces/ISpoke.sol
miguelmtzinf: ok, but feels mandatory for `UsingAsCollateral`

src/contracts/Spoke.sol
DhairyaSethi: how about `whitelistPositoinManager` or `add/updatePositionManager` for the gov methods then instead

src/interfaces/ISpoke.sol
DhairyaSethi: the reason to move to library like v3 is we've forgotten to emit the even on liquidate so it makes sense the abstraction which updates the state also emits an emit

src/contracts/Spoke.sol
DhairyaSethi: #546

tests/unit/Spoke/Spoke.PositionManager.t.sol
```diff
@@ -0,0 +1,211 @@
...
+
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.ApprovalForPositionManager(who, POSITION_MANAGER, true);
+    vm.prank(who);
+    spoke1.setApprovalForPositionManager(POSITION_MANAGER, true);
+
+    assertTrue(spoke1.isPositionManager(who, POSITION_MANAGER));
+    assertTrue(spoke1.isPositionManagerActive(POSITION_MANAGER));
+  }
+
```
miguelmtzinf, line 258: missing cases like:
1. action call via posm is correct but not active anymore
2. setApproval fails because posm is not active
3. setApproval false succeeds even if posm is not active
4. action call via posm fails because is active but user config was not set

src/contracts/Spoke.sol
miguelmtzinf: i like `updatePositionManager` or `allowPositionManager`

tests/unit/Spoke/Spoke.PositionManager.t.sol
DhairyaSethi: added! lmk

Issue #435, opened 2025-06-03
refactor - do not update collateral status if none remains


yan-man:
ref: https://avaraverse.slack.com/archives/C04HCBMEEBZ/p1748970648624429

PR #467, opened 2025-06-20
Liquidation Fee - Dedicated LH Method

### LINKED ISSUE #377: Implement Liq Protocol Fee
Currently, without treasury accounting, the treasury liq protocol fee amount is emitted in a temp event in Spoke, and read within tests to check its values 
- instead, read the values from the treasury accounting in hub (ie shares earned by treasury) 
- update `_assertProtocolFeeEarned` assertions, and rm event logging in Liq Call tests, and `_tmpGetProtocolFeeFromLiqEvent`
- rm event from Spoke/ISpoke

yan-man:
- during payment of Liquidation Fee (LF), remove only the seized collateral amount without LF. Create a separate dedicated method on LH to transfer existing supplied shares from current spoke to feeReceiver spoke
- can be invoked during liquidations to send seized collateral to LH as Liquidation Fee (LF)
- clean up temp event emitted during liquidations re: LF, associated prior liquidation test assertions on LF
- rename liquidation protocol fee -> liquidation fee
- undo changing `usingAsCollateral` status during liquidation

Benefits:
- allows direct `remove` to transfer seized collateral to liquidator directly (don't need to transfer collateral to spoke first, and then redistribute to liquidator, which was needed before to separate base seized collateral from LF)
- doesn't require an `add` type operation to external supply fee shares to feeReceiver, can just adjust accounting on LH (saves another token transfer action op)

See spec doc: https://s.craft.me/sCEeOkOcpSGDit for details

closes #377 
closes https://github.com/aave/aave-v4/issues/486
closes https://github.com/aave/aave-v4/issues/435

### DIFF
--- a/tests/unit/LiquidityHub/LiquidityHub.PayFee.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.PayFee.t.sol
@@ -73,16 +73,8 @@ contract LiquidityHubPayFeeTest is LiquidityHubBase {
       to: address(spoke1)
     });
 
-    _supplyAndDrawLiquidity({
-      assetId: daiAssetId,
-      supplyUser: bob,
-      supplySpoke: address(spoke2),
-      supplyAmount: 100e18,
-      drawUser: alice,
-      drawSpoke: address(spoke3),
-      drawAmount: 100e18,
-      skipTime: skipTime
-    });
+    _addLiquidity(daiAssetId, 100e18);
+    _drawLiquidity(daiAssetId, 100e18, true);
 
     uint256 spokeSharesBefore = hub.getSpokeSuppliedShares(daiAssetId, address(spoke1));
 


src/contracts/LiquidityHub.sol
```diff
@@ -317,6 +302,20 @@ contract LiquidityHub is ILiquidityHub {
     require(premiumDebtAfter + realizedPremiumTaken - premiumDebtBefore <= 2, InvalidDebtChange());
   }
 
+  /// @inheritdoc ILiquidityHub
+  function donate(uint256 assetId, uint256 amount, address from) external returns (uint256) {
```
DhairyaSethi: here from is the spoke address but on add it's the original supplier, think we can use different arg names for clarity?

DhairyaSethi: and would from always be equal to msg.sender?

src/contracts/LiquidityHub.sol
```diff
@@ -565,6 +564,33 @@ contract LiquidityHub is ILiquidityHub {
     emit SpokeAdded(assetId, spoke); // todo: emit config
   }
 
+  function _executeAdd(
+    uint256 assetId,
+    uint256 amount,
+    address from,
+    DataTypes.Asset storage asset,
+    DataTypes.SpokeData storage spoke
```
DhairyaSethi: nit: can invert order of items to storage pointers are first but nbd

src/contracts/Spoke.sol
```diff
@@ -1186,23 +1184,37 @@ contract Spoke is ISpoke, Multicall {
...
+          address(collateralReserveHub),
+          vars.totalLiquidationProtocolFeeAmount
+        );
+        collateralReserveHub.donate(
+          vars.collateralAssetId,
+          vars.totalLiquidationProtocolFeeAmount,
+          address(this)
+        );
+      } else {
+        vars.totalCollateralToLiquidate += vars.totalLiquidationProtocolFeeAmount;
```
DhairyaSethi: ah why do we omit taking a fee in this case? bc not worth it & possible donation side effects?

src/contracts/Spoke.sol
```diff
@@ -1186,23 +1184,37 @@ contract Spoke is ISpoke, Multicall {
...
-    );
+
+    if (vars.totalLiquidationProtocolFeeAmount > 0) {
+      if (
+        collateralReserveHub.convertToSuppliedShares(
+          vars.collateralAssetId,
+          vars.totalLiquidationProtocolFeeAmount
+        ) > 0
+      ) {
+        IERC20(collateralReserve.asset).safeIncreaseAllowance(
```
DhairyaSethi: think we can infinite approve hub on reserve listing, should be fine; especially bc hub will be immutable

src/contracts/LiquidityHub.sol
```diff
@@ -317,6 +302,20 @@ contract LiquidityHub is ILiquidityHub {
     require(premiumDebtAfter + realizedPremiumTaken - premiumDebtBefore <= 2, InvalidDebtChange());
   }
 
+  /// @inheritdoc ILiquidityHub
+  function donate(uint256 assetId, uint256 amount, address from) external returns (uint256) {
+    // TODO: authorization - only spokes
+
+    DataTypes.Asset storage asset = _assets[assetId];
+    DataTypes.SpokeData storage spoke = _spokes[assetId][asset.config.feeReceiver];
```
DhairyaSethi: feeReceiver can be null when liquidity fee (ie reserve factor) is null, probably need to note & change assertions there or just accept donation & short circuit in that case?

src/interfaces/ILiquidityHub.sol
```diff
@@ -173,6 +179,16 @@ interface ILiquidityHub {
     uint256 realizedPremiumTaken
   ) external;
 
+  /**
+   * @notice Donate asset on behalf of feeReceiver.
+   * @dev Only callable by spokes.
+   * @param assetId The identifier of the asset.
+   * @param amount The amount of asset liquidity to add/supply.
+   * @param from The address (user) from which we pull assets.
```
DhairyaSethi: no? in the impl this is the spoke address & it pulls funds from the spoke that it just sent on the withdraw

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -257,22 +264,18 @@ contract SpokeLiquidationBase is SpokeBase {
...
-      state.treasury.balanceChange // actual protocol fee shares, from tmp emitted event
-    );
+    uint256 liqProtocolFeeAmount = state.treasury.balanceChange;
     // TODO: resolve precision loss difference
     assertApproxEqAbs(
       liqProtocolFeeAmount,
-      totalLiqBonusAmount.percentMulUp(state.liquidationProtocolFee),
-      3,
+      totalLiqBonusAmount.percentMulUp(state.liquidationProtocolFeeBPS),
+      2,
```
DhairyaSethi: 🎉

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.CloseFactor.t.sol
```diff
@@ -457,9 +457,15 @@ contract LiquidationCallCloseFactorTest is SpokeLiquidationBase {
...
 
-    // logs to read protocol fee from tmp emitted event
-    // TODO: update when treasury accounting is done
-    vm.recordLogs();
+    state.liqProtocolFeeShares = hub.convertToSuppliedShares(
+      state.collateralAssetId,
+      state.liqProtocolFee
+    );
+
+    // if protocol fee equates to 0 shares, it is instead added to collateral to liquidate for the liquidator
```
DhairyaSethi: does this incentivise multiple small liquidations?

src/contracts/LiquidityHub.sol
yan-man: yes but its only the spoke address during liquidation for LPF, could also be a generic donation from any end user too (perhaps in future)

not always msg.sender, can be any from address

src/contracts/LiquidityHub.sol
yan-man: ok will do, does that save gas?

src/contracts/Spoke.sol
yan-man: was going to comment on this proactively - this is when there are invalid shares (amount rounds to 0) so thought it'd be better to give that back to liquidator rather than leaving dust in spoke. Can't donate in this case bc validation causes a revert with 0 shares

src/contracts/LiquidityHub.sol
yan-man: ah so feeReceiver can be 0 right? yeah prob need some validation there. 

edit: did validation and will revert if feeReceiver is 0. if feeReceiver is 0, i think liq protocol fee should also be 0 too. https://github.com/aave/aave-v4/pull/467/commits/06751295312fb9d04426c6e2e75ca7b24720c265

src/contracts/LiquidityHub.sol
DhairyaSethi: nope haha, just a standard i feel (see validate methods)

src/interfaces/ILiquidityHub.sol
yan-man: i can rm the `(user)` part, but otherwise it can be from anyone, just so happens in LPF it is from spoke. But donate is generic enough that donations can come from any user too (if a spoke were to implement that)

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.CloseFactor.t.sol
yan-man: balanced by gas costs of doing that, but yes. Future features related to this that can help mitigate:
- https://github.com/aave/aave-v4/issues/380 (under threshold you must liquidate all debt)
- potentially this: https://github.com/aave/aave-v4/issues/260?issue=aave%7Caave-v4%7C399

src/contracts/Spoke.sol
DhairyaSethi: yep yep, wondering if it's ok to accept 0 shares on donate

yan-man: not as I have it implemented but I can do that. I think I'd add a bool flag input to `_validateAdd` to allow accepting donations

and just to clarify, to accept these donations, it becomes a donation to the protocol rather than a donation to the feeReceiver (bc 0 shares, so this amount just gets added to available liquidity), therefore this would increase the supply ex rate. Perhaps would this be a inflation attack vector?

wdyt @CheyenneAtapour?

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/06751295312fb9d04426c6e2e75ca7b24720c265

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/06751295312fb9d04426c6e2e75ca7b24720c265

src/contracts/Spoke.sol
yan-man: decided today to prevent this donation (to protocol). Instead will just rename the `donate` method to be more clear that it is a donation to feeReceiver specifically and don't allow donations to protocol (as supply ex rate inflation attack vector): https://github.com/aave/aave-v4/pull/467/commits/87bfd919a8937735843c438619d1d0e82f0578cb

src/contracts/Spoke.sol
```diff
@@ -85,6 +85,8 @@ contract Spoke is ISpoke, Multicall {
     });
     _dynamicConfig[reserveId][dynamicConfigKey] = dynamicConfig;
 
+    IERC20(asset).forceApprove(address(config.hub), type(uint256).max);
```
miguelmtzinf: I hate this, but the opposite does not look nicer no? (we safeTransfer to hub and call `supplyToFeeReceiver`)

src/contracts/LiquidityHub.sol
```diff
@@ -317,6 +302,26 @@ contract LiquidityHub is ILiquidityHub {
     require(premiumDebtAfter + realizedPremiumTaken - premiumDebtBefore <= 2, InvalidDebtChange());
   }
 
+  /// @inheritdoc ILiquidityHub
+  function supplyToFeeReceiver(
```
miguelmtzinf: do not like it much, but needs to be `addToFeeReceiver` in any case

src/contracts/LiquidityHub.sol
```diff
@@ -317,6 +302,26 @@ contract LiquidityHub is ILiquidityHub {
...
+    uint256 assetId,
+    uint256 amount,
+    address from
+  ) external returns (uint256) {
+    // TODO: authorization - only spokes
+
+    DataTypes.Asset storage asset = _assets[assetId];
+    address feeReceiver = asset.config.feeReceiver;
+    require(feeReceiver != address(0), InvalidFeeReceiver());
+    DataTypes.SpokeData storage spoke = _spokes[assetId][feeReceiver];
```
miguelmtzinf: should we check existence?

src/contracts/LiquidityHub.sol
```diff
@@ -317,6 +302,26 @@ contract LiquidityHub is ILiquidityHub {
...
+
+    uint256 suppliedShares = _executeAdd(asset, spoke, assetId, amount, from);
+
+    emit SupplyToFeeReceiver(assetId, msg.sender, suppliedShares, amount);
```
miguelmtzinf: I think we should either use `Add` event, or the same event we have when accruing interest fees to feeReceiver

src/contracts/Spoke.sol
```diff
@@ -1186,23 +1188,33 @@ contract Spoke is ISpoke, Multicall {
       0,
       0
     );
-    vars.totalLiquidationProtocolFeeShares = collateralReserveHub.convertToSuppliedShares(
-      vars.collateralAssetId,
-      vars.totalLiquidationProtocolFeeAmount
-    );
+
+    if (vars.totalLiquidationProtocolFeeAmount > 0) {
+      if (
```
miguelmtzinf: shouldn't this be AND?
If fee > 0 and > minimumAmountToAdd 
then: do donation
otherwise: fee = 0

src/contracts/LiquidityHub.sol
```diff
@@ -317,6 +302,26 @@ contract LiquidityHub is ILiquidityHub {
...
+  function supplyToFeeReceiver(
+    uint256 assetId,
+    uint256 amount,
+    address from
+  ) external returns (uint256) {
+    // TODO: authorization - only spokes
+
+    DataTypes.Asset storage asset = _assets[assetId];
+    address feeReceiver = asset.config.feeReceiver;
+    require(feeReceiver != address(0), InvalidFeeReceiver());
```
miguelmtzinf: i don't like the fact that a liquidation can fail due to the donate.
This can happen because of this and because of `validateAdd` fn.
- re feeReceiver, i d impose a good config: if lpf != 0, feeReceiver cannot be zero
- re validateAdd, i d skip. fee should be accrued to fee receiver no matter what imo

src/contracts/LiquidityHub.sol
miguelmtzinf: alternatives: `sendFee`, `transferFee`, `payFee`, `addFee`

src/contracts/LiquidityHub.sol
```diff
@@ -565,6 +570,33 @@ contract LiquidityHub is ILiquidityHub {
...
+    DataTypes.Asset storage asset,
+    DataTypes.SpokeData storage spoke,
+    uint256 assetId,
+    uint256 amount,
+    address from
+  ) internal returns (uint256) {
+    asset.accrue(_spokes[assetId][asset.config.feeReceiver]);
+    _validateAdd(asset, spoke, amount, from);
+
+    asset.updateBorrowRate({liquidityAdded: amount, liquidityTaken: 0});
```
miguelmtzinf: bad thing here is that we are adding more available liquidity so rates are impacted.
the case of interest fees accrual it's kinda different, because liq is already part of the availableLiq so rates are not impacted.

src/contracts/Spoke.sol
```diff
@@ -1186,23 +1188,33 @@ contract Spoke is ISpoke, Multicall {
...
+        vars.totalLiquidationProtocolFeeAmount = 0;
+      }
+    }
 
     // transfer total liquidated collateral to liquidator
-    IERC20(collateralReserve.asset).safeTransfer(msg.sender, vars.totalCollateralToLiquidate);
-    // TODO: treasury accounting for protocol fee
-    // TODO: rm temp event
-    emit TmpLiquidationFee(vars.totalLiquidationProtocolFeeShares);
+    IERC20(collateralReserve.asset).safeTransfer(liquidator, vars.totalCollateralToLiquidate);
```
miguelmtzinf: transferring to the hub may not be a complete bad idea if we are already transferring to the liquidator

miguelmtzinf: alternative would be:
- don't remove lpf
- call `payFeeWithExistingLiquidity` so we move assets from the caller spoke to fee receiver -> just accounting, no transfer of assets or validations (only that caller spoke can decrease suppliedAssets by lpf amount)

src/contracts/Spoke.sol
yan-man: yes this can be AND, it's just a slightly less efficient operation for when `totalLiquidationProtocolFeeAmount == 0` due to fee config to 0, bc then we redundantly do the `else` op. But maybe less messy looking in the code

src/contracts/Spoke.sol
yan-man: > alternative would be

this draft pr impl is here: https://github.com/aave/aave-v4/pull/477

src/contracts/LiquidityHub.sol
yan-man: we have a todo on emitting the event for accruing interest to feeReceiver atm, but will rename to `AccrueFees` for now and rm `msg.sender` from the emitted args bc this won't exist in the regular interest accrual to feeReceiver anyhow

src/contracts/LiquidityHub.sol
yan-man: > re feeReceiver, i d impose a good config: if lpf != 0, feeReceiver cannot be zero

will impl this but note that LPF is defined on spoke Reserve but feeReceiver validation will be on the associated hub

miguelmtzinf: hmm didn't realize... a bit weird, because the Spoke can call the Hub to check non-zero feeReceiver if setting non-zero lpf, but the opposite won't happen (checking lpf is zero when setting feeReceiver to zero).
Do we have an alternative? Otherwise it looks like we need to come up with a good compromise here

src/contracts/Spoke.sol
miguelmtzinf: not only about efficiency.
IF fee > 0 AND fee is NOT > minimumAmountToAdd
then we do not do anything... which is bad right

src/contracts/LiquidityHub.sol
DhairyaSethi: can't rm msg.sender here bc they've had a balance change & we have to signal that

src/contracts/Spoke.sol
yan-man: > IF fee > 0 AND fee is NOT > minimumAmountToAdd

i think in this case we go into else branch then, which donates this intermediate amt to liquidator

but isn't relevant now in new impl

src/contracts/Spoke.sol
yan-man: continuing with the suggested impl here in this PR

src/contracts/LiquidityHub.sol
yan-man: decided we will only have non-zero feeReceivers

src/contracts/LiquidityHub.sol
yan-man: decided we will only allow non zero fee receiver

src/contracts/LiquidityHub.sol
yan-man: will use `payFeeWithExistingLiquidity`

src/contracts/Spoke.sol
yan-man: no longer relevant with new approach

src/contracts/LiquidityHub.sol
yan-man: with new approach, can only pay feeReceiver using a spoke's existing liquidity

src/contracts/Spoke.sol
yan-man: no longer relevant

src/contracts/LiquidityHub.sol
```diff
@@ -317,6 +319,28 @@ contract LiquidityHub is ILiquidityHub {
...
+    // TODO: authorization - only spokes
+    _validatePayFeeWithExistingLiquidity(feeShares);
+
+    DataTypes.Asset storage asset = _assets[assetId];
+    DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
+
+    uint256 payableShares = spoke.suppliedShares;
+    require(feeShares <= payableShares, SuppliedSharesExceeded(payableShares));
+
+    spoke.suppliedShares -= feeShares;
```
```suggestion
    unchecked { spoke.suppliedShares = payableShares - feeShares; } 
```
since you've already cached & enforced bounds 
can leave the unchecked opt for later since it's applicable at other places as well

src/contracts/LiquidityHub.sol
```diff
@@ -317,6 +319,28 @@ contract LiquidityHub is ILiquidityHub {
...
+    _spokes[assetId][asset.config.feeReceiver].suppliedShares += feeShares;
+
+    emit AccrueFees(assetId, feeShares);
```
DhairyaSethi: we need to signal the decrease in sender spoke shares, should emit a Remove event or make an exceptional AccrueFees
prefer the former 
side note: you mentioned somewhere we have a ticket to add feeReceiver to AccrueFees?

src/contracts/Spoke.sol
```diff
@@ -1088,18 +1101,29 @@ contract Spoke is ISpoke, Multicall {
...
       vars.totalRestoredShares += vars.restoredShares;
 
-      // liquidate collateral
-      vars.withdrawnShares = collateralReserveHub.remove(
+      // expected total withdrawn shares includes liquidation fee
+      vars.withdrawnShares = collateralReserveHub.convertToSuppliedSharesUp(
+        vars.collateralAssetId,
+        vars.liquidationFeeAmount + vars.collateralToLiquidate
+      );
+      vars.withdrawnLiquidatorShares = collateralReserveHub.convertToSuppliedSharesUp(
```
DhairyaSethi: cross posting so it's tracked: this isn't needed, can get this value from the remove call and calculate liquidationFeeShares after

yan-man: yes this was in progress, done

src/contracts/LiquidityHub.sol
yan-man: > add feeReceiver to AccrueFees

no, just that Miguel wanted to emit the same event initially as for a regular non-payment accrual, but that one won't have a `msg.sender` like in this scenario. 

So for now just emitted a different `PayFeeWithExistingLiquidity` event as it isn't exactly a `Remove` action and dont know if we should convert shares to amt just to emit, but open to hearing feedback cc @CheyenneAtapour @miguelmtzinf

src/contracts/LiquidityHub.sol
yan-man: decided we won't be having feeReceiver as zero addr

src/interfaces/ILiquidityHub.sol
yan-man: not relevant anymore, no users to pull from, just spoke

src/contracts/LiquidityHub.sol
yan-man: liquidity no longer affected with new approach, direct accounting transfer

src/contracts/LiquidityHub.sol
DhairyaSethi: it is a Remove, consider how transfer w address(0) is emitted on burn/mint in erc20s, remove signals decrease in supplied shares
agree w not converting to amount tho, didn't realise that, we can discuss if it's worth switching emitting shares here & amounts on hub since that was coin toss decision

yan-man: it's a remove in that supplied shares goes down but isn't exactly a remove in terms of the other implications - ie availableLiquidity does not change, accrue doesn't happen (but actually it probably should), borrow rate does not get updated

DhairyaSethi: bc it removes from one spoke and adds to another 
the deltas on everything else cancels out, so it can still be thought of as a remove

src/contracts/LiquidityHub.sol
```diff
@@ -268,6 +268,31 @@ contract LiquidityHub is ILiquidityHub {
...
+    uint256 feeShares
+  ) external returns (uint256) {
+    // TODO: authorization - only spokes
+    _validatePayFeeWithExistingLiquidity(feeShares);
+
+    DataTypes.Asset storage asset = _assets[assetId];
+    DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
+    DataTypes.SpokeData storage feeReceiverSpoke = _spokes[assetId][asset.config.feeReceiver];
+
+    asset.accrue(assetId, feeReceiverSpoke);
```
yan-man: this should escape early for liquidations but we still need this for generic `payFeeWithExistingLiquidity` invocations from spokes outside of liquidation

src/contracts/LiquidityHub.sol
```diff
@@ -268,6 +268,31 @@ contract LiquidityHub is ILiquidityHub {
...
+
+    uint256 payableShares = spoke.suppliedShares;
+    require(feeShares <= payableShares, SuppliedSharesExceeded(payableShares));
+
+    spoke.suppliedShares = payableShares - feeShares;
+    feeReceiverSpoke.suppliedShares += feeShares;
+
+    emit PayFeeWithExistingLiquidity(assetId, msg.sender, feeShares);
```
yan-man: used this event for now but wanted to bring discussion here on emitting events:
- https://github.com/aave/aave-v4/pull/467/files/775a1417be9290c37340c3a2f56e3ac11fbfeb5e#r2169027886
- https://github.com/aave/aave-v4/pull/467#discussion_r2172902857

i guess we can emit 2 events here, 1 which is same for a general accrual, and one specifically for the msg.sender suppliedShares change (which could be `Remove`)? 
@DhairyaSethi @miguelmtzinf

DhairyaSethi: yep as suggested before, that works 

thinking more, one caveat w Remove is it signals the receiver param receiving underlying assets whereas here it'd be the feeReceiver receiving shares (the addition is signalled by AccrueFee)

tests/invariant/LiquidityHubHandler.t.sol
```diff
@@ -126,7 +126,7 @@ contract LiquidityHubHandler is Test {
     s.userSupplied[assetId][user] -= amount;
   }
 
-  function donate(uint256 assetId, address user, uint256 amount) public {
+  function payFeeWithExistingLiquidity(uint256 assetId, address user, uint256 amount) public {
```
DhairyaSethi: this will remain donate here

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -19,7 +18,7 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
       )
     );
     vm.prank(address(spoke1));
-    hub.add(daiAssetId, amount, address(spoke1));
+    hub.add(daiAssetId, amount, makeAddr('randomUser'));
```
DhairyaSethi, line 21:
```suggestion
    hub.add(daiAssetId, amount, makeAddr('randomSpoke'));
```
don't want to give the idea hub interactions are meant for users for the reader (but obv check for it)

src/contracts/Spoke.sol
```diff
@@ -993,13 +993,13 @@ contract Spoke is ISpoke, Multicall {
...
   /// @return collateralToLiquidate The amount of collateral to liquidate.
-  /// @return liquidationProtocolFeeAmount The amount of protocol fee.
   function _executeLiquidationCall(
     DataTypes.Reserve storage collateralReserve,
     DataTypes.Reserve storage debtReserve,
     address[] memory users,
-    uint256[] memory debtsToCover
-  ) internal returns (address, address, uint256, uint256, uint256) {
+    uint256[] memory debtsToCover,
+    address liquidator
```
DhairyaSethi, line 1028: why are we adding this again?

snapshots/Spoke.Operations.json
```diff
@@ -1,17 +1,17 @@
...
-  "supply: 0 debt, collateralDisabled": "172003",
-  "supply: 0 debt, collateralEnabled": "171979",
-  "supply: 1 debt": "172027",
-  "supply: 2 debt": "172038",
-  "supply: 3 debt": "171991",
-  "updateUserRiskPremium": "174207",
-  "updateUserRiskPremium: 2 debts": "221738",
+  "borrow": "491792",
+  "liquidationCall: full": "464147",
+  "liquidationCall: partial": "645418",
```
DhairyaSethi: test includes non zero fee receiver, right?

DhairyaSethi: and current behaviour is we revert with zero fee receiver? pls document

src/contracts/Spoke.sol
yan-man: passing the liquidator you mean? bc we had discussed before that internal methods shouldn't call msg.sender directly

snapshots/Spoke.Operations.json
yan-man: assuming this will be implemented: https://github.com/aave/aave-v4/issues/480 so there will never be a zero fee receiver

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
yan-man: but this will be a user during normal flow bc it's the address to pull tokens from

tests/invariant/LiquidityHubHandler.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/35bf34f370caae27d19154d79af835b3868d2311

src/contracts/LiquidityHub.sol
miguelmtzinf: > thinking more, one caveat w Remove is it signals the receiver param receiving underlying assets whereas here it'd be the feeReceiver receiving shares (the addition is signalled by AccrueFee)

wdym? the `Remove` does not have a receiver of underlying param

miguelmtzinf: I think we should use a `Remove` for the caller spoke, and thinking that we may want to use `Add` for whatever fee we wanna accrue (here and in `AssetLogic.accrue`). Otherwise, indexer wouldn't work well in the sense that fee receiver has more supplied assets than the data says.
Not against emitting a dedicated event to reflect FeeAccrual too.

src/contracts/LiquidityHub.sol
```diff
@@ -268,6 +268,31 @@ contract LiquidityHub is ILiquidityHub {
...
+  function payFeeWithExistingLiquidity(
+    uint256 assetId,
+    uint256 feeShares
+  ) external returns (uint256) {
+    // TODO: authorization - only spokes
+    _validatePayFeeWithExistingLiquidity(feeShares);
+
+    DataTypes.Asset storage asset = _assets[assetId];
+    DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
+    DataTypes.SpokeData storage feeReceiverSpoke = _spokes[assetId][asset.config.feeReceiver];
```
miguelmtzinf: i think `feeReceiver` is just fine

src/contracts/LiquidityHub.sol
```diff
@@ -268,6 +268,31 @@ contract LiquidityHub is ILiquidityHub {
...
+
+    asset.accrue(assetId, feeReceiverSpoke);
+
+    uint256 payableShares = spoke.suppliedShares;
+    require(feeShares <= payableShares, SuppliedSharesExceeded(payableShares));
```
miguelmtzinf: do we do the check in asset terms so we can make use of `SuppliedAmountExceeded`?

src/contracts/LiquidityHub.sol
```diff
@@ -268,6 +268,31 @@ contract LiquidityHub is ILiquidityHub {
     require(premiumDebtAfter + realizedPremiumTaken - premiumDebtBefore <= 2, InvalidDebtChange());
   }
 
+  /// @inheritdoc ILiquidityHub
+  function payFeeWithExistingLiquidity(
```
miguelmtzinf: I think `payFee` is just fine. This name is smt I gave to give an idea of the implementation

src/contracts/Spoke.sol
```diff
@@ -1095,31 +1095,28 @@ contract Spoke is ISpoke, Multicall {
...
-      vars.withdrawnShares = collateralReserveHub.remove(
+      // expected total withdrawn shares includes liquidation fee
+      vars.withdrawnShares = collateralReserveHub.convertToSuppliedSharesUp(
         vars.collateralAssetId,
-        vars.collateralToLiquidate + vars.liquidationProtocolFeeAmount,
-        address(this) // must be sent to spoke first before distributing to treasury/liquidator
+        vars.liquidationFeeAmount + vars.collateralToLiquidate
+      );
+      // remove collateral, send liquidated collateral directly to liquidator
+      vars.withdrawnLiquidatorShares = collateralReserveHub.remove(
```
miguelmtzinf: cannot we do this at the end? so we have sum of liquidated collateral and sum of liquidation fee that gets `remove`d and `payFee`d at the end

tests/unit/LiquidityHub/LiquidityHub.PayFeeWithExistingLiquidity.t.sol
```diff
@@ -0,0 +1,170 @@
...
+    uint256 feeShares = hub.getSpokeSuppliedShares(daiAssetId, address(spoke1));
+
+    // supply ex rate increases due to interest
+    assertGt(hub.getSpokeSuppliedAmount(daiAssetId, address(spoke1)), feeShares);
+
+    vm.expectRevert(
+      abi.encodeWithSelector(ILiquidityHub.SuppliedSharesExceeded.selector, feeShares)
+    );
+    vm.prank(address(spoke1));
+    hub.payFeeWithExistingLiquidity(daiAssetId, feeShares + 1);
```
miguelmtzinf: hmm does make more sense to calc this amount in this test case in particular by `hub.getSpokeSuppliedAssets().toShares + 1`?

miguelmtzinf: thinking that this test is a bit dumb given that feeShares does not change overtime

tests/unit/LiquidityHub/LiquidityHub.PayFeeWithExistingLiquidity.t.sol
```diff
@@ -0,0 +1,170 @@
...
+    );
+
+    assertEq(spokeSharesAfter, spokeSharesBefore - feeShares, 'spoke supplied shares after');
+    assertEq(
+      feeReceiverSharesAfter,
+      feeReceiverSharesBefore + feeShares,
+      'fee receiver supplied shares after'
+    );
+  }
+  function test_payFeeWithExistingLiquidity_revertsWith_InvalidFeeShares() public {
```
miguelmtzinf: newline

tests/unit/LiquidityHub/LiquidityHub.PayFeeWithExistingLiquidity.t.sol
```diff
@@ -0,0 +1,170 @@
...
+    );
+  }
+
+  function test_payFeeWithExistingLiquidity_fuzz_with_interest(
+    uint256 addAmount,
+    uint256 feeShares,
+    uint256 skipTime
+  ) public {
+    addAmount = bound(addAmount, 1, MAX_SUPPLY_AMOUNT);
+    skipTime = bound(skipTime, 1, MAX_SKIP_TIME);
```
miguelmtzinf: can the above one be replaced if we change to 0 as min skipTime?

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -38,7 +38,7 @@ contract SpokeLiquidationBase is SpokeBase {
     uint256 liquidationBonus;
     uint256 collateralAssetId;
     uint256 debtAssetId;
-    uint256 liquidationProtocolFee;
+    uint256 liquidationFeeBPS;
```
miguelmtzinf: using `liquidationFee` in src

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -109,7 +110,7 @@ contract SpokeLiquidationBase is SpokeBase {
     liqConfig = _bound(liqConfig);
     liqBonus = bound(liqBonus, MIN_LIQUIDATION_BONUS, MAX_LIQUIDATION_BONUS);
     desiredHf = bound(desiredHf, 0.1e18, HEALTH_FACTOR_LIQUIDATION_THRESHOLD - 0.01e18);
-    liquidationProtocolFee = bound(liquidationProtocolFee, 0, 100_00);
+    liquidationFee = bound(liquidationFee, 0, 100_00);
```
miguelmtzinf: should we fetch from PercentageMath.max?

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -170,12 +171,22 @@ contract SpokeLiquidationBase is SpokeBase {
...
-    // TODO: update when treasury accounting is done
-    vm.recordLogs();
+    state.liquidationFeeShares =
+      hub.convertToSuppliedSharesUp(
+        state.collateralReserve.assetId,
+        state.collToLiq + state.liquidationFeeAmount
+      ) -
+      hub.convertToSuppliedSharesUp(state.collateralReserve.assetId, state.collToLiq);
+
+    // due to restore donation, exact feeShares args may be inaccurate
```
miguelmtzinf: there is a `calculateEffectiveSupplyAmount` or smt like that. wondering if would make sense to create one for the restore that we can use here

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -241,7 +265,7 @@ contract SpokeLiquidationBase is SpokeBase {
         finalHf,
         _getCloseFactor(spoke),
         _approxRelFromBps(20),
-        'HF matches closeFactor within 0.1%'
+        'HF matches closeFactor within 0.25%'
```
miguelmtzinf: is it not 20%? based on `_approxRelFromBps(20),`

src/contracts/LiquidityHub.sol
yan-man: yes we can do this instead, then can also emit amount and shares in `Remove` event

https://github.com/aave/aave-v4/commit/897903f8bb6e81094596859c3e07da7d05cd489c

tests/unit/LiquidityHub/LiquidityHub.PayFeeWithExistingLiquidity.t.sol
yan-man: no bc skip time must be > 0 in order to increase the supply ex rate. 

also wdym in terms of what above would be replaced with?

tests/unit/LiquidityHub/LiquidityHub.PayFeeWithExistingLiquidity.t.sol
yan-man: yes true, but now with re-using SuppliedAmountExceeded event, the fee amount will change with interest in the event arg

src/contracts/Spoke.sol
yan-man: i think the problem with doing the aggregated action at the end for both remove/restore on LH is the rounding - doing an action on round(totalAmount) won't equal round(amount1) + round(amount2) +... so accounting will be inaccurate. We don't see that issue atm with just 1 liquidation, but it should show up with parallel liqs. 

It only works with `payFee` if we use shares directly and dont round

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/897903f8bb6e81094596859c3e07da7d05cd489c

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/897903f8bb6e81094596859c3e07da7d05cd489c

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/897903f8bb6e81094596859c3e07da7d05cd489c

tests/unit/LiquidityHub/LiquidityHub.PayFeeWithExistingLiquidity.t.sol
yan-man: https://github.com/aave/aave-v4/commit/897903f8bb6e81094596859c3e07da7d05cd489c

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/commit/897903f8bb6e81094596859c3e07da7d05cd489c

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/commit/897903f8bb6e81094596859c3e07da7d05cd489c

src/contracts/LiquidityHub.sol
miguelmtzinf: I believe we should emit an `Add` event too, for the fee receiver spoke (an do the same at accrual)

src/contracts/LiquidityHub.sol
```diff
@@ -276,6 +276,31 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     require(premiumDebtAfter + realizedPremiumTaken - premiumDebtBefore <= 2, InvalidDebtChange());
   }
 
+  /// @inheritdoc ILiquidityHub
+  function payFee(uint256 assetId, uint256 feeShares) external returns (uint256) {
```
miguelmtzinf: question: what d be the difference between this function and a generic function `move` that just updates accounting between 2 spokes. Is the dedicated event for accrual the only difference? :thinking:

src/contracts/LiquidityHub.sol
```diff
@@ -276,6 +276,31 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+    // TODO: authorization - only spokes
```
```suggestion
    require(_spokes[assetId][msg.sender].config.active, SpokeNotActive());
```

src/contracts/LiquidityHub.sol
```diff
@@ -276,6 +276,31 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+    _validatePayFee(feeShares);
+
+    DataTypes.Asset storage asset = _assets[assetId];
+    DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
+    DataTypes.SpokeData storage feeReceiver = _spokes[assetId][asset.config.feeReceiver];
+
+    asset.accrue(assetId, feeReceiver);
+
+    uint256 payableShares = spoke.suppliedShares;
+    uint256 payableAmount = asset.toSuppliedAssetsDown(payableShares);
```
miguelmtzinf: `spokeSuppliedShares` and `spokeSuppliedAssets`
seems clearer to me

src/contracts/LiquidityHub.sol
```diff
@@ -532,4 +557,9 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     if (b >= 0) return a + uint256(b);
     return a - uint256(-b);
   }
+
+  function _validatePayFee(uint256 feeShares) internal pure {
+    // TODO: validate valid asset
```
miguelmtzinf, line 560: should we block this if the asset is not active?
unclear to me how we wanna work with spoke.active

src/contracts/Spoke.sol
miguelmtzinf: Sounds doable... I d like to try and do 1 single aggregated action at the end. Can you create a ticket for this please?

src/libraries/types/DataTypes.sol
```diff
@@ -166,14 +167,13 @@ library DataTypes {
...
-    uint256 totalLiquidationProtocolFeeShares;
+    uint256 totalliquidationFeeShares;
     int256 totalUserDebtPremiumDrawnSharesDelta;
     int256 totalUserDebtPremiumOffsetDelta;
     int256 totalUserCollateralPremiumDrawnSharesDelta;
     int256 totalUserCollateralPremiumOffsetDelta;
     uint256 totalDebtToLiquidate;
     uint256 usersLength;
-    uint256 newUserSuppliedShares;
+    uint256 withdrawnLiquidatorShares;
```
miguelmtzinf: got confused for a moment because liquidator does not withdraw anything. Wondering if we should rename to `liquidatedSuppliedShares` or similar

tests/Base.t.sol
```diff
@@ -940,6 +940,18 @@ abstract contract Base is Test {
     liquidityHub.updateAssetConfig(assetId, assetConfig);
   }
 
+  function updateAssetFeeReceiver(
```
miguelmtzinf: i thought this already exists

tests/Base.t.sol
```diff
@@ -978,18 +990,18 @@ abstract contract Base is Test {
     assertEq(spoke.getReserve(reserveId).config.liquidationBonus, newLiquidationBonus);
   }
 
-  function updateLiquidationProtocolFee(
+  function updateLiquidationFee(
     ISpoke spoke,
     uint256 reserveId,
-    uint256 newLiquidationProtocolFee
+    uint256 newliquidationFee
```
```suggestion
    uint256 newLiquidationFee
```
please search for similar occurrences to fix them

tests/Base.t.sol
```diff
@@ -978,18 +990,18 @@ abstract contract Base is Test {
...
   ) internal {
     DataTypes.ReserveConfig memory config = spoke.getReserve(reserveId).config;
-    config.liquidationProtocolFee = newLiquidationProtocolFee;
+    config.liquidationFee = newliquidationFee;
 
     vm.prank(SPOKE_ADMIN);
     spoke.updateReserveConfig(reserveId, config);
 
-    assertEq(spoke.getReserve(reserveId).config.liquidationProtocolFee, newLiquidationProtocolFee);
+    assertEq(spoke.getReserve(reserveId).config.liquidationFee, newliquidationFee);
```
miguelmtzinf: i think it d be fine just to remove, but all good

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -258,22 +295,18 @@ contract SpokeLiquidationBase is SpokeBase {
     }
   }
 
-  // todo: utilize treasury accounting to assert protocol fee
-  function _assertProtocolFeeEarned(
+  function _assertliquidationFeeEarned(
```
miguelmtzinf: same

tests/unit/LiquidityHub/LiquidityHub.PayFee.t.sol
```diff
@@ -0,0 +1,179 @@
...
+    Utils.add({
+      hub: hub,
+      assetId: daiAssetId,
+      spoke: address(spoke1),
+      amount: addAmount,
+      user: alice,
+      to: address(spoke1)
+    });
+
+    _supplyAndDrawLiquidity({
```
miguelmtzinf: i d use the combo `addLiquidity` and `drawLiquidity` instead.. way cleaner and by random spoke

src/contracts/LiquidityHub.sol
miguelmtzinf: non-zero fee receiver is just an address on the Asset config struct, but does not enforce that existence of the corresponding fee receiver spoke. We are using `lastUpdateTimestamp` for existence for now.

tests/unit/LiquidityHub/LiquidityHub.PayFeeWithExistingLiquidity.t.sol
miguelmtzinf: `test_payFee_fuzz` and `test_payFee_fuzz_with_interest` are almost identical. I think we can just simplify and have only one where skipTime can be 0 as well.

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
miguelmtzinf: maybe better to get the number out of the msg

tests/Base.t.sol
yan-man: I just saw the _get, but didn't see an update helper

src/contracts/LiquidityHub.sol
yan-man: this is basically a specific subset of that generic function move, except that it only goes to feeReceiver. Besides the event, we just need some validation on valid recipient spoke, and its supply caps. should I implement a generic move too?

src/contracts/LiquidityHub.sol
DhairyaSethi: i think `FeeAccrual` signals that, `Add` on top seems excessive, would do one or the other then; with add i'm concerned about user's seeing an add event to a random (from their pov) spoke on their actions, so `FeeAccrue` can act as an exception `Add` for the treasury spoke

src/contracts/LiquidityHub.sol
DhairyaSethi: access control is in now

lib/forge-std
DhairyaSethi, line 1: from the diffs doesn't look like this is needed but nbd

src/contracts/LiquidityHub.sol
yan-man: ok will use dedicated event for now - `Remove, AccrueFees`.

src/interfaces/ILiquidityHub.sol
```diff
@@ -52,6 +52,7 @@ interface ILiquidityHub is IAccessManaged {
     uint256 realizedPremiumAdded,
     uint256 realizedPremiumTaken
   );
+  event AccrueFees(uint256 indexed assetId, uint256 shares);
```
DhairyaSethi, line 55: let's just add this to AssetLogic.accrue as well now? (and resolve that todo together)

src/contracts/LiquidityHub.sol
yan-man: see dupe comment https://github.com/aave/aave-v4/pull/467#discussion_r2175451357

src/interfaces/ILiquidityHub.sol
yan-man: see https://github.com/aave/aave-v4/pull/467#discussion_r2175621353

src/contracts/LiquidityHub.sol
yan-man: @DhairyaSethi on second thought don't want to touch `assetLogic` atm, so that the scope of this PR doesn't expand, needing to test event emission on generic accruals. Also want to wait for `_assertEventNotEmitted` from https://github.com/aave/aave-v4/pull/491/files

created ticket for it here: https://github.com/aave/aave-v4/issues/494

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/issues/495

tests/unit/LiquidityHub/LiquidityHub.PayFee.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/0d61e6cbc228071dd9df7f32190e9249b680b0d0

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/0d61e6cbc228071dd9df7f32190e9249b680b0d0

tests/Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/0d61e6cbc228071dd9df7f32190e9249b680b0d0

src/libraries/types/DataTypes.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/0d61e6cbc228071dd9df7f32190e9249b680b0d0

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/0d61e6cbc228071dd9df7f32190e9249b680b0d0

src/contracts/LiquidityHub.sol
yan-man: added into `_validatePayFee` method https://github.com/aave/aave-v4/pull/467/commits/0d61e6cbc228071dd9df7f32190e9249b680b0d0

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/0d61e6cbc228071dd9df7f32190e9249b680b0d0

tests/unit/LiquidityHub/LiquidityHub.PayFeeWithExistingLiquidity.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/0d61e6cbc228071dd9df7f32190e9249b680b0d0

src/contracts/LiquidityHub.sol
yan-man: added ticket for it https://github.com/aave/aave-v4/issues/497

but also for frozen/paused same q. I think 1 of the flags should block any spoke action altogether (including payFee).

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: updated with conditional assertions - with the same coll/debt asset involved in the liquidation, `restore` creates wei rounding differences in calc of total debt that slightly change the ex rate and feeShare calc, making the exact value difficult to calc precisely

src/contracts/LiquidityHub.sol
yan-man: created issue here: https://github.com/aave/aave-v4/issues/500

src/contracts/LiquidityHub.sol
yan-man: given that this same pattern is used across `add`, `remove`, etc. I think if we want to check existence it should happen in `accrue` in asset logic

created ticket for that here: https://github.com/aave/aave-v4/issues/501

tests/Base.t.sol
yan-man: left as is, but did you mean the assertion altogether? would really like to have these across all the helpers as a sanity check that the change was applied properly

src/contracts/LiquidityHub.sol
```diff
@@ -276,6 +276,30 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+
+    uint256 spokeSuppliedShares = spoke.suppliedShares;
+    uint256 spokeSuppliedAssets = asset.toSuppliedAssetsDown(spokeSuppliedShares);
+    uint256 feeAmount = asset.toSuppliedAssetsDown(feeShares);
+    require(feeAmount <= spokeSuppliedAssets, SuppliedAmountExceeded(spokeSuppliedAssets));
+
+    spoke.suppliedShares = spokeSuppliedShares - feeShares;
+    feeReceiver.suppliedShares += feeShares;
+
+    emit Remove(assetId, msg.sender, feeShares, feeAmount);
```
DhairyaSethi, line 298: wdyt about emitting shares on the hub & assets on the spoke instead? that choice was still arbitrary iirc & saves us the unnecessary conversion here just for the event

src/contracts/LiquidityHub.sol
```diff
@@ -276,6 +276,30 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+    emit AccrueFees(assetId, feeShares);
+
+    return feeShares;
```
DhairyaSethi: we never update this, don't think returning is needed

src/contracts/LiquidityHub.sol
yan-man: refer to https://github.com/aave/aave-v4/pull/467#discussion_r2173679138

Miguel wanted to re-use existing `SuppliedAmountExceeded`  error so the conversion was for that as well 
but also dont know if it makes sense to create a new Remove method just to split up the shares/assets emission

DhairyaSethi: i mean just for the event (which is malleable) & error we shouldn't have to convert tbh

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/0ae241a259173ea0c50024fe0b4baa05f9d298bf

src/contracts/LiquidityHub.sol
yan-man: wdym by event is malleable, you mean using a diff one - if using `Remove` as is we need shares/amount, no?

src/contracts/LiquidityHub.sol
```diff
@@ -277,6 +277,28 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+
+    uint256 spokeSuppliedShares = spoke.suppliedShares;
+    uint256 spokeSuppliedAssets = asset.toSuppliedAssetsDown(spokeSuppliedShares);
+    uint256 feeAmount = asset.toSuppliedAssetsDown(feeShares);
+    require(feeAmount <= spokeSuppliedAssets, SuppliedAmountExceeded(spokeSuppliedAssets));
+
+    spoke.suppliedShares = spokeSuppliedShares - feeShares;
+    feeReceiver.suppliedShares += feeShares;
+
+    emit Remove(assetId, msg.sender, feeShares, feeAmount);
```
miguelmtzinf, line 298: @yan-man we need to emit an `Add` event here

tests/unit/LiquidityHub/LiquidityHub.PayFee.t.sol
```diff
@@ -0,0 +1,127 @@
...
+      assetId: daiAssetId,
+      supplyUser: bob,
+      supplySpoke: address(spoke2),
+      supplyAmount: 100e18,
+      drawUser: alice,
+      drawSpoke: address(spoke3),
+      drawAmount: 100e18,
+      skipTime: skipTime
+    });
```
miguelmtzinf:
```suggestion
    _addLiquidity(daiAssetId, 100e18);
    _drawLiquidity(daiAssetId, 100e18, true);
```

tests/unit/LiquidityHub/LiquidityHub.PayFee.t.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    uint256 spokeSharesBefore = hub.getSpokeSuppliedShares(daiAssetId, address(spoke1));
+
+    // supply ex rate increases due to interest
+    assertGe(
```
miguelmtzinf, line 82: we are doing this across the test suite, and would be nice to have a specific helper/assert for this.
can you add it or add this item for the refactor issue #430 please

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
miguelmtzinf: did you create a ticket for this at the end?

src/contracts/LiquidityHub.sol
yan-man: as discussed on slack will resolve here: https://github.com/aave/aave-v4/issues/494

tests/unit/LiquidityHub/LiquidityHub.PayFee.t.sol
yan-man: added to issue - but difficult thing about that helper is that we typically want to check supply ex rate before/after an action, so can't just call it in 1 place

tests/unit/LiquidityHub/LiquidityHub.PayFee.t.sol
yan-man: https://github.com/aave/aave-v4/pull/467/commits/2d676612a6bb546d98136f160e0d14eacbd9fa16

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: ah my bad initially thought you were creating. Created it here https://github.com/aave/aave-v4/issues/503

PR #468, opened 2025-06-22
feat: rm reserve accounting on spoke

DhairyaSethi:
Removes reserve accounting on the spoke level to save unnecessary writes. Spoke will fetch reserve accounting from hub.
**CAVEAT**: We cannot have two reserve's registered with the same asset(id) configured on the same hub. Potential workaround is listing the asset again in the Hub, so the new reserve gets linked to the new asset.

**TODO**: opt _refreshPremiumDebt method since the sloads for config.hub & reserveId can be omitted. Out of spoke for this PR bc of stack too deep and some refactor around premium updates in general is pending.

Followup needed: https://github.com/aave/aave-v4/issues/579#issuecomment-3139208628


yan-man:
> CAVEAT: We cannot have two reserve's registered for the same asset(id) on the same hub

hm trying to think if there is a use case for this as a requirement. would this affect the investment manager impl that you preferred @DhairyaSethi

src/contracts/Spoke.sol
```diff
@@ -434,11 +421,6 @@ contract Spoke is ISpoke, Multicall {
     return baseDebt + premiumDebt;
   }
 
-  function getReserveRiskPremium(uint256 reserveId) external view returns (uint256) {
-    DataTypes.Reserve storage reserve = _reserves[reserveId];
-    return reserve.premiumDrawnShares.rayDiv(reserve.baseDrawnShares); // trailing
-  }
```
yan-man, line 475: we can never read a reserve risk premium anymore, as it doesnt exist on hub. do you think we'll never need it

src/contracts/Spoke.sol
```diff
@@ -400,16 +387,16 @@ contract Spoke is ISpoke, Multicall {
     return baseDebt + premiumDebt;
   }
 
+  /// @dev We do not differentiate between duplicate reserves (assetId) on the same hub
```
CheyenneAtapour, line 421: Hmm, ok so if we want to relist the same reserve on a spoke, we must also relist the asset on the hub basically, correct?

src/contracts/Spoke.sol
CheyenneAtapour: I think this is something that can be computed by backend if needed for rate calculations, etc.

DhairyaSethi: yes and it can be added on the hub or data provider since the underlying relevant fields (premium & shares) are still exposed on the hub with getSpokeData

src/contracts/Spoke.sol
DhairyaSethi: yeah or have the asset listed on another hub, haven't added validation for this yet bc it'll be o(n) unless we add an asset => reserve reverse map

tests/Base.t.sol
```diff
@@ -977,23 +991,26 @@ abstract contract Base is Test {
...
   }
 
   function updateCollateralFlag(ISpoke spoke, uint256 reserveId, bool newCollateralFlag) internal {
-    DataTypes.Reserve memory reserveData = spoke.getReserve(reserveId);
-    reserveData.config.collateral = newCollateralFlag;
-    spoke.updateReserveConfig(reserveId, reserveData.config);
+    DataTypes.ReserveConfig memory config = spoke.getReserveConfig(reserveId);
+    config.collateral = newCollateralFlag;
+    vm.prank(SPOKE_ADMIN);
+    spoke.updateReserveConfig(reserveId, config);
```
yan-man: on these, let's do an assertion on reserveConfig after to make sure it matches the updated value

tests/Base.t.sol
```diff
@@ -977,23 +991,26 @@ abstract contract Base is Test {
...
     uint256 reserveId,
     bool newBorrowable
   ) internal {
-    DataTypes.Reserve memory reserveData = spoke.getReserve(reserveId);
-    reserveData.config.borrowable = newBorrowable;
-    spoke.updateReserveConfig(reserveId, reserveData.config);
+    DataTypes.ReserveConfig memory config = spoke.getReserveConfig(reserveId);
+    config.borrowable = newBorrowable;
+    vm.prank(SPOKE_ADMIN);
+    spoke.updateReserveConfig(reserveId, config);
```
yan-man: same

tests/Base.t.sol
```diff
@@ -141,6 +141,20 @@ abstract contract Base is Test {
     uint256 liquidityPremium;
   }
 
+  struct ReservePosition {
```
yan-man: given only the accounting on hub remains, should we rename this to `SpokePosition` to align with hub

tests/Base.t.sol
```diff
@@ -1109,6 +1126,36 @@ abstract contract Base is Test {
...
+    return getReservePosition(spoke, reserveIdFn(spoke));
+  }
+
+  function getReservePosition(
+    ISpoke spoke,
+    uint256 reserveId
+  ) internal view returns (ReservePosition memory) {
+    uint256 assetId = spoke.getReserve(reserveId).assetId;
+    DataTypes.SpokeData memory spokeData = hub.getSpoke(assetId, address(spoke));
+    (uint256 baseDebt, uint256 premiumDebt) = spoke.getReserveDebt(reserveId);
```
yan-man: should we also rm spoke getReserveDebt and just refer to hub for spoke debt data?

tests/Base.t.sol
DhairyaSethi: hmm would prefer to do it with the entire test refactor tbh bc this then would require changing a lot of names in the tests

tests/Base.t.sol
DhairyaSethi: think its good to keep tbh

tests/Base.t.sol
DhairyaSethi: sure, will do it for all other helpers in this file (i was just cleaning these two methods) and good to note in the tests refactor PR so we propagate it for consistency

src/contracts/Spoke.sol
miguelmtzinf: we could also keep this getter and fetch from the hub right?

DhairyaSethi: feels more in scope for the data provider but yes

src/contracts/Spoke.sol
miguelmtzinf: this is bad...
cannot we just do `inputHub.getAsset(inputAssetId).lastUpdateTimestamp == address(0)`?

DhairyaSethi: hub doesn't have any knowledge of which reserves are listed against its assets 
the validation in question is reserve.(hub, assetId) tuple is unique in the reserveList 
thinking more, will add an O(n) validation on addReserve since hub, assetId are static

miguelmtzinf: updated the PR desc to add the workaround of adding the asset again on the hub.

miguelmtzinf: ah you are right. My suggestion was checking the asset exists, not if any of the Spoke reserves are linked to the Hub asset.
I'm sorry to say we need to add a validation, even if it's O(n)... and this is a big con

DhairyaSethi: as agreed with Emilio there's few use cases for that tbh, especially since IR strategy lives on the hub

snapshots/Spoke.Operations.json
miguelmtzinf, line 1: Saving ~80k in borrow, but not crazy on the rest.... 20-40k
And this is highly impacted by current non-opt storage layout....
I think savings will be way less relevant once we pack everything.

src/contracts/Spoke.sol
miguelmtzinf: can you provide one of the use cases pls? or add directly to the desc please

snapshots/Spoke.Operations.json
DhairyaSethi: the writes this removes (balances) would each take half a slot each after packing so that wouldn't diminish the returns we get here 

in general tho i concur, but the lost config (multiple reserves for the same hub, asset) is worth it to me bc i dont see a use case

yan-man:
pls link this to an issue

miguelmtzinf:
As agreed, we will get back to this once structs data are more optimized (mainly packed), so we can get a better sense of the gas savings of this change.

DhairyaSethi:
yes @CheyenneAtapour

PR #470, opened 2025-06-24
feat: New IR Interface & correct borrow rate refresh in the hub

### LINKED ISSUE #469: Wrong `totalDebt` passed to the Interest Rate Strategy
In V3 and V4, the Interest Rate Strategy takes in `totalDebt`, `availableLiquidity`, `liquidityAdded`, and `liquidityTaken`. `availableLiquidity`  is not updated with the ongoing action, since it is reflected through `liquidityAdded` and `liquidityTaken` -- this is true in both V3 and V4 and gives flexibility for potential future strategies. However, the ongoing action should be reflected in the `totalDebt` -- see [this](https://github.com/aave-dao/aave-v3-origin/blob/496c6eaf9056a25729a28c9e98b7757f975dc54a/src/contracts/protocol/libraries/logic/BorrowLogic.sol#L101) and [this](https://github.com/aave-dao/aave-v3-origin/blob/496c6eaf9056a25729a28c9e98b7757f975dc54a/src/contracts/protocol/libraries/logic/ReserveLogic.sol#L169) in V3. This discrepancy is needed because we cannot distinguish between `add` and `restore`, or between `remove` and `draw` by only looking at `liquidityAdded` and `liquidityTaken`. 

The issue is that V4 does not update `baseDrawnShares` before updating the borrow rate in `draw` and `restore` operations. This is very problematic, as seen in the following example:
- user A adds 1000 USDC --> no debt, usage ratio = 0%
- user B draws 1000 USDC --> still no debt (given that `baseDrawnShares` was not updated), usage ratio still 0%. Instead, usage ration should be 100% and borrow rate at max possible rate.

This issue can be observed by refreshing the borrow rate just before the end of the `draw` and `restore` functions. Doing `asset.updateBorrowRate({assetId: assetId, liquidityAdded: 0, liquidityTaken: 0});` after any operation should never change the asset's borrow rate, which is not true in the current implementation.

Next steps:
1. Fix the bug either by:
    1. update `baseDrawnShares` before updating the borrow rate in `draw` and `restore`.
    2. Pass 2 more variables to the IR Strategy: `debtAdded` and `debtTaken` 
    3. Replace `liquidityAdded` and `liquidityTaken` with `liquidityDelta`, and add `debtDelta`
4. Add invariant in util functions, where we check that `asset.updateBorrowRate({assetId: assetId, liquidityAdded: 0, liquidityTaken: 0});` has no effect on the asset's borrow rate after the operation. We should also add some explicit fuzz tests

avniculae:
Close #469 

- Changes the interest rate strategy interface; to calculate the interest rate, the strategy requires the **updated** available liquidity, base and premium debt values. While the current IR strategy does not use premium debt, it is passed in case we want to use it in the future
- Fixes the delta by which availableLiquidity is updated when sent to the IR strategy in the `restore` function
- Adjusts existing tests
- Adds borrow rate invariant in the targeted tests
- Asserts all arguments sent to the IR strategy contract by using `expectCall` in the targeted tests

### DIFF
--- a/tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
@@ -518,9 +518,9 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     );
 
     vm.prank(alice);
-    IERC20(hub.getAsset(daiAssetId).underlying).approve(address(hub), supplyAmount);
+    tokenList.dai.approve(address(hub), supplyAmount);
 
-    vm.expectEmit(hub.getAsset(daiAssetId).underlying);
+    vm.expectEmit(address(tokenList.dai));
     emit IERC20.Transfer(alice, address(hub), supplyAmount);
     vm.expectEmit(address(hub));
     emit ILiquidityHub.Add(daiAssetId, address(spoke2), shares, supplyAmount);


src/contracts/LiquidityHub.sol
```diff
@@ -188,17 +188,15 @@ contract LiquidityHub is ILiquidityHub {
...
     require(suppliedShares != 0, InvalidSharesAmount());
-
-    asset.availableLiquidity += amount;
     asset.suppliedShares += suppliedShares;
-
     spoke.suppliedShares += suppliedShares;
 
+    asset.updateBorrowRate({liquidityAdded: amount, liquidityTaken: 0});
+    asset.availableLiquidity += amount;
+
```
avniculae: these changes are not needed, but I think it's good to align ordering of operations to the other actions

src/contracts/LiquidityHub.sol
```diff
@@ -408,6 +402,10 @@ contract LiquidityHub is ILiquidityHub {
     return _assets[assetId].toDrawnSharesDown(assets);
   }
 
+  function convertToDrawnSharesUp(uint256 assetId, uint256 assets) external view returns (uint256) {
```
avniculae, line 392: this fn is useful to compute number of shares that will be drawn

src/contracts/LiquidityHub.sol
yan-man: which change are you referring to which is not needed?

DhairyaSethi: @yan-man the reorder on add and remove, they are not needed bc these methods don't touch debt so they are technically fine right now

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
```diff
@@ -450,4 +455,48 @@ contract LiquidityHubDrawTest is LiquidityHubBase {
     vm.prank(address(spoke1));
     hub.draw({assetId: daiAssetId, amount: drawAmount, to: address(spoke1)});
   }
+
+  function _checkedDraw(address caller, uint256 assetId, uint256 amount, address to) internal {
```
CheyenneAtapour: Maybe a /// @dev comment describing what this helper is supposed to do would make it clearer

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -847,4 +770,43 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
       skip(randomizer(1 days, 365 days));
     }
   }
+
+  function _checkedAdd(address caller, uint256 assetId, uint256 amount, address from) public {
```
CheyenneAtapour: Same for this, comment to describe this function

avniculae: [7396cba](https://github.com/aave/aave-v4/pull/470/commits/7396cba6e1c9d319592c27a6f9a4bd3f28a41175)

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
avniculae: [7396cba](https://github.com/aave/aave-v4/pull/470/commits/7396cba6e1c9d319592c27a6f9a4bd3f28a41175)

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
```diff
@@ -231,7 +232,8 @@ contract SpokeMultipleHubIsolationModeTest is SpokeMultipleHubBase {
 
     // Bob still cannot borrow asset B from the new hub because there is no liquidity
     vm.expectRevert(abi.encodeWithSelector(ILiquidityHub.NotAvailableLiquidity.selector, 0));
-    Utils.borrow(newSpoke, isolationVars.reserveBId, bob, 1, bob);
+    vm.prank(bob);
+    newSpoke.borrow(isolationVars.reserveBId, 1, bob);
```
yan-man: any reason for doing this specifically instead? with Utils.borrow we can get the `checkBorrowRateInvariant`

avniculae: exactly, and here we don't want to check the invariant bc we expect a revert, and the invariant does not hold if no actions were performed on an asset because we initialise the borrow rate to 0 in `addAsset`.

If we were to initialise it to the base borrow rate (by calling the ir strategy with 0 debt, 0 avail liq, 0 liq added, 0 liq taken in the `addAsset` function), then the invariant would have been true even if no actions were performed. I actually think this is better, because we also test the fact that the ir strategy is not reverting upon asset registration. wdyt @yan-man  @miguelmtzinf @DhairyaSethi ?

tests/Utils.sol
```diff
@@ -23,7 +24,11 @@ library Utils {
     vm.stopPrank();
 
     vm.prank(spoke);
-    return hub.add(assetId, amount, user);
+    uint256 sharesAdded = hub.add(assetId, amount, user);
+
+    checkBorrowRateInvariant(hub, assetId, 'hub.add');
```
DhairyaSethi: this is not the place, especially bc this is not the only post action invariant we will be checking 
it's good we validated this & sanity ran it against the whole test suite, but this would explicitly be run in the dedicated invariant test suite
we can talk about keeping these temporarily here if you feel very strongly about this, tbh i would just add them manually to a few of the dedicated scenario tests of each method, especially on the IR strategy e2e tests that you'll be writing that check all points of the curve through spoke actions 

LH: saw the checked ops, which are great & sufficient for now imo, would add them in more places like^

avniculae: ikik, I wanted to add it as an invariant but noticed that the invariant test suite is work in progress. I added them here temporarily, but I agree: I will undo this check from the utils

avniculae: [773b61f](https://github.com/aave/aave-v4/pull/470/commits/773b61f17e11b16bd8fd279996645365719d99a1)

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
avniculae: reverted [dc1b64b](https://github.com/aave/aave-v4/pull/470/commits/dc1b64b7edecc3dd199fd7eb04b68ca4be5ef081) bc of this https://github.com/aave/aave-v4/pull/470#discussion_r2168815850

avniculae: Irrespective of this revert, questions from above still stands

> If we were to initialise it to the base borrow rate (by calling the ir strategy with 0 debt, 0 avail liq, 0 liq added, 0 liq taken in the addAsset function), then the invariant would have been true even if no actions were performed. I actually think this is better, because we also test the fact that the ir strategy is not reverting upon asset registration. wdyt @yan-man @miguelmtzinf @DhairyaSethi ?

tests/Base.t.sol
```diff
@@ -1755,4 +1755,51 @@ abstract contract Base is Test {
...
+      address(irStrategy),
+      abi.encodeCall(
+        IBasicInterestRateStrategy.calculateInterestRate,
+        (assetId, availableLiquidity, totalDebt, liquidityAdded, liquidityTaken)
+      ),
+      abi.encode(interestRateRay)
+    );
+  }
+
+  function checkBorrowRateInvariant(
```
miguelmtzinf: assert pls

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
```diff
@@ -677,60 +677,60 @@ contract SpokeAccrueInterestTest is SpokeBase {
...
-      vm.mockCall(
-        address(irStrategy),
-        abi.encodeCall(
-          IBasicInterestRateStrategy.calculateInterestRate,
-          (daiAssetId, asset.availableLiquidity, baseDebt, 0, amounts.daiBorrowAmount)
-        ),
-        abi.encode(rates.daiBaseBorrowRate)
-      );
+      uint256 daiBorrowShares = hub.convertToDrawnSharesUp(daiAssetId, amounts.daiBorrowAmount);
+      _mockInterestRateRay({
```
miguelmtzinf, line 682: iirc @DhairyaSethi wanted this, being more explicit. is it just for checking the params used to call the mocked fn ?

tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
```diff
@@ -1071,7 +1071,7 @@ contract SpokeRepayScenarioTest is SpokeBase {
     assertApproxEqAbs(
       spoke1.getUserTotalDebt(_daiReserveId(spoke1), bob),
       bobDaiBefore.totalDebt + action2.borrowAmount,
-      3,
+      4,
```
miguelmtzinf, line 1074: why such big diff now?

src/contracts/LiquidityHub.sol
```diff
@@ -243,15 +239,13 @@ contract LiquidityHub is ILiquidityHub {
...
-
     uint256 drawnShares = asset.toDrawnSharesUp(amount); // non zero since we round up
-
-    asset.availableLiquidity -= amount;
     asset.baseDrawnShares += drawnShares;
-
     spoke.baseDrawnShares += drawnShares;
 
+    asset.updateBorrowRate({liquidityAdded: 0, liquidityTaken: amount});
+    asset.availableLiquidity -= amount;
```
miguelmtzinf, line 203: i do not like it at all. New order of actions is confusing and counter-intuitive....
I d either do
i) all accounting first, `updateBorrowRate()` afterwards with no added params
ii) do `updateBorrowRate()` first with 4 added params for liquidity and debt, accounting afterwards.

This is related to #481

avniculae: I agree, but I thought we agreed to do this temporary fix for now

src/contracts/LiquidityHub.sol
DhairyaSethi: is it needed in src for now? 
regardless we need to figure out a different naming convention like previewAction for these getters

tests/Utils.sol
```diff
@@ -77,6 +78,11 @@ library Utils {
     spoke.supply(reserveId, amount);
   }
 
+  function supply(ITreasurySpoke spoke, uint256 reserveId, address user, uint256 amount) internal {
```
DhairyaSethi: i wouldn't add these tbh, atleast for now 
treasury spoke's supply. withdraw should align with the spoke standard (to be defined)

src/contracts/LiquidityHub.sol
avniculae: no, it's not needed in src now

tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
DhairyaSethi: +1 
this needs to be investigated

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -286,8 +287,8 @@ contract SpokeRiskPremiumEdgeCasesTest is SpokeBase {
       onBehalfOf: bob
     });
 
-    // Deploy liquidity for weth borrow
-    _openSupplyPosition(spoke2, _wethReserveId(spoke2), wethBorrowAmount);
+    // Deploy liquidity for weth borrow such that usage ratio will be at 45%
+    _openSupplyPosition(spoke2, _wethReserveId(spoke2), wethBorrowAmount * 100 / 45);
```
DhairyaSethi:
```suggestion
    _openSupplyPosition(spoke2, _wethReserveId(spoke2), wethBorrowAmount.percentDiv(45_00));
```

tests/unit/Spoke/Spoke.AccrueInterest.t.sol
avniculae: yes, exaclty

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
avniculae: [1ba06f1](https://github.com/aave/aave-v4/pull/470/commits/1ba06f1b21b2b93ec8189acda6a37025111d8ace)

tests/Utils.sol
avniculae: [1ba06f1](https://github.com/aave/aave-v4/pull/470/commits/1ba06f1b21b2b93ec8189acda6a37025111d8ace)

tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
avniculae: [1ba06f1](https://github.com/aave/aave-v4/pull/470/commits/1ba06f1b21b2b93ec8189acda6a37025111d8ace)

tests/Base.t.sol
avniculae: [1ba06f1](https://github.com/aave/aave-v4/pull/470/commits/1ba06f1b21b2b93ec8189acda6a37025111d8ace)

tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
DhairyaSethi: can you link the commit again?think it's lost due to the force push

avniculae: Sorry, replied to the wrong comment. commit above does not address this comment

tests/Base.t.sol
```diff
@@ -1866,12 +1866,13 @@ abstract contract Base is Test {
     require(a.frozen == b.frozen, 'assertEq(AssetConfig): rozen');
     require(a.feeReceiver == b.feeReceiver, 'assertEq(AssetConfig): feeReceiver');
     require(a.liquidityFee == b.liquidityFee, 'assertEq(AssetConfig): liquidityFee');
-    require(address(a.irStrategy) == address(b.irStrategy), 'assertEq(AssetConfig): irStrategy');
+    require(a.irStrategy == b.irStrategy, 'assertEq(AssetConfig): irStrategy');
```
DhairyaSethi: change all of these to assertEqs pls 
and add another sanity assertion for abi encoded a & b matching

tests/Base.t.sol
```diff
@@ -1866,12 +1866,13 @@ abstract contract Base is Test {
...
   }
 
   function assertEq(DataTypes.SpokeConfig memory a, DataTypes.SpokeConfig memory b) internal pure {
     require(a.supplyCap == b.supplyCap, 'assertEq(SpokeConfig): supplyCap');
     require(a.drawCap == b.drawCap, 'assertEq(SpokeConfig): drawCap');
+    require(a.active == b.active, 'assertEq(SpokeConfig): active');
```
DhairyaSethi: same here

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -83,6 +83,12 @@ contract AssetInterestRateStrategyTest is Base {
     );
   }
 
+  function test_setInterestRateData_revertsWith_OnlyLiquidityHub() public {
+    vm.expectRevert(IAssetInterestRateStrategy.OnlyLiquidityHub.selector);
+    vm.prank(makeAddr('randomCaller'));
```
DhairyaSethi, line 88: not sure what convention we use for this but just be consistent

src/contracts/LiquidityHub.sol
yan-man: maybe can rm for now if it's not needed yet

src/contracts/LiquidityHub.sol
```diff
@@ -232,15 +226,15 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-
-    asset.availableLiquidity += totalRestoredAmount;
     asset.baseDrawnShares -= baseDrawnSharesRestored;
-
     spoke.baseDrawnShares -= baseDrawnSharesRestored;
+    uint256 totalRestoredAmount = baseAmount + premiumAmount;
+    asset.availableLiquidity += totalRestoredAmount;
+
+    /// @dev premium debt must be restored in `refreshPremiumDebt` before calling this function
+    asset.updateBorrowRate(assetId);
```
miguelmtzinf, line 237: I still think our approach is incorrect, because the change on premium debt actually happens in `refresh`... as the comment states

(separately, i d change the comment to a normal one: `// premium ...`)

src/contracts/LiquidityHub.sol
miguelmtzinf: i'm keeping it, but we need to come up with an strategy for this

src/interfaces/IBasicInterestRateStrategy.sol
```diff
@@ -10,17 +10,15 @@ interface IBasicInterestRateStrategy {
   /**
    * @notice Calculates the interest rate depending on the asset's state and configurations
    * @param assetId The id of the asset
-   * @param availableLiquidity The available liquidity of the asset
-   * @param totalDebt The total debt of the asset
-   * @param liquidityAdded The amount of liquidity added to the asset
-   * @param liquidityTaken The amount of liquidity taken from the asset
+   * @param availableLiquidity The updated available liquidity of the asset
+   * @param baseDebt The updated base debt of the asset
+   * @param premiumDebt The updated premium debt of the asset
```
miguelmtzinf: someone can ask `updated` based on what? I think it's better to use `current` if so

tests/unit/AssetInterestRateStrategy.t.sol
miguelmtzinf: we are used to fuzz the address as input and assume, but we don't like that no? because of efficiency
what's your preference @DhairyaSethi

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -159,121 +165,75 @@ contract AssetInterestRateStrategyTest is Base {
...
     ) = _generateCalculateInterestRateParams(utilizationRatioRay);
 
     uint256 variableBorrowRate = rateStrategy.calculateInterestRate({
       assetId: mockAssetId,
       availableLiquidity: availableLiquidity,
-      totalDebt: totalDebt,
-      liquidityAdded: liquidityAdded,
-      liquidityTaken: liquidityTaken
+      baseDebt: baseDebt,
+      premiumDebt: vm.randomUint() // unused
```
miguelmtzinf: I think it's better to use a calculated value from `_generateCalculateInterestRateParams`

miguelmtzinf: also good to add a test where we change premium debt and IR stays the same

tests/unit/AssetInterestRateStrategy.t.sol
miguelmtzinf, line 1: do we have a test case covering/showcasing the exact issue you found?

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -261,95 +260,61 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
     });
   }
 
-  function test_supply_single_asset() public {
```
miguelmtzinf, line 264: why removing this entirely?

tests/unit/Spoke/Spoke.Repay.EdgeCases.t.sol
```diff
@@ -372,12 +375,13 @@ contract SpokeRepayEdgeCaseTest is SpokeBase {
     repayAmount = baseRestored + premiumRestored;
 
     assertEq(bobDaiDataAfter.suppliedShares, bobDaiDataBefore.suppliedShares);
-    assertEq(
+    assertApproxEqAbs(
```
miguelmtzinf, line 378: why we need this now?

tests/unit/Spoke/Spoke.Repay.EdgeCases.t.sol
```diff
@@ -372,12 +375,13 @@ contract SpokeRepayEdgeCaseTest is SpokeBase {
...
       spoke1.getUserTotalDebt(_daiReserveId(spoke1), bob),
       bobDaiBefore.totalDebt - baseRestored - premiumRestored,
+      1,
       'bob dai debt final balance'
     );
-    assertEq(bobDaiAfter.premiumDebt, 0, 'bob dai premium debt final balance');
+    assertApproxEqAbs(bobDaiAfter.premiumDebt, bobDaiBefore.premiumDebt - premiumRestored, 1, 'bob dai premium debt final balance');
```
miguelmtzinf, line 384: same

tests/unit/Spoke/Spoke.Repay.EdgeCases.t.sol
```diff
@@ -491,7 +495,7 @@ contract SpokeRepayEdgeCaseTest is SpokeBase {
     );
 
     assertEq(bobDaiDataAfter.suppliedShares, bobDaiDataBefore.suppliedShares);
-    assertApproxEqAbs(bobDaiAfter.baseDebt, daiBorrowAmount, 1, 'bob dai base debt final balance');
+    assertApproxEqAbs(bobDaiAfter.baseDebt, daiBorrowAmount, 2, 'bob dai base debt final balance');
```
miguelmtzinf, line 498: same

tests/unit/Spoke/Spoke.Repay.EdgeCases.t.sol
```diff
@@ -592,7 +596,7 @@ contract SpokeRepayEdgeCaseTest is SpokeBase {
     );
 
     assertEq(bobDaiDataAfter.suppliedShares, bobDaiDataBefore.suppliedShares);
-    assertEq(bobDaiAfter.baseDebt, daiBorrowAmount, 'bob dai base debt final balance');
+    assertApproxEqAbs(bobDaiAfter.baseDebt, daiBorrowAmount, 2, 'bob dai base debt final balance');
```
miguelmtzinf, line 599: here too

tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
miguelmtzinf: this is concerning

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -300,11 +300,11 @@ contract SpokeRepayTest is SpokeBase {
 
     assertEq(bobDaiDataAfter.suppliedShares, bobDaiDataBefore.suppliedShares);
     assertApproxEqAbs(bobDaiAfter.premiumDebt, 0, 1, 'bob dai premium debt final balance');
-    assertEq(bobDaiAfter.baseDebt, daiBorrowAmount, 'bob dai base debt final balance');
+    assertApproxEqAbs(bobDaiAfter.baseDebt, daiBorrowAmount, 1, 'bob dai base debt final balance');
```
miguelmtzinf, line 303: yikes

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: what value to generate? premium debt is not used, hence any value works

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: yes, I have added the invariant check to a few tests, as well as asserted call to IR strat

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: we have the fuzzed version, which is identical

src/interfaces/IBasicInterestRateStrategy.sol
avniculae: [8f8adcd](https://github.com/aave/aave-v4/pull/470/commits/8f8adcd9f9596e5ee2dd4380f51ca341826fba19)

tests/Base.t.sol
avniculae: [8f8adcd](https://github.com/aave/aave-v4/pull/470/commits/8f8adcd9f9596e5ee2dd4380f51ca341826fba19)

tests/Base.t.sol
avniculae: [8f8adcd](https://github.com/aave/aave-v4/pull/470/commits/8f8adcd9f9596e5ee2dd4380f51ca341826fba19)

src/contracts/LiquidityHub.sol
avniculae: the change in premium does happen in refresh, but the IR strat is called only once, and all that we have to make sure if that it's called with the latest values, which we do in this case. don't you think this is correct?

tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
avniculae: investigated it: there is 1 wei delta in premium (which is expected, we know this can go up to 2 wei), and there is 3 wei delta in base debt. we know that bc the borrowAmount is converted into shares, there can be up to 1 wei delta in terms of shares. given drawn index was 2.8 in the scenario I investigated, the 3 wei delta in terms of assets is expected. the theoretical delta here is `hub.convertToDrawnAssets(daiAssetId, 1) + 2`

avniculae: I fuzzed 100k runs with `hub.convertToDrawnAssets(daiAssetId, 1) + 2` as delta and the test passed. I am confident that this discrepancy is not coming from the IR strat changes (although delta was smaller before bc the borrow rate was smaller due to incorrect updates)

I think we should create a ticket to assess the deltas used in our tests, as they are a bit messy atm. wdyt?

tests/unit/AssetInterestRateStrategy.t.sol
miguelmtzinf: random, just from there, so can be easily changed in the future

src/contracts/LiquidityHub.sol
miguelmtzinf: this is fine as long as we assume `refresh` will be always called alongside `restore`. This is the assumption we are working with already, so all good. But again, the premium change actually happens in the `refresh`

tests/unit/Spoke/Spoke.Repay.Scenario.t.sol
miguelmtzinf: not sure I follow. Interested to know the rationale behind that value for the delta.
Anyway, this is something for Certora too

tests/Base.t.sol
```diff
@@ -1782,7 +1782,7 @@ abstract contract Base is Test {
     );
   }
 
-  function checkBorrowRateInvariant(
+  function assertBorrowRateInvariant(
```
miguelmtzinf: but this fn only works if no time has passed no?

tests/Utils.sol
```diff
@@ -57,48 +57,56 @@ library Utils {
...
   }
 
   function addSpoke(
     ILiquidityHub hub,
+    address hubAdmin,
     uint256 assetId,
     address spoke,
     DataTypes.SpokeConfig memory spokeConfig
   ) internal {
+    vm.prank(hubAdmin);
```
DhairyaSethi, line 76: i would rather we fetch the hub admin from the hub rather than pass it in the Utils

tests/Base.t.sol
avniculae: yes exactly. this invariant must be true immediately after each operation

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
```diff
@@ -531,25 +496,37 @@ contract LiquidityHubSupplyTest is LiquidityHubBase {
...
+      abi.encodeCall(
+        IBasicInterestRateStrategy.calculateInterestRate,
+        (daiAssetId, availableLiquidityBefore + supplyAmount, baseDebtBefore, premiumDebtBefore)
+      )
+    );
+
+    vm.prank(alice);
+    IERC20(hub.getAsset(daiAssetId).underlying).approve(address(hub), supplyAmount);
+
+    vm.expectEmit(hub.getAsset(daiAssetId).underlying);
```
DhairyaSethi:
```suggestion
    vm.expectEmit(address(tokenList.dai));
```

tests/Utils.sol
CheyenneAtapour: That sounds better to me than this current approach, but I also don't like coupling these Utils with assumptions about access control... I was pranking outside of using Utils. Wdyt?

tests/unit/AssetInterestRateStrategy.t.sol
CheyenneAtapour: We could just use `Bob` or `Alice` here imo. They are basically the standard users with additional privileges in our test suite. We have addresses for generic `USER1` and 2, but there are comments to remove those.

Also I believe Dhairya is fine with those kinds of assumes when fuzzing because they are narrow (in this case would be just assuming not the liquidity hub)

src/contracts/LiquidityHub.sol
yan-man: strategy as in naming convention?

tests/unit/AssetInterestRateStrategy.t.sol
yan-man: yeah with addresses we can't bound so got to use vm assume

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
yan-man: was this failing before under the 0 input case? is this q still relevant

tests/Base.t.sol
yan-man: maybe can leave a dev comment to clarify

tests/Utils.sol
avniculae: how can I do this? I can fetch the AccessManager by calling authority(), but I'm not sure I can fetch accounts that have a particular role

tests/Base.t.sol
miguelmtzinf: Would it be bad if just create 1 helpers instead to calculate borrow rate based on current state, and then do smt like: `assertEq(hub.getAsset(assetId).baseBorrowRate, calculateBorrowRate(hub, assetId)`

Othwerise, i d suggest renaming to `assertBorrowRateSynced`

avniculae: [a1f585f](https://github.com/aave/aave-v4/pull/470/commits/a1f585f11eea0c186c8ee42d0a4b2efdf378ad18)

tests/Utils.sol
DhairyaSethi: true, my preference would be not have util helpers for admin & gov stuff; keep them in internal methods (where hub admin most likely exists) bc would like this abstracted

tests/unit/LiquidityHub/LiquidityHub.Supply.t.sol
avniculae: [969da67](https://github.com/aave/aave-v4/pull/470/commits/969da67da870868bb2677e7d5a2ed794345812b5)

src/contracts/LiquidityHub.sol
```diff
@@ -177,14 +175,12 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
 
-    asset.updateBorrowRate({assetId: assetId, liquidityAdded: 0, liquidityTaken: amount});
-
     uint256 withdrawnShares = asset.toSuppliedSharesUp(amount); // non zero since we round up
-
-    asset.availableLiquidity -= amount;
     asset.suppliedShares -= withdrawnShares;
-
     spoke.suppliedShares -= withdrawnShares;
+    asset.availableLiquidity -= amount;
```
yan-man, line 181: nit: any reason for moving this? I think was there just to keep asset changes altogether

avniculae: wdym? there is no other asset change here right?

tests/unit/Spoke/Spoke.MultipleHub.IsolationMode.t.sol
avniculae: I haven't modified the initialisation value, the borrow rate is still initialised at 0 rate. I think it's up to us if we want to change this

src/contracts/LiquidityHub.sol
yan-man: meant moving this line down from where it was before, where we updated asset accoutning and then spoke accoutning. now we have asset / spoke / asset so wondering if that was on purpose specifically

PR #472, opened 2025-06-24
Deficit Accounting - Dedicated LH method

### LINKED ISSUE #387: Deficit accounting

yan-man:
- implement deficit accounting by creating a dedicated method in LH for reporting deficit

closes #387

### DIFF
--- a/tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
@@ -77,94 +77,14 @@ contract LiquidityHubReportDeficitTest is LiquidityHubBase {
     hub.reportDeficit(usdxAssetId, baseAmount, premiumAmount);
   }
 
-  function test_reportDeficit() public {
+  function test_reportDeficit_with_premium() public {
     uint256 drawnAmount = 10_000e6;
-    test_reportDeficit_fuzz({
+    test_reportDeficit_fuzz_with_premium({
       drawnAmount: drawnAmount,
       baseAmount: drawnAmount / 2,
       premiumAmount: 0,
       skipTime: 365 days
     });
-  }
-
-  function test_reportDeficit_fuzz(
-    uint256 drawnAmount,
-    uint256 baseAmount,
-    uint256 premiumAmount,
-    uint256 skipTime
-  ) public {
-    drawnAmount = bound(drawnAmount, 1, MAX_SUPPLY_AMOUNT);
-    skipTime = bound(skipTime, 1, MAX_SKIP_TIME);
-
-    ReportDeficitTestParams memory params;
-
-    // create premium debt via spoke1
-    (params.baseDebt, params.premiumDebt) = _drawLiquidityFromSpoke(
-      address(spoke1),
-      usdxAssetId,
-      drawnAmount,
-      skipTime,
-      true
-    );
-
-    baseAmount = bound(baseAmount, 0, params.baseDebt);
-    premiumAmount = bound(premiumAmount, 0, params.premiumDebt);
-    vm.assume(baseAmount + premiumAmount > 0);
-
-    params.deficitBefore = getDeficit(hub, usdxAssetId);
-    params.supplyExchangeRateBefore = hub.convertToSuppliedAssets(usdxAssetId, WadRayMath.RAY);
-    params.availableLiquidityBefore = hub.getAvailableLiquidity(usdxAssetId);
-    params.balanceBefore = IERC20(hub.getAsset(usdxAssetId).underlying).balanceOf(address(spoke1));
-    uint256 baseDrawnSharesBefore = hub.getAsset(usdxAssetId).baseDrawnShares;
-    uint256 totalDeficit = baseAmount + premiumAmount;
-
-    vm.expectEmit(address(hub));
-    emit ILiquidityHub.DeficitReported(
-      usdxAssetId,
-      address(spoke1),
-      hub.convertToDrawnShares(usdxAssetId, baseAmount),
-      totalDeficit
-    );
-
-    // Report the deficit
-    vm.prank(address(spoke1));
-    hub.reportDeficit(usdxAssetId, baseAmount, premiumAmount);
-
-    (params.baseDebtAfter, params.premiumDebtAfter) = hub.getAssetDebt(usdxAssetId);
-
-    params.deficitAfter = getDeficit(hub, usdxAssetId);
-    params.supplyExchangeRateAfter = hub.convertToSuppliedAssets(usdxAssetId, WadRayMath.RAY);
-    params.availableLiquidityAfter = hub.getAvailableLiquidity(usdxAssetId);
-    params.balanceAfter = IERC20(hub.getAsset(usdxAssetId).underlying).balanceOf(address(spoke1));
-    uint256 baseDrawnSharesAfter = hub.getAsset(usdxAssetId).baseDrawnShares;
-
-    // due to rounding of donation, base debt can differ by asset amount of one share
-    // and 1 wei imprecision
-    assertApproxEqAbs(
-      params.baseDebtAfter,
-      params.baseDebt - baseAmount,
-      minimumAssetsPerDrawnShare(hub, usdxAssetId) + 1,
-      'base debt'
-    );
-    assertEq(
-      baseDrawnSharesAfter,
-      baseDrawnSharesBefore - hub.convertToDrawnShares(usdxAssetId, baseAmount),
-      'base drawn shares'
-    );
-    assertEq(params.premiumDebtAfter, params.premiumDebt, 'premium debt');
-    assertEq(params.balanceAfter, params.balanceBefore, 'balance change');
-    assertEq(
-      params.availableLiquidityAfter,
-      params.availableLiquidityBefore,
-      'available liquidity'
-    );
-    assertEq(params.deficitAfter, params.deficitBefore + totalDeficit, 'deficit accounting');
-    assertGe(
-      params.supplyExchangeRateAfter,
-      params.supplyExchangeRateBefore,
-      'supply exchange rate should increase'
-    );
-    assertBorrowRateSynced(hub, usdxAssetId, 'reportDeficit');
   }
 
   function test_reportDeficit_fuzz_with_premium(
@@ -242,6 +162,89 @@ contract LiquidityHubReportDeficitTest is LiquidityHubBase {
       'base drawn shares'
     );
     assertEq(params.premiumDebtAfter, params.premiumDebt - premiumAmount, 'premium debt');
+    assertEq(params.balanceAfter, params.balanceBefore, 'balance change');
+    assertEq(
+      params.availableLiquidityAfter,
+      params.availableLiquidityBefore,
+      'available liquidity'
+    );
+    assertEq(params.deficitAfter, params.deficitBefore + totalDeficit, 'deficit accounting');
+    assertGe(
+      params.supplyExchangeRateAfter,
+      params.supplyExchangeRateBefore,
+      'supply exchange rate should increase'
+    );
+    assertBorrowRateSynced(hub, usdxAssetId, 'reportDeficit');
+  }
+
+  // unexpected behavior case: report deficit without refreshing premium debt beforehand
+  // premium debt is incorrectly left in debt, but still duplicated into deficit, increasing exchange rate
+  // remove test after settling premium is integrated into hub restore
+  function test_reportDeficit_fuzz_invalid_premium_accounting(
+    uint256 drawnAmount,
+    uint256 baseAmount,
+    uint256 premiumAmount,
+    uint256 skipTime
+  ) public {
+    drawnAmount = bound(drawnAmount, 1, MAX_SUPPLY_AMOUNT);
+    skipTime = bound(skipTime, 1, MAX_SKIP_TIME);
+
+    ReportDeficitTestParams memory params;
+
+    // create premium debt via spoke1
+    (params.baseDebt, params.premiumDebt) = _drawLiquidityFromSpoke(
+      address(spoke1),
+      usdxAssetId,
+      drawnAmount,
+      skipTime,
+      true
+    );
+
+    baseAmount = bound(baseAmount, 0, params.baseDebt);
+    premiumAmount = bound(premiumAmount, 0, params.premiumDebt);
+    vm.assume(baseAmount + premiumAmount > 0);
+
+    params.deficitBefore = getDeficit(hub, usdxAssetId);
+    params.supplyExchangeRateBefore = hub.convertToSuppliedAssets(usdxAssetId, WadRayMath.RAY);
+    params.availableLiquidityBefore = hub.getAvailableLiquidity(usdxAssetId);
+    params.balanceBefore = IERC20(hub.getAsset(usdxAssetId).underlying).balanceOf(address(spoke1));
+    uint256 baseDrawnSharesBefore = hub.getAsset(usdxAssetId).baseDrawnShares;
+    uint256 totalDeficit = baseAmount + premiumAmount;
+
+    vm.expectEmit(address(hub));
+    emit ILiquidityHub.DeficitReported(
+      usdxAssetId,
+      address(spoke1),
+      hub.convertToDrawnShares(usdxAssetId, baseAmount),
+      totalDeficit
+    );
+
+    // Report the deficit
+    vm.prank(address(spoke1));
+    hub.reportDeficit(usdxAssetId, baseAmount, premiumAmount);
+
+    (params.baseDebtAfter, params.premiumDebtAfter) = hub.getAssetDebt(usdxAssetId);
+
+    params.deficitAfter = getDeficit(hub, usdxAssetId);
+    params.supplyExchangeRateAfter = hub.convertToSuppliedAssets(usdxAssetId, WadRayMath.RAY);
+    params.availableLiquidityAfter = hub.getAvailableLiquidity(usdxAssetId);
+    params.balanceAfter = IERC20(hub.getAsset(usdxAssetId).underlying).balanceOf(address(spoke1));
+    uint256 baseDrawnSharesAfter = hub.getAsset(usdxAssetId).baseDrawnShares;
+
+    // due to rounding of donation, base debt can differ by asset amount of one share
+    // and 1 wei imprecision
+    assertApproxEqAbs(
+      params.baseDebtAfter,
+      params.baseDebt - baseAmount,
+      minimumAssetsPerDrawnShare(hub, usdxAssetId) + 1,
+      'base debt'
+    );
+    assertEq(
+      baseDrawnSharesAfter,
+      baseDrawnSharesBefore - hub.convertToDrawnShares(usdxAssetId, baseAmount),
+      'base drawn shares'
+    );
+    assertEq(params.premiumDebtAfter, params.premiumDebt, 'premium debt');
     assertEq(params.balanceAfter, params.balanceBefore, 'balance change');
     assertEq(
       params.availableLiquidityAfter,


tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -347,27 +434,107 @@ contract SpokeLiquidationBase is SpokeBase {
...
+      state.debtReserve.assetId,
+      WadRayMath.RAY
+    ) /
+      WadRayMath.RAY +
+      1; // add 1 to divUp
+    // outstanding debt should be moved to deficit
+    assertApproxEqAbs(
+      state.deficit.balanceChange,
+      state.outstandingDebt,
+      assetAmountOfOneShare,
```
yan-man: due to rounding in restore (donation), the remaining debt after a restore can diff from the amount repaid within 1 share amt of assets

src/contracts/Spoke.sol
```diff
@@ -1001,6 +1006,69 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+  ) internal {
+    // validation should already have occurred during liquidation
+    DataTypes.ExecuteReportDeficitLocalVars memory vars;
+
+    ILiquidityHub hub = reserve.hub;
+    vars.assetId = reserve.assetId;
+    (vars.baseDebt, vars.premiumDebt) = _getUserDebt(hub, vars.assetId, userPosition);
+    vars.userPremiumDrawnShares = userPosition.premiumDrawnShares;
+    vars.userPremiumOffset = userPosition.premiumOffset;
+    vars.accruedPremium = vars.premiumDebt - userPosition.realizedPremium;
```
yan-man: we must realize premium debt here first because this op could involve other reserves besides the one involved in liquidation (if user has borrowed others)

yan-man:
converted to draft until rebased to main, and set to blocked until issues sorted with contract size on spoke

src/contracts/LiquidityHub.sol
```diff
@@ -254,6 +255,32 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+
+    uint256 totalDeficitAmount = baseAmount + premiumAmount;
+    uint256 baseDrawnSharesRestored = asset.toDrawnSharesDown(baseAmount);
+    asset.baseDrawnShares -= baseDrawnSharesRestored;
+    spoke.baseDrawnShares -= baseDrawnSharesRestored;
+    asset.deficitAmount += totalDeficitAmount;
+    asset.updateBorrowRate(assetId);
+
+    emit DeficitCreated(assetId, msg.sender, baseDrawnSharesRestored, totalDeficitAmount);
+    emit Restore(assetId, msg.sender, baseDrawnSharesRestored, totalDeficitAmount);
```
yan-man: not sure whether to emit Restore here, figured it is consistent with payFee

src/contracts/Spoke.sol
```diff
@@ -1009,6 +1000,57 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+          premiumDebtRestored
+        );
+        uint256 deficitShares = hub.reportDeficit(assetId, baseDebtRestored, premiumDebtRestored); // settle base debt here by reporting deficit
+
+        reserve.baseDrawnShares -= deficitShares;
+        userPosition.baseDrawnShares -= deficitShares;
+        // newUserRiskPremium is 0 due to no collateral remaining
+        // notify unneeded as each debt reserve's deficit will be individually reported
+        userPosition.premiumDrawnShares = 0;
+        userPosition.premiumOffset = 0;
```
yan-man: did this directly instead of going via _notifyRiskPremiumUpdate bc each reserve's deficit will be reported individually

src/contracts/Spoke.sol
```diff
@@ -1103,87 +1161,58 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-      vars.totalDebtToLiquidate += vars.baseDebtToLiquidate + vars.premiumDebtToLiquidate;
+
+      emit LiquidationCall(
+        vars.collateralUnderlying,
+        vars.debtUnderlying,
+        vars.user,
+        vars.baseDebtToLiquidate + vars.premiumDebtToLiquidate,
+        vars.collateralToLiquidate,
+        liquidator
+      );
```
yan-man, line 1198: is there a desire to emit liquidation fee shares here?

src/contracts/Spoke.sol
```diff
@@ -1009,6 +1000,56 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+        );
+        uint256 deficitShares = hub.reportDeficit(assetId, baseDebtRestored, premiumDebtRestored); // settle base debt here by reporting deficit
+
+        reserve.baseDrawnShares -= deficitShares;
+        userPosition.baseDrawnShares -= deficitShares;
+        // newUserRiskPremium is 0 due to no collateral remaining
+        userPosition.premiumDrawnShares = 0;
+        userPosition.premiumOffset = 0;
+
+        _refreshPremiumDebt(reserve, hub, assetId, reserveId, user, 0, 0, 0, 0);
```
DhairyaSethi: this does nothing and can be removed, _settlePremiumDebt above already resets premiumShares, offset (by design), & realizedPremium (since all premium debt repaid) to 0

src/contracts/LiquidityHub.sol
DhairyaSethi: i don't think so, since this restored amount is not added to available liq 
in payFee it's different bc there's complementary Add & Remove which make change in avail liq 0 

here deficit created event should signal reduction in debt

src/contracts/Spoke.sol
DhairyaSethi: not needed bc settle invocation does this

snapshots/Spoke.Operations.json
```diff
@@ -1,20 +1,20 @@
...
-  "supply + enable usingAsCollateral": "241074",
-  "supply: 0 borrows, collateral disabled": "176592",
-  "supply: 0 borrows, collateral enabled": "179553",
-  "supply: 1 borrow": "179542",
-  "supply: second action, same reserve": "128253",
-  "updateUserRiskPremium: 1 borrow": "161998",
-  "updateUserRiskPremium: 2 borrows": "208279",
+  "borrow": "474973",
+  "liquidationCall: full": "404021",
+  "liquidationCall: partial": "589680",
```
DhairyaSethi: can we add some snapshot scenarios with deficit being created, can take cases from v3 snapshots

src/contracts/Spoke.sol
```diff
@@ -658,13 +644,15 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     uint256 healthFactor,
     uint256 collateralFactor
   ) internal view {
+    DataTypes.ReserveConfig memory collateralConfig = collateralReserve.config;
```
DhairyaSethi: why loading the whole thing in memory

yan-man: seems preferable to 2 sloads on active/paused

DhairyaSethi: this load in memory will cause more sloads (bc it loads the whole config) + the cost of memory expansion and mload 

it's preferable to cache in memory if the flags are read more than once

DhairyaSethi: would leave caching opts for later

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/e0cc49383f16c45999cfd871018733ceece95680

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/5d73b285744fcaa77b1de2ea974b5ce94a833e7d

src/libraries/types/DataTypes.sol
```diff
@@ -29,6 +29,7 @@ library DataTypes {
     uint256 baseDebtIndex;
     uint256 baseBorrowRate;
     uint256 lastUpdateTimestamp;
+    uint256 deficitAmount;
```
miguelmtzinf: i'm fine with just `deficit`

tests/Base.t.sol
```diff
@@ -1244,6 +1242,14 @@ abstract contract Base is Test {
     return hub.convertToSuppliedAssetsUp(assetId, 1);
   }
 
+  /// @dev Helper function to calculate asset amount corresponding to single drawn share
+  function minimumAssetsPerDrawnShare(
+    ILiquidityHub hub,
+    uint256 assetId
+  ) internal view returns (uint256) {
+    return hub.convertToDrawnAssets(assetId, 1);
```
miguelmtzinf: shouldn't be `convertToDrawnAssetsDown` instead?

tests/Base.t.sol
```diff
@@ -2086,6 +2160,31 @@ abstract contract Base is Test {
     );
   }
 
+  /// @dev Adds liquidity to the Hub via a random spoke
+  function _addLiquidity(uint256 assetId, uint256 amount) public {
```
miguelmtzinf: this is already part of the HubBase i think, and there is a similar one in SpokeBase

src/contracts/LiquidityHub.sol
miguelmtzinf: agree

src/interfaces/ILiquidityHub.sol
```diff
@@ -57,6 +57,12 @@ interface ILiquidityHub is IAccessManaged {
     uint256 realizedPremiumAdded,
     uint256 realizedPremiumTaken
   );
+  event DeficitCreated(
+    uint256 indexed assetId,
+    address indexed spoke,
+    uint256 baseSharesCreated,
+    uint256 amountCreated
```
miguelmtzinf:
```suggestion
    uint256 baseSharesRestored,
    uint256 totalDeficit
```
?

src/contracts/Spoke.sol
```diff
@@ -1103,87 +1154,58 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-
-      // TODO: realize bad debt
-      (vars.newUserRiskPremium, , , , ) = _calculateUserAccountData(vars.user);
-
-      if (userDebtPosition.baseDrawnShares == 0) {
-        _positionStatus[vars.user].setBorrowing(vars.debtReserveId, false);
+      if (vars.hasDeficit) {
+        _reportDeficits(vars.user);
+      } else {
+        if (userDebtPosition.baseDrawnShares == 0) {
```
miguelmtzinf: why this only happens if there is no deficit? need to confirm in `_calculateLiquidationParameters`

src/contracts/Spoke.sol
```diff
@@ -1103,87 +1154,58 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-      // collateral accounting
-      userCollateralPosition.suppliedShares -= vars.withdrawnShares;
-
-      // TODO: realize bad debt
-      (vars.newUserRiskPremium, , , , ) = _calculateUserAccountData(vars.user);
-
-      if (userDebtPosition.baseDrawnShares == 0) {
-        _positionStatus[vars.user].setBorrowing(vars.debtReserveId, false);
+      if (vars.hasDeficit) {
+        _reportDeficits(vars.user);
```
miguelmtzinf, line 1181: kind of an example of how messy is this. why not passing `vars` everywhere? lol

src/contracts/Spoke.sol
miguelmtzinf: i think it's fine as is, too many params already

src/contracts/Spoke.sol
yan-man: specifically bc you mentioned not wanting to pass vars struct around

tests/Base.t.sol
yan-man: `convertToDrawnAssetsDown` was not exposed from hub, and I extended this helper based on existing `minimumAssetsPerDrawnShare` helper but can add that here and update given we're adding all the conversion helpers anyway in Cheyenne's PR 

edit: https://github.com/aave/aave-v4/pull/472/commits/99aa94e2f5e85b1e53c549108f67823bdba66ed0

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -117,7 +117,8 @@ contract SpokeRepayTest is SpokeBase {
     assertGt(pos.baseDrawnShares, 0, 'user baseDrawnShares after repay');
     assertGt(hub.convertToDrawnAssets(daiAssetId, pos.baseDrawnShares), 0, 'user baseDrawnAssets');
 
-    Utils.repay(spoke1, _daiReserveId(spoke1), bob, type(uint256).max, bob);
+    vm.prank(bob);
+    spoke1.repay(_daiReserveId(spoke1), UINT256_MAX, bob);
```
yan-man: would like to get some clarity on this approach - using Utils helpers in the test at hand vs calling the actual method

tests/Base.t.sol
yan-man: yes I just moved it from hub base here to be used

src/contracts/Spoke.sol
yan-man: this is only needed here bc if deficit, we are intending anyway that remaining debt on all reserves should be 0

I can add this check into `_calculateLiquidationParameters` but it adds some complexity to the input params

tests/unit/Spoke/Spoke.Repay.t.sol
DhairyaSethi: would continue using Utils here since we don't expect anything before

src/interfaces/ILiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/99aa94e2f5e85b1e53c549108f67823bdba66ed0

src/libraries/types/DataTypes.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/99aa94e2f5e85b1e53c549108f67823bdba66ed0

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/99aa94e2f5e85b1e53c549108f67823bdba66ed0#diff-93ce7444093a2d59b6002ce0c047aafe8b166c6c72757bdb601764970c0d046c

snapshots/Spoke.Operations.json
yan-man: would like to add this as a separate PR: https://github.com/aave/aave-v4/issues/580 to keep this one smaller

src/contracts/Spoke.sol
```diff
@@ -1036,6 +1025,52 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     return vars.premiumIncrease;
   }
 
+  /**
+   * @dev Reports deficits for all borrowing reserves of the user.
+   * Includes the debt reserve being repaid during liquidation.
```
CheyenneAtapour: Perhaps this should be prefixed with a `@dev` tag?

src/contracts/Spoke.sol
```diff
@@ -1036,6 +1025,52 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     return vars.premiumIncrease;
   }
 
+  /**
+   * @dev Reports deficits for all borrowing reserves of the user.
```
CheyenneAtapour, line 1018: I'm fine with the name of the function, and probably is already fine, but is saying something like "Accounts for deficits for all borrowing reserves of the user." clearer?

src/contracts/Spoke.sol
```diff
@@ -1055,27 +1090,26 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     emit UserDynamicConfigRefreshedSingle(user, reserveId);
   }
 
-  /// @return collateralAsset The address of the underlying asset used as collateral, to receive as result of the liquidation.
-  /// @return debtAsset The address of the underlying borrowed asset to be repaid with the liquidation.
-  /// @return totalDebtToLiquidate The total amount of debt to be repaid.
-  /// @return collateralToLiquidate The amount of collateral to liquidate.
```
CheyenneAtapour, line 1050: imo might be nice to give an `@dev` comment detailing at a high level what happens in this function

src/interfaces/ILiquidityHub.sol
```diff
@@ -199,6 +207,20 @@ interface ILiquidityHub is IAccessManaged {
    */
   function payFee(uint256 assetId, uint256 shares) external;
 
+  /**
+   * @notice Reports deficit.
+   * @dev Only callable by spokes.
+   * @param assetId The identifier of the asset.
+   * @param baseAmount The base debt to report as deficit.
+   * @param premiumAmount The premium debt to report as deficit.
+   * @return The amount of debt reported as deficit.
```
CheyenneAtapour: This function returns `baseDrawnSharesRestored`, so is it more accurate to say `The amount of _base debt shares_ reported as deficit.`?

tests/Base.t.sol
```diff
@@ -1841,6 +1867,66 @@ abstract contract Base is Test {
...
+    return
+      totalCollateralBase
+        .percentMulDown(currentAvgCollateralFactor.dewadifyDown())
+        .percentMulDown(99_00)
+        .wadDivDown(desiredHf) - totalDebtBase;
+    // buffer to force debt lower (ie making sure resultant debt creates HF that is gt desired)
+  }
+
+  /// @dev Borrow to be below a certain healthy health factor
+  /// This function validates HF and does not mock price, thus it will cache user RP properly
```
CheyenneAtapour: could add `@dev` prefix here

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
```diff
@@ -0,0 +1,167 @@
...
+    uint256 supplyExchangeRateBefore;
+    uint256 supplyExchangeRateAfter;
+    uint256 availableLiquidityBefore;
+    uint256 availableLiquidityAfter;
+    uint256 balanceBefore;
+    uint256 balanceAfter;
+    uint256 baseDebtAfter;
+    uint256 premiumDebtAfter;
+  }
+  function setUp() public override {
```
CheyenneAtapour, line 22: small nit: Can add whitespace newline above this

src/contracts/Spoke.sol
yan-man: any reason why you prefer "accounts for deficits" instead of "reports deficits"? mildly feel the latter is clearer bc it's literally being reported, and "accounts for" seems more generic

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
yan-man: ah thanks i think it's my IDE settings, I always miss these 

https://github.com/aave/aave-v4/pull/472/commits/f93ac75fae244a628cc3500604afa67e70420eac

tests/Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/f93ac75fae244a628cc3500604afa67e70420eac

src/interfaces/ILiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/f93ac75fae244a628cc3500604afa67e70420eac

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/f93ac75fae244a628cc3500604afa67e70420eac

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/f93ac75fae244a628cc3500604afa67e70420eac

tests/unit/Spoke/Spoke.Repay.t.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/e87be1357176006cca31508fb9878c9f2d4322b8

snapshots/Spoke.Operations.json
DhairyaSethi: it's just one test to be added on `Spoke.operations.gas.t.sol` which is a unit test with deficit accrued, a test we've already written for this change as the happy path basic test with deficit (we can re-use helpers like we do in liquidation with `_borrowToBeBelowHf`) so imo very straightforward and small to add now
gives us an idea of the cost of the implementation as well

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -223,116 +282,172 @@ contract SpokeLiquidationBase is SpokeBase {
...
   ) internal view {
-    _assertUserAccountData(state, spoke, label);
     _assertLiquidationFeeEarned(state, label);
     _assertLiquidationBonusEarned(state, label);
     _assertSupplyExchangeRate(state, label);
-    _assertSetUsingAsCollateral(spoke, alice, state, label);
-    _assertRemainingSpokeCollateral(state, spoke, label);
+    _assertSetUsingAsCollateral(state, label);
+    if (state.hasDeficit) {
+      _assertBadDebt(state, label);
```
DhairyaSethi, line 308: we need to also check other important flags like for the user there is now no borrowed & supplied assets (flags + balances)

src/contracts/Spoke.sol
CheyenneAtapour: Hmm, when I first read the term, I thought maybe it was clearer to say "accounts", but now I see what you're saying. I think it's fine as is then

src/contracts/Spoke.sol
CheyenneAtapour: This commit did not add docs to this function

tests/Base.t.sol
CheyenneAtapour: I think Miguel means to use this one in hubBase, and the spokeBase one if working with spoke tests, if you can

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -361,27 +547,101 @@ contract SpokeLiquidationBase is SpokeBase {
...
   }
 
-  /// @notice Calculate output from LiquidationLogic.calculateAvailableCollateralToLiquidate.
-  /// @param spoke Spoke contract.
-  /// @param state LiquidationTestLocalParams struct containing local params.
-  /// @param debtToCover Desired amount of debt to cover.
-  /// @return actualCollateralToLiquidate Amount of actual collateral to liquidate.
-  /// @return actualDebtToLiquidate Amount of actual debt to liquidate.
-  /// @return liquidationFeeAmount Amount of protocol fee (in asset).
+  /// assertions when bad debt remains and is reported as deficit
```
CheyenneAtapour, line 567: small nit: Can add `@dev` tag just to be consistent

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -361,27 +547,101 @@ contract SpokeLiquidationBase is SpokeBase {
...
+    // outstanding debt should be moved to deficit
+    assertApproxEqAbs(
+      state.deficit.balanceChange,
+      state.outstandingDebt,
+      state.assetAmountOfOneDrawnShare + 1,
+      string.concat('deficit should match restored amount ', label)
+    );
+  }
+
+  /// generic assertions in non bad debt scenarios
```
CheyenneAtapour, line 600: Same, `@dev` tag

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.MultiReserve.BadPremiumDebt.t.sol
```diff
@@ -0,0 +1,515 @@
...
+      supplyAmount: 1.5e18,
+      liquidationFee: 5_00,
+      collateralReserveId: collateralReserveId,
+      skipTime: 365 days,
+      skipTimeToAccruePremium: 365 days * 2,
+      debtReserveIndex: 0
+    });
+  }
+
+  /// fuzz test - bad debt: weth, wbtc, usdy
```
CheyenneAtapour, line 139: `@dev` prefix

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.MultiReserve.BadPremiumDebt.t.sol
```diff
@@ -0,0 +1,515 @@
...
+      supplyAmount: 1.5e18,
+      liquidationFee: 5_00,
+      collateralReserveId: collateralReserveId,
+      skipTime: 365 days,
+      skipTimeToAccruePremium: 365 days * 4,
+      debtReserveIndex: 0
+    });
+  }
+
+  /// fuzz test - bad debt: dai, usdx, usdy
```
CheyenneAtapour, line 220: We can decide if we want / don't want `@dev` before lines describing tests

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.MultiReserve.BadPremiumDebt.t.sol
```diff
@@ -0,0 +1,515 @@
...
+      skipTimeToAccruePremium
+    );
+
+    string memory label = 'test_liquidationCall_fuzz_multi_reserve_badPremiumDebt_scenario3';
+    _checkLiquidation(state, label);
+    _checkDebtReserveDeficits(state, debtReserveIds, alice);
+  }
+
+  /// execute fuzz tests with bad debt across multiple debt reserves
+  /// multiple debt reserves, single collateral reserve
```
CheyenneAtapour: do you want to just say this in one line, `execute fuzz tests with bad debt across multiple debt reserves, single collateral reserve`? or too long?

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
```diff
@@ -0,0 +1,162 @@
...
+
+    vm.expectEmit(address(hub));
+    emit ILiquidityHub.DeficitCreated(
+      usdxAssetId,
+      address(spoke1),
+      hub.convertToDrawnShares(usdxAssetId, baseAmount),
+      totalDeficit
+    );
+
+    // Restore the deficit
```
DhairyaSethi: report

src/contracts/Spoke.sol
```diff
@@ -1025,6 +1014,52 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+          user,
+          premiumDebtRestored
+        );
+        uint256 deficitShares = hub.reportDeficit(assetId, baseDebtRestored, premiumDebtRestored); // settle base debt here by reporting deficit
+
+        reserve.baseDrawnShares -= deficitShares;
+        userPosition.baseDrawnShares -= deficitShares;
+        // newUserRiskPremium is 0 due to no collateral remaining
+        // non-zero deficit means user ends up with zero total debt
+        positionStatus.setBorrowing(reserve.reserveId, false);
```
DhairyaSethi, line 1054: debt & col balance changes are asserted in post liquidation checks but not this flag, pls also include this in _checkLiquidation if hasDeficit branch

DhairyaSethi: to confirm: we had decided not to set collateral flag to false, yeah?

src/contracts/LiquidityHub.sol
```diff
@@ -247,6 +248,31 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+  ) external returns (uint256) {
+    DataTypes.Asset storage asset = _assets[assetId];
+    DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
+
+    asset.accrue(assetId, _spokes[assetId][asset.config.feeReceiver]);
+
+    _validateReportDeficit(asset, spoke, baseAmount, premiumAmount);
+
+    uint256 totalDeficitAmount = baseAmount + premiumAmount;
+    uint256 baseDrawnSharesRestored = asset.toDrawnSharesDown(baseAmount);
```
avniculae: in `restore` we use `previewRestoreByAssets`, should we do the same here?

src/contracts/LiquidityHub.sol
```diff
@@ -247,6 +248,31 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+    asset.baseDrawnShares -= baseDrawnSharesRestored;
+    spoke.baseDrawnShares -= baseDrawnSharesRestored;
+    asset.deficit += totalDeficitAmount;
+    asset.updateBorrowRate(assetId);
```
avniculae, line 271: should we add a dev comment to mention the same thing as in `restore`: `premium debt must be restored in `refreshPremiumDebt` before calling this function` ?

src/contracts/LiquidityHub.sol
```diff
@@ -247,6 +248,31 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+
+    emit DeficitCreated(assetId, msg.sender, baseDrawnSharesRestored, totalDeficitAmount);
```
avniculae: Should we call it `DeficitReported` ?

src/contracts/Spoke.sol
avniculae: If an user has a single collateral left, and two debt reserves left, where one debt reserve has the exact debt amount (in base currency) as the collateral amount, then you'll not be able to liquidate that debt reserve because user will have deficit (bc they have another debt reserve active), you will restore the debt reserve above (but not set borrowing to false), and reportDeficits will fail because both base and premium debt is 0 for this reserve (hence deficit validation on hub will fail). This is ofc very unlikely for a long period of time, but I think the logic should do the following:

```
if (userDebtPosition.baseDrawnShares == 0) {
  _positionStatus[vars.user].setBorrowing(vars.debtReserveId, false);
}
if (vars.hasDeficit) {
  _reportDeficits(vars.user)
} else {
  notify..
}
```

wdyt?

src/contracts/Spoke.sol
DhairyaSethi: debatable, let's tackle in liquidation refactor separately

src/contracts/Spoke.sol
```diff
@@ -1119,87 +1170,65 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
 
-    if (vars.totalLiquidationFeeShares > 0) {
-      collateralReserveHub.payFee(vars.collateralAssetId, vars.totalLiquidationFeeShares);
-    }
-
     // TODO: rm when dupe reserve accounting is rm
     debtReserve.baseDrawnShares -= vars.totalRestoredShares;
     collateralReserve.suppliedShares -= vars.totalWithdrawnShares;
 
-    debtReserveHub.refreshPremiumDebt(
```
avniculae, line 1168: couldn't understand why tests were not failing before, but it looks like `vars.userPremiumDrawnShares` was not populated, hence `vars.totalUserDebtPremiumDrawnSharesDelta` was zero. same for `vars.totalUserDebtPremiumOffsetDelta`. for the collateral side the total values weren't even updated

avniculae: looks like the issue was introduced in this PR https://github.com/aave/aave-v4/pull/473

DhairyaSethi: yes known issue, had pointed before, this (and subsequent coll hub refresh) are unnecessary since the refresh are covered by the notify calls above

@yan-man we can also remove unused local memory vars as well, here or separately

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -91,43 +91,63 @@ library LiquidationLogic {
...
+      vars.collateralAmount =
+        ((vars.maxCollateralToLiquidate * params.collateralAssetUnit) / params.collateralAssetPrice)
           .fromWadDown() +
         1;
-      debtAmountNeeded = params.actualDebtToLiquidate;
+      vars.debtAmountNeeded = params.actualDebtToLiquidate;
+      vars.collateralToLiquidateInBaseCurrency =
+        (vars.collateralAmount * params.collateralAssetPrice).toWad() /
+        params.collateralAssetUnit;
+      vars.debtToLiquidateInBaseCurrency = vars.baseCollateral;
```
avniculae, line 132: I think `baseCollateral` is very misleading here, but logic is correct. nbd now, we can address this when refactoring

src/contracts/Spoke.sol
yan-man: > include this in _checkLiquidation if hasDeficit branch

will add: https://github.com/aave/aave-v4/pull/472/commits/05be275c8486e8092772a4845f51dfaac471e424

> decided not to set collateral flag to false

yes: https://github.com/aave/aave-v4/issues/435

src/libraries/logic/LiquidationLogic.sol
yan-man: used terminology from v3

src/contracts/Spoke.sol
yan-man: mixed it up with _calculateLiquidationParameters, will add

https://github.com/aave/aave-v4/pull/472/commits/05be275c8486e8092772a4845f51dfaac471e424

src/contracts/Spoke.sol
yan-man: > unnecessary since the refresh are covered by the notify calls above

as discussed, this was necessary previously bc notify skipped an `assetId`, it was only in this PR: https://github.com/aave/aave-v4/commit/5b0ae7495194c2f54f39acde37e44feb8f0810d9 which removed the `assetIdToAvoid`

> remove unused local memory vars

I think they have been removed from ExecuteLiquidationLocalVars in DataTypes and here

src/contracts/Spoke.sol
yan-man: actually @avniculae yeah good point, setting borrowing to false would resolve that edge case

https://github.com/aave/aave-v4/pull/472/commits/05be275c8486e8092772a4845f51dfaac471e424

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: see below

src/contracts/Spoke.sol
avniculae: > this (and subsequent coll hub refresh) are unnecessary since the refresh are covered by the notify calls above

I think given that this is covered by notify (and before by `_updatePremiumDebt` that was introduced in PR #473, as mentioned above), the refresh is not only unnecessary, but actually wrong (since you refresh twice).
However, since PR #473, the `total` vars are 0 (as mentioned above), hence all good!

avniculae: > as discussed, this was necessary previously bc notify skipped an assetId, it was only in this PR: https://github.com/aave/aave-v4/commit/5b0ae7495194c2f54f39acde37e44feb8f0810d9 in which removed the assetIdToAvoid

I think they were not needed since PR #473. In fact, I think they would have resulted in wrong behaviour if they were not combined with the additional bug where we did not update the `total` vars

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
yan-man: prefer not to here bc I don't think we've been consistent, see `spoke.repay.t.sol`. Once we settle on the convention is we can go back and update all of these

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.MultiReserve.BadPremiumDebt.t.sol
yan-man: yeah see above, think it's all over the place atm so prefer to leave for later to update all at once

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/05be275c8486e8092772a4845f51dfaac471e424

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/05be275c8486e8092772a4845f51dfaac471e424

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/05be275c8486e8092772a4845f51dfaac471e424

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/05be275c8486e8092772a4845f51dfaac471e424

tests/Base.t.sol
yan-man: actually I can move it back as is: https://github.com/aave/aave-v4/pull/472/commits/bd5ab44935702e17e6a979f35a28243d210b4578

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
```diff
@@ -0,0 +1,162 @@
...
+    emit ILiquidityHub.DeficitReported(
+      usdxAssetId,
+      address(spoke1),
+      hub.convertToDrawnShares(usdxAssetId, baseAmount),
+      totalDeficit
+    );
+
+    // Report the deficit
+    vm.prank(address(spoke1));
+    hub.reportDeficit(usdxAssetId, baseAmount, premiumAmount);
```
DhairyaSethi, line 224: this is is insufficient because premiumAmount has not been restored, it's supposed to be restored with a refresh premium debt before, no?

this should become more intuitive in general for us with #574 bc we'll add deltas on this method directly

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
```diff
@@ -0,0 +1,162 @@
...
+      params.availableLiquidityAfter,
+      params.availableLiquidityBefore,
+      'available liquidity'
+    );
+    assertEq(params.deficitAfter, params.deficitBefore + totalDeficit, 'deficit accounting');
+    assertGe(
+      params.supplyExchangeRateAfter,
+      params.supplyExchangeRateBefore,
+      'supply exchange rate should increase'
+    );
```
DhairyaSethi, line 259: can we also check borrow rate is updated correctly after this method? and also after all liquidation tests, this is done in v3 liq tests as well https://github.com/aave-dao/aave-v3-origin/blob/v3.5.0/tests/protocol/pool/Pool.Deficit.sol#L57
helper: `assertBorrowRateSynced`

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
DhairyaSethi: tests should catch this tho imo, since debt wasn't accounted for correctly here 

since in drawLiquidity withPremium boolean we place all premium debt in realised the settle premium debt needed before should be very straightforward, just realisedPremiumTaken: premiumDebt needed, all other params zero i think

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
CheyenneAtapour: sg

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
yan-man: agree it should perhaps be added but it applies to all the Pool methods, not just liquidations, so doest need to go here specifically. Feel free to add to the clearDeficit PR if you wish

DhairyaSethi: not following, applies to all spoke methods including liquidations, so makes sense to add here now, we can continue adding to the rest of tests as we go along imo

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
yan-man: the premiumAmount isn't restored in a reportDeficit, it must be restored with a refresh before, yes. Added a separate test for this: https://github.com/aave/aave-v4/pull/472/commits/688c19a6f6061ad58fd6b3e7d68160d25148495b

src/contracts/LiquidityHub.sol
```diff
@@ -291,7 +291,7 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     _refresh(
       assetId,
       msg.sender,
-      premiumDrawnShareDelta,
+      premiumDrawnSharesDelta,
```
avniculae, line 294: let's fix this in the `_refresh` function below as well

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
DhairyaSethi: > the premiumAmount isn't restored in a reportDeficit, it must be restored with a refresh before

yes exactly, and it's added to deficit on reportDeficit (parallel to how it's added to availableLiquidity on a restore), so without that settle, excess value is being removed from the deficit counter (bc premiumDebt was never restored, unlike base debt) which is my main point here

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
yan-man: it's not specific to deficit or liquidations (applies to everywhere borrow rate is updated - ie every hub method) and doesn't need to be added here

src/contracts/LiquidityHub.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/de8cd660707f3327902dda6e0aa114316262f0f8

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
DhairyaSethi: why does it not need to be added here? these are the post deficit update w/o time skip assertions, no?
agree it needs to be checked on each method which updates the borrow rate, incl liquidations (especially imp here bc borrowRate is updated through restore, remove, and this) so we start adding this assertion here, and add to remaining tests

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
DhairyaSethi: apart from separate test, it needs to be added on this helper as well

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
avniculae: we have this for some other tests already, eg https://github.com/aave/aave-v4/blob/6b59856b0744c01b94b03b9a89cfb7dfefa3dd93/tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol#L22-L23

I think it's a good idea to add this now as well

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
yan-man: added as a separate test bc although we expect this refresh to occur beforehand think both behaviors should be tested, as a spoke technically can call reportDeficit without settling premium debt part first

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
avniculae: draw test from above also expects the event and calls `assertBorrowRateSynced` after the draw op

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
DhairyaSethi: > as a spoke technically can call reportDeficit without settling premium debt part first

NO it should not, that would be incorrect! this has the same trust assumption as we have with restore - if restoring premium debt, spoke HAS to call settle it beforehand with a refresh

tests/unit/LiquidityHub/LiquidityHub.ReportDeficit.t.sol
yan-man: https://github.com/aave/aave-v4/pull/472/commits/7f908a55b5ccc9057789077a653f252f95dd2ad7

PR #473, opened 2025-06-25
Refactor Premium Debt Accrual Logic

### LINKED ISSUE #471: Refactor logic in Spoke for accruing premium before the user action and refreshing premium data afterwards
These two code snippets are repeated across the code. Let's refactor please.
https://github.com/aave/aave-v4/blob/b3db28c0acdcf2118d7967f6a6ef9c7f875bf537/src/contracts/Spoke.sol#L217-L223
https://github.com/aave/aave-v4/blob/b3db28c0acdcf2118d7967f6a6ef9c7f875bf537/src/contracts/Spoke.sol#L242-L248

CheyenneAtapour:
Moves logic for accruing premium debt and updating across a user's position into 2 dedicated helper functions.
This logic was duplicated across the code in the spoke functions `borrow`, `repay`, and `withdraw`

Closes #471

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -114,7 +114,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     });
     _dynamicConfig[reserveId][dynamicConfigKey] = dynamicConfig;
 
-    emit ReserveAdded(reserveId, assetId);
+    emit ReserveAdded(reserveId, assetId, hub);
     emit ReserveConfigUpdated(reserveId, config);
     emit DynamicReserveConfigUpdated(reserveId, dynamicConfigKey, dynamicConfig);
 
--- a/src/interfaces/ISpoke.sol
+++ b/src/interfaces/ISpoke.sol
@@ -12,7 +12,7 @@ import {DataTypes} from 'src/libraries/types/DataTypes.sol';
  * @notice Basic interface for Spoke
  */
 interface ISpoke is IMulticall, IAccessManaged {
-  event ReserveAdded(uint256 indexed reserveId, uint256 indexed assetId);
+  event ReserveAdded(uint256 indexed reserveId, uint256 indexed assetId, address indexed hub);
   event ReserveConfigUpdated(uint256 indexed reserveId, DataTypes.ReserveConfig config);
   event LiquidityPremiumUpdated(uint256 indexed reserveId, uint256 liquidityPremium);
   event DynamicReserveConfigUpdated(
--- a/tests/unit/Spoke/Spoke.Config.t.sol
+++ b/tests/unit/Spoke/Spoke.Config.t.sol
@@ -403,7 +403,7 @@ contract SpokeConfigTest is SpokeBase {
     address reserveSource = _deployMockPriceFeed(spoke1, 2000e8);
 
     vm.expectEmit(address(spoke1));
-    emit ISpoke.ReserveAdded(reserveId, wethAssetId);
+    emit ISpoke.ReserveAdded(reserveId, wethAssetId, address(hub));
     vm.expectEmit(address(spoke1));
     emit ISpoke.ReserveConfigUpdated(reserveId, newReserveConfig);
     vm.expectEmit(address(spoke1));
--- a/tests/unit/Spoke/Spoke.Multicall.t.sol
+++ b/tests/unit/Spoke/Spoke.Multicall.t.sol
@@ -132,9 +132,9 @@ contract SpokeMulticall is SpokeBase {
     );
 
     vm.expectEmit(address(spoke1));
-    emit ISpoke.ReserveAdded(dai2ReserveId, daiAssetId);
+    emit ISpoke.ReserveAdded(dai2ReserveId, daiAssetId, address(hub));
     vm.expectEmit(address(spoke1));
-    emit ISpoke.ReserveAdded(dai3ReserveId, daiAssetId);
+    emit ISpoke.ReserveAdded(dai3ReserveId, daiAssetId, address(hub));
 
     // Execute the multicall
     vm.prank(SPOKE_ADMIN);


src/contracts/Spoke.sol
```diff
@@ -660,6 +561,64 @@ contract Spoke is ISpoke, Multicall {
...
+    ); // unnecessary but we realize premium debt here
+  }
+
+  function _updatePremiumDebt(
+    ILiquidityHub hub,
+    uint256 assetId,
+    address user,
+    uint256 newUserRiskPremium,
+    DataTypes.Reserve storage reserve,
+    DataTypes.UserPosition storage userPosition
```
yan-man: I think could update arg order to be like other internal methods (_refreshPremiumDebt) with reserve or userPosition first

src/contracts/Spoke.sol
```diff
@@ -660,6 +561,64 @@ contract Spoke is ISpoke, Multicall {
     return (amount - premiumDebt, premiumDebt);
   }
 
+  function _accruePremium(
+    ILiquidityHub hub,
+    uint256 assetId,
+    address user,
+    uint256 premiumDebtRestored,
+    DataTypes.Reserve storage reserve,
+    DataTypes.UserPosition storage userPosition
```
yan-man: same for arg order

src/contracts/Spoke.sol

src/contracts/Spoke.sol

src/contracts/Spoke.sol
```diff
@@ -660,6 +561,64 @@ contract Spoke is ISpoke, Multicall {
...
+
+    _refreshPremiumDebt(
+      reserve,
+      user,
+      assetId,
+      -int256(userPremiumDrawnShares),
+      -int256(userPremiumOffset),
+      accruedPremium,
+      premiumDebtRestored
+    ); // unnecessary but we realize premium debt here
```
DhairyaSethi: would move this comment up everywhere this method is invoked

src/contracts/Spoke.sol
```diff
@@ -660,6 +561,64 @@ contract Spoke is ISpoke, Multicall {
...
+    uint256 userPremiumDrawnShares = userPosition.premiumDrawnShares;
+    uint256 userPremiumOffset = userPosition.premiumOffset;
+    uint256 accruedPremium = hub.convertToDrawnAssets(assetId, userPremiumDrawnShares) -
+      userPremiumOffset; // assets(premiumShares) - offset should never be < 0
+    userPosition.premiumDrawnShares = 0;
+    userPosition.premiumOffset = 0;
+    userPosition.realizedPremium =
+      userPosition.realizedPremium +
+      accruedPremium -
+      premiumDebtRestored;
```
DhairyaSethi, line 656: we are mixing usage of this function here, it accrues but also realises premium debt restored (and this is only needed for restore), i would do something else here to keep separation of concerns 

i'm good with having _accruePremium only accrue and _accrueAndRealisePremium do this instead (which is only called once on repay)

src/contracts/Spoke.sol
```diff
@@ -660,6 +561,64 @@ contract Spoke is ISpoke, Multicall {
...
+    uint256 premiumDebtRestored
+  ) internal {
+    uint256 userPremiumDrawnShares = userPosition.premiumDrawnShares;
+    uint256 userPremiumOffset = userPosition.premiumOffset;
+    uint256 accruedPremium = hub.convertToDrawnAssets(assetId, userPremiumDrawnShares) -
+      userPremiumOffset; // assets(premiumShares) - offset should never be < 0
+    userPosition.premiumDrawnShares = 0;
+    userPosition.premiumOffset = 0;
+    userPosition.realizedPremium =
+      userPosition.realizedPremium +
```
DhairyaSethi, line 654: @CheyenneAtapour you were curious about the gas increase, on first look i think it's bc of:
- the additional sload on each operation for realisedPremium here (unneeded on repay)
- passing user as a param so it lives in the stack, maintaining it and dup before each consumption - earlier was cheaper bc `caller()` was being used for it everywhere 
- not reusing userPremiumDrawnShares, userPremiumOffset vars across these two fns anymore (insteading creating again)

DhairyaSethi: and maintaining the additional premiumDebtRestored on the stack for all operations except repay where it's 0, compiler **cannot** optimise that out since it's inlining the same fn on all methods where it's not unused only on repay and it has to inline the same code

src/contracts/Spoke.sol
yan-man: lol think this one was your comment originally @DhairyaSethi

DhairyaSethi: yep yep i think it's necessary, we can keep track of it in a ticket too, im just saying to move it wherever we call accrue and this applies 

it doesn't make sense inside this method, it says it's unnecessary to accrue before an action but this method is a general accrue abstraction

src/contracts/Spoke.sol
yan-man: imo disagree here, based on the same rationale we did to merge `_refreshPremiumDebt` and `_settlePremiumDebt` vs splitting into 2

DhairyaSethi: i disagree there as well, and this isn't exactly the same. especially bc there is an extraneous premiumDebtRestored arg which is zero for all but repay case here

src/contracts/Spoke.sol
CheyenneAtapour: It also seems strange to me to have it on the invocations of this function though. Everything related to accrual could just be abstracted away, and we can omit the comment

src/contracts/Spoke.sol
CheyenneAtapour: It's also annoying tho imo to duplicate code just for the extra variable. @miguelmtzinf do you have thoughts here?

src/contracts/Spoke.sol
CheyenneAtapour: Alright cool, thanks! So sounds like the only gas saving we can do is split the `_accruePremium` into 2 to not have `premiumDebtRestored` except for the one called during `repay` call

src/contracts/Spoke.sol
DhairyaSethi: it's important bc it's not talking about the accrual impl but rather when we're accruing

src/contracts/Spoke.sol
DhairyaSethi: that also saves an additional unneeded warm sload (100 gas) on repay

src/contracts/Spoke.sol
CheyenneAtapour: kk, done

CheyenneAtapour: @DhairyaSethi what about within the `_executeLiquidationCall` function? It's necessary to accrue there, right? Looks like necessary for debt, but not collateral

src/contracts/Spoke.sol
```diff
@@ -1001,59 +1002,31 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+        userCollateralPosition,
+        collateralReserveHub,
         vars.collateralAssetId,
-        -int256(vars.userPremiumDrawnShares),
-        -int256(vars.userPremiumOffset),
-        vars.userRealizedPremium,
+        vars.user,
         0
-      ); // unnecessary but settle premium debt here for consistency
+      ); // Unecessary but settle premium debt here for consistency
```
yan-man: typo? `Unnecessary`

CheyenneAtapour: Thanks! Fixed

src/contracts/Spoke.sol
```diff
@@ -296,23 +232,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
     );
 
-    uint256 userPremiumDrawnShares = userPosition.premiumDrawnShares;
-    uint256 userPremiumOffset = userPosition.premiumOffset;
-    uint256 accruedPremium = premiumDebt - userPosition.realizedPremium;
-
-    userPosition.premiumDrawnShares = 0;
-    userPosition.premiumOffset = 0;
-    userPosition.realizedPremium = premiumDebt - premiumDebtRestored;
+    _accruePremium(reserve, userPosition, hub, assetId, msg.sender, premiumDebtRestored); // unnecessary but we realize premium debt here
```
```suggestion
    _accruePremium(reserve, userPosition, hub, assetId, msg.sender, premiumDebtRestored); // settle premium debt here
```
necessary and we settle here 
do you see how it's confusing with one name for both? 
now it should be even more compelling to split this in two bc settle is invoked twice, on repay & liquidate

yan-man: why is it necessary now but not before in a withdraw? 

and I don't think it's **exactly** the same, but why do you feel there is a big enough distinction here to have separate methods vs keeping settle/refresh premium debt as separate methods?

src/contracts/Spoke.sol
```diff
@@ -707,6 +607,69 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+  }
+
+  function _calcAndRefreshPremiumDebt(
+    DataTypes.Reserve storage reserve,
+    DataTypes.UserPosition storage userPosition,
+    ILiquidityHub hub,
+    uint256 assetId,
+    address user,
+    uint256 newUserRiskPremium
+  ) internal returns (uint256, uint256) {
```
yan-man, line 676: given how `_notifyRiskPremiumUpdate` also provides bool output that is only used conditionally, feel this should be ok here too

src/contracts/Spoke.sol
CheyenneAtapour: we need @miguelmtzinf to settle the vote here haha. 

Imo, I think the comment is fine to clarify the slight difference. It's also doing the same action, just in one case it's optional and the other it's not

I mean, we could even argue two functions that are almost exactly the same except for one subtraction is also confusing, and we are just adding more code for that

DhairyaSethi: > why is it necessary now but not before in a withdraw?

we've discussed a bunch before, it's necessary bc here we accrue + settle premium debt 
(accrue part of this is unnecessary, like in the case of borrow and withdraw) 

> and I don't think it's exactly the same, but why do you feel there is a big enough distinction here to have separate methods vs keeping settle/refresh premium debt as separate methods?

it **is exactly** the same for settling before in restore and liquidation, they're trying to do the same action. 
on the latter you're confusing two concepts, and there we combine just for the indexer

DhairyaSethi: > we could even argue two functions that are almost exactly the same except for one subtraction is also confusing

not at all, they do completely different things. you have realise the subtlety, this is exactly how bugs are introduced. if you introduce complexity in reading the code by having the overhead of remembering when the last param is 0 the accrue method accrues but when it's non 0 the accrue method settles - you just got confused yourself as the author of the PR
we need different names so it's easier to reason about the code and not have this unnecessary overhead bc you want to merge in one fn (i would somewhat understand if it was aesthetics but this code is nowhere in that ballpark)

yan-man: > it is exactly the same for settling before in restore and liquidation, they're trying to do the same action.
on the latter you're confusing two concepts, and there we combine just for the indexer

lmao no I think you're confusing what I'm asking, I'm not asking about settling. there is no former and latter, just 1 q I am asking, continuing from above where you mentioned about settle/refresh premium. You're saying that we combined for the indexer, but we initially kept them separate as settle premium / refresh premium. But why not continue to keep those separate for the same reasons you prefer to keep this method separate?

src/contracts/Spoke.sol
CheyenneAtapour: @DhairyaSethi won here, went with 2 functions, gg

src/contracts/Spoke.sol
CheyenneAtapour: I have 2 functions now

src/contracts/Spoke.sol
```diff
@@ -170,48 +170,16 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
     _validateWithdraw(reserve, userPosition, amount);
 
-    uint256 userPremiumDrawnShares = userPosition.premiumDrawnShares;
-    uint256 userPremiumOffset = userPosition.premiumOffset;
-    uint256 accruedPremium = hub.convertToDrawnAssets(assetId, userPremiumDrawnShares) -
-      userPremiumOffset; // assets(premiumShares) - offset should never be < 0
-    userPosition.premiumDrawnShares = 0;
-    userPosition.premiumOffset = 0;
-    userPosition.realizedPremium += accruedPremium;
+    _accruePremium(reserve, userPosition, hub, assetId, msg.sender); // unnecessary but we realize premium debt here
```
miguelmtzinf: remove comment pls

src/contracts/Spoke.sol
```diff
@@ -707,6 +608,95 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+      .baseDrawnShares
+      .percentMulUp(newUserRiskPremium);
+    uint256 userPremiumOffset = userPosition.premiumOffset = hub.previewOffset(
+      assetId,
+      userPremiumDrawnShares
+    );
+
+    if (userPosition.baseDrawnShares == 0) {
+      _positionStatus[user].setBorrowing(reserve.reserveId, false);
+    }
```
miguelmtzinf: this does not belong to this fn for sure

src/contracts/Spoke.sol
```diff
@@ -707,6 +608,95 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+
+    _refreshPremiumDebt(
+      reserve,
+      user,
+      assetId,
+      -int256(userPremiumDrawnShares),
+      -int256(userPremiumOffset),
+      accruedPremium,
+      0
+    );
```
```suggestion
    _accrueAndRealizePremium({
        reserve: reserve,
        userPosition: userPosition,
        hub: hub,
        assetId: assetId,
        user: user,
        premiumDebtRestored: 0
    });
```

src/contracts/Spoke.sol
```diff
@@ -707,6 +608,95 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+      user,
+      assetId,
+      -int256(userPremiumDrawnShares),
+      -int256(userPremiumOffset),
+      accruedPremium,
+      premiumDebtRestored
+    );
+  }
+
+  function _calcAndRefreshPremiumDebt(
```
miguelmtzinf: we could write a general function, similar to this one, where we pass the new `userPremiumDrawnShares` and
if `userPremiumDrawnShares == 0`: means we are closing position, so accruing, so everything to 0 and we propagate
if `userPremiumDrawnShares` != 0: means we are are opening position, so update data and propagate

not sure if would be confusing, but we could merge the 3 fns into 1. Another alternative is to have 1 general one and 3 internal wrappers with very good and clear names

miguelmtzinf:
Refactor/rename suggestions for this PR:
- `_accruePremium` -> `_accruePremiumDebt`
- `_accrueAndRealizePremium` -> `_accrueAndSettlePremiumDebt`
- `_calcAndRefreshPremiumDebt` -> `_updatePremiumDebt`

Created new issue for general refactor: https://github.com/aave/aave-v4/issues/508

src/contracts/Spoke.sol
miguelmtzinf: We can use this fn inside `notify`

miguelmtzinf:
Please @CheyenneAtapour , emit an event for `UserRiskPremiumUpdate` whenever we refresh the user rp.

src/contracts/Spoke.sol
DhairyaSethi: i still think it’s necessary here bc this accrue is optional here, we can cleanup all of these comments together with a decision separately. did you have some other thoughts here?

src/contracts/Spoke.sol
CheyenneAtapour: But this one passes `accruedUserPremium` as 0 into `_refreshPremiumDebt`

CheyenneAtapour:
> Please @CheyenneAtapour , emit an event for `UserRiskPremiumUpdate` whenever we refresh the user rp.

For sure, how about we just do this at the end of `_notifyRiskPremiumUpdate`, since this must be called whenever user rp is updated?

src/contracts/Spoke.sol

src/contracts/Spoke.sol

src/contracts/Spoke.sol
```diff
@@ -1167,53 +1105,37 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-        .percentMulUp(vars.newUserRiskPremium);
-      vars.userPremiumOffset = userDebtPosition.premiumOffset = debtReserveHub.previewOffset(
-        vars.debtAssetId,
-        userDebtPosition.premiumDrawnShares
-      );
-      vars.totalUserDebtPremiumDrawnSharesDelta += int256(vars.userPremiumDrawnShares);
-      vars.totalUserDebtPremiumOffsetDelta += int256(vars.userPremiumOffset);
-
-      _refresh(
+      (vars.userPremiumDrawnShares, vars.userPremiumOffset) = _updatePremiumDebt(
```
miguelmtzinf, line 1096: we update premium debt data here and we don't emit `UserRiskPremiumUpdate`. This is the only place where there is no `notify` after `updatePremiumDebt`

src/contracts/Spoke.sol
miguelmtzinf: i thought we decided to keep these accruals at the beginning of each function for consistency, even if not completely necessary
we can discuss again

src/contracts/Spoke.sol
CheyenneAtapour: It happens later on in the function, is that not fine? It's all an atomic operation anyway

miguelmtzinf: hmm good point

PR #477, opened 2025-06-26
Liquidation Protocol Fee - Dedicated LH Method V2

yan-man:
- v2 impl: during payment of LPF, remove only the seized collateral amount without LPF. Create a separate dedicated method on LH to transfer existing supplied shares from current spoke to feeReceiver spoke

Benefits:
- allows direct `remove` to transfer seized collateral to liquidator directly (don't need to transfer collateral to spoke first, and then redistribute to liquidator, which was needed before to separate base seized collateral from LPF)
- doesn't require an `add` type operation to supply fee shares to feeReceiver, can just adjust accounting on LH

### DIFF
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -1,7 +1,7 @@
 {
   "borrow": "517795",
-  "liquidationCall: full": "494460",
-  "liquidationCall: partial": "677266",
+  "liquidationCall: full": "503427",
+  "liquidationCall: partial": "686233",
   "repay: full": "268089",
   "repay: partial": "356100",
   "supply: 0 debt, collateralDisabled": "174957",


src/contracts/LiquidityHub.sol
```diff
@@ -462,7 +485,7 @@ contract LiquidityHub is ILiquidityHub {
     return _assets[assetId].totalSuppliedShares();
   }
 
-  function getSpokeSuppliedAmount(uint256 assetId, address spoke) external view returns (uint256) {
+  function getSpokeSuppliedAmount(uint256 assetId, address spoke) public view returns (uint256) {
```
miguelmtzinf: revert

src/contracts/Spoke.sol
```diff
@@ -583,6 +588,12 @@ contract Spoke is ISpoke, Multicall {
       config.liquidationProtocolFee <= PercentageMathExtended.PERCENTAGE_FACTOR,
       InvalidLiquidationProtocolFee()
     );
+    if (config.liquidationProtocolFee > 0) {
+      require(
+        config.hub.getAssetConfig(assetId).feeReceiver != address(0),
```
```suggestion
      require(
        config.liquidationProtocolFee == 0 || config.hub.getAssetConfig(assetId).feeReceiver != address(0),
```

src/contracts/LiquidityHub.sol
yan-man: just had it for testing, yes will revert

src/contracts/LiquidityHub.sol
```diff
@@ -303,23 +320,31 @@ contract LiquidityHub is ILiquidityHub {
...
-    emit SupplyToFeeReceiver(assetId, msg.sender, suppliedShares, amount);
+    // console.log('supplied %e %e', getSpokeSuppliedAmount(assetId, msg.sender), feeShares);
 
-    return suppliedShares;
+    spoke.suppliedShares -= feeShares;
+    _spokes[assetId][feeReceiver].suppliedShares += feeShares; // if feeReceiver is not defined, these
+
+    // console.log('supplied %e', getSpokeSuppliedAmount(assetId, msg.sender));
+
+    emit AccrueFees(assetId, feeShares);
```
DhairyaSethi, line 337: we should emit the sender & likely receiver (current) spokes as well 
smh gas

src/contracts/Spoke.sol
```diff
@@ -1096,12 +1107,56 @@ contract Spoke is ISpoke, Multicall {
       // debt accounting
       userDebtPosition.baseDrawnShares -= vars.restoredShares;
       vars.totalRestoredShares += vars.restoredShares;
+      vars.withdrawnShares = 0;
 
-      // liquidate collateral
-      vars.withdrawnShares = collateralReserveHub.remove(
+      vars.liquidationProtocolFeeShares = collateralReserveHub.convertToSuppliedShares(
```
```suggestion
      vars.liquidationProtocolFeeShares = collateralReserveHub.convertToSuppliedSharesUp(
```

src/contracts/LiquidityHub.sol
yan-man: miguel had a comment on re-using event from accrual: https://github.com/aave/aave-v4/pull/467/files#r2169027886

in that case cannot emit sender as there won't be one in the pure accrual case

DhairyaSethi: hmm this accrual is different from a pure accrual tho

DhairyaSethi: there is a balance change for the sender spoke, we need to include this information in some event (rn it's not in any) 
idc about (fee) receiver since we signal an 'accrual'

src/contracts/Spoke.sol
yan-man: done anyway. but moving changes back to original PR and closing this one

src/contracts/Spoke.sol
yan-man: no longer relevant

src/contracts/Spoke.sol
```diff
@@ -1097,11 +1108,22 @@ contract Spoke is ISpoke, Multicall {
...
+      vars.withdrawnShares = collateralReserveHub.convertToSuppliedSharesUp(
+        vars.collateralAssetId,
+        vars.liquidationFeeAmount + vars.collateralToLiquidate
+      );
+      vars.withdrawnLiquidatorShares = collateralReserveHub.convertToSuppliedSharesUp(
         vars.collateralAssetId,
-        vars.collateralToLiquidate + vars.liquidationProtocolFeeAmount,
-        address(this) // must be sent to spoke first before distributing to treasury/liquidator
+        vars.collateralToLiquidate
+      );
```
DhairyaSethi, line 1119: don't need to do this, remove call gives you this amt and can calc liq fee shares later

yan-man: this is leftover code. Please refer to the other comment, don't leave additional comments on this pr

DhairyaSethi: which one?

src/contracts/LiquidityHub.sol
yan-man: continue convo in that thread, not here

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/477#discussion_r2172813949

https://github.com/aave/aave-v4/pull/467

src/contracts/LiquidityHub.sol
DhairyaSethi: but the code is here 😃

yan-man: im going to merge this into that branch and continue there

PR #478, opened 2025-06-26
feat: Aave Oracle

### LINKED ISSUE #7: Simple Chainlink-based Oracle
We need a simple oracle contract able to provide prices from Chainlink price feeds.
We can re-use V3 implementation of the AaveOracle, but removing whatever is secondary/complementary (fallback oracle for example).
We also need a mock oracle for tests

### LINKED ISSUE #512: bug: refresh debt referring to wrong hub
https://github.com/aave/aave-v4/blob/5845bbb0f1d85d9e8b2be534731f608601e2258a/src/contracts/Spoke.sol#L1219

should be `debtReserveHub`

avniculae:
Closes #7 
- introduces a minimal AaveOracle inspired from v3, with some modifications: 
  - it now talks to the v3 aggregator interface from chainlink, which is not supporting `latestAnswer()` anymore; 
  - it drops support for the fallback oracle, which was never used in v3
  - does not contain base currency reference anymore; this can be achieved by registering a unit price feed. any other metadata can be included in the description string
  - it makes sure that registered price feeds use the same number of decimals as the oracle
  - uses oz's AccessManaged for permissioning

Closes #512

### DIFF
--- a/src/misc/Multicall.sol
+++ b/src/misc/Multicall.sol
@@ -23,5 +23,6 @@ abstract contract Multicall is IMulticall {
 
       results[i] = res;
     }
+    return results;
   }
 }
--- a/tests/unit/Spoke/Spoke.Multicall.t.sol
+++ b/tests/unit/Spoke/Spoke.Multicall.t.sol
@@ -122,8 +122,14 @@ contract SpokeMulticall is SpokeBase {
 
     // Set up the multicall
     bytes[] memory calls = new bytes[](2);
-    calls[0] = abi.encodeCall(ISpoke.addReserve, (daiAssetId, address(hub), _deployMockPriceFeed(spoke1, 1e8), dai2Config, dai2DynConfig));
-    calls[1] = abi.encodeCall(ISpoke.addReserve, (daiAssetId, address(hub), _deployMockPriceFeed(spoke1, 1e8), dai3Config, dai3DynConfig));
+    calls[0] = abi.encodeCall(
+      ISpoke.addReserve,
+      (daiAssetId, address(hub), _deployMockPriceFeed(spoke1, 1e8), dai2Config, dai2DynConfig)
+    );
+    calls[1] = abi.encodeCall(
+      ISpoke.addReserve,
+      (daiAssetId, address(hub), _deployMockPriceFeed(spoke1, 1e8), dai3Config, dai3DynConfig)
+    );
 
     vm.expectEmit(address(spoke1));
     emit ISpoke.ReserveAdded(dai2ReserveId, daiAssetId);
@@ -174,5 +180,24 @@ contract SpokeMulticall is SpokeBase {
     // Check the reserve configs
     assertEq(spoke1.getReserve(daiReserveId), newDai);
     assertEq(spoke1.getReserve(usdxReserveId), newUsdx);
+  }
+
+  function test_multicall_getters() public {
+    bytes[] memory calls = new bytes[](5);
+    calls[0] = abi.encodeCall(ISpoke.supply, (_daiReserveId(spoke1), 120e18));
+    calls[1] = abi.encodeCall(ISpoke.setUsingAsCollateral, (_daiReserveId(spoke1), true));
+    calls[2] = abi.encodeCall(ISpoke.borrow, (_daiReserveId(spoke1), 80e18, alice));
+    calls[3] = abi.encodeCall(ISpoke.getUserRiskPremium, (alice));
+    calls[4] = abi.encodeCall(ISpoke.getUserDebt, (_daiReserveId(spoke1), alice));
+
+    vm.prank(alice);
+    bytes[] memory ret = spoke1.multicall(calls);
+
+    assertEq(ret.length, calls.length);
+    assertEq(ret[0].length, 0);
+    assertEq(ret[1].length, 0);
+    assertEq(ret[2].length, 0);
+    assertEq(ret[3], abi.encode(_calculateExpectedUserRP(alice, spoke1)));
+    assertEq(ret[4], abi.encode(80e18, 0));
   }
 }


src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  function _getSourcePrice(uint256 reserveId) internal view returns (uint256) {
+    AggregatorV3Interface source = reserveSource[reserveId];
+    if (address(source) == address(0)) {
+      revert InvalidSource(reserveId);
+    }
+
+    (, int256 price, , , ) = source.latestRoundData();
+    if (price < 0) {
```
avniculae: v3 fallbacks to the fallback oracle if price is <= 0, but to me 0 is a valid price, wdyt?

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.0;
+
+import {IPriceOracle} from './IPriceOracle.sol';
+
+interface IAaveOracle is IPriceOracle {
```
avniculae, line 11: TODO: natspec

src/contracts/AaveOracle.sol
DhairyaSethi: that's not a valid price unfortunately, most oracles return 0 when stale as well

avniculae: imo staleness should be checked against the timestamp returned, not the price. This approach would also allow to register zero price feeds, where the price feed always return a constant 0 price (if we would like to do this for any particular reason)

src/misc/UnitPriceFeed.sol
avniculae, line 1: FYI The aave oracle does not have an explicit base currency anymore (as in v3). If there will ever be a reserve with the base currency as the asset, we can register this UnitPriceFeed, which will just return 1 (with decimals precision)

src/contracts/Spoke.sol
```diff
@@ -54,6 +54,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     require(oracleAddress != address(0), InvalidOracleAddress());
 
     oracle = IPriceOracle(oracleAddress);
+    emit OracleUpdated(oracleAddress);
```
avniculae: shouldn't we have a function in the spoke to update the oracle? I think we should

src/contracts/AaveOracle.sol
avniculae: addressed, please see latest code

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,83 @@
...
+  /// @inheritdoc IPriceOracle
+  uint8 public immutable override DECIMALS;
+  /// @inheritdoc IPriceOracle
+  string public override DESCRIPTION;
+
+  mapping(uint256 reserveId => AggregatorV3Interface source) internal _reserveSource;
+
+  /**
+   * @dev Constructor.
+   * @dev The authority should implement the AccessManaged interface to control access.
```
yan-man:
```suggestion
   * @dev The authority must implement the AccessManaged interface to control access.
```
?

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,83 @@
...
+  function getReservePrice(uint256 reserveId) external view override returns (uint256) {
+    return _getSourcePrice(reserveId);
+  }
+
+  /// @inheritdoc IAaveOracle
+  function getReservesPrices(
+    uint256[] calldata reserveIds
+  ) external view override returns (uint256[] memory) {
+    uint256[] memory prices = new uint256[](reserveIds.length);
+    for (uint256 i = 0; i < reserveIds.length; i++) {
```
yan-man:
```suggestion
uint256 reserveIdsLength = reserveIds.length;
    for (uint256 i; i < reserveIdsLength; ++i) {
```

src/dependencies/chainlink/AggregatorV3Interface.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+// Imported form https://github.com/smartcontractkit/chainlink/blob/e8490e910274279a2c394ba37eb328943322ac6b/contracts/src/v0.8/shared/interfaces/AggregatorV3Interface.sol
```
yan-man:
```suggestion
// Imported from https://github.com/smartcontractkit/chainlink/blob/e8490e910274279a2c394ba37eb328943322ac6b/contracts/src/v0.8/shared/interfaces/AggregatorV3Interface.sol
```

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,63 @@
...
+   * @return prices The prices of the reserves
+   */
+  function getReservesPrices(
+    uint256[] calldata reserveIds
+  ) external view returns (uint256[] memory);
+
+  /**
+   * @notice Returns the price feed source of a reserve
+   * @param reserveId The id of the reserve
+   * @return source The price feed source of the reserve
```
yan-man: periods on natspec

src/interfaces/ISpoke.sol
```diff
@@ -45,7 +45,7 @@ interface ISpoke is IMulticall, IAccessManaged {
     uint256 realizedPremiumAdded,
     uint256 realizedPremiumTaken
   );
-  event OracleUpdated(uint256 indexed reserveId, address indexed oracle);
+  event OracleUpdated(address indexed oracle);
```
yan-man, line 49: can you add a test for emitting this event

src/misc/UnitPriceFeed.sol
```diff
@@ -0,0 +1,68 @@
...
+   * @notice Thrown when the operation is not supported.
+   */
+  error OperationNotSupported();
+
+  /**
+   * @dev Constructor.
+   * @param decimals_ The number of decimals used to represent the unit price.
+   * @param description_ The description of the unit price feed.
+   */
+  constructor(uint8 decimals_, string memory description_) {
```
yan-man, line 30: i would like to maintain consistency on constructor args, and do this everywhere (`variable_`) if everyone is aligned with @miguelmtzinf (I personally would also like to do this pattern in normal functions too ideally...) 

https://github.com/aave/aave-v4/pull/490#discussion_r2176617432

cc @DhairyaSethi @CheyenneAtapour

tests/Base.t.sol
```diff
@@ -800,10 +794,28 @@ abstract contract Base is Test {
...
+    _mockReservePrice(spoke2, spokeInfo[spoke2].dai.reserveId, 1e8);
+    _mockReservePrice(spoke2, spokeInfo[spoke2].usdx.reserveId, 1e8);
+    _mockReservePrice(spoke2, spokeInfo[spoke2].usdy.reserveId, 1e8);
+
+    _mockReservePrice(spoke3, spokeInfo[spoke3].dai.reserveId, 1e8);
+    _mockReservePrice(spoke3, spokeInfo[spoke3].usdx.reserveId, 1e8);
+    _mockReservePrice(spoke3, spokeInfo[spoke3].weth.reserveId, 2000e8);
+    _mockReservePrice(spoke3, spokeInfo[spoke3].wbtc.reserveId, 50_000e8);
+
+    _mockReservePrice(spoke2, spokeInfo[spoke2].dai2.reserveId, 1e8);
```
yan-man: can you move this up w other spoke2 ones

tests/unit/AaveOracle.t.sol
```diff
@@ -0,0 +1,186 @@
...
+  function test_setReserveSource_revertsWith_InvalidSource() public {
+    _mockSourceDecimals(address(0), _decimals);
+
+    vm.expectRevert(abi.encodeWithSelector(IAaveOracle.InvalidSource.selector, reserveId1));
+
+    vm.prank(ORACLE_ADMIN);
+    oracle.setReserveSource(reserveId1, address(0));
+  }
+
+  function test_setReserveSource_revertsWith_InvalidPrice() public {
```
yan-man, line 78: worth it to fuzz on this <= 0?

tests/Base.t.sol
```diff
@@ -1866,4 +1869,18 @@ abstract contract Base is Test {
       abi.encode(interestRateBps.bpsToRay())
     );
   }
+
+  function _mockReservePrice(ISpoke spoke, uint256 reserveId, uint256 price) internal {
```
yan-man, line 1937: to clarify, every time we do this func call, it'd deploy a new price feed rather than updating just the previous one's price right? do you think worth it to add some way like before where we can just update the price on the old one without needing to deploy a whole new oracle

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,83 @@
...
+
+import {AggregatorV3Interface} from 'src/dependencies/chainlink/AggregatorV3Interface.sol';
+import {AccessManaged} from 'src/dependencies/openzeppelin/AccessManaged.sol';
+import {IAaveOracle, IPriceOracle} from 'src/interfaces/IAaveOracle.sol';
+
+/**
+ * @title AaveOracle contract
+ * @author Aave Labs
+ * @notice Oracle contract for the Aave protocol
+ * @dev Oracles are spoke-specific: one oracle CAN'T be used across different spoke
```
CheyenneAtapour: Technically if the spokes had the same reserves in the same order (same reserveIds), then ofc you could reuse the oracle. It's nbd, but I'm fine with omitting this part about not using across different spokes and just writing it as 
`Oracles are spoke-specific due to the usage of reserve id as index of the _reserveSource`

If you want to keep as is, please change it to say `...used across different spokes` plural

src/contracts/AaveOracle.sol
CheyenneAtapour: ah man, we have `must` on LiquidityHub and `should` on Spoke. But yeah, I think `must` is best for all

src/contracts/AaveOracle.sol
CheyenneAtapour: If caching there, can also cache before and use during instantiating `prices` array

src/misc/UnitPriceFeed.sol
CheyenneAtapour: Fine with me, unsure I like it on normal functions tho. Makes sense in constructor because usually variables are instantiated, which have a naming collision with the params

src/contracts/AaveOracle.sol
avniculae: i think @miguelmtzinf changed it to `should` if I am not mistaken, I think should is correct bc authority can be something else as well as long as it implements the necessary interface

tests/Base.t.sol
avniculae: we can't because of ongoing prank (we cannot override prank)

tests/Base.t.sol
avniculae: I think current version is fine given it's tests, but I have no strong opinion

tests/unit/AaveOracle.t.sol
avniculae: I think fuzz is not worth it, but I will add 0 and -100 to test more values, wdyt?

src/contracts/AaveOracle.sol
avniculae: @CheyenneAtapour wdym?

DhairyaSethi: array is in calldata, very little to no benefit from caching & can be avoided here

DhairyaSethi: but preincrement yes (++i) bc it saves a dup

src/contracts/AaveOracle.sol
avniculae: [87faf72](https://github.com/aave/aave-v4/pull/478/commits/87faf72cf676e72a4becbdb7740f2aecf9425f34)

tests/unit/AaveOracle.t.sol
avniculae: [87faf72](https://github.com/aave/aave-v4/pull/478/commits/87faf72cf676e72a4becbdb7740f2aecf9425f34)

src/misc/UnitPriceFeed.sol
avniculae: Made the change in the spoke as well [87faf72](https://github.com/aave/aave-v4/pull/478/commits/87faf72cf676e72a4becbdb7740f2aecf9425f34)

src/interfaces/ISpoke.sol
avniculae: good catch, [87faf72](https://github.com/aave/aave-v4/pull/478/commits/87faf72cf676e72a4becbdb7740f2aecf9425f34)

src/interfaces/IAaveOracle.sol
avniculae: [87faf72](https://github.com/aave/aave-v4/pull/478/commits/87faf72cf676e72a4becbdb7740f2aecf9425f34)

src/dependencies/chainlink/AggregatorV3Interface.sol
avniculae: [87faf72](https://github.com/aave/aave-v4/pull/478/commits/87faf72cf676e72a4becbdb7740f2aecf9425f34)

src/contracts/AaveOracle.sol
avniculae: [87faf72](https://github.com/aave/aave-v4/pull/478/commits/87faf72cf676e72a4becbdb7740f2aecf9425f34)

src/contracts/Spoke.sol
avniculae: [87faf72](https://github.com/aave/aave-v4/pull/478/commits/87faf72cf676e72a4becbdb7740f2aecf9425f34)

tests/Base.t.sol
yan-man: I meant just to L812, does that also have ongoing prank issues

tests/unit/AaveOracle.t.sol
yan-man: yeah just one with zero is fine too

tests/unit/UnitPriceFeed.t.sol
```diff
@@ -0,0 +1,44 @@
...
+
+  function test_getRoundData_revertsWith_OperationNotSupported() public {
+    vm.expectRevert(UnitPriceFeed.OperationNotSupported.selector);
+    unitPriceFeed.getRoundData(0);
+  }
+
+  function test_fuzz_latestRoundData(uint80 blockTimestamp) public {
+    vm.warp(blockTimestamp);
+    (uint80 roundId, int256 answer, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound) = unitPriceFeed.latestRoundData();
+    assertEq(roundId, block.timestamp);
```
yan-man: i think we were wanting to use `vm.getBlockTimestamp` in tests due to issues in skipping: https://github.com/foundry-rs/foundry/issues/6180

tests/unit/UnitPriceFeed.t.sol
```diff
@@ -0,0 +1,44 @@
...
+    assertEq(unitPriceFeed.version(), 1);
+  }
+
+  function test_getRoundData_revertsWith_OperationNotSupported() public {
+    vm.expectRevert(UnitPriceFeed.OperationNotSupported.selector);
+    unitPriceFeed.getRoundData(0);
+  }
+
+  function test_fuzz_latestRoundData(uint80 blockTimestamp) public {
+    vm.warp(blockTimestamp);
```
yan-man, line 36: and `skip` instead, but forget if this was just for consistency or any issues with warp

tests/unit/AaveOracle.t.sol
```diff
@@ -0,0 +1,194 @@
...
+  function _mockSourceDecimals(address source, uint8 decimals) internal {
+    vm.mockCall(source, abi.encodeCall(AggregatorV3Interface.decimals, ()), abi.encode(decimals));
+  }
+
+  function _mockSourceLatestRoundData(address source, int256 price) internal {
+    vm.mockCall(
+      source,
+      abi.encodeCall(AggregatorV3Interface.latestRoundData, ()),
+      abi.encode(
+        uint80(block.timestamp),
```
yan-man: same with vm.getBlockTimestamp

src/contracts/AaveOracle.sol
yan-man: it can be something else but still must implement the AccessManaged interface then right?

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,63 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.0;
+
+import {IPriceOracle} from './IPriceOracle.sol';
```
yan-man: nit: consistency on absolute imports

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,63 @@
...
+interface IAaveOracle is IPriceOracle {
+  /**
+   * @notice Emitted when the Aave Oracle is created.
+   * @param decimals The number of decimals used to return prices.
+   * @param description The description of the oracle.
+   */
+  event AaveOracleCreated(uint256 indexed decimals, string indexed description);
+  /**
+   * @notice Emitted when the price feed source of a reserve is updated.
+   * @param reserveId The id of the reserve.
```
yan-man: nit: consistency on "unique identifier" or however else we refer to ids in other places

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,63 @@
...
+   * @param source The price feed source of the reserve.
+   */
+  event ReserveSourceUpdated(uint256 indexed reserveId, address indexed source);
+
+  /**
+   * @notice Thrown when the price feed source uses a different number of decimals than the oracle.
+   * @param reserveId The id of the reserve.
```
yan-man: same

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,63 @@
...
+   */
+  error InvalidSourceDecimals(uint256 reserveId);
+  /**
+   * @notice Thrown when the price feed source is invalid (zero address).
+   * @param reserveId The id of the reserve.
```
yan-man: same, also below

src/contracts/Spoke.sol
```diff
@@ -32,7 +32,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
 
   uint256 public constant HEALTH_FACTOR_LIQUIDATION_THRESHOLD = WadRayMathExtended.WAD;
   uint256 public constant MAX_LIQUIDITY_PREMIUM = 1000_00; // 1000.00%
-  IPriceOracle public immutable oracle;
+  IPriceOracle public oracle;
```
yan-man: nit: given it's not immutable now, can move this down with the other normie public vars

src/contracts/AaveOracle.sol
```diff
@@ -8,9 +8,9 @@ import {IAaveOracle, IPriceOracle} from 'src/interfaces/IAaveOracle.sol';
 /**
  * @title AaveOracle contract
  * @author Aave Labs
- * @notice Oracle contract for the Aave protocol
- * @dev Oracles are spoke-specific: one oracle CAN'T be used across different spoke
- *   due to the usage of reserve id as index of the _reserveSource
+ * @notice Oracle contract for the Aave protocol.
+ * @dev Oracles are spoke-specific: one oracle CAN'T be used across different spokes
+ *   due to the usage of reserve id as index of the _reserveSource (unless spokes use the same reserve ids).
```
CheyenneAtapour: I don't like the parentheses, you can just revert the change, but keep the plural `spokes`

src/contracts/AaveOracle.sol
CheyenneAtapour: yeah it still is a must

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -9,7 +9,7 @@ import {IAssetInterestRateStrategy, IBasicInterestRateStrategy} from 'src/interf
  * @author Aave Labs
  * @notice Asset interest rate strategy used by the Aave protocol
  * @dev Strategies are hub-specific: one strategy CAN'T be used across different Aave hubs
```
CheyenneAtapour: My preferred suggestion is 

`@dev Strategies are hub specific, due to the usage of asset id as index of the _interestRateData`

src/contracts/AaveOracle.sol
avniculae: Ah, correct, it says interface. Will change to must

tests/unit/UnitPriceFeed.t.sol
avniculae: skip adds delta, warp sets absolute value. in this case I want to do warp, not skip

tests/Base.t.sol
avniculae: ah I see what you mean. this follows the same order of adding the reserves above, but will change both

src/interfaces/IAaveOracle.sol
avniculae: looks like we use unique only for return params. for inputs we use "identifier"

src/contracts/AssetInterestRateStrategy.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

src/contracts/AaveOracle.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

src/contracts/Spoke.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

src/interfaces/IAaveOracle.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

src/interfaces/IAaveOracle.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

src/interfaces/IAaveOracle.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

src/interfaces/IAaveOracle.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

tests/unit/AaveOracle.t.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

tests/unit/UnitPriceFeed.t.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

tests/Base.t.sol
avniculae: [ac71d91](https://github.com/aave/aave-v4/pull/478/commits/ac71d91cf95c88024bb08ece2ed064e3f51b50bd)

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.0;
+
+import {AggregatorV3Interface} from 'src/dependencies/chainlink/AggregatorV3Interface.sol';
+import {AccessManaged} from 'src/dependencies/openzeppelin/AccessManaged.sol';
+import {IAaveOracle, IPriceOracle} from 'src/interfaces/IAaveOracle.sol';
+
+/**
+ * @title AaveOracle contract
```
miguelmtzinf:
```suggestion
 * @title AaveOracle
```

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * @dev Constructor.
+   * @dev The authority must implement the AccessManaged interface to control access.
+   * @param authority The address of the authority contract which manages permissions.
+   * @param decimals The number of decimals for the oracle.
+   * @param description The description of the oracle.
+   */
+  constructor(
+    address authority,
```
miguelmtzinf: wondering if updates should go through the spoke, since this is spoke specific

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,82 @@
...
+    emit ReserveSourceUpdated(reserveId, source);
+  }
+
+  /// @inheritdoc IPriceOracle
+  function getReservePrice(uint256 reserveId) external view override returns (uint256) {
+    return _getSourcePrice(reserveId);
+  }
+
+  /// @inheritdoc IAaveOracle
+  function getReservesPrices(
```
miguelmtzinf, line 51: why do we need this? or just for flexibility

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,82 @@
...
+  mapping(uint256 reserveId => AggregatorV3Interface source) internal _reserveSource;
+
+  /**
+   * @dev Constructor.
+   * @dev The authority must implement the AccessManaged interface to control access.
+   * @param authority The address of the authority contract which manages permissions.
+   * @param decimals The number of decimals for the oracle.
+   * @param description The description of the oracle.
+   */
+  constructor(
```
miguelmtzinf: `variable_`

src/misc/UnitPriceFeed.sol
miguelmtzinf: not really following. How can i know which asset is used as base currency? not possible right

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,82 @@
...
+  /// @inheritdoc IAaveOracle
+  function getReserveSource(uint256 reserveId) external view override returns (address) {
+    return address(_reserveSource[reserveId]);
+  }
+
+  function _getSourcePrice(uint256 reserveId) internal view returns (uint256) {
+    AggregatorV3Interface source = _reserveSource[reserveId];
+    if (address(source) == address(0)) {
+      revert InvalidSource(reserveId);
+    }
```
DhairyaSethi: let's use require style here instead 
same below

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: AGPL-3.0
```
DhairyaSethi: MIT it for now

src/misc/UnitPriceFeed.sol
avniculae: unless I am mistaken, I don't think the address of the base currency was read/needed in other contracts, but rather it represented just some metadata for the outside world. This can be achieved through `description` now. In any case, bc the v4 oracle operates on reserves and not on assets, I think base currency does not make sense anymore (but rather base reserve).

I think we can replace description with base reserve if we want (or maybe keep both), but in any case I don't think we should have special logic in `getReservePrice` which returns the unit price if reserve is the base reserve; I think registering a UnitPriceFeed is much cleaner. This is the special logic that I referring to, from v3:
https://github.com/aave-dao/aave-v3-origin/blob/496c6eaf9056a25729a28c9e98b7757f975dc54a/src/contracts/misc/AaveOracle.sol#L101-L105

src/contracts/AaveOracle.sol
avniculae: just for flexibility

src/contracts/AaveOracle.sol
avniculae: [f6da2ab](https://github.com/aave/aave-v4/pull/478/commits/f6da2ab796fc45faa05cce6f90ae36b02a204621)

src/contracts/AaveOracle.sol
avniculae: [f6da2ab](https://github.com/aave/aave-v4/pull/478/commits/f6da2ab796fc45faa05cce6f90ae36b02a204621)

src/contracts/AaveOracle.sol
avniculae: [f6da2ab](https://github.com/aave/aave-v4/pull/478/commits/f6da2ab796fc45faa05cce6f90ae36b02a204621)

src/contracts/AaveOracle.sol
avniculae: [f6da2ab](https://github.com/aave/aave-v4/pull/478/commits/f6da2ab796fc45faa05cce6f90ae36b02a204621)

src/contracts/AaveOracle.sol
avniculae: I went back and forth on this as well when I initially implemented it. Main issue is that if we take the spoke in the constructor, we create a circular dependency: oracle requiring spoke in the constructor, and spoke requiring the oracle in the constructor. Shall we remove the oracle form the spoke constructor?

src/contracts/Spoke.sol
```diff
@@ -604,6 +609,12 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     // todo validate user not trying to repay more
   }
 
+  function _updateOracle(address newOracle) internal {
+    require(newOracle != address(0), InvalidOracleAddress());
```
DhairyaSethi: would like to re-use ZeroAddress error for all of these checks, slightly reduce bytecode size so all jump to the same revert destination 
having a separate name does not add a lot of benefit since these are mostly in constructors & gov operations 
lmk wdyt

avniculae: I think there's benefit for separate names in user facing fns, but constructors & gov operations are fine imo

avniculae: looks like there's not ZeroAddress in the spoke though

DhairyaSethi: pls add it and replace for all zero addy checks on gov & constructor 
let's make sure to do the same for hub (but can be tackled later if needed)

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,78 @@
...
+ * @notice Oracle contract for the Aave protocol.
+ * @dev Oracles are spoke-specific, due to the usage of reserve id as index of the _reserveSource.
+ */
+contract AaveOracle is IAaveOracle, AccessManaged {
+  /// @inheritdoc IPriceOracle
+  uint8 public immutable override DECIMALS;
+  /// @inheritdoc IPriceOracle
+  string public override DESCRIPTION;
+
+  mapping(uint256 reserveId => AggregatorV3Interface source) internal _reserveSource;
```
miguelmtzinf: personally like `_sources`

src/contracts/AaveOracle.sol
miguelmtzinf, line 1: please add a gas snapshot for calling `getReservePrice`

src/contracts/Configurator.sol
```diff
@@ -8,6 +8,12 @@ import {ILiquidityHub} from 'src/interfaces/ILiquidityHub.sol';
 import {IConfigurator} from 'src/interfaces/IConfigurator.sol';
 import {IAssetInterestRateStrategy} from 'src/interfaces/IAssetInterestRateStrategy.sol';
 
+/**
+ * @title Configurator contract
```
miguelmtzinf:
```suggestion
 * @title Configurator
```

src/contracts/Spoke.sol
miguelmtzinf: yeah, agree, let's do it separately pls

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,63 @@
...
+ */
+interface IAaveOracle is IPriceOracle {
+  /**
+   * @notice Emitted when the Aave Oracle is created.
+   * @param decimals The number of decimals used to return prices.
+   * @param description The description of the oracle.
+   */
+  event AaveOracleCreated(uint256 indexed decimals, string indexed description);
+  /**
+   * @notice Emitted when the price feed source of a reserve is updated.
```
miguelmtzinf: we use dev for events, and for errors we can use dev as well, no strong opinion

src/misc/UnitPriceFeed.sol
miguelmtzinf: I understand now. I do not think `description` is a good place to give reference about base currency and what not (because it's just an string, so for informative purposes only).

If we want to provide a reference of the base currency, we can do
`immutable BASE_CURRENCY` initialized at constructor
`constant USD_ADDRESS` as a sentinel value for USD

and then, we register this price feed as source as you are suggesting.

src/interfaces/IPriceOracle.sol
```diff
@@ -3,13 +3,26 @@ pragma solidity ^0.8.0;
...
+   * @notice Returns the number of decimals used to return prices
+   * @return The number of decimals
+   */
+  function DECIMALS() external view returns (uint8);
+
+  /**
+   * @notice Returns the description of the oracle
+   * @return The description of the oracle
+   */
+  function DESCRIPTION() external view returns (string memory);
```
miguelmtzinf: can we keep this interface as simple as possible? why do we need to bring these functions into here?
I guess `DECIMALS` is kinda mandatory now, but can we leave `DESCRIPTION` out?

src/contracts/AaveOracle.sol
avniculae: [70b1ca1](https://github.com/aave/aave-v4/pull/478/commits/70b1ca1a15957e531819c1da0d027338aced7bca)

tests/Base.t.sol
```diff
@@ -156,6 +160,21 @@ abstract contract Base is Test {
     uint256 outstandingPremium;
   }
 
+  struct ConfigureTokenListVars {
```
miguelmtzinf: i wonder why we don't reuse the same config structs

tests/Base.t.sol
miguelmtzinf: i'm fine with how it is now. we can also `vm.mockCall` right? or we want to make this permanent and don't care about the mocks

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,78 @@
...
+ * @title AaveOracle
+ * @author Aave Labs
+ * @notice Oracle contract for the Aave protocol.
+ * @dev Oracles are spoke-specific, due to the usage of reserve id as index of the _reserveSource.
+ */
+contract AaveOracle is IAaveOracle, AccessManaged {
+  /// @inheritdoc IPriceOracle
+  uint8 public immutable override DECIMALS;
+  /// @inheritdoc IPriceOracle
+  string public override DESCRIPTION;
```
miguelmtzinf, line 19: better to provide reference to the SPOKE imo

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
```diff
@@ -240,7 +239,11 @@ contract LiquidationCallValidationTest is SpokeLiquidationBase {
     assertFalse(spoke1.getUsingAsCollateral(wethReserveId, alice));
 
     // usdx collateral value drop, make sure that HF < threshold and position is liquidatable
-    oracle.setReservePrice(usdxReserveId, 0);
+    vm.mockCall(
```
miguelmtzinf, line 242: do we wanna use the helper here?

tests/Base.t.sol
```diff
@@ -1950,6 +1969,20 @@ abstract contract Base is Test {
...
 
+  function _mockReservePrice(ISpoke spoke, uint256 reserveId, uint256 price) internal {
+    require(price > 0, 'mockReservePrice: price must be positive');
+    AaveOracle oracle = AaveOracle(address(spoke.oracle()));
+    address mockPriceFeed = address(new MockPriceFeed(oracle.DECIMALS(), oracle.DESCRIPTION(), price));
+    vm.prank(ADMIN);
+    oracle.setReserveSource(reserveId, mockPriceFeed);
+  }
+
+  function _changeMockReservePriceByPercentage(ISpoke spoke, uint256 reserveId, uint256 percentage) internal {
```
```suggestion
  function _mockReservePriceByPercent(ISpoke spoke, uint256 reserveId, uint256 percent) internal {
```

tests/unit/UnitPriceFeed.t.sol
```diff
@@ -0,0 +1,44 @@
...
+    (uint80 roundId, int256 answer, uint256 startedAt, uint256 updatedAt, uint80 answeredInRound) = unitPriceFeed.latestRoundData();
+    assertEq(roundId, blockTimestamp);
+    assertEq(answer, int256(10 ** _decimals));
```
miguelmtzinf, line 39: create another fuzzing test for the answer specifically please
where you deploy a new one and call this fn to get answer

tests/unit/AaveOracle.t.sol
```diff
@@ -0,0 +1,194 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.10;
+
+import {AggregatorV3Interface} from 'src/dependencies/chainlink/AggregatorV3Interface.sol';
```
miguelmtzinf: move to Base

tests/unit/AaveOracle.t.sol
```diff
@@ -0,0 +1,194 @@
...
+  address private _source1 = makeAddr('SOURCE1');
+  address private _source2 = makeAddr('SOURCE2');
+
+  address private user = makeAddr('USER');
+
+  uint256 private constant reserveId1 = 0;
+  uint256 private constant reserveId2 = 1;
+
+  function setUp() public override {
+    super.setUp();
```
miguelmtzinf: why do we need to run super setUp?

src/contracts/AaveOracle.sol
```diff
@@ -21,23 +23,21 @@ contract AaveOracle is IAaveOracle, AccessManaged {
...
+    SPOKE = spoke_;
     DECIMALS = decimals_;
     DESCRIPTION = description_;
     emit AaveOracleCreated(decimals_, description_);
   }
 
-  /// @inheritdoc IAaveOracle
-  function setReserveSource(uint256 reserveId, address source) external override restricted {
+  /// @inheritdoc IPriceOracle
+  function configureReserve(uint256 reserveId, bytes calldata configData) external override {
```
miguelmtzinf: idk this is needed tbh. We made this in Hub because it's immutable and we foresee changes on the input params. It's not the case here

src/contracts/Spoke.sol
```diff
@@ -48,12 +48,9 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   /**
    * @dev Constructor.
    * @dev The authority should implement the AccessManaged interface to control access.
-   * @param oracle_ The address of the price oracle contract used for asset valuations.
    * @param authority_ The address of the authority contract which manages permissions.
    */
-  constructor(address oracle_, address authority_) AccessManaged(authority_) {
-    _updateOracle(oracle_);
-
+  constructor(address authority_) AccessManaged(authority_) {
```
miguelmtzinf, line 52: damn, i d like to enforce the oracle setup at deployment...

tests/Base.t.sol
avniculae: we can do mockCall, but they are a bit problematic bc sometimes we do `clearAllMocks`

src/contracts/Spoke.sol
avniculae: updated this issue https://github.com/aave/aave-v4/issues/397

src/contracts/AaveOracle.sol
avniculae: wdym?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
avniculae: we cannot use the helper here bc 0 is an invalid price

src/contracts/AaveOracle.sol
avniculae: Then the oracle in the spoke will be of type IAaveOracle instead of IPriceOracle. is this something we want?

src/contracts/Spoke.sol
avniculae: I discussed this with Dhairya as well -- I think it's fine bc we enforce configuring a reserve price when adding a new reserve, so we know for sure oracle is configured before any reserve is added

avniculae: wdyt?

src/contracts/AaveOracle.sol
avniculae: I don't think it's very relevant, given that we use a mock price feed, wdyt?

tests/unit/UnitPriceFeed.t.sol
avniculae: what is the benefit of this, can you pls elaborate?

src/contracts/AaveOracle.sol
miguelmtzinf: yes

src/contracts/Spoke.sol
```diff
@@ -41,27 +44,37 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+    oracle = IPriceOracle(newOracle);
+    emit OracleUpdated(newOracle);
+  }
+
+  function updateOracleConfig(
+    uint256 reserveId,
+    bytes calldata oracleConfigData
+  ) external restricted {
+    require(reserveId < reserveCount, ReserveNotListed());
+    require(address(oracle) != address(0), InvalidOracleAddress());
```
miguelmtzinf: dup

tests/unit/AaveOracle.t.sol
avniculae: [f6eeceb](https://github.com/aave/aave-v4/pull/478/commits/f6eecebb1aa1e6346886ca43a11486722e7f4517)

tests/unit/AaveOracle.t.sol
avniculae: [f6eeceb](https://github.com/aave/aave-v4/pull/478/commits/f6eecebb1aa1e6346886ca43a11486722e7f4517)

tests/Base.t.sol
avniculae: [f6eeceb](https://github.com/aave/aave-v4/pull/478/commits/f6eecebb1aa1e6346886ca43a11486722e7f4517)

tests/Base.t.sol
avniculae: [f6eeceb](https://github.com/aave/aave-v4/pull/478/commits/f6eecebb1aa1e6346886ca43a11486722e7f4517)

src/interfaces/IPriceOracle.sol
avniculae: [f6eeceb](https://github.com/aave/aave-v4/pull/478/commits/f6eecebb1aa1e6346886ca43a11486722e7f4517)

src/interfaces/IAaveOracle.sol
avniculae: [f6eeceb](https://github.com/aave/aave-v4/pull/478/commits/f6eecebb1aa1e6346886ca43a11486722e7f4517)

src/contracts/Configurator.sol
avniculae: [f6eeceb](https://github.com/aave/aave-v4/pull/478/commits/f6eecebb1aa1e6346886ca43a11486722e7f4517)

src/contracts/AaveOracle.sol
avniculae: [f6eeceb](https://github.com/aave/aave-v4/pull/478/commits/f6eecebb1aa1e6346886ca43a11486722e7f4517)

src/misc/UnitPriceFeed.sol
avniculae: created #510

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,77 @@
...
+   * @dev Constructor.
+   * @param spoke_ The address of the spoke contract.
+   * @param decimals_ The number of decimals for the oracle.
+   * @param description_ The description of the oracle.
+   */
+  constructor(address spoke_, uint8 decimals_, string memory description_) {
+    SPOKE = spoke_;
+    DECIMALS = decimals_;
+    DESCRIPTION = description_;
+    emit AaveOracleCreated(decimals_, description_);
```
DhairyaSethi: do not see the use case of this event tbh, would avoid

src/contracts/Spoke.sol
```diff
@@ -604,6 +620,11 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     // todo validate user not trying to repay more
   }
 
+  function _configureReservePrice(uint256 reserveId, bytes calldata oracleConfigData) internal {
```
DhairyaSethi: pls call this configure (reserve) price feed or something for clarity

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  /// @inheritdoc IPriceOracle
+  function configureReserve(uint256 reserveId, bytes calldata configData) external override {
+    require(msg.sender == SPOKE, OnlySpoke());
+    address source = abi.decode(configData, (address));
+    AggregatorV3Interface targetSource = AggregatorV3Interface(source);
```
DhairyaSethi:
```suggestion
    AggregatorV3Interface source = AggregatorV3Interface(abi.decode(configData, (address)));
```
and just cast for the event

tests/unit/AaveOracle.t.sol
```diff
@@ -0,0 +1,184 @@
...
+  function test_decimals() public {
+    assertEq(oracle.DECIMALS(), _decimals);
+  }
+
+  function test_description() public {
+    assertEq(oracle.DESCRIPTION(), _description);
+  }
+
+  function test_configureReserve_revertsWith_OnlySpoke() public {
+    vm.expectRevert(abi.encodeWithSelector(IPriceOracle.OnlySpoke.selector));
```
DhairyaSethi:
```suggestion
    vm.expectRevert(IPriceOracle.OnlySpoke.selector);
```

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  function _getSourcePrice(uint256 reserveId) internal view returns (uint256) {
+    AggregatorV3Interface source = _sources[reserveId];
+    require(address(source) != address(0), InvalidSource(reserveId));
+
+    (, int256 price, , , ) = source.latestRoundData();
+    require(price > 0, InvalidPrice(reserveId));
+
+    return uint256(price);
```
DhairyaSethi, line 73: fallback oracle no, fine but what about staleness checks?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
```diff
@@ -240,7 +239,11 @@ contract LiquidationCallValidationTest is SpokeLiquidationBase {
     assertFalse(spoke1.getUsingAsCollateral(wethReserveId, alice));
 
     // usdx collateral value drop, make sure that HF < threshold and position is liquidatable
-    oracle.setReservePrice(usdxReserveId, 0);
```
DhairyaSethi, line 243: for oracle prices i would still set instead of mocking since we already use a mock price feed

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.EdgeCases.t.sol
```diff
@@ -91,10 +84,7 @@ contract LiquidationCallEdgeCasesTest is SpokeLiquidationBase {
     Utils.borrow(spoke1, _usdyReserveId(spoke1), alice, borrowAmount2, alice);
 
     // price drops to reach liquidatable state
-    oracle.setReservePrice(
-      _wethReserveId(spoke1),
-      calcNewPrice(oracle.getReservePrice(_wethReserveId(spoke1)), 50_00)
-    ); // weth price drops by 50%
+    _mockReservePriceByPercent(spoke1, _wethReserveId(spoke1), 50_00);
```
DhairyaSethi, line 87: same

src/contracts/AaveOracle.sol
avniculae: noted it here #510

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
avniculae: we cannot set mock price feed here bc 0 price is invalid

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.EdgeCases.t.sol
avniculae: same to what? can you pls elaborate?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
DhairyaSethi: why not? the mock price feed setter has no checks
while adding on reserve it doesn't need to have 0 price

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.EdgeCases.t.sol
DhairyaSethi: use mock price feeds setter instead of mocking the call

tests/unit/AaveOracle.t.sol
avniculae: [69081e8](https://github.com/aave/aave-v4/pull/478/commits/69081e8dcb27ed301cc565f8f05d2a16709f4348)

src/contracts/AaveOracle.sol
avniculae: [69081e8](https://github.com/aave/aave-v4/pull/478/commits/69081e8dcb27ed301cc565f8f05d2a16709f4348)

src/contracts/Spoke.sol
avniculae: N/A anymore: [69081e8](https://github.com/aave/aave-v4/pull/478/commits/69081e8dcb27ed301cc565f8f05d2a16709f4348)

src/contracts/AaveOracle.sol
avniculae: [69081e8](https://github.com/aave/aave-v4/pull/478/commits/69081e8dcb27ed301cc565f8f05d2a16709f4348)

src/contracts/Spoke.sol
avniculae: [69081e8](https://github.com/aave/aave-v4/pull/478/commits/69081e8dcb27ed301cc565f8f05d2a16709f4348)

src/contracts/AaveOracle.sol
avniculae: [69081e8](https://github.com/aave/aave-v4/pull/478/commits/69081e8dcb27ed301cc565f8f05d2a16709f4348)

src/contracts/Spoke.sol
avniculae: (https://github.com/aave/aave-v4/pull/478#discussion_r2185139567)

src/contracts/Spoke.sol
```diff
@@ -41,27 +43,33 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
   // Governance
   // /////
 
+  function updateOracle(address newOracle) external restricted {
+    require(newOracle != address(0), InvalidOracle());
+    oracle = IAaveOracle(newOracle);
+    emit OracleUpdated(newOracle);
+  }
+
+  function updateReserveSource(uint256 reserveId, address source) external restricted {
```
miguelmtzinf: wondering if we should rename to `updateReservePriceSource(reserveId, priceSource)`

src/contracts/Spoke.sol
```diff
@@ -73,6 +81,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   function addReserve(
     uint256 assetId,
     address hub,
+    address source,
```
```suggestion
    address priceSource,
```

src/interfaces/ISpoke.sol
```diff
@@ -46,7 +46,8 @@ interface ISpoke is IMulticall, IAccessManaged {
...
+  event ReserveSourceUpdated(uint256 indexed reserveId, address indexed source);
```
miguelmtzinf: same

avniculae: [6aff710](https://github.com/aave/aave-v4/pull/478/commits/6aff7106c36625f6c83475036e381ccb1ade1e3c)

src/contracts/Spoke.sol
avniculae: [6aff710](https://github.com/aave/aave-v4/pull/478/commits/6aff7106c36625f6c83475036e381ccb1ade1e3c)

src/contracts/Spoke.sol
avniculae: [6aff710](https://github.com/aave/aave-v4/pull/478/commits/6aff7106c36625f6c83475036e381ccb1ade1e3c)

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
avniculae: `setReserveSource` has checks though

DhairyaSethi: yes and initial price doesn't need to be zero when setting

tests/Base.t.sol
```diff
@@ -199,18 +206,21 @@ abstract contract Base is Test {
     // Grant roles with 0 delay
     accessManager.grantRole(Roles.HUB_ADMIN_ROLE, ADMIN, 0);
     accessManager.grantRole(Roles.SPOKE_ADMIN_ROLE, ADMIN, 0);
+    accessManager.grantRole(Roles.ORACLE_ADMIN_ROLE, ADMIN, 0);
     accessManager.grantRole(Roles.HUB_ADMIN_ROLE, HUB_ADMIN, 0);
     accessManager.grantRole(Roles.SPOKE_ADMIN_ROLE, SPOKE_ADMIN, 0);
+    accessManager.grantRole(Roles.ORACLE_ADMIN_ROLE, ORACLE_ADMIN, 0);
```
miguelmtzinf: i think using SPOKE_ADMIN_ROLE is fine for now, but nbd

tests/Base.t.sol
```diff
@@ -1950,6 +1937,25 @@ abstract contract Base is Test {
...
+    spoke.updateReservePriceSource(reserveId, mockPriceFeed);
+  }
+
+  function _mockReservePriceByPercent(ISpoke spoke, uint256 reserveId, uint256 percentage) internal {
+    uint256 initialPrice = spoke.oracle().getReservePrice(reserveId);
+    uint256 newPrice = initialPrice.percentMulDown(percentage);
+    _mockReservePrice(spoke, reserveId, newPrice);
+  }
+
+  function _getMockReserveSource(ISpoke spoke, uint256 price) internal returns (address) {
```
miguelmtzinf: such a weird getter, you pass a price that you later return with no change at all
also, I d rename to `_getReservePriceSource` since it does not need to be mocked (we are returning an address)

miguelmtzinf: ah, you are deploying a new one with this price, let's call it `deployMockReservePriceSource` then, or just `deployMockPriceFeed`

tests/unit/AaveOracle.t.sol
```diff
@@ -0,0 +1,182 @@
...
+    test_spoke();
+    test_decimals();
+    test_description();
+  }
+
+  function test_spoke() public {
+    assertEq(oracle.SPOKE(), address(spoke1));
+  }
+
+  function test_decimals() public {
```
miguelmtzinf, line 45: can we create a new test to fuzz decimals at deployment? to test 10**decimals work well in `getReservePrice`?

avniculae: [f86b744](https://github.com/aave/aave-v4/pull/478/commits/f86b744eeb936d33d500387fba96489fdc9fd5c5)

tests/Base.t.sol
avniculae: [f86b744](https://github.com/aave/aave-v4/pull/478/commits/f86b744eeb936d33d500387fba96489fdc9fd5c5)

tests/Base.t.sol
avniculae: [f86b744](https://github.com/aave/aave-v4/pull/478/commits/f86b744eeb936d33d500387fba96489fdc9fd5c5)

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.EdgeCases.t.sol
avniculae: this is using mock price feeds

src/contracts/Spoke.sol
```diff
@@ -604,6 +615,12 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     // todo validate user not trying to repay more
   }
 
+  function _updateReservePriceSource(uint256 reserveId, address priceSource) internal {
+    require(address(oracle) != address(0), InvalidOracle());
```
CheyenneAtapour, line 619: Don't you want to validate the new `priceSource`?

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,65 @@
...
+   * @dev Thrown when the price feed source returns an invalid price (non-positive).
+   * @param reserveId The identifier of the reserve.
+   */
+  error InvalidPrice(uint256 reserveId);
+
+  /**
+   * @notice Sets the price feed source of a reserve.
+   * @dev Must be called by the spoke.
+   * @param reserveId The identifier of the reserve.
+   * @param source The price feed source of the reserve.
```
CheyenneAtapour, line 40: Do you want to leave a dev comment explaining what kind of contract / interface the source must be/implement?

tests/invariant/LiquidityHubHandler.t.sol
```diff
@@ -3,13 +3,15 @@ pragma solidity ^0.8.0;
...
+import {IPriceOracle} from 'src/interfaces/IPriceOracle.sol';
+import {AaveOracle} from 'src/contracts/AaveOracle.sol';
 import {LiquidityHub} from 'src/contracts/LiquidityHub.sol';
 import {Spoke} from 'src/contracts/Spoke.sol';
 import {TreasurySpoke} from 'src/contracts/TreasurySpoke.sol';
 import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
 import {AccessManager} from 'src/dependencies/openzeppelin/AccessManager.sol';
-import '../mocks/MockPriceOracle.sol';
 import '../mocks/MockERC20.sol';
+import {MockPriceFeed} from '../mocks/MockPriceFeed.sol';
```
CheyenneAtapour, line 14: nbd because it's a test, and also not your fault here, but just fyi on the orderings here: https://www.notion.so/avara/Smart-Contracts-Style-Guide-e4e5b80ba227401bbda022f0d4d79518?source=copy_link#0d75dcfb4c724ada941b6ba8770d2ed4

tests/unit/AaveOracle.t.sol
```diff
@@ -0,0 +1,191 @@
+// SPDX-License-Identifier: BUSL-1.1
```
yan-man: different than other tests, was that on purpose?

src/contracts/Spoke.sol
yan-man: happens a few layers down in `_getSourcePrice` but maybe also worth it here in case the validation would ever change on AaveOracle

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
avniculae: but if the price feed will ever return 0, then the getReservePrice will fail

src/contracts/Spoke.sol
avniculae: I thought of this as well (ie not relying on the aave oracle to revert when you set a price feed). while we can ensure that the price is valid now, it can be invalid in the next block. hence, it felt like a check to be done by the oracle, if they want to. wdyt?

tests/unit/AaveOracle.t.sol
avniculae: I might have looked at the wadmath tests, [7754080](https://github.com/aave/aave-v4/pull/478/commits/7754080248b60ba8b3629a2ade3b508d02d562ce)

tests/invariant/LiquidityHubHandler.t.sol
avniculae: I'm aware of the convention, but not sure what's wrong with this line. did some reordering though, lmk your thoughts [7754080](https://github.com/aave/aave-v4/pull/478/commits/7754080248b60ba8b3629a2ade3b508d02d562ce)

src/interfaces/IAaveOracle.sol
avniculae: [7754080](https://github.com/aave/aave-v4/pull/478/commits/7754080248b60ba8b3629a2ade3b508d02d562ce)

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,66 @@
...
+ * @author Aave Labs
+ * @notice Interface for the Aave Oracle
+ */
+interface IAaveOracle is IPriceOracle {
+  /**
+   * @dev Emitted when the price feed source of a reserve is updated.
+   * @param reserveId The identifier of the reserve.
+   * @param source The price feed source of the reserve.
+   */
+  event ReserveSourceUpdated(uint256 indexed reserveId, AggregatorV3Interface indexed source);
```
miguelmtzinf: I d suggest to emit source as address here

src/interfaces/IAaveOracle.sol
miguelmtzinf: move it up, right before any param

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,74 @@
...
+  constructor(address spoke_, uint8 decimals_, string memory description_) {
+    SPOKE = spoke_;
+    DECIMALS = decimals_;
+    DESCRIPTION = description_;
+  }
+
+  /// @inheritdoc IAaveOracle
+  function setReserveSource(uint256 reserveId, address source) external override {
+    require(msg.sender == SPOKE, OnlySpoke());
+    AggregatorV3Interface targetSource = AggregatorV3Interface(source);
```
miguelmtzinf, line 38: this intermediate variable feels unnecessary

src/contracts/AaveOracle.sol
```diff
@@ -0,0 +1,74 @@
...
+    require(targetSource.decimals() == DECIMALS, InvalidSourceDecimals(reserveId));
+    _sources[reserveId] = targetSource;
```
miguelmtzinf, line 40: i d cast here, and use `source` directly on the rest

src/interfaces/IAaveOracle.sol
avniculae: I casted bc of this comment https://github.com/aave/aave-v4/pull/478#discussion_r2189254550

src/contracts/AaveOracle.sol
avniculae: I need to cast it to call decimals() on the line above as well

src/contracts/AaveOracle.sol
avniculae: replied https://github.com/aave/aave-v4/pull/478#discussion_r2191806914

src/interfaces/IAaveOracle.sol
avniculae: [5f31885](https://github.com/aave/aave-v4/pull/478/commits/5f31885754d2cac368b0cc8fedfd150c83ac9ca2)

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: AGPL-3.0
+pragma solidity ^0.8.0;
+
+import {AggregatorV3Interface} from 'src/dependencies/chainlink/AggregatorV3Interface.sol';
```
miguelmtzinf: rm

src/interfaces/IAaveOracle.sol
```diff
@@ -0,0 +1,66 @@
...
+   * @param reserveId The identifier of the reserve.
+   */
+  error InvalidPrice(uint256 reserveId);
+
+  /**
+   * @notice Sets the price feed source of a reserve.
+   * @dev Must be called by the spoke.
+   * @param reserveId The identifier of the reserve.
+   * @param source The price feed source of the reserve.
+   * @dev The source must implement the AggregatorV3Interface.
```
miguelmtzinf: move up pls

avniculae: [d6e78c6](https://github.com/aave/aave-v4/pull/478/commits/d6e78c6ccf75203f4db0311b444dc140ea498fa7)

src/interfaces/IAaveOracle.sol
avniculae: [d6e78c6](https://github.com/aave/aave-v4/pull/478/commits/d6e78c6ccf75203f4db0311b444dc140ea498fa7)

src/contracts/Spoke.sol
yan-man: less on oracle price but more on the `address(source) != address(0)` imo

tests/invariant/LiquidityHubHandler.t.sol
CheyenneAtapour: Wasn't that line in particular. This ordering is great! Thanks

src/contracts/Spoke.sol
CheyenneAtapour: I see, it looks good as is to me

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Validation.t.sol
DhairyaSethi: yes and protocol works under that assumption

PR #489, opened 2025-06-29
fix: Simplify permission check of updateUserRp

miguelmtzinf:
Based on [insights from OZ team about AccessManaged](https://forum.openzeppelin.com/t/complex-conditional-logic-for-function-access-with-accessmanaged/44414/2), I suggest the simplification of the `updateUserRiskPremium` permission check.


tests/Base.t.sol
```diff
@@ -199,15 +199,15 @@ abstract contract Base is Test {
...
 
     // Grant responsibilities to roles
     // Spoke Admin functionalities
-    bytes4[] memory selectors = new bytes4[](4);
+    bytes4[] memory selectors = new bytes4[](5);
     selectors[0] = ISpoke.updateLiquidationConfig.selector;
     selectors[1] = ISpoke.addReserve.selector;
     selectors[2] = ISpoke.updateReserveConfig.selector;
     selectors[3] = ISpoke.updateDynamicReserveConfig.selector;
+    selectors[4] = ISpoke.updateUserRiskPremium.selector;
```
DhairyaSethi, line 211: since we need granularity, this should still come under a separate role, no?

src/contracts/Spoke.sol
```diff
@@ -380,10 +378,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   function updateUserRiskPremium(address user) external {
     (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
     bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
-    require(
-      !premiumIncrease || msg.sender == user || _hasUpdateUserRiskPremiumRole(msg.sender),
-      Unauthorized()
-    );
+    // check permissions if premium increases or not called by user
+    if (premiumIncrease || user != msg.sender) {
+      _checkCanCall(_msgSender(), _msgData());
```
DhairyaSethi, line 401: personally does not feel cleaner haha 
we can go this way so we have to room to expand the responsibilities for this auth

tests/Base.t.sol
CheyenneAtapour: I had this under a governor role, but Miguel wanted to move it to `SPOKE_ADMIN`. We can always move that around whenever we like, so I think nbd

DhairyaSethi: we loose granularity to give those role to a contract without giving access to ALL other spoke admin selectors, no?

DhairyaSethi: ah you mean with this setup it can changed later, sure

miguelmtzinf: yeah, no need to create a separate one now

src/contracts/Spoke.sol
```diff
@@ -380,10 +378,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   function updateUserRiskPremium(address user) external {
     (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(user);
     bool premiumIncrease = _notifyRiskPremiumUpdate(type(uint256).max, user, userRiskPremium);
-    require(
-      !premiumIncrease || msg.sender == user || _hasUpdateUserRiskPremiumRole(msg.sender),
-      Unauthorized()
-    );
+    // check permissions if premium increases or not called by user
+    if (premiumIncrease || user != msg.sender) {
```
```suggestion
    // check permissions if premium increases and not called by user
    if (premiumIncrease && user != msg.sender) {
```

miguelmtzinf:
This was added in #455

PR #490, opened 2025-06-29
fix: Makes fee receiver mandatory

### LINKED ISSUE #480: Do not allow zero address as fee receiver
We have a bit of complexity in a couple of places just because we wanna allow `address(0)` as `feeReceiver`.
- `liquidityFee` cannot be non-zero if `feeReceiver` is zero
- `spoke.liquidationFee` cannot be non-zero if `feeReceiver` is zero.

Recommendation is to not allow having zero address as fee receiver, so code is simpler (removing edge cases when we accrue fees to zero address as fee receiver).

miguelmtzinf:
- A non-zero address for the fee receiver is now mandatory when adding an asset.
- The update function also enforces a non-zero value.
- This ensures that accrued fees are always directed to a valid, configured fee receiver.
- The fee receiver must be set up separately as a spoke via `Hub.addSpoke` to define its configuration, such as caps and active status.
- The Configurator automatically configures the fee receiver as Spoke when adding the asset.

Closes #480

### DIFF
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -1010,6 +1010,18 @@ abstract contract Base is Test {
     liquidityHub.updateAssetConfig(assetId, assetConfig);
   }
 
+  function updateAssetFeeReceiver(
+    ILiquidityHub liquidityHub,
+    uint256 assetId,
+    address newFeeReceiver
+  ) internal {
+    DataTypes.AssetConfig memory assetConfig = liquidityHub.getAsset(assetId).config;
+    assetConfig.feeReceiver = newFeeReceiver;
+
+    vm.prank(HUB_ADMIN);
+    liquidityHub.updateAssetConfig(assetId, assetConfig);
+  }
+
   function updateReserveFrozenFlag(ISpoke spoke, uint256 reserveId, bool newFrozenFlag) internal {
     DataTypes.ReserveConfig memory config = spoke.getReserve(reserveId).config;
     config.frozen = newFrozenFlag;
--- a/tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
@@ -290,6 +290,42 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
     assertEq(hub.getSpokeSuppliedShares(assetId, newFeeReceiver), newFees);
   }
 
+  /// Updates the fee receiver to an existing spoke of the hub, so ends up with existing supplied shares plus accrued fees
+  function test_updateAssetConfig_fuzz_UseExistingSpokeAsFeeReceiver(uint256 assetId) public {
+    assetId = bound(assetId, 0, hub.getAssetCount() - 1);
+
+    address oldFeeReceiver = _getFeeReceiver(assetId);
+    address newFeeReceiver = address(spoke1);
+
+    uint256 amount = 1000e18;
+    _addLiquidity(assetId, amount);
+    _drawLiquidity(assetId, amount, true);
+
+    uint256 oldReceiverFees = hub.getSpokeSuppliedShares(assetId, oldFeeReceiver);
+    assertTrue(oldReceiverFees > 0);
+
+    // spoke1 adds some assets
+    Utils.add({
+      hub: hub,
+      assetId: assetId,
+      spoke: address(spoke2),
+      amount: amount,
+      user: bob,
+      to: address(spoke2)
+    });
+    uint256 newReceiverFees = hub.getSpokeSuppliedShares(assetId, newFeeReceiver);
+
+    updateAssetFeeReceiver(hub, assetId, newFeeReceiver);
+
+    skip(365 days);
+
+    // new fee receiver keeps the existing supplied shares and earns more via fees accrual
+    assertTrue(hub.getSpokeSuppliedShares(assetId, newFeeReceiver) > newReceiverFees);
+
+    // old fee receiver keeps the accrued fees
+    assertEq(hub.getSpokeSuppliedShares(assetId, oldFeeReceiver), oldReceiverFees);
+  }
+
   /// Triggers accrual when liquidity fee update, based on old liquidity fee
   function test_updateAssetConfig_fuzz_LiquidityFee(uint256 assetId, uint256 liquidityFee) public {
     assetId = bound(assetId, 0, hub.getAssetCount() - 1);


src/contracts/Configurator.sol
```diff
@@ -31,20 +31,28 @@ contract Configurator is Ownable, IConfigurator {
...
   function addAsset(
     address hub,
-    address asset,
+    address underlying,
     uint8 decimals,
+    address feeReceiver,
     address irStrategy
   ) external override onlyOwner returns (uint256) {
-    return ILiquidityHub(hub).addAsset(asset, decimals, irStrategy);
+    return ILiquidityHub(hub).addAsset(underlying, decimals, feeReceiver, irStrategy);
```
avniculae: I think the configurator should now perform `addSpoke` to configure the `feeReceiver` spoke before calling `addAsset`. wdyt?

tests/Base.t.sol
```diff
@@ -304,7 +304,12 @@ abstract contract Base is Test {
...
     vm.startPrank(ADMIN);
     // add WETH
-    hub.addAsset(address(tokenList.weth), tokenList.weth.decimals(), address(irStrategy));
+    hub.addAsset(
+      address(tokenList.weth),
+      tokenList.weth.decimals(),
+      address(treasurySpoke),
+      address(irStrategy)
+    );
     hub.addSpoke(wethAssetId, address(treasurySpoke), spokeConfig);
```
avniculae, line 313: To be removed once we solve https://github.com/aave/aave-v4/pull/490/files#r2173853657

avniculae: same for the `addSpoke` operations below

tests/unit/Configurator.t.sol
```diff
@@ -304,33 +314,30 @@ contract ConfiguratorTest is LiquidityHubBase {
...
+    address caller
   ) public {
     vm.assume(caller != CONFIGURATOR_ADMIN);
     vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, caller));
     vm.prank(caller);
     configurator.updateFeeReceiver(address(hub), vm.randomUint(), vm.randomAddress());
   }
 
-  function test_updateFeeReceiver_fuzz_revertsWith_InvalidFeeReceiver(uint256 assetId) public {
+  function test_updateFeeReceiver_fuzz_zero_feeReceiver_reverts(uint256 assetId) public {
```
avniculae: why not `revertsWith_InvalidSpoke` ?

tests/unit/Configurator.t.sol
```diff
@@ -465,74 +466,69 @@ contract ConfiguratorTest is LiquidityHubBase {
...
     expectedConfig.feeReceiver = feeReceiver;
     expectedConfig.liquidityFee = liquidityFee;
 
-    vm.expectCall(
-      address(hub),
-      abi.encodeCall(ILiquidityHub.updateAssetConfig, (assetId, expectedConfig))
-    );
+    // vm.expectCall(
+    //   address(hub),
+    //   abi.encodeCall(ILiquidityHub.updateAssetConfig, (assetId, expectedConfig))
```
avniculae: why commented out?

tests/unit/Configurator.t.sol
```diff
@@ -543,21 +539,16 @@ contract ConfiguratorTest is LiquidityHubBase {
   function test_updateFeeConfig_Scenario() public {
     // set same fee receiver and change liquidity fee
     test_updateFeeConfig_fuzz(daiAssetId, 18_00, address(treasurySpoke));
-    // set new fee receiver and liquidity fee
+    // // set new fee receiver and liquidity fee
```
avniculae: double `//`

avniculae: same below

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
avniculae, line 1: shall we add a test to set fee to 0, and then back on, and make sure fee receiver did not earn any fees for the period when the fee was 0?

tests/unit/Configurator.t.sol
```diff
@@ -304,33 +314,30 @@ contract ConfiguratorTest is LiquidityHubBase {
...
-    vm.expectRevert(ILiquidityHub.InvalidFeeReceiver.selector);
+    // reverts when adding zero as new spoke
+    vm.expectRevert(ILiquidityHub.InvalidSpoke.selector);
     vm.prank(CONFIGURATOR_ADMIN);
     configurator.updateFeeReceiver(address(hub), assetId, address(0));
   }
 
   function test_updateFeeReceiver_revertsWith_InvalidFeeReceiver(uint256 assetId) public {
-    test_updateFeeReceiver_fuzz_revertsWith_InvalidFeeReceiver(daiAssetId);
+    test_updateFeeReceiver_fuzz_zero_feeReceiver_reverts(daiAssetId);
```
yan-man: continuation from what Alex mentioned

tests/Base.t.sol
miguelmtzinf: but we are using hub directly, not configurator.

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
miguelmtzinf: is `test_updateAssetConfig_fuzz_FromZeroLiquidityFee` what you are thinking about?

tests/Base.t.sol
avniculae: you are right, all good then

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
avniculae: is a subset of what I was thinking, most probably is enough

src/contracts/Configurator.sol
```diff
@@ -31,20 +31,55 @@ contract Configurator is Ownable, IConfigurator {
   /// @inheritdoc IConfigurator
   function addAsset(
     address hub,
-    address asset,
+    address underlying,
+    address feeReceiver,
     address irStrategy
   ) external override onlyOwner returns (uint256) {
-    return ILiquidityHub(hub).addAsset(asset, IERC20Metadata(asset).decimals(), irStrategy);
+    ILiquidityHub targetHub = ILiquidityHub(hub);
```
yan-man, line 38: should we also change arg param input to `targetHub` if you want to refer to it that way in the function too

tests/unit/Configurator.t.sol
```diff
@@ -465,63 +477,58 @@ contract ConfiguratorTest is LiquidityHubBase {
     address feeReceiver
   ) public {
     assetId = bound(assetId, 0, hub.getAssetCount() - 1);
-    if (feeReceiver == address(0)) {
-      liquidityFee = 0;
-    } else {
-      liquidityFee = bound(liquidityFee, 0, PercentageMathExtended.PERCENTAGE_FACTOR);
-    }
+    liquidityFee = bound(liquidityFee, 0, PercentageMathExtended.PERCENTAGE_FACTOR);
+    vm.assume(feeReceiver != address(0));
```
yan-man:
```suggestion
    vm.assume(feeReceiver != address(0) && feeReceiver != oldConfig.feeReceiver );
```
?

tests/unit/Configurator.t.sol
yan-man, line 1: not sure what file it should go into but perhaps we should add a test for if a regular already-registered spoke gets set as feeReceiver, and make sure that the accrual works properly for it

src/contracts/Configurator.sol
miguelmtzinf: we cannot use the same name for the input param and local variable, and I d prefer to pass an address as input

tests/unit/Configurator.t.sol
miguelmtzinf: kinda similar to `test_updateAssetConfig_fuzz_ReuseFeeReceiver`
also note that we cannot re-use a spoke when adding an asset (because all spokes are new).

src/contracts/Configurator.sol
yan-man: you only want to use underscore param arg inputs on constructor in general? `variable_`

tests/unit/Configurator.t.sol
yan-man: `test_updateAssetConfig_fuzz_ReuseFeeReceiver` is specifically for a previous feeReceiver, but are there tests for a non-treasury spoke being used as `feeReceiver` (we don't intend to prevent this behavior right)? ie during an `updateAssetConfig` instead of an `addAsset`

src/contracts/Configurator.sol
```diff
@@ -31,20 +31,55 @@ contract Configurator is Ownable, IConfigurator {
   /// @inheritdoc IConfigurator
   function addAsset(
     address hub,
-    address asset,
+    address underlying,
```
CheyenneAtapour, line 34: what's the motivation for this name change?

src/interfaces/IConfigurator.sol
```diff
@@ -31,27 +31,38 @@ interface IConfigurator {
...
+   * @notice Adds a new asset to the hub.
+   * @dev Retrieves the decimals of the underlying asset from its ERC20 contract.
+   * @dev The fee receiver is automatically added as a spoke with maximum caps.
    * @param hub The address of the Hub contract.
-   * @param asset The address of the asset.
+   * @param underlying The address of the underlying asset.
+   * @param feeReceiver The address of the fee receiver spoke.
    * @param irStrategy The address of the interest rate strategy contract.
-   * @return The id of the registered asset.
+   * @return The unique identifier of the added asset.
```
CheyenneAtapour, line 41: This kinda suggests you can't relist on the hub, but not a big deal. Could just call it "identifier" though

src/contracts/Configurator.sol
```diff
@@ -31,20 +31,55 @@ contract Configurator is Ownable, IConfigurator {
...
+        drawCap: type(uint256).max,
+        active: true
+      })
+    );
+
+    return assetId;
   }
 
   /// @inheritdoc IConfigurator
   function addAsset(
```
CheyenneAtapour, line 61: Should we just have the 1 function to `addAsset` above which just fetches decimals from the token contract, rather than this one extra which takes it as input?

src/interfaces/ILiquidityHub.sol
```diff
@@ -74,13 +74,32 @@ interface ILiquidityHub is IAccessManaged {
...
-  error InvalidAssetAddress();
+  error InvalidUnderlying();
   error InvalidDebtChange();
   error InvalidFeeReceiver();
   error SpokeNotActive();
 
-  function addAsset(address asset, uint8 decimals, address irStrategy) external returns (uint256);
+  /**
+   * @notice Adds a new asset to the hub.
+   * @dev The fee receiver must be configured as a Spoke separately
```
CheyenneAtapour: nit: Period at the end of this line

tests/unit/Configurator.t.sol
```diff
@@ -304,33 +325,30 @@ contract ConfiguratorTest is LiquidityHubBase {
...
+  function test_updateFeeReceiver_revertsWith_InvalidSpoke() public {
+    test_updateFeeReceiver_fuzz_revertsWith_InvalidSpoke(daiAssetId);
   }
 
   function test_updateFeeReceiver_fuzz(uint256 assetId, address feeReceiver) public {
     assetId = bound(assetId, 0, hub.getAssetCount() - 1);
-    if (feeReceiver == address(0)) {
-      test_updateLiquidityFee_fuzz(assetId, 0);
-    }
+    vm.assume(feeReceiver != address(0));
```
CheyenneAtapour: We kinda wanted to avoid using `vm.assume` and introducing more where possible so eventually we can run at very high fuzz runs.

tests/unit/Configurator.t.sol
```diff
@@ -545,19 +548,14 @@ contract ConfiguratorTest is LiquidityHubBase {
     test_updateFeeConfig_fuzz(daiAssetId, 18_00, address(treasurySpoke));
```
CheyenneAtapour: If you change the order here so this one comes after the following: `test_updateFeeConfig_fuzz(daiAssetId, 4_00, makeAddr('newFeeReceiver'));`, the tests will pass. It's currently failing because on the assume, the old fee receiver matches the one passed by argument.

But the test wants to test same fee receiver, so perhaps can handle this case in `test_updateFeeConfig_fuzz` test

CheyenneAtapour: On second look, I think this test case should just be a test call to `updateLiquidityFee` on the configurator

src/contracts/Configurator.sol
miguelmtzinf: that was my intention, yes

src/contracts/Configurator.sol
miguelmtzinf: it was introduced in the configurator PR iirc, so this is aligned. Reasoning is that we have already `asset` for the asset data struct, so we took `underlying` from V3 to refer the asset address

src/interfaces/IConfigurator.sol
miguelmtzinf: added a comment to clarify that an asset can be added multiple times. Lmk if you still think that we need to rephrase the return line

src/contracts/Configurator.sol
miguelmtzinf: we wanted to have the 2 versions for flexibility

tests/unit/Configurator.t.sol
miguelmtzinf: changed to condition + expectRevert

tests/unit/Configurator.t.sol
miguelmtzinf: added

tests/unit/Configurator.t.sol
DhairyaSethi: oh god no this is the perfect use of vm.assume 😭

PR #491, opened 2025-06-29
fix: Make Spoke.useAsCollateral no-op

### LINKED ISSUE #483: Behavior when Spoke.useAsCollateral does not change state (revert vs noOp)
Currently, `Spoke.useAsCollateral` does not revert when collateral status is not changed. (https://github.com/aave/aave-v4/pull/407)

Design decision: revert or noOp.

Integrators would benefit from the noOp but they can check status beforehand i guess? Unless the action happens async (via permit for example)

TODO: Add a test reflecting this.

miguelmtzinf:
No action is performed if the collateral status remains unchanged, and no events are emitted to maintain accurate indexing.

An alternative approach would be to revert in such cases, but this would increase friction for integrations.
- Integrators can verify the collateral status beforehand using `Spoke.getUsingAsCollateral`, but
- if `setUsingAsCollateral` is executed async, another action could front-run it, causing the entire transaction to revert.

Closes #483

### DIFF
--- a/tests/unit/Spoke/Spoke.Config.t.sol
+++ b/tests/unit/Spoke/Spoke.Config.t.sol
@@ -121,21 +121,27 @@ contract SpokeConfigTest is SpokeBase {
     spoke1.setUsingAsCollateral(daiReserveId, true);
   }
 
+  /// no action taken when collateral status is unchanged
   function test_setUsingAsCollateral_collateralStatusUnchanged() public {
     uint256 daiReserveId = _daiReserveId(spoke1);
 
     // ensure DAI is allowed as collateral
     updateCollateralFlag(spoke1, daiReserveId, true);
+
+    // slight update in collateral factor so user is subject to dynamic risk config refresh
+    updateCollateralFactor(spoke1, daiReserveId, _getCollateralFactor(spoke1, daiReserveId) + 1_00);
+    // slight update liquidity premium so user is subject to risk premium refresh
+    updateLiquidityPremium(spoke1, daiReserveId, _getLiquidityPremium(spoke1, daiReserveId) + 1_00);
+
     // Bob not using DAI as collateral
     assertFalse(spoke1.getUsingAsCollateral(daiReserveId, bob), 'bob not using as collateral');
-
-    vm.startPrank(bob);
 
     // No action taken, because collateral status is already false
     DynamicConfig[] memory bobDynConfig = _getUserDynConfigKeys(spoke1, bob);
     uint256 bobRp = _getUserRpStored(spoke1, daiReserveId, bob);
 
     vm.recordLogs();
+    vm.prank(bob);
     spoke1.setUsingAsCollateral(daiReserveId, false);
     _assertEventNotEmitted(ISpoke.UsingAsCollateral.selector);
 
@@ -144,22 +150,27 @@ contract SpokeConfigTest is SpokeBase {
     assertEq(_getUserDynConfigKeys(spoke1, bob), bobDynConfig);
 
     // Bob can change dai collateral status to true
+    vm.prank(bob);
     spoke1.setUsingAsCollateral(daiReserveId, true);
     assertTrue(spoke1.getUsingAsCollateral(daiReserveId, bob), 'bob using as collateral');
+
+    // slight update in collateral factor so user is subject to dynamic risk config refresh
+    updateCollateralFactor(spoke1, daiReserveId, _getCollateralFactor(spoke1, daiReserveId) + 1_00);
+    // slight update liquidity premium so user is subject to risk premium refresh
+    updateLiquidityPremium(spoke1, daiReserveId, _getLiquidityPremium(spoke1, daiReserveId) + 1_00);
 
     // No action taken, because collateral status is already true
     bobDynConfig = _getUserDynConfigKeys(spoke1, bob);
     bobRp = _getUserRpStored(spoke1, daiReserveId, bob);
 
     vm.recordLogs();
+    vm.prank(bob);
     spoke1.setUsingAsCollateral(daiReserveId, true);
     _assertEventNotEmitted(ISpoke.UsingAsCollateral.selector);
 
     assertTrue(spoke1.getUsingAsCollateral(daiReserveId, bob));
     assertEq(_getUserRpStored(spoke1, daiReserveId, bob), bobRp);
     assertEq(_getUserDynConfigKeys(spoke1, bob), bobDynConfig);
-
-    vm.stopPrank();
   }
 
   function test_setUsingAsCollateral() public {


tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -165,7 +167,11 @@ contract SpokeConfigTest is SpokeBase {
     spoke1.setUsingAsCollateral(daiReserveId, usingAsCollateral);
 
     DataTypes.UserPosition memory userData = spoke1.getUserPosition(daiReserveId, bob);
-    assertEq(spoke1.getUsingAsCollateral(daiReserveId, bob), usingAsCollateral, 'wrong usingAsCollateral');
+    assertEq(
+      spoke1.getUsingAsCollateral(daiReserveId, bob),
+      usingAsCollateral,
+      'wrong usingAsCollateral'
+    );
```
avniculae, line 200: I think it would be good to explicitly test this 2 branches as well, wdyt?
```
if (usingAsCollateral) {
        _refreshDynamicConfig(msg.sender);
      } else {
        uint256 newUserRiskPremium = _refreshAndValidateUserPosition(msg.sender); // validates HF
        _notifyRiskPremiumUpdate(type(uint256).max, msg.sender, newUserRiskPremium);
      }
```

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -131,16 +131,18 @@ contract SpokeConfigTest is SpokeBase {
...
 
     // Bob can change dai collateral status to true
     spoke1.setUsingAsCollateral(daiReserveId, true);
     assertTrue(spoke1.getUsingAsCollateral(daiReserveId, bob), 'bob using as collateral');
 
-    // Bob can't change dai collateral status to true, because already true
-    vm.expectRevert(ISpoke.CollateralStatusUnchanged.selector);
+    // No action taken, because collateral status is already true
+    vm.expectEmit(address(spoke1), uint64(0));
+    emit ISpoke.UsingAsCollateral(daiReserveId, bob, true);
```
avniculae: would it be an overkill to make sure that dynamic risk config and rp do not change (even if outdated)?

tests/unit/Spoke/Spoke.Config.t.sol
miguelmtzinf: there are tests for this in `Spoke.DynamicConfig.Triggers.t.sol`

tests/unit/Spoke/Spoke.Config.t.sol
miguelmtzinf: Added it. Removed the expect emit with 0 count, which was not working

tests/unit/Spoke/Spoke.Config.t.sol
avniculae: I see there's `test_usingAsCollateral_triggers_dynamicConfigUpdate`, which focuses on the `if` branch, but it also checks that `setUsingAsCollateral` reverts due to HF. however, it seems that it does not check the happy path of the `else` statement

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -132,18 +132,29 @@ contract SpokeConfigTest is SpokeBase {
...
-    vm.expectEmit(address(spoke1), uint64(0));
-    emit ISpoke.UsingAsCollateral(daiReserveId, bob, false);
+    DynamicConfig[] memory bobDynConfig = _getUserDynConfigKeys(spoke1, bob);
+    uint256 bobRp = _getUserRpStored(spoke1, daiReserveId, bob);
+
     spoke1.setUsingAsCollateral(daiReserveId, false);
 
+    assertFalse(spoke1.getUsingAsCollateral(daiReserveId, bob));
+    assertEq(_getUserRpStored(spoke1, daiReserveId, bob), bobRp);
+    assertEq(_getUserDynConfigKeys(spoke1, bob), bobDynConfig);
```
avniculae, line 150: shouldn't we add a newer dyn config key, and also change some collateral factor before `setUsingAsCollateral`? this way we can truly check that they stay unchanged, wdyt?

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -132,18 +132,29 @@ contract SpokeConfigTest is SpokeBase {
     vm.startPrank(bob);
 
     // No action taken, because collateral status is already false
-    vm.expectEmit(address(spoke1), uint64(0));
-    emit ISpoke.UsingAsCollateral(daiReserveId, bob, false);
+    DynamicConfig[] memory bobDynConfig = _getUserDynConfigKeys(spoke1, bob);
```
avniculae, line 140: I think we can recordLogs and assert they are empty / do not include ISpoke.UsingAsCollateral, wdyt?

src/contracts/Spoke.sol
```diff
@@ -389,7 +389,25 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
 
   /// @inheritdoc ISpoke
   function setUsingAsCollateral(uint256 reserveId, bool usingAsCollateral) external {
-    _setUsingAsCollateral(reserveId, msg.sender, usingAsCollateral);
+    DataTypes.PositionStatus storage positionStatus = _positionStatus[msg.sender];
+
+    // process only if collateral status changes
+    if (positionStatus.isUsingAsCollateral(reserveId) != usingAsCollateral) {
```
```suggestion
    if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) return;
```
prefer early return is better for cc but nbd

tests/unit/Spoke/Spoke.Config.t.sol
DhairyaSethi: else path is in `Spoke.Config.t.sol` haha

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -695,6 +695,12 @@ contract SpokeBase is Base {
     address user
   ) internal view returns (uint256) {
     DataTypes.UserPosition memory pos = spoke.getUserPosition(reserveId, user);
+    // sanity check
+    assertTrue(
+      pos.baseDrawnShares > 0 || pos.premiumDrawnShares == 0,
+      'if base is zero, premium must be zero'
+    );
+    if (pos.baseDrawnShares == 0) return 0;
     return pos.premiumDrawnShares.percentDiv(pos.baseDrawnShares);
```
DhairyaSethi: this would need explicit rounding dir as well since we've made `premium = base.percentMulUp(rp)`
looks like it was overlooked and could be causing the relative assertion precision issues

miguelmtzinf: done. They weren't involved in any relative assert, so all good.

tests/unit/Spoke/Spoke.Config.t.sol
DhairyaSethi: that test *does* check the else path, last section of the test is disabling as collateral 

https://github.com/aave/aave-v4/blob/edf026aa55ef8bc07542b2dcc2934646ecb883b0/tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol#L158

tests/unit/Spoke/Spoke.Config.t.sol
miguelmtzinf: added

tests/unit/Spoke/Spoke.Config.t.sol
avniculae: that test goes into else section, but it does not assert the new rp right?

avniculae: > else path is in Spoke.Config.t.sol haha

where?

PR #492, opened 2025-06-30
fix: only refresh relevant reserve when enabling as collateral

### LINKED ISSUE #493: DynamicRiskConfig: refresh on enabling as collateral
While enabling a reserve as collateral currently we refresh the dynamic risk configuration of all existing collateral reserves in the user position. But this is unfair to the user given it's a positive action on their position and with this feature our aim is to update the hooked configuration when the user increases risk/performs a negative action on their position; namely withdraw, borrow, or disabling as collateral.

DhairyaSethi:
closes #493

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -402,8 +402,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     positionStatus.setUsingAsCollateral(reserveId, usingAsCollateral);
 
     if (usingAsCollateral) {
-      _userPositions[msg.sender][reserveId].configKey = reserve.dynamicConfigKey;
-      emit UserDynamicConfigRefreshed(msg.sender, reserveId);
+      _refreshDynamicConfig(msg.sender, reserveId);
     } else {
       // If unsetting, check HF and update user rp
       uint256 newUserRiskPremium = _refreshAndValidateUserPosition(msg.sender); // validates HF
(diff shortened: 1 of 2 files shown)


src/interfaces/ISpoke.sol
```diff
@@ -21,6 +21,7 @@ interface ISpoke is IMulticall, IAccessManaged {
     DataTypes.DynamicReserveConfig config
   );
   event UserDynamicConfigRefreshed(address indexed user);
+  event UserDynamicConfigRefreshed(address indexed user, uint256 reserveId);
```
yan-man, line 24: should we index on reserveId? 

should we put reserveId first like in other actions?

DhairyaSethi: sure x2
it's slightly different here; the first event signals user dyn config refreshed for all their reserves, and the second for only a specific reserve; open to suggestions on an alternate approach here as well

yan-man: ah missed that it's another event. Perhaps we can rename the first one so it signals all reserves are refreshed, also so that it won't be as confusing during indexing

DhairyaSethi: i think of it as an overloaded event so it makes sense for the same name. any name ideas for your suggestion tho?

yan-man: `UserDynamicConfigsRefreshed` ?

CheyenneAtapour: An argument to leave the name the same for both events is if they are only dealing with 1 dynamic config to refresh, then these the two events should refer to exactly the same thing, and it's also not plural. So I'm fine leaving as is

yan-man: good point, also don't feel too strongly

src/contracts/Spoke.sol
```diff
@@ -1318,7 +1318,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     positionStatus.setUsingAsCollateral(reserveId, usingAsCollateral);
 
     if (usingAsCollateral) {
-      _refreshDynamicConfig(user);
+      userPosition.configKey = _reserves[reserveId].dynamicConfigKey;
+      emit UserDynamicConfigRefreshed(user, reserveId);
```
miguelmtzinf: I think better to create an internal method for this, so we have 2 flavors of `_refreshDynamicConfig`: for all, and for only 1.

src/interfaces/ISpoke.sol
miguelmtzinf, line 24: Feels inconsistent we emit 1 for all reserves, and 1 for 1 specific reserve...
Emitting 1 event per reserve is overkill, so thinking that we can just emit `UserDynamicConfigRefreshed` for all and indexers can check if this comes together a `setUsingAsCollateral` event?

src/contracts/Spoke.sol
DhairyaSethi: yeah this is only used once here but sure

src/interfaces/ISpoke.sol
DhairyaSethi: feels risky & complex for indexers considering logic needed to handle multicall txns which could include both variants of the events 

also one event meaning two things means we should align on the Add/AccrueFee debate as well haha

yan-man: :( already started https://github.com/aave/aave-v4/pull/506

PR #496, opened 2025-06-30
feat: collateral count, opt calcUserAccountData

DhairyaSethi:
- Introduce solady's LibBit as a dependency.
- Implement collateralCount leveraging solady's optimised `popCount`.
- Reduce number of loops in calculateUserAccountData from 3 to 2.
- Gas saving are more significant with a more realistic number of reserves for the gas snapshots (current reserve count is 4) as it scales linearly.

Snapshot diff for 100 reserves listed
```
  "borrow": "1146431", -> "1068089"
  "liquidationCall: full": "1198891" -> "1042206"
  "liquidationCall: partial": "1380162" -> "1223477"
  "repay: full": "851067" -> "772727"
  "repay: partial": "923738" -> "845395"
  "supply: 0 debt, collateralDisabled": "174154" -> "174154"
  "supply: 0 debt, collateralEnabled": "174130" -> "174130"
  "supply: 1 debt": "174178" -> "174178"
  "supply: 2 debt": "174189" -> "174189"
  "supply: 3 debt": "174142" -> "174142"
  "updateUserRiskPremium": "771772" -> "693430"
  "updateUserRiskPremium: 2 debts": "821459" -> "743117"
  "usingAsCollateral": "118732" -> "118732"
  "withdraw: full": "885442" -> "807103"
  "withdraw: partial": "896514" -> "818175"
```

### DIFF
--- a/src/libraries/configuration/PositionStatus.sol
+++ b/src/libraries/configuration/PositionStatus.sol
@@ -70,7 +70,7 @@ library PositionStatus {
     uint256 reserveId
   ) internal view returns (bool) {
     unchecked {
-      return (self.getMapSlot(reserveId) >> ((reserveId % 128) << 1)) & 3 != 0;
+      return (self.getBucketWord(reserveId) >> ((reserveId % 128) << 1)) & 3 != 0;
     }
   }
 
@@ -85,7 +85,7 @@ library PositionStatus {
     uint256 reserveId
   ) internal view returns (bool) {
     unchecked {
-      return (self.getMapSlot(reserveId) >> ((reserveId % 128) << 1)) & 1 != 0;
+      return (self.getBucketWord(reserveId) >> ((reserveId % 128) << 1)) & 1 != 0;
     }
   }
 
@@ -100,7 +100,7 @@ library PositionStatus {
     uint256 reserveId
   ) internal view returns (bool) {
     unchecked {
-      return (self.getMapSlot(reserveId) >> (((reserveId % 128) << 1) + 1)) & 1 != 0;
+      return (self.getBucketWord(reserveId) >> (((reserveId % 128) << 1) + 1)) & 1 != 0;
     }
   }
 
@@ -127,11 +127,11 @@ library PositionStatus {
   }
 
   /**
-   * @dev Returns the slot containing the reserve state in the bitmap.
+   * @dev Returns the word containing the reserve state in the bitmap.
    * @param self The configuration object.
-   * @return The slot containing the state of the reserve.
+   * @return The word containing the state of the reserve.
    */
-  function getMapSlot(
+  function getBucketWord(
     DataTypes.PositionStatus storage self,
     uint256 reserveId
   ) internal view returns (uint256) {
--- a/tests/unit/libraries/PositionStatus.t.sol
+++ b/tests/unit/libraries/PositionStatus.t.sol
@@ -207,6 +207,41 @@ contract PositionStatusTest is Base {
     assertEq(p.collateralCount(reserveCount), collateralCount);
   }
 
+  function test_setters_use_correct_slot(uint256 a) public {
+    uint256 bucket = a / 128;
+    bytes32 slot = keccak256(abi.encode(bucket, p.slot()));
+
+    vm.record();
+    p.setUsingAsCollateral(a, vm.randomBool());
+    (bytes32[] memory reads, bytes32[] memory writes) = vm.accesses(address(p));
+    assertEq(writes.length, 1);
+    assertEq(reads.length, 2);
+
+    assertEq(writes[0], slot);
+    assertEq(reads[0], slot);
+    assertEq(reads[1], slot);
+
+    vm.record();
+    p.setBorrowing(a, vm.randomBool());
+    (reads, writes) = vm.accesses(address(p));
+    assertEq(writes.length, 1);
+    assertEq(reads.length, 2);
+
+    assertEq(writes[0], slot);
+    assertEq(reads[0], slot);
+    assertEq(reads[1], slot);
+  }
+
+  function test_getBucketWord(uint256 a) public {
+    uint256 bucket = a / 128;
+    vm.record();
+    p.getBucketWord(a);
+    (bytes32[] memory reads, bytes32[] memory writes) = vm.accesses(address(p));
+    assertEq(writes.length, 0);
+    assertEq(reads.length, 1);
+    assertEq(reads[0], keccak256(abi.encode(bucket, p.slot())));
+  }
+
   function test_popCount(bytes32) public {
     uint256 bits = vm.randomUint();
     assertEq(LibBit.popCount(bits), _popCountNaive(bits));
--- a/tests/mocks/PositionStatusWrapper.sol
+++ b/tests/mocks/PositionStatusWrapper.sol
@@ -40,7 +40,13 @@ contract PositionStatusWrapper {
     return _p.collateralCount(reserveCount);
   }
 
-  function getMapSlot(uint256 reserveId) external view returns (uint256) {
-    return _p.getMapSlot(reserveId);
+  function getBucketWord(uint256 reserveId) external view returns (uint256) {
+    return _p.getBucketWord(reserveId);
+  }
+
+  function slot() external pure returns (bytes32 s) {
+    assembly ('memory-safe') {
+      s := _p.slot
+    }
   }
 }


src/contracts/Spoke.sol
```diff
@@ -826,6 +799,16 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
         }
       }
 
+      if (positionStatus.isBorrowing(vars.reserveId)) {
+        vars.totalDebtInBaseCurrency += _getUserDebtInBaseCurrency(
+          userPosition,
+          vars.assetId,
```
yan-man, line 839: can rm L779 as `assetId` is already defined in L766 

also for `assetPrice` on L781

src/contracts/Spoke.sol
```diff
@@ -849,7 +832,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     vars.i = 0;
     // @dev from this point onwards, `collateralCounterInBaseCurrency` represents running collateral
     // value used in risk premium, `debtCounterInBaseCurrency` represents running outstanding debt
-    while (vars.i < vars.collateralReserveCount && vars.debtCounterInBaseCurrency > 0) {
+    while (vars.i < list.length() && vars.debtCounterInBaseCurrency > 0) {
```
yan-man, line 868: should we cache `list.length()` into a var

src/libraries/configuration/PositionStatus.sol
```diff
@@ -112,6 +114,27 @@ library PositionStatus {
     }
   }
 
+  /**
+   * @dev Counts the number reserves enabled as collateral.
```
yan-man:
```suggestion
   * @dev Counts the number of reserves enabled as collateral.
```

tests/unit/libraries/PositionStatus.t.sol
yan-man, line 1: I think need a test for `InvalidReserveCount`?

src/contracts/Spoke.sol
DhairyaSethi: the list is in the memory so it's the same cost if we cache in localVars obj, only slightly helps if we break consistency and cache on the stack here

tests/unit/libraries/PositionStatus.t.sol
DhairyaSethi: theoretically should never hit, it's a sanity assert. will probably add for dirty bits as well & assert both

src/libraries/configuration/PositionStatus.sol
```diff
@@ -112,6 +114,27 @@ library PositionStatus {
...
+   * @dev Takes all bits in the last bucket after `reserveCount` as well.
```
CheyenneAtapour: Can we be clearer here about what is meant by "take"? Does it count them?

DhairyaSethi: yes

CheyenneAtapour: Then perhaps just say "Counts all bits in the last bucket after `reserveCount` as well."

src/libraries/configuration/PositionStatus.sol
```diff
@@ -112,6 +114,27 @@ library PositionStatus {
...
+   */
+  function collateralCount(
+    DataTypes.PositionStatus storage self,
+    uint256 reserveCount
+  ) internal view returns (uint256) {
+    require(reserveCount <= MAX_RESERVES_COUNT, InvalidReserveCount()); // assertion: should never hit
+    unchecked {
+      uint256 count;
+      uint256 bucket = 1 + (reserveCount >> 7);
+      while (bucket-- != 0) {
```
miguelmtzinf: this is not the only loop you wrote that count down instead of incrementing, it's just personal preference?
since there are some incrementing, i think it's better to remain consistent and do increasing iteration everywhere

src/dependencies/solady/LibBit.sol
```diff
@@ -0,0 +1,325 @@
...
+        byte(
+          and(div(0xd76453e0, shr(r, x)), 0x1f),
+          0x001f0d1e100c1d070f090b19131c1706010e11080a1a141802121b1503160405
+        )
+      )
+    }
+  }
+
+  /// @dev Returns the number of set bits in `x`.
+  function popCount(uint256 x) internal pure returns (uint256 c) {
```
miguelmtzinf, line 11: I do think we should just keep this function and remove the rest (giving proper reference ofc)

DhairyaSethi: yeah, we're not doing the same with OZ tho & hence didn't do it for consistency

src/libraries/configuration/PositionStatus.sol
DhairyaSethi: i didn't to avoid maintaining two stack vars, one incrementing from 0 & the other holding the endBucket, this is ever so slightly cheaper

src/dependencies/solady/LibBit.sol
miguelmtzinf: hmmm good point, i think OZ is kinda different, but no strong opinion

src/libraries/configuration/PositionStatus.sol
```diff
@@ -112,6 +113,30 @@ library PositionStatus {
...
+   * @dev Panics if `reserveCount` is greater than MAX_RESERVES_COUNT, & disregards
+   * potential dirty bits set after `reserveCount`.
+   * @param self The configuration object.
+   * @param reserveCount The current reserveCount, to avoid reading uninitialized buckets.
+   */
+  function collateralCount(
+    DataTypes.PositionStatus storage self,
+    uint256 reserveCount
+  ) internal view returns (uint256) {
+    assert(reserveCount <= MAX_RESERVES_COUNT); // should never hit
```
miguelmtzinf: why do we need this if we are only taking `reserveCount` into account

DhairyaSethi: sanity 
as discussed in the sync today, all of these checks are removed

tests/unit/libraries/PositionStatus.t.sol
```diff
@@ -1,142 +1,199 @@
...
-import 'tests/Base.t.sol';
+import {LibBit} from 'src/dependencies/solady/LibBit.sol';
+import {Test} from 'forge-std/Test.sol';
 
-contract PositionStatusTest is Base {
-  using PositionStatus for DataTypes.PositionStatus;
+import {PositionStatusWrapper, PositionStatus} from 'tests/mocks/PositionStatusWrapper.sol';
 
-  DataTypes.PositionStatus internal positionStatus;
+contract PositionStatusTest is Test {
```
miguelmtzinf: we kinda decided to use always Base, override setup and move all imports

tests/unit/libraries/PositionStatus.t.sol
```diff
@@ -1,142 +1,199 @@
...
+    p.setBorrowing(127, true);
+    assertEq(p.isBorrowing(127), true);
 
-    positionStatus.setBorrowing(127, false);
-    assertEq(positionStatus.isBorrowing(127), false);
+    p.setBorrowing(127, false);
+    assertEq(p.isBorrowing(127), false);
   }
 
   function test_setBorrowing_slot1() public {
```
miguelmtzinf, line 40: do we wanna check this is actually the second slot?

tests/unit/libraries/PositionStatus.t.sol
```diff
@@ -1,142 +1,199 @@
...
+    p.setUsingAsCollateral(0, true);
+    p.setBorrowing(0, true);
 
-    assertEq(positionStatus.isUsingAsCollateralOrBorrowing(0), true);
+    assertEq(p.isUsingAsCollateralOrBorrowing(0), true);
 
-    positionStatus.setUsingAsCollateral(0, false);
-    positionStatus.setBorrowing(0, false);
+    p.setUsingAsCollateral(0, false);
+    p.setBorrowing(0, false);
```
miguelmtzinf, line 111:
```suggestion
    p.setUsingAsCollateral(0, false);
    assertEq(p.isUsingAsCollateralOrBorrowing(0), true);

    p.setBorrowing(0, false);
```

can we do it in 2 steps, for completeness

tests/unit/libraries/PositionStatus.t.sol
```diff
@@ -1,142 +1,199 @@
...
+    p.setUsingAsCollateral(255, false);
+    assertEq(p.isUsingAsCollateral(255), false);
+  }
+
+  function test_collateralCount() public {
+    p.setUsingAsCollateral(127, true);
+    assertEq(p.collateralCount(128), 1);
+
+    // ignore invalid bits
+    assertEq(p.collateralCount(100), 0);
```
miguelmtzinf, line 151: can we have a clear example of this with more than 1 slot? dedicated test case may make sense

tests/unit/libraries/PositionStatus.t.sol
```diff
@@ -1,142 +1,199 @@
...
+
+  function test_collateralCount(uint256 reserveCount) public {
+    reserveCount = bound(reserveCount, 0, 1 << 10); // gas limit
+    vm.setArbitraryStorage(address(p));
+
+    uint256 collateralCount;
+    for (uint256 reserveId; reserveId < reserveCount; ++reserveId) {
+      if (p.isUsingAsCollateral(reserveId)) ++collateralCount;
+      // reserveId is 0-base indexed
+      assertEq(p.collateralCount({reserveCount: reserveId + 1}), collateralCount);
```
miguelmtzinf: what does this mean sorry

DhairyaSethi: input to `collateralCount` is reserveCount, and here we're checking on each iteration of the loop with the current counter, at any step the reserveCount until then is reserveId + 1 bc reserveId iterate from 0 - n so with 1 reserveCount we have the first reserve id as 0 & so on

tests/unit/libraries/PositionStatus.t.sol
```diff
@@ -1,142 +1,199 @@
...
+    p.setUsingAsCollateral(255, true);
+    assertEq(p.isUsingAsCollateral(255), true);
+
+    p.setUsingAsCollateral(255, false);
+    assertEq(p.isUsingAsCollateral(255), false);
+  }
+
+  function test_collateralCount() public {
+    p.setUsingAsCollateral(127, true);
+    assertEq(p.collateralCount(128), 1);
```
miguelmtzinf, line 144: can we add the limit case please?
slot goes from 0 to 127, so the following should pass right:
```
    p.setUsingAsCollateral(128, true);
    assertEq(p.collateralCount(128), 0);
```

tests/unit/libraries/PositionStatus.t.sol
miguelmtzinf: i see what you are saying, this is a subset of L186 on every iteration, with a tighter scope. can you elaborate a bit more in the comment pls?

tests/unit/libraries/PositionStatus.t.sol
DhairyaSethi: think that's overkill but lmk if you had something else in mind
[1d07850](https://github.com/aave/aave-v4/pull/496/commits/1d07850bb19021130e5acff76d544dc9cfc3eb03)

tests/unit/libraries/PositionStatus.t.sol
DhairyaSethi: added lmk [c4c3cea](https://github.com/aave/aave-v4/pull/496/commits/c4c3cea413dd135607363fc47be9b1c1cb14c7aa)

tests/unit/libraries/PositionStatus.t.sol
DhairyaSethi: lmk [c4c3cea](https://github.com/aave/aave-v4/pull/496/commits/c4c3cea413dd135607363fc47be9b1c1cb14c7aa)

src/libraries/configuration/PositionStatus.sol
```diff
@@ -127,11 +127,11 @@ library PositionStatus {
...
 
   /**
-   * @dev Returns the slot containing the reserve state in the bitmap.
+   * @dev Returns the word containing the reserve state in the bitmap.
    * @param self The configuration object.
-   * @return The slot containing the state of the reserve.
+   * @return The word containing the state of the reserve.
    */
-  function getMapSlot(
+  function getBucketWord(
```
miguelmtzinf, line 134: i actually like more `getMapSlot` or `getMapWord` personally, but nbd

DhairyaSethi: used bucket bc i refer it it like that above as well, can rename the var to `bucket` as well (renaming the helpers in collateralCount as map doesn't make sense to me lol)

PR #507, opened 2025-07-04
Move Liquidation Bonus to Dynamic Reserve Config

### LINKED ISSUE #459: Move LiquidationBonus to DynamicConfig
In the current implementation we can end up with a collateral factor incompatible with the liquidationBonus for a user position.

CheyenneAtapour:
closes #459

### DIFF
--- a/tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
@@ -105,7 +105,7 @@ contract SpokeLiquidationBase is SpokeBase {
     LiquidationTestLocalParams memory state;
     state.collateralReserve = spoke1.getReserve(collateralReserveId);
     state.debtReserve = spoke1.getReserve(debtReserveId);
-    state.collDynConfig = spoke1.getDynamicReserveConfig(collateralReserveId);
+    state.collDynConfig = _getUserDynConfig(spoke1, alice, collateralReserveId);
 
     liqConfig = _bound(liqConfig);
     liqBonus = bound(
--- a/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.CloseFactor.MultiReserve.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.CloseFactor.MultiReserve.t.sol
@@ -188,10 +188,7 @@ contract LiquidationCallCloseFactorMultiReserveTest is SpokeLiquidationBase {
 
     for (uint256 i = 0; i < collateralReserveIds.length; i++) {
       state.collateralReserves[i] = spoke1.getReserve(collateralReserveIds[i]);
-      state.collDynConfigs[i] = spoke1.getDynamicReserveConfig(
-        collateralReserveIds[i],
-        state.collateralReserves[i].dynamicConfigKey
-      ); // utilize latest dynamic config
+      state.collDynConfigs[i] = _getUserDynConfig(spoke1, alice, collateralReserveIds[i]); // utilize user's dynamic config
     }
     for (uint256 i = 0; i < debtReserveIds.length; i++) {
       state.debtReserves[i] = spoke1.getReserve(debtReserveIds[i]);
--- a/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenario.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenario.t.sol
@@ -45,7 +45,7 @@ contract LiquidationCallScenarioTest is SpokeLiquidationBase {
     state.daiReserveId = _daiReserveId(spoke1);
     state.wbtcReserveId = _wbtcReserveId(spoke1);
 
-    state.liqBonus = spoke1.getDynamicReserveConfig(state.wbtcReserveId).liquidationBonus;
+    state.liqBonus = _getUserDynConfig(spoke1, alice, state.wbtcReserveId).liquidationBonus;
 
     // collateral: wbtc/dai
     state.collAmount.wbtc = 1 * 10 ** decimals.wbtc; // $50k wbtc
@@ -143,7 +143,7 @@ contract LiquidationCallScenarioTest is SpokeLiquidationBase {
     state.daiReserveId = _daiReserveId(spoke1);
     state.wbtcReserveId = _wbtcReserveId(spoke1);
 
-    state.liqBonus = spoke1.getDynamicReserveConfig(state.wbtcReserveId).liquidationBonus;
+    state.liqBonus = _getUserDynConfig(spoke1, alice, state.wbtcReserveId).liquidationBonus;
 
     // collateral: wbtc/dai
     state.collAmount.wbtc = 1 * 10 ** decimals.wbtc; // $50k wbtc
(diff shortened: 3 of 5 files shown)


src/contracts/Spoke.sol
```diff
@@ -501,7 +502,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     return
       _liquidationConfig.calculateVariableLiquidationBonus(
         healthFactor,
-        _reserves[reserveId].config.liquidationBonus,
+        _dynamicConfig[reserveId][_reserves[reserveId].dynamicConfigKey].liquidationBonus,
```
DhairyaSethi: we don't want to use the latest config key on liquidations but the users last stored key

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -234,19 +229,38 @@ contract SpokeConfigTest is SpokeBase {
 
   function test_updateReserveConfig_fuzz_revertsWith_ReserveNotListed(
     uint256 reserveId,
-    uint256 liquidationBonus
+    uint256 liquidityPremium
```
yan-man: actually can we just rm this input altogether

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -265,22 +279,41 @@ contract SpokeConfigTest is SpokeBase {
     spoke1.updateDynamicReserveConfig(daiReserveId, config);
   }
 
-  function test_updateReserveConfig_revertsWith_InvalidLiquidationBonus() public {
-    uint256 liquidationBonus = PercentageMath.PERCENTAGE_FACTOR + 1;
-    test_updateReserveConfig_fuzz_revertsWith_InvalidLiquidationBonus(liquidationBonus);
+  function test_updateDynamicReserveConfig_revertsWith_InvalidCollateralFactorAndLiquidationBonus()
```
yan-man: can this also just call the `test_updateDynamicReserveConfig_fuzz_revertsWith_InvalidCollateralFactorAndLiquidationBonus`

tests/unit/Spoke/Spoke.Config.t.sol
CheyenneAtapour: True, we can just throw a config of all 0's in there too, I think that's just simpler - done

tests/unit/Spoke/Spoke.Config.t.sol
CheyenneAtapour: Yep, done

src/contracts/Spoke.sol
CheyenneAtapour: Done, added a user param

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -108,7 +108,11 @@ contract SpokeLiquidationBase is SpokeBase {
     state.collDynConfig = spoke1.getDynamicReserveConfig(collateralReserveId);
 
     liqConfig = _bound(liqConfig);
-    liqBonus = bound(liqBonus, MIN_LIQUIDATION_BONUS, MAX_LIQUIDATION_BONUS);
+    liqBonus = bound(
+      liqBonus,
+      MIN_LIQUIDATION_BONUS,
+      PercentageMathExtended.PERCENTAGE_FACTOR.percentDivDown(state.collDynConfig.collateralFactor)
```
yan-man, line 114: do tests fail without changing this upper bound?

yan-man: discussed today, it fails due to initially updating the LB in the test set up but not in the assertions part (this is what I was wondering)

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -165,10 +169,17 @@ contract SpokeLiquidationBase is SpokeBase {
...
+    state.liquidationBonus = spoke1.getVariableLiquidationBonus(
+      collateralReserveId,
+      alice,
+      hfAfterBorrow
+    );
 
     state = _getAccountingInfoBeforeLiq(state);
 
+    // Get user's dynamic config key before liquidation
+    uint16 configKeyBefore = spoke1.getUserPosition(collateralReserveId, alice).configKey;
```
DhairyaSethi:
```suggestion
    // Get user's dynamic config key before liquidation
    DynamicConfig[] memory configKeysBefore = _getUserDynConfigKeys(spoke1, alice);
```
we need to validate all config keys didn't change 
another way could be assert the update event was not emitted in the normal triggers test actually

src/contracts/Spoke.sol
```diff
@@ -648,6 +646,15 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
       InvalidCollateralFactor()
     ); // max 100.00%
+    require(
+      config.liquidationBonus >= PercentageMathExtended.PERCENTAGE_FACTOR,
+      InvalidLiquidationBonus()
+    ); // min 100.00%
+    require(
+      config.collateralFactor.percentMulUp(config.liquidationBonus) <=
+        PercentageMathExtended.PERCENTAGE_FACTOR,
+      InvalidCollateralFactorAndLiquidationBonus()
```
DhairyaSethi: can we rename this error to something more apt?

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
DhairyaSethi: #526

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenario.t.sol
```diff
@@ -45,7 +45,7 @@ contract LiquidationCallScenarioTest is SpokeLiquidationBase {
     state.daiReserveId = _daiReserveId(spoke1);
     state.wbtcReserveId = _wbtcReserveId(spoke1);
 
-    state.liqBonus = spoke1.getReserve(state.wbtcReserveId).config.liquidationBonus;
+    state.liqBonus = spoke1.getDynamicReserveConfig(state.wbtcReserveId).liquidationBonus;
```
DhairyaSethi: this call gets the latest get dynamic config, which is wrong bc we need to get the liq bonus at the user's last config key

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.CloseFactor.MultiReserve.t.sol
```diff
@@ -188,10 +188,7 @@ contract LiquidationCallCloseFactorMultiReserveTest is SpokeLiquidationBase {
 
     for (uint256 i = 0; i < collateralReserveIds.length; i++) {
       state.collateralReserves[i] = spoke1.getReserve(collateralReserveIds[i]);
-      state.collDynConfigs[i] = spoke1.getDynamicReserveConfig(
-        collateralReserveIds[i],
-        state.collateralReserves[i].dynamicConfigKey
-      ); // utilize latest dynamic config
+      state.collDynConfigs[i] = _getUserDynConfig(spoke1, alice, collateralReserveIds[i]); // utilize user's dynamic config
```
yan-man, line 191: nit: can rm this comment

PR #509, opened 2025-07-04
Refactor premium restore logic

### LINKED ISSUE #505: Premium Debt Restore
Currently, the premium debt is restored when the spoke `repay` function calls `hub.refreshPremiumDebt`
https://github.com/aave/aave-v4/blob/2f9767e0d61d22d310c0256028a6aed6bea17e94/src/contracts/Spoke.sol#L307-L315

Hence, hub's `refreshPremiumDebt` function has 3 roles: realise premium, refresh premium according to new RP and restore premium. Out of these 3 features, the restore premium is the only one which changes the premium debt (if we ignore the 2 wei increase due to rounding). It might be less error prone if the `refreshPremiumDebt` only focuses on the first 2 roles (realise & refresh premium), and make the hub responsible for restoring the premium debt during `restore`. This is relevant around the update to the borrow rate: we need to make sure we update it once all values were updated

avniculae:
Closes #505 
- it refactors the premium logic to restore the premium debt in `hub.restore` instead of `hub.refreshPremiumDebt`. `hub.refreshPremiumDebt` has only 2 functionalities now: realise premium debt and refresh premium according to new RP

### DIFF
--- a/tests/unit/SpokeConfigurator.t.sol
+++ b/tests/unit/SpokeConfigurator.t.sol
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.10;
 
-import 'tests/Base.t.sol';
+import 'tests/unit/Spoke/SpokeBase.t.sol';
 
-contract SpokeConfiguratorTest is Base {
+contract SpokeConfiguratorTest is SpokeBase {
   SpokeConfigurator public spokeConfigurator;
   address public SPOKE_CONFIGURATOR_ADMIN = makeAddr('SPOKE_CONFIGURATOR_ADMIN');
 
@@ -14,7 +14,6 @@ contract SpokeConfiguratorTest is Base {
 
   function setUp() public virtual override {
     super.setUp();
-    initEnvironment();
 
     spokeConfigurator = new SpokeConfigurator(SPOKE_CONFIGURATOR_ADMIN);
     spokeAddr = address(spoke1);
@@ -362,7 +361,7 @@ contract SpokeConfiguratorTest is Base {
       .getDynamicReserveConfig(reserveId);
     expectedDynamicReserveConfig.collateralFactor = newCollateralFactor;
 
-    uint16 expectedConfigKey = _nextDynamicConfigKey(spoke1);
+    uint16 expectedConfigKey = _nextDynamicConfigKey(spoke1, reserveId);
 
     vm.expectCall(
       spokeAddr,
@@ -394,7 +393,7 @@ contract SpokeConfiguratorTest is Base {
       .getDynamicReserveConfig(reserveId);
     expectedDynamicReserveConfig.liquidationBonus = newLiquidationBonus;
 
-    uint16 expectedConfigKey = _nextDynamicConfigKey(spoke1);
+    uint16 expectedConfigKey = _nextDynamicConfigKey(spoke1, reserveId);
 
     vm.expectCall(
       spokeAddr,
@@ -425,7 +424,7 @@ contract SpokeConfiguratorTest is Base {
       .getDynamicReserveConfig(reserveId);
     expectedDynamicReserveConfig.liquidationFee = newLiquidationFee;
 
-    uint16 expectedConfigKey = _nextDynamicConfigKey(spoke1);
+    uint16 expectedConfigKey = _nextDynamicConfigKey(spoke1, reserveId);
 
     vm.expectCall(
       spokeAddr,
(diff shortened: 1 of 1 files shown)


src/contracts/LiquidityHub.sol
```diff
@@ -304,8 +304,7 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     address spokeAddress,
     int256 premiumDrawnShareDelta,
     int256 premiumOffsetDelta,
-    uint256 realizedPremiumAdded,
-    uint256 realizedPremiumTaken
+    uint256 realizedPremium
```
miguelmtzinf, line 303: we could infer this thanks to the passed data `premiumDrawnShareDelta` and `premiumOffsetDelta` by
`premiumDrawnShareDelta.toShares - premiumOffsetDelta`. However, we incur in the 2 wei instant debt creation for spoke and hub side (not for the user, because when we update rp, we set the accruedPremium to 0 always)

src/contracts/Spoke.sol
```diff
@@ -209,7 +208,6 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
       assetId,
       int256(userPremiumDrawnShares),
       int256(userPremiumOffset),
-      0,
```
miguelmtzinf: does not this mean that we can underflow? because if all users repay
total premium debt is subject to 2 wei diff but not the user, because we manually set to 0

cc @DhairyaSethi cannot recall the reason about this

src/contracts/LiquidityHub.sol
```diff
@@ -254,8 +256,7 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     uint256 assetId,
     int256 premiumDrawnShareDelta,
     int256 premiumOffsetDelta,
-    uint256 realizedPremiumAdded,
-    uint256 realizedPremiumTaken
+    uint256 realizedPremium
```
miguelmtzinf, line 254: @avniculae pointing that we could remove `realizedPremiumAdded` and keep `realizedPremiumTaken`, meaning this PR does not have anything to do with this idea of infering accruedPremium by using `drawnShares` and `offset`.

src/contracts/Spoke.sol
DhairyaSethi: global (spoke) precision loss over time converges to the same loss as local (user) i think but not i'm not certain. need to think a bit more

src/contracts/LiquidityHub.sol
avniculae: would it work here https://github.com/aave/aave-v4/blob/4e8f553a60f20d979943d26a91d2c433b99b420d/src/contracts/Spoke.sol#L977 ? I think we'll need to make an additional refresh per reserve is we want to infer it from the deltas, wdyt?

DhairyaSethi: yeah you're right, will need an additional closePremium on notify

src/contracts/Spoke.sol
```diff
@@ -738,20 +726,18 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
     int256 premiumDrawnSharesDelta,
     int256 premiumOffsetDelta,
-    uint256 realizedPremiumAdded,
-    uint256 realizedPremiumTaken
+    uint256 realizedPremium
   ) internal {
     reserve.premiumDrawnShares = _add(reserve.premiumDrawnShares, premiumDrawnSharesDelta);
     reserve.premiumOffset = _add(reserve.premiumOffset, premiumOffsetDelta);
-    reserve.realizedPremium = reserve.realizedPremium + realizedPremiumAdded - realizedPremiumTaken;
+    reserve.realizedPremium = reserve.realizedPremium + realizedPremium;
```
CheyenneAtapour:
```suggestion
    reserve.realizedPremium += realizedPremium;
```

nbd, but this feels cleaner to me

src/contracts/Spoke.sol
miguelmtzinf: I think this PR is not changing the existing behaviour, because although we set to 0 the accrued premium of the user, spoke and asset, the inflight calc can result in 2 wei diff as well.
nothing new, but created a ticket to track this: https://github.com/aave/aave-v4/issues/527

src/contracts/Spoke.sol
avniculae: [e3b2827](https://github.com/aave/aave-v4/pull/509/commits/e3b2827cf5e641ca7d3a2c56732125262f3c7f27)

miguelmtzinf:
The current implementation does not enforce a specific order between restore and refresh, which is why it is preferred (we need to accrue first so there is enough premium debt to be repaid later). The main drawback is that premium debt is settled before the actual restore (and the resulting borrow rate update).

Issue #482, opened 2025-06-27
Analysis of maximum number of reserves in a Spoke


miguelmtzinf:
Currently, we support 1024 reserves in the Spoke, which means we use 8 slots for each user in their `PositionStatus` map.

```
uint256 public constant MAX_RESERVES_COUNT = 1024;
```

Questions:
1. Do we need to enforce a maximum number of reserves, why?
2. If yes, what's the proper value as a maximum number.

miguelmtzinf:
reference: https://github.com/aave/aave-v4/pull/407#discussion_r2164083865

miguelmtzinf:
related to https://github.com/aave/aave-v4/pull/496
we decided to remove max amount from here, and put it somewhere else... maybe in `Spoke.addReserves`?

DhairyaSethi:
not planned for now

maximum number of assetId is limited on a spoke to 65 as we pack it to uint16

PR #513, opened 2025-07-07
Spoke Configurator

### LINKED ISSUE #465: Configurator functions for Spoke
see parent issue for description #445

avniculae:
Closes #465 
- renames existing Configurator to HubConfigurator
- implements new SpokeConfigurator, with granular functions for setting each field from the `LiquidationConfig`, `ReserveConfig` and `DyanmicReserveConfig` struct
- it also contains wrapper functions for `addReserve`, `updateOracle` and `updateReservePriceSource` --> while these fns can be called directly on the spoke, it will be more manageable to give access only to the configurators. also, a general forwarder function (such as `callSpoke(bytes spokeCalldata)` would be too complex for operators
- reverts in spoke if reserve is not listed when updating dynamic risk config
- converts most fuzzed hub configurator tests to simple tests

### DIFF
--- a/tests/unit/Spoke/SpokeBase.t.sol
+++ b/tests/unit/Spoke/SpokeBase.t.sol
@@ -878,7 +878,7 @@ contract SpokeBase is Base {
   function getReserveIdByAssetId(ISpoke spoke, uint256 assetId) internal view returns (uint256) {
     for (uint256 i; i < spoke.getReserveCount(); ++i) {
       if (assetId == spoke.getReserve(i).assetId) {
-        return assetId;
+        return i;
       }
     }
     revert('not found');


src/contracts/Spoke.sol
```diff
@@ -124,8 +125,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     DataTypes.ReserveConfig calldata config
   ) external restricted {
     // TODO: AccessControl, More sophisticated
+    require(reserveId < reserveCount, ReserveNotListed());
```
CheyenneAtapour: Why this approach over the previous, just to revert earlier?

src/contracts/SpokeConfigurator.sol
```diff
@@ -0,0 +1,166 @@
...
+    uint256 reserveId,
+    uint16 collateralFactor
+  ) external onlyOwner {
+    ISpoke targetSpoke = ISpoke(spoke);
+    DataTypes.DynamicReserveConfig memory dynamicReserveConfig = targetSpoke
+      .getDynamicReserveConfig(reserveId);
+    dynamicReserveConfig.collateralFactor = collateralFactor;
+    targetSpoke.updateDynamicReserveConfig(reserveId, dynamicReserveConfig);
+  }
+}
```
CheyenneAtapour, line 186: Do we also want a general function `updateReserveConfig` to update all aspects of a reserve's config at once? (taking `ReserveConfig` as parameter). 

If so, perhaps also one for `updateDynamicReserveConfig` as well

src/interfaces/IHubConfigurator.sol
```diff
@@ -4,21 +4,21 @@ pragma solidity ^0.8.0;
...
+interface IHubConfigurator {
   /**
    * @notice Thrown when the the list of assets and spoke configs are not the same length in `addSpokeToAssets`.
    */
   error MismatchedConfigs();
 
   /**
    * @notice Registers the same spoke for multiple assets with the hub, each with their own configuration.
-   * @dev The i-th asset id in `assetIds` corresponds to the i-th configuration in `configs`.
+   * @dev The i-th asset identifier in `assetIds` corresponds to the i-th configuration in `configs`.
```
CheyenneAtapour, line 19: not really a big deal, but imo, I'd say it's alright in the `dev` comment to just say `id` instead of `identifier`. Honestly, I think it's more clear that way in this context since the variable is also called `assetIds`.

I personally read this twice to understand

miguelmtzinf:
Can we tackle #482 here? @DhairyaSethi had some insights about the theoretical max number of reserves we could have.

src/contracts/Spoke.sol
```diff
@@ -79,8 +79,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   }
 
   function addReserve(
-    uint256 assetId,
     address hub,
+    uint256 assetId,
```
miguelmtzinf, line 84: we could assert `assetId < hub.assetCount()` instead of `asset.unedrlying != addrses(0)`

src/contracts/Spoke.sol
```diff
@@ -79,8 +79,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   }
 
   function addReserve(
```
miguelmtzinf, line 82: in this fn we can impose a max reserves limit

src/interfaces/ISpokeConfigurator.sol
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * @notice Updates the liquidation close factor of a spoke.
+   * @param spoke The address of the spoke.
+   * @param closeFactor The new liquidation close factor.
+   */
+  function updateLiquidationCloseFactor(address spoke, uint256 closeFactor) external;
+
+  /**
+   * @notice Updates the liquidation health factor for max bonus of a spoke.
```
miguelmtzinf: i d rephrase a bit, it's the health factor at which the liquidation bonus is maximized.

miguelmtzinf: and make sure this docs is aligned with `ISpoke.getLiquidationHealthFactorForMaxBonus` or `ISpoke.getLiquidationConfig` getter

src/interfaces/ISpokeConfigurator.sol
```diff
@@ -0,0 +1,155 @@
...
+   */
+  function updateLiquidationFee(address spoke, uint256 reserveId, uint256 liquidationFee) external;
+
+  /**
+   * @notice Updates the collateral factor of a reserve.
+   * @param spoke The address of the spoke.
+   * @param reserveId The identifier of the reserve.
+   * @param collateralFactor The new collateral factor.
+   */
+  function updateCollateralFactor(
```
miguelmtzinf: we may need changes on dynamic risk config update functions once #459 is merged.

src/interfaces/IHubConfigurator.sol
yan-man: agree but for consistency we been using `identifier` everywhere so imo should just stick with this

src/contracts/Spoke.sol
yan-man: yes seems that is the reason, but were we going to rm `reserveCount`? I thought we were planning to rm also `_assetCount`. If we keep it should we make reserveCount internal to be consistent too

src/contracts/SpokeConfigurator.sol
```diff
@@ -0,0 +1,166 @@
...
+    targetSpoke.updateLiquidationConfig(liquidationConfig);
+  }
+
+  /// @inheritdoc ISpokeConfigurator
+  function addReserve(
+    address spoke,
+    address hub,
+    uint256 assetId,
+    address priceSource,
+    DataTypes.ReserveConfig calldata config,
```
yan-man, line 72: given 2 configs in this arg list, would prefer this be named `reserveConfig` for clarity

src/interfaces/ISpokeConfigurator.sol
yan-man:
```suggestion
   * @notice Updates the health factor for max liquidation bonus of a spoke.
```
yea something like this

src/interfaces/ISpokeConfigurator.sol
```diff
@@ -0,0 +1,155 @@
...
+   * @param closeFactor The new liquidation close factor.
+   */
+  function updateLiquidationCloseFactor(address spoke, uint256 closeFactor) external;
+
+  /**
+   * @notice Updates the liquidation health factor for max bonus of a spoke.
+   * @param spoke The address of the spoke.
+   * @param healthFactorForMaxBonus The new liquidation health factor for max bonus.
+   */
+  function updateLiquidationHealthFactorForMaxBonus(
```
yan-man: yeah this maybe should be `updateHealthFactorForMaxBonus` to be consistent or `updateHealthFactorForMaxLiquidationBonus` but then need to update everywhere else too

src/interfaces/ISpokeConfigurator.sol
```diff
@@ -0,0 +1,155 @@
...
+   * @param config The configuration of the reserve.
+   * @param dynamicConfig The dynamic configuration of the reserve.
+   * @return reserveId The identifier of the reserve.
+   */
+  function addReserve(
+    address spoke,
+    address hub,
+    uint256 assetId,
+    address priceSource,
+    DataTypes.ReserveConfig calldata config,
```
yan-man, line 65: also would prefer `reserveConfig` here

tests/unit/HubConfigurator.t.sol
```diff
@@ -0,0 +1,474 @@
...
+pragma solidity ^0.8.10;
+
+import 'tests/unit/LiquidityHub/LiquidityHubBase.t.sol';
+
+contract HubConfiguratorTest is LiquidityHubBase {
+  HubConfigurator internal hubConfigurator;
+
+  address internal HUB_CONFIGURATOR_ADMIN = makeAddr('HUB_CONFIGURATOR_ADMIN');
+
+  uint256 internal assetId;
```
yan-man: perhaps make this public or keep `_assetId` convention?

tests/unit/HubConfigurator.t.sol
```diff
@@ -0,0 +1,474 @@
...
+    assertEq(hub.getAssetCount(), assetId + 1, 'asset count');
+    assertEq(hub.getAsset(assetId).decimals, decimals, 'asset decimals');
+    assertEq(hub.getAssetConfig(assetId), expectedConfig);
+    assertEq(hub.getSpokeConfig(assetId, feeReceiver), expectedSpokeConfig);
+  }
+
+  function test_updateActive_revertsWith_OwnableUnauthorizedAccount() public {
+    vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, alice));
+    vm.prank(alice);
+    hubConfigurator.updateActive(address(hub), vm.randomUint(), vm.randomBool());
```
yan-man, line 173: any reason to do it like this instead of fuzzing these 2 vals?

tests/unit/SpokeConfigurator.t.sol
```diff
@@ -0,0 +1,416 @@
...
+import 'tests/Base.t.sol';
+
+contract SpokeConfiguratorTest is Base {
+  SpokeConfigurator internal spokeConfigurator;
+  address internal SPOKE_CONFIGURATOR_ADMIN = makeAddr('SPOKE_CONFIGURATOR_ADMIN');
+
+  address internal spokeAddr;
+  ISpoke internal spoke;
+  uint256 internal reserveId;
+  uint256 internal invalidReserveId;
```
yan-man: yeah forget what the convention we decided on for internal vars, I know even for tests we wanted to keep `_function()`

tests/unit/SpokeConfigurator.t.sol
```diff
@@ -0,0 +1,416 @@
...
+  }
+
+  function test_updateLiquidationCloseFactor_revertsWith_OwnableUnauthorizedAccount() public {
+    vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, alice));
+    vm.prank(alice);
+    spokeConfigurator.updateLiquidationCloseFactor(spokeAddr, 0);
+  }
+
+  function test_updateLiquidationCloseFactor() public {
+    uint256 newCloseFactor = ISpoke(spoke).HEALTH_FACTOR_LIQUIDATION_THRESHOLD() * 2;
```
yan-man, line 71: can we fuzz this

src/contracts/SpokeConfigurator.sol
miguelmtzinf: +1, let's add the corresponding one to HubConfigurator too

tests/unit/HubConfigurator.t.sol
miguelmtzinf: this is a revert case and we have agreed not to fuzz to save time on tests running?

tests/unit/SpokeConfigurator.t.sol
miguelmtzinf: for `Configurator` tests, do we need to fuzz? I was thinking that we can fuzz the Hub/Spoke, and just have minimal test suite for their own logic. Othwerise we dup tests?

tests/unit/HubConfigurator.t.sol
DhairyaSethi: in general: 
fuzzing the caller for auth tests is worth it bc it could potentially reveal misconfigurations 
we need not fuzz for out of bound values like invalid bps, decimals etc 

for this pr: 
i agree with the below comment to have minimal unit tests for configurator to avoid dup

src/contracts/Spoke.sol
avniculae: how can we remove `reserveCount`? we need it when adding new reserves in the current model

src/contracts/Spoke.sol
avniculae: at the cost of one extra external call, but I think it's fine given it's a governance only function

src/interfaces/ISpokeConfigurator.sol
avniculae: not to this functions, but I fixed conflicts [c159cea](https://github.com/aave/aave-v4/pull/513/commits/c159cea80cac5aa19c42b8424a4a68851fdd332d)

tests/unit/HubConfigurator.t.sol
avniculae: these are tests, do we use this convention here as well?

src/interfaces/IHubConfigurator.sol
avniculae: I also prefer id (in all cases, not just for dev comments), but will keep identifier for now to align with the rest of comments.

src/contracts/SpokeConfigurator.sol
avniculae: I used the same var names as Spoke. Should we change them there as well?

src/interfaces/ISpokeConfigurator.sol
avniculae: [e6af63a](https://github.com/aave/aave-v4/pull/513/commits/e6af63a32c67962f4acfd4adff793a57d89f486b)

src/interfaces/ISpokeConfigurator.sol
avniculae: [e6af63a](https://github.com/aave/aave-v4/pull/513/commits/e6af63a32c67962f4acfd4adff793a57d89f486b)

src/contracts/Spoke.sol
avniculae: [e6af63a](https://github.com/aave/aave-v4/pull/513/commits/e6af63a32c67962f4acfd4adff793a57d89f486b)

src/contracts/SpokeConfigurator.sol
avniculae: [e6af63a](https://github.com/aave/aave-v4/pull/513/commits/e6af63a32c67962f4acfd4adff793a57d89f486b)

src/interfaces/ISpokeConfigurator.sol
avniculae: https://github.com/aave/aave-v4/pull/513#discussion_r2207724601

tests/unit/HubConfigurator.t.sol
avniculae: configurators are ownable, so I think it's fine to not fuzz caller, wdyt?

src/contracts/Spoke.sol
yan-man: does it differ from the rationale for `_assetCount`? is that one also needed to add new assets

src/contracts/SpokeConfigurator.sol
yan-man: my pref personally would be to have it as `reserveConfig` to differentiate it from `dynamicConfig` there too, but idk how others feel

src/contracts/Spoke.sol
avniculae: yes, it's also needed to add new assets

src/interfaces/IHubConfigurator.sol
CheyenneAtapour: thumbs up under duress

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Yeah, we are supposed to be using that convention even in tests (need to fix the ones that aren't eventually)

tests/Base.t.sol
```diff
@@ -1920,6 +1921,34 @@ abstract contract Base is Test {
...
+    assertEq(a.liquidityPremium, b.liquidityPremium, 'assertEq(ReserveConfig): liquidityPremium');
+    assertEq(a.liquidationFee, b.liquidationFee, 'assertEq(ReserveConfig): liquidationFee');
+    assertEq(abi.encode(a), abi.encode(b), 'assertEq(ReserveConfig): all fields');
+  }
+
+  function assertEq(DataTypes.DynamicReserveConfig memory a, DataTypes.DynamicReserveConfig memory b) internal pure {
+    assertEq(a, b, 'assertEq(DynamicReserveConfig): ');
+  }
+
+  function assertEq(DataTypes.DynamicReserveConfig memory a, DataTypes.DynamicReserveConfig memory b, string memory label) internal pure {
```
DhairyaSethi: the label idea is inconsistent here bc the assertEq(struct) suffix won't apply if this is used directly (it can be merged with the above) 
would rm assertEq(struct) suffix in general

src/contracts/SpokeConfigurator.sol
CheyenneAtapour: I like `config` and `dynamicConfig`, it's clear they both refer to a reserve in this context. I also prefer the brevity

tests/unit/SpokeConfigurator.t.sol
```diff
@@ -0,0 +1,495 @@
...
+        ISpoke.addReserve,
+        (address(hub), daiAssetId, newPriceSource, config, dynamicConfig)
+      )
+    );
+    vm.expectEmit(address(spoke));
+    emit ISpoke.ReserveAdded(expectedReserveId, daiAssetId, address(hub));
+    vm.expectEmit(address(spoke));
+    emit ISpoke.ReserveConfigUpdated(expectedReserveId, config);
+    vm.expectEmit(address(spoke));
+    emit ISpoke.DynamicReserveConfigUpdated(expectedReserveId, 0, dynamicConfig);
```
DhairyaSethi, line 190: let's add another test on top of this just for completeness sake which updates dyn risk config again bc it has an inbuilt configKey counter we can validate with _nextConfigKey

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: I think we can fuzz caller here, because that is basically what is causing the revert, and thus the expected behavior is related to the addresses calling it. We don't need to fuzz the other stuff

tests/unit/HubConfigurator.t.sol
```diff
@@ -0,0 +1,541 @@
...
+
+  function test_updateFrozen_revertsWith_OwnableUnauthorizedAccount() public {
+    vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, alice));
+    vm.prank(alice);
+    hubConfigurator.updateFrozen(address(hub), vm.randomUint(), vm.randomBool());
+  }
+
+  function test_updateFrozen() public {
+    DataTypes.AssetConfig memory expectedConfig = hub.getAssetConfig(assetId);
+    expectedConfig.frozen = !expectedConfig.frozen;
```
CheyenneAtapour, line 220: I wonder if it's worth it for these flag updating tests to explicitly test e.g. here going from unfrozen to frozen and frozen to unfrozen, to know we've exhausted the relevant possibilities. Or just fuzz the bool. Or better: Update the current value, then update it again, so we've tested both

src/contracts/SpokeConfigurator.sol
avniculae: yeah, the reason I don't like `reserveConfig` is bc both of them apply to reserves

src/contracts/Spoke.sol
avniculae: we agreed to tackle this later

tests/unit/HubConfigurator.t.sol
avniculae: honestly it feels like they should live in the asset config tests. in fact, we have fuzz tests there that fuzz the entire struct, so I think it's fine to only do this here, wdyt?

src/contracts/Spoke.sol
avniculae: [e824dba](https://github.com/aave/aave-v4/pull/513/commits/e824dba3a8ceeda0c3a5b1e535b34cedc65cd860)

tests/unit/HubConfigurator.t.sol
avniculae: [e824dba](https://github.com/aave/aave-v4/pull/513/commits/e824dba3a8ceeda0c3a5b1e535b34cedc65cd860)

tests/unit/SpokeConfigurator.t.sol
avniculae: [e824dba](https://github.com/aave/aave-v4/pull/513/commits/e824dba3a8ceeda0c3a5b1e535b34cedc65cd860)

tests/Base.t.sol
avniculae: [e824dba](https://github.com/aave/aave-v4/pull/513/commits/e824dba3a8ceeda0c3a5b1e535b34cedc65cd860)

tests/unit/SpokeConfigurator.t.sol
```diff
@@ -0,0 +1,495 @@
...
+  }
+
+  function test_updateCollateralFactor() public {
+    uint16 newCollateralFactor = uint16(PercentageMathExtended.PERCENTAGE_FACTOR / 2);
+
+    DataTypes.DynamicReserveConfig memory expectedDynamicReserveConfig = spoke
+      .getDynamicReserveConfig(reserveId);
+    expectedDynamicReserveConfig.collateralFactor = newCollateralFactor;
+
+    uint16 expectedConfigKey = spoke.getReserve(reserveId).dynamicConfigKey + 1;
```
DhairyaSethi, line 397:
```suggestion
    uint16 expectedConfigKey = _nextDynConfigKey(spoke1);
```

src/contracts/SpokeConfigurator.sol
```diff
@@ -0,0 +1,185 @@
...
+    uint256 liquidationFee
+  ) external onlyOwner {
+    ISpoke targetSpoke = ISpoke(spoke);
+    DataTypes.ReserveConfig memory reserveConfig = targetSpoke.getReserveConfig(reserveId);
+    reserveConfig.liquidationFee = liquidationFee;
+    targetSpoke.updateReserveConfig(reserveId, reserveConfig);
+  }
+
+  /// @inheritdoc ISpokeConfigurator
+  function updateReserveConfig(
```
miguelmtzinf: can you move this down, right before `updateDynamicReserveConfig`
for consistency

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: well, I think it's nice to show the configurator can do both actions. For example, in a different project, there are stewards that can set a variable from nonzero to 0, but not the other way around. So imo, it's nice to show the configurator can set the frozen flag in both cases.

src/contracts/SpokeConfigurator.sol
avniculae: [1b52107](https://github.com/aave/aave-v4/pull/513/commits/1b521075742666da31682ca35d9f9e124f07c05a)

tests/unit/HubConfigurator.t.sol
avniculae: [541dae2](https://github.com/aave/aave-v4/pull/513/commits/541dae24180a845b5a9e5e352cdeac1c9a7e63d9)

tests/unit/SpokeConfigurator.t.sol
avniculae: Addressed in a different PR: [3e2182b](https://github.com/aave/aave-v4/pull/509/commits/3e2182b42d66adcc7ec9fdec140e403b12555e7b)

PR #515, opened 2025-07-08
fix: multicall return

DhairyaSethi:
closes: #516

### DIFF
--- a/src/misc/Multicall.sol
+++ b/src/misc/Multicall.sol
@@ -23,5 +23,6 @@ abstract contract Multicall is IMulticall {
 
       results[i] = res;
     }
+    return results;
   }
 }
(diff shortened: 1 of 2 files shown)


src/misc/Multicall.sol
```diff
@@ -23,5 +23,6 @@ abstract contract Multicall is IMulticall {
 
       results[i] = res;
     }
+    return results;
```
miguelmtzinf, line 26: Add return docs param to `IMulticall` please

DhairyaSethi: exists already
https://github.com/aave/aave-v4/blob/137a8a01ebe5cd1f166fd6efdaa84222d938eb7b/src/interfaces/IMulticall.sol#L5-L10

PR #521, opened 2025-07-10
feat: Allow Users to Update Their Dynamic Config

### LINKED ISSUE #520: Allow User to Update Dynamic Risk Config
We need to allow user to update their dynamic risk config

Linear: https://linear.app/avara/issue/AAVE-741/allow-user-to-update-their-dynamic-risk-config

CheyenneAtapour:
* Adds 1 function allowing users to update their dynamic config for all their collateral reserves

closes #520


src/contracts/Spoke.sol
```diff
@@ -433,6 +433,14 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     emit UserRiskPremiumUpdate(user, userRiskPremium);
   }
 
+  function updateUserDynamicConfig(uint256 reserveId) external {
+    _refreshDynamicConfig(msg.sender, reserveId);
+  }
```
DhairyaSethi: i think this has some implications we need to think through, refresh all is fine

yan-man: what kinds of implications are you worried about here

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -169,4 +169,104 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
...
   }
+
+  function test_manual_single_triggers_dynamicConfigUpdate() public {
+    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 95_00);
+    DynamicConfig[] memory configs = _getUserDynConfigKeys(spoke1, alice);
+
+    // no action yet, so user config should not change
+    assertEq(_getUserDynConfigKeys(spoke1, alice), configs);
+    assertNotEq(_getSpokeDynConfigKeys(spoke1), configs);
```
yan-man, line 182: in the `assertEq` and `assertNotEq` helpers, can you also add the check for liq bonus field based on the PR you just merged

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -169,4 +169,104 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
     assertNotEq(_getUserDynConfigKeys(spoke1, alice), configs);
     assertEq(_getSpokeDynConfigKeys(spoke1), _getUserDynConfigKeys(spoke1, alice));
   }
+
+  function test_manual_single_triggers_dynamicConfigUpdate() public {
+    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 95_00);
```
yan-man, line 177: can you like set `95_00` as a var like `newCollateralFactor`, and then do an assert that it doesn't equal the current value

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -169,4 +169,104 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
...
+    assertEq(_getSpokeDynConfigKeys(spoke1), _getUserDynConfigKeys(spoke1, alice));
+  }
+
+  /// @dev Tests user having two collaterals, but only updates one of their dynamic configs.
+  function test_manual_single_collateral_update_triggers_dynamicConfigUpdate() public {
+    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
+    Utils.supplyCollateral(spoke1, _wethReserveId(spoke1), alice, 1e18, alice);
+
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 95_00);
+    updateCollateralFactor(spoke1, _wethReserveId(spoke1), 90_00);
```
yan-man, line 200: same on making sure these new CF values aren't equal to existing ones

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -169,4 +169,104 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
...
+    assertNotEq(_getSpokeDynConfigKeys(spoke1), configs);
+
+    // alice only manually triggers weth dyn config update
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.UserDynamicConfigRefreshed(alice, _wethReserveId(spoke1));
+    vm.prank(alice);
+    spoke1.updateUserDynamicConfig(_wethReserveId(spoke1));
+
+    // user config should change for weth
+    assertNotEq(abi.encode(_getUserDynConfigKeys(spoke1, alice)), abi.encode(configs));
```
yan-man: can this use the `assertNotEq` helper instead

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -169,4 +169,104 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
...
+    spoke1.updateUserDynamicConfig(_usdxReserveId(spoke1));
+
+    // user config should change for usdx
+    assertNotEq(_getUserDynConfigKeys(spoke1, alice), configs);
+    assertEq(_getSpokeDynConfigKeys(spoke1), _getUserDynConfigKeys(spoke1, alice));
+  }
+
+  function test_manual_single_collateral_updateAll_triggers_dynamicConfigUpdate() public {
+    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 95_00);
```
yan-man: same

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -169,4 +169,104 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
...
+    assertNotEq(_getUserDynConfigKeys(spoke1, alice), configs);
+    assertEq(_getSpokeDynConfigKeys(spoke1), _getUserDynConfigKeys(spoke1, alice));
+  }
+
+  function test_manual_updateAll_triggers_dynamicConfigUpdate() public {
+    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
+    Utils.supplyCollateral(spoke1, _wethReserveId(spoke1), alice, 1e18, alice);
+
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 95_00);
+    updateCollateralFactor(spoke1, _wethReserveId(spoke1), 90_00);
```
yan-man, line 225: same

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
CheyenneAtapour: I'd have to change the struct type that @DhairyaSethi wrote, wdyt Dhairya?

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
CheyenneAtapour: Well, even if it's the same value, the key is updated, which is what's checked. But I can do that if you think it's necessary

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
CheyenneAtapour: Looks like that helper checks that ALL of the keys DON'T match, but it's just the case that one of them doesn't match

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
DhairyaSethi: the method gets config **keys** bc it's better to compare keys rather than the whole structs bc two values can the be the same across keys. think liq x, update to y and then update x

CheyenneAtapour: Yep, cool, just like below then, I think it's fine to look at the keys rather than the values, meaning it doesn't matter the value is actually changed, we are only concerned about keys. Also more maintainable this way, because for example, there is another PR I'll make later changing what dynamic config itself consists of

src/contracts/Spoke.sol
DhairyaSethi: being gameable, not sure of a scenario yet

CheyenneAtapour: Ah, I see, we currently don't update just one existing collateral's dynamic config anywhere. I decided to just remove this

DhairyaSethi: > we currently don't update just one existing collateral's dynamic config anywhere

we do when enabling as collateral

src/contracts/Spoke.sol
```diff
@@ -339,6 +339,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     }
   }
 
+  function updateUserDynamicConfig() external {
```
DhairyaSethi: optional: add user input and leave it untouched for now bc i'll make this callable by posm 
can do in the posm PR also so nbd

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -171,4 +171,47 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
     assertNotEq(_getUserDynConfigKeys(spoke1, alice), configs);
     assertEq(_getSpokeDynConfigKeys(spoke1), _getUserDynConfigKeys(spoke1, alice));
   }
+
+  function test_manual_single_collateral_updateAll_triggers_dynamicConfigUpdate() public {
```
DhairyaSethi, line 175: also can we think of some scenario tests for this new method as well, can note in ticket for now also

src/contracts/Spoke.sol
CheyenneAtapour: But they can't game it in the same way I mean. Because if they already have that particular asset set as collateral. They can't just set it again, since it does a no-op and doesn't update dynamic config. Also if they unset and set again, then it actually just triggers all dynamic config to be updated on the unset.

src/contracts/Spoke.sol
CheyenneAtapour: Maybe should do in that pr

DhairyaSethi: posm got merged, can make it doable through posm as well 😅

src/contracts/Spoke.sol
```diff
@@ -382,6 +382,11 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     }
   }
 
+  function updateUserDynamicConfig(address user) external {
+    require(user == msg.sender || _isPositionManager(user, msg.sender), Unauthorized());
+    _refreshDynamicConfig(user);
+  }
```
```suggestion
  /// @inheritdoc ISpoke
  function updateUserDynamicConfig(address onBehalfOf) external onlyPositionManager(onBehalfOf) {
    _refreshDynamicConfig(onBehalfOf);
  }
```

src/interfaces/ISpoke.sol
```diff
@@ -270,6 +270,13 @@ interface ISpoke is IMulticall, IAccessManaged {
    */
   function updateUserRiskPremium(address user) external;
 
+  /**
+   * @notice Allows a user to update their dynamic configuration for all collateral reserves.
+   * @dev An authorized position manager can call this on behalf of the user.
+   * @param user The address of the user whose dynamic configuration is to be updated.
+   */
+  function updateUserDynamicConfig(address user) external;
```
```suggestion
   * @dev Caller must be `onBehalfOf` or an authorized position manager for `onBehalfOf`.
   * @param onBehalfOf The owner of the position being modified.
   */
  function updateUserDynamicConfig(address onBehalfOf) external;
```

DhairyaSethi: also would rework the notice since it's not exactly `Allows a user to` anymore completely

tests/unit/Spoke/Spoke.PositionManager.t.sol
```diff
@@ -238,6 +238,36 @@ contract SpokePositionManagerTest is SpokeBase {
...
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.UserDynamicConfigRefreshedAll(alice);
+    vm.prank(POSITION_MANAGER);
+    spoke1.updateUserDynamicConfig(alice);
+
+    assertNotEq(_getUserDynConfigKeys(spoke1, alice), configs);
+    assertEq(_getSpokeDynConfigKeys(spoke1), _getUserDynConfigKeys(spoke1, alice));
+
+    _disablePositionManager();
+    vm.expectRevert(abi.encodeWithSelector(ISpoke.Unauthorized.selector));
```
DhairyaSethi:
```suggestion
    vm.expectRevert(ISpoke.Unauthorized.selector);
```

CheyenneAtapour: Done

src/contracts/Spoke.sol

src/interfaces/ISpoke.sol
CheyenneAtapour: Done

DhairyaSethi: potentially we could rework the notice text as well to be consistent with others

CheyenneAtapour: How so?

src/contracts/Spoke.sol
```diff
@@ -382,6 +382,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     }
   }
 
+  function updateUserDynamicConfig(address onBehalfOf) external onlyPositionManager(onBehalfOf) {
```
```suggestion
  /// @inheritdoc ISpoke
  function updateUserDynamicConfig(address onBehalfOf) external onlyPositionManager(onBehalfOf) {
```

CheyenneAtapour: Done, sorry, thought this was there before

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -171,4 +171,77 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
...
+    vm.prank(alice);
+    spoke1.updateUserDynamicConfig(alice);
+
+    // user config should change
+    assertNotEq(_getUserDynConfigKeys(spoke1, alice), configs);
+    assertEq(_getSpokeDynConfigKeys(spoke1), _getUserDynConfigKeys(spoke1, alice));
+  }
+
+  function test_manual_withPermission_triggers_dynamicConfigUpdate(address caller) public {
+    vm.assume(caller != alice);
```
DhairyaSethi, line 219: wanted to use alex assume helpers here for it not being forge, zero, precompile etc? nbd, we'll tackle for all tests together

CheyenneAtapour: But that doesn't matter here for just the caller. More relevant when it's a contract or something where it having code actually matters and affects the test

PR #523, opened 2025-07-10
Claim from Treasury

### LINKED ISSUE #450: Add functions to transfer assets out from TreasurySpoke
Enrich funcionality of TreasurySpoke by adding functions to transfer ERC20 tokens out, to be used for liquidity mgmt or just rescue actions.

foodaka:
closes #450 

PR to transfer erc20 tokens out of the treasury spoke via owner role

### DIFF
--- a/tests/unit/Spoke/TreasurySpoke.t.sol
+++ b/tests/unit/Spoke/TreasurySpoke.t.sol
@@ -4,11 +4,11 @@ pragma solidity ^0.8.0;
 import 'tests/unit/Spoke/SpokeBase.t.sol';
 
 contract TreasurySpokeTest is SpokeBase {
-  MockERC20 internal testToken;
+  MockERC20 internal _testToken;
 
   function setUp() public virtual override {
     super.setUp();
-    testToken = new MockERC20();
+    _testToken = new MockERC20();
   }
 
   function test_initial_state() public view {
@@ -145,15 +145,15 @@ contract TreasurySpokeTest is SpokeBase {
     amount = bound(amount, 1, type(uint128).max);
     transferAmount = bound(transferAmount, 1, amount);
 
-    testToken.mint(address(treasurySpoke), amount);
+    _testToken.mint(address(treasurySpoke), amount);
 
-    vm.expectEmit(address(testToken));
+    vm.expectEmit(address(_testToken));
     emit IERC20.Transfer(address(treasurySpoke), recipient, transferAmount);
     vm.prank(TREASURY_ADMIN);
-    treasurySpoke.transfer(address(testToken), recipient, transferAmount);
+    treasurySpoke.transfer(address(_testToken), recipient, transferAmount);
 
-    assertEq(testToken.balanceOf(address(treasurySpoke)), amount - transferAmount);
-    assertEq(testToken.balanceOf(recipient), transferAmount);
+    assertEq(_testToken.balanceOf(address(treasurySpoke)), amount - transferAmount);
+    assertEq(_testToken.balanceOf(recipient), transferAmount);
   }
 
   function _treasurySpoke() internal view returns (ISpoke) {


src/contracts/TreasurySpoke.sol
```diff
@@ -52,6 +61,16 @@ contract TreasurySpoke is Ownable, ITreasurySpoke {
     return HUB.getSpokeSuppliedShares(reserveId, address(this));
   }
 
-  // todo: add functions to transfer ERC20 out, assuming this can hold assets
+  /// @inheritdoc ITreasurySpoke
+  function transfer(address token, address to, uint256 amount) external onlyOwner {
+    if (token == address(0)) revert ZeroAddress();
+    if (to == address(0)) revert ZeroAddress();
+    if (amount == 0) revert ZeroAmount();
```
miguelmtzinf: i d remove checks completely given this is under owner control and we can assume safe calls. we don't need the event neither imho

src/contracts/TreasurySpoke.sol
```diff
@@ -13,9 +15,16 @@ import {ITreasurySpoke} from 'src/interfaces/ITreasurySpoke.sol';
  * @dev Allows withdraw to claim fees and supply to invest back into the Hub via this dedicated spoke.
  */
 contract TreasurySpoke is Ownable, ITreasurySpoke {
+  using SafeERC20 for IERC20;
+
   /// @inheritdoc ITreasurySpoke
   ILiquidityHub public immutable HUB;
 
+  error ZeroAddress();
+  error ZeroAmount();
```
yan-man: think we can add these to `ITreasurySpoke`

src/contracts/TreasurySpoke.sol
```diff
@@ -52,6 +61,16 @@ contract TreasurySpoke is Ownable, ITreasurySpoke {
     return HUB.getSpokeSuppliedShares(reserveId, address(this));
   }
 
-  // todo: add functions to transfer ERC20 out, assuming this can hold assets
+  /// @inheritdoc ITreasurySpoke
+  function transfer(address token, address to, uint256 amount) external onlyOwner {
+    if (token == address(0)) revert ZeroAddress();
+    if (to == address(0)) revert ZeroAddress();
```
yan-man:
```suggestion
    if (token == address(0) || to == address(0)) revert ZeroAddress();
```
if we don't rm these checks, can combine these 2 together

src/contracts/TreasurySpoke.sol
yan-man: think we dont need the event also bc the safeTransfer will emit an event anyway too

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -119,6 +130,85 @@ contract TreasurySpokeTest is SpokeBase {
...
+  }
+
+  function test_transfer_revertsWith_ZeroAmount() public {
+    vm.prank(TREASURY_ADMIN);
+    vm.expectRevert(TreasurySpoke.ZeroAmount.selector);
+    treasurySpoke.transfer(address(testToken), address(1), 0);
+  }
+
+  function test_transfer_revertsWith_InsufficientBalance(uint256 amount) public {
+    vm.assume(amount > 0);
```
yan-man, line 127:
```suggestion
    amount = bound(amount, 1, UINT256_MAX);
```
preferring bound over vm assumes

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -119,6 +130,85 @@ contract TreasurySpokeTest is SpokeBase {
...
+    treasurySpoke.transfer(address(testToken), address(0), 1);
+  }
+
+  function test_transfer_revertsWith_ZeroAmount() public {
+    vm.prank(TREASURY_ADMIN);
+    vm.expectRevert(TreasurySpoke.ZeroAmount.selector);
+    treasurySpoke.transfer(address(testToken), address(1), 0);
+  }
+
+  function test_transfer_revertsWith_InsufficientBalance(uint256 amount) public {
```
yan-man, line 126: i think can also fuzz the token address too then

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -119,6 +130,85 @@ contract TreasurySpokeTest is SpokeBase {
...
+    vm.assume(recipient != address(0));
+    vm.assume(recipient != address(treasurySpoke));
+    amount = bound(amount, 1, 1000000e18);
+
+    testToken.mint(address(treasurySpoke), amount);
+    assertEq(testToken.balanceOf(address(treasurySpoke)), amount);
+    assertEq(testToken.balanceOf(recipient), 0);
+
+    vm.prank(TREASURY_ADMIN);
+    vm.expectEmit(true, true, false, true, address(treasurySpoke));
```
yan-man:
```suggestion
    vm.expectEmit(address(treasurySpoke));
```
don't need the checks here

if we aren't doing the `ERC20Transferred` can still expectEmit the ERC20 transfer event instead

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -119,6 +130,85 @@ contract TreasurySpokeTest is SpokeBase {
...
+    emit ERC20Transferred(address(testToken), recipient, amount);
+    treasurySpoke.transfer(address(testToken), recipient, amount);
+
+    assertEq(testToken.balanceOf(address(treasurySpoke)), 0);
+    assertEq(testToken.balanceOf(recipient), amount);
+  }
+
+  function test_transfer_success_partialAmount(address recipient) public {
```
yan-man: can also fuzz `transferAmount`

then do `transferAmount = bound(transferAmount, 1, totalAmount);` after `totalAmount` is defined

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -119,6 +130,85 @@ contract TreasurySpokeTest is SpokeBase {
...
+
+    uint256 totalAmount = 1000e18;
+    uint256 transferAmount = 400e18;
+
+    testToken.mint(address(treasurySpoke), totalAmount);
+    assertEq(testToken.balanceOf(address(treasurySpoke)), totalAmount);
+    assertEq(testToken.balanceOf(recipient), 0);
+
+    vm.prank(TREASURY_ADMIN);
+    vm.expectEmit(true, true, false, true, address(treasurySpoke));
```
yan-man:
```suggestion
    vm.expectEmit(address(treasurySpoke));
```

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -119,6 +130,85 @@ contract TreasurySpokeTest is SpokeBase {
...
+    vm.assume(recipient != address(treasurySpoke));
+
+    uint256 totalAmount = 1000e18;
+    uint256 transferAmount = 400e18;
+
+    testToken.mint(address(treasurySpoke), totalAmount);
+    assertEq(testToken.balanceOf(address(treasurySpoke)), totalAmount);
+    assertEq(testToken.balanceOf(recipient), 0);
+
+    vm.prank(TREASURY_ADMIN);
```
yan-man: also we wanted to have the vm.prank right before the action, so moved to L205

tests/unit/Spoke/TreasurySpoke.t.sol
DhairyaSethi: ser assume is good here bc it’s a very broad range 
https://getfoundry.sh/reference/cheatcodes/assume#assume

src/contracts/TreasurySpoke.sol
foodaka: @miguelmtzinf all 3 then? on just the amount check

foodaka: removed in 3ff47b562b7997ee39ba138c67a875f3ab0b4221

src/contracts/TreasurySpoke.sol
foodaka: removed because we wont use 3ff47b562b7997ee39ba138c67a875f3ab0b4221

src/contracts/TreasurySpoke.sol
foodaka: removed 3ff47b562b7997ee39ba138c67a875f3ab0b4221

tests/unit/Spoke/TreasurySpoke.t.sol
foodaka: if its ok i kept it the same, otherwise let me know and i will update

tests/unit/Spoke/TreasurySpoke.t.sol
foodaka: added fuzz 3ff47b562b7997ee39ba138c67a875f3ab0b4221

tests/unit/Spoke/TreasurySpoke.t.sol
foodaka: so you mean directly before the transfer right?

yan-man: yeah exactly, right before the `treasurySpoke.transfer(` action

tests/unit/Spoke/TreasurySpoke.t.sol
foodaka: added in cd2370385c1a41ff730f252a7e29131d9bea4d49

tests/unit/Spoke/TreasurySpoke.t.sol
foodaka: added in cd2370385c1a41ff730f252a7e29131d9bea4d49

tests/unit/Spoke/TreasurySpoke.t.sol
yan-man: https://github.com/aave/aave-v4/pull/523/commits/4d913d0408d0a3777a32c44ce91265f011de8b0c

tests/unit/Spoke/TreasurySpoke.t.sol
yan-man: https://github.com/aave/aave-v4/pull/523/commits/4d913d0408d0a3777a32c44ce91265f011de8b0c

tests/unit/Spoke/TreasurySpoke.t.sol
yan-man: this was a bad idea, I reverted it: https://github.com/aave/aave-v4/pull/523/commits/4d913d0408d0a3777a32c44ce91265f011de8b0c

src/contracts/TreasurySpoke.sol
yan-man: also removed the event: https://github.com/aave/aave-v4/pull/523/commits/4d913d0408d0a3777a32c44ce91265f011de8b0c

src/interfaces/ITreasurySpoke.sol
```diff
@@ -45,6 +45,16 @@ interface ITreasurySpoke {
...
   function getSuppliedShares(uint256 reserveId) external view returns (uint256);
 
+  /**
+   * @notice Transfers ERC20 tokens held by this contract to a specified address.
+   * @dev Only callable by owner.
+   * @param token The address of the ERC20 token to transfer.
+   * @param to The recipient address.
+   * @param amount The amount of tokens to transfer.
+   */
+
```
CheyenneAtapour: remove this newline

yan-man: https://github.com/aave/aave-v4/pull/523/commits/f0b381b78e4b2d598862ae150f1cdfc36388954b

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -119,6 +128,61 @@ contract TreasurySpokeTest is SpokeBase {
...
+    vm.expectRevert(
+      abi.encodeWithSelector(
+        IERC20Errors.ERC20InsufficientBalance.selector,
+        address(treasurySpoke),
+        0,
+        amount
+      )
+    );
+    treasurySpoke.transfer(token, vm.randomAddress(), amount);
```
CheyenneAtapour, line 139: I don't like using `vm.randomAddress()` here because it's possible to generate `address(0)`, which would cause a different revert condition than expected. This would cause tests / ci to fail in the very rare case that it is generated, so perhaps better to just use a constant like `address(1)`

DhairyaSethi: > use a constant like address(1)

disagree with this, addr(1) points to ecrecover precompile which can cause unintended behaviour in general (fine here)
preference should be to use makeAddr('temp') in these cases

yan-man: was originally address(1) but changed it to this. Normally use makeAddr instead but here even if it is address(0) it should not revert in this context. 

but that brings up a point, perhaps in spoke `withdraw` we should validate against `to` being zero address, which I don't think we do atm

PR #534, opened 2025-07-15
Move Liquidation Fee to Dynamic Config

### LINKED ISSUE #529: Liquidation Fee may belong to dynamic risk configuration to avoid
The `liquidationFee` is a percentage of the collateral bonus to be paid on liquidation. 

Example: 
- 1000 USDC is the liquidated collateral
- 10% of liquidation bonus, 5% of liquidation fee
- (1000 * 10% =) 100 USDC is the collateral bonus
  - (100 * 5% =) 5 USDC as fees 
  - (100 - 5 =) 95 USDC to liquidator
 
Since the collateral factor and liquidation bonus are dynamic and user-specific, it may be convenient to include the liquidation fee in the dynamic risk configuration as well, for greater flexibility. Otherwise, situations could arise where a significantly increased liquidation bonus results in a disproportionately large liquidation fee for users under newer dynamic configurations, while it remains negligible for those on earlier configurations.

Following the above example:
- USDC risk profile drastically changed
- 40% of liquidation bonus, 5% of liquidation fee (unchanged)
- 1000 USDC is the liquidated collateral
- (1000 * 40% =) 400 USDC is the collateral bonus
  - (400 * 5% =) 20 USDC as fees 
  - (400 - 20 =) 380 USDC to liquidator

As a result, changes to the liquidation bonus may be less impactful than intended, since the non-dynamic-config-based liquidation fee offsets its effect. Adjusting the fee can address this for the latest configurations but won’t resolve discrepancies for earlier ones.

Alternatively, the liquidation fee can remain in the static reserve configuration and be based on the collateral amount rather than the liquidation bonus. This way, liquidations consider two factors: the liquidation bonus and the liquidation fee, both calculated relative to the collateral liquidated.

Related to #507

CheyenneAtapour:
Closes #529

### DIFF
--- a/tests/unit/Spoke/Spoke.Config.t.sol
+++ b/tests/unit/Spoke/Spoke.Config.t.sol
@@ -111,6 +111,32 @@ contract SpokeConfigTest is SpokeBase {
     assertEq(spoke1.getReserveConfig(daiReserveId), newReserveConfig);
   }
 
+  function test_updateDynamicReserveConfig_fuzz(
+    DataTypes.DynamicReserveConfig memory newConfig
+  ) public {
+    newConfig.liquidationFee = bound(
+      newConfig.liquidationFee,
+      0,
+      PercentageMathExtended.PERCENTAGE_FACTOR
+    );
+    newConfig.collateralFactor = uint16(bound(newConfig.collateralFactor, 0, 80_00));
+    newConfig.liquidationBonus = bound(
+      newConfig.liquidationBonus,
+      PercentageMathExtended.PERCENTAGE_FACTOR,
+      125_00
+    );
+
+    uint256 daiReserveId = _daiReserveId(spoke1);
+    uint16 dynamicConfigKey = _nextDynamicConfigKey(spoke1, daiReserveId);
+
+    vm.expectEmit(address(spoke1));
+    emit ISpoke.DynamicReserveConfigUpdated(daiReserveId, dynamicConfigKey, newConfig);
+    vm.prank(SPOKE_ADMIN);
+    spoke1.updateDynamicReserveConfig(daiReserveId, newConfig);
+
+    assertEq(spoke1.getDynamicReserveConfig(daiReserveId), newConfig);
+  }
+
   function test_setUsingAsCollateral_revertsWith_ReserveCannotBeUsedAsCollateral() public {
     bool newCollateralFlag = false;
     bool usingAsCollateral = true;


src/contracts/Spoke.sol
```diff
@@ -1290,7 +1294,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     vars.closeFactor = _liquidationConfig.closeFactor;
     vars.collateralAssetPrice = oracle.getReservePrice(vars.collateralReserveId);
     vars.collateralAssetUnit = 10 ** collateralReserve.decimals;
-    vars.liquidationFee = collateralReserve.config.liquidationFee;
+    vars.liquidationFee = _getLiquidationFee(vars.collateralReserveId, user);
```
DhairyaSethi: on line 1265 we already get the correct dynamic config, think it's better we reuse that and either instantiate a storage pointer to that or pass as args

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -870,6 +869,8 @@ contract SpokeBase is Base {
     DataTypes.DynamicReserveConfig memory b
   ) internal pure {
     assertEq(a.collateralFactor, b.collateralFactor, 'collateral factor');
+    assertEq(a.liquidationBonus, b.liquidationBonus, 'liquidation bonus');
+    assertEq(a.liquidationFee, b.liquidationFee, 'liquidation protocol fee');
     assertEq(abi.encode(a), abi.encode(b)); // sanity
```
DhairyaSethi, line 874: this is why the sanity check is so imp bc we missed liq bonus the last time lol

src/contracts/Spoke.sol
CheyenneAtapour: Good call, done

src/contracts/Spoke.sol
```diff
@@ -1180,9 +1180,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     vars.debtReserveId = debtReserve.reserveId;
     vars.userCollateralBalance = getUserSuppliedAmount(vars.collateralReserveId, user);
     vars.totalDebt = baseDebt + premiumDebt;
-    vars.collateralFactor = _dynamicConfig[vars.collateralReserveId][
-      _userPositions[user][vars.collateralReserveId].configKey
-    ].collateralFactor;
+    DataTypes.DynamicReserveConfig storage collateralDynConfig = _dynamicConfig[
```
DhairyaSethi, line 1183: would be great if we eventually use this instead of the getter for `liquidationBonus` as well; let's note in the ticket for now

CheyenneAtapour: That one does a calculation though, so I feel the getter makes sense

DhairyaSethi: so do the others tho? not following your point

Issue #525, opened 2025-07-10
Offboarding a spoke / existence check


CheyenneAtapour:
We discussed to remove the existence check on a spoke (our current approach is checking the lastUpdatedTimestamp)
Only concern is when we update config of a spoke that doesn't exist, we don't have validation that we actually updated the config of a spoke in existence

CheyenneAtapour:
We are doing existence checks now https://github.com/aave/aave-v4/pull/553

PR #536, opened 2025-07-16
feat: Remove existence check on spokes

### LINKED ISSUE #524: Old Treasury Spoke Should Remain Active
When offboarding a treasury spoke, we must not set it to inactive such that funds can still be withdrawn from it

CheyenneAtapour:
* Removes existence check on spokes
* Removes `SpokeData.lastUpdateTimestamp`
* Ensures offboarded fee receiver spoke is still active, in case funds need to be drawn

Closes #524  #525

### DIFF
--- a/tests/unit/HubConfigurator.t.sol
+++ b/tests/unit/HubConfigurator.t.sol
@@ -485,11 +485,10 @@ contract HubConfiguratorTest is LiquidityHubBase {
 
   /// @dev Test update fee receiver and fees can still be withdrawn from old fee receiver
   function test_updateFeeReceiver_WithdrawFromOldSpoke() public {
-    // Ensure current fee receiver is the treasury spoke
     assertEq(
       hub.getAssetConfig(daiAssetId).feeReceiver,
       address(treasurySpoke),
-      'old fee receiver mismatch'
+      'current fee receiver matches treasury spoke'
     );
 
     // Create debt to build up fees on the existing treasury spoke
@@ -504,14 +503,11 @@ contract HubConfiguratorTest is LiquidityHubBase {
     vm.prank(HUB_CONFIGURATOR_ADMIN);
     hubConfigurator.updateFeeReceiver(address(hub), daiAssetId, address(newTreasurySpoke));
 
-    // Ensure fee receiver was updated
     assertEq(
       hub.getAssetConfig(daiAssetId).feeReceiver,
       address(newTreasurySpoke),
-      'new fee receiver mismatch'
+      'new fee receiver updated'
     );
-
-    // Ensure old fee receiver is still active
     assertTrue(
       hub.getSpokeConfig(daiAssetId, address(treasurySpoke)).active,
       'old fee receiver is not active'
@@ -520,13 +516,11 @@ contract HubConfiguratorTest is LiquidityHubBase {
     // Withdraw fees from the old treasury spoke
     Utils.withdraw(_treasurySpoke(), daiAssetId, TREASURY_ADMIN, fees, address(treasurySpoke));
 
-    // Check that the old treasury spoke is empty
-    assertEq(treasurySpoke.getSuppliedAmount(daiAssetId), 0);
+    assertEq(treasurySpoke.getSuppliedAmount(daiAssetId), 0, 'old treasury spoke should be empty');
 
     // Accrue more fees, this time to new fee receiver
     skip(365 days);
 
-    // Check that new fee receiver is getting the fees, and not old treasury spoke
     assertGt(
       newTreasurySpoke.getSuppliedAmount(daiAssetId),
       0,


CheyenneAtapour:
Note that currently to remove a spoke, we would just set it's config to 0 caps and `active: false`. Up for discussion

src/contracts/Configurator.sol
```diff
@@ -176,30 +176,17 @@ contract Configurator is Ownable, IConfigurator {
     hub.updateSpokeConfig(
       assetId,
       config.feeReceiver,
-      DataTypes.SpokeConfig({supplyCap: 0, drawCap: 0, active: false})
+      DataTypes.SpokeConfig({supplyCap: 0, drawCap: 0, active: true})
```
DhairyaSethi: can we add a regression test that an old feeReceiver can still withdraw leftover balance

CheyenneAtapour: Added a test for this, unsure what you mean by regression test though

src/contracts/LiquidityHub.sol
```diff
@@ -104,29 +104,27 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     DataTypes.SpokeConfig calldata config
   ) external restricted {
     require(assetId < _assetCount, AssetNotListed());
-    require(spoke != address(0), InvalidSpoke()); // todo: how to remove spoke
```
CheyenneAtapour, line 118: Note that I think it's sufficient to just set spoke's caps to 0 and set it inactive to essentially remove it. If a spoke is upgradeable, then we can overwrite this config if we want to make it active again

tests/unit/Configurator.t.sol
```diff
@@ -427,6 +409,46 @@ contract ConfiguratorTest is LiquidityHubBase {
...
+    );
+
+    // Ensure old fee receiver is still active
+    assertTrue(
+      hub.getSpokeConfig(daiAssetId, address(treasurySpoke)).active,
+      'old fee receiver is not active'
+    );
+
+    // Withdraw fees from the old treasury spoke
+    Utils.withdraw(_treasurySpoke(), daiAssetId, TREASURY_ADMIN, fees, address(treasurySpoke));
```
DhairyaSethi: dont need so many comments imo, looks ai generated, but nbd. also can we do a quick assertion that fees from now on go to newFeeReceiver and old fee receiver's balance remains static for sanity

src/contracts/Configurator.sol
DhairyaSethi: means a test for the bug which fails w/o this fix

tests/unit/HubConfigurator.t.sol
```diff
@@ -450,6 +603,28 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
 
+  function test_updateInterestRateStrategy_fuzz_revertsWith_InvalidIrStrategy(
+    uint256 assetId
+  ) public {
+    assetId = bound(assetId, 0, hub.getAssetCount() - 1);
+
+    vm.expectRevert(ILiquidityHub.InvalidIrStrategy.selector);
+    vm.prank(HUB_CONFIGURATOR_ADMIN);
+    hubConfigurator.updateInterestRateStrategy(address(hub), assetId, address(0));
+  }
```
DhairyaSethi:
```suggestion
  function test_updateInterestRateStrategy_revertsWith_InvalidIrStrategy() public {
    uint256 assetId = vm.randomUint(0, hub.getAssetCount() - 1);

    vm.expectRevert(ILiquidityHub.InvalidIrStrategy.selector);
    vm.prank(HUB_CONFIGURATOR_ADMIN);
    hubConfigurator.updateInterestRateStrategy(address(hub), assetId, address(0));
  }
```

assetId does not need to be fuzzed since the revert cond is focussed on irStrategy address
similar scheme for all added tests

src/contracts/HubConfigurator.sol
```diff
@@ -192,30 +192,17 @@ contract HubConfigurator is Ownable, IHubConfigurator {
     hub.updateSpokeConfig(
       assetId,
       oldConfig.feeReceiver,
-      DataTypes.SpokeConfig({supplyCap: 0, drawCap: 0, active: false})
+      DataTypes.SpokeConfig({supplyCap: 0, drawCap: 0, active: true})
```
miguelmtzinf: i d just keep whatever value i had before. (not setting to true, because it could have set to false)

src/contracts/HubConfigurator.sol
```diff
@@ -192,30 +192,17 @@ contract HubConfigurator is Ownable, IHubConfigurator {
...
-        })
-      );
-    }
+    hub.updateSpokeConfig(
+      assetId,
+      newFeeReceiver,
+      DataTypes.SpokeConfig({
+        supplyCap: type(uint256).max,
+        drawCap: type(uint256).max,
+        active: true
```
miguelmtzinf: same here? if `feeReceiver` is new, it ll be active because of `addAsset` setting it to true.
otherwise, whatever value it was. (DAO can update if needed, separately)

src/contracts/LiquidityHub.sol
miguelmtzinf: can you expand on how it would work? let's imagine we have a flawed spoke with thousands of users.
How do we offboard the spoke? or we fix it with an upgrade?

src/contracts/LiquidityHub.sol
```diff
@@ -104,29 +104,27 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
     emit SpokeConfigUpdated(assetId, spoke, config);
   }
 
+  /// @inheritdoc ILiquidityHub
   function updateSpokeConfig(
     uint256 assetId,
     address spoke,
     DataTypes.SpokeConfig calldata config
   ) external restricted {
-    require(_spokes[assetId][spoke].lastUpdateTimestamp != 0, SpokeNotListed());
```
miguelmtzinf, line 139: just realizing no way of iterating over the spokes of an asset

src/contracts/LiquidityHub.sol
miguelmtzinf: if we use this fn to overwrite, we ll be emitting `SpokeAdded` twice (eg. for the same spoke address, but upgraded)

src/contracts/LiquidityHub.sol
miguelmtzinf: we don't really know if there were an spoke here or not...
meaning that for new spokes we could either use `addSpoke` or `updateSpokeConfig` no?

CheyenneAtapour: That's correct, `addSpoke` is basically becoming redundant now

src/contracts/LiquidityHub.sol
CheyenneAtapour: In the nice case, I was thinking we freeze, let people get their positions out of that spoke, and then they are free to move to another one.

The tricky thing would be if there is a serious bug or if they are not allowed to do withdraw etc, so upgrading would be the best route there. Reason being that accounting and funds live on the hub, and it's tied to the address of this problematic spoke, so potentially not possible to withdraw from another address

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Done

tests/unit/Configurator.t.sol
CheyenneAtapour: Done

src/contracts/HubConfigurator.sol
CheyenneAtapour: It doesn't make sense if the new fee receiver is not active though, right? We can save ourselves a call to get the new fee receiver's config at least. But since this is called by governance, that cost doesn't matter so much. Happy to make this change still if you think that's better

src/contracts/HubConfigurator.sol
CheyenneAtapour: Done

tests/unit/HubConfigurator.t.sol
```diff
@@ -345,7 +478,39 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+
+    // reverts when updating asset's fee receiver to zero spoke
+    vm.expectRevert(ILiquidityHub.InvalidFeeReceiver.selector);
+    vm.prank(HUB_CONFIGURATOR_ADMIN);
+    hubConfigurator.updateFeeConfig(address(hub), assetId, liquidityFee, address(0));
+  }
+
+  function test_updateFeeConfig_fuzz_revertsWith_InvalidLiquidityFee(
+    uint256 liquidityFee,
+    address feeReceiver
```
DhairyaSethi: don't need to be fuzzed, would write the same way `test_updateFeeConfig_revertsWith_InvalidFeeReceiver` is written

tests/unit/HubConfigurator.t.sol
```diff
@@ -325,6 +406,58 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+    );
+
+    // Withdraw fees from the old treasury spoke
+    Utils.withdraw(_treasurySpoke(), daiAssetId, TREASURY_ADMIN, fees, address(treasurySpoke));
+
+    // Check that the old treasury spoke is empty
+    assertEq(treasurySpoke.getSuppliedAmount(daiAssetId), 0);
+
+    // Accrue more fees, this time to new fee receiver
+    skip(365 days);
```
DhairyaSethi: apologies unsure if already exists: test to change feeReceiver, not withdraw all, accrue on new and ensure existing is still owed the same (in **share** terms) - would prefer it on this test where existing feeReceiver only withdraws half its owed shares before the second skip (which accrues for new receiver) and then withdraws full (which is half of initial)

tests/unit/HubConfigurator.t.sol
```diff
@@ -325,6 +406,58 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+    TreasurySpoke newTreasurySpoke = new TreasurySpoke(HUB_ADMIN, address(hub));
+    vm.prank(HUB_CONFIGURATOR_ADMIN);
+    hubConfigurator.updateFeeReceiver(address(hub), daiAssetId, address(newTreasurySpoke));
+
+    // Ensure fee receiver was updated
+    assertEq(
+      hub.getAssetConfig(daiAssetId).feeReceiver,
+      address(newTreasurySpoke),
+      'fee receiver mismatch'
+    );
```
DhairyaSethi: would keep one or the other, the label feels unnecessary here but nbd

tests/unit/HubConfigurator.t.sol
```diff
@@ -270,44 +369,26 @@ contract HubConfiguratorTest is LiquidityHubBase {
           (
             assetId,
             oldConfig.feeReceiver,
-            DataTypes.SpokeConfig({supplyCap: 0, drawCap: 0, active: false})
+            DataTypes.SpokeConfig({supplyCap: 0, drawCap: 0, active: true})
```
DhairyaSethi: can we test one where oldFeeReceiver has been deactivated

tests/unit/HubConfigurator.t.sol
```diff
@@ -236,17 +301,42 @@ contract HubConfiguratorTest is LiquidityHubBase {
     hubConfigurator.updateLiquidityFee(address(hub), vm.randomUint(), vm.randomUint());
   }
 
+  function test_updateLiquidityFee_fuzz_revertsWith_InvalidLiquidityFee(
+    uint256 liquidityFee
```
DhairyaSethi: doesnt need to be fuzzed imo

tests/unit/HubConfigurator.t.sol
```diff
@@ -21,8 +21,16 @@ contract HubConfiguratorTest is LiquidityHubBase {
   }
 
   function test_addSpokeToAssets_revertsWith_OwnableUnauthorizedAccount() public {
```
avniculae: if we add fuzz versions for these, then I would remove this non-fuzzed versions, wdyt?

tests/unit/HubConfigurator.t.sol
```diff
@@ -46,6 +54,20 @@ contract HubConfiguratorTest is LiquidityHubBase {
     hubConfigurator.addSpokeToAssets(address(hub), address(spoke1), assetIds, spokeConfigs);
   }
 
+  function test_addSpokeToAssets_revertsWith_InvalidSpoke() public {
```
avniculae, line 61: I feel this test should not live here, but rather in the hub config tests, wdyt? (we already have it there)

tests/unit/HubConfigurator.t.sol
```diff
@@ -46,6 +54,20 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+    DataTypes.SpokeConfig[] memory spokeConfigs = new DataTypes.SpokeConfig[](2);
+    spokeConfigs[0] = DataTypes.SpokeConfig({active: true, supplyCap: 1, drawCap: 2});
+    spokeConfigs[1] = DataTypes.SpokeConfig({active: true, supplyCap: 3, drawCap: 4});
+
+    vm.expectRevert(ILiquidityHub.InvalidSpoke.selector);
+    vm.prank(HUB_CONFIGURATOR_ADMIN);
+    hubConfigurator.addSpokeToAssets(address(hub), address(0), assetIds, spokeConfigs);
+  }
+
   function test_addSpokeToAssets() public {
```
avniculae, line 75: I think we should add an expectCall in this test, wdyt?

avniculae: I think I missed it last time

tests/unit/HubConfigurator.t.sol
avniculae: it applies to other tests below as well

tests/unit/HubConfigurator.t.sol
```diff
@@ -110,6 +138,33 @@ contract HubConfiguratorTest is LiquidityHubBase {
     _addAsset(fetchErc20Decimals, underlying, decimals, feeReceiver, interestRateStrategy);
   }
 
+  function test_addAsset_fuzz_revertsWith_InvalidUnderlying(
```
avniculae: same comment as above: feels that this test should not live here, but rather in the hub config tests (it's already there), wdyt?

tests/unit/HubConfigurator.t.sol
```diff
@@ -110,6 +138,33 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+    uint8 decimals,
+    address feeReceiver,
+    address interestRateStrategy
+  ) public {
+    vm.expectRevert(ILiquidityHub.InvalidUnderlying.selector, address(hub));
+    vm.prank(HUB_CONFIGURATOR_ADMIN);
+    _addAsset(fetchErc20Decimals, address(0), decimals, feeReceiver, interestRateStrategy);
+  }
+
+  function test_addAsset_fuzz_revertsWith_InvalidIrStrategy(
```
avniculae: same

tests/unit/HubConfigurator.t.sol
avniculae: same comment as above, I think we can remove this test

tests/unit/HubConfigurator.t.sol
```diff
@@ -236,17 +301,42 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+  }
+
   function test_updateLiquidityFee() public {
     DataTypes.AssetConfig memory expectedConfig = hub.getAssetConfig(assetId);
     expectedConfig.liquidityFee = PercentageMathExtended.PERCENTAGE_FACTOR - 1;
 
+    test_updateLiquidityFee_fuzz(assetId, expectedConfig.liquidityFee);
+  }
+
+  function test_updateLiquidityFee_fuzz(uint256 assetId, uint256 liquidityFee) public {
```
avniculae, line 346: i don't think fuzz is needed here, but if we do it then I think we can remove non-fuzzed version

tests/unit/HubConfigurator.t.sol
```diff
@@ -257,6 +347,15 @@ contract HubConfiguratorTest is LiquidityHubBase {
     hubConfigurator.updateFeeReceiver(address(hub), vm.randomUint(), vm.randomAddress());
   }
 
+  function test_updateFeeReceiver_revertsWith_InvalidFeeReceiver() public {
```
avniculae, line 371: same, I don't think we need this here

tests/unit/HubConfigurator.t.sol
```diff
@@ -450,6 +597,25 @@ contract HubConfiguratorTest is LiquidityHubBase {
     assertEq(hub.getAssetConfig(assetId), expectedConfig);
   }
 
+  function test_updateInterestRateStrategy_revertsWith_InvalidIrStrategy() public {
```
avniculae, line 747: same

tests/unit/HubConfigurator.t.sol
```diff
@@ -325,6 +406,58 @@ contract HubConfiguratorTest is LiquidityHubBase {
     }
   }
 
+  /// @dev Test update fee receiver and fees can still be withdrawn from old fee receiver
+  function test_updateFeeReceiver_WithdrawFromOldSpoke() public {
```
avniculae, line 487: this feels duplicate to `test_updateAssetConfig_fuzz_NewFeeReceiver` in hub.config, wdyt?

tests/unit/HubConfigurator.t.sol
avniculae, line 1: main comment I have for this file is: I think we don't need to test (and hence duplicate) the hub functionality here. I think we should only expectCall, expectEmit and check end result. The edge cases are tested in the hub.config test file, wdyt?

src/contracts/HubConfigurator.sol
avniculae: wdym @miguelmtzinf ? `addAsset` is not touching the spoke config right?

miguelmtzinf: saying that i d leave the old active value.
- fees can be accrued anyway even if not active
- if governance wants to activate, they can
- if fee receiver is new, added when adding the asset, it will be active
- if fee receiver is new, not added when adding asset, will respect the existing config

src/contracts/LiquidityHub.sol
```diff
@@ -104,29 +104,27 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-    require(spoke != address(0), InvalidSpoke()); // todo: how to remove spoke
+    require(spoke != address(0), InvalidSpoke());
 
     _spokes[assetId][spoke] = DataTypes.SpokeData({
       suppliedShares: 0,
       baseDrawnShares: 0,
       premiumDrawnShares: 0,
       premiumOffset: 0,
       realizedPremium: 0,
-      lastUpdateTimestamp: block.timestamp,
```
avniculae, line 126: why do we need addReserve in this approach anymore? can we not delete it?

src/contracts/HubConfigurator.sol
CheyenneAtapour: Ok done. Btw, I see this is actually using asset config, but should be using spoke config, right?

src/contracts/LiquidityHub.sol
CheyenneAtapour: ig this is being addressed in a different pr

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: What do you mean by one or the other? I'm updating the two labels to be different and clearer

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Should we allow the hub configurator to update the fee receiver's config btw? Then I wouldn't have to deactivate old fee receiver through the hub directly

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: I think it's just nice for completeness to also have here, but it's nbd imo

src/contracts/LiquidityHub.sol
CheyenneAtapour: Maybe there's a case where we want to reset the accounting on the hub side after upgrading a spoke and wishing to re-use it, but perhaps it's quite niche, and we'd prefer just using a new spoke. @miguelmtzinf @DhairyaSethi thoughts?

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: That's true, since we want to keep this test suite more simple, a scenario style test like this I think better left in the hub config test suite

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Done

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: doesn't look like this exists, but I think I'm aligned with keeping this test suite simple (still do want to test all cases / reverts for completeness, don't have to be fuzzed). So if aligned, I can move the new test I wrote for this into `LiquidityHub.Config.t.sol`

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Imo, I think keep such tests, don't need to be fuzzed. I've applied such changes

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Added

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: See my above comments. I think it's nice for completeness, but they mostly don't need fuzzing

tests/unit/HubConfigurator.t.sol
DhairyaSethi: yeah sounds good as long as the test exists

tests/unit/HubConfigurator.t.sol
DhairyaSethi: either the label or the comment

CheyenneAtapour: Ah yeah, great point, I cleaned this up. 

In general btw, I definitely like the label because it makes debugging easier, instead of just getting some revert

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: @avniculae  is tackling this in other pr^

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Actually, I am kinda coming back on this after some thought, because on the hub config tests, it's basically just a unit test involving a singular function `hub.updateAssetConfig`, but here on the hub configurator, there are multiple different functions, and so imo, nice to have the test to ensure if we ever change the logic on hub configurator, we are not messing things up

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: I changed my mind on this (test still exists ofc): https://github.com/aave/aave-v4/pull/536#discussion_r2227269557

miguelmtzinf:
@CheyenneAtapour can we close in favor of #553 ?

CheyenneAtapour:
Closing in favor of #575, which just contains now 1 simple natspec and some extra tests for hub configurator (if we'd like to include)

PR #540, opened 2025-07-18
feat: Add Hub Conversion Functions

### LINKED ISSUE #514: Hub utility functions for shares-assets conversions
## Current State
- operations
  - `add` uses assets to shares round down
  - `remove` uses assets to shares round up
  - `draw` uses assets to shares round up
  - `restore` uses assets to shares round down
- converter functions: 
  - `previewOffset` use shares to assets round down
  - `convertToSuppliedAssets` -> rounds down
  - `convertToSuppliedAssetsUp` -> rounds up
  - `convertToSuppliedShares` -> rounds down
  - `convertToSuppliedSharesUp` -> rounds up
  - `convertToDrawnAssets` -> rounds up
  - `convertToDrawnShares` -> rounds down
- getters: 
  - `getSpokeDebt` -> uses shares to assets round up
  - `getAssetSuppliedAmount` -> uses shares to assets round down
  - `getAssetSuppliedShares` -> uses shares to assets round down
  - `getSpokeSuppliedAmount` -> uses shares to assets round down
  - `getTotalSuppliedAssets` -> for `totalAssets` (`= availableLiq + totalDebt`)
  - `getTotalSuppliedShares` -> for `totalShares` (`= totalShares + feeShares`)

Main **issue** now is the great amount of different functions with similar names which can be confusing.

### Option A

Make use of `preview` functions, so we don't have Up/Down converter query functions

- `previewAdd` -> `convertToSuppliedSharesDown`
- `previewRemove` -> `convertToSuppliedSharesUp`
- `previewDraw` -> `convertToDrawnSharesUp`
- `previewRestore` -> `convertToDrawnSharesDown`

**Questions:**
- Conversions from assets to shares are covered, what about conversions from shares to assets?
  - `convertToSuppliedAssetsDown` is used by Spoke to calculate withdrawable amounts
  - `convertToSuppliedAssetsUp` is used only for `minimumAssetsPerSuppliedShare` helper -> how much assets I need for 1 share.
  - `convertToDrawnAssets` is used by Spoke to i) calculate debt and ii) calculate premium (premiumDrawnShares into premiumDrawnAssets)
  - `convertToDrawnAssetsDown` does not exist but would be needed for `minimumAssetsPerDrawnShare` -> how much assets I need to restore for 1 share.

### Option B

Create very open/flexible converter functions with both rounding directions, so integrators (and spokes) can choose:
- `convertToSuppliedSharesDown`
- `convertToSuppliedSharesUp`
- `convertToSuppliedAssetsDown`
- `convertToSuppliedAssetsUp`
- `convertToDrawnSharesDown`
- `convertToDrawnSharesUp`
- `convertToDrawnAssetsDown`
- `convertToDrawnAssetsUp`

(not sold on the naming)

The above can be the only one functions we have, and relay on integrators goodwill, or just build some data adapters.
Alternative would be adding also functions to help integrations (like preview ones, and getters). 

### Option C

Just rename functions and add whatever converter function we are missing.
- Should we continue using 4626 terms or move away from that? Is using 4626 terms confusing given that we are not 4626 complaint?

CheyenneAtapour:
Closes #514

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -435,6 +435,7 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     return _assets[assetId].toDrawnSharesDown(assets);
   }
 
+  /// @inheritdoc ILiquidityHub
   function getAssetDrawnIndex(uint256 assetId) external view returns (uint256) {
     return _assets[assetId].getDrawnIndex();
   }
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -593,12 +593,14 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   }
 
   /**
-   * @dev Helper function to get user premium offset in assets amount from shares amount using current debt index.
+   * @dev Calculates the user's premium debt offset in assets amount from a given share amount.
    * @dev Rounds down to the nearest assets amount.
+   * @dev Uses the opposite rounding direction of the debt shares-to-assets conversion to prevent underflow
+   * in premium debt.
    * @param hub The liquidity hub of the reserve.
-   * @param assetId The asset identifier.
+   * @param assetId The identifier of the asset.
    * @param shares The amount of shares to convert to assets amount.
-   * @return The amount of assets converted from shares amount, corresponding to user premium offset.
+   * @return The amount of assets converted corresponding to user's premium offset.
    */
   function _previewOffset(
     ILiquidityHub hub,


src/interfaces/ILiquidityHub.sol
```diff
@@ -193,6 +193,65 @@ interface ILiquidityHub is IAccessManaged {
    */
   function payFee(uint256 assetId, uint256 shares) external;
 
+  /**
+   * @notice Converts the specified amount of asset to shares supplied upon an Add action.
+   * @param assetId The identifier of the asset.
+   * @param amount The amount of asset to convert to shares.
+   */
```
avniculae, line 208: add `@return` natspec (applicable to fns below as well)

src/contracts/LiquidityHub.sol
```diff
@@ -352,6 +352,49 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+  /// @inheritdoc ILiquidityHub
+  function previewBaseRestoreByShares(
+    uint256 assetId,
+    uint256 baseDrawnShares
+  ) public view returns (uint256) {
+    return _assets[assetId].toDrawnAssetsUp(baseDrawnShares);
+  }
+
   // todo 4626 getter naming
   function convertToSuppliedAssets(
```
avniculae, line 413: should we remove convert fns now?

src/contracts/LiquidityHub.sol
```diff
@@ -352,6 +352,49 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+    return _assets[assetId].toSuppliedAssetsUp(shares);
+  }
+
+  /// @inheritdoc ILiquidityHub
+  function previewRemove(uint256 assetId, uint256 amount) public view returns (uint256) {
+    return _assets[assetId].toSuppliedSharesUp(amount);
+  }
+
+  /// @inheritdoc ILiquidityHub
+  function previewRemoveByShares(uint256 assetId, uint256 shares) public view returns (uint256) {
```
avniculae, line 388: use it in `payFee` and `getSpokeSuppliedAmount`

src/contracts/LiquidityHub.sol
```diff
@@ -352,6 +352,49 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
   }
 
+  /// @inheritdoc ILiquidityHub
+  function previewAdd(uint256 assetId, uint256 amount) public view returns (uint256) {
+    return _assets[assetId].toSuppliedSharesDown(amount);
+  }
+
+  /// @inheritdoc ILiquidityHub
+  function previewAddByShares(uint256 assetId, uint256 shares) public view returns (uint256) {
+    return _assets[assetId].toSuppliedAssetsUp(shares);
```
avniculae, line 379: use it in `_validateAdd`

src/contracts/LiquidityHub.sol
```diff
@@ -352,6 +352,49 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+  }
+
+  /// @inheritdoc ILiquidityHub
+  function previewDraw(uint256 assetId, uint256 amount) public view returns (uint256) {
+    return _assets[assetId].toDrawnSharesUp(amount);
+  }
+
+  /// @inheritdoc ILiquidityHub
+  function previewDrawByShares(uint256 assetId, uint256 shares) public view returns (uint256) {
+    return _assets[assetId].toDrawnAssetsDown(shares);
```
avniculae, line 399: `previewOffset` uses it. should we delete `previewOffset`, or use `previewDrawByShares` in `previewOffset`?

src/contracts/LiquidityHub.sol
```diff
@@ -352,6 +352,49 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+  function previewBaseRestore(uint256 assetId, uint256 baseAmount) public view returns (uint256) {
+    return _assets[assetId].toDrawnSharesDown(baseAmount);
+  }
+
+  /// @inheritdoc ILiquidityHub
+  function previewBaseRestoreByShares(
+    uint256 assetId,
+    uint256 baseDrawnShares
+  ) public view returns (uint256) {
+    return _assets[assetId].toDrawnAssetsUp(baseDrawnShares);
```
avniculae: use it in `_getSpokeDebt`

src/contracts/LiquidityHub.sol
CheyenneAtapour: Determined in sync we would keep the conversions that don't have the rounding direction in their names

src/contracts/LiquidityHub.sol
CheyenneAtapour: We decided to leave `previewOffset` as is for now

src/contracts/LiquidityHub.sol
CheyenneAtapour: @miguelmtzinf @DhairyaSethi are these fine as is? I thought we mostly care about what we are exposing

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: Done, thanks!

src/contracts/LiquidityHub.sol
CheyenneAtapour: Here I'd have to pass an extra param for `assetId`, which feels annoying

src/contracts/LiquidityHub.sol
CheyenneAtapour: some of these also make the gas costs worse if we are adding a new param to a function, see above

src/contracts/LiquidityHub.sol
```diff
@@ -415,20 +440,17 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-      _spokes[assetId][spoke]
-    );
+    (uint256 baseDebt, uint256 premiumDebt) = _getSpokeDebt(_spokes[assetId][spoke], assetId);
     return baseDebt + premiumDebt;
   }
 
   function getAssetSuppliedAmount(uint256 assetId) external view returns (uint256) {
     DataTypes.Asset storage asset = _assets[assetId];
-    return asset.toSuppliedAssetsDown(asset.suppliedShares);
+    return previewRemoveByShares(assetId, asset.suppliedShares);
```
DhairyaSethi: i mean if we're keeping the convert methods exposed we can just use those, here specifically i would keep it as it as well; same for getSpoke/AssetDebt

src/contracts/LiquidityHub.sol
```diff
@@ -488,21 +512,22 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     require(!asset.config.frozen, AssetFrozen());
     require(
       spoke.config.supplyCap == type(uint256).max ||
-        asset.toSuppliedAssetsUp(spoke.suppliedShares) + amount <= spoke.config.supplyCap,
+        previewAddByShares(assetId, spoke.suppliedShares) + amount <= spoke.config.supplyCap,
```
DhairyaSethi: would keep as is
but really with 4626 convention we should use the maxAdd(assetId, spoke) call here

src/contracts/LiquidityHub.sol
DhairyaSethi, line 1: let's add maxAdd(asset, spoke), maxRemove, maxDraw, maxRestore based on the current available spoke caps as well imo 
(to continue borrowing from 4626 convention, plus these are valuable getters for spokes needed onchain)

src/contracts/LiquidityHub.sol
DhairyaSethi: imo should use the convert functions as usual there

src/contracts/LiquidityHub.sol
CheyenneAtapour: @miguelmtzinf if aligned, I can add. I thought we were trying to cut down on the number of these getters, and unsure how much of a priority it is to match erc4626

src/contracts/LiquidityHub.sol
CheyenneAtapour: We're only really concerned with what we expose, so fine imo to use the asset library elsewhere. I think it also makes sense to use previews in the action functions themselves.

@DhairyaSethi seems aligned too: https://github.com/aave/aave-v4/pull/540#discussion_r2221563833

src/contracts/LiquidityHub.sol
CheyenneAtapour: Open to it if aligned: https://github.com/aave/aave-v4/pull/540#discussion_r2224344929

src/contracts/LiquidityHub.sol
miguelmtzinf: i do think we could call `previewDrawByShares` from an internal method `_previewOffset` in `Spoke`.
separately, we ll remove this if the PR of rounding in the same direction turns out well

src/contracts/Spoke.sol
miguelmtzinf, line 1: I think
- `getReserveSuppliedAmount` and `getUserSuppliedAmount` must use `previewRemoveByShares`
- `_getReserveDebt` must use `previewRestoreByShares` for `baseDebt`
- `_getUserBalanceInBaseCurrency` must use `previewRemoveByShares`
- `_getUserDebt` must use `previewRestoreByShares` for `baseDebt`

src/contracts/LiquidityHub.sol
miguelmtzinf: we d need to split into `byShares` and `byAssets` too..
If we do it, i think:
- `withdraw` can use `maxWithdraw` once added, for full-balance withdrawal
- `validateWithdraw` can use `maxAdd` once added, for maximum amount of assets that can be supplied

src/interfaces/ILiquidityHub.sol
```diff
@@ -193,26 +193,102 @@ interface ILiquidityHub is IAccessManaged {
    */
   function payFee(uint256 assetId, uint256 shares) external;
 
+  /**
+   * @notice Converts the specified amount of assets to shares supplied upon an Add action.
```
miguelmtzinf: 1. should we specify this is for supply assets/shares?
2. i d add a `@dev` on each function, saying which rounding direction we use

src/contracts/LiquidityHub.sol
miguelmtzinf: we could keep it as is, and set preview/max getters  as external only

src/contracts/LiquidityHub.sol
miguelmtzinf: we agreed on not doing this. If so, we can add them to an external contract adapter

src/interfaces/ILiquidityHub.sol
yan-man: > should we specify this is for supply assets/shares?

it does mention this, right? or not clear enough

maybe just `amount of assets to shares added`?

src/contracts/LiquidityHub.sol
yan-man: seems like these are good candidates to be moved to external contract tbh

src/interfaces/ILiquidityHub.sol
```diff
@@ -193,26 +193,102 @@ interface ILiquidityHub is IAccessManaged {
...
+   * @param assetId The identifier of the asset.
+   * @param shares The amount of drawn shares to convert to asset amount.
+   * @return The amount of drawn assets converted from shares.
+   */
   function convertToDrawnAssets(uint256 assetId, uint256 shares) external view returns (uint256);
 
+  /**
+   * @notice Converts the specified amount of drawn assets to shares.
+   * @param assetId The identifier of the asset.
+   * @param assets The amount of drawn assets to convert to shares.
```
```suggestion
   * @param assets The amount of drawn assets to convert to shares amount.
```
for consistency, update these or rm `amount` from `_ByShares` methods

src/contracts/LiquidityHub.sol
CheyenneAtapour: Done

src/contracts/Spoke.sol
CheyenneAtapour: I did these, but for example, I think keeping `convertToSuppliedShares` for `getReserveSuppliedAmount` is more intuitive, cause it's just converting the supplied shares to amount. I think `previewRemoveByShares` looks less intuitive. We are keeping `convertToSuppliedShares` in hub anyway.

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: I tried to specify `amounts` on assets, because that is perhaps not clear it represents an amount, whereas for shares, I think that's clear. For example, "assets" could just mean multiple assets, like usdc, eth, dai, etc.

CheyenneAtapour: But yeah, I think actually easier and better just to be verbose in the natspec, so I did amount for both

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: Specified rounding direction with a dev comment, and changed now to "added", since that will be the term going forwards

src/contracts/Spoke.sol
miguelmtzinf: Imo keeping convert functions in hub it's separate topic. So I disagree, but nbd

src/contracts/LiquidityHub.sol
```diff
@@ -418,10 +418,6 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     return _assets[assetId].toDrawnSharesDown(assets);
   }
 
-  function previewOffset(uint256 assetId, uint256 shares) external view returns (uint256) {
-    return _assets[assetId].toDrawnAssetsDown(shares);
-  }
-
   function previewDrawnIndex(uint256 assetId) external view returns (uint256) {
```
miguelmtzinf: damn, should we rename this now?

src/contracts/Spoke.sol
```diff
@@ -604,6 +604,14 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     // todo NoExplicitAmountToRepayOnBehalf
   }
 
+  function _previewOffset(
```
miguelmtzinf, line 605: add docs specifying rounding direction pls

src/contracts/LiquidityHub.sol
miguelmtzinf: does `getNextDrawnIndex` work? or just `getAssetDrawnIndex`?
I like the latter because it's what we are doing something similar in v3: `getPoolReserveNormalizedIndex` or smt like that gives the next one and `getReserve` the latest one.
Using `next` term could be confusing because this is changing all the time until the next action

CheyenneAtapour: Yeah, `getAssetDrawnIndex` is fine for me

src/contracts/Spoke.sol
CheyenneAtapour: Right now, it's done as you like btw

src/contracts/LiquidityHub.sol
CheyenneAtapour: On second thought, I'm also fine leaving as is. We also have `previewFeeShares` on `AssetLogic` as well. I think they basically follow the same convention to look at what "would" happen when an action happens

src/contracts/Spoke.sol
CheyenneAtapour: Done

DhairyaSethi: @CheyenneAtapour we don't need full natspec on internal methods, a dev comment about the rounding dir & reason for diff with premiumShares is enough

CheyenneAtapour: I reduced it down now to 2 dev comments because it was a long description

miguelmtzinf: @DhairyaSethi i was actually pushing for full natspec though, for clarity

src/contracts/LiquidityHub.sol
miguelmtzinf: but that's internal right

src/contracts/Spoke.sol
CheyenneAtapour: Reverted this change

DhairyaSethi: @CheyenneAtapour its missing the main point: reason for diff with premiumShares rounding

CheyenneAtapour: should I add to the dev comment "to avoid underflow when subtracting from premium drawn shares"?

PR #541, opened 2025-07-19
feat: update existing dyn config

DhairyaSethi:
Introduces add and update dynamic config. 
Both emit separate events such that indexer is able to track the latest reserve config key.

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -123,7 +123,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
 
     emit ReserveAdded(reserveId, assetId, hub);
     emit ReserveConfigUpdated(reserveId, config);
-    emit DynamicReserveConfigUpdated(reserveId, dynamicConfigKey, dynamicConfig);
+    emit DynamicReserveConfigAdded(reserveId, dynamicConfigKey, dynamicConfig);
 
     return reserveId;
   }
@@ -151,7 +151,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     unchecked {
       configKey = ++_reserves[reserveId].dynamicConfigKey;
     }
-    _setDynamicReserveConfig(reserveId, configKey, dynamicConfig);
+    _validateDynamicReserveConfig(dynamicConfig);
+    _dynamicConfig[reserveId][configKey] = dynamicConfig;
     emit DynamicReserveConfigAdded(reserveId, configKey, dynamicConfig);
     return configKey;
   }
@@ -165,7 +166,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     require(reserveId < _reserveCount, ReserveNotListed());
     // @dev sufficient check since min liquidationBonus is 100_00
     require(_dynamicConfig[reserveId][configKey].liquidationBonus != 0, ConfigKeyUninitialized());
-    _setDynamicReserveConfig(reserveId, configKey, dynamicConfig);
+    _validateDynamicReserveConfig(dynamicConfig);
+    _dynamicConfig[reserveId][configKey] = dynamicConfig;
     emit DynamicReserveConfigUpdated(reserveId, configKey, dynamicConfig);
   }
 
@@ -1051,15 +1053,6 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   function _refreshDynamicConfig(address user, uint256 reserveId) internal {
     _userPositions[user][reserveId].configKey = _reserves[reserveId].dynamicConfigKey;
     emit UserDynamicConfigRefreshedSingle(user, reserveId);
-  }
-
-  function _setDynamicReserveConfig(
-    uint256 reserveId,
-    uint16 configKey,
-    DataTypes.DynamicReserveConfig calldata dynamicConfig
-  ) internal {
-    _validateDynamicReserveConfig(dynamicConfig);
-    _dynamicConfig[reserveId][configKey] = dynamicConfig;
   }
 
   /// @return collateralAsset The address of the underlying asset used as collateral, to receive as result of the liquidation.
--- a/tests/unit/Spoke/Spoke.Config.t.sol
+++ b/tests/unit/Spoke/Spoke.Config.t.sol
@@ -155,7 +155,7 @@ contract SpokeConfigTest is SpokeBase {
     vm.expectEmit(address(spoke1));
     emit ISpoke.ReserveConfigUpdated(reserveId, newReserveConfig);
     vm.expectEmit(address(spoke1));
-    emit ISpoke.DynamicReserveConfigUpdated({
+    emit ISpoke.DynamicReserveConfigAdded({
       reserveId: reserveId,
       configKey: 0,
       config: newDynReserveConfig
--- a/src/interfaces/ISpoke.sol
+++ b/src/interfaces/ISpoke.sol
@@ -14,17 +14,45 @@ import {DataTypes} from 'src/libraries/types/DataTypes.sol';
 interface ISpoke is IMulticall, IAccessManaged {
   event ReserveAdded(uint256 indexed reserveId, uint256 indexed assetId, address indexed hub);
   event ReserveConfigUpdated(uint256 indexed reserveId, DataTypes.ReserveConfig config);
+
+  /**
+   * @notice Emitted when a dynamic reserve config is added.
+   * @dev The config key is the next available key for the reserve, which is now the latest config
+   * key of the reserve. It can be an existing key that was previously used and is now being
+   * overridden.
+   * @param reserveId The identifier of the reserve.
+   * @param configKey The key of the added dynamic config.
+   * @param config The dynamic reserve config.
+   */
   event DynamicReserveConfigAdded(
     uint256 indexed reserveId,
     uint16 indexed configKey,
     DataTypes.DynamicReserveConfig config
   );
+
+  /**
+   * @notice Emitted when a dynamic reserve config is updated.
+   * @param reserveId The identifier of the reserve.
+   * @param configKey The key of the updated dynamic config.
+   * @param config The dynamic reserve config.
+   */
   event DynamicReserveConfigUpdated(
     uint256 indexed reserveId,
     uint16 indexed configKey,
     DataTypes.DynamicReserveConfig config
   );
+
+  /**
+   * @notice Emitted when a user's dynamic config is refreshed for all reserves to their latest config key.
+   * @param user The address of the user.
+   */
   event UserDynamicConfigRefreshedAll(address indexed user);
+
+  /**
+   * @notice Emitted when a user's dynamic config is refreshed for a single reserve to its latest config key.
+   * @param user The address of the user.
+   * @param reserveId The identifier of the reserve.
+   */
   event UserDynamicConfigRefreshedSingle(address indexed user, uint256 reserveId);
 
   /**
@@ -206,7 +234,7 @@ interface ISpoke is IMulticall, IAccessManaged {
 
   /**
    * @notice Updates the dynamic reserve config for a given reserve at the specified key.
-   * @dev Reverts with `ConfigKeyUninitialized` if the config key has not been uninitialized yet.
+   * @dev Reverts with `ConfigKeyUninitialized` if the config key has not been initialized yet.
    * @param reserveId The identifier of the reserve.
    * @param configKey The key of the config to update.
    * @param dynamicConfig The dynamic reserve config to update.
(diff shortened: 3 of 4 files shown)


src/contracts/Spoke.sol
```diff
@@ -133,6 +133,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     emit ReserveConfigUpdated(reserveId, config);
   }
 
+  /// @inheritdoc ISpoke
   function updateDynamicReserveConfig(
```
avniculae: this can call `updateExistingDynamicReserveConfig` now. any other refactoring is fine as long as we avoid code duplication

tests/Base.t.sol
```diff
@@ -2091,12 +2101,12 @@ abstract contract Base is Test {
   }
 
   function assertBorrowRateSynced(
-    ILiquidityHub hub,
+    ILiquidityHub liquidityHub,
```
avniculae: I thought we want to migrate towards Hub (drop `liquidity`), did this change?

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
```diff
@@ -6,45 +6,59 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
...
+      abi.encodeWithSelector(IAccessManaged.AccessManagedUnauthorized.selector, caller)
     );
+    vm.prank(caller);
+    spoke1.updateDynamicReserveConfig(reserveId, dynConf);
+
+    vm.expectRevert(
+      abi.encodeWithSelector(IAccessManaged.AccessManagedUnauthorized.selector, caller)
+    );
+    vm.prank(caller);
+    spoke1.updateExistingDynamicReserveConfig(reserveId, _randomConfigKey(), dynConf);
```
avniculae: I would create a separate test for this

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
```diff
@@ -6,45 +6,59 @@ import 'tests/unit/Spoke/SpokeBase.t.sol';
...
-        collateralFactor: 80_00,
-        liquidationBonus: 100_00,
-        liquidationFee: 0
-      })
-    );
+    spoke1.updateDynamicReserveConfig(invalidReserveId, dynConf);
+
+    vm.expectRevert(abi.encodeWithSelector(ISpoke.ReserveNotListed.selector, invalidReserveId));
+    vm.prank(SPOKE_ADMIN);
+    spoke1.updateExistingDynamicReserveConfig(invalidReserveId, _randomConfigKey(), dynConf);
```
avniculae: I would create a separate test for this

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
avniculae, line 1: where are we testing the happy path for `updateExistingDynamicReserveConfig` ?

src/contracts/Spoke.sol
miguelmtzinf: we can do that and do some renaming... I d suggest something verbose like `addDynamicReserveConfig` and `updateDynamicReserveConfig` for old keys.
An alternative would be having one single `updateDynamicReserveConfig(reserveId, key, config)` fn, which adds a new config key if key is max_uint. This simplifies code in Spoke, and we can have the split in `SpokeConfigurator`

src/contracts/Spoke.sol
```diff
@@ -152,6 +153,22 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+  /// @inheritdoc ISpoke
+  function updateExistingDynamicReserveConfig(
+    uint256 reserveId,
+    uint16 configKey,
+    DataTypes.DynamicReserveConfig calldata dynamicConfig
+  ) external restricted {
+    require(reserveId < _reserveCount, ReserveNotListed());
+    _validateDynamicReserveConfig(dynamicConfig);
+    // @dev sufficient check since min liquidationBonus is 100_00
+    require(_dynamicConfig[reserveId][configKey].liquidationBonus != 0, ConfigKeyUninitialized());
```
miguelmtzinf, line 168: should not we do this instead?
```suggestion
    require(configKey <= reserve.dynamicConfig, ConfigKeyUninitialized());
```

src/contracts/Spoke.sol
DhairyaSethi: yea i wanted to the max_uint sentinel approach but was sure it wouldn't be well received haha

tests/Base.t.sol
DhairyaSethi: it's a shadowed var in the test bc we have a global hub, this is cleanup
(we will be removing a global hub bc that written under the assumption spoke -> hub, but now its reserve -> hub)

src/contracts/Spoke.sol
DhairyaSethi: after first pass of uint16.max, every config key is initialized (see _nextUintializedConfigKey helper in tests)

miguelmtzinf: oh, you are completely right

src/contracts/Spoke.sol
miguelmtzinf: I'm fine doing `addDynamicReserveConfig` and `updateDynamicReserveConfig`, calling to the same internal fn (we can do max_uint sentinal thing there if needed)

src/contracts/Spoke.sol
```diff
@@ -14,6 +14,7 @@ import {KeyValueListInMemory} from 'src/libraries/helpers/KeyValueListInMemory.s
 import {DataTypes} from 'src/libraries/types/DataTypes.sol';
 import {LiquidationLogic} from 'src/libraries/logic/LiquidationLogic.sol';
 import {PositionStatus} from 'src/libraries/configuration/PositionStatus.sol';
+import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
```
miguelmtzinf: pls move it up together with OZ ones

src/contracts/Spoke.sol
```diff
@@ -1045,6 +1028,40 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+
+    uint16 nextConfigKey;
+    if (configKey == type(uint256).max) {
+      // @dev overflow is desired, we implicitly invalidate & override stale config
+      unchecked {
+        nextConfigKey = ++_reserves[reserveId].dynamicConfigKey;
+      }
+    } else {
+      nextConfigKey = configKey.toUint16();
+      // @dev sufficient check since min liquidationBonus is 100_00
```
miguelmtzinf: this feels a bit hacky but okay.
on the other hand, we could argue that this is an authorized call so there is no need to check this and should be actually allowed...

src/contracts/SpokeConfigurator.sol
```diff
@@ -175,12 +175,22 @@ contract SpokeConfigurator is Ownable, ISpokeConfigurator {
...
+    DataTypes.DynamicReserveConfig calldata dynamicConfig
+  ) external onlyOwner returns (uint16 configKey) {
+    return ISpoke(spoke).addDynamicReserveConfig(reserveId, dynamicConfig);
+  }
+
   /// @inheritdoc ISpokeConfigurator
   function updateDynamicReserveConfig(
     address spoke,
     uint256 reserveId,
+    uint256 configKey,
```
miguelmtzinf: should we use `uint16` here? how will be our strategy once struct packing is done?

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -132,274 +135,33 @@ contract SpokeConfigTest is SpokeBase {
     vm.expectEmit(address(spoke1));
     emit ISpoke.DynamicReserveConfigUpdated(daiReserveId, dynamicConfigKey, newConfig);
     vm.prank(SPOKE_ADMIN);
-    spoke1.updateDynamicReserveConfig(daiReserveId, newConfig);
+    spoke1.addDynamicReserveConfig(daiReserveId, newConfig);
 
     assertEq(spoke1.getDynamicReserveConfig(daiReserveId), newConfig);
   }
 
-  function test_setUsingAsCollateral_revertsWith_ReserveCannotBeUsedAsCollateral() public {
```
DhairyaSethi: usingAsCollateral tests moved to independent file: `Spoke.SetUsingAsCollateral.t.sol`

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -132,274 +135,33 @@ contract SpokeConfigTest is SpokeBase {
...
-
+    uint256 reserveId = vm.randomUint(spoke1.getReserveCount() + 1, type(uint256).max);
     DataTypes.ReserveConfig memory config;
 
     vm.expectRevert(ISpoke.ReserveNotListed.selector);
     vm.prank(SPOKE_ADMIN);
     spoke1.updateReserveConfig(reserveId, config);
   }
 
-  function test_updateDynamicReserveConfig_revertsWithInvalidLiquidationBonus() public {
```
DhairyaSethi, line 283: moved to `Spoke.DynamicConfig.t.sol`

src/contracts/SpokeConfigurator.sol
miguelmtzinf: kinda weird we use `uin256` here and returning `uint16` in `add`... I understand the reason behind accepting `uint256` as input, but this is a bit confusing

DhairyaSethi: its kinda strategy independent tbh 
we can make this a uint16 input as well, but i made it uint256 bc the internal method needs that (its fine its given given a uint256 from the external input param bc solidity will silently cast up which is ok in the update case) but then the toSafe16() cast is unecessary; so its just for consistency with the internal method and being able to assert that safe cast
can revert to uint16 input no strong opinion

src/contracts/Spoke.sol
DhairyaSethi: allowing updating a future config key is kinda a footgun bc when you reach that configKey in the future (ie with an `addDynamicConfig`) it will be overwritten
we do not provide a way to just increment configKey (and not update data) bc thats an even worse footgun since it assumes data is pre populated 

this is anyway only enforced on the first pass, bc after uint16.max keys are exhausted, then all keys are past AND future keys bc of the cycle

miguelmtzinf: yeah, i'm fine with this

src/contracts/SpokeConfigurator.sol
miguelmtzinf: I think we can change to uint16 input so we can save the safeCast on Spoke internal method.
We would need to safely cast from uint16 to uint256 to uint16. As an alternative, we can calculate the `nextDynamicConfigKey` before calling the internal `_updateDynamicReserveConfig` (think it's cleaner)

Separately, note that this is aligned with using actual data types as inputs instead of general uint256 to safe cast later. Do you have any opinion/preference on this?

DhairyaSethi: > We would need to safely cast from uint16 to uint256 to uint16

uint16 to uint256 is always a safe cast, so solc does it implicitely and that's fine

> we can calculate the nextDynamicConfigKey before calling the internal _updateDynamicReserveConfig (think it's cleaner)

would disagree bc for add we need to update, so its kind of multiple writes
```solidity
function add() {
  uint16 newKey = uint16(uint256(_reserves[reserveId].configKey) + 1) % type(uint16).max));
  _update() // will overwrite _reserves[reserveId]).configKey 
}
```

have kept the first approach for now, lmk

miguelmtzinf: good point, you still wanna keep the safeCast? it's unnecessary

DhairyaSethi: it 100% is, i just don't feel comfortable unsafe casting, but can do it with a comment lmk

miguelmtzinf: how much gas it is? anyway, personal opinion is that we could remove and just have a comment
i d be similar to do unchecked in some parts of the code.

DhairyaSethi: ok sure, added [8552db4](https://github.com/aave/aave-v4/pull/541/commits/8552db424ed69614c761ed84e30d4894a862f546)
agree with unchecked, whenever we add anywhere it will be with a `// SAFETY:` comment

src/contracts/Spoke.sol
DhairyaSethi: removed this abstraction now bc i realised we need two different events for update and add dyn reserve such that backend can know which is configKey is the LATEST for each reserve (needed for refreshDynConf)
now the two fn's are sufficiently different enough that they don't warrant this

PR #542, opened 2025-07-19
rft: premium refresh

DhairyaSethi:
Consolidate all premium update logic in two methods: settle (accrues to realized & removes premiumRestored) & notify (updates new rp for all debt reserves). 

- Removes optional refreshPremium to 0 in withdraw, borrow, & liquidate.
- Removes optional refresh of collateral asset in liquidate.
- Removes updatePremiumDebt in withdraw, borrow, repay, & liquidate.
- Refactors notify to handle premium update on all debt reserves (rm `assetIdToAvoid` logic).
- Few liquidationCall opts.

### Notes:
- There is an unnecessary sload & sub for the reserve in question in repay bc we have set premiumShares, & offset to 0 while settling but that's worth it to keep notify cleaner.
- Notify can still do with a refactor by caching repeatedly read storage values (pos.premiumDrawnShares, pos.premiumOffset) but that is left for a future refactor (when we move to libraries for logic).
- Evaluate isBorrowingAny() early return in notify.


src/contracts/Spoke.sol
```diff
@@ -264,8 +265,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     }
 
     (uint256 newUserRiskPremium, , , , ) = _calculateUserAccountData(msg.sender);
-    _updatePremiumDebt(reserve, userPosition, hub, assetId, msg.sender, newUserRiskPremium);
-    _notifyRiskPremiumUpdate(assetId, msg.sender, newUserRiskPremium);
+    _notifyRiskPremiumUpdate(msg.sender, newUserRiskPremium);
```
DhairyaSethi: note: bc we always settle premium debt **first**, the updatePremiumDebt call was also unnecessary here when **entire** debt was repaid 
this is now skipped in the notify as we set isBorrowing to false before (L265)

to investigate: there are still some places we refresh with 0 deltas

src/contracts/Spoke.sol
```diff
@@ -641,11 +641,12 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     return (amount - premiumDebt, premiumDebt);
   }
 
-  function _accruePremiumDebt(
+  function _settlePremiumDebt(
```
avniculae, line 710: should we call it `_realizePremiumDebt` ?

DhairyaSethi: imo `settle` is more apt since this does realize (ie move accrued premium to realized) and settle (reduce realized by premiumDebtRestored)

avniculae: ah fair, we could call it realize if #505 would be merged

DhairyaSethi: yep agree!

snapshots/Spoke.Operations.json
yan-man, line 1: wow nice really cut down on withdraw here

DhairyaSethi: lmao yeah bc of our redundant refresh's there primarily; here the withdraw snapshot test is done w/o any borrowed asset so it short circuits the hf & notify calls
thinking adding two more withdraw snapshot tests: partial & full on 1 borrow. wdyt

src/contracts/Spoke.sol
```diff
@@ -284,7 +281,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
       vars.baseDebtRestored,
       vars.premiumDebtRestored,
       msg.sender
-    ); // we settle base debt here
+    ); // settle base debt
```
```suggestion
    );
```

src/contracts/Spoke.sol
```diff
@@ -734,62 +724,41 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-    return (userPremiumDrawnShares, userPremiumOffset);
-  }
-
   function _refreshPremiumDebt(
     DataTypes.Reserve storage reserve,
-    address user,
+    ILiquidityHub hub,
     uint256 assetId,
+    uint256 reserveId,
+    address userAddress,
```
```suggestion
    address user,
```

src/contracts/Spoke.sol
```diff
@@ -734,62 +724,41 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-
-    return (userPremiumDrawnShares, userPremiumOffset);
-  }
-
   function _refreshPremiumDebt(
     DataTypes.Reserve storage reserve,
-    address user,
+    ILiquidityHub hub,
     uint256 assetId,
+    uint256 reserveId,
```
miguelmtzinf: is it v bad if we take this from the `reserve` struct?

DhairyaSethi: just an unnecessary warm sload, 100 additional gas 
nbd if you think its making the fn too ugly, we can tackle caching later on

miguelmtzinf: we are passing a ton of params to fns in multiple places. I think it's better to pass less params when possible for better/clean code at the cost of slight gas increase.
This is something we can tackle separately if so, good to go as is

src/contracts/Spoke.sol
miguelmtzinf: are all of them removed/fixed now?

DhairyaSethi: you mean when we're sending 0 deltas?
for that, from what i can tell:
- when riskPremium is 0, thinking about a correct shortcircuit for that bc there could be existing premium debt which needs a refresh on another reserve 
- excess withdraw, borrow; which doesn't change rp 

going to handle this in a separate PR, want to keep this PR to the point bc we're doing things in a rush now

src/contracts/Spoke.sol
DhairyaSethi: yeah lets handle separately?

src/contracts/Spoke.sol
```diff
@@ -236,16 +235,14 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
       positionStatus.setBorrowing(reserveId, true);
```
miguelmtzinf: does not belong to this PR, but can you move down the flag update after accounting updates? to align with repay method

src/contracts/Spoke.sol
miguelmtzinf: - we only send 0 deltas if old and new risk premiums are 0.
- we can check if user risk premium has changed

let's create a ticket with a short desc pls

src/contracts/Spoke.sol
```diff
@@ -739,62 +729,41 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
     int256 premiumDrawnSharesDelta,
     int256 premiumOffsetDelta,
     uint256 realizedPremiumAdded,
     uint256 realizedPremiumTaken
   ) internal {
-    _refresh(
-      reserve,
+    reserve.premiumDrawnShares = _add(reserve.premiumDrawnShares, premiumDrawnSharesDelta);
+    reserve.premiumOffset = _add(reserve.premiumOffset, premiumOffsetDelta);
+    reserve.realizedPremium = reserve.realizedPremium + realizedPremiumAdded - realizedPremiumTaken;
```
miguelmtzinf, line 756:
```suggestion
    reserve.realizedPremium += realizedPremiumAdded - realizedPremiumTaken;
```
any reason we don't do this?

src/contracts/Spoke.sol
```diff
@@ -739,62 +729,41 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+
+    emit RefreshPremiumDebt(
```
miguelmtzinf, line 758: did we flip the order of actions for any reason?
we usually do hub events first, spoke events second

src/contracts/Spoke.sol
DhairyaSethi: https://github.com/aave/aave-v4/issues/556

src/contracts/Spoke.sol
DhairyaSethi: unsafe, if Taken > Added then this underflows

image


since reserve.realizedPremium is an unsigned int

src/contracts/Spoke.sol
DhairyaSethi: so on this, its always been like this, but its a good call, the thing we apply state updates then also after hub call but will need a whole new refactor bc we first update userPosition two different ways (in a settle, and in a notify), and then we call this fn which updates spoke state and calls hub, i think it makes sense it emits after spoke state for now, i think we can note this down in a ticket and tackle in a seaparate rft?

miguelmtzinf: [557](https://github.com/aave/aave-v4/issues/557)

src/contracts/Spoke.sol
miguelmtzinf: https://github.com/aave/aave-v4/issues/558

src/contracts/Spoke.sol
```diff
@@ -1180,35 +1122,14 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-        collateralReserve,
-        userCollateralPosition,
-        collateralReserveHub,
-        vars.collateralAssetId,
-        vars.user,
-        vars.newUserRiskPremium
-      );
-
-      _notifyRiskPremiumUpdate(vars.debtAssetId, vars.user, vars.newUserRiskPremium);
+      _notifyRiskPremiumUpdate(vars.user, vars.newUserRiskPremium);
```
yan-man, line 1132: think this will need to return `userPremiumDrawnShares` and `userPremiumOffset` to be aggregated across liquidation 

currently didnt cause any errors bc the vars just default to 0

DhairyaSethi: we realised this is was unnecessary in liquidation

PR #543, opened 2025-07-19
rft: explicit rounding

DhairyaSethi:
- Merge `WadRayMathExtended` into `WadRayMath` & rm extended variant.
- Merge `PercentageMathExtended` into `PercentMath` & rm extended variant.
- Source rounding directions:
  - Round down on health factor and avgColFactor calculations in calculateUserAccountData.
  - Round up in `getUserDebtInBaseCurrency`
- Tests rounds directions:
  - Round up in `_calcMinimumCollAmount`.

Note:
- A lot of liquidation test and helpers do: x.wadify() / y which can be replaced with x.wadDiv(y) -> Left for dedicated liquidation refactor.

Reference: https://github.com/aave/aave-v4/pull/26

### DIFF
--- a/src/libraries/math/WadRayMath.sol
+++ b/src/libraries/math/WadRayMath.sol
@@ -167,7 +167,7 @@ library WadRayMath {
    * @param a The number
    * @return b (= a * 1e18)
    */
-  function wadify(uint256 a) internal pure returns (uint256 b) {
+  function toWad(uint256 a) internal pure returns (uint256 b) {
     // to avoid overflow, b/WAD == a
     assembly {
       b := mul(a, WAD)
@@ -183,7 +183,7 @@ library WadRayMath {
    * @param a The number in Wad precision
    * @return b (= a / 1e18)
    */
-  function dewadifyDown(uint256 a) internal pure returns (uint256 b) {
+  function fromWadDown(uint256 a) internal pure returns (uint256 b) {
     assembly {
       b := div(a, WAD)
     }
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -1244,8 +1244,7 @@ abstract contract Base is Test {
     IPriceOracle oracle = spoke.oracle();
     uint256 assetId = spoke.getReserve(reserveId).assetId;
     return
-      (amount * oracle.getReservePrice(reserveId).wadify()) /
-      (10 ** hub.getAsset(assetId).decimals);
+      (amount * oracle.getReservePrice(reserveId).toWad()) / (10 ** hub.getAsset(assetId).decimals);
   }
 
   /// @notice Convert 1 asset amount to equivalent amount in another asset.
@@ -1560,7 +1559,7 @@ abstract contract Base is Test {
     uint256 assetPrice,
     uint256 assetUnit
   ) internal pure returns (uint256) {
-    return (amount * assetPrice).wadify() / assetUnit;
+    return (amount * assetPrice).toWad() / assetUnit;
   }
 
   function _convertBaseCurrencyToAmount(
@@ -1584,7 +1583,7 @@ abstract contract Base is Test {
     uint256 assetPrice,
     uint256 assetUnit
   ) internal pure returns (uint256) {
-    return ((baseCurrencyAmount * assetUnit) / assetPrice).dewadifyDown();
+    return ((baseCurrencyAmount * assetUnit) / assetPrice).fromWadDown();
   }
 
   /**
--- a/tests/unit/Spoke/SpokeBase.t.sol
+++ b/tests/unit/Spoke/SpokeBase.t.sol
@@ -441,7 +441,7 @@ contract SpokeBase is Base {
 
     return
       normalizedDebtAmount.wadDivUp(
-        normalizedCollPrice.wadify().percentMulDown(colDynConf.collateralFactor)
+        normalizedCollPrice.toWad().percentMulDown(colDynConf.collateralFactor)
       );
   }
 
@@ -465,8 +465,8 @@ contract SpokeBase is Base {
     uint256 normalizedCollPrice = (collAmount * collPrice).wadDivDown(collAssetUnits);
 
     uint256 maxDebt = (
-      (normalizedCollPrice.wadify().percentMulDown(colDynConf.collateralFactor) /
-        normalizedDebtAmount.wadify())
+      (normalizedCollPrice.toWad().percentMulDown(colDynConf.collateralFactor) /
+        normalizedDebtAmount.toWad())
     );
 
     return maxDebt > 1 ? maxDebt - 1 : maxDebt;
--- a/src/libraries/logic/LiquidationLogic.sol
+++ b/src/libraries/logic/LiquidationLogic.sol
@@ -68,7 +68,7 @@ library LiquidationLogic {
   ) internal pure returns (uint256) {
     // represents the effective value loss from the collateral per unit of debt repaid
     // the greater the penalty, the more debt must be repaid to restore the user's health factor
-    uint256 effectiveLiquidationPenalty = (params.liquidationBonus.wadify())
+    uint256 effectiveLiquidationPenalty = (params.liquidationBonus.toWad())
       .percentMulDown(params.collateralFactor)
       .fromBpsDown();
 
@@ -82,7 +82,7 @@ library LiquidationLogic {
       (((params.totalDebtInBaseCurrency * params.debtAssetUnit) *
         (params.closeFactor - params.healthFactor)) /
         ((params.closeFactor - effectiveLiquidationPenalty + 1) * params.debtAssetPrice))
-        .dewadifyDown();
+        .fromWadDown();
   }
 
   /**
@@ -97,10 +97,10 @@ library LiquidationLogic {
   ) internal pure returns (uint256, uint256, uint256) {
     // convert existing collateral to base currency
     uint256 userCollateralBalanceInBaseCurrency = (params.userCollateralBalance *
-      params.collateralAssetPrice).wadify() / params.collateralAssetUnit;
+      params.collateralAssetPrice).toWad() / params.collateralAssetUnit;
 
     // find collateral in base currency that corresponds to the debt to cover
-    uint256 baseCollateral = (params.actualDebtToLiquidate * params.debtAssetPrice).wadify() /
+    uint256 baseCollateral = (params.actualDebtToLiquidate * params.debtAssetPrice).toWad() /
       params.debtAssetUnit;
 
     // account for additional collateral required due to liquidation bonus
@@ -111,12 +111,12 @@ library LiquidationLogic {
     if (maxCollateralToLiquidate >= userCollateralBalanceInBaseCurrency) {
       collateralAmount = params.userCollateralBalance;
       debtAmountNeeded = ((params.debtAssetUnit * userCollateralBalanceInBaseCurrency)
-        .percentDivDown(params.liquidationBonus) / params.debtAssetPrice).dewadifyDown();
+        .percentDivDown(params.liquidationBonus) / params.debtAssetPrice).fromWadDown();
     } else {
       // add 1 to round collateral amount up, ensuring HF is always <= close factor
       collateralAmount =
         ((maxCollateralToLiquidate * params.collateralAssetUnit) / params.collateralAssetPrice)
-          .dewadifyDown() +
+          .fromWadDown() +
         1;
       debtAmountNeeded = params.actualDebtToLiquidate;
     }
--- a/tests/mocks/WadRayMathWrapper.sol
+++ b/tests/mocks/WadRayMathWrapper.sol
@@ -48,12 +48,12 @@ contract WadRayMathWrapper {
     return WadRayMath.rayDivUp(a, b);
   }
 
-  function wadify(uint256 a) public pure returns (uint256) {
-    return WadRayMath.wadify(a);
+  function toWad(uint256 a) public pure returns (uint256) {
+    return WadRayMath.toWad(a);
   }
 
-  function dewadifyDown(uint256 a) public pure returns (uint256) {
-    return WadRayMath.dewadifyDown(a);
+  function fromWadDown(uint256 a) public pure returns (uint256) {
+    return WadRayMath.fromWadDown(a);
   }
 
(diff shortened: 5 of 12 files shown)


src/contracts/Spoke.sol
```diff
@@ -473,7 +470,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
 
   function getReserveRiskPremium(uint256 reserveId) external view returns (uint256) {
     DataTypes.Reserve storage reserve = _reserves[reserveId];
-    return reserve.premiumDrawnShares.rayDiv(reserve.baseDrawnShares); // trailing
+    return reserve.premiumDrawnShares.rayDivDown(reserve.baseDrawnShares); // trailing
```
miguelmtzinf, line 482: this is because we are doing `rayMulUp` when calculating `premiumDrawnShares` right?

DhairyaSethi: yes

src/libraries/math/PercentageMath.sol
```diff
@@ -6,56 +6,97 @@ pragma solidity ^0.8.0;
...
- * @dev Operations are rounded. If a value is >=.5, will be rounded up, otherwise rounded down.
  */
 library PercentageMath {
   // Maximum percentage factor (100.00%)
   uint256 internal constant PERCENTAGE_FACTOR = 1e4;
 
-  // Half percentage factor (50.00%)
-  uint256 internal constant HALF_PERCENTAGE_FACTOR = 0.5e4;
+  /**
+   * @notice Executes a percentage multiplication, rounded down
```
miguelmtzinf: i think we can make all of them `@dev`

src/libraries/math/PercentageMath.sol
```diff
@@ -6,56 +6,97 @@ pragma solidity ^0.8.0;
...
-      ) {
+  function percentMulUp(uint256 value, uint256 percentage) internal pure returns (uint256 result) {
+    // to avoid overflow, value <= type(uint256).max / percentage
+    assembly ('memory-safe') {
+      if iszero(or(iszero(percentage), iszero(gt(value, div(not(0), percentage))))) {
         revert(0, 0)
       }
-
-      result := div(add(mul(value, percentage), HALF_PERCENTAGE_FACTOR), PERCENTAGE_FACTOR)
+      result := mul(value, percentage)
```
miguelmtzinf, line 48:
```suggestion

      result := mul(value, percentage)
```

src/libraries/math/PercentageMath.sol
```diff
@@ -6,56 +6,97 @@ pragma solidity ^0.8.0;
...
-    // to avoid overflow, value <= (type(uint256).max - HALF_PERCENTAGE_FACTOR) / percentage
-    assembly {
-      if iszero(
-        or(
-          iszero(percentage),
-          iszero(gt(value, div(sub(not(0), HALF_PERCENTAGE_FACTOR), percentage)))
-        )
-      ) {
+  function percentMulUp(uint256 value, uint256 percentage) internal pure returns (uint256 result) {
+    // to avoid overflow, value <= type(uint256).max / percentage
```
miguelmtzinf, line 43: shouldn't we do `type(uint256).max - 1`? to account for rounding up (+1)

src/libraries/math/PercentageMath.sol
```diff
@@ -6,56 +6,97 @@ pragma solidity ^0.8.0;
...
+   * @param percentage The percentage of the value to be calculated (in BPS)
+   * @return result value percentDiv percentage
+   */
+  function percentDivUp(uint256 value, uint256 percentage) internal pure returns (uint256 result) {
+    // to avoid overflow, value <= type(uint256).max / PERCENTAGE_FACTOR
+    assembly ('memory-safe') {
+      if or(iszero(percentage), iszero(iszero(gt(value, div(not(0), PERCENTAGE_FACTOR))))) {
+        revert(0, 0)
+      }
+      result := mul(value, PERCENTAGE_FACTOR)
```
miguelmtzinf, line 89:
```suggestion

      result := mul(value, PERCENTAGE_FACTOR)
```

src/libraries/math/PercentageMath.sol
miguelmtzinf: same in `percentDivUp`

src/libraries/math/WadRayMath.sol
```diff
@@ -1,165 +1,200 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
-
 /**
- * @title WadRayMath library
+ * @title WadRayMath library, with explicit rounding
```
miguelmtzinf: i d move this to the notice or dev instad. same in Percent lib

src/libraries/math/WadRayMath.sol
```diff
@@ -1,165 +1,200 @@
...
   /**
-   * @notice Truncates number from Ray precision
-   * @param a The number in Ray precision
-   * @return b (= a / 1e27)
+   * @notice Truncates number from Wad precision
+   * @param a The number in Wad precision
+   * @return b (= a / 1e18)
    */
-  function derayify(uint256 a) internal pure returns (uint256 b) {
+  function dewadifyDown(uint256 a) internal pure returns (uint256 b) {
```
miguelmtzinf: this is completely off from `fromBps`.
wondering if we should change any of them to make them consistent.

src/libraries/math/PercentageMath.sol
miguelmtzinf, line 1: Wondering if we should put `PercentageMath` and `WadRayMath` together, since they are used together the most of time.

tests/Base.t.sol
```diff
@@ -1675,9 +1675,7 @@ abstract contract Base is Test {
     ) = spoke.getUserAccountData(user);
 
     requiredDebtInBaseCurrency =
-      totalCollateralBase.percentMul(currentAvgCollateralFactor.dewadifyDown() + 1).wadDivUp(
-        desiredHf
-      ) -
+      (totalCollateralBase.percentMulUp(currentAvgCollateralFactor + 1) / desiredHf) -
```
miguelmtzinf, line 1624: i'm okay with this change, but not sure why we sum +1 at the end instead of adding it to avgCf

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -424,6 +423,8 @@ contract SpokeBase is Base {
     uint256 debtReserveId,
     uint256 debtAmount
   ) internal view returns (uint256) {
+    if (debtAmount == 0) return 1;
```
miguelmtzinf, line 427: why

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -434,12 +435,13 @@ contract SpokeBase is Base {
     uint256 debtAssetUnits = 10 ** hub.getAsset(debtData.assetId).decimals;
     uint256 debtPrice = oracle.getReservePrice(debtReserveId);
 
-    uint256 normalizedDebtAmount = (debtAmount * debtPrice).wadify() / debtAssetUnits;
-    uint256 normalizedCollPrice = collPrice.wadify() / collAssetUnits;
+    uint256 normalizedDebtAmount = (debtAmount * debtPrice).wadDivDown(debtAssetUnits);
```
miguelmtzinf, line 439: shouldn't we overestimate here?

miguelmtzinf: oh, we prefer to underestimate since it's numerator later?

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -458,11 +460,11 @@ contract SpokeBase is Base {
...
     uint256 debtPrice = oracle.getReservePrice(debtReserveId);
 
-    uint256 normalizedDebtAmount = (debtPrice).wadify() / debtAssetUnits;
-    uint256 normalizedCollPrice = (collAmount * collPrice).wadify() / collAssetUnits;
+    uint256 normalizedDebtAmount = (debtPrice).wadDivDown(debtAssetUnits);
+    uint256 normalizedCollPrice = (collAmount * collPrice).wadDivDown(collAssetUnits);
 
     uint256 maxDebt = (
-      (normalizedCollPrice.wadify().percentMul(colDynConf.collateralFactor) /
+      (normalizedCollPrice.wadify().percentMulDown(colDynConf.collateralFactor) /
```
miguelmtzinf: not following this operation tbh. is `normalizedCollPrice` actually a price?

src/contracts/Spoke.sol
miguelmtzinf: wondering if makes sense to have a couple of tests showcasing this

src/contracts/Spoke.sol
```diff
@@ -614,29 +611,20 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-      InvalidCollateralFactor()
-    ); // max 100.00%
-    require(
-      config.liquidationBonus >= PercentageMathExtended.PERCENTAGE_FACTOR,
-      InvalidLiquidationBonus()
-    ); // min 100.00%
-    require(
-      config.collateralFactor.percentMulUp(config.liquidationBonus) <=
-        PercentageMathExtended.PERCENTAGE_FACTOR,
+      config.liquidationBonus.percentMulUp(config.collateralFactor) <=
```
avniculae, line 646: nit: why did you reverse the order here? I would keep the original order given the error name

src/contracts/Spoke.sol
```diff
@@ -857,12 +845,12 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     // strip BPS factor from result, because running avgCollateralFactor sum has been scaled by collateralFactor (in BPS) above
```
avniculae, line 871: comment above `(avgCollateralFactor / totalCollateral) * totalCollateral can be simplified to avgCollateralFactor` is confusing, what is it trying to say?

src/libraries/math/PercentageMath.sol
avniculae: I think this is a good idea, do you have any name suggestions? We can call it `FixedPointMath`/ `DecimalMath`

src/libraries/math/PercentageMath.sol
DhairyaSethi: nope this is just a check on intermediary overflow, ie value * percentage, the following division + ceil rounding can make it overflow since its always less than the origin mul (ie value * percentage) bc of the division

src/libraries/math/PercentageMath.sol
DhairyaSethi: i think they're separate enough tbh, we just have the conversions between the two which are a bit scattered 

> We can call it FixedPointMath/ DecimalMath

yea the two are subset of fixed point math, we have oz dep for it where we use the mulDiv for shares math but here since we don't allow overflow (ie take calculation temporarily into 512 bit territory unlike share math where we do) it'll be very inconsistent to use this naming

tests/Base.t.sol
DhairyaSethi: @yan-man

tests/unit/Spoke/SpokeBase.t.sol
DhairyaSethi: bc some fuzz tests allow bounds of 0 debt amount and this tests expects to return minimum col amount of 1 based on previous usage, that goes away bc of the explicit rounding addition on debtAssetUnits being rounded up instead of down from before, hence this is needed

it's sister method, ie calcMinDebtAmount achieves this behaviour with a ternary cond in the return (as before)

tests/unit/Spoke/SpokeBase.t.sol
DhairyaSethi: yep correct

tests/unit/Spoke/SpokeBase.t.sol
DhairyaSethi: its actually normalizedCollAmount, so collAmount * collPrice * WAD / collUnits, can rename
the extra wadify() here & in the denominator is for extra precision i imagine (bc when i remove it (since it cancels out) a lot tests fail with hf below 1), right @yan-man?

src/contracts/Spoke.sol
DhairyaSethi: bc coll factor is a uint16 and had to cast lib to it, it's a temp measure anyway before packing so nbd

src/contracts/Spoke.sol
DhairyaSethi: the multiplier and the denominator in the multiplicand (ie `totalCollateral`) cancel out

src/libraries/math/WadRayMath.sol
DhairyaSethi: agree, made it consistent with fromBpsDown 
[033211b](https://github.com/aave/aave-v4/pull/543/commits/033211b5db78a0721fdec81ce656610b24503852)

src/libraries/math/WadRayMath.sol
DhairyaSethi: [033211b](https://github.com/aave/aave-v4/pull/543/commits/033211b5db78a0721fdec81ce656610b24503852)

src/libraries/math/PercentageMath.sol
DhairyaSethi: [033211b](https://github.com/aave/aave-v4/pull/543/commits/033211b5db78a0721fdec81ce656610b24503852)

src/libraries/math/PercentageMath.sol
DhairyaSethi: [033211b](https://github.com/aave/aave-v4/pull/543/commits/033211b5db78a0721fdec81ce656610b24503852)

src/libraries/math/PercentageMath.sol
DhairyaSethi: [033211b](https://github.com/aave/aave-v4/pull/543/commits/033211b5db78a0721fdec81ce656610b24503852)

src/contracts/Spoke.sol
DhairyaSethi: noted, not too concerned about this since this is an external getter not used in the system anywhere which we might move to data provider contract instead

PR #549, opened 2025-07-21
feat: emit `AssetUpdated` event

### LINKED ISSUE #538: Emission of borrow rate after updating on Hub.updateBorrowRate
Aave V3 emits information about rates (supply/variableDebt/staticDebt) and indexes (supply/debt) after updating the rates.
https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/protocol/libraries/logic/ReserveLogic.sol#L165-L167

We can consider emitting an event in `updateBorrowRate` such as:
`AssetUpdated(uint256 indexed assetId, uint256 drawnIndex, uint256 baseBorrowRate, uint256 latestUpdateTimestamp);`

We could remove the `DrawnIndexUpdate` event emitted in `AssetLogic.accrue()` function and emit everything at once.

Is there anything else we must add? 

Note that `latestUpdateTimestamp` can be removed later down the line, once all networks support the new RPC call where block.timestamp is also added in `eth_getLogs` rpc call.

yan-man:
- emit `AssetUpdated` event after `updateBorrowRate`
- remove `DrawnIndexUpdate` event

closes #538

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -50,8 +50,6 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     require(irStrategy != address(0), InvalidIrStrategy());
 
     uint256 assetId = _assetCount++;
-    uint256 baseDebtIndex = WadRayMathExtended.RAY;
-    uint256 lastUpdateTimestamp = block.timestamp;
     IAssetInterestRateStrategy(irStrategy).setInterestRateData(assetId, data);
     uint256 baseBorrowRate = IAssetInterestRateStrategy(irStrategy).calculateInterestRate({
       assetId: assetId,
@@ -59,6 +57,9 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
       baseDebt: 0,
       premiumDebt: 0
     });
+
+    uint256 baseDebtIndex = WadRayMathExtended.RAY;
+    uint256 lastUpdateTimestamp = block.timestamp;
     DataTypes.AssetConfig memory config = DataTypes.AssetConfig({
       active: true,
       paused: false,


src/interfaces/ILiquidityHub.sol
```diff
@@ -19,7 +19,12 @@ interface ILiquidityHub is IAccessManaged {
     address indexed spoke,
     DataTypes.SpokeConfig config
   );
-  event DrawnIndexUpdate(uint256 indexed assetId, uint256 drawnIndex, uint256 lastUpdateTimestamp);
+  event AssetUpdated(
```
miguelmtzinf, line 22: Not sold on the name, open to alternatives

src/libraries/logic/AssetLogic.sol
```diff
@@ -106,13 +106,22 @@ library AssetLogic {
...
       .calculateInterestRate({
         assetId: assetId,
         availableLiquidity: asset.availableLiquidity,
         baseDebt: asset.baseDebt(),
         premiumDebt: asset.premiumDebt()
       });
+    asset.baseBorrowRate = newBorrowRate;
+
+    // asset accrual should have already occurred
+    emit ILiquidityHub.AssetUpdated(
```
miguelmtzinf, line 119: should we emit on `addAsset`? thinking that we d never emit base variable borrow rate as `newBorrowRate` and 1 ray as `baseDebtIndex`

src/interfaces/ILiquidityHub.sol
yan-man: I'm in favor of the name given it is more generic than a borrow rate we're trying to emit, and includes index and timestamp as well (so makes sense as an asset update). But maybe gets confusing bc we also have AssetConfigUpdated, which specifies which part of the asset is being updated. 

But don't feel too strongly

src/libraries/logic/AssetLogic.sol
yan-man: imo addAsset instantiates rather than updates the asset, and already we emit `AssetAdded` and `AssetConfigUpdated` there

miguelmtzinf: thing is that we are missing the first value of borrow rate and index.

yan-man: as you mentioned above index would just be 1 ray, and base borrow rate is just ir strat baseVariableBorrowRate so backend can retrieve this when asset is added?

miguelmtzinf: backend will need to call IR strategy to know the borrow rate this asset starts out with

miguelmtzinf: @avniculae raised a similar point about not emitting the borrow rate when listing.

src/contracts/LiquidityHub.sol
```diff
@@ -68,13 +74,14 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
       premiumOffset: 0,
       realizedPremium: 0,
       baseDebtIndex: WadRayMathExtended.RAY,
-      baseBorrowRate: 0,
+      baseBorrowRate: baseBorrowRate,
```
yan-man, line 81: adding this change has some downstream effects: `calculateInterestRate` can only be called when ir strat has been initialized (`rateData.optimalUsageRatio != 0`), so prior to `addAsset`, we must make sure that `setInterestRateData` is set already (see all accompanying changes in the test setups). 

edit: motivation was mainly to have the initial borrow rate that the asset starts with be available via event

CheyenneAtapour: I didn't necessarily see an explicit motivation in the ticket for adding this change, but I do remember we talked about this at some point. Is there either another ticket to link on this PR, or can you give a brief explanation of the motivation for this? I see it that this could still be 0, and the event could emit the same?

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: These are all things related to the asset's borrow rate. Immediately I would think this relates to an asset's config without additional clarification.

So what about `AssetRateUpdated`?

src/contracts/LiquidityHub.sol
CheyenneAtapour: Ah I see the convo is here: https://github.com/aave/aave-v4/pull/549/files#r2220133066

Well, with no supply either, can't we argue there shouldn't be a non-zero borrow rate at listing? Because even if it was nonzero, it's impossible to be used

tests/Base.t.sol
```diff
@@ -353,16 +353,7 @@ abstract contract Base is Test {
     });
 
     // Add all assets to the Liquidity Hub
```
CheyenneAtapour: Maybe can just move this comment down to when assets are being added. Then can just change next comment to "initialize interest rate strategy data". Can leave next comment too, I know point of this pr is that the timing of initialization of ir data is important now, so nbd either way

tests/Base.t.sol
```diff
@@ -353,16 +353,7 @@ abstract contract Base is Test {
...
-    // add WETH
-    hub.addAsset(
-      address(tokenList.weth),
-      tokenList.weth.decimals(),
-      address(treasurySpoke),
-      address(irStrategy)
-    );
-    hub.addSpoke(wethAssetId, address(treasurySpoke), spokeConfig);
-    vm.stopPrank();
+    // initialize interest rate strategy data first
```
CheyenneAtapour: imo feels cleaner to initialize ir data for all assets (similar to hub 2 fixture) before adding the assets, so we don't need to keep changing prank, but nbd, or can tackle separately after discussion in a refactoring pr

tests/Base.t.sol
```diff
@@ -939,23 +957,8 @@ abstract contract Base is Test {
...
-    );
-    vm.startPrank(address(hub2));
-    hub2IrStrategy.setInterestRateData(wethAssetId, encodedIrData);
-    hub2IrStrategy.setInterestRateData(usdxAssetId, encodedIrData);
-    hub2IrStrategy.setInterestRateData(daiAssetId, encodedIrData);
-    hub2IrStrategy.setInterestRateData(wbtcAssetId, encodedIrData);
-    vm.stopPrank();
-
     setUpRoles(hub2, spoke1, accessManager2);
+    vm.stopPrank();
```
CheyenneAtapour: I thought it would complain about overriding a prank in the previous function. To me feels safer to stop prank before calling `setUpRoles` anyway, and feels like better practice imo, but I guess it's not having impact, so nbd

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
```diff
@@ -119,15 +120,29 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
...
     uint256 expectedAssetId = hub.getAssetCount();
+
+    address interestRateStrategy = address(new AssetInterestRateStrategy(address(hub)));
+    vm.prank(address(hub));
+    IAssetInterestRateStrategy(interestRateStrategy).setInterestRateData(
+      expectedAssetId,
+      abi.encode(
+        IAssetInterestRateStrategy.InterestRateData({
+          optimalUsageRatio: 90_00, // 90.00%
+          baseVariableBorrowRate: baseVariableBorrowRate, // 5.00%
```
CheyenneAtapour: this is being fuzzed now, right? So not always 5%

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
```diff
@@ -119,15 +120,29 @@ contract LiquidityHubConfigTest is LiquidityHubBase {
...
-    address interestRateStrategy
+    uint32 baseVariableBorrowRate
   ) public {
     assumeUnusedAddress(underlying);
     assumeNotZeroAddress(feeReceiver);
-    assumeNotZeroAddress(interestRateStrategy);
 
     decimals = uint8(bound(decimals, 0, hub.MAX_ALLOWED_ASSET_DECIMALS()));
-
+    baseVariableBorrowRate = uint32(bound(baseVariableBorrowRate, 0, MAX_BORROW_RATE / 2)); // 1000%
```
CheyenneAtapour: isn't max borrow rate 1000, so div by 2 would be 500%?

src/contracts/LiquidityHub.sol
miguelmtzinf: it's actually not a bad idea to make sure the IR data is ready for the listing, so not worried about enforcing the order of actions (first `setInterestRateData`, second `addAsset`) to avoid misconfigurations.

src/contracts/LiquidityHub.sol
```diff
@@ -68,13 +74,14 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
       lastUpdateTimestamp: block.timestamp,
       config: config
     });
 
     emit AssetAdded(assetId, underlying, decimals);
     emit AssetConfigUpdated(assetId, config);
+    emit AssetUpdated(assetId, WadRayMathExtended.RAY, baseBorrowRate, block.timestamp);
```
miguelmtzinf: i d declare a variable for the index, so it's clearer (not concerned about gas costs)

src/interfaces/ILiquidityHub.sol
miguelmtzinf: i d prefer to refer to the index if we change the current name tbh

src/contracts/LiquidityHub.sol
yan-man: @CheyenneAtapour i think just a matter of preference to have the initial borrow rate be available to backend without needing to call the IR strategy, 0 also works (as is)

tests/Base.t.sol
yan-man: yeah was not sure whether it'd be preferable to have this enforced IR strat prior to addAsset but now I'll go back and clean this up

edit: https://github.com/aave/aave-v4/pull/549/commits/9c8ae6de03185a611e6bc90e8ac7d62fc7e5b2cd

src/contracts/LiquidityHub.sol
yan-man: also added a var for `lastUpdateTimestamp`, lmk if you think that's overkill 

https://github.com/aave/aave-v4/pull/549/commits/9c8ae6de03185a611e6bc90e8ac7d62fc7e5b2cd

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
yan-man: instead just bounded it based on the max rate considering the rate1/rate2 vals

https://github.com/aave/aave-v4/pull/549/commits/9c8ae6de03185a611e6bc90e8ac7d62fc7e5b2cd

tests/unit/LiquidityHub/LiquidityHub.Config.t.sol
yan-man: https://github.com/aave/aave-v4/pull/549/commits/9c8ae6de03185a611e6bc90e8ac7d62fc7e5b2cd

tests/Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/549/commits/9c8ae6de03185a611e6bc90e8ac7d62fc7e5b2cd

tests/Base.t.sol
yan-man: https://github.com/aave/aave-v4/pull/549/commits/9c8ae6de03185a611e6bc90e8ac7d62fc7e5b2cd

src/contracts/LiquidityHub.sol
miguelmtzinf: no strong opinion, we will remove!

src/interfaces/ILiquidityHub.sol
```diff
@@ -19,7 +19,12 @@ interface ILiquidityHub is IAccessManaged {
...
+    uint256 indexed assetId,
+    uint256 drawnIndex,
+    uint256 baseBorrowRate,
+    uint256 latestUpdateTimestamp // TODO: rm when block.timestamp is added in eth_getLogs rpc call
```
miguelmtzinf: remove this TODO pls. Create a ticket from now on instead
https://github.com/aave/aave-v4/issues/554

tests/Base.t.sol
```diff
@@ -352,6 +352,25 @@ abstract contract Base is Test {
...
+      IAssetInterestRateStrategy.InterestRateData({
+        optimalUsageRatio: 90_00, // 90.00%
+        baseVariableBorrowRate: 5_00, // 5.00%
+        variableRateSlope1: 5_00, // 5.00%
+        variableRateSlope2: 5_00 // 5.00%
+      })
+    );
+
+    // initialize interest rate strategy data first
+    vm.startPrank(address(hub));
```
DhairyaSethi: lets rm existing hub pranks as well 😭

src/contracts/LiquidityHub.sol
avniculae, line 1: as mentioned in the sync, we have a circular dependency bc we need to set IR params before registering an asset, but this must happen through the hub (which will access the asset storage, which is empty).

one solution is to take in the IR params in the addAsset fn

yan-man: @miguelmtzinf will impl the solution w IR params in `addAsset` and can reevaluate

https://github.com/aave/aave-v4/pull/549/commits/dc35039dde26a88f9b80a3cb470a76cd2054e9de

tests/Base.t.sol
yan-man: actually good catch @avniculae we were pranking hub in test setups but should have been going thru configurator anyway

https://github.com/aave/aave-v4/pull/549/commits/dc35039dde26a88f9b80a3cb470a76cd2054e9de

src/contracts/LiquidityHub.sol
miguelmtzinf: not following @avniculae . The state is zero at the beginning so it's fine to call IR with zeroed values.
even if you access asset storage, you will pass zero

miguelmtzinf: doing the `setInterestRateData` inside `addAsset` feels it belongs more to `HubConfigurator` than `Hub` imho.

yan-man: before the fix:
- in hub, when calling `addAsset` we want to calculate the `baseBorrowRate` via `calculateInterestRate`
- however, `calculateInterestRate` depends on `_interestRateData` being set in `AssetInterestRateStrategy` for a given `assetId`
- `_interestRateData` must be set via `setInterestRateData`, only by hub's `setInterestRateData` - which also requires asset to exist already 
- previously it was working in tests bc we were pranking the hub address to call `setInterestRateData` on `AssetInterestRateStrategy` directly, but this is not correct in practice (as we should instead call hub's `setInterestRateData` that depends on asset already existing)
- thus we dont have a way of setting IR prior to asset's existence

src/contracts/LiquidityHub.sol
```diff
@@ -41,14 +41,24 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
     uint256 assetId = _assetCount++;
+    uint256 baseDebtIndex = WadRayMathExtended.RAY;
+    uint256 lastUpdateTimestamp = block.timestamp;
+    IAssetInterestRateStrategy(irStrategy).setInterestRateData(assetId, data);
+    uint256 baseBorrowRate = IAssetInterestRateStrategy(irStrategy).calculateInterestRate({
+      assetId: assetId,
+      availableLiquidity: 0,
+      baseDebt: 0,
+      premiumDebt: 0
+    });
```
```suggestion
    IAssetInterestRateStrategy(irStrategy).setInterestRateData(assetId, data);
    uint256 baseBorrowRate = IAssetInterestRateStrategy(irStrategy).calculateInterestRate({
      assetId: assetId,
      availableLiquidity: 0,
      baseDebt: 0,
      premiumDebt: 0
    });

    uint256 baseDebtIndex = WadRayMathExtended.RAY;
    uint256 lastUpdateTimestamp = block.timestamp;
```

PR #550, opened 2025-07-21
feat: withdraw gas snapshot with active borrows

DhairyaSethi:
to ideally merge before #542

### DIFF
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -4,6 +4,7 @@
   "liquidationCall: partial": "664563",
   "repay: full": "250783",
   "repay: partial": "336183",
+  "supply + enable usingAsCollateral": "241074",
   "supply: 0 borrows, collateral disabled": "176592",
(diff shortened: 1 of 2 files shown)


snapshots/Spoke.Operations.json
```diff
@@ -4,14 +4,15 @@
...
-  "supply: 0 debt, collateralDisabled": "176188",
-  "supply: 0 debt, collateralEnabled": "176164",
-  "supply: 1 debt": "176212",
-  "supply: 2 debt": "176223",
-  "supply: 3 debt": "176176",
+  "supply: 0 borrow, collateralDisabled": "176188",
+  "supply: 0 borrow, collateralEnabled": "176164",
+  "supply: 1 borrow": "176212",
+  "supply: 2 borrows": "176223",
+  "supply: 3 borrows": "176176",
```
DhairyaSethi: should rm these three tbh, since supply does not touch debt data

miguelmtzinf: do you wanna remove now?

PR #553, opened 2025-07-22
feat: Refactor flags

### LINKED ISSUE #453: Revisit hub and spoke flags
Hub has the following flags per asset: `active`, `paused`, `frozen`
- `active` and `paused` has identical functionality (they are perfect complements): `add`, `remove`, `draw` and `restore` make sure that `active` is `true` and `paused` is `false`.
- `frozen` must be `false` for `add` and `draw` actions, but has no effect on `remove` and `restore`.
- Note: the flags have no effect on `refreshPremiumDebt` and `settlePremiumDebt`.

Spoke has the following flags per reserve: `active`, `paused`, `frozen`, `borrowable`, `collateral`
- `active` and `paused` has identical functionality (they are perfect complements): `supply`, `withdraw`, `borrow`, `repay`, `liquidationCall` and `setUsingAsCollateral` make sure that `active` is `true` and `paused` is `false`.
- `frozen` must be `false` for `supply`, `borrow` and `setUsingAsCollateral`, but has no effect on `withdraw`, `repay` and `liquidationCall`
- `borrowable` must be `false` for `borrow`, has no other effect
- `collateral` must be `false` for `setUsingAsCollateral`, has no other effect

[Since Aave v3.1](https://governance.aave.com/t/bgd-aave-v3-1-and-aave-origin/17305#h-7-ltv0-on-freezing-11), LTV is set to 0 whenever an asset is frozen. This makes the collateral power of the frozen asset 0, until it is unfrozen. However, this does not affect the user's HF given that liquidation threshold stays the same. In V4, we could achieve a similar result (but not exactly the same) by setting `collateralFactor` to 0 whenever an asset is frozen: with dynamic risk configurations, this will take effect only on a user action, or AFTER a liquidation. However, once this takes effect, user's HF will be affected (in contrast, user's HF would have not been affected in V3 even after an user action). Alternatively, a more complex solution would require changes to the spoke, where the collateralFactor would be considered 0 if asset is frozen and action is anything but `liquidationCall`.

Outstanding points:
- what should be the behaviour of the spoke when a reserve is frozen?
- shall we remove the `active` flag, given it's a perfect complement of `paused`?

avniculae:
Close #453 

- Hub
  - AssetConfig: no flags
  - SpokeConfig: active, caps
  - Add EnumerableSet in storage to store all registered spokes for each asset. Spokes are not removed from sets.
- Spoke
  - ReserveConfig: paused, frozen, borrowable
- HubConfigurator
  - adds additional functions to freeze/pause assets, as well as to update spoke configs
  - keeps fee receiver active flag constant
- SpokeConfigurator
  - adds additional functions to freeze/pause all reserves

<hr>

Questions
- How do we offboard a reserve from the Spoke?
  - reserveConfig.frozen = true
  - increase reserve premium to force users to pay debt
- How do we offboard an asset's Spoke from the Hub?
  - set caps to 0 for that asset's Spoke Config
  - **There's no way to force spoke's users to pay debt faster only through the Hub**
- How do we offboard a Spoke from the Hub?
  - iterate through all assets where the spoke is registered, and set caps to 0
  - **There's no way to force spoke's users to pay debt faster only through the Hub**
- How do we offboard an asset from the Hub?
  - iterate through all spokes registered for the asset and set caps to 0
  - increase borrow rate by changing IR params to force users to pay debt
- How to pause an asset in the Hub in case of emergency?
  - set asset.active = false

### DIFF
--- a/snapshots/Hub.Operations.json
+++ b/snapshots/Hub.Operations.json
@@ -1,6 +1,6 @@
 {
   "add": "167675",
-  "draw": "140871",
+  "draw": "140887",
   "remove: full": "82662",
   "remove: partial": "98948",
   "restore: full": "116093",
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -1,5 +1,5 @@
 {
-  "borrow": "474034",
+  "borrow": "474050",
   "liquidationCall: full": "422436",
   "liquidationCall: partial": "605911",
   "repay: full": "220485",


src/contracts/LiquidityHub.sol
```diff
@@ -26,6 +26,9 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   mapping(uint256 assetId => DataTypes.Asset assetData) internal _assets;
   mapping(uint256 assetId => mapping(address spokeAddress => DataTypes.SpokeData spokeData))
     internal _spokes;
+  mapping(uint256 assetId => address[] spokeAddresses) internal _spokesList;
+  mapping(uint256 assetId => mapping(address spokeAddress => bool isListed))
+    internal _isSpokeListed;
```
DhairyaSethi: we can use oz enumerable set for this instead, internally it also stores an array & reverse map but provides convenience helpers like checking for dup additions, rming etc but ok with how it is if we won't need most of the abstracions

src/contracts/LiquidityHub.sol
```diff
@@ -115,15 +115,18 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+
+    _isSpokeListed[assetId][spoke] = true;
+    _spokesList[assetId].push(spoke);
 
     _spokes[assetId][spoke] = DataTypes.SpokeData({
       suppliedShares: 0,
       baseDrawnShares: 0,
       premiumDrawnShares: 0,
       premiumOffset: 0,
       realizedPremium: 0,
```
miguelmtzinf: we can probably skip this and just set the config

src/contracts/LiquidityHub.sol
```diff
@@ -494,9 +508,6 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     require(spoke.config.active, SpokeNotActive());
     require(amount != 0, InvalidAddAmount());
     require(from != address(this), InvalidAddFromHub());
```
```suggestion
    require(amount != 0, InvalidAddAmount());
    require(from != address(this), InvalidAddFromHub());
    require(spoke.config.active, SpokeNotActive());
```
let's keep checks on spoke flags together. apply to the rest pls

src/contracts/LiquidityHub.sol
```diff
@@ -511,8 +522,6 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   ) internal view {
     require(spoke.config.active, SpokeNotActive());
     require(amount != 0, InvalidRemoveAmount());
-    require(asset.config.active, AssetNotActive());
-    require(!asset.config.paused, AssetPaused());
     uint256 withdrawable = asset.toSuppliedAssetsDown(spoke.suppliedShares);
     require(amount <= withdrawable, SuppliedAmountExceeded(withdrawable));
```
miguelmtzinf, line 521:
```suggestion
    require(amount <= asset.toSuppliedAssetsDown(spoke.suppliedShares), SuppliedAmountExceeded(withdrawable));
```

src/contracts/LiquidityHub.sol
```diff
@@ -526,9 +535,6 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   ) internal view {
     require(spoke.config.active, SpokeNotActive());
     require(amount > 0, InvalidDrawAmount());
```
miguelmtzinf, line 532: can we choose to do != or > consistently?

miguelmtzinf: i d go with >, specially because we also have `int256` in the code (although i do not think we compare them)

src/contracts/LiquidityHub.sol
```diff
@@ -526,9 +535,6 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-    require(asset.config.active, AssetNotActive());
-    require(!asset.config.paused, AssetPaused());
-    require(!asset.config.frozen, AssetFrozen());
     require(
       drawCap == type(uint256).max || amount + asset.totalDebt() <= drawCap,
```
miguelmtzinf: damn, bug here
this is spoke-specific and we are checking against total debt

src/contracts/LiquidityHub.sol
```diff
@@ -494,9 +508,6 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     require(spoke.config.active, SpokeNotActive());
     require(amount != 0, InvalidAddAmount());
     require(from != address(this), InvalidAddFromHub());
-    require(asset.config.active, AssetNotActive());
-    require(!asset.config.paused, AssetPaused());
-    require(!asset.config.frozen, AssetFrozen());
     require(
       spoke.config.supplyCap == type(uint256).max ||
         asset.toSuppliedAssetsUp(spoke.suppliedShares) + amount <= spoke.config.supplyCap,
```
```suggestion
      spoke.config.supplyCap == type(uint256).max ||
        spoke.config.supplyCap >= asset.toSuppliedAssetsUp(spoke.suppliedShares) + amount,
```

src/contracts/LiquidityHub.sol
avniculae: good catch, should have been caught by tests (need to fix tests as well)

miguelmtzinf: i d remove `drawCap` as input param here. That's why we err here

avniculae: https://github.com/aave/aave-v4/issues/560

src/contracts/LiquidityHub.sol
miguelmtzinf: we aren't doing this `from` check in any other fn, should we?

src/contracts/Spoke.sol
```diff
@@ -696,9 +691,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     uint256 reserveId,
     bool usingAsCollateral
   ) internal view {
-    require(reserve.config.active, ReserveNotActive());
```
miguelmtzinf, line 688: `dev` docs referring to `enabled`.
can you fix and check if there is any other reference to active in this contract?

src/contracts/LiquidityHub.sol
miguelmtzinf: i d prefer to go with OZ for simplicity

src/contracts/HubConfigurator.sol
```diff
@@ -20,19 +20,6 @@ contract HubConfigurator is Ownable, IHubConfigurator {
...
-    address spoke,
-    uint256[] calldata assetIds,
-    DataTypes.SpokeConfig[] calldata configs
-  ) external onlyOwner {
-    require(assetIds.length == configs.length, MismatchedConfigs());
-    for (uint256 i; i < assetIds.length; i++) {
-      ILiquidityHub(hub).addSpoke(assetIds[i], spoke, configs[i]);
-    }
-  }
-
```
miguelmtzinf, line 35: why are we removing this?

src/contracts/HubConfigurator.sol
```diff
@@ -182,6 +145,109 @@ contract HubConfigurator is Ownable, IHubConfigurator {
...
+    uint256 assetId,
+    address spoke,
+    DataTypes.SpokeConfig calldata config
+  ) external override onlyOwner {
+    ILiquidityHub targetHub = ILiquidityHub(hub);
+    targetHub.updateSpokeConfig(assetId, spoke, config);
+  }
+
+  /// @inheritdoc IHubConfigurator
+  function addSpokeToAssets(
```
miguelmtzinf: i d move it up, before any spoke update fn

src/contracts/SpokeConfigurator.sol
```diff
@@ -75,14 +75,6 @@ contract SpokeConfigurator is Ownable, ISpokeConfigurator {
...
-  /// @inheritdoc ISpokeConfigurator
-  function updateActive(address spoke, uint256 reserveId, bool active) external onlyOwner {
-    ISpoke targetSpoke = ISpoke(spoke);
-    DataTypes.ReserveConfig memory reserveConfig = targetSpoke.getReserveConfig(reserveId);
-    reserveConfig.active = active;
-    targetSpoke.updateReserveConfig(reserveId, reserveConfig);
-  }
-
   /// @inheritdoc ISpokeConfigurator
   function updatePaused(address spoke, uint256 reserveId, bool paused) external onlyOwner {
```
miguelmtzinf, line 79: is it fine to have `freezeAsset` and `pauseAsset` on Hub
and `updatePaused` and `updateFrozen` on Spoke? We could do `pauseReserve` and `freezeReserve` instead
Also, how do we completely pause/freeze an spoke (all reserves)? we need fns for that

src/contracts/HubConfigurator.sol
yan-man: +1

src/contracts/LiquidityHub.sol
avniculae: we emit withdrawable in the error, that's why it's a separate variable

src/contracts/LiquidityHub.sol
avniculae: I think the way it is at the moment if more clear, but will address

src/contracts/HubConfigurator.sol
avniculae: it's not removed, it's moved below

src/contracts/HubConfigurator.sol
avniculae: exactly, this is what I did. first the asset update fns, and then the spoke update fns

src/contracts/SpokeConfigurator.sol
avniculae: > is it fine to have freezeAsset and pauseAsset on Hub
and updatePaused and updateFrozen on Spoke?

I think it is, since in the hub case we don't case paused/frozen flags, whereas in the spoke we have (so we call them update, as any other fn in the configurators)

avniculae: > Also, how do we completely pause/freeze an spoke (all reserves)? we need fns for that

Sure, can add this

avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/Spoke.sol
avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/LiquidityHub.sol
avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/LiquidityHub.sol
avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/LiquidityHub.sol
avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/LiquidityHub.sol
avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/LiquidityHub.sol
avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/LiquidityHub.sol
avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/LiquidityHub.sol
avniculae: [1511769](https://github.com/aave/aave-v4/pull/553/commits/151176905d89abdb55d5549d67a9b67d22daafb8)

src/contracts/LiquidityHub.sol
```diff
@@ -26,6 +28,7 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   mapping(uint256 assetId => DataTypes.Asset assetData) internal _assets;
   mapping(uint256 assetId => mapping(address spokeAddress => DataTypes.SpokeData spokeData))
     internal _spokes;
+  mapping(uint256 assetId => EnumerableSet.AddressSet spokeAddresses) internal _spokesList;
```
```suggestion
  mapping(uint256 assetId => EnumerableSet.AddressSet spoke) internal _spokesList;
```
not sure if we should rename to smt different like `_assetToSpokes` or `_assetToSpokeSet`

src/contracts/LiquidityHub.sol
```diff
@@ -116,19 +116,10 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   ) external restricted {
     require(assetId < _assetCount, AssetNotListed());
     require(spoke != address(0), InvalidSpoke());
-    require(!_isSpokeListed[assetId][spoke], SpokeAlreadyListed());
+    require(!_spokesList[assetId].contains(spoke), SpokeAlreadyListed());
```
miguelmtzinf: is `SpokeAlreadyAdded` better?

src/contracts/LiquidityHub.sol
```diff
@@ -116,19 +116,10 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-    _spokes[assetId][spoke] = DataTypes.SpokeData({
-      suppliedShares: 0,
-      baseDrawnShares: 0,
-      premiumDrawnShares: 0,
-      premiumOffset: 0,
-      realizedPremium: 0,
-      config: config
-    });
+    _spokes[assetId][spoke].config = config;
+    _spokesList[assetId].add(spoke);
```
miguelmtzinf: let's flip the order so it's easier to read: we check it's not added yet and we add

src/contracts/LiquidityHub.sol
```diff
@@ -352,15 +343,15 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-    return _spokesList[assetId].length;
+    return _spokesList[assetId].length();
   }
 
   function getSpokeAddress(uint256 assetId, uint256 index) external view returns (address) {
-    return _spokesList[assetId][index];
+    return _spokesList[assetId].at(index);
   }
 
   function isSpokeListed(uint256 assetId, address spoke) external view returns (bool) {
```
miguelmtzinf, line 354: here we could do `isSpoke` or `isAssetSpoke` idk

src/contracts/LiquidityHub.sol
miguelmtzinf: ah, good point.
can we have withdrawable and availableLiq on the left side of the comparison? so it's aligned with caps checks

src/contracts/LiquidityHub.sol
miguelmtzinf: i think restore is the only one mandatory actually.. the check on draw/remove are just to avoid transferring tokens by mistake.
I don't have a strong opinion, but I d try to reduce number of errors: `InvalidFromAddress` and `InvalidToAddress` (if we decide to keep it)

src/contracts/SpokeConfigurator.sol
miguelmtzinf: given that we have `pauseAllReserves` now, I think it makes more sense to have `pauseReserve`
same for freeze

src/contracts/HubConfigurator.sol
miguelmtzinf: so i d move it up, right before `updateSpokeActive`

also, realizing we don't have the single add, `addSpoke`
i think we do need it

src/contracts/SpokeConfigurator.sol
avniculae: `updatePaused` accepts a boolean to set paused to true/false. `pauseAllReserves` sets paused to true for all reserves. I think unpausing all introduces issues, since you might want to keep some paused. To me the current naming still makes sense, wdyt?

src/contracts/LiquidityHub.sol
avniculae: to me this is very unintuitive tbh, curious what others think as well @DhairyaSethi @CheyenneAtapour @yan-man

src/contracts/LiquidityHub.sol
avniculae: The reason why I don't like `assetToSpoke` naming is because the mapping above does a similar thing: maps assetId to spoke address, and spoke address to spoke data

Also, we use spokeAddress for the mapping above, this is why I used it here as well

src/contracts/LiquidityHub.sol
avniculae: I used `Listed` bc we also have `AssetNotListed`, `SpokeNotListed`

src/contracts/LiquidityHub.sol
avniculae: [339f68d](https://github.com/aave/aave-v4/pull/553/commits/339f68d83f64ce4733147887c845ce4feac3a673)

src/contracts/LiquidityHub.sol
avniculae: [339f68d](https://github.com/aave/aave-v4/pull/553/commits/339f68d83f64ce4733147887c845ce4feac3a673)

src/contracts/HubConfigurator.sol
avniculae: we have `addSpokeToAssets`, isn't that enough?

src/contracts/LiquidityHub.sol
avniculae: > i think restore is the only one mandatory actually.. the check on draw/remove are just to avoid transferring tokens by mistake.

I agree, only add and restore are mandatory. However, I think it's fine to add the check to all 4, wdyt?

avniculae: [8d6511c](https://github.com/aave/aave-v4/pull/553/commits/8d6511c9c44e12b61dcc488e1f4e145b3c9af205)

src/contracts/HubConfigurator.sol
```diff
@@ -275,15 +267,20 @@ contract HubConfigurator is Ownable, IHubConfigurator {
...
-          supplyCap: type(uint256).max,
-          drawCap: type(uint256).max,
-          active: true
-        })
-      );
+      _updateSpokeCaps(hub, assetId, newFeeReceiver, type(uint256).max, type(uint256).max);
     }
   }
+
+  function _updateSpokeCaps(
```
miguelmtzinf, line 300: should we use this in `freezeAsset`?
let's add natspec dev docs as well please

src/contracts/HubConfigurator.sol
```diff
@@ -251,18 +247,14 @@ contract HubConfigurator is Ownable, IHubConfigurator {
   function _updateFeeReceiverSpokeConfig(
```
miguelmtzinf, line 265: can we add dev docs, specifying that active status of the old one, and new one if reused, remains untouched?

src/interfaces/ILiquidityHub.sol
```diff
@@ -61,10 +61,8 @@ interface ILiquidityHub is IAccessManaged {
 
   error InvalidSharesAmount();
   error InvalidAddAmount();
-  error InvalidAddFromHub();
-  error InvalidRemoveToHub();
-  error InvalidDrawToHub();
-  error InvalidRestoreFromHub();
+  error InvalidFromAddress();
+  error InvalidToAddress();
```
miguelmtzinf, line 65: now that i look at this, it's hard to know what is referring to... open to suggestions

src/contracts/LiquidityHub.sol
miguelmtzinf: I think my issue is that i'm used to the term `list` for assets/reserves and not spokes...
I guess it's better to use the the `list` term here and may in `isAssetSpoke`? as long as the verb `list` for spoke makes sense

src/contracts/LiquidityHub.sol
miguelmtzinf: re the mapping above, i'm kinda fine with `_spokes` because it provides general info about spokes (aligned with `_assets`), but i d change from `spokeAddress` to `spoke` as well.
re the mapping in question, i think is clearer `assetToSpokes` or `assetToSpokeList` or `assetToSpokeSet`, but not strong opinion tbh. If you discard this suggestion, do you think List is fine or we should change it to Set?

src/contracts/SpokeConfigurator.sol
miguelmtzinf: damn, how can we unpause or unfreeze all? missing this piece, which could be useful in case of temporary measures

src/contracts/HubConfigurator.sol
miguelmtzinf: if we add only 1, we need to build an array. idk, nbd

src/contracts/LiquidityHub.sol
avniculae: I don't think `listed` refers to lists (ie arrays). I think it's fine to align with the other errors

src/interfaces/ILiquidityHub.sol
avniculae: we have `Invalid(Add|Remove|Draw|Restore)Amount` errors, so I'm fine with: `Invalid(Add|Remove|Draw|Restore)(From|To)Address` (so 4 in total again). wdyt?

src/contracts/HubConfigurator.sol
avniculae: [526841c](https://github.com/aave/aave-v4/pull/553/commits/526841c551a7ffef0ad499ff7696d5ec7819becd)

src/contracts/HubConfigurator.sol
avniculae: [526841c](https://github.com/aave/aave-v4/pull/553/commits/526841c551a7ffef0ad499ff7696d5ec7819becd)

src/contracts/LiquidityHub.sol
avniculae: [526841c](https://github.com/aave/aave-v4/pull/553/commits/526841c551a7ffef0ad499ff7696d5ec7819becd)

src/contracts/HubConfigurator.sol
```diff
@@ -244,6 +241,13 @@ contract HubConfigurator is Ownable, IHubConfigurator {
...
   }
 
+  /**
+   * @dev Updates the spoke configs for the old and new fee receivers.
+   *  - updates the caps for the old fee receiver to 0.
+   *  - if new fee receiver is not already a spoke, it adds it with max caps and active flag set to true.
+   *  - if new fee receiver is already a spoke, it updates the caps to max, without changing the active flag.
+   * @dev If the old and new fee receivers are the same, it does nothing.
+   */
   function _updateFeeReceiverSpokeConfig(
```
miguelmtzinf, line 265: let's add `@param` too
same for `_updateSpokeCaps`

src/contracts/LiquidityHub.sol
```diff
@@ -343,15 +343,15 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   }
 
   function getSpokeCount(uint256 assetId) external view returns (uint256) {
-    return _spokesList[assetId].length();
+    return _assetToSpokes[assetId].length();
   }
 
   function getSpokeAddress(uint256 assetId, uint256 index) external view returns (address) {
```
miguelmtzinf, line 349: should we just return `SpokeData` instead?

miguelmtzinf: it's probably fine, there is no other way to fetch the address of the spoke by index.

src/contracts/SpokeConfigurator.sol
miguelmtzinf: i'm fine with update(bool) and pauseAllReserves.
Created this ticket to describes this aspect and some functions I am missing
https://github.com/aave/aave-v4/issues/569

src/contracts/LiquidityHub.sol
miguelmtzinf: reverted back to `isSpokeListed`

src/contracts/HubConfigurator.sol
miguelmtzinf: [6ef66c3](https://github.com/aave/aave-v4/pull/553/commits/6ef66c30542074efec6cd182378ae9b8552d25e8)

src/contracts/LiquidityHub.sol
miguelmtzinf: [6ef66c3](https://github.com/aave/aave-v4/pull/553/commits/6ef66c30542074efec6cd182378ae9b8552d25e8)

src/contracts/HubConfigurator.sol
miguelmtzinf: added in [6ef66c3](https://github.com/aave/aave-v4/pull/553/commits/6ef66c30542074efec6cd182378ae9b8552d25e8)

src/contracts/LiquidityHub.sol
yan-man: favor consistency above all else as it just feels subjective. agree with alex here but validateAdd has caps on the left, so can just go with that

before I think we agreed that "thing of interest to be checked" goes on the left, ie for checks or tests

src/contracts/HubConfigurator.sol
```diff
@@ -176,30 +139,142 @@ contract HubConfigurator is Ownable, IHubConfigurator {
...
+  function pauseAsset(address hub, uint256 assetId) external override onlyOwner {
+    ILiquidityHub targetHub = ILiquidityHub(hub);
+    uint256 spokesCount = targetHub.getSpokeCount(assetId);
+    for (uint256 i = 0; i < spokesCount; ++i) {
+      address spokeAddress = targetHub.getSpokeAddress(assetId, i);
+      DataTypes.SpokeConfig memory config = targetHub.getSpokeConfig(assetId, spokeAddress);
+      config.active = false;
+      targetHub.updateSpokeConfig(assetId, spokeAddress, config);
+    }
+  }
```
CheyenneAtapour, line 168: To me these seem less efficient to iterate all spokes like this than to just set an asset-wide flag as before to pause/freeze all actions on an asset, no?

src/contracts/HubConfigurator.sol
CheyenneAtapour, line 1: I thought we would also add `setInterestRateData` here, but can do in a separate pr

src/contracts/HubConfigurator.sol
miguelmtzinf: yes. we could have a general switch as well, but feels a bit cumbersome though

src/contracts/HubConfigurator.sol
miguelmtzinf: ah true. Can you create a ticket for this please?

src/contracts/LiquidityHub.sol
CheyenneAtapour: I like how it's currently written in the code as `amount <= withdrawable`

src/contracts/Spoke.sol
```diff
@@ -675,7 +670,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
 
   /**
    * @dev Validates the reserve can be set as collateral.
-   * @dev Collateral can be disabled if the reserve is frozen but not enabled.
+   * @dev Collateral can be disabled if the reserve is frozen.
```
CheyenneAtapour, line 673: could say frozen but not paused here, but nbd

src/contracts/HubConfigurator.sol
CheyenneAtapour: created: https://github.com/orgs/aave/projects/14?pane=issue&itemId=121524390&issue=aave%7Caave-v4%7C571

src/contracts/Spoke.sol
CheyenneAtapour: yeah on second thought, paused is the new active flag, and nothing can be done on paused, so it's fine as is

src/contracts/LiquidityHub.sol
avniculae: yes, exactly, we already have a `getSpoke` fn to get the SpokeData

PR #564, opened 2025-07-24
feat: allow dao to upd user dyn config

### LINKED ISSUE #570: Allow DAO to update any user dynamic risk config
The user can refresh their own dynamic risk configuration, but we wanna allow the DAO to rebalance any user config as well (same approach as `updateUserRiskPremium`).
https://github.com/aave/aave-v4/blob/5b0ae7495194c2f54f39acde37e44feb8f0810d9/src/contracts/Spoke.sol#L375-L377

If the DAO wants to change any user dynamic risk configuration, it can:
- change the dynamic risk config key the user is using -> this impact other users sharing the key
- update the dynamic risk config key to the latest config.

DhairyaSethi:
Closes #570


src/contracts/Spoke.sol
```diff
@@ -372,7 +372,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   }
 
   /// @inheritdoc ISpoke
-  function updateUserDynamicConfig(address onBehalfOf) external onlyPositionManager(onBehalfOf) {
+  function updateUserDynamicConfig(address onBehalfOf) external {
+    if (!_isPositionManager({user: onBehalfOf, manager: msg.sender})) {
+      _checkCanCall(msg.sender, msg.data);
```
avniculae, line 377:
```suggestion
      _checkCanCall(_msgSender(), _msgData());
```

to stay consistent with the AccessManaged impl, wdyt?

tests/Base.t.sol
```diff
@@ -2165,4 +2176,11 @@ abstract contract Base is Test {
     assertEq(a.premiumDebt, b.premiumDebt, 'premiumDebt');
     assertEq(abi.encode(a), abi.encode(b)); // sanity check
   }
+
+  modifier resumePrank() {
+    (VmSafe.CallerMode callerMode, address msgSender, address txOrigin) = vm.readCallers();
```
avniculae, line 2090: this is nice

avniculae: I would call the function `pausePrank`, wdyt?

DhairyaSethi: good call, that's better

src/contracts/Spoke.sol
DhairyaSethi: it's a dep so we don't need to follow its convention and we explicitly decided not to use those abstractions (https://github.com/aave/aave-v4/pull/462#discussion_r2216494315)
otherwise should update all invocations msg.sender w the internal abstraction

tests/Base.t.sol
```diff
@@ -1055,7 +1066,7 @@ abstract contract Base is Test {
     ISpoke spoke,
     uint256 reserveId,
     uint256 newLiquidationFee
-  ) internal {
+  ) internal resumePrank {
```
miguelmtzinf: adding this for only one function helper can make the code confusing and harder to maintain. If your plan is to apply to all later i'm good with it

tests/Base.t.sol
```diff
@@ -232,40 +233,50 @@ abstract contract Base is Test {
   }
 
   function setUpRoles(
-    ILiquidityHub hub,
+    ILiquidityHub liquidityHub,
```
miguelmtzinf: i d rename to `targetHub`. Also, the global `hub` could become `_hub` no? i know its a big change i we probably prefer to tackle separately

tests/Base.t.sol
DhairyaSethi: thought it's a cool modifier, can add it all and it might be reused later. the idea is that these helper abstractions can be used inside an ongoing startPrank

src/contracts/Spoke.sol
```diff
@@ -372,7 +372,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+    }
     _refreshDynamicConfig(onBehalfOf);
```
miguelmtzinf, line 379: should we allow refreshing only one reserve?

DhairyaSethi: we discussed in the last PR, it feels gameable to me (no attack in mind yet) 
and right we anyway offer refresh all w/o this method with a borrow(1 wei) or withdraw(1 wei) similar to update user risk premium 
BUT 
we don't already allow single rp refresh force bc that's only possible while enabling as collateral & we short circuit on that method

miguelmtzinf: https://github.com/aave/aave-v4/issues/572

PR #567, opened 2025-07-25
feat: Change rounding direction for offset

### LINKED ISSUE #548: Apply loose conditions on Hub refresh method
Whenever a refresh occurs, we ensure the premium debt does not increase by more than 2 wei due to accruals. In theory, no increase should occur, as the operation simply transitions accounting from in-flight to realized premium debt.
https://github.com/aave/aave-v4/blob/afe9eda2dfd4a79c09bef92901d00397fea3836b/src/contracts/LiquidityHub.sol#L265-L272

However, during the computation of premium debt accruals (both at the user and spoke level), we may see up to a 2 wei increase. This is due to applying opposite rounding directions: we round up when converting premium drawn shares to assets and down when calculating the offset.
https://github.com/aave/aave-v4/blob/afe9eda2dfd4a79c09bef92901d00397fea3836b/src/contracts/Spoke.sol#L726-L727

These opposite roundings were introduced to prevent underflows in asset-level accounting. That said, we aim to revisit this logic and:
1. Align the rounding direction for both calculations and identify any resulting failures.
2. Remove the condition on Hub refresh.
3. Introduce a more relaxed computation of asset-level realized premium on the Hub side: if the offset exceeds the converted premium drawn shares, we set the value at zero.

This approach allows us to eliminate the instant creation of premium debt, which significantly impacts overcollateralization (see issue [#421](https://github.com/aave/aave-v4/issues/421) and https://github.com/aave/aave-v4/issues/527)). To ensure safety, we must coordinate with Certora to formally verify that the premium offset never exceeds the converted premium drawn shares by more than 2 wei—effectively bounding the rounding error within acceptable limits.

CheyenneAtapour:
closes #548


src/contracts/LiquidityHub.sol
```diff
@@ -279,7 +279,18 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+      require(
+        premiumDebtBefore - (premiumDebtAfter + realizedPremiumTaken) <= 2,
+        InvalidDebtChange()
+      );
+    } else {
+      require(
+        premiumDebtAfter + realizedPremiumTaken - premiumDebtBefore <= 2,
+        InvalidDebtChange()
+      );
+    }
```
CheyenneAtapour, line 287: Will remove this, just verifying the diff

src/contracts/LiquidityHub.sol
```diff
@@ -319,13 +330,21 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
     asset.accrue(assetId, _spokes[assetId][asset.config.feeReceiver]);
 
+    uint256 realizedPremium = asset.realizedPremium + realizedPremiumAdded;
+
     asset.premiumDrawnShares = _add(asset.premiumDrawnShares, premiumDrawnShareDelta);
     asset.premiumOffset = _add(asset.premiumOffset, premiumOffsetDelta);
-    asset.realizedPremium = asset.realizedPremium + realizedPremiumAdded - realizedPremiumTaken;
+    asset.realizedPremium = realizedPremiumTaken <= realizedPremium
+      ? realizedPremium - realizedPremiumTaken
+      : 0;
```
CheyenneAtapour: Thinking to add a `_sub` helper method which subtracts values but returns 0 instead in case of underflow

DhairyaSethi: yes, branchless impl: https://github.com/morpho-org/morpho-blue/blob/main/src/libraries/UtilsLib.sol#L33

src/contracts/LiquidityHub.sol
DhairyaSethi: diff should ideally be 0 with this (no reference of two should remain)

src/contracts/LiquidityHub.sol
CheyenneAtapour: Added that

miguelmtzinf:
We proved 2 wei is the maximum delta on premium.

PR #574, opened 2025-07-25
feat: simplify restore flow

DhairyaSethi:
Consolidates restore accounting by removing responsibility of `refreshPremiumDebt` to settle premium by adding an additional `PremiumDelta` param on `restore`.

### DIFF
--- a/src/libraries/logic/AssetLogic.sol
+++ b/src/libraries/logic/AssetLogic.sol
@@ -179,6 +179,7 @@ library AssetLogic {
       return 0;
     }
 
+    // liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee
     // prettier-ignore
     uint256 feesAmount = (
       asset.baseDrawnShares.rayMulDown(nextDrawnIndex - currentDrawnIndex) +


src/contracts/LiquidityHub.sol
```diff
@@ -237,42 +238,34 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     uint256 totalRestoredAmount = baseAmount + premiumAmount;
     asset.availableLiquidity += totalRestoredAmount;
 
-    /// @dev premium debt must be restored in `refreshPremiumDebt` before calling this function
     asset.updateBorrowRate(assetId);
 
     IERC20(asset.underlying).safeTransferFrom(from, address(this), totalRestoredAmount);
 
-    emit Restore(assetId, msg.sender, baseDrawnSharesRestored, totalRestoredAmount);
+    emit Restore(assetId, msg.sender, baseDrawnSharesRestored, premiumDelta, totalRestoredAmount);
```
miguelmtzinf: no way to know how much premium was generated and repaid

DhairyaSethi: premiumDelta, but yeah not explicitly enough, will have to split totalRestoredAmount to base & premiumAmount

src/contracts/LiquidityHub.sol
```diff
@@ -220,15 +220,17 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-
     DataTypes.Asset storage asset = _assets[assetId];
     DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
 
     asset.accrue(assetId, _spokes[assetId][asset.config.feeReceiver]);
 
+    uint256 premiumDebtBefore = asset.premiumDebt();
+    _applyPremiumDelta(asset, spoke, premiumDelta);
+    require(asset.premiumDebt() + premiumAmount - premiumDebtBefore <= 2, PremiumDebtChanged());
     _validateRestore(asset, spoke, baseAmount, premiumAmount, from);
```
avniculae, line 231: I would personally do this check before accrual above

DhairyaSethi: why? makes sense to follow state changes after accrual pattern (irrelevant for debt state changes, like here, bc index)

src/contracts/LiquidityHub.sol
```diff
@@ -237,42 +239,41 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-      realizedPremiumAdded,
-      realizedPremiumTaken
-    );
-    uint256 premiumDebtAfter = asset.premiumDebt();
+    _applyPremiumDelta(asset, spoke, premiumDelta);
     // can increase due to precision loss on premium debt (base unchanged)
     // todo mathematically find premium diff ceiling and replace the `2`
-    // if no premium debt is restored, premium debt remains unchanged
-    require(premiumDebtAfter + realizedPremiumTaken - premiumDebtBefore <= 2, InvalidDebtChange());
+    require(asset.premiumDebt() - premiumDebtBefore <= 2, PremiumDebtChanged());
```
avniculae: this check makes sure realizedDelta is not negative here right? (to make sure premium debt is not restored through refresh)

what I mean is that: while before we could decide to remove it, now it's kind of needed right?

src/contracts/Spoke.sol
```diff
@@ -302,7 +304,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     (vars.newUserRiskPremium, , , , ) = _calculateUserAccountData(onBehalfOf);
     _notifyRiskPremiumUpdate(onBehalfOf, vars.newUserRiskPremium);
 
-    emit Repay(reserveId, msg.sender, onBehalfOf, vars.restoredShares);
+    emit Repay(reserveId, msg.sender, onBehalfOf, vars.restoredShares); // todo: add premiumDelta
```
DhairyaSethi, line 307: this is to be added in this event in a separate PR since this requires some refactor repay tests since we already don't expect emit the previous refresh event during settling, the equivalence of debt state changes are asserted by the current test suite

tests/gas/LiquidityHub.Operations.gas.t.sol
```diff
@@ -56,14 +56,14 @@ contract LiquidityHubOperations_Gas_Tests is Base {
     skip(1000);
 
     (baseDebtRemaining, premiumDebtRemaining) = hub.getSpokeDebt(daiAssetId, address(spoke1));
-    hub.restore(daiAssetId, baseDebtRemaining / 2, premiumDebtRemaining, alice);
+    hub.restore(daiAssetId, baseDebtRemaining / 2, 0, DataTypes.PremiumDelta(0, 0, 0), alice);
```
avniculae, line 59: why 0?

DhairyaSethi: bc in this test we set any premium debt (so premium debt regardless is 0)
this will be made more accurate with some additional helpers for liquidity hub draw tests

src/contracts/LiquidityHub.sol
DhairyaSethi: we could not remove it before either, why do you say that?

this check ensures `premium debt is not restored through refresh` by checking total premium debt remains unchained (within the 2 wei precision diff due premiumShares/offset fix we have)

DhairyaSethi: we have a trust assumption between spoke/hub but should strive to minimise it wherever possible

src/contracts/LiquidityHub.sol
DhairyaSethi: had missunderstood haha, as discussed in sync moved after accrual but before state change [de3639a](https://github.com/aave/aave-v4/pull/574/commits/de3639a0739b8a44df5dc44785117e53fcbca70e)

src/contracts/LiquidityHub.sol
```diff
@@ -298,36 +299,18 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
   }
 
-  function _refresh(
-    uint256 assetId,
-    address spokeAddress,
-    int256 premiumDrawnShareDelta,
-    int256 premiumOffsetDelta,
-    uint256 realizedPremiumAdded,
-    uint256 realizedPremiumTaken
+  function _applyPremiumDelta(
```
yan-man, line 334: can you add some dev comments here for this method

src/contracts/LiquidityHub.sol
```diff
@@ -220,59 +220,60 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-    int256 premiumOffsetDelta,
-    uint256 realizedPremiumAdded,
-    uint256 realizedPremiumTaken
+    DataTypes.PremiumDelta calldata premiumDelta
   ) external {
-    require(_spokes[assetId][msg.sender].config.active, SpokeNotActive());
-
     DataTypes.Asset storage asset = _assets[assetId];
+    DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
+    require(spoke.config.active, SpokeNotActive());
```
yan-man: I think we can move these 2 lines above L263 to fail earlier

tests/Utils.sol
```diff
@@ -49,20 +49,19 @@ library Utils {
...
     ILiquidityHub hub,
     uint256 assetId,
     address caller,
     uint256 baseAmount,
-    uint256 premiumAmount,
     address repayer
   ) internal returns (uint256) {
     vm.startPrank(repayer);
-    IERC20(hub.getAsset(assetId).underlying).approve(address(hub), (baseAmount + premiumAmount));
+    IERC20(hub.getAsset(assetId).underlying).approve(address(hub), baseAmount);
```
yan-man: should we use that `_approve` helper you created

yan-man: and should this approve be for base + premium?

tests/gas/LiquidityHub.Operations.gas.t.sol
yan-man: > we set any premium debt

can you clarify wdym here

src/interfaces/ILiquidityHub.sol
```diff
@@ -157,43 +156,38 @@ interface ILiquidityHub is IAccessManaged {
...
+    DataTypes.PremiumDelta calldata premiumDelta,
     address from
   ) external returns (uint256);
 
   /**
    * @notice Refreshes premium debt accounting.
-   * @dev To be called when moving accrued premium to realized premium.
-   * @dev Only callable by spokes.
-   * @dev Premium debt can only decrease by at most the amount of realized premium taken.
+   * @dev Only callable by active spokes, reverts with `SpokeNotActive` otherwise.
```
CheyenneAtapour, line 186: I don't think the revert message here is really necessary, but fine nbd

src/interfaces/ILiquidityHub.sol
```diff
@@ -157,43 +156,38 @@ interface ILiquidityHub is IAccessManaged {
...
-    int256 premiumOffsetDelta,
-    uint256 realizedPremiumAdded,
-    uint256 realizedPremiumTaken
+    DataTypes.PremiumDelta calldata premiumDelta
   ) external;
 
   /**
    * @notice Pay existing liquidity to feeReceiver.
-   * @dev Only callable by spokes.
+   * @dev Only callable by active spokes.
```
CheyenneAtapour, line 198: The purpose of these comments imo is more so for access control, but also nbd

tests/Utils.sol
CheyenneAtapour: Just base, right? why premium?

tests/gas/LiquidityHub.Operations.gas.t.sol
CheyenneAtapour: Dhairya typo'ed I think and meant to say we *don't* set any premium debt here

src/contracts/LiquidityHub.sol
DhairyaSethi: wdym, the fn starts at 263, this is the earliest we can fail

tests/Utils.sol
DhairyaSethi: yep rebased with approve helper, should just be base bc this restores base debt only

src/contracts/Spoke.sol
```diff
@@ -710,34 +716,19 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-      hub,
-      assetId,
-      reserveId,
-      user,
-      -int256(userPremiumDrawnShares),
-      -int256(userPremiumOffset),
-      accruedPremium,
-      premiumDebtRestored
-    );
+    userPosition.realizedPremium = _add(userPosition.realizedPremium, premiumDelta.realizedDelta);
```
avniculae, line 724: should we just inline this function in the fn above?

DhairyaSethi: yep sg

src/contracts/Spoke.sol
```diff
@@ -1031,22 +1007,24 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-          userPosition,
-          hub,
-          assetId,
-          reserveId,
-          user,
-          premiumDebtRestored
-        );
-        uint256 deficitShares = hub.reportDeficit(assetId, baseDebtRestored, premiumDebtRestored); // settle base debt here by reporting deficit
-
+        _settlePremiumDebt(reserve, userPosition, premiumDelta);
```
avniculae, line 1027: do we test that we settle premium debt?

DhairyaSethi: yes ofc, 9 tests fail when you comment that line (this is called mutation testing, good to do when refactoring)
mainly 
image

src/contracts/Spoke.sol
DhairyaSethi: #586

src/contracts/LiquidityHub.sol
yan-man: ie can load asset after spoke / SpokeNotActive

tests/Utils.sol
yan-man: yep yep you guys are right, base only

src/contracts/LiquidityHub.sol
DhairyaSethi: ah sure sure, not worth breaking consistency and readability imo but nbd, feel free to do in a future pr

PR #575, opened 2025-07-28
Add Hub Configurator Tests

CheyenneAtapour:
This PR adds some leftover tests from the previous #536 PR, which is no longer needed

### DIFF
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -19,8 +19,6 @@ import {TreasurySpoke, ITreasurySpoke} from 'src/contracts/TreasurySpoke.sol';
 import {HubConfigurator, IHubConfigurator} from 'src/contracts/HubConfigurator.sol';
 import {SpokeConfigurator, ISpokeConfigurator} from 'src/contracts/SpokeConfigurator.sol';
 import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
-import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
-import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {SharesMath} from 'src/libraries/math/SharesMath.sol';
 import {MathUtils} from 'src/libraries/math/MathUtils.sol';
--- a/tests/unit/HubConfigurator.t.sol
+++ b/tests/unit/HubConfigurator.t.sol
@@ -76,7 +76,14 @@ contract HubConfiguratorTest is HubBase {
 
     vm.expectRevert(IHub.InvalidAssetDecimals.selector, address(hub1));
     vm.prank(HUB_CONFIGURATOR_ADMIN);
-    _addAsset(true, underlying, decimals, feeReceiver, interestRateStrategy, encodedIrData);
+    _addAsset(
+      fetchErc20Decimals,
+      underlying,
+      decimals,
+      feeReceiver,
+      interestRateStrategy,
+      encodedIrData
+    );
   }
 
   function test_addAsset_revertsWith_InvalidUnderlying() public {
--- a/tests/unit/Spoke/SpokeBase.t.sol
+++ b/tests/unit/Spoke/SpokeBase.t.sol
@@ -223,13 +223,6 @@ contract SpokeBase is Base {
     return tempUser;
   }
 
-  function deal(ISpoke spoke, uint256 reserveId, address user, uint256 amount) internal {
-    IERC20 underlying = IERC20(spoke.getReserve(reserveId).underlying);
-    if (underlying.balanceOf(user) < amount) {
-      deal(address(underlying), user, amount);
-    }
-  }
-
   // @dev Borrows reserve by minimum required collateral for the same reserve
   function _backedBorrow(
     ISpoke spoke,
@@ -248,6 +241,13 @@ contract SpokeBase is Base {
     Utils.approve(spoke, collateralReserveId, user, UINT256_MAX);
     Utils.supplyCollateral(spoke, collateralReserveId, user, supplyAmount, user);
     Utils.borrow(spoke, debtReserveId, user, borrowAmount, user);
+  }
+
+  function deal(ISpoke spoke, uint256 reserveId, address user, uint256 amount) internal {
+    IERC20 underlying = IERC20(spoke.getReserve(reserveId).underlying);
+    if (underlying.balanceOf(user) < amount) {
+      deal(address(underlying), user, amount);
+    }
   }
 
   // increase share conversion index on given reserve


src/interfaces/ILiquidityHub.sol
```diff
@@ -112,6 +112,12 @@ interface ILiquidityHub is IAccessManaged {
 
   function addSpoke(uint256 assetId, address spoke, DataTypes.SpokeConfig calldata params) external;
```
DhairyaSethi: can we also just add natspec for addSpoke here

tests/unit/HubConfigurator.t.sol
```diff
@@ -60,29 +62,35 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
-    address feeReceiver,
-    address interestRateStrategy
-  ) public {
-    assumeUnusedAddress(underlying);
-    assumeNotZeroAddress(feeReceiver);
-    assumeNotZeroAddress(interestRateStrategy);
-
-    decimals = uint8(bound(decimals, hub.MAX_ALLOWED_ASSET_DECIMALS() + 1, type(uint8).max));
+  function test_addAsset_revertsWith_InvalidAssetDecimals() public {
+    uint8 decimals = hub.MAX_ALLOWED_ASSET_DECIMALS() + 1;
```
DhairyaSethi: can use randomUint(range..) as well for consistency but nbd 
same below w randomAddress

tests/unit/HubConfigurator.t.sol
```diff
@@ -171,27 +179,53 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
     assertEq(hub.getAssetConfig(assetId), expectedConfig);
   }
 
-  function test_updateFeeReceiver_revertsWith_OwnableUnauthorizedAccount() public {
-    vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, alice));
-    vm.prank(alice);
+  function test_updateFeeReceiver_fuzz_revertsWith_OwnableUnauthorizedAccount(
+    address caller
+  ) public {
+    vm.assume(caller != HUB_CONFIGURATOR_ADMIN);
```
DhairyaSethi, line 224: we said we'll make an exception for these access control tests bc it's ownable (so only one possible owner at a time) but nbd i'm good with this

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: I'll skip because we already have better tests for this within liquidity hub config tests, and there is a chance for the random address to be something already in use

src/interfaces/ILiquidityHub.sol
```diff
@@ -110,8 +110,20 @@ interface ILiquidityHub is IAccessManaged {
    */
   function updateAssetConfig(uint256 assetId, DataTypes.AssetConfig calldata config) external;
 
+  /**
+   * @notice Registers a new spoke for a specific asset in the hub's internal accounting.
```
```suggestion
   * @notice Registers a new spoke for a specific asset in the hub.
```
nit but nbd

tests/Base.t.sol
```diff
@@ -1659,6 +1659,86 @@ abstract contract Base is Test {
...
+    uint256 normalizedDebtAmount = (debtAmount * debtPrice).wadify() / debtAssetUnits;
+    uint256 normalizedCollPrice = collPrice.wadify() / collAssetUnits;
+
+    return
+      (normalizedDebtAmount.wadify() /
+        normalizedCollPrice.wadify().percentMul(colDynConf.collateralFactor)) + 1;
+  }
+
+  /// @dev Opens a debt position for a random user, using same asset as collateral and borrow
+  function _openDebtPosition(
```
miguelmtzinf: why are we moving this from SpokeBase? we have similar functions for the Hub side if needed

tests/unit/HubConfigurator.t.sol
```diff
@@ -264,6 +298,124 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+  /// @dev Test update fee receiver and fees can still be withdrawn from old fee receiver
+  function test_updateFeeReceiver_WithdrawFromOldSpoke() public {
+    assertEq(
+      hub.getAssetConfig(daiAssetId).feeReceiver,
+      address(treasurySpoke),
+      'current fee receiver matches treasury spoke'
+    );
+
+    // Create debt to build up fees on the existing treasury spoke
+    _openDebtPosition(spoke1, _daiReserveId(spoke1), 100e18, true);
```
miguelmtzinf: check `_addLiquidity` and `_drawLiquidity`

tests/unit/HubConfigurator.t.sol
```diff
@@ -264,6 +298,124 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+    skip(365 days);
+
+    assertGe(treasurySpoke.getSuppliedShares(daiAssetId), 0);
+    uint256 fees = treasurySpoke.getSuppliedAmount(daiAssetId);
+
+    // Change the fee receiver
+    TreasurySpoke newTreasurySpoke = new TreasurySpoke(HUB_ADMIN, address(hub));
```
miguelmtzinf: cannot recall how other tests are doing, but here we could even use a mocked address (no need to use a real treasury spoke)

tests/unit/HubConfigurator.t.sol
```diff
@@ -264,6 +298,124 @@ contract HubConfiguratorTest is LiquidityHubBase {
...
+
+    // Ensure old fee receiver is still active
+    assertTrue(
+      hub.getSpokeConfig(daiAssetId, address(treasurySpoke)).active,
+      'old fee receiver is not active'
+    );
+
+    // Withdraw half the fee shares from the old treasury spoke
+    Utils.withdraw(
+      _treasurySpoke(),
```
miguelmtzinf: a bit awkward that this fn now is in Base.sol and now is referring to the old receiver. 
I d keep the helper as internal in TreasurySpoke tbh and here use the full thing (only used 3 times)

tests/unit/HubConfigurator.t.sol
DhairyaSethi: i specifically just mean for these revert case tests only, for consistency

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: Done

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Here I'm actually calling withdraw on the treasury spoke though, so I think it's nice to use a real treasury spoke

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: Done

tests/Base.t.sol
CheyenneAtapour: Moved back

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Using now

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Done

PR #577, opened 2025-07-28
[DRAFT] feat: Inflation Attack Protection

avniculae:


avniculae:
Implemented a second approach #588

PR #581, opened 2025-07-29
feat: split restore flow

DhairyaSethi:
todo: some tests in restore need adjustment which expect InvalidRestoreAmounts reverts.

## Design

Split current restore flow in two, ie take premium settlement independantly.


DhairyaSethi:
decided to move forward with #574

since both are essentially the same but 574 offers a simpler integration with our current spoke model & infrastructure, and is cheaper in the case of repaying both base + premium debt

PR #583, opened 2025-07-30
feat: eliminate deficit

### LINKED ISSUE #555: Add function to burn supply shares on Hub (umbrella compatibility)
We are able to report deficit but there is no way to restore that.
Umbrella can work as spoke where it clears the deficit under certain situations.

Whenever is bad debt:
- decrease debt
- increase deficit
- supplied assets should be equal to debt assets + deficit assets, so LPs cannot fully withdraw

Whenever the bad debt is cleared: 
- reduce deficit
- reduce supplied shares but keep liquidity
- supplied assets should be equal to debt assets, so LPs can fully withdraw

references:
- https://github.com/aave-dao/aave-v3-origin/blob/312a899aaa9f7f4685d051d13185ada1902a2730/docs/3.3/Aave-v3.3-properties.md?plain=1#L7
- https://github.com/bgd-labs/aave-umbrella

DhairyaSethi:
closes #555 
addresses #585 

Adds a method on hub to clear deficit by burning caller spoke supplied balance (instead of pulling assets).

This PR is complementary to an additional method to move supplied shares across spokes on the hub (which in turn requires #468 to avoid needing a complex callback on receiver spoke) and potentially replace payFee method.

### DIFF
--- a/src/contracts/LiquidityHub.sol
+++ b/src/contracts/LiquidityHub.sol
@@ -293,15 +293,12 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
 
     asset.accrue(assetId, _spokes[assetId][asset.config.feeReceiver]);
-    _validateEliminateDeficit(spoke, amount);
-
-    uint256 deficit = asset.deficit;
-    if (amount > deficit) amount = deficit;
+    _validateEliminateDeficit(asset, spoke, amount);
 
     uint256 removedShares = previewRemoveByAssets(assetId, amount);
     asset.suppliedShares -= removedShares;
     spoke.suppliedShares -= removedShares;
-    asset.deficit = deficit - amount;
+    asset.deficit -= amount;
 
     asset.updateBorrowRate(assetId);
 
@@ -348,31 +345,6 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
 
     emit Remove(assetId, msg.sender, feeShares, feeAmount);
     emit Add(assetId, feeReceiver, feeShares, feeAmount);
-  }
-
-  /**
-   * @dev Applies premium deltas on asset and spoke debt, and validates that total premium debt
-   * cannot decrease by more than `premiumAmount`.
-   */
-  function _applyPremiumDelta(
-    DataTypes.Asset storage asset,
-    DataTypes.SpokeData storage spoke,
-    DataTypes.PremiumDelta calldata premium,
-    uint256 premiumAmount
-  ) internal {
-    uint256 premiumDebtBefore = asset.premiumDebt();
-
-    asset.premiumDrawnShares = _add(asset.premiumDrawnShares, premium.drawnSharesDelta);
-    asset.premiumOffset = _add(asset.premiumOffset, premium.offsetDelta);
-    asset.realizedPremium = _add(asset.realizedPremium, premium.realizedDelta);
-
-    spoke.premiumDrawnShares = _add(spoke.premiumDrawnShares, premium.drawnSharesDelta);
-    spoke.premiumOffset = _add(spoke.premiumOffset, premium.offsetDelta);
-    spoke.realizedPremium = _add(spoke.realizedPremium, premium.realizedDelta);
-
-    // can increase due to precision loss on premium debt (base unchanged)
-    // todo mathematically find premium diff ceiling and replace the `2`
-    require(asset.premiumDebt() + premiumAmount - premiumDebtBefore <= 2, InvalidDebtChange());
   }
 
   /// @inheritdoc ILiquidityHub
@@ -562,6 +534,31 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   // Internal
   //
 
+  /**
+   * @dev Applies premium deltas on asset and spoke debt, and validates that total premium debt
+   * cannot decrease by more than `premiumAmount`.
+   */
+  function _applyPremiumDelta(
+    DataTypes.Asset storage asset,
+    DataTypes.SpokeData storage spoke,
+    DataTypes.PremiumDelta calldata premium,
+    uint256 premiumAmount
+  ) internal {
+    uint256 premiumDebtBefore = asset.premiumDebt();
+
+    asset.premiumDrawnShares = _add(asset.premiumDrawnShares, premium.drawnSharesDelta);
+    asset.premiumOffset = _add(asset.premiumOffset, premium.offsetDelta);
+    asset.realizedPremium = _add(asset.realizedPremium, premium.realizedDelta);
+
+    spoke.premiumDrawnShares = _add(spoke.premiumDrawnShares, premium.drawnSharesDelta);
+    spoke.premiumOffset = _add(spoke.premiumOffset, premium.offsetDelta);
+    spoke.realizedPremium = _add(spoke.realizedPremium, premium.realizedDelta);
+
+    // can increase due to precision loss on premium debt (base unchanged)
+    // todo mathematically find premium diff ceiling and replace the `2`
+    require(asset.premiumDebt() + premiumAmount - premiumDebtBefore <= 2, InvalidDebtChange());
+  }
+
   function _validateAdd(
     DataTypes.Asset storage asset,
     DataTypes.SpokeData storage spoke,
@@ -649,17 +646,12 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   }
 
   function _validateEliminateDeficit(
+    DataTypes.Asset storage asset,
     DataTypes.SpokeData storage spoke,
     uint256 amount
   ) internal view {
     require(spoke.config.active, SpokeNotActive());
-    require(amount != 0, InvalidDeficitAmount());
-  }
-
-  // handles underflow
-  function _add(uint256 a, int256 b) internal pure returns (uint256) {
-    if (b >= 0) return a + uint256(b);
-    return a - uint256(-b);
+    require(amount != 0 && amount <= asset.deficit, InvalidDeficitAmount());
   }
 
   function _validatePayFee(
@@ -668,5 +660,11 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
   ) internal view {
     require(senderSpoke.config.active, SpokeNotActive());
     require(feeShares != 0, InvalidFeeShares());
+  }
+
+  // handles underflow
+  function _add(uint256 a, int256 b) internal pure returns (uint256) {
+    if (b >= 0) return a + uint256(b);
+    return a - uint256(-b);
   }
 }
--- a/tests/unit/LiquidityHub/LiquidityHub.EliminateDeficit.t.sol
+++ b/tests/unit/LiquidityHub/LiquidityHub.EliminateDeficit.t.sol
@@ -4,11 +4,25 @@ pragma solidity ^0.8.0;
 import 'tests/unit/LiquidityHub/LiquidityHubBase.t.sol';
 
 contract LiquidityHubEliminateDeficitTest is LiquidityHubBase {
-  function test_eliminateDeficit_revertsWith_InvalidDeficitAmount() public {
+  function test_eliminateDeficit_revertsWith_InvalidDeficitAmount_zero() public {
     uint256 assetId = _randomAssetId(hub);
     vm.expectRevert(ILiquidityHub.InvalidDeficitAmount.selector);
     vm.prank(address(spoke1));
     hub.eliminateDeficit(assetId, 0);
+
+    _createDeficit(assetId, spoke1, 1000e6);
+    assertEq(hub.getDeficit(assetId), 1000e6);
+    vm.expectRevert(ILiquidityHub.InvalidDeficitAmount.selector);
+    vm.prank(address(spoke1));
+    hub.eliminateDeficit(assetId, 0);
+  }
+
+  function test_eliminateDeficit_revertsWith_InvalidDeficitAmount_excess() public {
+    uint256 assetId = _randomAssetId(hub);
+    _createDeficit(assetId, spoke1, 1000e6);
+    vm.expectRevert(ILiquidityHub.InvalidDeficitAmount.selector);
+    vm.prank(address(spoke1));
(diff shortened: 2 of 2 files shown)


tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
```diff
@@ -0,0 +1,135 @@
...
+
+    assertEq(removedShares, expectedRemoveShares);
+    assertEq(hub.getDeficit(assetId), deficit - clearedDeficit);
+    assertEq(hub.getAssetSuppliedShares(assetId), assetSuppliedShares - expectedRemoveShares);
+    assertEq(
+      hub.getSpokeSuppliedShares(assetId, address(spoke1)),
+      spokeSuppliedShares - expectedRemoveShares
+    );
+    assertGe(getSupplyExRate(assetId), supplyExRate);
+    assertBorrowRateSynced(hub, assetId, 'clearDeficit');
```
yan-man: nice

src/contracts/LiquidityHub.sol
```diff
@@ -627,6 +654,11 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     // we should have already restored premium debt
   }
 
+  function _validateClearDeficit(DataTypes.SpokeData storage spoke, uint256 amount) internal view {
```
yan-man: should we also validate that the spoke has no debt: https://github.com/aave-dao/aave-v3-origin/blob/312a899aaa9f7f4685d051d13185ada1902a2730/docs/3.3/Aave-v3.3-properties.md?plain=1#L14

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
yan-man: might as well add this to every test where borrow rate is updated then

yan-man:
closes https://github.com/aave/aave-v4/issues/555 too?

src/contracts/LiquidityHub.sol
```diff
@@ -275,6 +275,29 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     return baseDrawnSharesRestored;
   }
 
+  /// @inheritdoc ILiquidityHub
+  function clearDeficit(uint256 assetId, uint256 amount) external returns (uint256) {
```
yan-man: nbd, should we re-use method name from v3: `eliminateReserveDeficit`

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
DhairyaSethi: in a separate rft mate, have noted, let's remember to add it on new tests from now on atleast

src/contracts/LiquidityHub.sol
DhairyaSethi: it's not a reserve, but rather asset, so would anyway be different from v3 exactly, either `eliminateDeficit`, or `eliminateAssetDeficit`, chose a different shorter name since we won't exactly re-use, open to suggestions

src/contracts/LiquidityHub.sol
DhairyaSethi: nope idts, we allow any arbitrary active spoke to clear deficit here bc the hub does not have collateralisation constraints (that's on the spoke) unlike v3 where umbrella is an user which, if with debt, would require hf checks on eliminateDeficit

src/contracts/LiquidityHub.sol
yan-man: yes meant to rm that remove part, perhaps just `eliminateDeficit` instead of clearDeficit, but nbd

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
```diff
@@ -0,0 +1,135 @@
...
+    uint256 removedShares = hub.clearDeficit(assetId, clearedDeficit);
+
+    assertEq(removedShares, expectedRemoveShares);
+    assertEq(hub.getDeficit(assetId), deficit - clearedDeficit);
+    assertEq(hub.getAssetSuppliedShares(assetId), assetSuppliedShares - expectedRemoveShares);
+    assertEq(
+      hub.getSpokeSuppliedShares(assetId, address(spoke1)),
+      spokeSuppliedShares - expectedRemoveShares
+    );
+    assertGe(getSupplyExRate(assetId), supplyExRate);
```
yan-man: perhaps can add some error msgs here and in other assertions in this test file for consistency

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  function test_clearDeficit_excess() public {
+    uint256 assetId = _randomAssetId(hub);
+    uint256 deficit = 1000e6;
+
+    _createDeficit(assetId, spoke1, deficit);
+    // arbitrarily inflate index
+    Utils.add(hub, assetId, address(spoke2), vm.randomUint(1, MAX_SUPPLY_AMOUNT), bob);
+    Utils.add(hub, assetId, address(spoke3), vm.randomUint(1, MAX_SUPPLY_AMOUNT), derl);
```
yan-man: should this be a borrow with skipTime to inflate index?

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
```diff
@@ -0,0 +1,135 @@
...
+
+    _supply(hub, spoke1, assetId, deficit);
+    assertGe(hub.getSpokeSuppliedAmount(assetId, address(spoke1)), deficit);
+
+    uint256 expectedRemoveShares = hub.previewRemoveByAssets(assetId, deficit);
+
+    vm.expectEmit(address(hub));
+    emit ILiquidityHub.DeficitCleared(assetId, address(spoke1), expectedRemoveShares, deficit);
+    vm.prank(address(spoke1));
+    uint256 removedShares = hub.clearDeficit(assetId, vm.randomUint(deficit, type(uint256).max));
```
yan-man: nit: we can just use `UINT256_MAX` in the tests

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
```diff
@@ -0,0 +1,135 @@
...
+    _addLiquidity(assetId, amount);
+    _drawLiquidityFromSpoke(address(spoke), assetId, amount, 322 days, true);
+    vm.prank(address(spoke));
+    hub.reportDeficit(assetId, amount, 0);
+
+    assertEq(hub.getDeficit(assetId), amount);
+  }
+
+  function _supply(
+    ILiquidityHub liqHub,
```
yan-man: nit: only place where we're using `liqHub` name i think so a bit odd

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
```diff
@@ -0,0 +1,135 @@
...
+    uint256 removedShares = hub.clearDeficit(assetId, vm.randomUint(deficit, type(uint256).max));
+
+    assertEq(removedShares, expectedRemoveShares);
+    assertEq(hub.getDeficit(assetId), 0);
+    assertBorrowRateSynced(hub, assetId, 'clearDeficit');
+  }
+
+  function _createDeficit(uint256 assetId, ISpoke spoke, uint256 amount) internal {
+    _addLiquidity(assetId, amount);
+    _drawLiquidityFromSpoke(address(spoke), assetId, amount, 322 days, true);
```
yan-man: perhaps we can set `322 days` as a constant in base to re-use, I know you mentioned wanting to use something other than 365 days everywhere

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
```diff
@@ -0,0 +1,135 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import './LiquidityHubBase.t.sol';
```
yan-man: nit: absolute path; can you also update this for reportDeficit test where I should have done that too

src/contracts/LiquidityHub.sol
yan-man, line 1: also I know it isn't part of this PR, but should some of the convert methods that Cheyenne added be external by default rather than public?

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
DhairyaSethi: it's arbitrary, don't think it needs to be a constant but we can do later. maybe call it default skip time? idk 
main point being different than 365 days is more interesting imo

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
DhairyaSethi: it's to avoid shadowing with 'hub' internal var, will rename this to liquidityHub as i've done w prior fixes

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
DhairyaSethi: yes good call ty

tests/unit/LiquidityHub/LiquidityHub.ClearDeficit.t.sol
DhairyaSethi: personally not a fan of that, they make it such that test file reads very cluttered. good for debugging & signalling info, will add whenever applicable

src/contracts/LiquidityHub.sol
DhairyaSethi: i'm agnostic, @CheyenneAtapour @avniculae thoughts?

src/contracts/LiquidityHub.sol
DhairyaSethi: all convert methods are external right now, which ones are you talking about?

yan-man: preview ones: previewRemoveByShares, previewDrawByShares, previewRestoreByShares, previewAddByShares

tests/Base.t.sol
```diff
@@ -1371,13 +1371,13 @@ abstract contract Base is Test {
     assertApproxEqAbs(
       actualPremiumDebt,
       expectedPremiumDebt,
-      2,
+      3,
```
yan-man, line 1374: any reason why clearing deficit would reduce precision here

src/contracts/LiquidityHub.sol
DhairyaSethi: ~they cannot be made external bc they're using by the normal operations (add, remove etc), if we make them external will have to call them with this.preview which will be needlessly expensive bc additional staticcall~

ah you mean the one's NOT used, sure we can, i like them as is for consistency with others but no strong opinion

tests/unit/LiquidityHub/LiquidityHub.Draw.t.sol
```diff
@@ -401,6 +403,7 @@ contract LiquidityHubDrawTest is LiquidityHubBase {
     updateDrawCap(hub, daiAssetId, address(spoke2), drawCap);
 
     _supplyAndDrawLiquidity({
+      liquidityHub: hub,
```
yan-man, line 406: think maybe we can follow spoke set up, and call it hub1, hub2 so we can keep the fields as hub rather than liquidityHub, as we simply won't have the concept of liquidityHubs after renaming

DhairyaSethi: yes want to do that only, but it's a lot of changes in the test suite hence haven't done it, this is temporary measure

tests/Base.t.sol
DhairyaSethi: unsure, i actually ran the failing test on a month old main commit and it needed this, just hasn't been caught by all of our main/pr ci fuzz runs until now lmao, still investigating but i think we can move forward here

src/contracts/LiquidityHub.sol
```diff
@@ -287,6 +287,29 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+  /// @inheritdoc ILiquidityHub
+  function eliminateDeficit(uint256 assetId, uint256 amount) external returns (uint256) {
+    DataTypes.Asset storage asset = _assets[assetId];
+    DataTypes.SpokeData storage spoke = _spokes[assetId][msg.sender];
+
+    asset.accrue(assetId, _spokes[assetId][asset.config.feeReceiver]);
+    _validateEliminateDeficit(spoke, amount);
+
+    uint256 deficit = asset.deficit;
+    if (amount > deficit) amount = deficit;
```
yan-man: hm we should also do this for the Remove/Restore as well

src/contracts/LiquidityHub.sol
yan-man, line 1: not part of this PR but we should also move `_applyPremiumDelta` down w the other internal methods

src/contracts/LiquidityHub.sol
DhairyaSethi: i think so, but changed to follow mostly the same convention for consistency for now

tests/Base.t.sol
yan-man: hm weird that our main CI wouldnt catch this either, given the high number of runs there

src/interfaces/ILiquidityHub.sol
```diff
@@ -217,6 +231,15 @@ interface ILiquidityHub is IAccessManaged {
     DataTypes.PremiumDelta calldata premiumDelta
   ) external returns (uint256);
 
+  /**
+   * @notice Eliminates deficit by removing supplied shares of caller spoke.
+   * @dev Only callable by active spokes.
+   * @param assetId The identifier of the asset.
+   * @param amount The amount of deficit to eliminate.
+   * @return The amount of shares removed.
```
CheyenneAtapour, line 239: This isn't a big deal, but should we call this variable `shares` instead of amount, so it's consistent with the preview functions

yan-man: hm but the input param to this function is an amount instead of a share

PR #589, opened 2025-07-31
feat: spoke/base minimal interfaces

### LINKED ISSUE #545: Spoke & Hub Standard
Create a minimal SpokeBase interface which is mandatory for all spokes to follow. This should contain all user actions and events; provide no restriction to governance methods, internal accounting or logic.
Brief:
- supply method
  - MUST emit Supply event
- borrow method
  - MUST emit Borrow event
- withdraw method
  - MUST emit Withdraw event
- restore method
  - MUST emit Restore event
- liquidationCall method
  - MUST emit liquidate event

Avoid:
- setUsingAsCollateral, updateUserRiskPremium, updateUserDynamicRiskConfiguration, gov methods etc

ISpoke should inherit ISpokeBase.
Util helpers should use ISpokeBase

yan-man:
- implement minimal interfaces for spoke/hub

closes #545

### DIFF
--- a/src/contracts/TreasurySpoke.sol
+++ b/src/contracts/TreasurySpoke.sol
@@ -5,7 +5,7 @@ import {Ownable} from 'src/dependencies/openzeppelin/Ownable.sol';
 import {SafeERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
 import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
 import {ILiquidityHub} from 'src/interfaces/ILiquidityHub.sol';
-import {ITreasurySpoke} from 'src/interfaces/ITreasurySpoke.sol';
+import {ITreasurySpoke, ISpokeBase} from 'src/interfaces/ITreasurySpoke.sol';
 
 /**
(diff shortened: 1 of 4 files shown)


src/contracts/TreasurySpoke.sol
```diff
@@ -60,4 +60,24 @@ contract TreasurySpoke is Ownable, ITreasurySpoke {
...
+  }
+
+  /// @inheritdoc ISpokeBase
+  function liquidationCall(
+    uint256 collateralReserveId,
+    uint256 debtReserveId,
+    address user,
+    uint256 debtToCover
+  ) external {
+    /// intentionally left blank
```
yan-man, line 81: left it like this for now, can create another PR to revert if desired, but should add some tests for that too

DhairyaSethi: let's add revert and tests for that in this PR itself please

yan-man: why? that is for spoke and we're trying to resolve hub first, prefer to just leave that for a separate pr 

https://github.com/aave/aave-v4/issues/592

PR #590, opened 2025-07-31
feat: Transfer added shares between spokes

### LINKED ISSUE #500: Investigate adding LH generic "move" method
For liquidation fees, these are achieved thru a method on spoke that allows using existing supplied shares from a spoke and transferring to the feeReceiver spoke. Investigate a more generic version of this method that allows any spoke to arbitrarily update supplied/borrowed shares accounting to another spoke. 

Use cases:
- migration during spoke off-boarding, to transfer accounting to the new spoke
- flexibility for allowing to move shares between spokes without needing actual token transfers

ref: question: what d be the difference between this function and a generic function `move` that just updates accounting between 2 spokes. Is the dedicated event for accrual the only difference? :thinking:

_Originally posted by @miguelmtzinf in https://github.com/aave/aave-v4/pull/467#discussion_r2175438341_

CheyenneAtapour:
Closes #500 

Depends on #468


src/contracts/LiquidityHub.sol
```diff
@@ -327,6 +327,25 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     emit Add(assetId, feeReceiver, feeShares, feeAmount);
   }
 
+  /// @inheritdoc ILiquidityHub
+  function moveSuppliedShares(uint256 assetId, uint256 shares, address toSpoke) external {
+    DataTypes.SpokeData storage fromSpoke = _spokes[assetId][msg.sender];
+    require(fromSpoke.config.active, SpokeNotActive());
+    DataTypes.SpokeData storage targetSpoke = _spokes[assetId][toSpoke];
+
+    require(shares > 0 && shares <= fromSpoke.suppliedShares, InvalidSharesAmount());
```
DhairyaSethi: pls move to validate internal method for consistency for now

src/contracts/LiquidityHub.sol
```diff
@@ -327,6 +327,25 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+    require(
+      targetSpoke.suppliedShares + _assets[assetId].toSuppliedAssetsDown(shares) <=
+        targetSpoke.config.supplyCap,
+      SupplyCapExceeded(targetSpoke.config.supplyCap)
+    );
```
DhairyaSethi: adding share and asset values, doesn't seem right, would need to convert the sum of shares into asset terms & compare

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
```diff
@@ -0,0 +1,100 @@
...
+
+  function test_moveSuppliedShares_exceedingCap_revertsWith_InvalidSharesAmount() public {
+    uint256 supplyAmount = 1000e18;
+    deal(address(tokenList.dai), address(spoke1), supplyAmount);
+    Utils.add(hub, daiAssetId, address(spoke1), supplyAmount, address(spoke1));
+
+    uint256 suppliedShares = hub.getSpokeSuppliedShares(daiAssetId, address(spoke1));
+    assertEq(suppliedShares, hub.convertToSuppliedAssets(daiAssetId, supplyAmount));
+
+    uint256 newSupplyCap = supplyAmount - 1;
```
DhairyaSethi: i would make this difference larger maybe so the bug could have been caught?

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
```diff
@@ -0,0 +1,100 @@
...
+    _updateSupplyCap(daiAssetId, address(spoke2), newSupplyCap);
+
+    // attempting transfer of supplied shares exceeding cap on spoke2
```
DhairyaSethi: ideally could check against remaining supply cap but nbd

src/contracts/LiquidityHub.sol
```diff
@@ -347,6 +347,25 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+
+    targetSpoke.suppliedShares += shares;
+    fromSpoke.suppliedShares -= shares;
+
+    emit MovedSuppliedShares(assetId, shares, msg.sender, toSpoke);
```
yan-man: consistency with other events like Add, Remove, etc. so prob should be MoveAddedShares

src/interfaces/ILiquidityHub.sol
```diff
@@ -231,6 +237,15 @@ interface ILiquidityHub is IAccessManaged {
...
   ) external returns (uint256);
 
+  /**
+   * @notice Allows a spoke to transfer its supplied shares of an asset to another spoke.
+   * @dev Only callable by spokes.
+   * @param assetId The identifier of the asset.
+   * @param shares The amount of shares to move.
+   * @param toSpoke The address of the spoke to move shares to.
+   */
+  function moveSuppliedShares(uint256 assetId, uint256 shares, address toSpoke) external;
```
yan-man: can we call it "added" instead of "supplied" to make the renaming easier later

src/contracts/LiquidityHub.sol
```diff
@@ -347,6 +347,25 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     emit Add(assetId, feeReceiver, feeShares, feeAmount);
   }
 
+  /// @inheritdoc ILiquidityHub
+  function moveSuppliedShares(uint256 assetId, uint256 shares, address toSpoke) external {
```
yan-man: not a huge fan of the name but dont have a better suggestion

src/contracts/LiquidityHub.sol
```diff
@@ -347,6 +347,25 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+    DataTypes.SpokeData storage fromSpoke = _spokes[assetId][msg.sender];
+    require(fromSpoke.config.active, SpokeNotActive());
```
yan-man: do we need to also validate the targetSpoke being active

src/contracts/LiquidityHub.sol
yan-man: in a similar way to payFee, should this just be Add/Remove events

src/contracts/LiquidityHub.sol
DhairyaSethi: same, can maybe call it transfer shares or something

src/contracts/LiquidityHub.sol
DhairyaSethi: yes we do @CheyenneAtapour

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
CheyenneAtapour: In these tests, the exchange rate is 1:1 anyway

src/contracts/LiquidityHub.sol
CheyenneAtapour: We decided today to keep this event, but I am renaming to `TransferredAddedShares`

src/contracts/LiquidityHub.sol

src/contracts/LiquidityHub.sol

src/interfaces/ILiquidityHub.sol
CheyenneAtapour: Done

src/contracts/LiquidityHub.sol
CheyenneAtapour: transferAddedShares now

src/contracts/LiquidityHub.sol

src/contracts/LiquidityHub.sol

src/contracts/LiquidityHub.sol
```diff
@@ -681,6 +665,23 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
+    DataTypes.SpokeData storage targetSpoke,
+    uint256 assetId,
+    uint256 shares
+  ) internal view {
+    require(fromSpoke.config.active && targetSpoke.config.active, SpokeNotActive());
+    require(shares > 0, InvalidSharesAmount());
+    require(shares <= fromSpoke.suppliedShares, SuppliedSharesExceeded(fromSpoke.suppliedShares));
+    require(
+      convertToSuppliedAssets(assetId, targetSpoke.suppliedShares) +
+        convertToSuppliedAssets(assetId, shares) <=
```
DhairyaSethi: i think we should convert the total share balance once bc we underestimate assets on each conversion and this is incorrect bc we can technically be over the addCap slightly and this wouldn't catch that

also we can pass asset storage pointer here directly and use explicit rounding for consistency with other validate abstractions

src/contracts/LiquidityHub.sol
```diff
@@ -336,34 +336,24 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
     receiver.suppliedShares += feeShares;
 
-    emit Remove(assetId, msg.sender, feeShares, feeAmount);
-    emit Add(assetId, feeReceiver, feeShares, feeAmount);
+    emit TransferredAddedShares(assetId, feeShares, msg.sender, feeReceiver);
   }
 
   /// @inheritdoc ILiquidityHub
-  function moveSuppliedShares(uint256 assetId, uint256 shares, address toSpoke) external {
+  function transferAddedShares(uint256 assetId, uint256 shares, address toSpoke) external {
```
DhairyaSethi: so in @yan-man naming PR, he's using `shares` in intermediary variables to signal supplied/added shares and `drawnShares` for debt shares, i think `transferShares` here is slightly cleaner and is ok imo if you're worried about confusion with drawnShares

DhairyaSethi: same for event, wdyt?

src/contracts/LiquidityHub.sol
```diff
@@ -336,34 +336,24 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-    require(shares > 0 && shares <= fromSpoke.suppliedShares, InvalidSharesAmount());
-    require(
-      targetSpoke.suppliedShares + _assets[assetId].toSuppliedAssetsDown(shares) <=
-        targetSpoke.config.supplyCap,
-      SupplyCapExceeded(targetSpoke.config.supplyCap)
-    );
+    _validateTransferAddedShares(fromSpoke, targetSpoke, assetId, shares);
 
     targetSpoke.suppliedShares += shares;
     fromSpoke.suppliedShares -= shares;
```
DhairyaSethi: i would prefer to do it like `payFee` where we cache sender's suppliedShares, do the balance check revert and commit the subtraction in storage using the cached value, it's cheaper (we can also do the sub in unchecked block in both after that)

also would call these `sender` and `receiver` instead of `targetSpoke` & `fromSpoke` like `payFee`

DhairyaSethi: note: the difference between `payFee` and `transferShares` is that `payFee` omits receiver spoke active check, receiver spoke addCap check

src/contracts/LiquidityHub.sol
```diff
@@ -336,15 +336,24 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
-    emit Add(assetId, feeReceiver, feeShares, feeAmount);
+    emit TransferredAddedShares(assetId, feeShares, msg.sender, feeReceiver);
+  }
+
+  /// @inheritdoc ILiquidityHub
+  function transferAddedShares(uint256 assetId, uint256 shares, address toSpoke) external {
+    DataTypes.SpokeData storage fromSpoke = _spokes[assetId][msg.sender];
+    DataTypes.SpokeData storage targetSpoke = _spokes[assetId][toSpoke];
+    _validateTransferAddedShares(fromSpoke, targetSpoke, assetId, shares);
+
```
DhairyaSethi, line 351: asset.accrue is missing

src/contracts/LiquidityHub.sol
yan-man: minor but Add, Remove etc have present tense events so perhaps `TransferAddedShares`?

src/contracts/LiquidityHub.sol
yan-man:
```suggestion
      convertToSuppliedAssets(assetId, targetSpoke.suppliedShares+shares) <= 
```

agreed, rather than 2 rounding ops

src/contracts/LiquidityHub.sol
yan-man: yeah and for present tense to be consistent (ie "TransferShares")

src/contracts/LiquidityHub.sol
yan-man: i think we could have an internal method to execute the logic that both payFee and transferShares use, and just the validation logic and receiver spoke differ between the 2

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
```diff
@@ -0,0 +1,102 @@
...
+pragma solidity ^0.8.0;
+
+import 'tests/unit/LiquidityHub/LiquidityHubBase.t.sol';
+
+contract LiquidityHubMoveSuppliedSharesTest is LiquidityHubBase {
+  function test_moveSuppliedShares() public {
+    test_moveSuppliedShares_fuzz(1000e18, 1000e18);
+  }
+
+  function test_moveSuppliedShares_fuzz(uint256 supplyAmount, uint256 moveAmount) public {
```
yan-man: renaming move -> transfer in downstream places

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
```diff
@@ -0,0 +1,102 @@
...
+    supplyAmount = bound(supplyAmount, 1, MAX_SUPPLY_AMOUNT);
+    moveAmount = bound(moveAmount, 1, supplyAmount);
+
+    // supply from spoke1
+    deal(address(tokenList.dai), address(spoke1), supplyAmount);
+    Utils.add(hub, daiAssetId, address(spoke1), supplyAmount, address(spoke1));
```
yan-man: minor - can add from user like Alice and rm the deal part i think

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
yan-man, line 1: also test name -> Transfer

src/contracts/LiquidityHub.sol
CheyenneAtapour: Done

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
```diff
@@ -0,0 +1,102 @@
...
+    vm.prank(address(spoke1));
+    hub.transferShares(daiAssetId, moveAmount, address(spoke2));
+
+    assertEq(hub.getSpokeSuppliedShares(daiAssetId, address(spoke1)), suppliedShares - moveAmount);
+    assertEq(hub.getSpokeSuppliedShares(daiAssetId, address(spoke2)), moveAmount);
+    assertEq(hub.getAssetSuppliedShares(daiAssetId), assetSuppliedShares);
+  }
+
+  /// @dev Test moving more shares than a spoke has supplied
+  function test_moveSuppliedShares_fuzz_exceedingSuppliedShares_revertsWith_SuppliedSharesExceeded(
```
yan-man:
```suggestion
  function test_moveSuppliedShares_fuzz_revertsWith_SuppliedSharesExceeded(
```

in addition to move->transfer

src/contracts/LiquidityHub.sol
CheyenneAtapour: Yeah, this is added

src/contracts/LiquidityHub.sol

tests/unit/Spoke/Spoke.Access.t.sol
```diff
@@ -33,6 +36,12 @@ contract SpokeAccessTest is SpokeBase {
     hub.restore(daiAssetId, 500e18, 0, DataTypes.PremiumDelta(0, 0, 0), address(spoke1));
     hub.remove(daiAssetId, 1000e18, address(spoke1));
     hub.refreshPremiumDebt(daiAssetId, DataTypes.PremiumDelta(0, 0, 0));
+    hub.add(daiAssetId, 1000e18, address(spoke1));
+    hub.transferShares(
```
DhairyaSethi, line 44: nice, let's add payFee here too for consistency since there as well sender spoke needs to be active. think it must already be checked there 

also the add above is unneeded bc spoke1 has added 1000 above

CheyenneAtapour: but it removed above too

DhairyaSethi: oh yeah 😄

src/contracts/LiquidityHub.sol

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
CheyenneAtapour: Done

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
CheyenneAtapour: Done

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
CheyenneAtapour: Done

tests/unit/LiquidityHub/LiquidityHub.MoveSuppliedShares.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Access.t.sol
CheyenneAtapour: Done

src/contracts/LiquidityHub.sol
```diff
@@ -559,6 +556,22 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
...
   }
 
+  function _transferShares(
+    DataTypes.Asset storage asset,
+    DataTypes.SpokeData storage sender,
+    DataTypes.SpokeData storage receiver,
+    uint256 assetId,
+    uint256 shares
+  ) internal {
+    asset.accrue(assetId, _spokes[assetId][asset.config.feeReceiver]);
```
yan-man: minor but I think we can do the accrue outside the `_transferShares` so that in payFee we can call `asset.accrue(assetId, receiver)` instead of needing to call `_spokes[assetId][asset.config.feeReceiver]` again

src/contracts/LiquidityHub.sol
```diff
@@ -662,6 +675,22 @@ contract LiquidityHub is ILiquidityHub, AccessManaged {
     require(feeShares != 0, InvalidFeeShares());
   }
 
+  function _validateTransferShares(
+    DataTypes.Asset storage asset,
+    DataTypes.SpokeData storage senderSpoke,
+    DataTypes.SpokeData storage receiverSpoke,
```
yan-man: I think can rename arg inputs to sender/receiver to match method

PR #593, opened 2025-08-04
feat: Check existence when adding reserve

### LINKED ISSUE #579: Add check for reserve existence on the Spoke
Even if `addReserve` function is permissioned, the function must check if the reserve is already listed.
Otherwise, a misconfiguration could overwrite the accounting of an existing reserve.
https://github.com/aave/aave-v4/blob/f58eb66b67e6ddc8c9790cf99b899b9d12d228d4/src/contracts/Spoke.sol#L89-L121

CheyenneAtapour:
* Adds a new internal mapping to the spoke which takes an address for a hub and an assetId and returns whether that reserve pairing has been added to the spoke via `addReserve`
* This mapping only ever sets existence of reserves to true (never sets any to false). We think this is fine because even when disabling spokes, we cannot be sure we've cleared the accounting
* Note that the default value of any pair of (hub, assetId) that doesn't exist yet will evaluate to false, which is desired.

closes #579


src/contracts/Spoke.sol
```diff
@@ -45,6 +45,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   mapping(uint256 reserveId => mapping(uint16 configKey => DataTypes.DynamicReserveConfig config))
     internal _dynamicConfig; // dictionary of dynamic configs per reserve
   DataTypes.LiquidationConfig internal _liquidationConfig;
+  mapping(address hub => mapping(uint256 assetId => bool exists)) internal _reserveExists;
```
DhairyaSethi, line 48: can we rename this var to something shorter & cleaner 
+ let's add a dev comment for now that this mapping is just used to existence check & nothing else 

something like _assetToReserve (we have similar assetToSpokes list mapping) maybe but will have to use use reserveId + 1 on the value bc 0 needs to be not found sentinel so something with bool like now is better

src/contracts/Spoke.sol
```diff
@@ -113,6 +115,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
       hub: IHub(hub)
     });
     _dynamicConfig[reserveId][dynamicConfigKey] = dynamicConfig;
+    _reserveExists[hub][assetId] = true;
```
DhairyaSethi, line 118: so this never resets, bc while disabling spokes we cannot ensure all accounting was cleared. makes sense. pls add this note + any caveats you think of the in the PR description (so we can reference later for dev docs)

src/contracts/Spoke.sol
CheyenneAtapour: I think this name is fitting and informative, especially since it resolves to a bool value, and it isn't a mapping that would return a reserveId or something.

Honestly, I'd argue the name makes it clear enough that it's only used to check existence of a reserve. But I can still add the dev comment if you're not fully convinced

avniculae:
subscribing to @DhairyaSethi 's comments, LGTM otherwise

PR #596, opened 2025-08-04
feat: consolidate common errors together

### LINKED ISSUE #535: Replace errors for non-zero address with InvalidAddress
We are using multiple errors when configuring and one of the input params is zero address.
Unless strictly necessary, we can simplify and use the same error code for all (`InvalidZeroAddress`).

In **LiquidityHub**:
- `addAsset`: `InvalidUnderlying`, `InvalidFeeReceiver` and `InvalidIrStrategy` checks
- `updateAssetConfig`: `InvalidFeeReceiver` and `InvalidIrStrategy` checks
- `addSpoke`: `InvalidSpoke` check

In **Spoke**:
- `updateOracle`: `InvalidOracle`
- `addReserve`: `InvalidHubAddress`

In **TreasurySpoke**:
- `constructor`: `InvalidHubAddress`

Fix tests and add whatever is needed to make sure the proper error is bubled up.
It may be more difficult to debug errors now... if it's hard to reason about errors on our tests, it means that we may need to revert this change.

yan-man:
- consolidate common errors; use internal method instead of modifier 
- currently changes account for only ~200 B
Screenshot 2025-08-23 at 00 37 49
Screenshot 2025-08-23 at 00 57 18

closes #535


src/interfaces/ISpokeBase.sol
```diff
@@ -8,9 +8,9 @@ pragma solidity ^0.8.0;
  */
 interface ISpokeBase {
   /**
-   * @notice Error thrown when the hub address is invalid.
+   * @notice Error thrown for invalid zero address.
    */
-  error InvalidHubAddress();
+  error InvalidZeroAddress();
```
DhairyaSethi: would move to ISpoke & IHub respectively, they don't need to be on base (if needed somewhere can be repeated)

src/contracts/Hub.sol
```diff
@@ -52,10 +52,11 @@ contract Hub is IHub, AccessManaged {
     address irStrategy,
     bytes calldata data
   ) external restricted returns (uint256) {
-    require(underlying != address(0), InvalidUnderlying());
+    require(
+      underlying != address(0) && feeReceiver != address(0) && irStrategy != address(0),
+      InvalidZeroAddress()
+    );
     require(decimals <= Constants.MAX_ALLOWED_ASSET_DECIMALS, InvalidAssetDecimals());
```
DhairyaSethi, line 59: we can also replace this with InvalidConfig, and reuse that for liqFee in updateAssetConfig 
wdyt?

DhairyaSethi: the idea with this suggestion was to cleanup single use errors on constructors & gov methods

src/contracts/Hub.sol
```diff
@@ -52,10 +52,11 @@ contract Hub is IHub, AccessManaged {
     address irStrategy,
     bytes calldata data
   ) external restricted returns (uint256) {
-    require(underlying != address(0), InvalidUnderlying());
+    require(
+      underlying != address(0) && feeReceiver != address(0) && irStrategy != address(0),
+      InvalidZeroAddress()
```
DhairyaSethi: let's keep this ZeroAddress or InvalidAddress instead imo, wdyt?

yan-man: you mean the renaming the error right? fine either way tbh, bc sometimes InvalidAddress can be non-zero address. But ok w ZeroAddress or InvalidZeroAddress

src/contracts/Hub.sol
yan-man: yeah wasn't sure how deep to go on re-using errors, rationale atm was that the zero address ones are all testing 0 address, but then this type of thing for config asset decimals can be bc some value is >= 18, whereas a diff config error can be for some other value issue. Unclear, open to thoughts @CheyenneAtapour @avniculae

src/contracts/Hub.sol
DhairyaSethi: > sometimes InvalidAddress can be non-zero address

not anywhere for us iirc, and we also use `InvalidSharesAmount` etc to signal when amount is 0

src/contracts/Spoke.sol
```diff
@@ -49,11 +49,6 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
   DataTypes.LiquidationConfig internal _liquidationConfig;
   mapping(address hub => mapping(uint256 assetId => bool exists)) internal _reserveExists;
 
-  modifier onlyPositionManager(address onBehalfOf) {
```
yan-man: found that changing this to internal saved ~100B, but can revert if this is preferred. Perhaps can get more significant savings from validationLib than doing this if it seems ugly, thoughts?

Screenshot 2025-08-24 at 13 56 38
Screenshot 2025-08-24 at 13 56 32

DhairyaSethi: prefer the current

src/contracts/Spoke.sol
```diff
@@ -108,7 +108,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     uint16 dynamicConfigKey; // 0 as first key to use
 
     DataTypes.Asset memory asset = IHub(hub).getAsset(assetId);
-    require(asset.underlying != address(0), AssetNotListed());
+    require(asset.underlying != address(0), InvalidAddress());
```
DhairyaSethi: this still needs to be what it was before i think, InvalidAddress makes sense if it's on an input

yan-man: https://github.com/aave/aave-v4/pull/596/commits/522cd2b650ab0ffc386be0fe5bb137d187ad72d6

src/contracts/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/596/commits/522cd2b650ab0ffc386be0fe5bb137d187ad72d6

Issue #503, opened 2025-07-02
test refactor - conditional assertions involving LH `restore`


yan-man:
update with conditional assertions - with the same coll/debt asset involved in a liquidation, `restore` creates wei rounding differences in calc of total debt that slightly change the ex rate and feeShare calc, making the exact value difficult to calc precisely. Also update this in other test cases where there is similar behavior.

Even in spoke repay tests, we're not able to exactly get the ex rate after a restore action, only within a wei. 

_Originally posted by @yan-man in https://github.com/aave/aave-v4/pull/467#discussion_r2175704457_

DhairyaSethi:
https://github.com/aave/aave-v4/pull/574/commits/e02912881ae084052d9ce56ad6026009a57bbe4f

PR #597, opened 2025-08-04
test: Hub Refresh Premium

### LINKED ISSUE #587: test: Hub.RefreshPremiumDebt
- snapshots
- validate does not change total debt by more than a 2 wei increase (decrease not allowed)

CheyenneAtapour:
Closes #587 
closes https://github.com/aave/aave-v4/issues/503


tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+    return PremiumDataLocal(asset.premiumShares, asset.premiumOffset, asset.realizedPremium);
+  }
+
+  function _applyPremiumDelta(
+    PremiumDataLocal memory premiumData,
+    DataTypes.PremiumDelta memory premiumDelta
+  ) internal pure returns (PremiumDataLocal memory) {
+    premiumData.premiumShares = uint256(
+      int256(premiumData.premiumShares) + premiumDelta.sharesDelta
+    );
```
DhairyaSethi: this can be done the same way it's done in source using the MathUtils.add method

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+      int256(premiumData.premiumOffset) + premiumDelta.offsetDelta
+    );
+    premiumData.realizedPremium = uint256(
+      int256(premiumData.realizedPremium) + premiumDelta.realizedDelta
+    );
+    return premiumData;
+  }
+
+  function _assertEq(PremiumDataLocal memory a, PremiumDataLocal memory b) internal pure {
+    assertEq(a.premiumShares, b.premiumShares, 'premium shares do not match');
```
DhairyaSethi:
```suggestion
  function assertEq(PremiumDataLocal memory a, PremiumDataLocal memory b) internal pure {
    assertEq(a.premiumShares, b.premiumShares, 'premium shares');
```

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+    assertEq(a.premiumOffset, b.premiumOffset, 'premium offset do not match');
+    assertEq(a.realizedPremium, b.realizedPremium, 'realized premium do not match');
+  }
+
+  function test_refreshPremium_revertsWith_SpokeNotActive() public {
+    DataTypes.PremiumDelta memory premiumDelta = DataTypes.PremiumDelta({
+      sharesDelta: 0,
+      offsetDelta: 0,
+      realizedDelta: 0
+    });
```
DhairyaSethi:
```suggestion
    DataTypes.PremiumDelta memory premiumDelta;
```
looks cleaner & achieves the same thing, or can populate with vm.randomInt()

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+    _assertEq(
+      _loadAssetPremiumData(daiAssetId),
+      _applyPremiumDelta(premiumDataBefore, premiumDelta)
+    );
+  }
+
+  /// @dev offsetDelta can't be more than sharesDelta or else underflow
+  /// @dev sharesDelta + realizedDelta can't be more than 2 more than offsetDelta
+  function test_refreshPremium(
+    uint256 sharesDelta,
```
DhairyaSethi: shouldn't we fuzz these as int's?

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+      offsetDeltaPos = asset.premiumOffset;
+    }
+
+    int256 sharesDelta = -int256(sharesDeltaPos);
+    int256 offsetDelta = -int256(offsetDeltaPos);
+    int256 realizedDelta = 0;
+    DataTypes.PremiumDelta memory premiumDelta = DataTypes.PremiumDelta({
+      sharesDelta: sharesDelta,
+      offsetDelta: offsetDelta,
+      realizedDelta: realizedDelta
```
DhairyaSethi: we don't need those intermediary vars, can just directly populate `premiumDelta` struct
also pls prefer to use safecast instead of manually casting in tests also (we decided on this as best practice, will add to style guide, benefit is sometimes bounds are incorrect and safecast will help catch that) even though it has no effect here

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+      _assertEq(
+        _loadAssetPremiumData(assetId),
+        _applyPremiumDelta(premiumDataBefore, premiumDelta)
+      );
+    }
+  }
+
+  function test_refreshPremium_negativeNumbers(
+    uint256 sharesDeltaPos,
+    uint256 offsetDeltaPos
```
DhairyaSethi: these are not really deltas but used to isolate a specific user position, right? maybe we can rename this, it's not super clear, and add comments if necessary

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+      realizedDelta: realizedDelta
+    });
+
+    vm.prank(address(spoke1));
+    hub1.refreshPremium(assetId, premiumDelta);
+
+    _assertEq(_loadAssetPremiumData(assetId), _applyPremiumDelta(premiumDataBefore, premiumDelta));
+  }
+
+  function test_refreshPremium_negativeNumbers_withAccrual(
```
DhairyaSethi: how about negative or negativeDeltas instead?

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+    bool reverting;
+
+    // Initial user position
+    userPremiumShares = bound(userPremiumShares, 0, asset.premiumShares);
+    userAccruedPremium = bound(
+      userAccruedPremium,
+      0,
+      hub1.convertToDrawnAssets(assetId, asset.premiumShares) - asset.premiumOffset
+    );
+    vm.assume(hub1.convertToDrawnAssets(assetId, userPremiumShares) >= userAccruedPremium);
```
DhairyaSethi, line 205: i don't think we need this assume, it is never true mathematically 
```solidity
// from first bound on L226
userAccruedPremium <= hub1.convertToDrawnAssets(assetId, asset.premiumShares) - asset.premiumOffset
// since asset.premiumOffset is an unsigned int, it's always >= 0, which implies the vm.assume
userAccruedPremium <= hub1.convertToDrawnAssets(assetId, asset.premiumShares) 
```

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+    uint256 userPremiumOffset = hub1.convertToDrawnAssets(assetId, userPremiumShares) -
+      userAccruedPremium;
+
+    // New user position
+    userPremiumSharesNew = bound(userPremiumSharesNew, 0, MAX_SUPPLY_AMOUNT / 2);
```
DhairyaSethi: it's a little weird here to bound shares to amount, i know why you did this, maybe we bound it some other constant like 1e30, and second to 1e30 - asset.premiumShares directly but nbd

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+    uint256 userPremiumOffsetNew = hub1.previewDrawByShares(assetId, userPremiumSharesNew);
+
+    DataTypes.PremiumDelta memory premiumDelta = DataTypes.PremiumDelta({
+      sharesDelta: int256(userPremiumSharesNew) - int256(userPremiumShares),
+      offsetDelta: int256(userPremiumOffsetNew) - int256(userPremiumOffset),
+      realizedDelta: int256(userAccruedPremium)
+    });
+
+    if (
```
DhairyaSethi, line 223: i dont think we need these cases if we follow the smt proof exactly (let's have one fuzz test exactly like that)

tests/unit/Hub/Hub.RefreshPremium.t.sol
DhairyaSethi, line 1: think we can also include https://gist.github.com/DhairyaSethi/428bc7156b14d4fe7490ce18a09c2e55 in this PR in `tests/misc` like we did with liquidity growth for now, wdyt

tests/unit/Hub/Hub.RefreshPremium.t.sol
DhairyaSethi: also let's move internal helpers at the end of the file for consistency?

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,267 @@
...
+    vm.expectEmit(address(hub1));
+    emit IHub.RefreshPremium(daiAssetId, address(spoke1), premiumDelta);
+
+    vm.prank(address(spoke1));
+    hub1.refreshPremium(daiAssetId, premiumDelta);
+
+    _assertEq(
+      _loadAssetPremiumData(daiAssetId),
+      _applyPremiumDelta(premiumDataBefore, premiumDelta)
+    );
```
DhairyaSethi: let's also assert the total debt before and after has only increased by at max two?

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Here I thought we wanted to keep all internal functions prefixed with `_`, but I suppose we can make an exception for assertEq, since it's basically an override?

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Yes. Also, this was intended just to be the positive deltas case. So I will update the function name to suggest as such

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: I didn't use that strategy here of isolating a specific user position. These are indeed the deltas. I just found it was easier for me to reason about the cases and do the logic when they were positive firstly, then I flip to negative and pass into the `premiumDelta` struct

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: I like negativeDeltas

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Hmm, is there something wrong with this approach then? We need this to be the case, otherwise the following line underflows

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: I can just convert that value (MAX_SUPPLY_AMOUNT / 2) to shares

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Yeah, for sure, I'm happy to include

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Hmm, I'm not sure here. My test does pretty much seem to exactly replicate the solver, or do things that produce equivalent bounds. Unsure why it has the underflow case

tests/gas/Hub.Operations.gas.t.sol
```diff
@@ -68,18 +68,16 @@ contract HubOperations_Gas_Tests is Base {
...
+  function test_refreshPremium() public {
+    vm.startPrank(address(spoke2));
+    hub1.add(daiAssetId, 1000e18, bob);
+    hub1.draw(daiAssetId, 500e18, bob);
+    vm.stopPrank();
+    vm.prank(address(spoke1));
+    hub1.draw(daiAssetId, 500e18, alice);
+    skip(100);
+    vm.prank(address(spoke1));
+    hub1.refreshPremium(daiAssetId, DataTypes.PremiumDelta(1, 1, 1));
```
DhairyaSethi: should we do a more realistic case? maybe i was thinking 
premiumSharesDelta = X
premiumOffsetDelta = X.toAssetsDown() (ie previewDrawByShares)
offsetDelta = 0

or something else, wdyt?

CheyenneAtapour: You mean realizedDelta = 0 at the end, but yeah I get you. I at least wanted everything non-zero. I feel it doesn't matter so much since the point is just to check gas, and as long as they are non-zero, I don't see an issue. 

But I'm happy to make it some other scenario that's more realistic if you like

DhairyaSethi: yep correct
I'm more thinking in the sense of when i am confused about a how a function works/is supposed to be called i go to its tests, the simplest test are usually in the <operation>.t.sol but in all cases if it's gas usage is checked it's present here so the first instinct is to always look at gas snapshots for a 101 example of how something works

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -0,0 +1,269 @@
...
+      .realizedPremium
+      .add(premiumDelta.realizedDelta)
+      .toUint128();
+    return premiumData;
+  }
+
+  function assertEq(PremiumDataLocal memory a, PremiumDataLocal memory b) internal pure {
+    assertEq(a.premiumShares, b.premiumShares, 'premium shares');
+    assertEq(a.premiumOffset, b.premiumOffset, 'premium offset');
+    assertEq(a.realizedPremium, b.realizedPremium, 'realized premium');
```
DhairyaSethi, line 267: for consistency we can add `assertEq(abi.encode(a), abi.encode(b))`

CheyenneAtapour: Done

tests/gas/Hub.Operations.gas.t.sol
CheyenneAtapour: understood, I adjusted it, lemme know what you think now

PR #602, opened 2025-08-05
feat: Pack Spoke Structs

avniculae:
- Packs spoke structs
- uses SafeCast in tests

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -45,6 +45,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
   mapping(uint256 reserveId => mapping(uint16 configKey => DataTypes.DynamicReserveConfig config))
     internal _dynamicConfig; // dictionary of dynamic configs per reserve
   DataTypes.LiquidationConfig internal _liquidationConfig;
+  mapping(address hub => mapping(uint256 assetId => bool exists)) internal _reserveExists;
 
   modifier onlyPositionManager(address onBehalfOf) {
     require(_isPositionManager({user: onBehalfOf, manager: msg.sender}), Unauthorized());
@@ -93,6 +94,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     DataTypes.DynamicReserveConfig calldata dynamicConfig
   ) external restricted returns (uint256) {
     require(hub != address(0), InvalidHubAddress());
+    require(!_reserveExists[hub][assetId], ReserveExists());
 
     _validateReserveConfig(config);
     uint256 reserveId = _reserveCount++;
@@ -115,6 +117,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
       collateralRisk: config.collateralRisk
     });
     _dynamicConfig[reserveId][dynamicConfigKey] = dynamicConfig;
+    _reserveExists[hub][assetId] = true;
 
     emit AddReserve(reserveId, assetId, hub);
     emit ReserveConfigUpdate(reserveId, config);
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -19,8 +19,6 @@ import {TreasurySpoke, ITreasurySpoke} from 'src/contracts/TreasurySpoke.sol';
 import {HubConfigurator, IHubConfigurator} from 'src/contracts/HubConfigurator.sol';
 import {SpokeConfigurator, ISpokeConfigurator} from 'src/contracts/SpokeConfigurator.sol';
 import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
-import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
-import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {SharesMath} from 'src/libraries/math/SharesMath.sol';
 import {MathUtils} from 'src/libraries/math/MathUtils.sol';
@@ -54,7 +52,6 @@ import {LibBit} from 'src/dependencies/solady/LibBit.sol';
 abstract contract Base is Test {
   using WadRayMath for uint256;
   using SharesMath for uint256;
-  using PercentageMath for uint256;
   using PercentageMath for uint256;
   using SafeCast for *;
 
@@ -1676,6 +1673,33 @@ abstract contract Base is Test {
     return spoke.getLiquidationConfig().closeFactor;
   }
 
+  function _calcMinimumCollAmount(
+    ISpoke spoke,
+    uint256 collReserveId,
+    uint256 debtReserveId,
+    uint256 debtAmount
+  ) internal view returns (uint256) {
+    if (debtAmount == 0) return 1;
+
+    IPriceOracle oracle = spoke.oracle();
+    DataTypes.Reserve memory collData = spoke.getReserve(collReserveId);
+    DataTypes.DynamicReserveConfig memory colDynConf = spoke.getDynamicReserveConfig(collReserveId);
+    uint256 collPrice = oracle.getReservePrice(collReserveId);
+    uint256 collAssetUnits = 10 ** hub1.getAsset(collData.assetId).decimals;
+
+    DataTypes.Reserve memory debtData = spoke.getReserve(debtReserveId);
+    uint256 debtAssetUnits = 10 ** hub1.getAsset(debtData.assetId).decimals;
+    uint256 debtPrice = oracle.getReservePrice(debtReserveId);
+
+    uint256 normalizedDebtAmount = (debtAmount * debtPrice).wadDivDown(debtAssetUnits);
+    uint256 normalizedCollPrice = collPrice.wadDivDown(collAssetUnits);
+
+    return
+      normalizedDebtAmount.wadDivUp(
+        normalizedCollPrice.toWad().percentMulDown(colDynConf.collateralFactor)
+      );
+  }
+
   /// @dev Helper function to borrow without health factor check
   function _borrowWithoutHfCheck(
     ISpoke spoke,
@@ -1744,11 +1768,10 @@ abstract contract Base is Test {
   }
 
   /// @dev Helper function to calculate burnt interest in assets terms (originated from virtual shares and assets)
-  function _calculateBurntInterest(
-    IHub hub,
-    uint256 assetId
-  ) internal view returns (uint256) {
-    return hub.getTotalAddedAssets(assetId) - hub.previewRemoveByShares(assetId, hub.getTotalAddedShares(assetId));
+  function _calculateBurntInterest(IHub hub, uint256 assetId) internal view returns (uint256) {
+    return
+      hub.getTotalAddedAssets(assetId) -
+      hub.previewRemoveByShares(assetId, hub.getTotalAddedShares(assetId));
   }
 
   /// @dev Helper function to withdraw fees from the treasury spoke
--- a/tests/unit/Spoke/SpokeBase.t.sol
+++ b/tests/unit/Spoke/SpokeBase.t.sol
@@ -167,7 +167,6 @@ contract SpokeBase is Base {
   }
 
   /// @dev Opens a debt position for a random user, using same asset as collateral and borrow
-  /// @return user address
   function _openDebtPosition(
     ISpoke spoke,
     uint256 reserveId,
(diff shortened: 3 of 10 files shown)


src/libraries/types/DataTypes.sol
```diff
@@ -99,6 +108,12 @@ library DataTypes {
     mapping(uint256 slot => uint256 status) map;
   }
 
+  struct LiquidationConfig {
+    uint128 closeFactor; // BPS, HF value to restore to during a liquidation
```
DhairyaSethi: close factor is not bps value, could you pls fix the comment

src/libraries/types/DataTypes.sol
```diff
@@ -99,6 +108,12 @@ library DataTypes {
...
+    uint64 healthFactorForMaxBonus; // health factor under which liquidation bonus is max
```
DhairyaSethi: how does this fit within 64 bits?

src/libraries/types/DataTypes.sol
```diff
@@ -59,34 +59,43 @@ library DataTypes {
...
+    IHub hub;
+    //
+    uint16 assetId;
+    bool paused;
+    bool frozen;
+    bool borrowable;
+    uint24 collateralRisk;
     uint16 dynamicConfigKey; // key of the last reserve config
     uint8 decimals;
     address underlying;
```
DhairyaSethi: kinda a nit but could we group the slot stuff together, thinking reserve config stuff at the end together such that if we were to append to the reserve config it all stays together (rn would have dyn configKey, decimals and underlying in bw for ex)

src/contracts/Spoke.sol
```diff
@@ -23,13 +24,14 @@ import {IAaveOracle} from 'src/interfaces/IAaveOracle.sol';
...
+  using PercentageMath for *;
   using KeyValueListInMemory for KeyValueListInMemory.List;
   using LiquidationLogic for DataTypes.LiquidationConfig;
   using PositionStatus for DataTypes.PositionStatus;
   using LiquidationLogic for DataTypes.LiquidationCallLocalVars;
-  using MathUtils for uint256;
+  using MathUtils for uint128;
 
   uint256 public constant HEALTH_FACTOR_LIQUIDATION_THRESHOLD = WadRayMath.WAD;
   uint256 public constant MAX_COLLATERAL_RISK = 1000_00; // 1000.00%
```
DhairyaSethi: let's move to Constants.sol

tests/Base.t.sol
```diff
@@ -73,9 +73,9 @@ abstract contract Base is Test {
   uint256 internal constant MIN_OPTIMAL_RATIO = 1_00; // 1.00% in BPS, matches AssetInterestRateStrategy
   uint256 internal constant MAX_OPTIMAL_RATIO = 99_00; // 99.00% in BPS, matches AssetInterestRateStrategy
   uint256 internal constant MAX_SKIP_TIME = 10_000 days;
-  uint256 internal constant MIN_LIQUIDATION_BONUS = PercentageMath.PERCENTAGE_FACTOR; // 100% == 0% bonus
-  uint256 internal constant MAX_LIQUIDATION_BONUS = 150_00; // 50% bonus
-  uint256 internal constant MAX_LIQUIDATION_BONUS_FACTOR = PercentageMath.PERCENTAGE_FACTOR; // 100%
+  uint32 internal constant MIN_LIQUIDATION_BONUS = uint32(PercentageMath.PERCENTAGE_FACTOR); // 100% == 0% bonus
+  uint32 internal constant MAX_LIQUIDATION_BONUS = 150_00; // 50% bonus
```
DhairyaSethi, line 74: is this a globally enforced limit? does that mean liq bonus can fit into uint16? if not, wondering why do we have this var here

src/libraries/types/DataTypes.sol
avniculae: we require `config.healthFactorForMaxBonus < HEALTH_FACTOR_LIQUIDATION_THRESHOLD`, where `HEALTH_FACTOR_LIQUIDATION_THRESHOLD` is 1e18. log2(1e18) = 59.7

DhairyaSethi: that constant can change but yeah only decrease so this should be safe. what's max close factor tho?

avniculae: we don't have a max close factor in the code, it just has to be above `HEALTH_FACTOR_LIQUIDATION_THRESHOLD` (1e18 now). We can make `closeFactor` smaller if we want, but atm there's no benefit in doing that, so I just decreased it to uint128

tests/Base.t.sol
avniculae: @yan-man do you have more insights here?

src/libraries/types/DataTypes.sol
avniculae: [eccba1f](https://github.com/aave/aave-v4/pull/602/commits/eccba1fdfbcf1b52899dfe0728ec1a35d0876867)

src/libraries/types/DataTypes.sol
avniculae: [eccba1f](https://github.com/aave/aave-v4/pull/602/commits/eccba1fdfbcf1b52899dfe0728ec1a35d0876867)

src/contracts/Spoke.sol
avniculae: [eccba1f](https://github.com/aave/aave-v4/pull/602/commits/eccba1fdfbcf1b52899dfe0728ec1a35d0876867)

src/contracts/Spoke.sol
```diff
@@ -23,16 +25,14 @@ import {IAaveOracle} from 'src/interfaces/IAaveOracle.sol';
 
 contract Spoke is ISpoke, Multicall, AccessManaged {
   using SafeERC20 for IERC20;
+  using SafeCast for *;
```
avniculae, line 28: FYI I used wildcard here and not just uint256 bc of this

https://github.com/aave/aave-v4/blob/461bb99d90ac1e0c047bb47b1f3cac3de45b321f/src/contracts/Spoke.sol#L282-L284

src/libraries/types/DataTypes.sol
```diff
@@ -59,34 +59,42 @@ library DataTypes {
...
+    // todo: remove reserveId
     uint256 reserveId;
-    uint256 assetId;
-    ReserveConfig config;
-    uint16 dynamicConfigKey; // key of the last reserve config
+    //
+    IHub hub;
+    //
+    uint16 assetId;
     uint8 decimals;
```
DhairyaSethi: the separator comment is incorrect; hub, assetId, and decimals will be in the same slot (which is good bc assetId & hub are always read together)

src/libraries/types/DataTypes.sol
```diff
@@ -59,34 +59,42 @@ library DataTypes {
...
     address underlying;
-    IHub hub;
-  }
-
-  struct ReserveConfig {
-    bool frozen;
+    uint16 dynamicConfigKey; // key of the last reserve config
```
DhairyaSethi, line 68: we can also move this in the first slot so there's more space in this second slot for potential future expansions but nbd bc it's fine even if we overflow later

src/libraries/types/DataTypes.sol
```diff
@@ -59,34 +59,42 @@ library DataTypes {
...
     uint256 reserveId;
-    uint256 assetId;
-    ReserveConfig config;
-    uint16 dynamicConfigKey; // key of the last reserve config
+    //
+    IHub hub;
+    //
+    uint16 assetId;
     uint8 decimals;
     address underlying;
```
DhairyaSethi: we currently use this for for existence check and it seems like only that pretty much, no other reason for `underlying` on reserve, so i think we can remove this & start using the `hub` address or something for existence instead

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -114,7 +115,7 @@ contract AssetInterestRateStrategyTest is Base {
 
   function test_setInterestRateData_revertsWith_InvalidMaxRate() public {
     rateData.baseVariableBorrowRate = rateData.variableRateSlope1 = rateData.variableRateSlope2 =
-      uint32(rateStrategy.MAX_BORROW_RATE()) /
+      rateStrategy.MAX_BORROW_RATE().toUint32() /
```
DhairyaSethi, line 118: probably this getter can just return a uint32, i would say that's better and consuming contracts can safely overcast if needed

tests/unit/Hub/Hub.Restore.t.sol
```diff
@@ -1422,7 +1422,7 @@ contract HubRestoreTest is HubBase {
 
     //     uint256 cumulatedBaseInterest = MathUtils.calculateLinearInterest(
     //       rate,
-    //       uint40(spoke1DaiData.lastUpdateTimestamp)
+    //       spoke1DaiData.lastUpdateTimestamp.toUint40()
```
DhairyaSethi, line 1425: pls tell me you wrote the regex expr this find and replace lmfao

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -623,13 +622,14 @@ contract SpokeBase is Base {
...
-    userPos.suppliedShares = hub1.convertToAddedShares(assetId, suppliedAmount);
+    userPos.drawnShares = hub1.convertToDrawnShares(assetId, debtAmount).toUint128();
+    userPos.premiumShares = hub1
+      .convertToDrawnShares(assetId, debtAmount)
+      .percentMulUp(riskPremium)
+      .toUint128();
+    userPos.premiumOffset = hub1.convertToDrawnAssets(assetId, userPos.premiumShares).toUint128();
+    userPos.realizedPremium = expectedRealizedPremium.toUint128();
+    userPos.suppliedShares = hub1.convertToAddedShares(assetId, suppliedAmount).toUint128();
   }
```
DhairyaSethi, line 605: ideally this helper should use another struct with all uint256's but we can tackle separately

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: hmm I remember we specifically wanted to keep uint256 in the IR strategy contract

tests/unit/Hub/Hub.Restore.t.sol
avniculae: I avoid regex replacements when possible to avoid unintended mistakes, I actually did this myself lol

tests/unit/Spoke/SpokeBase.t.sol
avniculae: why do you think it's better? to catch potential overflows?

src/libraries/types/DataTypes.sol
avniculae: [8e7d4ca](https://github.com/aave/aave-v4/pull/602/commits/8e7d4ca82cc586f90967ebf863605be90ed5165f)

src/libraries/types/DataTypes.sol
avniculae: n/a anymore I think [8e7d4ca](https://github.com/aave/aave-v4/pull/602/commits/8e7d4ca82cc586f90967ebf863605be90ed5165f)

src/libraries/types/DataTypes.sol
avniculae: this should be solved now [8e7d4ca](https://github.com/aave/aave-v4/pull/602/commits/8e7d4ca82cc586f90967ebf863605be90ed5165f)

src/contracts/Spoke.sol
```diff
@@ -1105,8 +1126,8 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
       vars.totalLiquidationFeeShares += vars.liquidationFeeShares;
 
       emit LiquidationCall(
-        vars.collateralUnderlying,
-        vars.debtUnderlying,
+        vars.collateralAssetId,
+        vars.debtAssetId,
```
DhairyaSethi, line 1133: imo this should emit reserve ids instead, unsure why we were doing the underlying before? @yan-man 
(if it's bc v3 does it, we have to recognise v3 uses underlying asset addresses as reserve identifiers and we use reserve ids)

tests/unit/Spoke/SpokeBase.t.sol
DhairyaSethi: yeah and not have these tests use & pass around internal data structures

tests/unit/AssetInterestRateStrategy.t.sol
DhairyaSethi: hmm, why tho? i don't have a strong opinion, fine w how it is also, just curious

this value is stored as a packed uint32 & casting up is safe 

regardless pr lgtm

src/contracts/Spoke.sol
avniculae: I agree, I think it makes more sense to emit the reserve ids

tests/unit/AssetInterestRateStrategy.t.sol
avniculae: It's a good point tbh, I think storing it as uint32 makes sense. I can't remember the exact reasoning from back then

src/contracts/Spoke.sol
avniculae: can we handle this in #605 @yan-man ?

DhairyaSethi: @yan-man https://github.com/aave/aave-v4/issues/608

PR #605, opened 2025-08-06
Feat: Prevent Liquidations Resulting in Dust Debt

### LINKED ISSUE #399: research - implement MIN_LEFTOVER_BASE threshold
see v3.3
- there is a final calc to determine if remaining debt or collateral after a liquidation is below a certain MIN_LEFTOVER_BASE threshold. If so, the operation reverts. (as well as forcing full liqs on either coll/debt) See if we want to implement something similar in v4

Image

- adds complexity but hard assurance that dust will not accumulate in protocol, coupled with #380 (which allows full liq of coll/debt based on starting health factor 

pros:
- no debt/coll dust remaining

cons:
- added complexity

notes:
- see LiquidationCallCloseFactorTest, at low values of dust coll supplied amount, can result in dust collateral remaining with bad debt. Update threshold for min amount and run fuzz tests. Can have dust on collateral fully being liquidated, but dust debt remains in deficit, due to rounding down in liquidation logic: https://github.com/aave/aave-v4/blob/9fed9cc17a294d68d71fa9b47d1f4ea9d8bc9cff/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.CloseFactor.t.sol#L407

### LINKED ISSUE #608: spoke: emit reserve id instead of asset id in liquidation call event

yan-man:
- prevent liquidations that result in dust debt unless all borrower's reserve collateral at hand is seized
- if liquidator input `debtToCover` is insufficient, revert. Otherwise, adjust debt to liquidate to a value that prevents dust to result 

closes #608 
closes #399

### DIFF
--- a/src/libraries/logic/LiquidationLogic.sol
+++ b/src/libraries/logic/LiquidationLogic.sol
@@ -62,11 +62,16 @@ library LiquidationLogic {
   ) internal pure returns (uint256) {
     uint256 maxLiquidatableDebt = debtToCover.min(params.totalBorrowerReserveDebt);
     uint256 actualDebtToLiquidate = maxLiquidatableDebt.min(params.debtToRestoreCloseFactor);
+
+    if (actualDebtToLiquidate == params.totalBorrowerReserveDebt) {
+      return actualDebtToLiquidate;
+    }
+
     uint256 remainingDebtInBaseCurrency = ((params.totalBorrowerReserveDebt -
       actualDebtToLiquidate) * params.debtAssetPrice).toWad() / params.debtAssetUnit;
 
-    // check for non zero debt dust remaining
-    if (remainingDebtInBaseCurrency < MIN_LEFTOVER_BASE && remainingDebtInBaseCurrency != 0) {
+    // check for (non zero) debt dust remaining
+    if (remainingDebtInBaseCurrency < MIN_LEFTOVER_BASE) {
       require(debtToCover >= params.totalBorrowerReserveDebt, MustNotLeaveDust());
       actualDebtToLiquidate = params.totalBorrowerReserveDebt;
     }
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -1,8 +1,8 @@
 {
   "borrow: first": "282483",
   "borrow: second action, same reserve": "267562",
-  "liquidationCall: full": "323708",
-  "liquidationCall: partial": "415193",
+  "liquidationCall: full": "323720",
+  "liquidationCall: partial": "415205",
   "repay: full": "189479",
   "repay: partial": "256010",
   "supply + enable collateral": "202528",


src/libraries/logic/LiquidationLogic.sol
```diff
@@ -43,19 +52,55 @@ library LiquidationLogic {
...
-    DataTypes.LiquidationCallLocalVars memory params
+    DataTypes.LiquidationCallLocalVars memory params,
+    uint256 debtToCover
   ) internal pure returns (uint256) {
-    uint256 maxLiquidatableDebt = params.totalDebt; // for current debt asset, in amount
-    uint256 debtToRestoreCloseFactor = params.calculateDebtToRestoreCloseFactor();
-    maxLiquidatableDebt = maxLiquidatableDebt.min(debtToRestoreCloseFactor);
-    return debtToCover.min(maxLiquidatableDebt);
+    uint256 maxLiquidatableDebt = debtToCover.min(params.totalBorrowerReserveDebt);
+    uint256 actualDebtToLiquidate = maxLiquidatableDebt.min(params.debtToRestoreCloseFactor);
```
yan-man, line 64: for reference:
- `debtToCover` is value input by liquidator
- `totalBorrowerReserveDebt` is the total debt for the user being liquidated in the debt reserve being repaid
- `debtToRestoreCloseFactor` is the amt of debt to repay to restore liquidated user to desired close factor

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -43,19 +52,55 @@ library LiquidationLogic {
...
+    console.log('LL actualDebtToLiquidate %e', actualDebtToLiquidate);
+    console.log('LL debtToRestoreCloseFactor %e', params.debtToRestoreCloseFactor);
+    console.log('LL actualDebtToLiquidate == debtToCover', actualDebtToLiquidate == debtToCover);
+
+    // only adjust actualDebtToLiquidate if there is non zero dust remaining
+    if (remainingDebtInBaseCurrency < MIN_LEFTOVER_BASE && remainingDebtInBaseCurrency != 0) {
+      console.log('LL dust');
+
+      // revert if debtToCover is the min value and has been set too low
+      if (debtToCover == actualDebtToLiquidate) {
```
yan-man: this means `debtToCover` is set too low by liquidator, thus revert. debtToCover is a hard cap, and we can never liquidate more than this value

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -43,19 +52,55 @@ library LiquidationLogic {
...
+        revert MustNotLeaveDust();
+      }
+
+      // at this point, debtToRestoreCloseFactor is min value
+      // if debtToCover also returns dust, revert
+      if (
+        debtToCover < params.totalBorrowerReserveDebt &&
+        ((params.totalBorrowerReserveDebt - debtToCover) * params.debtAssetPrice).toWad() /
+          params.debtAssetUnit <
+        MIN_LEFTOVER_BASE
```
yan-man: if it reaches this branch, it means that `debtToRestoreCloseFactor` is the min value. However, we are again hard capped by `debtToCover` so need to check if it is set to a value that is greater than `debtToRestoreCloseFactor` but still results in dust. If so revert

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -43,19 +52,55 @@ library LiquidationLogic {
...
+      ) {
+        revert MustNotLeaveDust();
+      } else {
+        // if debtToCover is valid, return min(debtToCover, totalBorrowerReserveDebt)
+        actualDebtToLiquidate = maxLiquidatableDebt;
```
yan-man: if `debtToCover` doesn't result in dust, increase the `actualDebtToLiquidate` to the max upper cap. Can never exceed `debtToCover` or `totalBorrowerReserveDebt`

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -136,6 +183,22 @@ library LiquidationLogic {
...
+      console.log('LL totalBorrowerReserveDebt %e', params.totalBorrowerReserveDebt);
+      console.log('LL debtAmountNeeded %e', vars.debtAmountNeeded);
+
+      vars.leftoverDebtBase =
+        ((params.totalBorrowerReserveDebt - vars.debtAmountNeeded) * params.debtAssetPrice)
+          .toWad() /
+        params.debtAssetUnit;
+
+      console.log('LL leftoverDebtBase %e', vars.leftoverDebtBase);
+      if (vars.leftoverDebtBase != 0 && vars.leftoverDebtBase < MIN_LEFTOVER_BASE) {
```
yan-man: we **must** also do this check bc of the scenario where debt value exceeds collateral value. Therefore max debt is not possible to be liquidated. 

Consider a multi-reserve case (assuming no liq bonus, no liq fee):
Collateral:
- asset A: $8k
- asset B: $5k (coll to seize)

Debt:
- asset C: $8k
- asset D: $5.4k (debt to liq)

if `actualDebtToLiquidate` ends up as $5.4k, ie liquidating all of asset D, by the time we come here to calc available collateral, there isn't enough collateral to liquidate to cover the whole debt and thus we can only liquidate $5k worth of asset D, leading to dust.

yan-man: resolved - this scenario is acceptable as long as all the reserve collateral is seized

src/contracts/Spoke.sol
```diff
@@ -318,13 +318,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     uint256[] memory debtsToCover = new uint256[](1);
     debtsToCover[0] = debtToCover;
```
yan-man, line 320: if we're also set on no parallel liqs, then can rm this array type input and the loop in liqCall (to be done in liquidation refactor)

src/libraries/logic/LiquidationLogic.sol
DhairyaSethi: at this point `maxLiquidatableDebt` is `totalBorrowerReserveDebt` (bc `debtToCover` > `totalborrowerReserveDebt` from the if part of this else branch, ie still respect liquidating within `debtToCover` input), right?
if we reach this branch, we are liquidating everything **on that reserve** and should not be left with dust, right?

yan-man: > bc debtToCover > totalborrowerReserveDebt

based on the if branch, which is basically `debtToCover < params.totalBorrowerReserveDebt && "remainingDebtFromDebtToCover leaving dust"`, then this branch is either: `debtToCover > totalBorrowerReserveDebt || remainingDebtFromDebtToCover not leaving dust`

so besides debtToCover > totalborrowerReserveDebt, it can also be that debtToCover < totalborrowerReserveDebt, but allows a remaining debt amount that doesn't leave dust, so a valid value. 

So here, `debtToRestoreCloseFactor < debtToCover < totalborrowerReserveDebt`, where `debtToRestoreCloseFactor` would leave dust but `debtToCover` wouldn't leave dust. Therefore adjust the `actualDebtToLiquidate` to min(debtToCover, totalborrowerReserveDebt) if debtToCover is valid

Not necessarily all of `totalborrowerReserveDebt`

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -13,6 +13,14 @@ library LiquidationLogic {
   using MathUtils for uint256;
   using LiquidationLogic for DataTypes.LiquidationCallLocalVars;
 
+  /**
+   * @dev This constant represents the minimum amount of assets in base currency that need to be leftover after a liquidation, if not clearing collateral on a position completely.
+   * @notice The default value assumes that the basePrice is usd denominated by 26 decimals.
+   */
+  uint256 constant MIN_LEFTOVER_BASE = 1000e26;
```
avniculae, line 20: currently this is the only assumption of the spoke that the aave oracle uses 8 decimals. I think we should hardcode 1000e18 and then multiply it by 10^(oracle.decimals()), wdyt?

DhairyaSethi: it's a pretty safe assumption, we can avoid doing this in the src code but good to add a test for this to flag non standard oracle decimals

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -43,19 +51,42 @@ library LiquidationLogic {
...
-    uint256 debtToRestoreCloseFactor = params.calculateDebtToRestoreCloseFactor();
-    maxLiquidatableDebt = maxLiquidatableDebt.min(debtToRestoreCloseFactor);
-    return debtToCover.min(maxLiquidatableDebt);
+    uint256 maxLiquidatableDebt = debtToCover.min(params.totalBorrowerReserveDebt);
+    uint256 actualDebtToLiquidate = maxLiquidatableDebt.min(params.debtToRestoreCloseFactor);
+    uint256 remainingDebtInBaseCurrency = ((params.totalBorrowerReserveDebt -
+      actualDebtToLiquidate) * params.debtAssetPrice).toWad() / params.debtAssetUnit;
+
+    // check for non zero debt dust remaining
+    if (remainingDebtInBaseCurrency < MIN_LEFTOVER_BASE && remainingDebtInBaseCurrency != 0) {
```
avniculae: i would rewrite this to `remainingDebtInBaseCurrency > 0 && remainingDebtInBaseCurrency < MIN_LEFTOVER_BASE`, but nbd

src/libraries/logic/LiquidationLogic.sol
avniculae: it can be the case that `params.debtToRestoreCloseFactor = debtToCover`, in which case the liquidator will need to input 1 wei more just to pass this check, since `actualDebtToLiquidate` will remain the same. hence, I would rewrite this as:
```
if (debtToCover < params.debtToRestoreCloseFactor)
```

wdyt?

we don't need to consider `params.totalBorrowerReserveDebt` since we checked `remainingDebtInBaseCurrency < MIN_LEFTOVER_BASE` above, so it must be the case that min is either `debtToCover` or `params.debtToRestoreCloseFactor`, right?

src/libraries/types/DataTypes.sol
```diff
@@ -170,14 +171,15 @@ library DataTypes {
...
-    uint256 userCollateralBalanceInBaseCurrency;
+    uint256 borrowerCollateralBalanceInBaseCurrency;
     uint256 baseCollateral;
     uint256 maxCollateralToLiquidate;
     uint256 collateralAmount;
     uint256 debtAmountNeeded;
     uint256 collateralToLiquidateInBaseCurrency;
     uint256 debtToLiquidateInBaseCurrency;
     bool hasDeficit;
+    uint256 leftoverDebtBase;
```
avniculae: is this used anywhere?

src/libraries/logic/LiquidationLogic.sol
CheyenneAtapour: just my 2 cents: I like the current order since the most important check is that there is a dust amount under the threshold

tests/unit/libraries/LiquidationLogic/LiquidationLogic.ActualDebtToLiquidate.Dust.t.sol
```diff
@@ -0,0 +1,374 @@
...
+  }
+
+  /// if totalBorrowerReserveDebt is the lowest value, then it is always returned, and unaffected by dust prevention
+  function test_calculateActualDebtToLiquidate_fuzz_totalBorrowerReserveDebt_lowest(
+    uint256 debtToCover,
+    TestDebtToRestoreCloseFactorParams memory params
+  ) public {
+    params = bound(params);
+    DataTypes.LiquidationCallLocalVars memory params = _setStructFields(params);
+    vm.assume(debtToCover > params.totalBorrowerReserveDebt);
```
avniculae, line 30: here we could bound debtToCover instead of assuming, wdyt?

src/libraries/logic/LiquidationLogic.sol
avniculae: [66deb06](https://github.com/aave/aave-v4/pull/605/commits/66deb06fcfc1d96e24ff5275def1ec5f777ec5b0)

src/libraries/types/DataTypes.sol
avniculae: [66deb06](https://github.com/aave/aave-v4/pull/605/commits/66deb06fcfc1d96e24ff5275def1ec5f777ec5b0)

src/libraries/logic/LiquidationLogic.sol
avniculae: It turns out that all that we care if there is any dust is for the debtCover to be >= totalDebt. This is bc even is the min value is the closeFactor, if debtToCover is < totalDebt then debtToCover must leave dust (bc closeFactor left dust, which is smaller). Simplified logic: [66deb06](https://github.com/aave/aave-v4/pull/605/commits/66deb06fcfc1d96e24ff5275def1ec5f777ec5b0)

tests/unit/libraries/LiquidationLogic/LiquidationLogic.ActualDebtToLiquidate.Dust.t.sol
avniculae: [6f36809](https://github.com/aave/aave-v4/pull/605/commits/6f36809dd9bc1d9fdf98d2b7e4e508d79ef004cb)

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -43,19 +51,27 @@ library LiquidationLogic {
...
+      require(debtToCover >= params.totalBorrowerReserveDebt, MustNotLeaveDust());
+      actualDebtToLiquidate = params.totalBorrowerReserveDebt;
+    }
+
+    return actualDebtToLiquidate;
```
DhairyaSethi: can we not optimise this further 
if params.totalBorrowerReserveDebt == actualDebtToLiquidate) return actualDebtToLiquidate

and now just compute remaining debt in base currency and do #69-72

avniculae: [5036781](https://github.com/aave/aave-v4/pull/605/commits/5036781c2c3d865465e5ae7d5d257e00df9eba0a)

PR #606, opened 2025-08-06
fix: fee share calc

DhairyaSethi:
closes: #607 

### Fix
 
Optimal:`base.rayMulUp(index2) + premium.rayMulUp(index2) - base.rayMulUp(index1) - premium.rayMulUp(index1)`
Optimised`base.rayMulDown(index2 - index1) + premium.rayMulDown(index2 - index1)`

SMT proof that the optimised fix always underestimates the accrued liquidity growth: https://gist.github.com/DhairyaSethi/cfbe18553ba9988aa754c37d952a58de

### Regression:
The existing tests have been fixed to catch the above two cases as mentioned in the issue. Replacing the current logic the previous two faulty iterations make the affected tests fail.
```solidity
    uint256 feesAmount = (
       (asset.drawnShares + asset.premiumShares).rayMulDown(indexDelta)
    ).percentMulDown(liquidityFee);

    uint256 feesAmount = (
       asset.drawnShares.rayMulDown(nextDrawnIndex - currentDrawnIndex) +
       asset.premiumShares.rayMulDown(nextDrawnIndex) - asset.premiumOffset
    ).percentMulDown(liquidityFee);
```


src/libraries/logic/AssetLogic.sol
```diff
@@ -68,7 +68,7 @@ library AssetLogic {
   }
 
   function totalAddedShares(DataTypes.Asset storage asset) internal view returns (uint256) {
-    return asset.addedShares + asset.getFeeShares(asset.getDrawnIndex(), asset.drawnIndex);
+    return asset.addedShares + asset.getFeeShares(asset.getDrawnIndex() - asset.drawnIndex);
```
DhairyaSethi, line 71: we can make this sub unchecked to shave off some more gas considering this is very frequently invoked in the lifecycle of any particular action

CheyenneAtapour: I know you want to wait until we do this all across the codebase, but honestly, imo that can delay our development, and we might forget to add these. So I'm happy for you to add unchecked here in this PR for this

src/libraries/logic/AssetLogic.sol
```diff
@@ -156,25 +156,22 @@ library AssetLogic {
...
-    uint256 nextDrawnIndex,
-    uint256 currentDrawnIndex
+    uint256 indexDelta
   ) internal view returns (uint256) {
+    if (indexDelta == 0) return 0;
     uint256 liquidityFee = asset.liquidityFee;
-    if (nextDrawnIndex == currentDrawnIndex || liquidityFee == 0) {
-      return 0;
-    }
+    if (liquidityFee == 0) return 0;
```
CheyenneAtapour, line 169: Is it not better to combine this and the check for indexDelta to be 0 and return 0 if either condition is met, similar to before?

src/libraries/logic/AssetLogic.sol
```diff
@@ -156,25 +156,22 @@ library AssetLogic {
...
 
     // liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee
     // prettier-ignore
     uint256 feesAmount = (
-      asset.drawnShares.rayMulDown(nextDrawnIndex - currentDrawnIndex) +
-      asset.premiumShares.rayMulDown(nextDrawnIndex) - asset.premiumOffset
+      asset.drawnShares.rayMulDown(indexDelta) +
+      asset.premiumShares.rayMulDown(indexDelta)
```
CheyenneAtapour: don't we still need to subtract out the `premiumOffset`?

src/libraries/logic/AssetLogic.sol
DhairyaSethi: would prefer independantly and do it all places it's relevant, tracking https://github.com/aave/aave-v4/issues/609

src/libraries/logic/AssetLogic.sol
DhairyaSethi: nope this is cheaper bc earlier in the case indexDelta is zero we still read liquidityFee unnecessarily from storage

src/libraries/logic/AssetLogic.sol
DhairyaSethi: no that's the incorrect implementation, pls see linked issue for a recap of the explanation
tldr premiumOffset != premiumShares.rayMulUp(lastIndex) for asset & spoke level (only true for user level accounting) bc we duplicate accrued for non active spokes/users & take fee from that twice

DhairyaSethi: example:
```
spoke A - premiumShares = 10, offset = 5
spoke B - premiumShares = 10, offset = 5
index = 1.2
accrued premium = 20 * 1.2 - 10 = 14 => fee taken on this
time passes, index = 1.4, spokeA accrues using a supply action
accrued premium = 20 * 1.4 - 10 = 18 => we cannot take fee on this 18 bc we've taken the fee on 14 already
actual untaxxed accrued premium = 20 * (1.4 - 1.2) = 4 
```

src/libraries/logic/AssetLogic.sol
```diff
@@ -156,25 +156,22 @@ library AssetLogic {
...
-      return 0;
-    }
+    if (liquidityFee == 0) return 0;
 
     // liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee
     // prettier-ignore
     uint256 feesAmount = (
-      asset.drawnShares.rayMulDown(nextDrawnIndex - currentDrawnIndex) +
-      asset.premiumShares.rayMulDown(nextDrawnIndex) - asset.premiumOffset
+      asset.drawnShares.rayMulDown(indexDelta) +
```
avniculae: can you pls add a dev comment to mention that this should not be compacted to (asset.drawnShares+asset.premiumShares).rayMulDown(indexDelta)?

I think one can easily think this the first time they see this code

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
```diff
@@ -166,32 +167,34 @@ contract SpokeAccrueLiquidityFeeEdgeCasesTest is SpokeBase {
...
     for (uint256 i; i < count; ++i) {
       address user = makeUser(i); // deterministic operation
       ISpoke spoke = spokes[i % spokes.length]; // deterministic operation
       uint256 reserveId = _reserveId(spoke, assetId);
-      uint256 debt = spoke.getUserTotalDebt(reserveId, user);
-      deal(spoke, reserveId, user, debt);
-      Utils.repay(spoke, reserveId, user, debt, user);
-      totalRepaid += debt;
-    }
+      Utils.repay(spoke, reserveId, user, 1, user); // accrue interest & realize premium
```
avniculae, line 189: how is this not reverting? repaying 1 asset should result in 0 shares repaid no? (given we skipped time)

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
```diff
@@ -166,32 +167,34 @@ contract SpokeAccrueLiquidityFeeEdgeCasesTest is SpokeBase {
...
+
+      uint256 totalOwedAfter = hub1.getAssetTotalOwed(assetId);
+      feesAccrued += totalOwedAfter - totalOwedBefore;
+      totalOwedBefore = totalOwedAfter;
 
-    uint256 feesAccruedToTreasury = hub1.getSpokeAddedAmount(assetId, address(treasurySpoke));
-    assertLe(feesAccruedToTreasury, totalRepaid - totalBorrowed, 'fees <= accrued');
-    assertApproxEqRel(feesAccruedToTreasury, totalRepaid - totalBorrowed, 0.0000001e18); // 0.00001%
+      uint256 actualFeesAccrued = hub1.getSpokeAddedAmount(assetId, address(treasurySpoke));
+      assertApproxEqRel(actualFeesAccrued, feesAccrued, 0.0000001e18); // 0.00001%
```
avniculae, line 196: when impl was wrong, `feesAccrued` was the correct value and `actualFeesAccrued` the wrong one right?

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
DhairyaSethi: repaying premium debt here

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
DhairyaSethi: yes correct haha, can rename these vars, `actual` just refers to what our implementation gives, and `expected` is the true value but open to suggestions if its confusing

src/libraries/logic/AssetLogic.sol
CheyenneAtapour: ok nice that we are tracking it in that issue at least

Issue #600, opened 2025-08-05
AssetInterestRateStrategy: include `deficit` & `sweeped`


DhairyaSethi:
Currently AssetInterestRateStrategy takes `liqudity`, `drawn` and `premium`. We should include passing `deficit` & `sweeped`.
`deficit` would be currently ignored in the IR strategy, `sweeped` will be added to `liquidity` as we don't want to infer it as withdrawn liquidity and retain the current utilisation.

PR #619, opened 2025-08-12
feat: reinvestment

### LINKED ISSUE #620: reinvestment feature

DhairyaSethi:
closes #620

### Design

Introduces two new hub methods:
- `sweep`: pulls available liquidity for the reinvestment strategy to generate yield on idle liquidity.
- `reclaim`: restore swept liquidity, only up to exact swept amount is allowed and callable by current reinvestment strategy.

### Notes

- The yield generated is expected to be distributed to suppliers taking additional risk completely offchain.
- The swept amount does not affect utilisation, ie `swept` should be included in `liquidity` here. (Note: In followup PR adjust IR strategy to be modular, #600)
- `addAsset` by default sets 0 reinvestment strategy, it is updatable through `updateAssetConfig`.
- Any additional caps, or protections for swept amount like the following are expected to be built using a strategy controller outside the core protocol.
  - Only allow a % of available liquidity at a time.
  - Sweeping more only allowed if existing swept has been completely repaid more than 1 block ago.

### DIFF
--- a/src/contracts/Hub.sol
+++ b/src/contracts/Hub.sol
@@ -316,8 +316,6 @@ contract Hub is IHub, AccessManaged {
     // no premium change allowed
     _applyPremiumDelta(asset, spoke, premiumDelta, 0);
 
-    asset.updateDrawnRate(assetId);
-
     emit RefreshPremium(assetId, msg.sender, premiumDelta);
   }
 
--- a/src/interfaces/IHub.sol
+++ b/src/interfaces/IHub.sol
@@ -42,14 +42,14 @@ interface IHub is IHubBase, IAccessManaged {
   event TransferShares(uint256 indexed assetId, uint256 shares, address sender, address receiver);
 
   /**
-   * @notice Emitted when an amount of liquidity is swept. (reinvested)
+   * @notice Emitted when an amount of liquidity is swept/reinvested.
    * @param assetId The identifier of the asset.
    * @param amount The amount swept.
    */
   event Sweep(uint256 indexed assetId, uint256 amount);
 
   /**
-   * @notice Emitted when an amount of liquidity is reclaimed. (from swept/reinvested liquidity)
+   * @notice Emitted when an amount of liquidity is reclaimed (from swept/reinvested liquidity).
    * @param assetId The identifier of the asset.
    * @param amount The amount reclaimed.
    */
--- a/tests/unit/Hub/Hub.Sweep.t.sol
+++ b/tests/unit/Hub/Hub.Sweep.t.sol
@@ -83,7 +83,7 @@ contract HubSweepTest is HubBase {
     updateAssetReinvestmentStrategy(hub1, daiAssetId, reinvestmentStrategy);
 
     _addLiquidity(daiAssetId, supplyAmount);
-    _drawLiquidity(daiAssetId, drawAmount, true);
+    _drawLiquidity(daiAssetId, drawAmount, false, false);
     uint256 swept = vm.randomUint(1, supplyAmount - drawAmount);
 
     uint256 drawnRate = hub1.getAssetDrawnRate(daiAssetId);
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -1,10 +1,10 @@
 {
-  "borrow: first": "289646",
-  "borrow: second action, same reserve": "274725",
-  "liquidationCall: full": "333032",
-  "liquidationCall: partial": "424505",
+  "borrow: first": "282483",
+  "borrow: second action, same reserve": "267562",
+  "liquidationCall: full": "325869",
+  "liquidationCall: partial": "417342",
   "repay: full": "189479",
-  "repay: partial": "263173",
+  "repay: partial": "256010",
   "supply + enable collateral": "202528",
   "supply: 0 borrows, collateral disabled": "140366",
   "supply: 0 borrows, collateral enabled": "143362",
@@ -12,12 +12,12 @@
   "supply: second action, same reserve": "109162",
   "updateUserDynamicConfig: 1 collateral": "54000",
   "updateUserDynamicConfig: 2 collaterals": "58612",
-  "updateUserRiskPremium: 1 borrow": "147309",
-  "updateUserRiskPremium: 2 borrows": "189945",
+  "updateUserRiskPremium: 1 borrow": "129646",
+  "updateUserRiskPremium: 2 borrows": "161119",
   "usingAsCollateral: 0 borrows, enable": "53792",
-  "usingAsCollateral: 1 borrow, disable": "159995",
+  "usingAsCollateral: 1 borrow, disable": "142332",
   "usingAsCollateral: 1 borrow, enable": "24717",
   "withdraw: 0 borrows, full": "145289",
   "withdraw: 0 borrows, partial": "151616",
-  "withdraw: 1 borrow, partial": "298019"
+  "withdraw: 1 borrow, partial": "286856"
 }
--- a/tests/unit/Hub/HubBase.t.sol
+++ b/tests/unit/Hub/HubBase.t.sol
@@ -87,7 +87,12 @@ contract HubBase is Base {
   }
 
   /// @dev Draws liquidity from the Hub via a random spoke
-  function _drawLiquidity(uint256 assetId, uint256 amount, bool withPremium) internal {
+  function _drawLiquidity(
+    uint256 assetId,
+    uint256 amount,
+    bool withPremium,
+    bool skipTime
+  ) internal {
     address tempSpoke = vm.randomAddress();
     address tempUser = vm.randomAddress();
 
@@ -109,7 +114,7 @@ contract HubBase is Base {
 
     Utils.draw(hub1, assetId, tempSpoke, tempUser, amount);
 
-    skip(365 days);
+    if (skipTime) skip(365 days);
 
     (uint256 drawn, uint256 premium) = hub1.getAssetOwed(assetId);
     assertGt(drawn, 0); // non-zero premium debt
@@ -123,6 +128,11 @@ contract HubBase is Base {
         DataTypes.PremiumDelta(-sharesDelta, -premiumOffsetDelta, int256(premium))
       );
     }
+  }
+
+  // @dev Draws liquidity from the Hub via a random spoke and skips time
+  function _drawLiquidity(uint256 assetId, uint256 amount, bool premium) internal {
+    _drawLiquidity(assetId, amount, premium, true);
   }
 
   /// @dev Draws liquidity from the Hub via a specific spoke which is already active


foundry.toml
```diff
@@ -11,6 +11,7 @@ optimizer = true
 optimizer_runs = 200
 bytecode_hash = "none"
 gas_snapshot_check = false
+gas_limit = 1099511627776
```
DhairyaSethi, line 14: original default gas_limit was 1073741824 (1024 ** 3)
needed bc `test_sharePriceWithMultipleDonations` runs out of gas, the inline configs to override this value or disable block gas limit aren't working

src/contracts/Hub.sol
```diff
@@ -81,14 +82,20 @@ contract Hub is IHub, AccessManaged {
       decimals: decimals,
       drawnRate: drawnRate.toUint128(),
       irStrategy: irStrategy,
+      reinvestmentStrategy: address(0),
```
CheyenneAtapour, line 85: I am scared about accidentally burning funds here, should we just check on `sweep` that this is not address(0)?

src/contracts/Hub.sol
```diff
@@ -619,7 +669,7 @@ contract Hub is IHub, AccessManaged {
     require(spoke.active, SpokeNotActive());
     uint256 withdrawable = previewRemoveByShares(assetId, spoke.addedShares);
     require(amount <= withdrawable, AddedAmountExceeded(withdrawable));
-    require(amount <= asset.liquidity, NotLiquidity(asset.liquidity));
+    require(amount <= asset.liquidity, InsufficientLiquidity(asset.liquidity));
```
CheyenneAtapour, line 670: Nice change here!

src/contracts/Hub.sol
```diff
@@ -705,4 +755,22 @@ contract Hub is IHub, AccessManaged {
...
     );
   }
+
+  function _validateSweep(
+    DataTypes.Asset storage asset,
+    address caller,
+    uint256 amount
+  ) internal view {
+    require(caller == asset.reinvestmentStrategy, InvalidReinvestmentStrategy());
+    require(amount > 0, InvalidSweepAmount());
```
CheyenneAtapour: do we also want `amount <= asset.liquidity`?

src/interfaces/IHub.sol
```diff
@@ -41,6 +41,20 @@ interface IHub is IHubBase, IAccessManaged {
...
 
+  /**
+   * @notice Emitted when an amount of liquidity is swept (reinvested).
+   * @param assetId The identifier of the asset.
+   * @param amount The amount swept.
+   */
+  event Sweep(uint256 indexed assetId, uint256 amount);
+
+  /**
+   * @notice Emitted when an amount of liquidity is reclaimed.
```
CheyenneAtapour: It's kinda unclear without context what this is. Want to say, "... is reclaimed from reinvestment"?

src/interfaces/IHub.sol
```diff
@@ -187,6 +203,22 @@ interface IHub is IHubBase, IAccessManaged {
    */
   function eliminateDeficit(uint256 assetId, uint256 amount) external returns (uint256);
 
+  /**
+   * @notice Sweeps an amount of liquidity of the corresponding asset and sends it to the configured reinvestment strategy.
+   * @dev The strategy handles the actual reinvestment of the funds and the redistribution of the interest, and investment caps.
```
CheyenneAtapour:
```suggestion
   * @dev The strategy handles the actual reinvestment of funds, redistribution of interest, and investment caps.
```

src/interfaces/IHub.sol
```diff
@@ -187,6 +203,22 @@ interface IHub is IHubBase, IAccessManaged {
...
+   * @param assetId The identifier of the asset.
+   * @param amount The amount swept.
+   */
+  function sweep(uint256 assetId, uint256 amount) external;
+
+  /**
+   * @notice Reclaims an amount of liquidity of the corresponding asset from the configured reinvestment strategy.
+   * @dev The strategy can only reclaim swept amount, all accrued interest is distributed offchain.
```
CheyenneAtapour:
```suggestion
   * @dev The strategy can only reclaim up to swept amount. All accrued interest is distributed offchain.
```

src/interfaces/IHub.sol
```diff
@@ -187,6 +203,22 @@ interface IHub is IHubBase, IAccessManaged {
...
+   * @param assetId The identifier of the asset.
+   * @param amount The amount swept.
```
CheyenneAtapour:
```suggestion
   * @param amount The amount to reclaim.
```

src/interfaces/IHub.sol
```diff
@@ -187,6 +203,22 @@ interface IHub is IHubBase, IAccessManaged {
    */
   function eliminateDeficit(uint256 assetId, uint256 amount) external returns (uint256);
 
+  /**
+   * @notice Sweeps an amount of liquidity of the corresponding asset and sends it to the configured reinvestment strategy.
+   * @dev The strategy handles the actual reinvestment of the funds and the redistribution of the interest, and investment caps.
+   * @param assetId The identifier of the asset.
+   * @param amount The amount swept.
```
CheyenneAtapour:
```suggestion
   * @param amount The amount to sweep.
```

Future tense because we are about to do it by passing these params to this function. Makes more sense as past tense when describing an event imo (since that's post-transaction), but nbd

tests/Base.t.sol
```diff
@@ -914,17 +920,31 @@ abstract contract Base is Test {
...
 
     vm.prank(HUB_ADMIN);
-    targetHub.updateAssetConfig(assetId, config);
+    hub.updateAssetConfig(assetId, config);
 
-    assertEq(targetHub.getAssetConfig(assetId), config);
+    assertEq(hub.getAssetConfig(assetId), config);
+  }
+
+  function updateAssetReinvestmentStrategy(
```
CheyenneAtapour, line 936: imo we should have newly introduced internal functions prefixed with `_` so we can reduce the refactor work later

src/contracts/Hub.sol
DhairyaSethi: how do you mean - burning funds at what stage?

src/contracts/Hub.sol
```diff
@@ -705,4 +755,22 @@ contract Hub is IHub, AccessManaged {
       AddCapExceeded(addCap)
     );
   }
+
+  function _validateSweep(
+    DataTypes.Asset storage asset,
+    address caller,
+    uint256 amount
+  ) internal view {
+    require(caller == asset.reinvestmentStrategy, InvalidReinvestmentStrategy());
```
CheyenneAtapour: I think we can consider renaming this error and the similar one in `_validateReclaim` to something like `OnlyReinvestmentStrategy` to make it clear this is a condition on the caller

src/contracts/Hub.sol
CheyenneAtapour: Ah nvm, I see caller must be the `reinvestmentStrategy`, so can't happen

tests/Base.t.sol
DhairyaSethi: this is likely the last such method to be introduced, let's do it all together now

src/interfaces/IHub.sol
```diff
@@ -41,6 +41,20 @@ interface IHub is IHubBase, IAccessManaged {
   event AccrueFees(uint256 indexed assetId, uint256 shares);
   event TransferShares(uint256 indexed assetId, uint256 shares, address sender, address receiver);
 
+  /**
+   * @notice Emitted when an amount of liquidity is swept by the reinvestment strategy.
```
CheyenneAtapour: Hate to nit pick again, but it's unclear what "swept" means. The "(reinvested)" is helpful like in `getSwept()`

src/interfaces/IHub.sol
```diff
@@ -187,6 +204,22 @@ interface IHub is IHubBase, IAccessManaged {
...
+   * @param assetId The identifier of the asset.
+   * @param amount The amount to sweep.
+   */
+  function sweep(uint256 assetId, uint256 amount) external;
+
+  /**
+   * @notice Reclaims an amount of liquidity of the corresponding asset from the configured reinvestment strategy.
+   * @dev The strategy can only reclaim up to swept amount. All accrued interest is distributed offchain.
+   * @param assetId The identifier of the asset.
+   * @param amount The amount reclaim.
```
CheyenneAtapour:
```suggestion
   * @param amount The amount to reclaim.
```

src/interfaces/IHub.sol
DhairyaSethi: fixed lmk

src/interfaces/IHub.sol
```diff
@@ -41,6 +41,20 @@ interface IHub is IHubBase, IAccessManaged {
   event AccrueFees(uint256 indexed assetId, uint256 shares);
   event TransferShares(uint256 indexed assetId, uint256 shares, address sender, address receiver);
 
+  /**
+   * @notice Emitted when an amount of liquidity is swept. (reinvested)
```
CheyenneAtapour:
```suggestion
   * @notice Emitted when an amount of liquidity is swept (reinvested).
```

src/interfaces/IHub.sol
```diff
@@ -41,6 +41,20 @@ interface IHub is IHubBase, IAccessManaged {
...
+   * @param assetId The identifier of the asset.
+   * @param amount The amount swept.
+   */
+  event Sweep(uint256 indexed assetId, uint256 amount);
+
+  /**
+   * @notice Emitted when an amount of liquidity is reclaimed. (from swept/reinvested liquidity)
```
CheyenneAtapour:
```suggestion
   * @notice Emitted when an amount of liquidity is reclaimed (from swept/reinvested liquidity).
```

src/contracts/Hub.sol
```diff
@@ -304,6 +316,8 @@ contract Hub is IHub, AccessManaged {
     // no premium change allowed
     _applyPremiumDelta(asset, spoke, premiumDelta, 0);
 
+    asset.updateDrawnRate(assetId);
```
CheyenneAtapour: We decided in sync today not to do this right? Because this adds more gas than necessary to make a precise drawn rate, but we are fine approximating and saving gas, as long as we document in the corresponding issue about this.

DhairyaSethi: correct, being documented at #621 
rmed [bd8533a](https://github.com/aave/aave-v4/pull/619/commits/bd8533a4292525b1cc9e8db9ddda754c582c7a00)

src/contracts/Hub.sol
```diff
@@ -229,10 +241,10 @@ contract Hub is IHub, AccessManaged {
     asset.accrue(assetId, _spokes[assetId][asset.feeReceiver]);
     _validateRestore(asset, spoke, assetId, drawnAmount, premiumAmount, from);
 
-    _applyPremiumDelta(asset, spoke, premiumDelta, premiumAmount);
     uint128 drawnShares = previewRestoreByAssets(assetId, drawnAmount).toUint128();
     asset.drawnShares -= drawnShares;
     spoke.drawnShares -= drawnShares;
+    _applyPremiumDelta(asset, spoke, premiumDelta, premiumAmount);
```
avniculae, line 247: this change is cosmetic rather than a logic bug right?

src/contracts/Hub.sol
```diff
@@ -337,6 +349,36 @@ contract Hub is IHub, AccessManaged {
...
+  function sweep(uint256 assetId, uint256 amount) external {
+    DataTypes.Asset storage asset = _assets[assetId];
+    _validateSweep(asset, msg.sender, amount);
+
+    asset.accrue(assetId, _spokes[assetId][asset.feeReceiver]);
+    asset.liquidity -= amount.toUint128();
+    asset.swept += amount.toUint128();
+    asset.updateDrawnRate(assetId);
+
+    IERC20(asset.underlying).safeTransfer(asset.reinvestmentStrategy, amount);
```
avniculae, line 362: here we could use msg.sender to save ~100 gas

src/contracts/Hub.sol
```diff
@@ -337,6 +349,36 @@ contract Hub is IHub, AccessManaged {
...
+  function reclaim(uint256 assetId, uint256 amount) external {
+    DataTypes.Asset storage asset = _assets[assetId];
+    _validateReclaim(asset, msg.sender, amount);
+
+    asset.accrue(assetId, _spokes[assetId][asset.feeReceiver]);
+    asset.liquidity += amount.toUint128();
+    asset.swept -= amount.toUint128();
+    asset.updateDrawnRate(assetId);
+
+    IERC20(asset.underlying).safeTransferFrom(asset.reinvestmentStrategy, address(this), amount);
```
avniculae, line 377: here we could use msg.sender to save ~100 gas

tests/unit/HubConfigurator.t.sol
```diff
@@ -602,6 +604,24 @@ contract HubConfiguratorTest is HubBase {
...
+    hubConfigurator.updateReinvestmentStrategy(address(hub1), vm.randomUint(), vm.randomAddress());
+  }
+
+  function test_updateReinvestmentStrategy() public {
+    address reinvestmentStrategy = makeAddr('newReinvestmentStrategy');
+    DataTypes.AssetConfig memory expectedConfig = hub1.getAssetConfig(assetId);
+    expectedConfig.reinvestmentStrategy = reinvestmentStrategy;
+    vm.expectCall(address(hub1), abi.encodeCall(IHub.updateAssetConfig, (assetId, expectedConfig)));
+    vm.prank(HUB_CONFIGURATOR_ADMIN);
+    hubConfigurator.updateReinvestmentStrategy(address(hub1), assetId, reinvestmentStrategy);
```
avniculae, line 622: in this file we also read the asset config after action and asset it, nbd

tests/unit/Hub/Hub.Sweep.t.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    vm.prank(reinvestmentStrategy);
+    hub1.sweep(daiAssetId, sweepAmount);
+
+    assertEq(hub1.getSwept(daiAssetId), sweepAmount);
+    assertEq(hub1.getLiquidity(daiAssetId), assetLiquidity - sweepAmount);
+    assertBorrowRateSynced(hub1, daiAssetId, 'sweep');
+  }
+
+  function test_swept_amount_is_not_withdrawable() public {
```
avniculae, line 61: this name does not follow the naming convention

src/contracts/Hub.sol
DhairyaSethi: yes

DhairyaSethi: i realised now that the conversion is using the debt share price which is unaffected by the ordering bc its index

src/contracts/Hub.sol
DhairyaSethi: thought as well, figured this was more clear and ok with loosing that gas since it's a gov (kind of) fn
but agnostic, ok with updating

avniculae: I'm fine keeping the way it is

DhairyaSethi:
@avniculae comments being addressed in #624 
tysm!

PR #623, opened 2025-08-13
opt: shrink drawn rate to 96 bits

DhairyaSethi:


src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -14,7 +14,7 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
   using WadRayMath for *;
 
   /// @inheritdoc IAssetInterestRateStrategy
-  uint256 public constant MAX_BORROW_RATE = 1000_00; // 1000.00% in BPS
+  uint256 public constant MAX_BORROW_RATE = 7922_81; // 7922.81% in BPS
```
avniculae: why?

DhairyaSethi: oh yeah forgot to update this was a miscalc, not needed

PR #624, opened 2025-08-13
feat: Add deficit and swept to IR Strategy

### LINKED ISSUE #600: AssetInterestRateStrategy: include `deficit` & `sweeped`

CheyenneAtapour:
closes #600


tests/unit/Hub/Hub.Sweep.t.sol
```diff
@@ -99,7 +99,9 @@ contract HubSweepTest is HubBase {
         assetId: daiAssetId,
         liquidity: supplyAmount - drawAmount,
```
DhairyaSethi:
```suggestion
        liquidity: supplyAmount - drawAmount - swept,
```

tests/unit/AssetInterestRateStrategy.t.sol
```diff
@@ -248,8 +272,12 @@ contract AssetInterestRateStrategyTest is Base {
     liquidity = drawn.rayMulUp(WadRayMath.RAY - targetUtilizationRatioRay).rayDivUp(
       targetUtilizationRatioRay
     );
+    // Take a random portion of liquidity as swept
+    swept = liquidity > 0 ? vm.randomUint(0, liquidity) : 0;
```
DhairyaSethi:
```suggestion
    swept = vm.randomUint(0, liquidity);
```
slightly cleaner

CheyenneAtapour: True

tests/unit/Hub/Hub.Sweep.t.sol
CheyenneAtapour: Ah I see, thanks mate!

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -110,7 +112,7 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
       return currentVariableBorrowRateRay;
     }
 
-    uint256 usageRatioRay = drawn.rayDivUp(liquidity + drawn);
+    uint256 usageRatioRay = drawn.rayDivUp(liquidity + drawn + swept);
```
avniculae, line 115: why don't we add deficit as well here?

tests/Base.t.sol
```diff
@@ -2049,7 +2062,16 @@ abstract contract Base is Test {
...
   ) internal {
-    _mockInterestRateRay(address(irStrategy), interestRateRay, assetId, liquidity, drawn, premium);
+    _mockInterestRateRay(
+      address(irStrategy),
+      interestRateRay,
+      assetId,
+      liquidity,
+      drawn,
+      premium,
+      0,
```
avniculae, line 2076: I think you should take these as args

src/contracts/AssetInterestRateStrategy.sol
DhairyaSethi: bc it is not available to be drawn 

we make an exception for `swept` by definition as being virtual liquidity still contributing to the utilisation, bc it's purpose to utilise the "idle" capital (if it's not included here then it would act as withdrawn)

avniculae: > bc it is not available to be drawn

why does it matter? This is relevant only for remove/draw actions. Here we overestimate the rate imo since we don't add the deficit (while in share price calc we do), wdyt?

DhairyaSethi: if it's not available to be drawn it cannot contribute to the drawn:available ratio

PR #628, opened 2025-08-15
feat: cache reserve on the stack

DhairyaSethi:


DhairyaSethi:
we decided to not move forward with this primarily bc the data is sparsed & gas savings here alone aren't worth it

PR #642, opened 2025-08-18
fix: Liquidation logic rounding

### LINKED ISSUE #460: Investigate rounding direction in liquidations
- At the moments the liquidation logic uses dewadify down. We should investigate whether we should do dewadify up instead.

avniculae:
Closes #460


src/libraries/logic/LiquidationLogic.sol
```diff
@@ -142,10 +142,9 @@ library LiquidationLogic {
         params.debtAssetUnit;
     } else {
       // add 1 to round collateral amount up, ensuring HF is always <= close factor
-      vars.collateralAmount =
-        ((vars.maxCollateralToLiquidate * params.collateralAssetUnit) / params.collateralAssetPrice)
-          .fromWadDown() +
-        1;
+      vars.collateralAmount = (vars.maxCollateralToLiquidate * params.collateralAssetUnit)
+        .wadDivUp(params.collateralAssetPrice.toWad())
+        .fromWadUp();
```
DhairyaSethi:
```suggestion
      vars.collateralAmount = (vars.maxCollateralToLiquidate * params.collateralAssetUnit)
        .wadDivUp(params.collateralAssetPrice);
```
equivalent, but do we loose precision this way?

also we just round up bc of the comment (and prev + 1) above, no? but wondering if we should enforce that somewhere else bc as with the convention to underestimate collateral & overestimate debt during HF calc

avniculae: I don't think they are equivalent: `vars.maxCollateralToLiquidate` is in WAD, hence `vars.maxCollateralToLiquidate * params.collateralAssetUnit` is in WAD. the suggestion above would do `wadNumber * WAD / collateralAssetPrice` (let's ignore rounding for now), which means number will have 36 decimals precision, right?

avniculae: > also we just round up bc of the comment (and prev + 1) above, no?

no, I tried (my best) to ignore existing comments and logic. I think when calculating the user's **existing** collateral and debt, we should underestimate collateral and overestimate debt. However, when calculating the amounts to liquidate, I think we should do the opposite: seize as much collateral as possible (hence overestimate collateral) for as little debt as possible (hence underestimate debt). wdyt?

src/libraries/math/WadRayMath.sol
```diff
@@ -179,16 +179,27 @@ library WadRayMath {
...
     }
   }
 
+  /**
+   * @notice Converts number from Wad precision by rounding up
+   * @param a The number in Wad precision
+   * @return b (= a / 1e18, rounded up)
+   */
+  function fromWadUp(uint256 a) internal pure returns (uint256 b) {
+    assembly {
```
DhairyaSethi: nice, could you tag all of these assembly blocks with "memory-safe". it's applicable here, not a lot to optimise here anyway so i don't expect (any) savings, but good practice 

https://docs.soliditylang.org/en/v0.8.30/assembly.html#memory-safety

avniculae: > Inline assembly that neither involves any operations that access memory nor assigns to any Solidity variables in memory is automatically considered memory-safe and does not need to be annotated.

This means that the compiler already considers this block memory-safe, even though we didn't tag it, right? In any case, I agree it's good to add these to be consistent

src/libraries/logic/LiquidationLogic.sol
DhairyaSethi: ah you're right, not equivalent, just don't like that we're rounding twice - unsure of a way this creates any discrepency
wdyt `vars.collateralAmount = vars.maxCollateralToLiquidate.mulDivUp(params.collateralAssetUnit, params.collateralAssetPrice.toWad());` (in 256 bit using a ceilDiv can also work)

avniculae: yeah, thought about `mulDivUp` as well, but we currently don't have such an implementation

src/libraries/math/WadRayMath.sol
DhairyaSethi: yes correct

src/libraries/logic/LiquidationLogic.sol
DhairyaSethi: we can add for this imo, would prefer over rounding twice generally

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -142,10 +142,9 @@ library LiquidationLogic {
...
       vars.debtAmountNeeded = params.actualDebtToLiquidate;
       vars.collateralToLiquidateInBaseCurrency =
         (vars.collateralAmount * params.collateralAssetPrice).toWad() /
```
DhairyaSethi: should we also round up here then if 
> when calculating the amounts to liquidate: seize as much collateral as possible for as little debt as possible

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -89,21 +89,20 @@ library LiquidationLogic {
   ) internal pure returns (uint256) {
     // represents the effective value loss from the collateral per unit of debt repaid
     // the greater the penalty, the more debt must be repaid to restore the user's health factor
-    uint256 effectiveLiquidationPenalty = (params.liquidationBonus.toWad())
-      .percentMulDown(params.collateralFactor)
-      .fromBpsDown();
+    uint256 effectiveLiquidationPenalty = params.liquidationBonus.bpsToWad().percentMulDown(
```
yan-man, line 94: just to clarify this one doesnt change the rounding but is just formatting right? ie both round down

DhairyaSethi: it's just slightly less precise from before 

before rounded down twice, the second time was with more precision

src/libraries/logic/LiquidationLogic.sol
avniculae: as we discussed, I think we want to round down here, as this is only used to assess if there is deficit

src/libraries/logic/LiquidationLogic.sol
avniculae: I don't think any precision is lost here, which also means that rounding direction does not matter here. this is bc both of these values are in bps values, and the result can be represented in wad without any loss of precision. percentMulDown will divide by 1e4, which will remove a few 0s from the 1e18 wad precision.

src/libraries/math/MathUtils.sol
```diff
@@ -77,4 +77,31 @@ library MathUtils {
       return a - b;
     }
   }
+
+  // performs 256 bit
```
DhairyaSethi: can you make the natspec more consistent with the others

src/libraries/math/MathUtils.sol
```diff
@@ -77,4 +77,31 @@ library MathUtils {
...
+  function mulDivDown(uint256 a, uint256 b, uint256 c) internal pure returns (uint256 d) {
```
DhairyaSethi, line 89: don't really need this but sure np

src/libraries/math/MathUtils.sol
avniculae: [23b5101](https://github.com/aave/aave-v4/pull/642/commits/23b5101c1a354da1967deb280dd7d3787465a64f)

src/contracts/Spoke.sol
```diff
@@ -1048,107 +1040,98 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+
+    if (vars.hasDeficit) {
+      _reportDeficits(user);
+    } else {
+      // new risk premium only needs to be propagated if no deficit exists
+      (vars.newUserRiskPremium, , , , ) = _calculateUserAccountData(user);
+      _notifyRiskPremiumUpdate(user, vars.newUserRiskPremium);
     }
+
+    if (vars.withdrawnShares - vars.liquidatedSuppliedShares > 0) {
```
yan-man: nit: minor opt?
```suggestion
    if (vars.withdrawnShares > vars.liquidatedSuppliedShares) {
```

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -842,10 +843,10 @@ contract SpokeLiquidationBase is SpokeBase {
         params.debtAssetUnit;
     } else {
       // add 1 to round collateral amount up, ensuring HF is always <= close factor
```
yan-man: rm this comment

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
```diff
@@ -825,12 +826,12 @@ contract SpokeLiquidationBase is SpokeBase {
...
-      params.debtAssetUnit;
+    vars.baseCollateral = (params.actualDebtToLiquidate * params.debtAssetPrice).wadDivUp(
+      params.debtAssetUnit
+    );
 
     // account for additional collateral required due to liquidation bonus
-    vars.maxCollateralToLiquidate = vars.baseCollateral.percentMulDown(params.liquidationBonus);
+    vars.maxCollateralToLiquidate = vars.baseCollateral.percentMulUp(params.liquidationBonus);
 
     if (vars.maxCollateralToLiquidate >= vars.borrowerCollateralBalanceInBaseCurrency) {
```
yan-man, line 836: previously made this into a >= due to the adding 1 from the bottom branch... i think this now can be > strictly aligning w v3

https://github.com/aave-dao/aave-v3-origin/blob/63d8a8f34aa756d6e8c699259529f05cbc9f4abb/src/contracts/protocol/libraries/logic/LiquidationLogic.sol#L631

tests/unit/libraries/LiquidationLogic/LiquidationLogic.AvailableCollateralToLiquidate.t.sol
```diff
@@ -331,11 +333,9 @@ contract LiquidationAvailableCollateralToLiquidateTest is LiquidationLogicBaseTe
...
   ) internal pure returns (uint256) {
     return
-      _convertAmountToBaseCurrency(
-        params.actualDebtToLiquidate,
-        params.debtAssetPrice,
-        params.debtAssetUnit
-      ).percentMulDown(params.liquidationBonus);
+      (params.actualDebtToLiquidate * params.debtAssetPrice)
+        .wadDivUp(params.debtAssetUnit)
+        .percentMulUp(params.liquidationBonus);
```
yan-man, line 338: can this not be same as before but `percentMulUp`? or does the `_convertAmountToBaseCurrency` helper need to be updated too?

tests/unit/libraries/LiquidationLogic/LiquidationLogic.AvailableCollateralToLiquidate.t.sol
```diff
@@ -207,7 +209,7 @@ contract LiquidationAvailableCollateralToLiquidateTest is LiquidationLogicBaseTe
     params.debtAssetUnit = 0;
     DataTypes.LiquidationCallLocalVars memory args = setStructFields(params);
 
-    vm.expectRevert(stdError.divisionError);
+    vm.expectRevert(); // WadMathRay reverts with no data if division by 0
```
yan-man:
```suggestion
    vm.expectRevert(); // WadRayMath reverts with no data if division by 0
```

tests/unit/libraries/LiquidationLogic/LiquidationLogic.AvailableCollateralToLiquidate.t.sol
avniculae: given that we need to maximise collateral seized, we are rounding up when converting amount to base currency. However, I don't think we should modify the `_convertAmountToBaseCurrency` function since you still want to round down when converting collateral in most cases

tests/unit/libraries/LiquidationLogic/LiquidationLogic.AvailableCollateralToLiquidate.t.sol
avniculae: I always do this lol

tests/unit/Spoke/Liquidations/Spoke.Liquidation.Base.t.sol
avniculae: I think it's actually better to do `>=` now, bc we want to correctly assess deficit (if we do `>` then it might be that `maxCollateralToLiquidate` is equal to `borrowerCollateralBalanceInBaseCurrency`, but bc of precision errors you don't report deficit)

src/contracts/Spoke.sol
avniculae: I expect the compiler to optimise this, but will address

tests/unit/libraries/LiquidationLogic/LiquidationLogic.AvailableCollateralToLiquidate.t.sol
DhairyaSethi: agree^
think i had added another helper called convertDebtAmountToBaseCurrency which rounded up but this is fine also

PR #644, opened 2025-08-19
feat: permitReserve & setUserPositionManagerWithSig

### LINKED ISSUE #576: Permit functions on Spoke
- Add permit functions to Spoke
- Evaluate which pattern we wanna use (normal permit vs permit2 vs ...)
- Does anything to do with PositionManager?

DhairyaSethi:
closes #576 

- uses Solady EIP712 to save on codesize & runtime gas costs.
  - contract is backwards compatible when we add upgradeability (see `_cachedDomainSeparatorInvalidated`)
- adds `permitReserve` to facilitate supplyWithPermit & repayWithPermit flows using multicall
  - spender is enforced as the reserve's hub
  - had to reintroduce `underlying` on the reserve struct since we consume permit on the token. gas diff is negligible since it takes the first slot & is never used anywhere except here
- adds `setUserPositionManagerWithSig ` to approve/revoke approval for a posm through a 712-typed signature
  - enables having complete gas sponsorship UX using posm - only 1 user sig approving posm would be required
  - uses forge's new `eip712` cheatcodes to validate typehash integrity

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -19,6 +19,7 @@ include = ["src/libraries/types/EIP712Types.sol"]
 
 [lint]
 ignore = ["src/dependencies/**/*", "tests/**/*"]
+lint_on_build = false
 
 [profile.default.fuzz]
 runs = 1000


tests/gas/Spoke.Operations.gas.t.sol
```diff
@@ -19,11 +19,6 @@ contract SpokeOperations_Gas_Tests is Base {
 
   function test_supply() public {
     vm.startPrank(alice);
-    vm.startSnapshotGas('Spoke.Operations', 'supply + enable collateral');
-    spoke1.supply(_daiReserveId(spoke1), 100e18, alice);
-    spoke1.setUsingAsCollateral(_daiReserveId(spoke1), true, alice);
-    vm.stopSnapshotGas();
-
```
yan-man, line 27: why this part removed?

tests/unit/Spoke/Spoke.SetUserPositionManagerWithSig.t.sol
```diff
@@ -0,0 +1,210 @@
...
+      params.user,
+      params.approve,
+      params.deadline,
+      v,
+      r,
+      s
+    );
+  }
+
+  function test_setUserPositionManagerWitSig() public {
```
yan-man:
```suggestion
  function test_setUserPositionManagerWithSig() public {
```

tests/gas/Spoke.Operations.gas.t.sol
DhairyaSethi: bc it's replaced with multicall below, this was showcasing two independant txns total gas, but would always batch through multicall now (for non smart accounts)

src/contracts/Spoke.sol
```diff
@@ -382,6 +409,32 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+    uint256 deadline,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external {
+    DataTypes.Reserve storage reserve = _reserves[reserveId];
+    address underlying = reserve.underlying;
+    require(underlying != address(0), ReserveNotListed());
+    try
+      IERC20Permit(underlying).permit({
```
yan-man, line 454: could also read from hub asset instead of reintroducing underlying field; any opposition to that approach?

DhairyaSethi: had it before but it costs more to read unnecessarily (bc this is supposed to be used as supply with permit & repay with permit flows of v3 with the multicall)

and the cost of reintroducing underlying here is minimal, bc it takes an independent slot and doesn't affect existing gas costs

tests/unit/Spoke/Spoke.SetUserPositionManagerWithSig.t.sol
```diff
@@ -0,0 +1,211 @@
...
+      params.deadline,
+      v,
+      r,
+      s
+    );
+
+    assertEq(spoke1.isPositionManager(user, params.positionManager), params.approve);
+  }
+
+  function test_useNonce_monotonic(bytes32) public {
```
yan-man, line 199: is this bytes32 input needed here

DhairyaSethi: it's not an input, it's a symbolic test (see setArbitraryStorage) this just makes it run for higher number of runs 
the fuzz input isn't used but rather symbolic value is generated through vm.random*

yan-man: hm not exactly a symbolic test i think, but i see - fuzz input just used for runs

src/contracts/Spoke.sol
```diff
@@ -1239,6 +1298,24 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
+    return ('Spoke', '1');
+  }
+
+  function _useNonce(address user) internal returns (uint256) {
+    unchecked {
+      return _nonces[user]++;
+    }
+  }
+
+  function _setUserPositionManager(address positionManager, address user, bool approve) private {
```
yan-man: nit: private method perhaps after internal

tests/gas/Spoke.Operations.gas.t.sol
```diff
@@ -194,4 +188,104 @@ contract SpokeOperations_Gas_Tests is Base {
     vm.snapshotGasLastCall('Spoke.Operations', 'updateUserDynamicConfig: 2 collaterals');
     vm.stopPrank();
   }
+
+  function test_multicall_ops() public {
```
yan-man, line 192: think we should add equivalent of this into spoke.supply and spoke.repay tests or maybe in spoke multicall

tests/unit/Spoke/Spoke.SetUserPositionManagerWithSig.t.sol
DhairyaSethi: no this is exactly how to write symbolic tests with foundry, the runner just needs to be an smt to solve it and forge just does the random storage bc of `vm.setArbitraryStorage(address(spoke1));`

src/libraries/types/EIP712Types.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: BUSL-1.1
```
yan-man: should we use a diff license here?

tests/unit/Spoke/Spoke.SetUserPositionManagerWithSig.t.sol
DhairyaSethi: it's just to prove that nonce always increases by one

tests/gas/Spoke.Operations.gas.t.sol
DhairyaSethi: we do have some on spoke.multicall but yeah we can more scenarios later, not worth the effort rn

src/libraries/types/EIP712Types.sol
DhairyaSethi: yeah good catch fixing

tests/gas/Spoke.Operations.gas.t.sol
yan-man: yeah doesnt necessarily need to be in this pr

src/contracts/Spoke.sol
```diff
@@ -382,6 +409,32 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     emit SetUserPositionManager(onBehalfOf, msg.sender, false);
   }
 
+  /// @inheritdoc ISpoke
+  function permitReserve(
```
yan-man, line 441: a *little* bit misleading bc spender is actually the reserve hub but understand

DhairyaSethi: yeah open to naming suggestion, i think current makes sense tbh 

your suggestion is permitReserveHub? can do tbh

Issue #609, opened 2025-08-07
opt: unchecked arithmetic wherever applicable


DhairyaSethi:
- AssetLogic: all index delta cals

PR #646, opened 2025-08-19
opt: optimally order hub struct, uncheckedSub

DhairyaSethi:
addresses #609


src/contracts/Hub.sol
```diff
@@ -702,40 +699,38 @@ contract Hub is IHub, AccessManaged {
...
   }
 
   function _validateEliminateDeficit(
-    DataTypes.Asset storage asset,
     DataTypes.SpokeData storage spoke,
     uint256 amount
   ) internal view {
     require(spoke.active, SpokeNotActive());
-    require(amount != 0 && amount <= asset.deficit, InvalidDeficitAmount());
+    require(amount > 0, InvalidDeficitAmount());
```
yan-man, line 725: question about this update, does the amount <= asset.deficit happen elsewhere and is unneeded here @DhairyaSethi ?

DhairyaSethi: happens on line 303

Issue #72, opened 2025-01-15
Settle on Solidity version


yan-man:
Some considerations:

> using a newer version can be problematic for i) integrations and ii) chain support.
> For i, solvable if we create interfaces/libs with flexible enough sol version. For ii, better off checking if it’s really an issue or it was just push0 the only one that caused issues in the past.

> for ii) it’s not a matter of solc but rather evm_version wrt supported opcodes, we can stick to an older globally supported version like paris. although idt push0 is an issue on most chains anymore as discussed before. 
> having a higher solc is good for optimizations and most importantly bug fixes, there have been known issues in the compiler with as late as 0.8.22 https://github.com/ethereum/solidity/blob/develop/docs/bugs_by_version.json#L1881
> so imo we should take the last stable version, if going with require(customError) then the last version which allowed its use w/o ir pipeline would be best imio

Reference: https://www.notion.so/V4-Code-Conventions-17a9d63a22de80898fb0ee1f6df6038b?d=17a9d63a22de80e9b460001c90f280d4&pvs=4#17a9d63a22de80d1b208c57aa1183ca9

DhairyaSethi:
note: specify optimizer settings explicitly, more info see https://github.com/bgd-labs/aave-proposals-v3/issues/570

DhairyaSethi:
IMP NOTE
currently we use `> 0` (& gt(X, 0)) for non zero checks bc it's cheaper on push0 chains than != 0 (since gt(x, 0) is gt(2), push0(3), and iszero(iszero(X)) is iszero (3) x2)

yan-man:
done in https://github.com/aave/aave-v4/pull/791

PR #647, opened 2025-08-19
fix: license updated on .sol files

CanonicalJP:
updated license in all .sol files to 

```
// SPDX-License-Identifier: UNLICENSED
```


src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -1,4 +1,5 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
 pragma solidity ^0.8.10;
```
yan-man, line 3: can be done on a separate PR but we should maintain some consistency on the solidity versions across the files too: https://github.com/aave/aave-v4/issues/72

Issue #645, opened 2025-08-19
reduce spoke codesize


DhairyaSethi:
main opts:
- move logic to ext lib
- via_ir config separate compilation profile for the spoke

PR #648, opened 2025-08-20
feat: extract user account logic to external lib

DhairyaSethi:
for #645


The-3D:
closed as it has been replaced by #666

PR #664, opened 2025-08-22
fix: spoke fixes

### LINKED ISSUE #652: spoke: oracle set improvements
- why is it not immutable?

1. https://github.com/aave/aave-v4/blob/d79ea65552a0d870d7f046d1fd6fa2e6160be579/src/contracts/Spoke.sol#L73 - move constant (8) to Constants lib

2. 
https://github.com/aave/aave-v4/blob/d79ea65552a0d870d7f046d1fd6fa2e6160be579/src/contracts/Spoke.sol#L70-L75
```
This function is not validating that the newOracle oracle is also supporting all the EXISTING assets already supported by this Spoke and the previously replaced oracle.

I think that any changes made by functions guarded by the restricted should never break the core functionality of a "working" Spoke.
```
either document or fix on configurator

### LINKED ISSUE #656: spoke.calcUserAccData isUsingAsCollateral branch
1.
https://github.com/aave/aave-v4/blob/d79ea65552a0d870d7f046d1fd6fa2e6160be579/src/contracts/Spoke.sol#L784

this branch should be `if (collateralFactor != 0 && isUsingAsCollateral)`

v3: https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/protocol/libraries/logic/GenericLogic.sol#L111

2.
remove line 842 here, unneeded
https://github.com/aave/aave-v4/blob/d79ea65552a0d870d7f046d1fd6fa2e6160be579/src/contracts/Spoke.sol#L841-L842

3.
cache `oracle` since it's a state var

4. getUserDebtInBaseCurrency should round both base & premium debt separately

DhairyaSethi:
closes #656 
closes #652


tests/unit/Spoke/Spoke.Borrow.Scenario.t.sol
```diff
@@ -576,4 +576,36 @@ contract SpokeBorrowScenarioTest is SpokeBase {
       label: 'bob dai data after'
     });
   }
+
+  function test_health_factor_does_not_include_zero_cf_collateral() public {
```
yan-man: maybe `test_userAccountData_...`, as this also involves in general `totalCollateralInBaseCurrency`

src/contracts/Spoke.sol
```diff
@@ -581,15 +582,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
 
   /**
    * @dev Calculates the user's premium debt offset in assets amount from a given share amount.
-   * @dev Rounds down to the nearest assets amount.
-   * @dev Uses the opposite rounding direction of the debt shares-to-assets conversion to prevent underflow
-   * in premium debt.
-   * @param hub The liquidity hub of the reserve.
-   * @param assetId The identifier of the asset.
-   * @param shares The amount of shares to convert to assets amount.
-   * @return The amount of assets converted corresponding to user's premium offset.
```
yan-man, line 685: why rm docs for params and return vals?

DhairyaSethi: unneeded here, it's very obvious from the first dev comment

src/libraries/helpers/Constants.sol
yan-man, line 1: (for future PRs) constants already in existing libs should remain there yes?

yan-man: if so, perhaps HEALTH_FACTOR_LIQUIDATION_THRESHOLD should go into LiquidationLogic lib

tests/Base.t.sol
```diff
@@ -1065,6 +1065,23 @@ abstract contract Base is Test {
     return configKey;
   }
 
+  function updateCollateralFactor(
```
yan-man: i think with this one we'll have 3 updateCollateralFactor helpers lol, any way to consolidate them?

tests/Base.t.sol
```diff
@@ -1867,6 +1899,13 @@ abstract contract Base is Test {
     return spoke.getDynamicReserveConfig(reserveId).collateralFactor;
   }
 
+  function _getCollateralFactor(
```
yan-man, line 1904: we also have an existing one of these, can we consolidate?

src/contracts/Spoke.sol
yan-man: imo shouldnt pick and choose for natspec and just include everything, even if obvious. Otherwise I think gets too opinionated as all the other "id" params are redundant and could be removed too. But can keep description shorter

tests/Base.t.sol
DhairyaSethi: the first two: update latest config, one takes reserveid input, other takes fn (ISpoke) returns (uint) for syntax sugar

and this new one updates at a specific config key, will rename this with `atKey`

tests/Base.t.sol
DhairyaSethi: both are for syntax sugar

src/libraries/helpers/Constants.sol
DhairyaSethi: yeah we can settle on this, these are global constants of the contracts, so can keep lib one's isolated there

HEALTH_FACTOR_LIQUIDATION_THRESHOLD is used on the Spoke also

src/contracts/Spoke.sol
DhairyaSethi: i'm thinking internal fn natspec brings clutter where it's not super useful

yan-man: ok as long as we stay consistent, but I'm in favor of keeping even for internal

PR #666, opened 2025-08-22
feat: external liquidation logic library

avniculae:
- move liquidation logic from Spoke to external LiquidationLogic lib


src/contracts/Spoke.sol
```diff
@@ -1031,195 +1005,41 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
...
-      collateralReserveId
-    ];
-    DataTypes.UserPosition storage userDebtPosition = _userPositions[user][debtReserveId];
-
-    (vars.drawnDebt, vars.premiumDebt, vars.accruedPremium) = _getUserDebt(
-      vars.debtReserveHub,
-      vars.debtAssetId,
-      userDebtPosition
-    );
+    DataTypes.LiquidationCallParams memory params;
```
yan-man: would like to keep consistent on this, was before using "vars" for struct defined within a method; "params" for struct passed into a method. Dont feel too strongly but if we keep it like this, would like to update the `executeLiquidationCall` method

src/contracts/Spoke.sol
```diff
@@ -1031,195 +1005,41 @@ contract Spoke is ISpoke, Multicall, AccessManaged {
     uint256 debtToCover,
```
yan-man: i think in general don't really need this _executeLiquidationCall method, can collapse logic into liquidationCall directly

Issue #554, opened 2025-07-22
Remove lastUpdateTimestamp parameter in Hub AssetUpdated event


miguelmtzinf:
We are emitting `block.timestamp` as `lastUpdateTimestamp` when emitting `AssetUpdated` event.

We can remove this param once the block timestamp is added to `eth_getLogs` rpc call.

Reference: https://ethereum-magicians.org/t/proposal-for-adding-blocktimestamp-to-logs-object-returned-by-eth-getlogs-and-related-requests/11183

miguelmtzinf:
It's live on Ethereum Mainne, Ethereum Sepolia, Base Mainnet, Base Sepolia. Rest of chains will be adding support overtime.
Additionally, we are in conversations with Alchemy so they can build custom RPC methods.

Updates on slack [thread](https://avaraxyz.slack.com/archives/C04JUBMUUV9/p1747228992101199)

yan-man:
decided today will do this later, otherwise will affect backend

PR #671, opened 2025-08-24
fix: remove unneeded getter methods on hub

### LINKED ISSUE #603: Remove getAssetAddedShares, getAssetAddedAmounts
- do not include fee shares, pointless as replaced getAssetTotalShares, getAssetTotalAmounts

yan-man:
- remove getAssetAddedShares, getAssetAddedAmounts 

closes #603


tests/Base.t.sol
```diff
@@ -1376,6 +1376,14 @@ abstract contract Base is Test {
     return (drawnRestored, premium);
   }
 
+  // get the asset added amount converted from addedShares
+  function getAssetAddedAmount(
```
avniculae: the name is a bit confusing given that on the hub we have getTotalAddedAmount, nbd. maybe we can document here the difference between the 2 fns?

tests/Base.t.sol
```diff
@@ -1889,8 +1897,8 @@ abstract contract Base is Test {
     return hub1.getAssetConfig(assetId).liquidityFee;
   }
 
-  function _getFeeReceiver(uint256 assetId) internal view returns (address) {
-    return hub1.getAssetConfig(assetId).feeReceiver;
+  function getFeeReceiver(IHub hub, uint256 assetId) internal view returns (address) {
```
avniculae: why did we remove `_`? we still use it in other fns right? do we want to migrate to this new convention?

tests/unit/Hub/HubAccrueInterest.t.sol
```diff
@@ -55,7 +55,7 @@ contract HubAccrueInterestTest is Base {
     assertEq(daiInfo.lastUpdateTimestamp, vm.getBlockTimestamp());
     assertEq(daiInfo.drawnIndex, WadRayMath.RAY);
     assertEq(daiInfo.realizedPremium, 0);
-    assertEq(hub1.getAssetAddedAmount(daiAssetId), 0);
+    assertEq(getAssetAddedAmount(hub1, daiAssetId), 0);
```
avniculae: I'm wondering: should we use `getAssetAddedAmount` in all of these places, or can some be replaced by `hub.getTotalAddedAmount`?

tests/Base.t.sol
yan-man: yes we want to migrate to this new convention for tests, we've been going between both. Just want to eventually have some consistency

tests/Base.t.sol
yan-man: renamed it to clarify a bit and added some comments https://github.com/aave/aave-v4/pull/671/commits/9483b8a72b1398f846905f344507164737f53b15

tests/unit/Hub/HubAccrueInterest.t.sol
yan-man: in some cases can be replaced, done where applicable. but also can be off by rounding or liquidityFee impact on the effective amount

https://github.com/aave/aave-v4/pull/671/commits/9483b8a72b1398f846905f344507164737f53b15

tests/Base.t.sol
CheyenneAtapour: what is the decided convention now? That we won't use `_` for internal methods within tests?

tests/unit/Hub/HubAccrueInterest.t.sol
```diff
@@ -241,11 +241,15 @@ contract HubAccrueInterestTest is Base {
 
     assertEq(elapsed, daiInfo.lastUpdateTimestamp - startTime);
     assertEq(daiInfo.drawnIndex, expectedDrawnIndex, 'drawnIndex');
-    assertEq(
-      hub1.getAssetAddedAmount(daiAssetId) + _calculateBurntInterest(hub1, daiAssetId),
-      addAmount + addAmount2 + interest,
-      'addAmount'
+    console.log(
```
CheyenneAtapour: This should be removed before merging in

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
```diff
@@ -80,7 +80,17 @@ contract SpokeWithdrawScenarioTest is SpokeBase {
     });
 
     uint256 treasuryFees = hub1.getSpokeAddedAmount(daiAssetId, address(treasurySpoke));
-    uint256 interestAccrued = hub1.getAssetAddedAmount(daiAssetId) - treasuryFees - supplyAmount;
+    uint256 interestAccrued = getConvertedAssetAddedAmount(hub1, daiAssetId) -
+      treasuryFees -
+      supplyAmount;
+
+    console.log('treasuryFees %e', treasuryFees);
```
CheyenneAtapour: Same with these console logs

tests/Base.t.sol
yan-man: yeah unless we decide otherwise, I'm in favor of no _ for internal helpers (just bc looks cleaner) but don't feel too strongly. Just want something consistent we can stick with.

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
yan-man: ah ty forgot about these 

https://github.com/aave/aave-v4/pull/671/commits/42fbca29e80bf1b1b15b981e23ea42968eeaa1d5

tests/unit/Hub/HubAccrueInterest.t.sol
yan-man: https://github.com/aave/aave-v4/pull/671/commits/42fbca29e80bf1b1b15b981e23ea42968eeaa1d5

tests/Base.t.sol
CheyenneAtapour: Honestly, I am in favor of keeping the `_` so at least we can differentiate between functions that are internal vs external

tests/Base.t.sol
```diff
@@ -1377,6 +1377,15 @@ abstract contract Base is Test {
     return (drawnRestored, premium);
   }
 
+  // get the asset added amount converted from total added shares
+  // effective addedAssetAmount which accounts for liquidity fee impact on exchange rate
+  function getConvertedAssetAddedAmount(
```
CheyenneAtapour: This name is slightly weird for me, but I don't atm have a better alternative

yan-man: not great, open to suggestions

tests/Base.t.sol
yan-man: for tests don't really need any external methods though, ie everything in base test is internal. I'd so far been using _ but a lot of the other ones are without, like all `update*` helpers

DhairyaSethi: if removing underscore then pls make the methods public (we've only done this on some assertEqs) 

no strong opinion from me, leaning in favour of _ internal

CheyenneAtapour: I agree with @DhairyaSethi , keep _ for internal at least

yan-man: decided today we'll keep _ for internal methods, public on assertEqs

tests/Base.t.sol
miguelmtzinf: is not true that this is fn should return the same as `getTotalAddedAssets` except rounding? so why cannot we use `getTotalAddedAssets` directly?
re naming, what about `getTotalAddedSharesIntoAssets`

tests/Base.t.sol
miguelmtzinf: public without `_` for all assert functions then? or only those that are overloading `assertEq` foundry fn

yan-man: > public without _ for all assert functions then

i'm ok with this, whatever is public can be without _

tests/Base.t.sol
yan-man: It is not the same as getTotalAddedAssets except rounding, ig as you mentioned with the virtual shares taken into account 

good w new name: https://github.com/aave/aave-v4/pull/671/commits/3a3d1f9887ce274f97a9313ff9fe8c67611a719a

tests/unit/Hub/HubAccrueInterest.t.sol
yan-man: updated this now everywhere, incorporating burnt interest

tests/Base.t.sol
yan-man: updated this now everywhere with getTotalAddedAssets, incorporating burnt interest

miguelmtzinf:
reverted back the change for removing `lastUpdateTimeStamp` from `AssetUpdate` event. We ll apply this change later, so this does not close #554

Issue #669, opened 2025-08-24
position status iterator improvements


DhairyaSethi:
1. make `next`, `nextBorrowing`, `nextCollateral` more consistent with `collateralCount` and disregard potential dirty bits set after `reserveCount` https://github.com/aave/aave-v4/blob/457775efe93a9b5fdea88cb98a5b54d9fd31cb4d/src/libraries/configuration/PositionStatus.sol#L134 OR remove that assumption from `collateralCount`
the change will needed on the `isolate`, `isolateBorrowingFrom`, `isolateCollateralFrom` bitmasks, to disregard count after `reserveCount`
this was not added in the first impl because it's not as straightforward - the iterators loop over buckets from start to end (unlike `collateralCount` which loops from end to start to avoid maintaining an end count & checking if start != end) and in this approach will need to to the check only on the last bucket OR pass input differently where bucket and offset are both counted

2. for invalid inputs where `startReserveId` > `reserveCount` https://github.com/aave/aave-v4/blob/457775efe93a9b5fdea88cb98a5b54d9fd31cb4d/src/libraries/configuration/PositionStatus.sol#L148-L149 we need to handle appropriately. either make end bucket min with the start bucket OR change https://github.com/aave/aave-v4/blob/457775efe93a9b5fdea88cb98a5b54d9fd31cb4d/src/libraries/configuration/PositionStatus.sol#L152 cond to bucket <= endBucket instead of !=

3. check if branchless is cheaper here https://github.com/aave/aave-v4/blob/457775efe93a9b5fdea88cb98a5b54d9fd31cb4d/src/libraries/configuration/PositionStatus.sol#L186

PR #676, opened 2025-08-25
feat : modify PositionStatus iterator behavior

Kogaroshi:
The methods `next`, `nextBorrowing`, `nextCollateral` are now exclusive with the starting parameter, and search from last to first, to follow the the behavior of `collateralCount` and disregard dirty bits.

For more, see #669


src/libraries/configuration/PositionStatus.sol
```diff
@@ -253,25 +244,25 @@ library PositionStatus {
...
     assembly ('memory-safe') {
-      ret := and(word, shl(shl(1, mod(reserveId, 128)), BORROWING_MASK))
+      ret := and(word, shr(sub(256, shl(1, mod(reserveCount, 128))), BORROWING_MASK))
     }
   }
 
   /**
-   * @dev Isolates the bits from word, disregarding bits before `reserveId`.
+   * @dev Isolates the bits from word, disregarding bits after `reserveCount`.
    */
```
miguelmtzinf:
```suggestion
   * @dev Isolates bits up to the given `reserveCount`, clearing all later reserves.
   * @param The 256-bit value encoding reserves configuration.
   * @param The number of reserves (2 bits each) to include.
   * @return The portion of word containing bits from the first reserve up to `reserveCount`.
   */
```

src/libraries/configuration/PositionStatus.sol
```diff
@@ -284,19 +275,6 @@ library PositionStatus {
...
-  function isolateCollateralFrom(
-    uint256 word,
-    uint256 reserveId
-  ) internal pure returns (uint256 ret) {
-    // ret = word & (COLLATERAL_MASK << ((reserveId % 128) << 1));
-    assembly ('memory-safe') {
-      ret := and(word, shl(shl(1, mod(reserveId, 128)), COLLATERAL_MASK))
-    }
-  }
-
```
miguelmtzinf, line 299: for `isolateCollateralUntil`
```
   * @dev Isolates collateral bits up to the given `reserveCount`, clearing all later reserves.
   * @param The 256-bit value encoding reserves configuration.
   * @param The number of reserves (2 bits each) to include.
   * @return The portion of word containing collateral bits from the first reserve up to `reserveCount`.
   */
```

miguelmtzinf: apply smt similar for `isolateBorrowingUntil` pls

src/libraries/configuration/PositionStatus.sol
```diff
@@ -164,49 +161,43 @@ library PositionStatus {
...
-   * @dev Does not disregard potential dirty bits set after `reserveCount` in it's bucket.
+   * @dev Returns NOT_FOUND if no such reserveId exists, starts searching from `lastSeenReserveId`, searches from last to first.
+   * @dev Disregard potential dirty bits set after `reserveCount` in it's bucket.
    * @param self The configuration object.
-   * @param startReserveId The reserveId to start searching from.
-   * @param reserveCount The current reserveCount, to avoid reading uninitialized buckets.
+   * @param lastSeenReserveId The reserveId to start searching from.
    * @return reserveId The next reserveId that is using as collateral.
    */
   function nextCollateral(
```
miguelmtzinf, line 195: i d rename to `previousCollateral` and `fromReserveId` to make clearer that we go backwards now

src/libraries/configuration/PositionStatus.sol
```diff
@@ -164,49 +161,43 @@ library PositionStatus {
...
    * @dev Returns the next reserveId that is using as collateral.
-   * @dev Returns NOT_FOUND if no such reserveId exists, starts searching from `startReserveId`.
-   * @dev Does not disregard potential dirty bits set after `reserveCount` in it's bucket.
+   * @dev Returns NOT_FOUND if no such reserveId exists, starts searching from `lastSeenReserveId`, searches from last to first.
+   * @dev Disregard potential dirty bits set after `reserveCount` in it's bucket.
    * @param self The configuration object.
-   * @param startReserveId The reserveId to start searching from.
-   * @param reserveCount The current reserveCount, to avoid reading uninitialized buckets.
+   * @param lastSeenReserveId The reserveId to start searching from.
    * @return reserveId The next reserveId that is using as collateral.
```
miguelmtzinf:
```
/**
 * @dev Finds the previous collateralized reserve strictly before `fromReserveId`.
 * @dev The search starts at `fromReserveId` (exclusive) and scans backward across buckets.
 * @dev Returns `NOT_FOUND` if no collateralized reserve exists before the bound.
 * @dev Ignores dirty bits beyond the configured `reserveCount` within the current bucket.
 * @param self The position status storing reserves bitmap.
 * @param fromReserveId The exclusive upper bound to start from (this reserveId is not considered).
 * @return reserveId The previous collateralized reserveId, or `NOT_FOUND` if none is found.
 */
```

src/contracts/Spoke.sol
```diff
@@ -783,19 +783,17 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     address user,
     bool refreshConfig
   ) internal returns (DataTypes.UserAccountData memory userAccountData) {
-    IAaveOracle aaveOracle = oracle;
-    uint256 reserveCount = _reserveCount;
-
     DataTypes.PositionStatus storage positionStatus = _positionStatus[user];
+
+    IAaveOracle aaveOracle = oracle;
+    uint256 reserveId = _reserveCount;
```
miguelmtzinf, line 789: why this? a bit confusing now, i d prefer keep on using `reserveCount`

miguelmtzinf: or use another name like `cursor` or smt

Kogaroshi: Because now we simply go from id to id until reaching 0, we want to start at the last one and go down.

src/contracts/Spoke.sol
```diff
@@ -783,19 +783,17 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
...
+      positionStatus.collateralCount(reserveId)
     );
-
-    uint256 reserveId;
     bool borrowing;
     bool collateral;
     while (true) {
-      (reserveId, borrowing, collateral) = positionStatus.next(reserveId, reserveCount);
+      (reserveId, borrowing, collateral) = positionStatus.next(reserveId);
       if (reserveId == PositionStatus.NOT_FOUND) break;
```
miguelmtzinf, line 797: can we move this into the while condition? doing this in `_notifyRiskPremiumUpdate` too

src/contracts/Spoke.sol
miguelmtzinf: i see, would it be cursor or pointer more confusing?

Kogaroshi: Yeah, we could rename to `cursor` or `iterator` for clarity

DhairyaSethi: it's used as an down the line as a reserveId in the hub.preview*() calls, cursor might be confusing there

src/contracts/Spoke.sol
miguelmtzinf: agreed not worth since it returns 3 params not one and we d need to call this fn twice (before and after the loop starts)

Issue #673, opened 2025-08-25
KeyValueListInMemory improvements


DhairyaSethi:
1. https://github.com/aave/aave-v4/blob/8f9301e10feeb81b528ea0d0bf54fe58ddb5d44c/src/libraries/helpers/KeyValueListInMemory.sol#L14-L15
Rename these to MAX_KEY & MAX_VALUE. or refactor implementation to use more accurate naming & consistent logic
logic rft references
- https://github.com/aave/aave-v4/pull/628/files#diff-bae38dbef5852aadab721ec16814dcd8fa0c94a24bc2ba0192c31401bafcebca
- https://github.com/aave-dao/aave-v3-origin/blob/v3.5.0/src/contracts/protocol/libraries/configuration/ReserveConfiguration.sol

2. to decrease the while loop len: https://github.com/aave/aave-v4/blob/8f9301e10feeb81b528ea0d0bf54fe58ddb5d44c/src/contracts/Spoke.sol#L885-L900
rft logic to sort in ASC by `key` and at clashes sort by `value` in DESC
one option is to `not` the `value` on packing & unpacking

3. Write tests for this library

3. Investigate if packing more than element in one the sorted list is worth the effort and gas save. the idea max key, ie collateralRisk is 1000_00, which can fit in 17 bits, max value can be taken as 1e30 which fits in 100 bits, hence we can fit two (key, value) pairs in a slot effectively reducing the size of the array by half

PR #679, opened 2025-08-26
refactor : change KVL value packing logic

Kogaroshi:
See #673 

> Write tests for this library. 

Tests were already existing, a new one was added

Current solution : 
Apply a not on the value while packing and unpacking. When sorting, any item no set in the list will end up as the 1st items of the sorted list. We do not want to unpack them (as they were never packed), and want to skip them in `_calculateAndPotentiallyRefreshUserAccountData` to save some gas during the iterations.

Alternative solutions : 
- when `collateralFactor == 0` for the reserve, before skipping, add an item with `key = type(uint32).max` & `value = 0` so it will always end up last in the List after sorting. This requires to pack & unpack more data, but avoids to loop over empty items later in the execution.
- implement a "next" iterator behavior to skip all unset items after sorting the list, until reaching the 1st non (0,0) value.

### DIFF
--- a/tests/unit/KeyListInMemory.t.sol
+++ b/tests/unit/KeyListInMemory.t.sol
@@ -28,6 +28,26 @@ contract KeyValueListInMemoryTest is Test {
     _assertSortedOrder(list);
   }
 
+  function test_fuzz_sortByKey_with_collision(uint256[] memory seed) public pure {
+    vm.assume(seed.length > 10);
+    uint256[] memory collisionKeys = new uint256[](seed.length / 10);
+    for (uint256 i; i < collisionKeys.length; ++i) {
+      collisionKeys[i] = seed[i];
+    }
+
+    vm.assume(seed.length > 0);
+    KeyValueListInMemory.List memory list = KeyValueListInMemory.init(seed.length);
+    for (uint256 i; i < seed.length; ++i) {
+      list.add(
+        i,
+        _truncateKey(collisionKeys[seed[i] % collisionKeys.length]),
+        _truncateValue(seed[i])
+      );
+    }
+    list.sortByKey();
+    _assertSortedOrder(list);
+  }
+
   function test_fuzz_get(uint256[] memory seed) public pure {
     vm.assume(seed.length > 0);
     KeyValueListInMemory.List memory list = KeyValueListInMemory.init(seed.length);
@@ -41,7 +61,7 @@ contract KeyValueListInMemoryTest is Test {
     }
   }
 
-  function test_fuzz_get_unsetData(uint256[] memory seed) public {
+  function test_fuzz_get_uninitialized(uint256[] memory seed) public {
     vm.assume(seed.length > 0);
     uint256 fillArrayTill = vm.randomUint(0, seed.length - 1);
     KeyValueListInMemory.List memory list = KeyValueListInMemory.init(seed.length);
@@ -50,7 +70,7 @@ contract KeyValueListInMemoryTest is Test {
     }
     for (uint256 i; i < seed.length; ++i) {
       (uint256 key, uint256 value) = list.get(i);
-      if(i < fillArrayTill) {
+      if (i < fillArrayTill) {
         assertEq(key, _truncateKey(seed[i]));
         assertEq(value, _truncateValue(seed[i]));
       } else {
@@ -60,7 +80,7 @@ contract KeyValueListInMemoryTest is Test {
     }
   }
 
-  function test_fuzz_get_unsetData_sorted(uint256[] memory seed) public {
+  function test_fuzz_get_uninitialized_sorted(uint256[] memory seed) public {
     vm.assume(seed.length > 0 && seed.length < 1e2);
     uint256 fillArrayTill = vm.randomUint(0, seed.length - 1);
     KeyValueListInMemory.List memory list = KeyValueListInMemory.init(seed.length);
@@ -70,7 +90,7 @@ contract KeyValueListInMemoryTest is Test {
     list.sortByKey();
     for (uint256 i; i < seed.length; ++i) {
       (uint256 key, uint256 value) = list.get(i);
-      if(i >= fillArrayTill) {
+      if (i >= fillArrayTill) {
         assertEq(key, 0);
         assertEq(value, 0);
       }
@@ -79,11 +99,15 @@ contract KeyValueListInMemoryTest is Test {
 
   function _assertSortedOrder(KeyValueListInMemory.List memory list) internal pure {
     // validate sorted order
-    (uint256 prevKey, ) = list.get(0);
+    (uint256 prevKey, uint256 prevValue) = list.get(0);
     for (uint256 i = 1; i < list.length(); ++i) {
-      (uint256 key, ) = list.get(i);
+      (uint256 key, uint256 value) = list.get(i);
       assertLe(prevKey, key);
+      if (prevKey == key) {
+        assertGe(prevValue, value);
+      }
       prevKey = key;
+      prevValue = value;
     }
   }
 


Kogaroshi:
New solution chosen after dev sync : 
- revert back to using OZ `Array.sort()`, since Solady's lib do not propose a descending order sort
- do not alter `value` when packing
- when packing `key`, pack it as `_MAX_KEY - key`

This allows to sort the list in descending order, where the output will still us the `key` in ascending order (when unpacked) and the `value` act as tie-breaker in descending order (allowing us to cover bigger amounts per iteration in the Spoke).
This solution removes the need to skip iterations at the beginning of the loop in `_calculateAndPotentiallyRefreshUserAccountData` when the un-set items in the list ended first after sorting, instead we have those unset items at the end of the list after sorting, and either they are not reached in the loop, or they are but do not affect the outcome (since `vars.userCollateralInBaseCurrency` would be 0).

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -43,23 +48,29 @@ library KeyValueListInMemory {
...
+    return _MAX_KEY - (data >> _KEY_SHIFT);
   }
 
   function unpackValue(uint256 data) internal pure returns (uint256) {
     return data & ((1 << _KEY_SHIFT) - 1);
   }
 
   function unpack(uint256 data) internal pure returns (uint256, uint256) {
+    // @dev no need to unpack data that was never packed
+    if(data == 0) return(0,0);
```
DhairyaSethi: is this necessary?

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -32,8 +37,8 @@ library KeyValueListInMemory {
   }
 
   function add(List memory self, uint256 idx, uint256 key, uint256 value) internal pure {
-    require(key <= _KEY_MASK, MaxKeySizeExceeded(key));
-    require(value <= _VALUE_MASK, MaxValueSizeExceeded(value));
+    require(key <= _MAX_KEY, MaxKeySizeExceeded(key));
+    require(value <= _MAX_VALUE, MaxValueSizeExceeded(value));
     self._inner[idx] = pack(key, value);
   }
```
DhairyaSethi, line 43: comment on `get` with a note/warning that unintitalised keys are returned as MAX_KEY, 0; so use appropriately 
and comment on pack, unpackKey to specify what sub(maxKey, ) is trying to achieve 
also pls fix the explicit comment on `sortByKey` that it results in ascending order by key, and desc by value on collission

snapshots/Spoke.Operations.json
```diff
@@ -1,26 +1,26 @@
 {
-  "borrow: first": "248921",
-  "borrow: second action, same reserve": "212393",
-  "liquidationCall: full": "309471",
-  "liquidationCall: partial": "384217",
-  "permitReserve + repay (multicall)": "269705",
+  "borrow: first": "249030",
```
DhairyaSethi: gas increase is only bc of the sorting dep change, right? the addition changes apart from this are a sub and push opcodes i think

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -2,7 +2,13 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity ^0.8.0;
 
-import {LibSort} from 'src/dependencies/solady/LibSort.sol';
```
DhairyaSethi, line 5: pls include in PR desc our rationale to switch sorting; solady doesn't provide convinient way to sort in desc

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -2,7 +2,13 @@
...
+/** 
+ * @notice Library to pack key-value pairs in a list to sort it
+ * @dev The key are packed as `_MAX_KEY - key`, so we can sort the list in descending order,
+ * with the value in ascending order in case of collision, and
+ * when unpacking have the smallest we get the smallest keys at first.
```
DhairyaSethi:
```suggestion
 * @notice Library to pack key-value pairs in a list.
 * @dev `sortByKey` helper sorts by asending order of the `key` & in case of collission by descending order of the `value`. This is acheived by sorting the packed `key-value` pair in descending order, but storing the invert of the `key` (ie `_MAX_KEY - key`).
 * WARNING: Uninitialized keys are returned as (key: `MAX_KEY`, value: 0).
```

DhairyaSethi: or something like this

src/libraries/helpers/KeyValueListInMemory.sol
DhairyaSethi: we cannot revert on uninitialized keys (bc in deficit case) we would reach that case so i think let's rename `get` to `tryGet` instead?

src/libraries/helpers/KeyValueListInMemory.sol
Kogaroshi: An item that was never packed will still be unpacked, applying the `_MAX_KEY - (data >> _KEY_SHIFT);` with a `key` that would be 0, so it will return `_MAX_KEY`.
If we consider it a non-issue, we can remove the line.

snapshots/Spoke.Operations.json
Kogaroshi: Yes, most of the gas increase is due to the change of sorting lib

src/libraries/helpers/KeyValueListInMemory.sol
DhairyaSethi: whats the gas increase bc of this? it is probably better this way, if we return (0, 0) for uninitialized keys rather than (MAX_KEY, 0) -> the warning can become lighter

Kogaroshi: gas increase is ~0.01% average, we can consider it a good tradeoff for the simplicity it brings

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -2,7 +2,13 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity ^0.8.0;
 
-import {LibSort} from 'src/dependencies/solady/LibSort.sol';
+/**
+ * @notice Library to pack key-value pairs in a list.
+ * @dev `sortByKey` helper sorts by asending order of the `key` & in case of collission by descending order of the `value`.
+ * This is acheived by sorting the packed `key-value` pair in descending order, but storing the invert of the `key` (ie `_MAX_KEY - key`).
+ * WARNING: Uninitialized keys are returned as (key: `MAX_KEY`, value: 0).
```
DhairyaSethi: pls fix this everywhere 
also another comment on sort uninitialised are at the end w 0,0

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -11,12 +18,11 @@ library KeyValueListInMemory {
   uint256 internal constant _MAX_KEY_BITS = 32;
   uint256 internal constant _MAX_VALUE_BITS = 224;
 
-  uint256 internal constant _KEY_MASK = (1 << _MAX_KEY_BITS) - 1;
-  uint256 internal constant _VALUE_MASK = (1 << _MAX_VALUE_BITS) - 1;
+  uint256 internal constant _MAX_KEY = type(uint32).max;
+  uint256 internal constant _MAX_VALUE = type(uint224).max;
```
DhairyaSethi: nit: let's reuse the bits value from above, such that only a change in those values is needed & these are updated dynamically if needed, find it cleaner

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -32,34 +38,45 @@ library KeyValueListInMemory {
   }
 
   function add(List memory self, uint256 idx, uint256 key, uint256 value) internal pure {
-    require(key <= _KEY_MASK, MaxKeySizeExceeded(key));
-    require(value <= _VALUE_MASK, MaxValueSizeExceeded(value));
+    require(key <= _MAX_KEY, MaxKeySizeExceeded(key));
+    require(value <= _MAX_VALUE, MaxValueSizeExceeded(value));
```
DhairyaSethi: let's combine these two errors into one, add operation will be slightly cheaper such that only one jumpi is checked

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -32,34 +38,45 @@ library KeyValueListInMemory {
...
   }
 
+  /// @dev Uninitialized keys are returned as (key: 0, value: 0)
   function get(List memory self, uint256 idx) internal pure returns (uint256, uint256) {
     return unpack(self._inner[idx]);
   }
 
+  /**
+   * @dev since `key` is in the MSB, we can sort by the key by sorting the array in descending order
+   * (so the keys are in ascending order when unpacking), and using value in case of collision, and all
```
DhairyaSethi: not clear key sorted in ascending (bc of the inversion when storing) and desc by value on collision 
nbd

tests/unit/KeyListInMemory.t.sol
```diff
@@ -28,7 +28,7 @@ contract KeyValueListInMemoryTest is Test {
     _assertSortedOrder(list);
   }
 
-  function test_fuzz_get(uint256[] memory seed) public pure {
+  function test_fuzz_tryGet(uint256[] memory seed) public pure {
```
DhairyaSethi: revert pls

tests/unit/KeyListInMemory.t.sol
```diff
@@ -41,6 +41,24 @@ contract KeyValueListInMemoryTest is Test {
     }
   }
 
+  function test_fuzz_tryGet_unsetData(uint256[] memory seed) public pure {
+    vm.assume(seed.length > 0);
+    KeyValueListInMemory.List memory list = KeyValueListInMemory.init(seed.length);
+    for (uint256 i; i < seed.length / 2; ++i) {
```
DhairyaSethi: ty for this test, let's choose the limit by randomUint also 
fillArrayTill = vm.randomUint(0, seed.length - 1)
also lets assert sorted order also until we reach this in the assertions below

snapshots/Spoke.Operations.json
miguelmtzinf, line 1: this does *not* decrease gas costs. Am I missing anything?

DhairyaSethi: we had switch back to OZ sort from solady (that's most of gas increase) bc solady doesn't support sorting in decreasing order efficiently 

this is cheaper in the general case bc of the logic

DhairyaSethi: when cf is the same for two assets, higher collateral value is consumed first, potentially reducing the iterations while loop of risk premium calc

src/libraries/helpers/KeyValueListInMemory.sol
```diff
@@ -32,34 +37,45 @@ library KeyValueListInMemory {
...
     return unpack(self._inner[idx]);
   }
 
+  /**
+   * @dev since `key` is in the MSB, we can sort by the key by sorting the array in descending order
+   * (so the keys are in ascending order when unpacking, due to inversion when packing),
+   * and using value in descending order in case of collision,
+   * and all uninitialized keys are placed at the end of the list after sorting.
+   */
   function sortByKey(List memory self) internal pure {
```
miguelmtzinf, line 55: the entire lib is ad-hoc to our case. I wonder if we should add smt to the name to tell about the sorting order

tests/unit/KeyListInMemory.t.sol
```diff
@@ -41,6 +41,42 @@ contract KeyValueListInMemoryTest is Test {
     }
   }
 
+  function test_fuzz_get_unsetData(uint256[] memory seed) public {
```
miguelmtzinf: why it's call `unsetData`? we are asserting set and unset data right

tests/unit/KeyListInMemory.t.sol
```diff
@@ -41,6 +41,42 @@ contract KeyValueListInMemoryTest is Test {
...
+        assertEq(key, _truncateKey(seed[i]));
+        assertEq(value, _truncateValue(seed[i]));
+      } else {
+        assertEq(key, 0);
+        assertEq(value, 0);
+      }
+    }
+  }
+
+  function test_fuzz_get_unsetData_sorted(uint256[] memory seed) public {
```
miguelmtzinf: this may include it, but missing an specific case where there is a tie on collateral risk and check it's ordered in desc by collateral value

tests/unit/KeyListInMemory.t.sol
DhairyaSethi: it's testing the cases when the entire array is not populated, ie you allocate x items but fill less than x,

maybe uninitialised is more descriptive than unset wdyt @Kogaroshi

src/libraries/helpers/KeyValueListInMemory.sol
Kogaroshi: `sortByKeyAscending` ? Or do we want something longer and more explicit ?

DhairyaSethi: i think sort by default means ascending (take solady for example) 
personally fine as is

tests/unit/KeyListInMemory.t.sol
Kogaroshi: yeah, we use uninitialized in comments, so I'll rename accordingly

tests/unit/KeyListInMemory.t.sol
Kogaroshi: Added a new test with key collisions to check on descending value order.

PR #685, opened 2025-09-02
feat : use SignatureChecker instead of only ECDSA

### LINKED ISSUE #680: support 1271 signatures for setUserPositionManagerWithSig
use https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol

for signature verification here
https://github.com/aave/aave-v4/blob/8f9301e10feeb81b528ea0d0bf54fe58ddb5d44c/src/contracts/Spoke.sol#L433
ensure we're under code size limit

Kogaroshi:
fixes #680

Following final decision in https://github.com/aave/aave-v4/issues/704, the SignatureChecker library used is now OZ v5.1, wrapped as an Helper lib, so the isValidSignatureNow function can be external, and not impact the Spoke code size anymore.
New Spoke code size after latest changes : 24,077 (- 151 from initial code size).


src/contracts/Spoke.sol
```diff
@@ -430,7 +430,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
         )
       )
     );
-    require(ECDSA.recover(hash, v, r, s) == user, InvalidSignature());
+    require(
+      SignatureChecker.isValidSignatureNow(user, hash, abi.encodePacked(r, s, v)),
```
DhairyaSethi: don't like that we pack here and unpack later again in the ECSA case, very inefficient, if it's needed for 1271 case let's just accept bytes signature in the call

Kogaroshi: The SignatureChecker expects the packed signature directly indeed, I'll change so we expect the same packed version in the call.

src/contracts/Spoke.sol
```diff
@@ -430,7 +428,10 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
         )
       )
     );
-    require(ECDSA.recover(hash, v, r, s) == user, InvalidSignature());
+    require(
+      SignatureChecker.isValidSignatureNow(user, hash, signature),
```
DhairyaSethi: can we use the isValidNowCalldata variant instead? think it'd be cheaper for ecdsa case

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/SignatureChecker.sol#L44

src/dependencies/openzeppelin/SignatureChecker.sol
```diff
@@ -0,0 +1,73 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.4.0) (utils/cryptography/SignatureChecker.sol)
```
DhairyaSethi: note all oz deps are pinned at 5.1, pls stick to that if possible (last i checked the only upgrade in 5.4 was non eoa signer verification but we're not using that, pls confirm)

src/dependencies/openzeppelin/SignatureChecker.sol
```diff
@@ -0,0 +1,73 @@
...
+/**
+ * @dev Signature verification helper that can be used instead of `ECDSA.recover` to seamlessly support:
+ *
+ * * ECDSA signatures from externally owned accounts (EOAs)
+ * * ERC-1271 signatures from smart contract wallets like Argent and Safe Wallet (previously Gnosis Safe)
+ * * ERC-7913 signatures from keys that do not have an Ethereum address of their own
+ *
+ * See https://eips.ethereum.org/EIPS/eip-1271[ERC-1271] and https://eips.ethereum.org/EIPS/eip-7913[ERC-7913].
+ */
+library SignatureChecker {
```
DhairyaSethi, line 14: fine to copy over everything for this file, we are not trimming oz deps, copying over as is

tests/mocks/MockERC1271Wallet.sol
```diff
@@ -0,0 +1,18 @@
...
+
+contract MockERC1271Wallet is IERC1271 {
+    address public owner;
+
+    constructor(address _owner) {
+        owner = _owner;
+    }
+
+    function isValidSignature(bytes32 _hash, bytes memory _signature) public view returns (bytes4) {
+        return ECDSA.recover(_hash, _signature) == owner ? this.isValidSignature.selector : bytes4(0);
```
DhairyaSethi: more realistic is a data hash that is approved, like in gnosis safe, bc the point is signing scheme can be anything here, and then we can test that a valid sig fromtrun by disable hash fails but nbd this is fine too

tests/unit/Spoke/Spoke.SetUserPositionManagerWithSig.t.sol
```diff
@@ -188,14 +186,48 @@ contract SpokeSetUserPositionManagerWithSigTest is SpokeBase {
...
-      v,
-      r,
-      s
+      signature
     );
 
     assertEq(spoke1.isPositionManager(user, params.positionManager), params.approve);
   }
 
+  function test_setUserPositionManagerWithSig_ERC1271() public {
```
DhairyaSethi, line 315: all negative cases too pls

src/contracts/Spoke.sol
Kogaroshi: Yeah, that shouldn't be an issue to use calldata

src/dependencies/openzeppelin/SignatureChecker.sol
Kogaroshi: I went with 5.4 because the version also supports ERC-7913, v5.1 does not. But if we do not care about ERC-7913 and just want to support ERC-1271, then we can go with v5.1 indeed

src/contracts/Spoke.sol
Kogaroshi: But note that going with v5.1 of SingatureChecker removes the `isValidNowCalldata` variant

src/dependencies/openzeppelin/SignatureChecker.sol
DhairyaSethi: yeah mate that's the non address signer i was talking about, we don't physically support that

Kogaroshi: Okay, I'll replace with v5.1 then

tests/mocks/MockERC1271Wallet.sol
Kogaroshi: I can add a storage for the expected hash, so the condition becomes `ECDSA.recover(_hash, _signature) == owner && hash == expectedHash`, wdyt ?

DhairyaSethi: nah that's kinda pointless no? i meant instead of ecrecover

src/contracts/Spoke.sol
DhairyaSethi: from the caveats the calldata variant looks to be worth it, how much gas does it save?

tests/mocks/MockERC1271Wallet.sol
Kogaroshi: to confirm the hash is the one expected by the smart wallet, and that the paylaod was signed by the smart wallet owner

src/contracts/Spoke.sol
Kogaroshi: Going with v5.1 pushes the Spoke size to 24,570, +255B, 6B off the limit

tests/mocks/MockERC1271Wallet.sol
DhairyaSethi: approvedHashes mapping stored by smart wallet 
owner approves hashes by a simple onlyOwner approveHash(hash) method 

idea is they sign and approve hashes off chain, or this contract implements a different signing scheme like p256 etc

Kogaroshi: Okay, I'm good with the mapping and owner gated checks.
Will change and replace the test

src/contracts/Spoke.sol
DhairyaSethi: what's the difference

Kogaroshi: `isValidERC1271SignatureNow` seems to be be lighter in v5.4

DhairyaSethi: didn't want to use solady here bc their ecdsa checks are weaker (don't check for malleability iirc)
can you do some analysis?

Kogaroshi: Yeah, will compare and see

Kogaroshi: Otherwise, what's the blocker with using v5.4 only for this contract ?
We do not go through the ERC-7913 cases since it's only when called via `function isValidSignatureNow(bytes memory signer, bytes32 hash, bytes memory signature)`, which is not our case

Kogaroshi: I checked with Solady, the lib was indeed lighter, but yeah it didn't pass a signature malleability check

Kogaroshi: After further discussions, analysis and tests, we decided to go with Solady's lib.
For the explanations and answer about malleability issues, see #704

snapshots/Spoke.Operations.json
```diff
@@ -8,6 +8,8 @@
   "permitReserve + supply + enable collateral (multicall)": "174118",
   "repay: full": "173610",
   "repay: partial": "228487",
+  "setUserPositionManagerWithSig: disable": "38317",
+  "setUserPositionManagerWithSig: enable": "79440",
```
DhairyaSethi: this is double the cost bc it's writing one boolean to one cold slot?

DhairyaSethi: oh and also nonces probably, can you burn one nonce to make it non zero before the enable and see

Kogaroshi: Yes, 1st one is more expensive because of cold slot, and was higher due to using nonce 0. Using nonce 1 reduces gas significantly (79440 -> 62340).

src/libraries/helpers/SignatureCheckerHelper.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {SignatureChecker} from 'src/dependencies/openzeppelin/SignatureChecker.sol';
+
+library SignatureCheckerHelper {
+  function isValidSignatureNow(
```
DhairyaSethi: let's add the same natspec and note it's just to make this lib ext, lgtm otherwise

src/libraries/helpers/SignatureCheckerHelper.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {SignatureChecker} from 'src/dependencies/openzeppelin/SignatureChecker.sol';
+
+/**
+ * @notice Simple wrapper contract for OpenZeppelin's SignatureChecker to make `isValidSignatureNow` an external function.
```
DhairyaSethi: wrapper library (not contract) to expose method externally or w delegatecall or smth

miguelmtzinf:
Closes https://github.com/aave/aave-v4/issues/704

src/dependencies/openzeppelin/IERC1271.sol
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * @dev Should return whether the signature provided is valid for the provided data
+   * @param hash      Hash of the data to be signed
+   * @param signature Signature byte array associated with `hash`
+   */
+  function isValidSignature(
+    bytes32 hash,
+    bytes calldata signature
+  ) external view returns (bytes4 magicValue);
+}
```
miguelmtzinf: wrong version, even if same code
```suggestion
// SPDX-License-Identifier: MIT
// OpenZeppelin Contracts (last updated v5.1.0) (interfaces/IERC1271.sol)

pragma solidity ^0.8.20;

/**
 * @dev Interface of the ERC-1271 standard signature validation method for
 * contracts as defined in https://eips.ethereum.org/EIPS/eip-1271[ERC-1271].
 */
interface IERC1271 {
    /**
     * @dev Should return whether the signature provided is valid for the provided data
     * @param hash      Hash of the data to be signed
     * @param signature Signature byte array associated with _data
     */
    function isValidSignature(bytes32 hash, bytes memory signature) external view returns (bytes4 magicValue);
}
```

src/libraries/helpers/SignatureCheckerHelper.sol
```diff
@@ -0,0 +1,25 @@
...
+ * @notice Simple wrapper library for OpenZeppelin's SignatureChecker to expose `isValidSignatureNow` externally.
+ */
+library SignatureCheckerHelper {
+  /**
+   * @dev Checks if a signature is valid for a given signer and data hash. If the signer is a smart contract, the
+   * signature is validated against that smart contract using ERC-1271, otherwise it's validated using `ECDSA.recover`.
+   *
+   * NOTE: Unlike ECDSA signatures, contract signatures are revocable, and the outcome of this function can thus
+   * change through time. It could return true at block N and false at block N+1 (or the opposite).
+   */
```
miguelmtzinf, line 17: i d write our own natspec docs, but nbd now. We can handle separately

src/interfaces/ISpoke.sol
```diff
@@ -222,9 +222,7 @@ interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
     address user,
     bool approve,
     uint256 deadline,
-    uint8 v,
-    bytes32 r,
-    bytes32 s
+    bytes memory signature
```
miguelmtzinf, line 226: add to docs pls

src/libraries/helpers/SignatureCheckerHelper.sol
DhairyaSethi: we can actually most likely remove this wrapper library and just simply use the dep since liquidation PR gives a headroom of 900 bytes for code size

miguelmtzinf: i d be fine with that

miguelmtzinf: https://github.com/aave/aave-v4/issues/723

Issue #580, opened 2025-07-28
add gas snapshot scenarios to tests


yan-man:
- [x] payFee
- [x] transferShares
- [ ] report/eliminate deficit: can we add some snapshot scenarios with deficit being created, can take cases from v3 snapshots (_Originally posted by @DhairyaSethi in https://github.com/aave/aave-v4/pull/472#discussion_r2227086426_)

yan-man:
catalogued in notion

PR #686, opened 2025-09-02
fix: Enforces feeReceiver as spoke on Hub

### LINKED ISSUE #668: hub/configurator: enforce `feeReceiver` invariants
we must ensure that the a feeReceiver is:

- always set (ie never address(0) in AssetConfig)
- always active
- always has MAX_CAP sentinel addCap

three issues right now:
1. 
there is a validation missing on configurator on a new potentially inactive spoke being made the feeReceiver that needs to be fixed
see https://github.com/aave/aave-v4/blob/457775efe93a9b5fdea88cb98a5b54d9fd31cb4d/src/contracts/HubConfigurator.sol#L296-L328

2. HubConfigurator.pauseAsset() - should skip current `feeReceiver`
3. HubConfigurator.freezeAsset() - should skip current `feeReceiver`

yan-man:
closes #668

### DIFF
--- a/src/interfaces/IHub.sol
+++ b/src/interfaces/IHub.sol
@@ -111,7 +111,7 @@ interface IHub is IHubBase, IAccessManaged {
 
   /**
    * @notice Updates the configuration of an asset.
-   * @dev If the fee receiver is updated, it is added as a new spoke with maximum add cap and zero draw cap, so it cannot already exist beforehand.
+   * @dev If the fee receiver is updated, it is added as a new spoke with maximum add cap and zero draw cap.
    * @param assetId The identifier of the asset.
    * @param config The new configuration for the asset.
    */


src/contracts/HubConfigurator.sol
```diff
@@ -61,15 +53,7 @@ contract HubConfigurator is Ownable2Step, IHubConfigurator {
...
-    uint256 assetId = targetHub.addAsset(underlying, decimals, feeReceiver, irStrategy, data);
-
-    targetHub.addSpoke(
-      assetId,
-      feeReceiver,
-      DataTypes.SpokeConfig({addCap: Constants.MAX_CAP, drawCap: Constants.MAX_CAP, active: true})
-    );
+    uint256 assetId = IHub(hub).addAsset(underlying, decimals, feeReceiver, irStrategy, data);
 
     return assetId;
```
miguelmtzinf:
```suggestion
    return IHub(hub).addAsset(underlying, decimals, feeReceiver, irStrategy, data);
```

src/contracts/Hub.sol
```diff
@@ -90,6 +90,12 @@ contract Hub is IHub, AccessManaged {
       feeReceiver: feeReceiver,
       liquidityFee: 0
     });
+    _addSpoke(assetId, feeReceiver);
+    _updateSpokeConfig(
+      assetId,
+      feeReceiver,
+      DataTypes.SpokeConfig({addCap: Constants.MAX_CAP, drawCap: Constants.MAX_CAP, active: true})
+    );
```
miguelmtzinf: we d need similar stuff in `updateAssetConfig`

yan-man: https://github.com/aave/aave-v4/pull/686/commits/9a4233f17f1ad54af7b0ed4b7c0dd45a77f52dc8

src/contracts/HubConfigurator.sol
yan-man: https://github.com/aave/aave-v4/pull/686/commits/9a4233f17f1ad54af7b0ed4b7c0dd45a77f52dc8

tests/gas/Hub.Operations.gas.t.sol
```diff
@@ -83,4 +83,33 @@ contract HubOperations_Gas_Tests is Base {
...
+
+    skip(100);
+
+    address feeReceiver = _getFeeReceiver(daiAssetId);
+
+    vm.startPrank(feeReceiver);
+    hub1.remove(daiAssetId, hub1.getSpokeAddedAmount(daiAssetId, feeReceiver) / 2, alice);
+    vm.stopPrank();
+
+    vm.snapshotGasLastCall('Hub.Operations', 'removeFeeShares: partial');
```
DhairyaSethi: why do we need this? also naming inconsistent, it's a remove op to feeReceiver. would make sense to add these gas snapshots for payFee, transferShares, report & eliminate deficit - pls create a ticket or add here. i would remove these new snapshots

tests/unit/HubConfigurator.t.sol
```diff
@@ -489,7 +484,7 @@ contract HubConfiguratorTest is HubBase {
     );
     address feeReceiver = hub1.getAssetConfig(assetId).feeReceiver;
 
-    vm.expectRevert(IHub.InvalidLiquidityFee.selector);
+    vm.expectRevert(IHub.InvalidLiquidityFee.selector, address(hub1));
```
DhairyaSethi, line 463: it is always so confusing to me that this is the emitter address & inconsistent with expectEmit

tests/gas/Hub.Operations.gas.t.sol
yan-man: @miguelmtzinf asked for this, i think to compare vs the other POCs on feeReceiver impls, will rm this now that we're going with updated version of this 

updated prior issue to include other snapshots here: https://github.com/aave/aave-v4/issues/580

src/contracts/Hub.sol
```diff
@@ -766,4 +779,9 @@ contract Hub is IHub, AccessManaged {
     require(caller == asset.reinvestmentController, OnlyReinvestmentController());
     require(amount > 0 && amount <= asset.swept, InvalidAmount());
   }
+
+  function _addSpoke(uint256 assetId, address spoke) internal {
+    require(_assetToSpokes[assetId].add(spoke), SpokeAlreadyListed());
```
yan-man, line 784: to clarify a nuance here though, an existing spoke can still be added as `feeReceiver` as long as the spoke hasn't been listed on the specific `assetId` of interest. (ie feeReceiver allowed even if spoke1 has been listed on other assetIds)

to fully prevent any prior existing spoke, i think we'll need to loop over all assetIds to check the new feeReceiver isn't listed

DhairyaSethi: that is totally fine imo, feeReceiver on other asset ids are not of concern

src/contracts/Hub.sol
```diff
@@ -122,13 +128,22 @@ contract Hub is IHub, AccessManaged {
...
 
     asset.updateDrawnRate(assetId);
 
+    if (asset.feeReceiver != config.feeReceiver) {
+      asset.feeReceiver = config.feeReceiver;
+      _addSpoke(assetId, config.feeReceiver);
+      _updateSpokeConfig(
+        assetId,
+        config.feeReceiver,
+        DataTypes.SpokeConfig({addCap: Constants.MAX_CAP, drawCap: Constants.MAX_CAP, active: true})
```
DhairyaSethi: should we do drawCap 0 instead?

src/contracts/Hub.sol
```diff
@@ -122,13 +128,22 @@ contract Hub is IHub, AccessManaged {
       InvalidReinvestmentController()
     );
 
-    asset.feeReceiver = config.feeReceiver;
     asset.liquidityFee = config.liquidityFee;
     asset.irStrategy = config.irStrategy;
     asset.reinvestmentController = config.reinvestmentController;
 
     asset.updateDrawnRate(assetId);
```
miguelmtzinf: move this down, so it's aligned with the rest of functions: i) accrue, ii) config updates ii) rate update

src/contracts/Hub.sol
miguelmtzinf: i'm fine being relaxed here (given immutability) and do max on both

src/contracts/HubConfigurator.sol
```diff
@@ -313,18 +299,7 @@ contract HubConfigurator is Ownable2Step, IHubConfigurator {
     if (oldFeeReceiver == newFeeReceiver) {
       return;
     }
-
```
miguelmtzinf, line 316: this function may be not needed now:
- In case of `updateFeeReceiver`: we do nothing if fee receiver does not change
- In case of `updateFeeConfig`: we change old fee receiver's caps only if changes.


I d remove `updateAssetConfig` function, because slightly defeats the purpose of this contract's functions. I d do the same for `updateSpokeConfig`, and same on `Spoke` side.

src/contracts/Hub.sol
```diff
@@ -90,6 +90,12 @@ contract Hub is IHub, AccessManaged {
       feeReceiver: feeReceiver,
       liquidityFee: 0
     });
+    _addSpoke(assetId, feeReceiver);
+    _updateSpokeConfig(
```
miguelmtzinf, line 94: leave a `@dev` comment about this. Also in `update`

tests/unit/HubConfigurator.t.sol
yan-man: yes true

src/contracts/Hub.sol
yan-man: https://github.com/aave/aave-v4/pull/686/commits/9c50005c3c5f279227899efc768c8d0dae9d5568

src/contracts/HubConfigurator.sol
yan-man: [i rm the](https://github.com/aave/aave-v4/pull/686/commits/9c50005c3c5f279227899efc768c8d0dae9d5568) `update_Configs` in the configurator methods, but: 

> In case of updateFeeConfig: we change old fee receiver's caps only if changes.

to clarify, if we rm this function, then we wont update old `feeReceiver` caps to 0 when there's a new `feeReceiver`. Is that acceptable?

src/contracts/Hub.sol
yan-man: https://github.com/aave/aave-v4/pull/686/commits/9c50005c3c5f279227899efc768c8d0dae9d5568

src/contracts/Hub.sol
miguelmtzinf: sorry, in the IHub fn I meant

src/contracts/HubConfigurator.sol
miguelmtzinf: yes yes, we still want to do it but saying that the existing internal `_updateFeeReceiverSpokeConfig` fn must be renamed and refactored, so
- we do nothing on `updateFeeReceiver` if fee receiver is the same. We change old `feeReceiver`'s caps if changed.
- we do smt similar for if `feeReceiver` is changed

tests/Base.t.sol
```diff
@@ -1890,8 +1884,8 @@ abstract contract Base is Test {
     return hub1.getAssetConfig(assetId).liquidityFee;
   }
 
-  function _getFeeReceiver(uint256 assetId) internal view returns (address) {
-    return hub1.getAssetConfig(assetId).feeReceiver;
+  function getFeeReceiver(IHub hub, uint256 assetId) internal view returns (address) {
```
CheyenneAtapour: underscore please

```suggestion
  function _getFeeReceiver(IHub hub, uint256 assetId) internal view returns (address) {
```

yan-man: https://github.com/aave/aave-v4/pull/686/commits/22ffbe8944e99d03922aed9e94d0fd273a94776a

src/contracts/Hub.sol
yan-man: https://github.com/aave/aave-v4/pull/686/commits/22ffbe8944e99d03922aed9e94d0fd273a94776a

src/contracts/HubConfigurator.sol
yan-man: it should be doing this now, updated name to `_updateOldFeeReceiverCaps` but open to naming suggestions otherwise

src/interfaces/IHub.sol
```diff
@@ -93,7 +93,7 @@ interface IHub is IHubBase, IAccessManaged {
   /**
    * @notice Adds a new asset to the hub.
    * @dev The same underlying asset address can be added as an asset multiple times.
-   * @dev The fee receiver must be configured as a Spoke separately.
+   * @dev Adds the fee receiver as a new spoke with maximum caps.
```
miguelmtzinf:
```suggestion
   * @dev The fee receiver is added as a new spoke with maximum caps, so it cannot already exist beforehand.
```

src/interfaces/IHub.sol
```diff
@@ -111,6 +111,7 @@ interface IHub is IHubBase, IAccessManaged {
 
   /**
    * @notice Updates the configuration of an asset.
+   * @dev If the fee receiver is updated, it must be a new spoke that has not yet been added on the asset.
```
miguelmtzinf:
```suggestion
   * @dev If the fee receiver is updated, it is added as a spoke, so it cannot already exist beforehand.
```

yan-man: https://github.com/aave/aave-v4/pull/686/commits/a29ee406f95772474f2cef26346169b37be42674

src/interfaces/IHub.sol
yan-man: https://github.com/aave/aave-v4/pull/686/commits/a29ee406f95772474f2cef26346169b37be42674

src/contracts/Hub.sol
DhairyaSethi: you mean feeReceiver being immutable, right? 
(bc regardless of hub, there is nothing stopping a fee receiver to draw) 

idt it will be in current plans, optics wise feels safer to have zero draw cap for this spoke even tho it's implementation doesn't have the draw functions implemented 

it's straightforward for us to maintain as well
max draw cap is likely something we'll never do in practice for regular spoke anyway i think

miguelmtzinf: good pt, ty

src/contracts/HubConfigurator.sol
```diff
@@ -294,17 +254,14 @@ contract HubConfigurator is Ownable2Step, IHubConfigurator {
...
-   *  - if new fee receiver is already a spoke, it updates the caps to max, without changing the active flag.
+   * @dev Updates the caps for the old fee receiver to 0.
    * @dev If the old and new fee receivers are the same, it does nothing.
    * @param hub The address of the Hub contract.
    * @param assetId The identifier of the asset.
    * @param oldFeeReceiver The old fee receiver.
    * @param newFeeReceiver The new fee receiver.
    */
-  function _updateFeeReceiverSpokeConfig(
+  function _updateOldFeeReceiverCaps(
```
DhairyaSethi: this method doesn't need `newFeeReceiver` anymore, could you pls rm

src/contracts/HubConfigurator.sol
```diff
@@ -313,18 +270,7 @@ contract HubConfigurator is Ownable2Step, IHubConfigurator {
     if (oldFeeReceiver == newFeeReceiver) {
       return;
     }
```
DhairyaSethi: not needed

src/interfaces/IHub.sol
```diff
@@ -93,7 +93,7 @@ interface IHub is IHubBase, IAccessManaged {
   /**
    * @notice Adds a new asset to the hub.
    * @dev The same underlying asset address can be added as an asset multiple times.
-   * @dev The fee receiver must be configured as a Spoke separately.
+   * @dev The fee receiver is added as a new spoke with maximum add cap and zero draw cap, so it cannot already exist beforehand.
```
```suggestion
   * @dev The fee receiver is added as a new spoke with maximum add cap and zero draw cap.
```
added implies it cannot exist before hand (we should specifiy in `addSpoke` that adding a spoke ensures spoke cannot be registered already, with something like `Reverts if spoke is already registered with SpokePresent()`

tests/unit/Hub/Hub.Config.t.sol
```diff
@@ -339,6 +352,28 @@ contract HubConfigTest is HubBase {
     uint256 liquidity = hub1.getLiquidity(assetId);
     (uint256 drawn, uint256 premium) = hub1.getAssetOwed(assetId);
 
+    console.log(
+      'newConfig.feeReceiver',
+      newConfig.feeReceiver,
+      hub1.isSpokeListed(assetId, newConfig.feeReceiver)
+    );
```
DhairyaSethi: pls rm

tests/unit/HubConfigurator.t.sol
```diff
@@ -504,69 +475,39 @@ contract HubConfiguratorTest is HubBase {
...
-              })
-            )
-          )
+          abi.encodeCall(IHub.updateAssetConfig, (assetId, expectedConfig))
         );
+      } else {
+        expectedConfig.liquidityFee = oldConfig.liquidityFee;
+        // if new fee receiver is different from old one, and is already listed, revert
+        vm.expectRevert(IHub.SpokeAlreadyListed.selector, address(hub1));
       }
```
DhairyaSethi, line 507: not a fan of so many nested branches in this test, would prefer to test each case individually but nbd for now

tests/unit/SpokeConfigurator.t.sol
```diff
@@ -507,41 +507,6 @@ contract SpokeConfiguratorTest is SpokeBase {
     assertEq(spoke.getDynamicReserveConfig(reserveId, configKey), expectedDynamicReserveConfig);
   }
 
-  function test_updateReserveConfig_revertsWith_OwnableUnauthorizedAccount() public {
```
DhairyaSethi, line 510: why was this rmed?

src/contracts/HubConfigurator.sol
```diff
@@ -138,29 +119,19 @@ contract HubConfigurator is Ownable2Step, IHubConfigurator {
...
     for (uint256 i = 0; i < spokesCount; ++i) {
       address spokeAddress = targetHub.getSpokeAddress(assetId, i);
-      _updateSpokeCaps(targetHub, assetId, spokeAddress, 0, 0);
+      _updateSpokeCaps({
+        hub: targetHub,
+        assetId: assetId,
+        spoke: spokeAddress,
+        addCap: 0,
+        drawCap: 0
+      });
```
DhairyaSethi: named args not needed here imo

tests/unit/HubConfigurator.t.sol
yan-man: prefer that as well generally but this test is called from other test scenarios so requires bigger refactor to separate

src/contracts/HubConfigurator.sol
yan-man: was thinking bc of addCap drawCap to include but they're both 0 anyway so good pt

edit: https://github.com/aave/aave-v4/pull/686/commits/bc04c7eeed118b7f6afa0dcc6a843c20d30b9dcd

tests/unit/SpokeConfigurator.t.sol
yan-man: see here: https://github.com/aave/aave-v4/pull/686#discussion_r2324415117

tests/unit/Hub/Hub.Config.t.sol
yan-man: https://github.com/aave/aave-v4/pull/686/commits/bc04c7eeed118b7f6afa0dcc6a843c20d30b9dcd

src/interfaces/IHub.sol
yan-man: https://github.com/aave/aave-v4/pull/686/commits/bc04c7eeed118b7f6afa0dcc6a843c20d30b9dcd

src/contracts/HubConfigurator.sol
yan-man: also see above, https://github.com/aave/aave-v4/pull/686/commits/bc04c7eeed118b7f6afa0dcc6a843c20d30b9dcd

src/contracts/HubConfigurator.sol
yan-man: decided to just rm this method to simplify as it just does one thing

https://github.com/aave/aave-v4/pull/686/commits/bc04c7eeed118b7f6afa0dcc6a843c20d30b9dcd

src/interfaces/IHub.sol
```diff
@@ -111,13 +111,15 @@ interface IHub is IHubBase, IAccessManaged {
 
   /**
    * @notice Updates the configuration of an asset.
+   * @dev If the fee receiver is updated, it is added as a new spoke with maximum add cap and zero draw cap, so it cannot already exist beforehand.
```
```suggestion
   * @dev If the fee receiver is updated, it is added as a new spoke with maximum add cap and zero draw cap.
```

yan-man: https://github.com/aave/aave-v4/pull/686/commits/c7bc58a196e519f6a0cd6c257c84fb37e578f713

PR #687, opened 2025-09-03
FeeReceiver as separate address instead of Spoke

yan-man:
Fee receiver as a separate address instead of acting as an Spoke.
- New function to claim fees
- New getters for fee shares and fee assets
- New mapping for fee receivers per asset (with memory-protection: old fee receivers will keep their shares forever)

Benefits:
- Increases clarity / maintainability, as fee receiver is just a regular address and no need to handle special/edge cases.

Downsides:
- Need of ad-hoc variables for amount of fee shares and fee receiver address (and corresponding events)
- Restricts functionality to only claim (e.g. cannot supply more or borrow against its position before claiming first)

Changes on system's invariants:
- `Total number of shares = sum of spokes' shares` -> `Total number of shares = sum of spokes' shares + sum of feeReceivers' shares`


src/contracts/Hub.sol
```diff
@@ -33,6 +33,8 @@ contract Hub is IHub, AccessManaged {
   mapping(uint256 assetId => DataTypes.Asset assetData) internal _assets;
   mapping(uint256 assetId => mapping(address spoke => DataTypes.SpokeData spokeData))
     internal _spokes;
+  mapping(uint256 assetId => mapping(address feeReceiver => DataTypes.FeeReceiver feeReceiverData))
```
miguelmtzinf, line 36: we could simplify by:
i) not granular by asset -> only 1 for all
ii) no update-friendly -> if feeReceiver is changed, all fees can be claimed by the new address

this d lead to only 2 storage vars -> `feeReceiver address` and `feeShares uint256`

src/contracts/Hub.sol
```diff
@@ -209,12 +211,35 @@ contract Hub is IHub, AccessManaged {
...
+    uint128 shares = previewRemoveByAssets(assetId, amount).toUint128(); // non zero since we round up
+    asset.addedShares -= shares;
+    feeReceiver.addedShares -= shares;
+    asset.liquidity = liquidity.uncheckedSub(amount).toUint128();
+
+    asset.updateDrawnRate(assetId);
+
+    IERC20(asset.underlying).safeTransfer(to, amount);
+
+    emit Remove(assetId, msg.sender, shares, amount);
```
miguelmtzinf: we d need a diff event i guess, since it's designed for spokes (2nd param is `spoke`)

src/contracts/Hub.sol
```diff
@@ -667,6 +688,19 @@ contract Hub is IHub, AccessManaged {
     require(amount <= withdrawable, AddedAmountExceeded(withdrawable));
   }
 
+  function _validateRemoveFeeShares(
+    DataTypes.FeeReceiver storage feeReceiver,
+    uint256 assetId,
+    uint256 amount,
+    address to
+  ) internal view {
+    require(msg.sender == asset.feeReceiver, OnlyFeeReceiver()); // TODO: pass msg.sender as an arg
```
miguelmtzinf: this is not consistent with storing a mapping of feeReceivers... old feeReceiver addresses must be able to claim their shares even if they are not active currenlty

src/contracts/Hub.sol
```diff
@@ -334,9 +359,13 @@ contract Hub is IHub, AccessManaged {
...
     _validatePayFee(sender, shares);
-    _transferShares(sender, receiver, shares);
+    uint256 addedShares = sender.addedShares;
+    require(shares <= addedShares, AddedSharesExceeded(addedShares));
+
+    sender.addedShares = addedShares.uncheckedSub(shares).toUint128();
+    receiver.addedShares += shares.toUint128();
     asset.updateDrawnRate(assetId);
 
     emit TransferShares(assetId, shares, msg.sender, feeReceiver);
```
miguelmtzinf, line 371: same here, we d need a diff event i guess

miguelmtzinf:
Benefits:
1. Increases clarity / maintainability, as fee receiver is just a regular address and no need to handle special/edge cases.

Downsides:
1. Need of ad-hoc variables for amount of fee shares and fee receiver address (and corresponding events)
3. Restricts functionality to only claim (e.g. cannot supply more or borrow against its position before claiming first)

src/contracts/Hub.sol
yan-man: good pt

src/contracts/Hub.sol
yan-man: agree but left it as is here given it doesnt affect the removeFeeShares or gas snapshots

src/contracts/Hub.sol
yan-man: however, feeShares would still need to be per-asset

edit: created a POC for the simple feeReceiver here: https://github.com/aave/aave-v4/pull/692/files

src/contracts/Hub.sol
```diff
@@ -554,6 +587,20 @@ contract Hub is IHub, AccessManaged {
     return _spokes[assetId][spoke].addedShares;
   }
 
+  function getFeeAmount(uint256 assetId, address feeReceiver) external view returns (uint256) {
+    DataTypes.Asset storage asset = _assets[assetId];
+    return
+      previewRemoveByShares(
+        assetId,
+        _feeReceivers[assetId][feeReceiver].addedShares + asset.unrealizedFeeShares()
```
miguelmtzinf, line 595: `unrealizedFeeShares` only matter for the active fee receiver. We should also accumulate unrealized shares into addShares of the fee receiver when updating.

src/contracts/Hub.sol
```diff
@@ -554,6 +587,20 @@ contract Hub is IHub, AccessManaged {
     return _spokes[assetId][spoke].addedShares;
```
miguelmtzinf, line 587: this function can be simplified now, as the fee receiver is not an spoke anymore

yan-man:
decided today feeReceiver is a spoke, see https://github.com/aave/aave-v4/pull/686

PR #688, opened 2025-09-03
fix: block enableAsCollateral when supplied balance is 0

### LINKED ISSUE #678: enableAsCollateral validation: revert when existing balance is 0
on https://github.com/aave/aave-v4/blob/8f9301e10feeb81b528ea0d0bf54fe58ddb5d44c/src/contracts/Spoke.sol#L376-L378
enabling as collateral, revert if existing user supply is 0

helps with edge cases mentioned in https://github.com/aave/aave-v4/issues/504 & follows v3 semantics https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/protocol/libraries/logic/SupplyLogic.sol#L281-L284

Kogaroshi:
Fixes #678 

WIP : need to find better order to keep all `SpokeOperations_Gas_Tests` gas checks in `test_supply`.


src/contracts/Spoke.sol
```diff
@@ -374,6 +374,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     positionStatus.setUsingAsCollateral(reserveId, usingAsCollateral);
 
     if (usingAsCollateral) {
+      if (_userPositions[onBehalfOf][reserveId].suppliedShares == 0) revert ReserveBalanceNull();
```
DhairyaSethi: check should be in validateSetUsingAsCollateral when usingAsCollateral is true

src/contracts/Spoke.sol
```diff
@@ -369,7 +369,11 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) return;
 
     DataTypes.Reserve storage reserve = _reserves[reserveId];
-    _validateSetUsingAsCollateral(reserve, usingAsCollateral);
+    _validateSetUsingAsCollateral(
+      reserve,
+      _userPositions[onBehalfOf][reserveId].suppliedShares,
```
DhairyaSethi: pls declare storage pointer above and pass that, following convention

src/contracts/Spoke.sol
```diff
@@ -733,16 +737,19 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
...
   function _validateSetUsingAsCollateral(
     DataTypes.Reserve storage reserve,
+    uint256 userSuppliedShares,
     bool usingAsCollateral
   ) internal view {
     require(address(reserve.hub) != address(0), ReserveNotListed());
     require(!reserve.paused, ReservePaused());
     // deactivation should be allowed
     require(!usingAsCollateral || !reserve.frozen, ReserveFrozen());
+    require(!usingAsCollateral || userSuppliedShares > 0, ReserveBalanceNull());
```
DhairyaSethi: these two can clubbed in one if branch, think it'd be cheaper? 

also can you pls see if we can reuse any existing error? maybe invalid reserve (even tho not technically correct) or something, would prefer to avoid adding a new error for code size but nbd

tests/unit/Spoke/Spoke.SetUsingAsCollateral.t.sol
```diff
@@ -51,9 +60,22 @@ contract SpokeConfigTest is SpokeBase {
...
+    spoke1.setUsingAsCollateral(daiReserveId, true, alice);
+  }
+
   /// no action taken when collateral status is unchanged
   function test_setUsingAsCollateral_collateralStatusUnchanged() public {
     uint256 daiReserveId = _daiReserveId(spoke1);
+    uint256 daiAmount = 100e18;
+
+    // Bob supply dai into spoke1
+    deal(address(tokenList.dai), bob, daiAmount);
```
DhairyaSethi: bob is already dealt max supply amount in base, this is not needed
and then declaring daiAmount separately is not needed as well

tests/unit/Spoke/Spoke.SetUsingAsCollateral.t.sol
```diff
@@ -51,9 +60,22 @@ contract SpokeConfigTest is SpokeBase {
     spoke1.setUsingAsCollateral(daiReserveId, true, alice);
   }
 
+  function test_setUsingAsCollateral_revertsWith_ReserveBalanceNull() public {
+    uint256 daiReserveId = _daiReserveId(spoke1);
+
+    vm.expectRevert(ISpoke.ReserveBalanceNull.selector);
+    vm.prank(alice);
+    spoke1.setUsingAsCollateral(daiReserveId, true, alice);
```
DhairyaSethi, line 68: pls test path when disabling and zero collateral remaining

tests/unit/Spoke/Spoke.PositionManager.t.sol
```diff
@@ -192,6 +192,9 @@ contract SpokePositionManagerTest is SpokeBase {
 
   function test_onlyPositionManager_on_usingAsCollateral() public {
     uint256 reserveId = _usdxReserveId(spoke1);
+    uint256 amount = 100e6;
```
DhairyaSethi: not needed, can be inlined

tests/gas/Spoke.Operations.gas.t.sol
```diff
@@ -24,9 +24,9 @@ contract SpokeOperations_Gas_Tests is Base {
     vm.snapshotGasLastCall('Spoke.Operations', 'supply: 0 borrows, collateral disabled');
     skip(100);
 
-    spoke1.setUsingAsCollateral(_wethReserveId(spoke1), true, alice);
     spoke1.supply(_wethReserveId(spoke1), 1e18, alice);
-    vm.snapshotGasLastCall('Spoke.Operations', 'supply: 0 borrows, collateral enabled');
+    spoke1.setUsingAsCollateral(_wethReserveId(spoke1), true, alice);
+    //vm.snapshotGasLastCall('Spoke.Operations', 'supply: 0 borrows, collateral enabled');
```
DhairyaSethi: yeah let's rm this, its fine 
also pls run lint

src/contracts/Spoke.sol
Kogaroshi: I didn't find an error matching the case, that's why I added a new one.
But if we merge the requires for `ReserveBalanceNull` &  `ReserveFrozen` in 1 require, we can replace `ReserveFrozen()` by something like `ReserveNotAllowed` or `ReserveNotCollateral`, wdyt ?

src/contracts/Spoke.sol
```diff
@@ -369,7 +369,12 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) return;
 
     DataTypes.Reserve storage reserve = _reserves[reserveId];
-    _validateSetUsingAsCollateral(reserve, usingAsCollateral);
+    uint256 userSuppliedShares = _userPositions[onBehalfOf][reserveId].suppliedShares;
+    _validateSetUsingAsCollateral(
+      reserve,
+      userSuppliedShares,
+      usingAsCollateral
+    );
```
```suggestion
    DataTypes.UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
    _validateSetUsingAsCollateral(
      reserve,
      userPosition,
      usingAsCollateral
    );
```

PR #690, opened 2025-09-03
feat : NativeToken Gateway for Spokes

### LINKED ISSUE #689: WrappedTokenGateway V4
Add a WrappedTokenGateway working as a PositionManager for Aave V4.

Ref V3 : https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/helpers/WrappedTokenGatewayV3.sol

The contract would also need to implement Multicall to enable approving the manager permission on the user position and execute the actions.

Kogaroshi:
Spoke-bound Gateways allowing to use wrap/unwrap native assets before interacting with a Spoke. The Gateway works as a Position Manager, executing the actions on behalf of users.
The Gateway also proposes a `multicall`, allowing users to allow the Position Manager and execute the action in a single transaction.

closes #689

### DIFF
--- a/src/interfaces/INativeTokenGateway.sol
+++ b/src/interfaces/INativeTokenGateway.sol
@@ -6,28 +6,28 @@ import {IRescuable} from 'src/interfaces/IRescuable.sol';
 
 interface INativeTokenGateway is IRescuable {
   /**
-  @notice Thrown when the given address is invalid.
-  */
+   * @notice Thrown when the given address is invalid.
+   */
   error InvalidAddress();
 
   /**
-  @notice Thrown when the given amount is invalid.
-  */
+   * @notice Thrown when the given amount is invalid.
+   */
   error InvalidAmount();
 
   /**
-  @notice Thrown when the given reserveId does not match the wrapped native asset reserveId on the Spoke.
-  */
+   * @notice Thrown when the given reserveId does not match the wrapped native asset reserveId on the Spoke.
+   */
   error InvalidReserveId();
 
   /**
-  @notice Thrown when the native amount sent does not match the given amount parameter.
-  */
+   * @notice Thrown when the native amount sent does not match the given amount parameter.
+   */
   error NativeAmountMismatch();
 
   /**
-  @notice Thrown when trying to call an unsupported action or sending native assets to this contract directly.
-  */
+   * @notice Thrown when trying to call an unsupported action or sending native assets to this contract directly.
+   */
   error UnsupportedAction();
 
   /**


src/dependencies/solady/ReentrancyGuard.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.4;
+
+/// @notice Reentrancy guard mixin.
+/// @author Solady (https://github.com/vectorized/solady/blob/main/src/utils/ReentrancyGuard.sol)
+abstract contract ReentrancyGuard {
```
DhairyaSethi: prefer to use OZ for standard stuff where gas is not a concern 
also lets use transient variant?

src/interfaces/IWrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+interface IWrappedTokenGatewayV4 {
+    error AmountNull();
```
DhairyaSethi: lint

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,165 @@
...
+import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+import {IWrappedTokenGatewayV4} from 'src/interfaces/IWrappedTokenGatewayV4.sol';
+import {ISpoke} from 'src/interfaces/ISpoke.sol';
+import {Multicall} from 'src/misc/Multicall.sol';
+
+contract WrappedTokenGatewayV4 is
+  IWrappedTokenGatewayV4,
+  Multicall,
+  ReentrancyGuard,
+  AccessManaged
```
DhairyaSethi: just for rescue, no? let's use ownable2step instead here

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,165 @@
...
+
+  function setUserPositionManagerWithSig(
+    address user,
+    bool approve,
+    uint256 deadline,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external {
+    require(user != address(0), AddressZero());
```
DhairyaSethi: should this be checked on the spoke instead? what's the thinking btw

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  function supplyNative(
+    uint256 reserveId,
+    uint256 amount,
+    address onBehalfOf
+  ) external payable nonReentrant {
+    _verifyParameters(reserveId, amount, onBehalfOf);
+
+    if (msg.value != amount) revert NativeAmountMismatch();
```
DhairyaSethi: require style for consistency

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,165 @@
...
+
+    _wrapNative(amount);
+    IERC20(address(WRAPPED_ASSET)).safeIncreaseAllowance(_getReserveHub(reserveId), amount);
```
DhairyaSethi: using safeErc20 for * or Weth9 would avoid this extra cast no?

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,165 @@
...
+    bytes32 s
+  ) external {
+    require(user != address(0), AddressZero());
+    SPOKE.setUserPositionManagerWithSig(address(this), user, approve, deadline, v, r, s);
+  }
+
+  function supplyNative(
+    uint256 reserveId,
+    uint256 amount,
+    address onBehalfOf
```
DhairyaSethi: if this is forever approved as posm on the users behalf then anyone can use it and bypass the posm check through this, no?

DhairyaSethi: ok we transfer assets to the user on withdraw, borrow so it should be fine

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: mistake when adding input checks in other methods, must have paste it here without noticing it wasn't needed.

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Yeah, idea is that it's always approved, but to deposit/repay it's the caller that pays, and when withdraw/borrow it's sent to the user. But maybe an extra check that on borrow/withdraw onBehalf must be the caller.

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Doesn't work, since SafeERC20 expects IERC20 parameter, hence the extra cast. Other solution is to skip using SafeERC20 and call the draw `approve()` on WETH9.

src/interfaces/IWrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,27 @@
...
+  error AmountNull();
+  error AddressZero();
+  error InvalidReserveId();
+  error InvalidCaller();
+  error NativeAmountMismatch();
+  error NativeTransferFailed();
+  error FallbackForbidden();
+  error ReceiveNotAllowed();
+
+  function supplyNative(uint256 reserveId, uint256 amount, address onBehalfOf) external payable;
```
DhairyaSethi: pls add natspec for all

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
+import {SafeERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
+import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
+import {ReentrancyGuardTransient} from 'src/dependencies/openzeppelin/ReentrancyGuardTransient.sol';
+
+import {WETH9} from 'src/dependencies/weth/WETH9.sol';
```
DhairyaSethi: can we use IWETH interface instead pls

DhairyaSethi: also follow order of imports convention as per style guide

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+
+contract WrappedTokenGatewayV4 is
+  IWrappedTokenGatewayV4,
+  Multicall,
+  ReentrancyGuardTransient,
+  Ownable2Step
+{
+  using SafeERC20 for IERC20;
+
+  WETH9 public immutable WRAPPED_ASSET;
```
DhairyaSethi:
```suggestion
  WETH9 public immutable WETH;
```
since we'll use weth interface and not INativeAsset, it's fine

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+
+  function setUserPositionManagerWithSig(
+    address user,
+    bool approve,
+    uint256 deadline,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external {
+    SPOKE.setUserPositionManagerWithSig(address(this), user, approve, deadline, v, r, s);
```
DhairyaSethi: pls also expose another fn for SPOKE.renouncePositionManager

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  function supplyNative(
+    uint256 reserveId,
+    uint256 amount,
+    address onBehalfOf
+  ) external payable nonReentrant {
+    _verifyParameters(reserveId, amount, onBehalfOf, msg.sender);
```
DhairyaSethi: _validateParams

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: lets use approve and rm safeerc20 here since we know it's a fixed underlying asset

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+  ) external nonReentrant {
+    _verifyParameters(reserveId, amount, onBehalfOf, msg.sender);
+
+    uint256 userSuppliedAmount = SPOKE.getUserSuppliedAmount(reserveId, onBehalfOf);
+    if (amount == type(uint256).max) {
+      amount = userSuppliedAmount;
+    }
+
+    SPOKE.withdraw(reserveId, amount, onBehalfOf);
+    _unwrapNative(amount);
```
DhairyaSethi: since the internal method only abstracts one line, let's avoid it imo

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+    _transferNative(onBehalfOf, amount);
```
DhairyaSethi: can you use OZ safe transfer eth from Address.sol or somewhere pls

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+    uint256 amount,
+    address onBehalfOf
+  ) external payable nonReentrant {
+    _verifyParameters(reserveId, amount, onBehalfOf, msg.sender);
+    require(msg.value == amount, NativeAmountMismatch());
+
+    uint256 userDebtAmount = SPOKE.getUserTotalDebt(reserveId, onBehalfOf);
+    uint256 leftovers;
+    if (amount > userDebtAmount) {
+      leftovers = amount - userDebtAmount;
```
DhairyaSethi: won't reach `leftovers` bc of the strict require on msg.value == amount on line 92, no? 

can make it >= instead like v3 https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/helpers/WrappedTokenGatewayV3.sol#L83

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+    require(_getReserveAsset(reserveId) == address(WRAPPED_ASSET), InvalidReserveId());
+  }
+
+  function _getReserveAsset(uint256 reserveId) internal view returns (address) {
+    return SPOKE.getReserve(reserveId).underlying;
+  }
+
+  function _getReserveHub(uint256 reserveId) internal view returns (address) {
+    return address(SPOKE.getReserve(reserveId).hub);
+  }
```
DhairyaSethi: combine these two usages in one call i think? would be cheaper

we should expose convenience methods, dont like it's reading the full Reserve struct

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+
+  function _unwrapNative(uint256 amount) internal {
+    WRAPPED_ASSET.withdraw(amount);
+  }
+
+  function _transferNative(address to, uint256 amount) internal {
+    (bool success, ) = to.call{value: amount}(new bytes(0));
+    require(success, NativeTransferFailed());
+  }
```
DhairyaSethi: rm, pls use OZ abstraction

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: My thoughts are : 
- check the inputed amount is the same as the msg.value
- check if the amount inputed is more than the debt, reduce `amount` to only repay the debt, and we'll have leftovers from msg.value
- we return the leftovers

That allows user to input an amount more than their current debt to account for the interest accrual between the tx signature and the execution.

in V3 there is a >= because they first decrease the amount to match current debt, and then check the msg.value

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: >  But maybe an extra check that on borrow/withdraw onBehalf must be the caller.

idt needed?

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: If so, we'd need to call this before `_validateParams`, and pass the asset address as input of `_validateParams`, and only alloc the Hub address in the case of deposit/repay, to prevent calling twice in the same execution.
If this behavior is acceptable, let's merge it yeah

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: ah ok makes sense!

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: in the cases of `approve == false`, so we save gas ?

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,164 @@
...
+
+import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+import {IWrappedTokenGatewayV4} from 'src/interfaces/IWrappedTokenGatewayV4.sol';
+import {ISpoke} from 'src/interfaces/ISpoke.sol';
+import {Multicall} from 'src/misc/Multicall.sol';
+
+contract WrappedTokenGatewayV4 is
+  IWrappedTokenGatewayV4,
+  Multicall,
+  ReentrancyGuardTransient,
```
DhairyaSethi: these two should work fine if we batch supply, borrow, repay, right?

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: yeah if you can find a clean way to do it

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Was solved by using IWETH that inherits IERC20

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Do we want anyone to be able to call `withdrawNative` with an user that approved the Position Manager, removing their funds (even if it sends to their wallet), it could allow griefing

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Since each call in the batch it's fully executed before starting the next one, the guard will reset after each item in the batch, I don't see a case where it would block.

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: no no

this fn allows a posm to renounce its posm approval for a given user. since this contract is a posm it needs to have all posm functionality exposed

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: yep

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Yeah forgot to update my comment, understood when checking the function, added a function exposing it in the contract

Kogaroshi: https://github.com/aave/aave-v4/blob/c6ee56357cd1d863ed675a5fe8a247e2fffacce5/src/misc/WrappedTokenGatewayV4.sol#L47

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Merge the helpers in c6ee56357cd1d863ed675a5fe8a247e2fffacce5, if not clean enough I can revert.

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: im thinking we always include renouncePosm so the posm status is not left hanging. with this check we disallow having gas sponsorship

Kogaroshi: True, tho I'm not fully comfortable letting such an option for impacting negatively an user position if the renounce isn't included in the batch in Multicall.

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,159 @@
...
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external {
+    SPOKE.setUserPositionManagerWithSig(address(this), user, approve, deadline, v, r, s);
+  }
+
+  function renouncePositionManagerRole() external {
+    SPOKE.renouncePositionManagerRole(msg.sender);
+  }
```
DhairyaSethi:
```suggestion
  function renouncePositionManagerRole(address user) external {
    SPOKE.renouncePositionManagerRole(user);
  }
```
let's discuss the drawbacks of making this permissionless - i think we should

Kogaroshi: Yeah agree, permissionless would be better, so we can remove the users that forgot to remove the authorization.
Only issue I could see with that is, if an user does not use Multicall, and does a pre-approve to then execute the action, you could call in-between to deny it.
So it kind of pushes the required behavior to use the Multicall.

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: yeah let's make this version not support gas sponsorship at all for now

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: yeah i think that's fine

src/interfaces/IWrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+interface IWrappedTokenGatewayV4 {
+  error AmountNull();
+  error AddressZero();
```
DhairyaSethi: we're using `InvalidAddress` for this everywhere in the core code, let's follow that
and `AmountZero` or something i think, let's use those for consistency

src/interfaces/IWrappedTokenGatewayV4.sol
DhairyaSethi: if the suffix is Native then let's use INative instead of IWETH or use supplyETH here, your call. i'm good with either, very slightly leaning towards generic INative but agnostic

Kogaroshi: I'd rather go for Native since we most likely will deploy on non-ETH networks ?

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,159 @@
...
+
+import {Multicall} from 'src/misc/Multicall.sol';
+
+import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
+import {SafeERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
+import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
+import {Address} from 'src/dependencies/openzeppelin/Address.sol';
+import {ReentrancyGuardTransient} from 'src/dependencies/openzeppelin/ReentrancyGuardTransient.sol';
+
+import {IWETH} from 'src/dependencies/weth/IWETH.sol';
```
DhairyaSethi:
```suggestion
import {ReentrancyGuardTransient} from 'src/dependencies/openzeppelin/ReentrancyGuardTransient.sol';
import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
import {SafeERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
import {Address} from 'src/dependencies/openzeppelin/Address.sol';
import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
import {IWETH} from 'src/dependencies/weth/IWETH.sol';
```

src/interfaces/IWrappedTokenGatewayV4.sol
DhairyaSethi: ofc yep poly, avax, bnb

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+
+  function test_constructor() public {
+    WrappedTokenGatewayV4 gateway = new WrappedTokenGatewayV4(
+      address(tokenList.weth),
+      address(spoke),
+      address(ADMIN)
+    );
+
+    assertEq(address(gateway.WETH()), address(tokenList.weth));
+    assertEq(address(gateway.SPOKE()), address(spoke));
```
DhairyaSethi: check ownable stuff

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+      r,
+      s
+    );
+
+    assertEq(spoke.isPositionManager(user, address(wrappedTokenGateway)), params.approve);
+  }
+
+  function test_renouncePositionManagerRole() public {
+    (address user, uint256 userPk) = makeAddrAndKey(string(vm.randomBytes(32)));
+    vm.label(user, 'user');
```
DhairyaSethi: not needed

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+    emit ISpokeBase.Supply(wethReserveId, address(wrappedTokenGateway), bob, amount);
+    vm.prank(bob);
+    wrappedTokenGateway.supplyNative{value: amount}(wethReserveId, amount, bob);
+
+    DataTypes.UserPosition memory finalUserData = _getUserData(bob);
+    DataTypes.UserPosition memory finalGatewayData = _getUserData(address(wrappedTokenGateway));
+    SpokePosition memory finalReserveData = getSpokePosition(spoke, wethReserveId);
+
+    // weth balance
+    assertEq(tokenList.weth.balanceOf(address(hub1)), amount, 'hub token balance after-supply');
```
DhairyaSethi: lets add these labels wherever they make sense and could help with debugging, im personally not a fan of how these are written from where you lifted this test, we can make it slightly cleaner

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+    // native balances
+    assertEq(bob.balance, mintAmount_WETH - amount, 'user native balance after-supply');
+    assertEq(
+      address(wrappedTokenGateway).balance,
+      0,
+      'wrappedTokenGateway native balance after-supply'
+    );
+  }
+
+  function test_supplyNative_fuzz(uint256 amount) public {
```
DhairyaSethi: `test_supplyNative` can just call `test_supplyNative_fuzz({amount: 100e18})`

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+      'wrappedTokenGateway native balance after-supply'
+    );
+  }
+
+  function test_supplyNative_revertsWith_AmountNull() public {
+    uint256 amount = 100e18;
+    deal(bob, mintAmount_WETH);
+
+    vm.prank(bob);
+    spoke.setUserPositionManager(address(wrappedTokenGateway), true);
```
DhairyaSethi: not needed to get to the revert, no?

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+contract WrappedTokenGatewayV4Test is Base {
+  WrappedTokenGatewayV4 public wrappedTokenGateway;
+  ISpoke public spoke;
+  uint256 public wethReserveId;
+
+  function setUp() public virtual override {
+    super.setUp();
+    initEnvironment();
+
+    spoke = ISpoke(address(spoke1));
```
DhairyaSethi: spoke1 is already casted as ISpoke?

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+
```
DhairyaSethi: pls rm unnecessary whitespace

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+    wrappedTokenGateway = new WrappedTokenGatewayV4(
+      address(tokenList.weth),
+      address(spoke),
+      address(ADMIN)
+    );
+
+    vm.prank(SPOKE_ADMIN);
+    spoke.updatePositionManager(address(wrappedTokenGateway), true);
+
+    deal(address(tokenList.weth), 1e36);
```
DhairyaSethi:
```suggestion
    deal(address(tokenList.weth), MAX_SUPPLY_AMOUNT);
```

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+  struct WithdrawLocalVars {
+    uint256 expectedSupplyShares;
+    uint256 prevUserSuppliedAmount;
+    uint256 prevHubBalance;
+    uint256 prevSpokeBalance;
+    uint256 prevHubAddedAmount;
+    uint256 prevSpokeSupplied;
+  }
+
+  function test_withdrawNative() public {
```
DhairyaSethi: same, can call fuzz variant

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+      wethReserveId,
+      address(wrappedTokenGateway),
+      bob,
+      vars.expectedSupplyShares
+    );
+    vm.prank(bob);
+    wrappedTokenGateway.withdrawNative(wethReserveId, UINT256_MAX, bob);
+
+    // treasury spoke withdraw fees
+    //withdrawLiquidityFees(wethAssetId, UINT256_MAX);
```
DhairyaSethi: rm?

src/interfaces/IWrappedTokenGatewayV4.sol
Kogaroshi: Okay, changing back to Native & WRAPPED_ASSET & renaming the interface.

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+
+  function test_receive_revertsWith_ReceiveNotAllowed() public {
+    deal(address(this), 1 ether);
+
+    vm.expectRevert(IWrappedTokenGatewayV4.ReceiveNotAllowed.selector);
+    address(wrappedTokenGateway).call{value: 1 ether}(new bytes(0));
+  }
+
+  // To receive recovered native assets
+  fallback() external payable {}
```
DhairyaSethi: can you pls use alice instead of the test contract instead?

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+
+    vm.prank(bob);
+    spoke.setUserPositionManager(address(wrappedTokenGateway), true);
+
+    // Bob supply dai collateral
+    Utils.supplyCollateral(spoke, _daiReserveId(spoke), bob, bobSupplyAmount, bob);
+    // Alice supply weth
+    Utils.supply(spoke, wethReserveId, alice, aliceSupplyAmount, alice);
+    //Bob borrows weth
+    Utils.borrow(spoke, wethReserveId, bob, borrowAmount, bob);
```
DhairyaSethi: comments unneeded imo

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,1295 @@
...
+
+    uint256 prevUserBalance = bob.balance;
+    uint256 prevHubBalance = tokenList.weth.balanceOf(address(hub1));
+    uint256 prevSpokeBalance = tokenList.weth.balanceOf(address(spoke));
+
+    /*vm.expectEmit(address(spoke));
+    emit ISpokeBase.Repay(wethReserveId, address(wrappedTokenGateway), bob, hub1.convertToDrawnShares(wethAssetId, repayAmount));*/
```
DhairyaSethi: to bring back, right?

Kogaroshi: Yeah, need to bring helper function to calculate correctly

src/dependencies/weth/INativeWrapper.sol
DhairyaSethi, line 1: pls move to src/interfaces

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,828 @@
...
+  function test_borrowNative_revertsWith_InvalidAddress() public {
+    uint256 borrowAmount = 5e18;
+
+    vm.prank(bob);
+    vm.expectRevert(IWrappedTokenGatewayV4.InvalidAddress.selector);
+    wrappedTokenGateway.borrowNative(wethReserveId, borrowAmount, address(0));
+  }
+
+  // Taken from SpokeBase.t.sol
+  function _getExpectedPremiumDelta(
```
DhairyaSethi: pls don't duplicate code, this helper can be moved to Base.t.sol and reused from there

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+    }
+  }
+
+  /**
+   * @notice Call multiple functions in the current contract and return the data from all of them if they all succeed.
+   * @dev We need to copy here instead of inheriting to make the function payable
+   * @param data The encoded function data for each of the calls to make to this contract.
+   * @return results The results from each of the calls passed in via data.
+   */
+  function multicall(bytes[] calldata data) external payable returns (bytes[] memory) {
```
DhairyaSethi: no haha i didn't mean to copy paste it here, we should inherit only

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function setUserPositionManagerWithSig(
+    address user,
+    bool approve,
+    uint256 deadline,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external payable {
+    SPOKE.setUserPositionManagerWithSig(address(this), user, approve, deadline, v, r, s);
```
DhairyaSethi: this can be mode into a delegatecall w the user param as msg.sender, with the normal setUserPosm (w/o withSig) - was my idea

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: I was inheriting at first, but not compatible with this contract, since we need to make the `multicall()` a payable method.
Otherwise in Multicall.sol we need to change the method to be payable, but not sure of the effects it could have for the Spoke.

DhairyaSethi: yeah i'd intentionally removed the payable initially but you can make it payable again now that we need it 
idt there's effects on the spoke, but let's evaluate effects here

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Solution was tested and not working. If we find another solution to bypass the need for signature, we'll introduce it later on.

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+ */
+contract WrappedTokenGatewayV4 is IWrappedTokenGatewayV4, ReentrancyGuardTransient, Ownable2Step {
+  using SafeERC20 for *;
+
+  /// @notice Native Wrapper contract
+  INativeWrapper public immutable NATIVE_WRAPPER;
+  /// @notice Spoke contract
+  ISpoke public immutable SPOKE;
+
+  constructor(address nativeWrapper_, address spoke_, address admin_) Ownable(admin_) {
```
DhairyaSethi: zero addr checks for constructor params

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+    DataTypes.Reserve memory reserveData = SPOKE.getReserve(reserveId);
+    return (reserveData.underlying, address(reserveData.hub));
+  }
+
+  /**
+   * @notice Recovers ERC20 tokens sent to this contract.
+   * @param token Address of the ERC20 token to recover.
+   * @param to Address to send the recovered tokens to.
+   **/
+  function recoverToken(address token, address to) external onlyOwner {
```
DhairyaSethi: these two external methods should go before _validateParams and inheritdoc natspec

tests/unit/WrappedTokenGatewayV4.t.sol
DhairyaSethi, line 1: could we add an assertion that for every happy path there is no eth or weth leftover in the gateway

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function supplyNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
+    (address reserveAsset, address hub) = _getReserveData(reserveId);
+    _validateParams(reserveAsset, amount);
+    require(msg.value == amount, NativeAmountMismatch());
+
+    NATIVE_WRAPPER.deposit{value: amount}();
+    NATIVE_WRAPPER.safeIncreaseAllowance(hub, amount);
+    SPOKE.supply(reserveId, amount, msg.sender);
```
yan-man: should we reset allowance after supply action?

src/misc/WrappedTokenGatewayV4.sol
yan-man: should we rename to `emergencyTokenTransfer` to be consistent w v3?

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+
+import {IWrappedTokenGatewayV4} from 'src/interfaces/IWrappedTokenGatewayV4.sol';
+import {INativeWrapper} from 'src/interfaces/INativeWrapper.sol';
+import {ISpoke} from 'src/interfaces/ISpoke.sol';
+
+/**
+ * @notice Contract allowing users to approve it as a Position Manager to wrap and unwrap the native asset
+ * before interacting with the Spoke.
+ */
+contract WrappedTokenGatewayV4 is IWrappedTokenGatewayV4, ReentrancyGuardTransient, Ownable2Step {
```
DhairyaSethi: actually realised we should expose setUsingAsCollateral helper also, such that it can be multicalled in one txn

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * @dev Only NATIVE_WRAPPER contract is allowed to do native transfer here. Prevent other addresses to send native assets to this contract.
+   */
+  receive() external payable {
+    require(msg.sender == address(NATIVE_WRAPPER), ReceiveNotAllowed());
+  }
+
+  /**
+   * @dev Revert fallback calls
```
yan-man: nit: period on end

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+   * @notice Recovers native asset left in this contract.
+   * @param to Address to send the recovered native asset to.
+   * @param amount Amount of native asset to recover.
+   **/
+  function recoverNative(address to, uint256 amount) external onlyOwner {
+    Address.sendValue(payable(to), amount);
+  }
+
+  /**
+   * @dev Only NATIVE_WRAPPER contract is allowed to do native transfer here. Prevent other addresses to send native assets to this contract.
```
yan-man:
```suggestion
   * @dev Only NATIVE_WRAPPER contract is allowed to do native transfer here. Prevent other addresses from sending native assets to this contract.
```

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,798 @@
...
+
+contract WrappedTokenGatewayV4Test is Base {
+  WrappedTokenGatewayV4 public wrappedTokenGateway;
+  uint256 public wethReserveId;
+
+  function setUp() public virtual override {
+    super.setUp();
+    initEnvironment();
+
+    wethReserveId = wethReserveId;
```
yan-man: what's this line meant for?

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: idt needed, bc we know supply will either consume 'amount' worth of allowance & pull that many tokens or revert (self note: to add this invariant in doc)

dangling allowances left IS bad (if somehow underlying spoke silently fails or is accidentally of different spec) but not exploitable here bc on each action, native tokens are pulled exactly

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,798 @@
...
+      address(ADMIN)
+    );
+
+    vm.prank(SPOKE_ADMIN);
+    spoke1.updatePositionManager(address(wrappedTokenGateway), true);
+
+    deal(address(tokenList.weth), MAX_SUPPLY_AMOUNT);
+  }
+
+  function _getUserData(address user) internal view returns (DataTypes.UserPosition memory) {
```
yan-man: pls move down below the tests

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+    uint256 deadline,
+    uint8 v,
+    bytes32 r,
+    bytes32 s
+  ) external payable {
+    SPOKE.setUserPositionManagerWithSig(address(this), user, approve, deadline, v, r, s);
+  }
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function renouncePositionManagerRole(address user) external payable {
```
yan-man: hm, can a user ever renouncePositionManagerRole for another user? I think perhaps this should accept no args and below is `SPOKE.renouncePositionManagerRole(msg.sender);` or am i mistaken

DhairyaSethi: yeah we'd kept it open intentionally bc the worst anyone can do is revoke users permission leading to slight dos 

but thinking more about, it's make this function onlyOwner @Kogaroshi 
renounce(msg.sender) is already possible with setUserPosmWithSig and isn't needed

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,798 @@
...
+
+    assertEq(bob.balance, prevUserBalance - amount);
+    assertEq(spoke1.getUserSuppliedAmount(wethReserveId, bob), prevUserSuppliedAmount + amount);
+    assertEq(tokenList.weth.balanceOf(address(hub1)), prevHubBalance + amount);
+    assertEq(address(wrappedTokenGateway).balance, 0);
+  }
+
+  function test_supplyNative_revertsWith_InvalidAmount() public {
+    uint256 amount = 100e18;
+    deal(bob, mintAmount_WETH);
```
yan-man: i think dont need these 2 lines here?

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,798 @@
...
+
+    vm.prank(bob);
+    vm.expectRevert(IWrappedTokenGatewayV4.InvalidReserveId.selector);
+    wrappedTokenGateway.supplyNative{value: amount}(wethReserveId + 1, amount);
+  }
+
+  function test_supplyNative_revertsWith_NativeAmountMismatch() public {
+    uint256 amount = 100e18;
+    deal(bob, mintAmount_WETH);
```
yan-man: i think maybe can deal to bob in the setup to avoid these?

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,798 @@
...
+
+    vm.prank(bob);
+    vm.expectRevert(IWrappedTokenGatewayV4.NativeAmountMismatch.selector);
+    wrappedTokenGateway.supplyNative{value: 0}(wethReserveId, amount);
+
+    vm.prank(bob);
+    vm.expectRevert(IWrappedTokenGatewayV4.NativeAmountMismatch.selector);
```
yan-man: our general convention has been 

expectRevert
prank
action

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,798 @@
...
+    // Bob borrows weth
+    Utils.borrow({
+      spoke: spoke1,
+      reserveId: _wethReserveId(spoke1),
+      caller: bob,
+      amount: borrowAmount,
+      onBehalfOf: bob
+    });
+
+    skip(365 days);
```
yan-man: i think we wanted to skip a more arbitrary time (less round), Dhairya has been using 322 days

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
...
+    NATIVE_WRAPPER.deposit{value: amount}();
+    NATIVE_WRAPPER.safeIncreaseAllowance(hub, amount);
+    SPOKE.supply(reserveId, amount, msg.sender);
+  }
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function withdrawNative(
+    uint256 reserveId,
+    uint256 amount,
+    address receiver
```
yan-man: via a regular posm, receiver on a withdraw is only allowed to be the user for whom the posm is acting on behalf of. Should we block this separate receiver for this gateway case for consistency?

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,798 @@
...
+    vm.prank(bob);
+    vm.expectRevert(IWrappedTokenGatewayV4.InvalidAddress.selector);
+    wrappedTokenGateway.borrowNative(wethReserveId, borrowAmount, address(0));
+  }
+
+  function test_repayNative() public {
+    test_repayNative_fuzz(5e18);
+  }
+
+  function test_repayNative_fuzz(uint256 repayAmount) public {
```
yan-man: this one might be good to skipTime to inflate the drawn index after borrow so that rounding causes some repay donation to check on dust being returned properly

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,798 @@
...
+    (userDrawnDebt, userPremiumDebt) = spoke1.getUserDebt(wethReserveId, bob);
+
+    assertEq(userDrawnDebt + userPremiumDebt, borrowAmount - repayAmount);
+    assertEq(tokenList.weth.balanceOf(address(hub1)), prevHubBalance + repayAmount);
+    assertEq(bob.balance, prevUserBalance - repayAmount);
+    assertEq(address(wrappedTokenGateway).balance, 0);
+
+    assertFalse(spoke1.isPositionManager(bob, address(wrappedTokenGateway)));
+  }
+}
```
DhairyaSethi: multicall revert case missing for coverage (pls check with `make coverage`)

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,188 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {ReentrancyGuardTransient} from 'src/dependencies/openzeppelin/ReentrancyGuardTransient.sol';
+import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
```
yan-man: jw on using ownable2step instead of vanilla ownable, just for extra safety?

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: renounce with setUserPosmWithSig() means making the user sign another payload, or add a branch that the approval is false we ignore the sig and call renounce on the Spoke.
It was left permissionless so it can be bundled at the end of each multicall, but we can otherwise have one that takes in msg.sender and another with `onlyOwner` to renounce for any given user ?

tests/unit/WrappedTokenGatewayV4.t.sol
Kogaroshi: it was lifted from another test, but can change it yeah

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: In the Gateway the `receiver` field is here to keep consistency with V3, do we prefer to keep the Spoke behavior instead ?

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: comes from this comment : https://github.com/aave/aave-v4/pull/690#discussion_r2321529506

tests/unit/WrappedTokenGatewayV4.t.sol
Kogaroshi: leftover I forgot to remove

tests/unit/WrappedTokenGatewayV4.t.sol
Kogaroshi: `deal` is needed so Bob has ETH for the supply call value

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: yeah i recall, we can do both here since other methods also expect caller to be user posm which is being modified

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,194 @@
...
+    bytes32 r,
+    bytes32 s
+  ) external payable {
+    SPOKE.setUserPositionManagerWithSig(address(this), user, approve, deadline, v, r, s);
+  }
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function renouncePositionManagerRole() external payable {
+    SPOKE.renouncePositionManagerRole(msg.sender);
+  }
```
miguelmtzinf: don't think this is needed as it exists in the Spoke contract itself

Kogaroshi: It allows to bundle the action in the Gateway multicall

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,194 @@
...
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function renouncePositionManagerRoleForUser(address user) external onlyOwner {
+    SPOKE.renouncePositionManagerRole(user);
+  }
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function supplyNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
+    (address reserveAsset, address hub) = _getReserveData(reserveId);
```
miguelmtzinf:
```suggestion
    (address underlying, address hub) = _getReserveData(reserveId);
```

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,194 @@
...
+  function recoverNative(address to, uint256 amount) external onlyOwner {
+    Address.sendValue(payable(to), amount);
+  }
+
+  /**
+   * @dev Validates the common parameters for all functions.
+   **/
+  function _validateParams(address reserveAsset, uint256 amount) internal {
+    require(amount > 0, InvalidAmount());
+    require(reserveAsset == address(NATIVE_WRAPPER), InvalidReserveId());
```
miguelmtzinf:
```suggestion
    require(reserveAsset == address(NATIVE_WRAPPER), InvalidReserveId());
    require(amount > 0, InvalidAmount());
```
nit

src/misc/WrappedTokenGatewayV4.sol
miguelmtzinf: so making this change at the end or nah?

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,194 @@
...
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function recoverToken(address token, address to) external onlyOwner {
+    IERC20(token).safeTransfer(to, IERC20(token).balanceOf(address(this)));
+  }
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function recoverNative(address to, uint256 amount) external onlyOwner {
+    Address.sendValue(payable(to), amount);
+  }
```
miguelmtzinf: we been using `recueTokens` in some contracts in the past. There is also this [helper from BGD](https://github.com/bgd-labs/solidity-utils/blob/main/src/contracts/utils/Rescuable.sol) which uses similar terms.
We ve never used the `recover` term for this so I d rename. In general we need to get aligned on these functions so they are the same across all contracts.

Kogaroshi: Yes, we just discussed it with @DhairyaSethi, all `recover` are renamed to `rescue`, and moved to a Rescuable.sol that will be inherited by both this Gateway and the Signature Posm.

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,194 @@
...
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function supplyNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
+    (address reserveAsset, address hub) = _getReserveData(reserveId);
+    _validateParams(reserveAsset, amount);
+    require(msg.value == amount, NativeAmountMismatch());
+
+    NATIVE_WRAPPER.deposit{value: amount}();
+    NATIVE_WRAPPER.safeIncreaseAllowance(hub, amount);
```
miguelmtzinf: why this over `forceApprove`

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,194 @@
...
+    SPOKE.renouncePositionManagerRole(msg.sender);
+  }
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function renouncePositionManagerRoleForUser(address user) external onlyOwner {
+    SPOKE.renouncePositionManagerRole(user);
+  }
+
+  /// @inheritdoc IWrappedTokenGatewayV4
+  function supplyNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
```
miguelmtzinf: we can just rm `Native` suffix imo

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,195 @@
...
+    address receiver
+  ) external payable nonReentrant {
+    (address underlying, ) = _getReserveData(reserveId);
+    _validateParams(underlying, amount);
+    require(receiver != address(0), InvalidAddress());
+
+    uint256 userSuppliedAmount = SPOKE.getUserSuppliedAmount(reserveId, msg.sender);
+    if (amount == type(uint256).max) {
+      amount = userSuppliedAmount;
+    }
```
miguelmtzinf: this is done internally. Alternatively we can balance diff before and after withdraw action. No strong opinion, but i d like to be consistent and do the same across the two posm (this and sigPosm)

src/misc/WrappedTokenGatewayV4.sol
```diff
@@ -0,0 +1,195 @@
...
+    uint256 reserveId,
+    uint256 amount,
+    address receiver
+  ) external payable nonReentrant {
+    (address underlying, ) = _getReserveData(reserveId);
+    _validateParams(underlying, amount);
+    require(receiver != address(0), InvalidAddress());
+
+    uint256 userSuppliedAmount = SPOKE.getUserSuppliedAmount(reserveId, msg.sender);
+    if (amount == type(uint256).max) {
```
miguelmtzinf: we are using `> currentBalance` condition in the Spoke.
Attack vector: an attacker can steal funds of this contract  (if any) by using an input amount greater than `userSuppliedAmount` (e.g. delta > 0): we d withdraw from aave `userSuppliedAmount` only, but withdraw `userSuppliedAmount + delta` from native wrapper

miguelmtzinf: ah no, we are using > for repay, == for withdraw. Worth looking into the attack vector anyway

src/misc/WrappedTokenGatewayV4.sol
miguelmtzinf, line 1: we can probably rename to `NativeTokenGateway`

src/misc/WrappedTokenGatewayV4.sol
miguelmtzinf: same with `setUsingAsCollateral`. I d only keep here functions related to native tokens.
If a user wants to deposit ETH and set as collateral we can multicall anyway no?

tests/unit/WrappedTokenGatewayV4.t.sol
```diff
@@ -0,0 +1,953 @@
...
+
+    assertEq(bob.balance, prevUserBalance - amount);
+    assertEq(
+      spoke1.getUserSuppliedAmount(_wethReserveId(spoke1), bob),
+      prevUserSuppliedAmount + amount
+    );
+    assertEq(tokenList.weth.balanceOf(address(hub1)), prevHubBalance + amount);
+    assertEq(address(wrappedTokenGateway).balance, 0);
+    assertEq(tokenList.weth.balanceOf(address(wrappedTokenGateway)), 0);
+    assertEq(tokenList.weth.allowance(address(wrappedTokenGateway), address(hub1)), 0);
```
miguelmtzinf: worth extracting a helper method

src/misc/Rescuable.sol
miguelmtzinf, line 1: this was copied from BGD's repo right? probably better to move to dependencies

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: Yeah, but we'd have to multicall from the Gateway, since it's the one doing the wrapping, that's why we expose those methods in the Gateway, and we can bundle the ETH deposit with `setUsingAsCollateral` here

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: force of habit, will replace

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: made this way to update `amount` for unwrapping the exact amount, but can replace with the `min()` check like in SigPosm ?

src/misc/Rescuable.sol
Kogaroshi: parts where, but the rescue methods are simpler here.
no strong opinion, we could remove and use BGD's version if we prefer

DhairyaSethi: prefer to keep ours, especially for consistency with sendNative 
we use OZ Address.sendValue, theirs implements the same

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: yes exactly, it's useful to have it here since we don't natively enable in supply action (unlike v3) and spoke supports multicall so this should as well

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: v3 supports withdrawTo and borrowTo natively, we don't for simplicity. but the argument then was to expose it on posm level so i think this makes sense and we should keep

DhairyaSethi: > receiver on a withdraw is only allowed to be the user for whom the posm is acting on behalf of


not really. receiver of funds is always caller - either posm or user

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: those were the names in the Gateway v3, but we can rename to match Spoke function names, no strong preference

src/misc/Rescuable.sol
```diff
@@ -0,0 +1,39 @@
...
+ * @notice This contracts allows to rescue funds sent to the contract by mistake or stuck after transactions.
+ */
+abstract contract Rescuable is IRescuable {
+  using SafeERC20 for *;
+
+  /// @notice modifier that checks that caller is allowed address
+  modifier onlyRescueGuardian() {
+    if (msg.sender != rescueGuardian()) {
+      revert OnlyRescueGuardian();
+    }
```
DhairyaSethi:
```suggestion
    require(msg.sender == rescueGuardian(), OnlyRescueGuardian());
```

DhairyaSethi: using SafeERC20 for *; => using SafeERC20 for IERC20;

DhairyaSethi: actually can you extract that in internal method and use like https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/Ownable.sol#L49-L67
we follow convention in posm checks on spoke also
it's internal so inlined so it's the same but just to follow convention

src/misc/WrappedTokenGatewayV4.sol
DhairyaSethi: #727 is changing this behaviour so we can stick to min approach on both posms for withdraw & repay

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,195 @@
...
+    SPOKE.setUsingAsCollateral(reserveId, usingAsCollateral, msg.sender);
+  }
+
+  /**
+   * @notice Call multiple functions in the current contract and return the data from all of them if they all succeed.
+   * @dev Function was inlined here so it can be payable.
+   * @param data The encoded function data for each of the calls to make to this contract.
+   * @return results The results from each of the calls passed in via data.
+   */
+  function multicall(bytes[] calldata data) external payable returns (bytes[] memory) {
```
DhairyaSethi: if i supply twice in a multicall i can steal any leftover funds in this contract, no?

Kogaroshi: You'd need to 1st supply the exact amount of funds left in the contract, so that `msg.value` equals the leftover balance, then the 2nd supply would be able to wrap and steal the leftover yes.

Kogaroshi: Outdated, multicall was removed, removing the issue.

src/misc/WrappedTokenGatewayV4.sol
Kogaroshi: `renouncePositionManagerRole()` && `setUsingAsCollateral()` removed when removing the multicall.

src/interfaces/INativeTokenGateway.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+interface INativeTokenGateway {
```
miguelmtzinf:
```suggestion

import {IRescuable} from 'src/interfaces/IRescuable.sol';

interface INativeTokenGateway is IRescuable {
```

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,139 @@
...
+    NATIVE_WRAPPER.deposit{value: amount}();
+    NATIVE_WRAPPER.forceApprove(hub, amount);
+    SPOKE.repay(reserveId, amount, msg.sender);
+
+    if (leftovers > 0) {
+      Address.sendValue(payable(msg.sender), leftovers);
+    }
+  }
+
+  /// @inheritdoc Rescuable
```
miguelmtzinf:
```suggestion
  /// @inheritdoc IRescuable
```

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,139 @@
...
+  INativeTokenGateway,
+  ReentrancyGuardTransient,
+  Rescuable,
+  Ownable2Step
+{
+  using SafeERC20 for *;
+
+  /// @notice Native Wrapper contract
+  INativeWrapper public immutable NATIVE_WRAPPER;
+  /// @notice Spoke contract
```
miguelmtzinf:
```suggestion
  /// @inheritdoc INativeTokenGateway
  INativeWrapper public immutable NATIVE_WRAPPER;
  /// @inheritdoc INativeTokenGateway
```
add getters functions to the end of INativeTokenGateway please

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,139 @@
...
+  ISpoke public immutable SPOKE;
+
+  constructor(address nativeWrapper_, address spoke_, address admin_) Ownable(admin_) {
```
miguelmtzinf:
```suggestion
  constructor(address nativeWrapper_, address spoke_, address initialOwner_) Ownable(initialOwner_) {
```

more aligned with the rest of contracts (e.g. SignatureGateway)

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,139 @@
...
+    SPOKE.supply(reserveId, amount, msg.sender);
+  }
+
+  /// @inheritdoc INativeTokenGateway
+  function withdrawNative(uint256 reserveId, uint256 amount, address receiver) external {
+    (address underlying, ) = _getReserveData(reserveId);
+    _validateParams(underlying, amount);
+    require(receiver != address(0), InvalidAddress());
+
+    amount = MathUtils.min(SPOKE.getUserSuppliedAmount(reserveId, msg.sender), amount);
```
miguelmtzinf:
```suggestion
    amount = MathUtils.min(amount, SPOKE.getUserSuppliedAmount(reserveId, msg.sender));
```
it reads better imho

miguelmtzinf: also, updating input param is bad practice... consider creating new state var pls

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,139 @@
...
+  function repayNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
+    (address underlying, address hub) = _getReserveData(reserveId);
+    _validateParams(underlying, amount);
+    require(msg.value == amount, NativeAmountMismatch());
+
+    uint256 userDebtAmount = SPOKE.getUserTotalDebt(reserveId, msg.sender);
+    uint256 leftovers;
+    if (amount > userDebtAmount) {
+      leftovers = amount - userDebtAmount;
+      amount = userDebtAmount;
```
miguelmtzinf: same here, updating input param

src/misc/WrappedTokenGatewayV4.sol
miguelmtzinf: resolved now

DhairyaSethi:
@Kogaroshi could you pls add `Rescuable` import to SignatureGateway as well now that's merged in main

also i forgot to add a comment about how all of those txns can be frontrun since they're caller agnostic to grief and that's ok

src/misc/Rescuable.sol
```diff
@@ -0,0 +1,47 @@
...
+  function rescueGuardian() external view returns (address) {
+    return _rescueGuardian();
+  }
+
+  /// @dev Returns the address that is allowed to rescue funds.
+  function _rescueGuardian() internal view virtual returns (address);
+
+  /**
+   * @dev Throws if the sender is not the rescue guardian.
+   */
```
DhairyaSethi:
```suggestion
   /// @dev Throws if the sender is not the rescue guardian.
```

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,149 @@
...
+  ReentrancyGuardTransient,
+  Rescuable,
+  Ownable2Step
+{
+  using SafeERC20 for *;
+
+  /// @inheritdoc INativeTokenGateway
+  INativeWrapper public immutable NATIVE_WRAPPER;
+  /// @inheritdoc INativeTokenGateway
+  ISpoke public immutable SPOKE;
```
miguelmtzinf: It d help integration if we make these 2 addresses because we d only need to import the interface and not `ISpoke` and `INativeWrapper`.
In turn we need to cast on every fn... feels acceptable to me. thoughts? (cc @DhairyaSethi )

src/interfaces/INativeTokenGateway.sol
```diff
@@ -0,0 +1,59 @@
...
+   * @param reserveId Reserve Id for the wrapped asset.
+   * @param amount Amount to wrap and repay.
+   **/
+  function repayNative(uint256 reserveId, uint256 amount) external payable;
+
+  /**
+   * @notice Renounces the positionManager approval given by an user.
+   * @param user The address of the user.
+   */
+  function renouncePositionManagerRole(address user) external;
```
miguelmtzinf, line 67: did you have plans for a shared interface for position managers where this is compulsory @DhairyaSethi ?

src/misc/NativeTokenGateway.sol
DhairyaSethi: sounds good to me! 

~these can also take ISpokeBase haha, thoughts?~ nvm, position manage stuff is not in ISpokeBase and rightfully so

src/interfaces/INativeTokenGateway.sol
DhairyaSethi: permissioned action, it's onlyOwner, i think it makes sense to expose since it's present on the spoke

src/misc/SignatureGateway.sol
```diff
@@ -141,7 +150,7 @@ contract SignatureGateway is EIP712, Multicall, Ownable2Step, ISignatureGateway
     underlying.safeTransfer(onBehalfOf, amount);
   }
 
-  // @inheritdoc ISignatureGateway
+  /// @inheritdoc ISignatureGateway
   function repayWithSig(
     uint256 reserveId,
     uint256 amount,
```
miguelmtzinf, line 160: we are updating this input param too here

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,149 @@
...
+  function _getReserveData(uint256 reserveId) internal view returns (address, address) {
+    DataTypes.Reserve memory reserveData = SPOKE.getReserve(reserveId);
+    return (reserveData.underlying, address(reserveData.hub));
+  }
+
+  /**
+   * @dev Only NATIVE_WRAPPER contract is allowed to do native transfer here. Prevent other addresses from sending native assets to this contract.
+   */
+  receive() external payable {
+    require(msg.sender == address(NATIVE_WRAPPER), ReceiveNotAllowed());
```
miguelmtzinf: we can use the classic `UnsupportedAction` we ve used in some other places. For `fallback` too

src/interfaces/INativeTokenGateway.sol
Kogaroshi: Should we introduce a BaseGateway that inherits Rescuable, Ownable2Step, and implements common methods (`renouncePositionManagerRole`, `_getReserveData` & `_rescueGuardian` ) ?

DhairyaSethi: it's fine for now i think, we can put these in a separate folder tho (the two posms) but nbd for now let's try to get this merged asap

Kogaroshi: #738 to do those changes later.

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,157 @@
...
+
+  /// @inheritdoc INativeTokenGateway
+  function NATIVE_WRAPPER() external view returns (address) {
+    return address(_NATIVE_WRAPPER);
+  }
+
+  /// @inheritdoc INativeTokenGateway
+  function SPOKE() external view returns (address) {
+    return address(_SPOKE);
+  }
```
miguelmtzinf: move down, at the beggining of read view fns

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,157 @@
...
+contract NativeTokenGateway is
+  INativeTokenGateway,
+  ReentrancyGuardTransient,
+  Rescuable,
+  Ownable2Step
+{
+  using SafeERC20 for *;
+
+  INativeWrapper private immutable _NATIVE_WRAPPER;
+  ISpoke private immutable _SPOKE;
```
miguelmtzinf:
```suggestion
  INativeWrapper internal immutable _nativeWrapper;
  ISpoke internal immutable _spoke;
```

src/interfaces/ISignatureGateway.sol
```diff
@@ -3,9 +3,10 @@
 pragma solidity ^0.8.0;
 
 import {IMulticall} from 'src/interfaces/IMulticall.sol';
-import {ISpoke} from 'src/interfaces/ISpoke.sol';
+import {IRescuable} from 'src/interfaces/IRescuable.sol';
 
-interface ISignatureGateway is IMulticall {
+interface ISignatureGateway is IMulticall, IRescuable {
+  error InvalidAddress();
```
DhairyaSethi, line 12: lets just add natspec for consistency for the error

src/misc/SignatureGateway.sol
```diff
@@ -2,54 +2,62 @@
...
-import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
 import {ISpoke} from 'src/interfaces/ISpoke.sol';
 
-contract SignatureGateway is EIP712, Multicall, Ownable2Step, ISignatureGateway {
+/**
+ * @notice Contract allowing users to approve it as a Position Manager to wrap and unwrap the native asset
+ * @dev All the batched calls in this contract can be frontrun and grief the caller, as they are agnostic, but we consider it
+ * acceptable as this would mean the intent signed by the user would be executed.
+ */
+contract SignatureGateway is ISignatureGateway, Multicall, Rescuable, Ownable2Step, EIP712 {
```
DhairyaSethi: > Contract allowing users to approve it as a Position Manager to wrap and unwrap the native asset

incorrect here, that's for NativeTokenGateway

i would remove from here, and add on interface instead something like
 Gateway to consume simple EIP-712 typed user intents for user actions

and something about caller agnostic, so can be griefied, no paymaster or native refunds or something

Kogaroshi: Changed the Natpsec, but left it on the contract to match what is done in other contract in the repo.

src/misc/NativeTokenGateway.sol
```diff
@@ -0,0 +1,157 @@
...
+contract NativeTokenGateway is
+  INativeTokenGateway,
+  ReentrancyGuardTransient,
+  Rescuable,
+  Ownable2Step
+{
+  using SafeERC20 for *;
+
+  INativeWrapper private immutable _nativeWrapper;
+  ISpoke private immutable _spoke;
```
miguelmtzinf:
```suggestion
  INativeWrapper internal immutable _nativeWrapper;
  ISpoke internal immutable _spoke;
```

src/misc/Rescuable.sol
```diff
@@ -0,0 +1,45 @@
...
+pragma solidity ^0.8.0;
+
+import {SafeERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
+import {Address} from 'src/dependencies/openzeppelin/Address.sol';
+import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
+
+import {IRescuable} from 'src/interfaces/IRescuable.sol';
+
+/**
+ * @notice This contracts allows to rescue funds sent to the contract by mistake or stuck after transactions.
```
yan-man: grammar

src/interfaces/INativeWrapper.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
+
+import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
+
+interface INativeWrapper is IERC20 {
+  function deposit() external payable;
+
+  function withdraw(uint256 wad) external;
```
yan-man: should we change this to be more generic to other native assets, will they always be in WAD

src/interfaces/INativeWrapper.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.20;
```
yan-man: inconsistent w other solidity versions atm

update license too?

DhairyaSethi: this should probably go to deps ngl
anyway, yes pls do

PR #692, opened 2025-09-03
FeeReceiver as separate address instead of Spoke (simple)

yan-man:
Same as #687 but with only 1 fee receiver address per asset.

No memory-protection, meaning that once the fee receiver is updated, all accumulated fees go to the new fee receiver.


yan-man:
decide feeReceiver is spoke, see https://github.com/aave/aave-v4/pull/686

PR #707, opened 2025-09-08
feat: Consolidate Interfaces

### LINKED ISSUE #670: consolidate interfaces: IHubBase, IBasicAssetInterestRateStrategy
- move all methods used by spoke on the hub to `IHubBase`
  - all `preview*` methods
  - replace `getAsset` usage for underlying & decimals on `addReserve`

Spoke should only have access to `IHubBase` interface
Move `setInterestRateData(uint256 assetId, bytes data)` to `IBasicAssetInterestRateStrategy` -> Hub should only interact with `IBasicAssetInterestRateStrategy`

CheyenneAtapour:
Closes #670 #713


src/interfaces/IAssetInterestRateStrategy.sol
```diff
@@ -71,13 +71,6 @@ interface IAssetInterestRateStrategy is IBasicInterestRateStrategy {
    */
   error InterestRateDataNotSet(uint256 assetId);
```
DhairyaSethi: this error can also move in IBasic

src/interfaces/IHubBase.sol
```diff
@@ -75,4 +75,142 @@ interface IHubBase {
...
+   */
+  function getSpokeAddedAmount(uint256 assetId, address spoke) external view returns (uint256);
+
+  /**
+   * @notice Returns the total amount of shares of the specified asset added to the hub by the specified spoke.
+   * @param assetId The identifier of the asset.
+   * @param spoke The address of the spoke.
+   * @return The amount of added shares.
+   */
+  function getSpokeAddedShares(uint256 assetId, address spoke) external view returns (uint256);
```
DhairyaSethi: what's our getter to get spoke debt balance in shares?

CheyenneAtapour: We don't have one. Even on spoke, we don't really deal with shares for debt, and instead deal in amounts because we have base drawn shares, and then the premium amount

src/interfaces/IAssetInterestRateStrategy.sol
CheyenneAtapour: nice, how about the errors used in `setInteresRateData` like `OnlyHub()` `InvalidOptimalUsageRatio` etc. should I move those too?

src/interfaces/IHubBase.sol
DhairyaSethi: yeah for consistency we should expose them 
something getSpokeDrawnShares -> drawnShares, getSpokePremiumShares -> (premiumShares, offset, realised)

src/interfaces/IHubBase.sol
```diff
@@ -75,4 +75,142 @@ interface IHubBase {
...
+  /**
+   * @notice Reports deficit.
+   * @dev Only callable by active spokes.
+   * @param assetId The identifier of the asset.
+   * @param drawnAmount The drawn amount to report as deficit.
+   * @param premiumAmount The premium amount to report as deficit.
+   * @param premiumDelta The premium delta to apply which signal premium deficit.
+   * @return The amount of drawn shares reported as deficit.
+   */
+  function reportDeficit(
```
miguelmtzinf, line 88: do we think deficit is a compulsory function of a hub? we could build Hubs that don't care about deficit.

CheyenneAtapour: I think hubs should care about deficit. Also note that these functions were moved into IHubBase as needed from usage in spoke. For any created hub, it would make sense that it implements all functions the canonical spoke uses imo

src/interfaces/IHubBase.sol
CheyenneAtapour: imo, I see an argument not to add them, because the protocol deals with debt in a particular way, and it can potentially be confusing to the end user of the getters (probably devs / integrators) to expose those, rather than encouraging use of the existing getters for debt.

I'm fine to add though, we can have a tiebreaker from @miguelmtzinf or anyone else

src/interfaces/IHubBase.sol
miguelmtzinf: they can care about deficit, but maybe it's not something they are fixing (e.g. v3 until umbrella came)

DhairyaSethi: > compulsory function of a hub

adding to interface doesn't mean it's compulsory, the hub can revert with UnsupportedAction here (like the treasury spoke does for borrow, repay, liquidate)

src/interfaces/IHubBase.sol
DhairyaSethi: we need consistency in getters, for collateral we expose getters for balance in shares, and assets. we should also expose to get debt in shares (already have for assets)

CheyenneAtapour: Ok, I will add

CheyenneAtapour: Added

src/interfaces/IAssetInterestRateStrategy.sol
DhairyaSethi: definitley not `InvalidOptimalUsageRatio` etc, getInterestRateData is explicitely not a part of the base interface for it to remain extensible, base interface is not conformed to kink based ir strategy

one of stermi's suggestion was for Hub to validate an IR strategy was configured for it by the hub checking irStrategy.hub() == address(this) which would mean base should expose a `hub()` getter (in which case i'm good `OnlyHub()` error here) but i don't have a strong opinion. i dont think we need that check so lmk wdyt

src/contracts/Spoke.sol
```diff
@@ -551,6 +551,20 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     return drawnDebt + premiumDebt;
   }
 
+  /// @inheritdoc ISpoke
```
DhairyaSethi: spokeBase probably if we're keeping these?

src/interfaces/IAssetInterestRateStrategy.sol
CheyenneAtapour: We can keep as is, the less in base, the better imo

src/contracts/Spoke.sol
CheyenneAtapour: we don't have the other getters in spoke base, and again I prefer simplicity

DhairyaSethi: the idea with base is minimal spec that has to be implemented, it makes sense to expect all spokes implement getters for their balances

src/contracts/Hub.sol
```diff
@@ -517,6 +534,20 @@ contract Hub is IHub, AccessManaged {
     return drawn + premium;
   }
 
+  /// @inheritdoc IHubBase
+  function getSpokeDrawnShares(uint256 assetId, address spoke) external view returns (uint256) {
+    return _spokes[assetId][spoke].drawnShares;
+  }
+
+  /// @inheritdoc IHubBase
+  function getSpokePremiumShares(
```
yan-man: just slight nit here, offset and realized premium are not in shares terms, but understand naming consistency w `getSpokeDrawnShares `

CheyenneAtapour: Do you have an idea for a better name?

yan-man: perhaps getSpokePremium or getSpokePremiumData?

CheyenneAtapour: Ok, I like `getSpokePremiumData`. I changed the corresponding for reserve too

src/interfaces/IBasicInterestRateStrategy.sol
```diff
@@ -8,6 +8,19 @@ pragma solidity ^0.8.10;
...
 interface IBasicInterestRateStrategy {
+  /**
+   * @notice Thrown when the interest rate data is not set for the asset.
+   * @param assetId The id of the asset with no interest rate data set.
+   */
+  error InterestRateDataNotSet(uint256 assetId);
+
+  /**
+   * @notice Sets interest rate data for an Aave rate strategy.
+   * @param assetId The id of the asset to update.
```
miguelmtzinf:
```suggestion
   * @param assetId The id of the asset.
```

src/interfaces/IBasicInterestRateStrategy.sol
```diff
@@ -8,6 +8,19 @@ pragma solidity ^0.8.10;
...
  */
 interface IBasicInterestRateStrategy {
+  /**
+   * @notice Thrown when the interest rate data is not set for the asset.
+   * @param assetId The id of the asset with no interest rate data set.
+   */
+  error InterestRateDataNotSet(uint256 assetId);
+
+  /**
+   * @notice Sets interest rate data for an Aave rate strategy.
```
miguelmtzinf:
```suggestion
   * @notice Sets the interest rate parameters for a specified asset.
```

src/interfaces/IBasicInterestRateStrategy.sol
```diff
@@ -8,6 +8,19 @@ pragma solidity ^0.8.10;
...
+   * @param assetId The id of the asset to update.
+   * @param data The interest rate data to apply to the given asset, all in bps, encoded in bytes.
```
miguelmtzinf:
```suggestion
   * @param data The encoded parameters used to configure the interest rate of the asset.
```

miguelmtzinf: not sold on this. I want to be generic because this interface can be used for different strategies, but we must document that current IR strategy is using bps for example. Should we  add these docs to the contract itself?

src/interfaces/ISpoke.sol
```diff
@@ -289,20 +291,31 @@ interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
...
+  function getReserveDebt(uint256 reserveId) external view returns (uint256, uint256);
+
   function getReserveTotalDebt(uint256 reserveId) external view returns (uint256);
 
+  /**
+   * @notice Returns the amount of drawn shares for a given reserve.
+   * @param reserveId The identifier of the reserve.
+   * @return The amount of drawn shares.
+   */
+  function getReserveDrawnShares(uint256 reserveId) external view returns (uint256);
```
miguelmtzinf: drawnShares? or debtShares
why do we need this new fn?

CheyenneAtapour: We use `drawnShares` throughout the spoke code

We don't need this, but we decided that we wanted getters for debt shares

(I was trying to argue against adding these haha :p so I'm down to remove)

src/interfaces/IBasicInterestRateStrategy.sol
CheyenneAtapour: We could do that to specify bps on the actual contract. I did this and the suggested change

src/interfaces/IBasicInterestRateStrategy.sol
CheyenneAtapour: Done

src/interfaces/IBasicInterestRateStrategy.sol
CheyenneAtapour: Done

src/contracts/Hub.sol
yan-man: lol let's see if there's other feedback though first from @DhairyaSethi or @miguelmtzinf

src/interfaces/IBasicInterestRateStrategy.sol
yan-man: minor: we use "identifier" everywhere for id

src/interfaces/ISpoke.sol
CheyenneAtapour: @DhairyaSethi suggests this to be in ISpokeBase: https://github.com/aave/aave-v4/pull/707/files#r2334657482

src/contracts/Hub.sol
CheyenneAtapour: Too late, but yeah agreed :)

src/interfaces/IBasicInterestRateStrategy.sol
CheyenneAtapour: Done

src/contracts/Hub.sol
```diff
@@ -439,50 +445,47 @@ contract Hub is IHub, AccessManaged {
...
     address spoke
   ) external view returns (DataTypes.SpokeConfig memory) {
+    DataTypes.SpokeData storage spoke = _spokes[assetId][spoke];
     return
-      DataTypes.SpokeConfig({
-        active: _spokes[assetId][spoke].active,
-        addCap: _spokes[assetId][spoke].addCap,
-        drawCap: _spokes[assetId][spoke].drawCap
-      });
+      DataTypes.SpokeConfig({active: spoke.active, addCap: spoke.addCap, drawCap: spoke.drawCap});
```
yan-man: this one perhaps doesnt need named args given it's clear anyway

src/contracts/Spoke.sol
```diff
@@ -595,12 +609,13 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
...
     return
       DataTypes.ReserveConfig({
-        paused: _reserves[reserveId].paused,
-        frozen: _reserves[reserveId].frozen,
-        borrowable: _reserves[reserveId].borrowable,
-        collateralRisk: _reserves[reserveId].collateralRisk
+        paused: reserve.paused,
+        frozen: reserve.frozen,
+        borrowable: reserve.borrowable,
+        collateralRisk: reserve.collateralRisk
```
yan-man, line 611: same, maybe named not needed but nbd

src/interfaces/IAssetInterestRateStrategy.sol
```diff
@@ -12,7 +12,7 @@ import {IBasicInterestRateStrategy} from './IBasicInterestRateStrategy.sol';
 interface IAssetInterestRateStrategy is IBasicInterestRateStrategy {
   /**
    * @notice Emitted when new interest rate data is set for an asset.
-   * @param assetId id of the asset that has new interest rate data set.
+   * @param assetId identifier of the asset that has new interest rate data set.
```
yan-man: nit: 
```suggestion
   * @param assetId Identifier of the asset that has new interest rate data set.
```

src/interfaces/IAssetInterestRateStrategy.sol
```diff
@@ -12,7 +12,7 @@ import {IBasicInterestRateStrategy} from './IBasicInterestRateStrategy.sol';
...
    * @param optimalUsageRatio The optimal borrow usage ratio, in bps.
    * @param baseVariableBorrowRate The base variable borrow rate, in bps.
    * @param variableRateSlope1 The slope of the variable interest curve, before hitting the optimal  borrow usage ratio, in bps.
```
yan-man: nit
```suggestion
   * @param variableRateSlope1 The slope of the variable interest curve, before hitting the optimal borrow usage ratio, in bps.
```

src/interfaces/IAssetInterestRateStrategy.sol
```diff
@@ -104,44 +91,44 @@ interface IAssetInterestRateStrategy is IBasicInterestRateStrategy {
...
+   * @param assetId The identifier of the asset to get the variable rate slope 2 for.
    * @return The variable rate slope, in bps.
    */
   function getVariableRateSlope2(uint256 assetId) external view returns (uint256);
 
   /**
    * @notice Returns the maximum variable borrow rate
-   * @param assetId The id of the asset to get the maximum variable borrow rate for
+   * @param assetId The identifier of the asset to get the maximum variable borrow rate for
    * @return The maximum variable borrow rate, in bps
```
yan-man: nit: periods

src/interfaces/IHub.sol
```diff
@@ -334,6 +229,11 @@ interface IHub is IHubBase, IAccessManaged {
    */
   function getAssetDrawnRate(uint256 assetId) external view returns (uint256);
 
+  /**
+   * @notice Returns a struct containing information about the specified asset.
```
yan-man, line 233: just for consistency sake - above is using "object" sometimes, here is struct. we can just choose one

src/interfaces/IHubBase.sol
yan-man: forgot where we landed on this

src/interfaces/IHubBase.sol
```diff
@@ -75,4 +75,163 @@ interface IHubBase {
...
+    uint256 assetId,
+    uint256 drawnAmount,
+    uint256 premiumAmount,
+    DataTypes.PremiumDelta calldata premiumDelta
+  ) external returns (uint256);
+
+  /**
+   * @notice Refreshes premium accounting.
+   * @dev Only callable by active spokes, reverts with `SpokeNotActive` otherwise.
+   * @dev Overall premium should not decrease, reverts with `InvalidPremiumChange` otherwise.
```
yan-man, line 98: we should also decide on convention - should we describe in every natspec the conditions of which cause reverts?

src/interfaces/ISpoke.sol
yan-man: i think convention we settled on:
drawn/premium (before base/premium)
-> owed in hub
-> debt in spoke 

so drawn and premium are both debt in spoke, ie drawn debt, premium debt

src/interfaces/IHubBase.sol
DhairyaSethi: some of these are good to have imo (like here), wouldn't be too strict on this either way tbh

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -36,7 +36,8 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
     HUB = hub_;
   }
 
-  /// @inheritdoc IAssetInterestRateStrategy
+  /// @inheritdoc IBasicInterestRateStrategy
+  /// @dev data contains bps values encoded in bytes
   function setInterestRateData(uint256 assetId, bytes calldata data) external {
```
DhairyaSethi: inheritdoc won't be picked up properly (forge doc, on etherscan etc) would move this comment inside the fn instead

src/interfaces/IHubBase.sol
yan-man: actually reportDeficit i think must be in hub base given that in spoke we're just importing hub base now

src/contracts/Spoke.sol
```diff
@@ -513,20 +512,21 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     return drawnDebt + premiumDebt;
   }
 
-  function getReserveSuppliedAmount(uint256 reserveId) external view returns (uint256) {
+  function getReserveSuppliedAssets(uint256 reserveId) external view returns (uint256) {
```
DhairyaSethi, line 517: should we call the debt getter `getReserve{Owed,Drawn}Assets`

yan-man: agree if there's a getReserveOwedShares too

src/contracts/Spoke.sol
DhairyaSethi: it's useful here DEXEX wise imo to easily know paused and frozen flags were not populated inversly for ex when returning (since we have more than arg of the same type next to each other)

src/contracts/Spoke.sol
CheyenneAtapour: But we used "owed" only on the hub side, and we use `drawn` on spoke side

yan-man: sorry yes @CheyenneAtapour  - spoke side would be: 

getReserveDebtShares
getReserveDebtAssets

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Done

src/interfaces/IHubBase.sol
CheyenneAtapour: Agreed

src/interfaces/IHub.sol
CheyenneAtapour: I chose to use struct everywhere instead of object

src/interfaces/IAssetInterestRateStrategy.sol
CheyenneAtapour: Done

src/interfaces/IAssetInterestRateStrategy.sol
CheyenneAtapour: Done

src/interfaces/IAssetInterestRateStrategy.sol
CheyenneAtapour: Done

src/contracts/Hub.sol
CheyenneAtapour: Done

src/interfaces/IHubBase.sol
```diff
@@ -75,4 +75,163 @@ interface IHubBase {
...
+   */
+  function getSpokeAddedAssets(uint256 assetId, address spoke) external view returns (uint256);
+
+  /**
+   * @notice Returns the total amount of shares of the specified asset added to the hub by the specified spoke.
+   * @param assetId The identifier of the asset.
+   * @param spoke The address of the spoke.
+   * @return The amount of added shares.
+   */
+  function getSpokeAddedShares(uint256 assetId, address spoke) external view returns (uint256);
```
DhairyaSethi, line 289: i think getTotalAddedAssets & getTotalAddedShares should also be in HubBase

src/contracts/Spoke.sol
CheyenneAtapour: Ok, but drawn is nice because it distinguishes from base debt, and also I thought we should do `getReservePremiumData` because "shares" doesn't make sense when returning 3 values (premiumShares, offset, realizedPremium)

src/interfaces/IHubBase.sol
CheyenneAtapour: Done

src/interfaces/IHub.sol
```diff
@@ -346,6 +246,22 @@ interface IHub is IHubBase, IAccessManaged {
...
+   * @param assetId The identifier of the asset.
+   * @return The premium shares of the asset.
+   * @return The premium offset of the asset.
+   * @return The realized premium of the asset.
+   */
+  function getAssetPremiumData(uint256 assetId) external view returns (uint256, uint256, uint256);
+
   function getTotalAddedAssets(uint256 assetId) external view returns (uint256);
 
   function getTotalAddedShares(uint256 assetId) external view returns (uint256);
```
```suggestion
```
these two need to be removed now that they're moved

src/interfaces/IHub.sol
DhairyaSethi, line 1: assets debt getters - assetDrawnShares & assetPremiumData is missing - in HubBase
also the newly added fns need to be tested to retain coverage 
image

src/contracts/Spoke.sol
avniculae: I agree with this, are we not doing it?

DhairyaSethi: we should imo

src/contracts/AssetInterestRateStrategy.sol
```diff
@@ -36,9 +36,10 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
     HUB = hub_;
   }
 
-  /// @inheritdoc IAssetInterestRateStrategy
+  /// @inheritdoc IBasicInterestRateStrategy
   function setInterestRateData(uint256 assetId, bytes calldata data) external {
     require(HUB == msg.sender, OnlyHub());
+    // data contains bps values encoded in bytes
```
miguelmtzinf: don't like that this is important and kinda hidden now... Maybe we can write our own natspec docs here? or in `IAssetInterestRateStrategy` by adding the functions again so we can provide more-detailed docs there (and inherit here)?

src/contracts/Hub.sol
```diff
@@ -411,6 +411,12 @@ contract Hub is IHub, AccessManaged {
     return _assets[assetId];
   }
 
+  /// @inheritdoc IHubBase
+  function getAssetUnderlyingAndDecimals(uint256 assetId) external view returns (address, uint8) {
```
miguelmtzinf, line 417: do not think we need this... what's the reason for adding it since we have `getAsset` getter already?

DhairyaSethi: bc spoke needs these two values on addReserve 
and wanted current spoke to use only IHubBase so far 
and we didn't want to add getAsset to IHubBase, since that's overly restrictive

src/interfaces/IHubBase.sol
miguelmtzinf: not an issue, the spoke can call `IHub(hub).reportDeficit` even if `reportDeficit` function is in `IHubBase` because `IHub` inherits `IHubBase`.

miguelmtzinf: I d just keep on `IHubBase` only what is strictly necessary for a `Hub`, but do not have a strong opinion.
This is smt that can change overtime depending on integrations and future hubs/spokes.

src/interfaces/IHubBase.sol
miguelmtzinf: i d suggest to document only those that are v relevant

src/contracts/Spoke.sol
```diff
@@ -19,7 +19,7 @@ import {LiquidationLogic} from 'src/libraries/logic/LiquidationLogic.sol';
 import {PositionStatus} from 'src/libraries/configuration/PositionStatus.sol';
 import {MathUtils} from 'src/libraries/math/MathUtils.sol';
 
-import {IHub} from 'src/interfaces/IHub.sol';
+import {IHubBase} from 'src/interfaces/IHubBase.sol';
```
miguelmtzinf, line 22: I d import and call `IHub` instead, as this `Spoke` contract is designed to call `Hub`.
At the end of the day, `IHub` defines an standard that all hubs must implement and `ISpoke` the same for all spokes. We created a Hub and a Spoke, that implement their corresponding standards. But this does not mean that our Spoke needs to use the standard `IHubBase`, it's completely fine and natural to use `IHub`.

src/contracts/Spoke.sol
```diff
@@ -108,16 +108,16 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     uint256 reserveId = _reserveCount++;
     uint16 dynamicConfigKey; // 0 as first key to use
 
-    DataTypes.Asset memory asset = IHub(hub).getAsset(assetId);
-    require(asset.underlying != address(0), AssetNotListed());
+    (address underlying, uint8 decimals) = IHubBase(hub).getAssetUnderlyingAndDecimals(assetId);
```
miguelmtzinf, line 111: gas cost increase of using `getAsset` is fine in this authorized function. Was `getAssetUnderlyingAndDecimals` added for any other reason?

src/contracts/Spoke.sol
miguelmtzinf: yep, I thought we were adding an `ISpokeBase` too @CheyenneAtapour

src/interfaces/ISpoke.sol
miguelmtzinf: I d use `getReserveDebtShares` and `getReservePremiumData` for internal data

miguelmtzinf: I saw we are using `getReserveDebt` for returning debt and premium and `getReserveTotalDebt` for returning the sum.
I d suggest:
- `getReserveTotalDebt` for sum
- `getReserveDebt` for base debt and premium
- `getReserveDebtAssets` for base debt
- `getReservePremiumAssets` for premium debt

another alternative is:
- `getReserveTotalDebt` for sum
- `getReserveDebt` for base debt
- `getReservePremium` for premium debt
- don't have any suggestion for returning base debt and premium :(

src/contracts/Spoke.sol
DhairyaSethi: yes https://github.com/aave/aave-v4/pull/707#discussion_r2342382493

src/contracts/Spoke.sol
yan-man: haha good point on Drawn for non premium debt shares (whereas debt is like drawn + premium)

src/contracts/Hub.sol
miguelmtzinf: `getSpokePremiumData` sounds good to me. Let's use `getSpokePremium` for the premium debt

src/interfaces/ISpoke.sol
miguelmtzinf: i d for the latter, using `getReserveDebtAndPremium` for returning both

src/contracts/Hub.sol
miguelmtzinf: last one is good pt but imho is completely fine to use `IHub` instead of `IHubBase`

src/interfaces/ISpoke.sol
miguelmtzinf: agreed to use `getReserveDebt` for returning both and do not return them individually to avoid confusion.

src/contracts/AssetInterestRateStrategy.sol
CheyenneAtapour: Yes, added our own natspec to the contract directly now

src/contracts/Spoke.sol
CheyenneAtapour: We decided for hub base

src/contracts/TreasurySpoke.sol
```diff
@@ -76,4 +76,44 @@ contract TreasurySpoke is ITreasurySpoke, Ownable {
   function liquidationCall(uint256, uint256, address, uint256) external pure {
     revert UnsupportedAction();
   }
+
+  /// @inheritdoc ISpokeBase
+  function getUserDebt(uint256 reserveId, address user) external pure returns (uint256, uint256) {
+    return (0, 0);
+  }
```
DhairyaSethi, line 83:
```suggestion
  function getUserDebt(uint256, address) external pure returns (uint256, uint256) {}
```
wouldn't add explicit zeros but just do like this 
and pls rm arg names where unused

src/contracts/TreasurySpoke.sol
```diff
@@ -76,4 +76,44 @@ contract TreasurySpoke is ITreasurySpoke, Ownable {
...
+    return 0;
+  }
+
+  /// @inheritdoc ISpokeBase
+  function getUserSuppliedShares(uint256 reserveId, address user) external pure returns (uint256) {
+    return 0;
+  }
+
+  /// @inheritdoc ISpokeBase
+  function getReserveDebt(uint256) external pure returns (uint256, uint256) {
```
DhairyaSethi, line 111: inconsistent w arg names here

PR #712, opened 2025-09-08
feat: signature gateway

### LINKED ISSUE #708: Credit Delegation PositionManager
Create a position manager to enable credit delegation. Two variants:

- V1: Simple wrapper for `withSig` variants for all user facing spoke methods which consumes a signature for each user facing method for the spoke. 
  - One time approval for each method, caller agnostic. Provides very simple gas sponsorship which can expanded to a 4337 paymaster (such that gas refund can be done onchain, here it's taken care of offchain).
  - Motivation: `"As a user i want to sign some typed data with my PK on the client > send it to a backend (any) who can send it on my behalf and they should have no other access enabling a secure gasless experience"`

- V2: aave v3 style granular approval amount credit delegation.

DhairyaSethi:
closes #708

### DIFF
--- a/src/misc/SignatureGateway.sol
+++ b/src/misc/SignatureGateway.sol
@@ -166,7 +166,7 @@ contract SignatureGateway is EIP712, Multicall, Ownable2Step, ISignatureGateway 
     require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
 
     (IERC20 underlying, address hub) = _getReserveData(reserveId);
-    amount = MathUtils.min(amounts, SPOKE.getUserTotalDebt(reserveId, onBehalfOf));
+    amount = MathUtils.min(amount, SPOKE.getUserTotalDebt(reserveId, onBehalfOf));
 
     underlying.safeTransferFrom(onBehalfOf, address(this), amount);
     underlying.forceApprove(hub, amount);


src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  // @inheritdoc ISignatureGateway
+  function setSelfAsUserPositionManagerWithSig(
+    address user,
+    bool approve,
+    uint256 deadline,
+    bytes calldata signature
+  ) external {
+    SPOKE.setUserPositionManagerWithSig(address(this), user, approve, deadline, signature);
```
DhairyaSethi, line 257: thinking about making this try catch since it can be frontrun to DOS multicall batches (similar rationale to permit for having it)

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,263 @@
...
+        deadline: deadline,
+        v: v,
+        r: r,
+        s: s
+      })
+    {} catch {}
+  }
+
+  // @inheritdoc ISignatureGateway
+  function renounceSelfAsUserPositionManager(address user) external onlyOwner {
```
miguelmtzinf, line 285: this is the only authorized fn in this contract, and i d like to get rid of it so these contracts are permissionless.
We can achieve same result by setting the positionManager `active` flag to `false` on the Spoke right?

do you have any edge case in mind where this is needed?

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,263 @@
...
+    address user,
+    bool approve,
+    uint256 deadline,
+    bytes calldata signature
+  ) external {
+    SPOKE.setUserPositionManagerWithSig(address(this), user, approve, deadline, signature);
+  }
+
+  // @inheritdoc ISignatureGateway
+  function permitReserve(
```
miguelmtzinf, line 261: you prefer to have this in place instead of relying on the underlying permit fn of the asset because of integration ease?
apart from that, not sold on the name since it's the `underlying` not the `reserve` itself

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,263 @@
...
+          deadline
+        )
+      )
+    );
+    require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
+
+    SPOKE.withdraw(reserveId, amount, onBehalfOf);
+
+    (IERC20 asset, ) = _getReserveData(reserveId);
+    asset.safeTransfer(onBehalfOf, amount);
```
miguelmtzinf: if amount is max, user will max withdraw and also steal any funds from the contract.

miguelmtzinf: this is one of the main edge cases where input amount may not be equal to the withdrawn assets, could there be more cases like this? In overall, feels safer to balance diff

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,263 @@
...
+        )
+      )
+    );
+    require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
+
+    (IERC20 asset, address hub) = _getReserveData(reserveId);
+    asset.safeTransferFrom(onBehalfOf, address(this), amount);
+    asset.forceApprove(hub, amount);
+
+    SPOKE.repay(reserveId, amount, onBehalfOf);
```
miguelmtzinf, line 174: we may want to block max repay to prevent front-running attack vector?

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,263 @@
...
+          _useNonce(onBehalfOf),
+          deadline
+        )
+      )
+    );
+    require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
+
+    SPOKE.borrow(reserveId, amount, onBehalfOf);
+
+    (IERC20 asset, ) = _getReserveData(reserveId);
```
miguelmtzinf: feels better to do this first for existence check, although i d revert on borrow too.

src/misc/SignatureGateway.sol
miguelmtzinf: just realized we have the same on the Spoke (same doubt there re name).
Do we wanna rely on that one instead?

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,263 @@
...
+          onBehalfOf,
+          _useNonce(onBehalfOf),
+          deadline
+        )
+      )
+    );
+    require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
+
+    SPOKE.setUsingAsCollateral(reserveId, useAsCollateral, onBehalfOf);
+  }
```
miguelmtzinf, line 202: missing functions for user rp and dyn config updates. is it in purpose? The gas sponsor can choose which actions to support on the backend so I think it may make sense to add them too (thinking that user rp / dyn config updates can be expensive so not always supported)

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,263 @@
...
+    SPOKE.renouncePositionManagerRole(user);
+  }
+
+  // @inheritdoc ISignatureGateway
+  function DOMAIN_SEPARATOR() external view returns (bytes32) {
+    return _domainSeparator();
+  }
+
+  // @inheritdoc ISignatureGateway
+  function useNonce() external {
```
miguelmtzinf: move it up pls

src/misc/SignatureGateway.sol
Kogaroshi: Otherwise, if max repay is allowed, need to add a sweep for unused assets in the contract back to the user.

src/misc/SignatureGateway.sol
miguelmtzinf: it can happen with any action here right? attacker can frontrun `repayWithSig` action of the batch and make the whole batch revert. Is this one different?

src/misc/SignatureGateway.sol
DhairyaSethi: for having ownable
contract is pretty permissionless regardless

for rescueTokens, debated adding that bc a property of this contract is that it should never be left with any leftover tokens or allowances 

we have this on wrappedtokengateway (v3 & our impl) 

for renounce 
have a plan for all positions managers to support this method for sanity

src/misc/SignatureGateway.sol
DhairyaSethi: it's adding to be able to multicall everything all actions from one place 
bc token approval is also needed for the first time

DhairyaSethi: it's always referenced from a reserve id hence called permitReserve

src/misc/SignatureGateway.sol
DhairyaSethi: yes exactly, only diff with this is only sig not verified here (like permit)
all other fns which can be front run are verified here

src/misc/SignatureGateway.sol
DhairyaSethi: damn yes ty. was cooked finishing this last night

src/misc/SignatureGateway.sol
DhairyaSethi: exceptionally left them bc those can also be called by specific roles given by governance and the sig type hashes were getting out of hand lol 

but in terms of true sponsorship, should add probably. wdyt

src/misc/SignatureGateway.sol
DhairyaSethi: > this is one of the main edge cases where input amount may not be equal to the withdrawn assets, could there be more cases like this? In overall, feels safer to balance diff


lil overkill since these are known underlying contracts but i see your point. i was thinking of writing to invariants doc that 
supply, borrow -> should consume the passed amount arg worth of allowance & transfer exactly that amount of underlying (so no rounding due to shares on this value, actual transfer expected) 

withdraw, repay cannot be exqct

miguelmtzinf: generally safer to balance diff always, preventing any attack vector related to stealing existing funds of the contract (it should not hold any)

src/misc/SignatureGateway.sol
miguelmtzinf: good point, there could be dust left behind also. let's block

src/misc/SignatureGateway.sol
miguelmtzinf: i d add it, since it's service that can be offered as well

src/misc/SignatureGateway.sol
miguelmtzinf: wondering if all these sig fn should be wrapped with try-catch because of this (modified version of multicall maybe?)

src/misc/SignatureGateway.sol
miguelmtzinf: i'm kinda okay with the name. Let's see if anyone else complains lol

src/misc/SignatureGateway.sol
miguelmtzinf: i d add rescueTokens even if this contract is not meant to hold funds.

`have a plan for all positions managers to support this method for sanity` -  wdym for sanity?

not opposed to have this fn in indeed... thinking that this contract can be deployed by an entity that wants to offer gasless txs, so makes sense for them to control this fn and `rescueTokens`

tests/unit/Hub/HubBase.t.sol
```diff
@@ -205,9 +205,8 @@ contract HubBase is Base {
     address user,
     uint256 reserveId,
     uint256 premiumRestored
-  ) internal returns (DataTypes.PremiumDelta memory) {
+  ) internal view override returns (DataTypes.PremiumDelta memory) {
```
miguelmtzinf, line 208: not following the diff between this and base one, and why the need to override... sorry but feels over-complicated

tests/unit/misc/SignatureGateway/SignatureGateway.Base.t.sol
```diff
@@ -0,0 +1,198 @@
...
+    uint256 deadline = vm.randomUint(0, MAX_SKIP_TIME - 1);
+    vm.warp(vm.randomUint(deadline + 1, MAX_SKIP_TIME));
+    return deadline;
+  }
+
+  /**
+   * @dev Warps to a random time before a randomly generated deadline.
+   * @return The randomly generated deadline.
+   */
+  function _warpUntilRandomDeadline() internal returns (uint256) {
```
miguelmtzinf:
```suggestion
  function _warpBeforeRandomDeadline() internal returns (uint256) {
```
because until is ==?

tests/unit/misc/SignatureGateway/SignatureGateway.Base.t.sol
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  function _typedDataHash(
+    ISignatureGateway _gateway,
+    bytes32 typeHash
+  ) internal view returns (bytes32) {
+    return keccak256(abi.encodePacked('\x19\x01', _gateway.DOMAIN_SEPARATOR(), typeHash));
+  }
+
+  function _assertGatewayHasNoBalanceOrAllowance(
```
miguelmtzinf, line 231: is it worth to create another one to ensure gateway has no active position in the spoke (supply/debt == 0)?

tests/unit/Hub/HubBase.t.sol
DhairyaSethi: was introduced one of @CheyenneAtapour PRs bc the diff in usage of how premium is restored on Spoke & Hub
Spoke enforces debt restore from premium first, but no such restriction on the hub (since hub only knows about spoke data, and not when user base debt is up) 
we can probably rename for clarity but i recall we added a comment on Cheyenne?

tests/unit/misc/SignatureGateway/SignatureGateway.Reverts.InvalidReserveId.t.sol
```diff
@@ -0,0 +1,44 @@
...
+    EIP712Types.Repay memory p = _repayData(spoke1, alice, _warpUntilRandomDeadline());
+    p.reserveId = _randomInvalidReserveId(spoke1);
+    bytes memory signature = _sign(alicePk, _getTypedDataHash(gateway, p));
+
+    vm.expectRevert(ISignatureGateway.InvalidReserveId.selector);
+    vm.prank(vm.randomAddress());
+    gateway.repayWithSig(p.reserveId, p.amount, alice, p.deadline, signature);
+  }
+
+  // withdraw, borrow revert earlier at Spoke
```
miguelmtzinf: it may be better devx/ux to follow same approach for all

src/misc/SignatureGateway.sol
miguelmtzinf: cross-posting: https://github.com/aave/aave-v4/pull/690/files#r2340179937

tests/unit/misc/SignatureGateway/SignatureGateway.Base.t.sol
DhairyaSethi: [3c01392](https://github.com/aave/aave-v4/pull/712/commits/3c013926d1745c103000b3ab477dc09f16c92f80)

src/misc/SignatureGateway.sol
DhairyaSethi: [1f00551](https://github.com/aave/aave-v4/pull/712/commits/1f005513fb7fd711201d3433dfea17586bdb35ac) so it's compatible when we revert to consistent behaviour

src/misc/SignatureGateway.sol
DhairyaSethi: will added once #689 is merged which introduces Rescuable dep

src/misc/SignatureGateway.sol
DhairyaSethi: no i don't think we wrap other fns (which are verified here, ie their nonce is incremented here) in try catch bc it becomes complex in the revert case to not increment nonce and miss edge cases, open to feedback

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,319 @@
...
+          onBehalfOf,
+          _useNonce(onBehalfOf),
+          deadline
+        )
+      )
+    );
+    require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
+
+    (IERC20 underlying, ) = _getReserveData(reserveId);
+    amount = MathUtils.min(SPOKE.getUserSuppliedAmount(reserveId, onBehalfOf), amount);
```
miguelmtzinf:
```suggestion
    amount = MathUtils.min(amount, SPOKE.getUserSuppliedAmount(reserveId, onBehalfOf));
```
feels more readable in the opposite order imho

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,319 @@
...
+          onBehalfOf,
+          _useNonce(onBehalfOf),
+          deadline
+        )
+      )
+    );
+    require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
+
+    (IERC20 underlying, address hub) = _getReserveData(reserveId);
+    amount = MathUtils.min(SPOKE.getUserTotalDebt(reserveId, onBehalfOf), amount);
```
miguelmtzinf:
```suggestion
    amount = MathUtils.min(amounts, SPOKE.getUserTotalDebt(reserveId, onBehalfOf));
```

src/interfaces/ISignatureGateway.sol
```diff
@@ -0,0 +1,219 @@
...
+    uint256 reserveId,
+    uint256 amount,
+    address onBehalfOf,
+    uint256 deadline,
+    bytes calldata signature
+  ) external;
+
+  /**
+   * @notice Facilitates withdraw action on connected SPOKE() with a typed signature from `onBehalfOf`.
+   * @dev Withdrawn assets are pushed to `onBehalfOf`.
```
miguelmtzinf:
```suggestion
   * @notice Facilitates withdraw action on connected SPOKE() with a typed signature from `onBehalfOf`.
   * @dev Providing an amount exceeding the user's current withdrawable balance indicates a request for a maximum withdrawal.
   * @dev Withdrawn assets are pushed to `onBehalfOf`.
```

src/interfaces/ISignatureGateway.sol
```diff
@@ -0,0 +1,219 @@
...
+    uint256 reserveId,
+    uint256 amount,
+    address onBehalfOf,
+    uint256 deadline,
+    bytes calldata signature
+  ) external;
+
+  /**
+   * @notice Facilitates repay action on connected SPOKE() with a typed signature from `onBehalfOf`.
+   * @dev Repay assets are pulled from `onBehalfOf`, prior approval to this gateway is required.
```
miguelmtzinf, line 73:
```suggestion
   * @dev Repay assets are pulled from `onBehalfOf`, prior approval to this gateway is required.
   * @dev Providing an amount greater than the user's current debt indicates a request to repay the maximum possible amount.
```

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,319 @@
...
+          onBehalfOf,
+          _useNonce(onBehalfOf),
+          deadline
+        )
+      )
+    );
+    require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
+
+    (IERC20 underlying, address hub) = _getReserveData(reserveId);
+    amount = MathUtils.min(amounts, SPOKE.getUserTotalDebt(reserveId, onBehalfOf));
```
DhairyaSethi:
```suggestion
    amount = MathUtils.min(amount, SPOKE.getUserTotalDebt(reserveId, onBehalfOf));
```

src/misc/SignatureGateway.sol
```diff
@@ -0,0 +1,319 @@
...
+  bytes32 public constant UPDATE_USER_RISK_PREMIUM_TYPEHASH =
+    0xb41e132023782c9b02febf1b9b7fe98c4a73f57ebc63ba44cd71f6365ea09eaf; // keccak256('UpdateUserRiskPremium(address spoke,address user,uint256 nonce,uint256 deadline)')
+  // @inheritdoc ISignatureGateway
+  bytes32 public constant UPDATE_USER_DYNAMIC_CONFIG_TYPEHASH =
+    0xba177b1f5b5e1e709f62c19f03c97988c57752ba561de58f383ebee4e8d0a71c; // keccak256('UpdateUserDynamicConfig(address spoke,address user,uint256 nonce,uint256 deadline)')
+
+  mapping(address user => uint256 nonce) internal _nonces;
+
+  constructor(address spoke_, address initialOwner_) Ownable(initialOwner_) {
+    assert(spoke_ != address(0) && initialOwner_ != address(0));
```
yan-man, line 48: a bit inconsistent with our constructors elsewhere with a require and error

DhairyaSethi: yeah didn't want to add an error which is only used once and won't be in abi in the interface

tests/unit/Spoke/Spoke.SetUserPositionManagerWithSig.t.sol
```diff
@@ -402,11 +402,10 @@ contract SpokeSetUserPositionManagerWithSigTest is SpokeBase {
     address user = vm.randomAddress();
 
     uint256 currentNonce = spoke1.nonces(user);
-    vm.assume(currentNonce != UINT256_MAX);
 
     vm.prank(user);
     spoke1.useNonce();
 
-    assertEq(spoke1.nonces(user), currentNonce + 1);
+    assertEq(spoke1.nonces(user), MathUtils.uncheckedAdd(currentNonce, 1));
```
yan-man, line 409: just to clarify, this isnt actually needed here right

DhairyaSethi: it is. bc i removed the assume that current nonce is not uint.max

src/misc/SignatureGateway.sol
miguelmtzinf: ah, missed this. Better to use require imo

DhairyaSethi: @Kogaroshi could you pls fix in #690

Issue #621, opened 2025-08-12
updateBorrowRate is omitted on hub.refreshPremium


DhairyaSethi:
Out of all actions, we currently do not `updateBorrowRate` on `Hub.refreshPremium`.

Rationale for: because of instant 2 wei premium creation, and general invocation due to spoke.updateUserRiskPremium w/o any supporting action which refreshes the borrow rate, borrow rate is always more exact.

Rationale against: borrow rate is inexact by only 2 wei on debt utilisation which is acceptable imprecision for the gas savings.

PR #716, opened 2025-09-09
feat: Update drawn rate in refresh premium

### LINKED ISSUE #695: Rate not updated on Hub::refreshPremium
[StErMi](https://github.com/StErMi):

[⚠️ Issue] This function does update the premium attributes (`premiumShares`, `premiumOffset` and `realizedPremium`) for both the asset and the spoke but does not update the asset's rate that could be influenced by that value.

The current IRS implementation does not (as far as AAVE said) use the premium value passed to `irs.calculateInterestRate(...)` but other IRS implementations could rely on that, and they would fail to update the borrowing rate properly.

This means that the next calculation of the rates would use an incorrect borrowing rate

CheyenneAtapour:
closes #695


miguelmtzinf:
Closes #621

PR #718, opened 2025-09-10
feat: snapshot repo pipeline

miguelmtzinf:
Re-issuing #711

### DIFF
--- a/.github/workflows/mirror.yml
+++ b/.github/workflows/mirror.yml
@@ -1,7 +1,7 @@
 name: Snapshot to private repo (cross-org)
 on:
   push:
-    branches: [testing-pipeline]
+    branches: [main]
 
 permissions:
   contents: read


.github/workflows/mirror.yml
```diff
@@ -0,0 +1,51 @@
...
+permissions:
+  contents: read
+
+jobs:
+  snapshot:
+    runs-on: aave-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
DhairyaSethi:
```suggestion
      - uses: actions/checkout@v4
        with:
          submodules: recursive
          token: ${{ secrets.READ_ONLY_PAT || github.token }}
```

absis: thanks for the tip @DhairyaSethi. I'll take a look to use github app instead of pat.

absis:
@miguelmtzinf @DhairyaSethi I fixed the workflow to get the submodules on checkout

PR #719, opened 2025-09-10
feat: Max getters for Hub

### LINKED ISSUE #699: Evaluate adding more 4626-standard functions to Hub for integration ease
We have `preview*` functions but we do not have `max*` functions.

CheyenneAtapour:
closes #699


src/contracts/Hub.sol
```diff
@@ -649,6 +694,14 @@ contract Hub is IHub, AccessManaged {
     );
   }
 
+  function _getSpokeTotalOwed(
+    DataTypes.SpokeData storage spoke,
+    uint256 assetId
```
miguelmtzinf, line 699: swap order

PR #672, opened 2025-08-25
Liquidation Refactoring

avniculae:
closes https://github.com/aave/aave-v4/issues/675
- refactors the liquidation logic library while also changing logic slightly (eg rounding directions)
- refactors the liquidation tests & test setup
- removes local var structs from Spoke
- renames liquidationBonus to maxLiquidationBonus, calculateVariableLiquidationBonus to calculateLiquidationBonus

PR #720, opened 2025-09-10
DataTypes removal so contracts are self-contained

### LINKED ISSUE #568: Remove DataTypes and move data structs to corresponding interfaces

miguelmtzinf:
Based on #672 
Blocked by #707 -> will need conflicts resolving

Partially closes #568

#### `PositionStatus` library was renamed to `PositionStatusMap`.
Reason being there is `PositionStatus` lib and struct currently in Spoke contract. We can either:
1. rename lib -> alternatives: `PositionStatusConfiguration` / `PositionConfiguration`)
2. rename struct -> alternatives:  `UserMap` / `PositionMap` / `UserPositionMap` / `UserStateMap`.


miguelmtzinf:
Closing in favor of #748

PR #721, opened 2025-09-10
Constants removal so contracts are self-contained

### LINKED ISSUE #568: Remove DataTypes and move data structs to corresponding interfaces

miguelmtzinf:
Based on #720 
Blocked by #707 -> will need conflicts resolving

Closes #568 

#### Only using Constants lib file on tests.
Otherwise we need to do smt like `IHub(hubInstance).CONSTANT()` every time and makes tests code messier.

#### Spoke constants are referring LiquidationLogic constants
Took the decision to move all libs' constants to main contract (`Spoke`) and refer them from there. Main reason is that one of the constants is used in both sides (Spoke and LiquidationLogic lib). Open to suggestions.

### DIFF
--- a/src/libraries/logic/LiquidationLogic.sol
+++ b/src/libraries/logic/LiquidationLogic.sol
@@ -6,7 +6,6 @@ import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
 import {IHub, IHubBase} from 'src/interfaces/IHub.sol';
 import {ISpoke, ISpokeBase} from 'src/interfaces/ISpoke.sol';
 import {IAaveOracle} from 'src/interfaces/IAaveOracle.sol';
-import {Constants} from 'src/libraries/helpers/Constants.sol';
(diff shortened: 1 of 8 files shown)


src/libraries/logic/LiquidationLogic.sol
```diff
@@ -103,10 +102,10 @@ library LiquidationLogic {
     uint256 borrowedAssetsCount;
   }
 
-  /**
-   * @dev This constant represents the minimum amount of assets in base currency that need to be leftover after a liquidation, if not clearing collateral on a position completely.
-   * @notice The default value assumes that the basePrice is usd denominated by 26 decimals.
-   */
+  /// @dev see ISpoke
```
miguelmtzinf, line 105: open to suggestions here: how to add proper docs to these two

yan-man: i am assuming you dont want to duplicate

miguelmtzinf:
Closing in favor of #748

PR #722, opened 2025-09-10
feat: Remove permissionless update rp logic 

### LINKED ISSUE #702: Spoke::notify returns whether the user premium has increased by using premiumShares and not real user rp
### Description

The `Spoke::notifyRiskPremiumUpdate` function returns whether there were an increase of user premium by using `premiumShares` instead of the real user rp (`= premiumShares/drawnShares`). This can be problematic because they are not completely equivalent: `premiumShares` are influenced by `drawnShares`.

Note that user rp (`= premiumShares/drawnShares`) must be the same for all reserves that are being borrowed by the user, but `premiumShares` are not the same.

https://github.com/aave/aave-v4/blob/875fc8c8da6c5f0f3860db7ef5d14c6099787a1f/src/contracts/Spoke.sol#L995-L1001

This flag is only used by `updateUserRiskPremium` to check if the action is callable by anyone (e.g. anyone can call it if user rp decreases). This is fine because the amount of debt is static for this action, but it d be wrong if this were to be used by any other function that updates the drawn amount.

**Note that current implementation is correct, but this worsen maintainability and can pose risks for future changes.**

### Potential solutions

1. Remove the conditional logic on this feature: this update could only be called by the user or DAO (no permissionless calls at all). This would make it more aligned with the `updateDynamicConfig` which it's only callable by user or DAO.

https://github.com/aave/aave-v4/blob/875fc8c8da6c5f0f3860db7ef5d14c6099787a1f/src/contracts/Spoke.sol#L386-L403

2. Calculate the user risk premium beforehand and compare against the new one. We d need to find the first reserve borrowed and calculate the current user rp by dividing `premiumShares` by `drawnShares`. This adds a bit more gas to this function, but removes a bit of logic on `notifyRiskPremiumUpdate` for all cases.

DhairyaSethi:
closes #702

### DIFF
--- a/src/interfaces/ISpoke.sol
+++ b/src/interfaces/ISpoke.sol
@@ -193,14 +193,14 @@ interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
   ) external;
 
   /**
-   * @notice Allows updating the risk premium on user position.
+   * @notice Allows updating the risk premium on onBehalfOf position.
    * @dev Caller must be `onBehalfOf`, an authorized position manager for `onBehalfOf`, or admin.
-   * @param user The address of the user.
+   * @param onBehalfOf The owner of the position being modified.
    */
-  function updateUserRiskPremium(address user) external;
+  function updateUserRiskPremium(address onBehalfOf) external;
 
   /**
-   * @notice Allows updating the dynamic configuration for all collateral reserves of a user position.
+   * @notice Allows updating the dynamic configuration for all collateral reserves on onBehalfOf position.
    * @dev Caller must be `onBehalfOf`, an authorized position manager for `onBehalfOf`, or admin.
    * @param onBehalfOf The owner of the position being modified.
    */


tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
```diff
@@ -5,25 +5,56 @@ pragma solidity ^0.8.0;
...
+
+  function setUp() public override {
+    super.setUp();
+
     _openSupplyPosition(spoke1, _daiReserveId(spoke1), 2500e18);
     Utils.supplyCollateral(spoke1, _wethReserveId(spoke1), alice, 1e18, alice); // 2k usd
-    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice); // 1k usd
-    Utils.borrow(spoke1, _daiReserveId(spoke1), alice, 2200e18, alice);
+    Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 2000e6, alice); // 2k usd
+    Utils.borrow(spoke1, _daiReserveId(spoke1), alice, 2500e18, alice);
```
yan-man: nit: like above can also specify 2.5k usd for consistency. 

why do these params need changing in the test, wouldnt rp still decrease by doubling the less risky coll price below?

src/contracts/Spoke.sol
```diff
@@ -385,13 +385,11 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
...
   function updateUserRiskPremium(address onBehalfOf) external {
-    (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(onBehalfOf);
-    bool premiumIncrease = _notifyRiskPremiumUpdate(onBehalfOf, userRiskPremium);
-
-    // check permissions if premium increases and not called by user
-    if (premiumIncrease && !_isPositionManager({user: onBehalfOf, manager: msg.sender})) {
+    if (!_isPositionManager({user: onBehalfOf, manager: msg.sender})) {
       _checkCanCall(msg.sender, msg.data);
     }
+    (uint256 userRiskPremium, , , , ) = _calculateUserAccountData(onBehalfOf);
```
yan-man: minor note - will need updating after liq pr that updates _calculateUserAccountData

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
DhairyaSethi: bc of the collateral factor configured

src/contracts/Spoke.sol
DhairyaSethi: yep

tests/unit/Spoke/Spoke.UpdateUserRiskPremium.t.sol
DhairyaSethi: going to remove this file tbh, these two tests are already covered in Spoke.PositionManager.t.sol, this file had a singular test which showed rp cannot be updated by random caller when it increased, no need for these special cases anymore

PR #725, opened 2025-09-10
fix: update events

### LINKED ISSUE #695: Rate not updated on Hub::refreshPremium

### LINKED ISSUE #717: add 'hub' on RateDataUpdated event

yan-man:
- merge after https://github.com/aave/aave-v4/pull/707

closes #695 
closes #717

### DIFF
--- a/src/contracts/Spoke.sol
+++ b/src/contracts/Spoke.sol
@@ -72,7 +72,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
       newOracle != address(0) && oracle.DECIMALS() == Constants.ORACLE_DECIMALS,
       InvalidOracle()
     );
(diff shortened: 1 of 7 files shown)


yan-man:
incorrect branch name, refer to https://github.com/aave/aave-v4/pull/744

PR #733, opened 2025-09-12
Docs/732 natspec

yan-man:


src/libraries/helpers/KeyValueList.sol
```diff
@@ -70,7 +69,6 @@ library KeyValueListInMemory {
   }
 
   function unpack(uint256 data) internal pure returns (uint256, uint256) {
-    // @dev no need to unpack data that was never packed
```
DhairyaSethi, line 73:
```
// @dev no need to unpack data that was never initialized
```

yan-man: done already separately

yan-man:
refer to https://github.com/aave/aave-v4/pull/767

PR #736, opened 2025-09-12
feat: make oracle immutable & hub, kvlist cleanup

DhairyaSethi:
- make spoke.oracle immutable -> we can upgrade oracle with implementation when needed after adding upgradeability
- reduces dup warm sload for reserve.hub
- key value list natspec and name cleanup

gas down 2k each action
code size limit 1.5k
image


src/contracts/Spoke.sol
```diff
@@ -65,16 +67,6 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
   // Governance
   // /////
 
-  /// @inheritdoc ISpoke
```
yan-man, line 64: so we need to upgrade the spoke to update oracle, is that the intention

src/interfaces/ISpoke.sol
```diff
@@ -301,6 +295,8 @@ interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
 
   function getReserveCount() external view returns (uint256);
 
+  function oracle() external view returns (IAaveOracle);
```
yan-man: any reason to move this up?

src/contracts/Spoke.sol
DhairyaSethi: upgrade the spoke implementation yes

src/interfaces/ISpoke.sol
DhairyaSethi: not really, just an artifact of a refactor

src/contracts/Spoke.sol
```diff
@@ -27,20 +27,20 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
   using SafeCast for *;
   using WadRayMath for uint256;
   using PercentageMath for *;
-  using KeyValueListInMemory for KeyValueListInMemory.List;
+  using KeyValueList for KeyValueList.List;
   using PositionStatus for *;
   using MathUtils for *;
 
-  IAaveOracle public oracle;
+  IAaveOracle public immutable oracle;
```
```suggestion
  IAaveOracle public immutable ORACLE;
```
do we wanna make it `address` instead?

src/contracts/Spoke.sol
```diff
@@ -200,11 +192,9 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
   ) external onlyPositionManager(onBehalfOf) {
     DataTypes.Reserve storage reserve = _reserves[reserveId];
     DataTypes.UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
+    IHubBase hub = _validateSupply(reserve);
```
miguelmtzinf: ok, this is v weird and not much benefit i d say

src/contracts/Spoke.sol
```diff
@@ -243,17 +233,12 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
...
-
-    _validateBorrow(reserve);
 
     uint256 drawnShares = hub.draw(assetId, amount, msg.sender);
-
     userPosition.drawnShares += drawnShares.toUint128();
-    if (!positionStatus.isBorrowing(reserveId)) {
-      positionStatus.setBorrowing(reserveId, true);
-    }
+    if (!positionStatus.isBorrowing(reserveId)) positionStatus.setBorrowing(reserveId, true);
```
miguelmtzinf: prefer the long version

src/contracts/Spoke.sol
DhairyaSethi: was suggested by stermi also and it saves one warm sload on reserve.hub -> we use it for existence check and then always interact with it -> no brainer opt imo

DhairyaSethi: i think as long as it's consistent with all validate fn's it's good to have but can rm lmk

miguelmtzinf: 100 gas units vs avoid "why a validate fn returns hub var we already have" question.
i prefer the second honestly

DhairyaSethi: it's a little more than 100 bc it's a packed var so solidity will read, and mask it out again twice but sure yep i see your point, will rm

Issue #734, opened 2025-09-12
Limit the number of Reserves on a Spoke


Kogaroshi:
In PositionStatus.sol, when not finding the next reserve the user an user is borrowing or using as collateral, we return `PositionStatus.NOT_FOUND`, which is `type(uint256).max`.
In `Spoke:_calculateAndPotentiallyRefreshUserAccountData()` and `Spoke:_notifyRiskPremiumUpdate()`, where we use `PositionStatus:next()` to find the next reserve, when receiving `PositionStatus.NOT_FOUND` we exit the loops.
But in the case `type(uint256).max` would be the latests Reserve listed, and so a valid ReserveId, that an user set as collateral or is borrowing, we'd exit the loop due to `next()` return value, instead of iterating like expected.

### Solution :
Add a new condition in `Spoke:addReserve()` to prevent that the new listed reserve's id is `type(uint256).max`.

miguelmtzinf:
Regarding limitations on the number of assets or reserves of the system:

**Hub**
There is no protocol-imposed maximum on the number of assets in the Hub. The asset identifier is `uint256`, allowing up to `2^256` assets to be added.

**Spoke**
It supports up to `2^256` reserves because the reserve identifier is `uint256`. However: 
1. some Spoke actions iterate over user reserves, risking out-of-gas errors with large sets (only matters if those reserves are involved in the user position). Governance should monitor reserve additions to ensure that, even if a user enables all available reserves, transactions remain executable and do not revert.
2. the `PositionStatusMap` library uses the maximum `uint256` value as a sentinel for `NOT_FOUND`. If a reserve with this identifier is used by an user, it will be skipped during processing.
3. when adding a reserve, the Spoke only allows asset identifiers less than `2^16-1` (`uint16`). In other words, a Spoke's reserve cannot connect to Hub assets with identifiers greater than this limit.

DhairyaSethi:
@miguelmtzinf this is talking about reserve ids on the spoke, not asset ids 


> when adding a reserve, the Spoke restricts asset identifiers to uint16 (less than 2^16-1).

this partially addresses the DOS issue, bc reserves are uniquely identifiable to (hub, assetId) pair & not just assetIds

miguelmtzinf:
ty @DhairyaSethi . corrected so it's clearer

miguelmtzinf:
Related to #639

miguelmtzinf:
[...]

PR #740, opened 2025-09-12
fix: Add max assetId check when adding reserve on Spoke

miguelmtzinf:
Partially addresses #734


src/contracts/Spoke.sol
```diff
@@ -116,7 +118,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
     _reserves[reserveId] = DataTypes.Reserve({
       underlying: underlying,
       hub: IHubBase(hub),
-      assetId: assetId.toUint16(),
+      assetId: uint16(assetId),
```
DhairyaSethi, line 117: would retain safe cast since gas is not a concern but nbd

src/libraries/types/DataTypes.sol
```diff
@@ -64,16 +64,15 @@ library DataTypes {
...
-    uint16 assetId;
     uint8 decimals;
     uint16 dynamicConfigKey; // key of the last reserve config
     bool paused;
     bool frozen;
     bool borrowable;
     uint24 collateralRisk;
+    IHubBase hub;
+    uint16 assetId;
+    address underlying;
```
DhairyaSethi:
```suggestion
    uint16 assetId;
    //
    address underlying;
```
using this convention to specify when new slots start

src/contracts/Spoke.sol
```diff
@@ -102,6 +103,7 @@ contract Spoke is ISpoke, Multicall, AccessManaged, EIP712 {
   ) external restricted returns (uint256) {
     require(hub != address(0), InvalidAddress());
     require(!_reserveExists[hub][assetId], ReserveExists());
+    require(assetId < type(uint16).max, InvalidAssetId());
```
DhairyaSethi: type(uint16).max is still a valid id, no? bc it's eq to 2 ** 16 - 1, ie fits in 16 bits
```suggestion
    require(assetId <= type(uint16).max, InvalidAssetId());
```

DhairyaSethi: and let's move this to Constants for now probably
similar to MAX_SUPPORTED_DECIMALS on Hub

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
```diff
@@ -170,7 +170,7 @@ contract SpokeDynamicConfigTest is SpokeBase {
     uint256 invalidReserveId = vm.randomUint(spoke1.getReserveCount(), type(uint256).max);
     DataTypes.DynamicReserveConfig memory dynConf;
 
-    vm.expectRevert(abi.encodeWithSelector(ISpoke.ReserveNotListed.selector, invalidReserveId));
+    vm.expectRevert(abi.encodeWithSelector(ISpoke.ReserveNotListed.selector));
```
DhairyaSethi:
```suggestion
    vm.expectRevert(ISpoke.ReserveNotListed.selector);
```

PR #746, opened 2025-09-12
feat: update old fee receiver caps in hub

### LINKED ISSUE #745: misc HubConfigurator
discussed 9.12.25 
- on hub, on updateAssetConfig with a new fee receiver, automatically set old fee receiver caps to 0

yan-man:
- update old fee receiver caps in the hub directly rather than from hub configurator 

closes #745

### DIFF
--- a/src/dependencies/openzeppelin/ReentrancyGuardTransient.sol
+++ b/src/dependencies/openzeppelin/ReentrancyGuardTransient.sol
@@ -1,0 +1,61 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.1.0) (utils/ReentrancyGuardTransient.sol)
+
+pragma solidity ^0.8.24;
+
+import {TransientSlot} from './TransientSlot.sol';
+
+/**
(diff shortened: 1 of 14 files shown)


src/contracts/Hub.sol
```diff
@@ -128,23 +124,25 @@ contract Hub is IHub, AccessManaged {
...
-      _updateSpokeConfig(
-        assetId,
-        config.feeReceiver,
-        DataTypes.SpokeConfig({addCap: Constants.MAX_CAP, drawCap: 0, active: true})
-      );
+      _addFeeReceiver(assetId, config.feeReceiver);
     }
 
+    asset.liquidityFee = config.liquidityFee;
+    asset.reinvestmentController = config.reinvestmentController;
```
DhairyaSethi, line 141: i was thinking we could very similarly have a separate "if changed condition" for reinvestment controller, exactly like above (on fee receiver and ir strategy) 
that if changed & new feeReciver is addr(0) then require that asset.swept is zero 


but it's shorter and cleaner as is so let's not do it

yan-man: yeah saw your other comment, was considering it but yeah let's save it for now

PR #747, opened 2025-09-12
feat: Move validation logic to library

### LINKED ISSUE #632: Move Validate Functions To Separate Library
Example: `_validateRepay` logic and hub equivalent can be moved to a dedicated / generic logic library

CheyenneAtapour:
closes #632


src/libraries/logic/ValidationLogic.sol
```diff
@@ -0,0 +1,127 @@
...
+  ) internal view {
+    IHub hub = IHub(hubAddress);
+    require(from != hubAddress, IHub.InvalidAddress());
+    require(amount > 0, IHub.InvalidAmount());
+    require(spoke.active, IHub.SpokeNotActive());
+    uint256 addCap = spoke.addCap;
+    require(
+      addCap == Constants.MAX_CAP ||
+        addCap * 10 ** asset.decimals >=
+        hub.previewAddByShares(assetId, spoke.addedShares) + amount,
```
DhairyaSethi, line 40: will have to use asset.convertTo here, this an external call back to hub which is super expensive

miguelmtzinf:
Closing given that it does bring much benefit in terms of gas costs or readability.

PR #748, opened 2025-09-13
Remove DataTypes and Constants from source code so contracts are self-contained.

### LINKED ISSUE #568: Remove DataTypes and move data structs to corresponding interfaces

miguelmtzinf:
In favor of #720 and #721 

Closes #568

#### `PositionStatus` library was renamed to `PositionStatusMap`.
Reason being there is `PositionStatus` lib and struct currently in Spoke contract. We can either:
1. rename lib -> alternatives: `PositionStatusConfiguration` / `PositionConfiguration`)
2. rename struct -> alternatives:  `UserMap` / `PositionMap` / `UserPositionMap` / `UserStateMap`.

#### Only using Constants lib file on tests.
Otherwise we need to do smt like `IHub(hubInstance).CONSTANT()` every time and makes tests code messier.

#### Spoke constants are referring LiquidationLogic constants
Took the decision to move all libs' constants to main contract (`Spoke`) and refer them from there. Main reason is that one of the constants is used in both sides (Spoke and LiquidationLogic lib). Open to suggestions.

### DIFF
--- a/src/interfaces/ISpoke.sol
+++ b/src/interfaces/ISpoke.sol
@@ -58,11 +58,11 @@ interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
 
   struct PositionManagerConfig {
     bool active;
-    mapping(address user => bool approved) approval;
+    mapping(address user => bool) approval;
   }
 
   struct PositionStatus {
-    mapping(uint256 slot => uint256 status) map;
+    mapping(uint256 slot => uint256) map;
   }
 
   struct UserAccountData {
--- a/src/contracts/AaveOracle.sol
+++ b/src/contracts/AaveOracle.sol
@@ -19,7 +19,7 @@ contract AaveOracle is IAaveOracle {
   /// @inheritdoc IAaveOracle
   string public override DESCRIPTION;
 
-  mapping(uint256 reserveId => AggregatorV3Interface source) internal _sources;
+  mapping(uint256 reserveId => AggregatorV3Interface) internal _sources;
 
   /**
    * @dev Constructor.
--- a/src/contracts/AssetInterestRateStrategy.sol
+++ b/src/contracts/AssetInterestRateStrategy.sol
@@ -27,7 +27,7 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
   address public immutable HUB;
 
   /// @dev Map of assetId and their interest rate data (assetId => interestRateData)
-  mapping(uint256 assetId => InterestRateData data) internal _interestRateData;
+  mapping(uint256 assetId => InterestRateData) internal _interestRateData;
 
   /**
    * @dev Constructor.
--- a/src/misc/SignatureGateway.sol
+++ b/src/misc/SignatureGateway.sol
@@ -64,7 +64,7 @@ contract SignatureGateway is ISignatureGateway, Multicall, Rescuable, Ownable2St
     // keccak256('UpdateUserDynamicConfig(address spoke,address user,uint256 nonce,uint256 deadline)')
     0xba177b1f5b5e1e709f62c19f03c97988c57752ba561de58f383ebee4e8d0a71c;
 
-  mapping(address user => uint256 nonce) internal _nonces;
+  mapping(address user => uint256) internal _nonces;
 
   constructor(address spoke_, address initialOwner_) Ownable(initialOwner_) {
     require(spoke_ != address(0) && initialOwner_ != address(0), InvalidAddress());


src/contracts/Hub.sol
```diff
@@ -8,13 +8,12 @@ import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
...
 import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
 
-import {DataTypes} from 'src/libraries/types/DataTypes.sol';
 import {AssetLogic} from 'src/libraries/logic/AssetLogic.sol';
 import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {SharesMath} from 'src/libraries/math/SharesMath.sol';
 import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
 import {MathUtils} from 'src/libraries/math/MathUtils.sol';
-import {Constants} from 'src/libraries/helpers/Constants.sol';
+import {Constants} from 'tests/Constants.sol';
```
DhairyaSethi: pls rm import

src/contracts/Spoke.sol
```diff
@@ -13,10 +13,9 @@ import {SignatureChecker} from 'src/dependencies/openzeppelin/SignatureChecker.s
 import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
 import {KeyValueListInMemory} from 'src/libraries/helpers/KeyValueListInMemory.sol';
-import {Constants} from 'src/libraries/helpers/Constants.sol';
-import {DataTypes} from 'src/libraries/types/DataTypes.sol';
+import {Constants} from 'tests/Constants.sol';
```
DhairyaSethi: pls rm import

src/contracts/Spoke.sol
```diff
@@ -33,22 +32,39 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
-  using PositionStatus for *;
+  using PositionStatusMap for *;
   using MathUtils for *;
 
+  /// @inheritdoc ISpoke
+  uint8 public constant ORACLE_DECIMALS = 8;
+
+  /// @inheritdoc ISpoke
+  uint64 public constant HEALTH_FACTOR_LIQUIDATION_THRESHOLD =
+    LiquidationLogic.HEALTH_FACTOR_LIQUIDATION_THRESHOLD;
```
DhairyaSethi: let's expose a getter for LiquidationLogic library address as well

and side note; maybe we rm this from here and have it be read from the lib instead? (same with leftover base

src/contracts/Spoke.sol
```diff
@@ -33,22 +32,39 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
+  /// @inheritdoc ISpoke
+  uint256 public constant MIN_LEFTOVER_BASE = LiquidationLogic.MIN_LEFTOVER_BASE;
+
+  /// @inheritdoc ISpoke
+  uint24 public constant MAX_COLLATERAL_RISK = 1000_00; // 1000.00%
+
+  /// @inheritdoc ISpoke
+  bytes32 public constant SET_USER_POSITION_MANAGER_TYPEHASH =
+    // keccak256('SetUserPositionManager(address positionManager,address user,bool approve,uint256 nonce,uint256 deadline)')
+    0x758d23a3c07218b7ea0b4f7f63903c4e9d5cbde72d3bcfe3e9896639025a0214;
```
DhairyaSethi, line 46: nice, ok with this, could you pls do the same for SignatureGateway?

src/contracts/Spoke.sol
miguelmtzinf: will do the getter
`maybe we rm this from here and have it be read from the lib instead?`
how, we d need to create a fn here then?

src/libraries/configuration/PositionStatusMap.sol
```diff
@@ -3,15 +3,15 @@
...
+import {ISpoke} from 'src/interfaces/ISpoke.sol';
 
 /**
  * @title PositionStatus
  * @author Aave Labs
  * @notice Implements the bitmap logic to handle the user configuration.
  */
-library PositionStatus {
-  using PositionStatus for *;
+library PositionStatusMap {
```
miguelmtzinf, line 13: are we good with this new name? or we prefer to rename `PositionStatus` struct. Maybe renaming that one is better indeed...

src/contracts/Hub.sol
```diff
@@ -26,13 +25,18 @@ contract Hub is IHub, AccessManaged {
...
+
+  /// @inheritdoc IHub
+  uint56 public constant MAX_CAP = type(uint56).max;
+
   uint256 internal _assetCount;
-  mapping(uint256 assetId => DataTypes.Asset assetData) internal _assets;
-  mapping(uint256 assetId => mapping(address spoke => DataTypes.SpokeData spokeData))
-    internal _spokes;
+  mapping(uint256 assetId => Asset assetData) internal _assets;
+  mapping(uint256 assetId => mapping(address spoke => SpokeData spokeData)) internal _spokes;
```
DhairyaSethi:
```suggestion
  mapping(uint256 assetId => Asset) internal _assets;
  mapping(uint256 assetId => mapping(address spoke => SpokeData)) internal _spokes;
```
feels those are redundant and this is cleaner but nbd

src/contracts/Spoke.sol
DhairyaSethi: no i mean it's public on the library so it also exposes that constant, the flow will be getLiquidationLogic() and then read constant from there

but makes things distributed, i'm ok with how it is here

src/contracts/Spoke.sol
```diff
@@ -786,10 +785,10 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     bool collateral;
     while (true) {
       (reserveId, borrowing, collateral) = positionStatus.next(reserveId);
-      if (reserveId == PositionStatus.NOT_FOUND) break;
+      if (reserveId == PositionStatusMap.NOT_FOUND) break;
```
DhairyaSethi, line 792: maybe we call it `PositionConfiguration`  bc of this?

src/interfaces/IHub.sol
```diff
@@ -13,30 +11,78 @@ import {IHubBase} from 'src/interfaces/IHubBase.sol';
...
+  struct AssetConfig {
+    address feeReceiver;
+    uint16 liquidityFee;
+    address irStrategy;
+    address reinvestmentController;
+  }
+
+  struct SpokeData {
+    //
+    uint128 premiumShares;
```
```suggestion
    uint128 premiumShares;
```

DhairyaSethi: still feels weird having internal data structures in the interface

src/interfaces/ISpoke.sol
```diff
@@ -2,20 +2,83 @@
...
+
+  struct LiquidationConfig {
+    uint128 targetHealthFactor; // WAD, HF value to restore to during a liquidation
+    uint64 healthFactorForMaxBonus; // WAD, health factor under which liquidation bonus is max
+    uint16 liquidationBonusFactor; // BPS, as a percentage of effective lb
+  }
+
+  struct UserPosition {
+    //
+    uint128 drawnShares;
```
```suggestion
    uint128 drawnShares;
```

src/interfaces/ISpoke.sol
```diff
@@ -267,31 +330,37 @@ interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
...
+  function MIN_LEFTOVER_BASE() external view returns (uint256);
+
+  function MAX_COLLATERAL_RISK() external view returns (uint24);
+
+  function SET_USER_POSITION_MANAGER_TYPEHASH() external view returns (bytes32);
+
   function getHealthFactor(address user) external view returns (uint256);
 
-  function getReserve(uint256 reserveId) external view returns (DataTypes.Reserve memory);
+  function getReserve(uint256 reserveId) external view returns (ISpoke.Reserve memory);
```
DhairyaSethi: all of these don't need ISpoke, can just be:
```suggestion
  function getReserve(uint256 reserveId) external view returns (Reserve memory);
```

src/libraries/configuration/PositionStatusMap.sol
DhairyaSethi: leaning towards PositionConfiguration

src/libraries/logic/LiquidationLogic.sol
```diff
@@ -315,14 +331,14 @@ library LiquidationLogic {
...
-    DataTypes.LiquidationConfig storage liquidationConfig,
-    DataTypes.DynamicReserveConfig storage collateralDynConfig,
-    DataTypes.LiquidateUserParams memory params
+    ISpoke.Reserve storage collateralReserve,
+    ISpoke.Reserve storage debtReserve,
+    ISpoke.UserPosition storage collateralPosition,
+    ISpoke.UserPosition storage debtPosition,
+    ISpoke.PositionStatus storage positionStatus,
+    ISpoke.LiquidationConfig storage liquidationConfig,
+    ISpoke.DynamicReserveConfig storage collateralDynConfig,
```
DhairyaSethi, line 338: would look much cleaner if just pass the mapping storage pointer and reserveIds, userAddress instead here tbh

src/contracts/Hub.sol
DhairyaSethi: @miguelmtzinf missed this

src/interfaces/IHub.sol
miguelmtzinf: exposing `SpokeData` in `getSpoke`

miguelmtzinf: do you wanna remove all `//`? or just this one as it's the first and don't needed

DhairyaSethi: just this only bc first and not needed, rest are all imp imo

DhairyaSethi: > exposing SpokeData in getSpoke

yep yep, still weird. i think these warrant some natspec also bc they're here but let's do that separately

PR #749, opened 2025-09-13
Remove individual getters for user hf and rp on spoke

### LINKED ISSUE #741: Naming of hub getters
The naming of `getAssetAddedAmount` is a bit strange because it's the only place we use "amount", but also `getAssetAddedAssets` is a bit strange

Rename `getAssetAddedAmount` to `getAddedAssets` and `getAssetAddedShares` to `getAddedShares`

miguelmtzinf:
We can remove these 2 getters on rely on `getUserAccountData`
closes #741

### DIFF
--- a/src/interfaces/ITreasurySpoke.sol
+++ b/src/interfaces/ITreasurySpoke.sol
@@ -16,7 +16,7 @@ interface ITreasurySpoke is ISpokeBase {
    * @notice Supplies a specified amount of the underlying asset to a given reserve.
    * @dev The Hub pulls the underlying asset from the caller, so prior approval is required.
    * @dev The reserve identifier **should match** corresponding asset identifier in the Hub.
-   * @param reserveId The identifier of the reserve
+   * @param reserveId The identifier of the reserve.
    * @param amount The amount of asset to supply.
    * @param onBehalfOf Unused parameter for this spoke.
    */


src/contracts/Spoke.sol
```diff
@@ -35,7 +35,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
   using MathUtils for *;
 
   /// @inheritdoc ISpoke
-  uint256 public constant MAX_RESERVE_ID = type(uint16).max;
+  uint256 public constant MAX_ASSET_ID = type(uint16).max;
```
```suggestion
  uint256 public constant MAX_SUPPORTED_ASSET_ID = type(uint16).max;
```
similar to how we have max_supported_asset_decimals 
supported convention is for something outside of this contract's scope or control, here asset id is a hub property so

src/contracts/Hub.sol
```diff
@@ -31,7 +31,7 @@ contract Hub is IHub, AccessManaged {
   uint8 public constant MAX_ALLOWED_ASSET_DECIMALS = 18;
 
   /// @inheritdoc IHub
-  uint56 public constant SPOKE_MAX_CAP = type(uint56).max;
+  uint56 public constant MAX_ALLOWED_SPOKE_CAP = type(uint56).max;
```
DhairyaSethi, line 34: this was fine before, max allowed doesn't make sense here since this is used as a sentinel value

src/contracts/Spoke.sol
```diff
@@ -35,10 +35,10 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
   using MathUtils for *;
 
   /// @inheritdoc ISpoke
-  uint256 public constant MAX_RESERVE_ID = type(uint16).max;
+  uint256 public constant MAX_ALLOWED_ASSET_ID = type(uint16).max;
 
   /// @inheritdoc ISpoke
-  uint24 public constant MAX_COLLATERAL_RISK = 1000_00; // 1000.00%
+  uint24 public constant MAX_ALLOWED_COLLATERAL_RISK = 1000_00; // 1000.00%
```
DhairyaSethi, line 41: maybe then also switch on hub max allowed token decimals const?

miguelmtzinf: it's called `MAX_ALLOWED_ASSET_DECIMALS`, maybe `MAX_ALLOWED_UNDERLYING_DECIMALS`?

PR #758, opened 2025-09-14
fix: reorg dir structure

### LINKED ISSUE #751: Directory structure
Few aspects to consider:
1. Hub and Spoke are independent
2. There will be many implementations of hubs and spokes in the future
3. Some libraries are shared by hub and spoke
4. there can be future implementations: new position managers, wrappers, data provider contracts, swap-adapters, etc.

Suggestion:
- src/deployments
- src/contracts
  - dependencies
  - libraries
  - hubs
    - libraries
    - interfaces
  - spokes
    - libraries
    - interfaces
  - position-managers
  - extensions
    - swap-adapters
  - misc
    - Multicall

open questions:
- what do we place a new implementation of the Spoke

miguelmtzinf:
Closes #751

### DIFF
--- a/src/dependencies/openzeppelin-upgradeable/AccessManagedUpgradeable.sol
+++ b/src/dependencies/openzeppelin-upgradeable/AccessManagedUpgradeable.sol
@@ -3,10 +3,10 @@
 
 pragma solidity ^0.8.20;
 
-import {IAuthority} from '@openzeppelin/contracts/access/manager/IAuthority.sol';
-import {AuthorityUtils} from '@openzeppelin/contracts/access/manager/AuthorityUtils.sol';
-import {IAccessManager} from '@openzeppelin/contracts/access/manager/IAccessManager.sol';
-import {IAccessManaged} from '@openzeppelin/contracts/access/manager/IAccessManaged.sol';
+import {IAuthority} from 'src/dependencies/openzeppelin/IAuthority.sol';
+import {AuthorityUtils} from 'src/dependencies/openzeppelin/AuthorityUtils.sol';
+import {IAccessManager} from 'src/dependencies/openzeppelin/IAccessManager.sol';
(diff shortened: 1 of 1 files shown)


src/contracts/hub/AssetInterestRateStrategy.sol
```diff
@@ -2,8 +2,8 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity ^0.8.10;
 
-import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
-import {IAssetInterestRateStrategy, IBasicInterestRateStrategy} from 'src/interfaces/IAssetInterestRateStrategy.sol';
+import {WadRayMath} from 'src/contracts/libraries/math/WadRayMath.sol';
```
DhairyaSethi: tbh not a fan of src/contracts 
was better before with everything directly in src and then can have deployments separately 
i think src/contracts is tech debt of v3 from hardhat days

src/contracts/dependencies/openzeppelin/AccessManager.sol
```diff
@@ -7,7 +7,7 @@ import {IAccessManager} from './IAccessManager.sol';
 import {IAccessManaged} from './IAccessManaged.sol';
 import {Address} from './Address.sol';
 import {Context} from './Context.sol';
-import {Multicall} from 'src/misc/Multicall.sol';
+import {Multicall} from 'src/contracts/utils/Multicall.sol';
```
DhairyaSethi: shouldn't have been changing this in the first place. would rather we copy oz multicall too just so deps have no change in them and are static copies

miguelmtzinf: brought OZ Multicall back

src/contracts/hub/AssetInterestRateStrategy.sol
miguelmtzinf: reverted

PR #768, opened 2025-09-16
feat: add deficit when checking draw cap

### LINKED ISSUE #764: Spoke caps can be bypassed by compromised spokes
While there is a strong trust assumption between Hubs and Spokes in V4, the impact of a compromised spoke is generally limited through configurations such as active flag and caps on the Hub side. 

However, a compromised spoke can bypass the draw cap by looping draw + report deficit operations. This is possible because caps are only checked against the current drawn amount of a spoke, without taking its deficit into consideration. A fix can add an additional `deficit` field in the `SpokeData` struct, which gets updated on `reportDeficit` and `eliminateDeficit`. To do this, the `eliminateDeficit` function must accept an additional argument to specify which spoke is the deficit covered for. This would then allow to check for draw cap by adding the draw amount with the active deficit of the spoke, hence protecting against this attack.

Similarly, a compromised spoke can steal an arbitrary amount of funds from users who approved their tokens to the same hub, even though they intended to interact with a different spoke. This is because spokes control the `from` address when the hub transfers funds. The spoke can easily bypass the add cap in this case by looping add + remove operations.

avniculae:
- deficit accounting on a per-spoke basis; must specify the spoke for which deficit is eliminated
- prevents bypassing the draw cap through a repeated draw + report deficit loop 

Partially closes #764

### DIFF
--- a/snapshots/Hub.Operations.json
+++ b/snapshots/Hub.Operations.json
@@ -1,6 +1,6 @@
 {
   "add": "118209",
-  "draw": "111089",
+  "draw": "108440",
   "eliminateDeficit: full": "62298",
   "eliminateDeficit: partial": "71874",
   "payFee": "77008",
--- a/snapshots/Spoke.Operations.ZeroRiskPremium.json
+++ b/snapshots/Spoke.Operations.ZeroRiskPremium.json
@@ -1,6 +1,6 @@
 {
-  "borrow: first": "225757",
-  "borrow: second action, same reserve": "206329",
+  "borrow: first": "223108",
+  "borrow: second action, same reserve": "203680",
   "liquidationCall: full": "311216",
   "liquidationCall: partial": "331678",
   "permitReserve + repay (multicall)": "263675",
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -1,6 +1,6 @@
 {
-  "borrow: first": "251257",
-  "borrow: second action, same reserve": "214729",
+  "borrow: first": "248608",
+  "borrow: second action, same reserve": "212080",
   "liquidationCall: full": "319832",
   "liquidationCall: partial": "340294",
   "permitReserve + repay (multicall)": "272291",


src/hub/Hub.sol
```diff
@@ -302,23 +304,29 @@ contract Hub is IHub, AccessManaged {
...
-  function eliminateDeficit(uint256 assetId, uint256 amount) external returns (uint256) {
+  function eliminateDeficit(
+    uint256 assetId,
+    uint256 amount,
+    address spokeToCover
+  ) external returns (uint256) {
     Asset storage asset = _assets[assetId];
-    SpokeData storage spoke = _spokes[assetId][msg.sender];
+    SpokeData storage senderSpoke = _spokes[assetId][msg.sender];
+    SpokeData storage receiverSpoke = _spokes[assetId][spokeToCover];
```
yan-man: nit: not exactly I think a sender/receiver in this case, also a bit inconsistent with the spokeToCover terminology. Maybe something like callerSpoke / coveredSpoke?

src/hub/Hub.sol
```diff
@@ -591,7 +604,7 @@ contract Hub is IHub, AccessManaged {
     return _assets[assetId].liquidity;
   }
 
-  function getDeficit(uint256 assetId) external view returns (uint256) {
+  function getAssetDeficit(uint256 assetId) external view returns (uint256) {
```
yan-man, line 607: ok so ig a new invariant we have now is that sum across all spoke deficits == asset deficit for a given asset. No rounding so this is exact

tests/unit/Hub/Hub.EliminateDeficit.t.sol
```diff
@@ -5,117 +5,100 @@ pragma solidity ^0.8.0;
...
-    hub1.eliminateDeficit(assetId, vm.randomUint());
+    hub1.eliminateDeficit(assetId, vm.randomUint(), receiverSpoke);
   }
 
-  function test_eliminateDeficit() public {
-    uint256 assetId = _randomAssetId(hub1);
-    uint256 deficit = 1000e6;
+  function test_eliminateDeficit(uint256) public {
+    _createDeficit(assetId, receiverSpoke, deficitAmount);
+    _createDeficit(assetId, otherSpoke, deficitAmount / 2);
```
yan-man: i think would be helpful to create another var for the 2nd spoke's deficit so it's clearer below where the /2 comes from

src/hub/Hub.sol
```diff
@@ -302,23 +304,29 @@ contract Hub is IHub, AccessManaged {
...
-    SpokeData storage spoke = _spokes[assetId][msg.sender];
+    SpokeData storage senderSpoke = _spokes[assetId][msg.sender];
+    SpokeData storage receiverSpoke = _spokes[assetId][spokeToCover];
 
     asset.accrue(assetId, _spokes[assetId][asset.feeReceiver]);
-    _validateEliminateDeficit(spoke, amount);
-    uint256 deficit = asset.deficit;
+    _validateEliminateDeficit(senderSpoke, amount);
+    uint256 deficit = receiverSpoke.deficit;
     require(amount <= deficit, InvalidAmount());
```
yan-man, line 319: ig by default, amount SHOULD be less than asset deficit, is it worth it to check that here as well and not just for a spoke's deficit

src/hub/interfaces/IHub.sol
```diff
@@ -343,6 +352,14 @@ interface IHub is IHubBase, IAccessManaged {
 
   function getSpoke(uint256 assetId, address spoke) external view returns (SpokeData memory);
 
+  /**
+   * @notice Returns the amount of deficit of the specified asset and spoke.
```
yan-man:
```suggestion
   * @notice Returns a spoke's deficit amount for the specified asset.
```

src/hub/Hub.sol
```diff
@@ -688,7 +701,8 @@ contract Hub is IHub, AccessManaged {
     uint256 addCap = spoke.addCap;
     require(
       addCap == MAX_ALLOWED_SPOKE_CAP ||
-        addCap * 10 ** asset.decimals >= previewAddByShares(assetId, spoke.addedShares) + amount,
+        addCap * uint256(10).uncheckedExp(asset.decimals) >=
```
```suggestion
        addCap * MathUtils.uncheckedExp(10, asset.decimals) >=
```

yan-man: just aesthetics, yes?

DhairyaSethi: consistency

src/hub/Hub.sol
DhairyaSethi: it would be by definition since sum of all spoke deficits is asset deficit 

for extra extra sanity we can rm the uncheckedSub on 324 (while keeping it on 325)

src/hub/interfaces/IHub.sol
```diff
@@ -207,9 +211,14 @@ interface IHub is IHubBase, IAccessManaged {
...
    * @param assetId The identifier of the asset.
    * @param amount The amount of deficit to eliminate.
+   * @param spokeToCover The spoke for which the deficit is eliminated.
    * @return The amount of shares removed.
    */
-  function eliminateDeficit(uint256 assetId, uint256 amount) external returns (uint256);
+  function eliminateDeficit(
+    uint256 assetId,
+    uint256 amount,
+    address spokeToCover
```
```suggestion
    address spoke
```

src/hub/interfaces/IHub.sol
```diff
@@ -207,9 +211,14 @@ interface IHub is IHubBase, IAccessManaged {
    * @dev Only callable by active spokes.
    * @param assetId The identifier of the asset.
    * @param amount The amount of deficit to eliminate.
+   * @param spokeToCover The spoke for which the deficit is eliminated.
```
```suggestion
   * @param spoke The spoke for which the deficit is eliminated.
```
to be consistent with implementation

src/hub/Hub.sol
```diff
@@ -302,23 +304,30 @@ contract Hub is IHub, AccessManaged {
...
+    uint256 deficit = coveredSpoke.deficit;
     require(amount <= deficit, InvalidAmount());
 
     uint128 shares = previewRemoveByAssets(assetId, amount).toUint128();
     asset.addedShares -= shares;
-    spoke.addedShares -= shares;
-    asset.deficit = deficit.uncheckedSub(amount).toUint128();
+    callerSpoke.addedShares -= shares;
+    // ensure that amount never exceeds asset.deficit
+    asset.deficit -= amount.toUint128();
```
```suggestion
    asset.deficit -= amount.toUint128();
```

src/hub/interfaces/IHub.sol
```diff
@@ -57,6 +57,8 @@ interface IHub is IHubBase, IAccessManaged {
     uint56 addCap;
     uint56 drawCap;
     bool active;
+    //
+    uint128 deficit;
```
DhairyaSethi, line 61: in the real world, it will be cheaper to keep caps, active in the same slot as deficit rather than added shares because drawCap should ideally never be infinite so it will always be read, so it makes sense that it's warm with the active, and drawCap

DhairyaSethi: but makes supply more expensive bc active is not warm by 2k

PR #774, opened 2025-09-17
feat : refactors hub transferFrom to be done by Spokes instead - option 1

Kogaroshi:
See #764

Change the fund transfer logic from Hub to Spoke for `add` & `restore` to instead call `transferAssetsFrom` in the Spoke. This means users wouldn't approve the Hubs anymore, but instead have allowances per Spoke.

TODO : 
- see how to manage donations
- fix tests
- add tests for new transferFrom method


src/spoke/Spoke.sol
```diff
@@ -487,6 +489,22 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+  function transferAssetsFrom(
+    uint256 assetId,
+    address underlying,
+    address from,
+    uint256 amount
+  ) external {
+    // also enforces the Hub is allowed to call this function
+    require(_reserveExists[msg.sender][assetId], AssetNotListed());
+    if (from == address(this)) {
+      underlying.safeTransfer(msg.sender, amount);
```
miguelmtzinf: this is not used at all no?

Kogaroshi: this was removed.

Kogaroshi:

PR #775, opened 2025-09-17
rft : refactor hub transferFrom to be done by Spokes instead - option 2

Kogaroshi:
See #764

Change the fund transfer logic from Hub to Spoke for `add` & `restore` to instead have the Spoke execute the `transferFrom` before calling the Hub, and the Hub only checking its balance is >= to the new recorded asset liquidity. This means users wouldn't approve the Hubs anymore, but instead have allowances per Spoke.

TODO : 
- see how to manage donations

https://github.com/aave/aave-v4/pull/774 As an alternate approach where Hub triggers pulling assets of users by spokes.


src/hub/Hub.sol
```diff
@@ -13,6 +13,7 @@ import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {AssetLogic} from 'src/hub/libraries/AssetLogic.sol';
 import {SharesMath} from 'src/hub/libraries/SharesMath.sol';
 import {IBasicInterestRateStrategy} from 'src/hub/interfaces/IBasicInterestRateStrategy.sol';
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
```
yan-man: rm

src/hub/Hub.sol
```diff
@@ -204,7 +205,11 @@ contract Hub is IHub, AccessManaged {
 
     asset.updateDrawnRate(assetId);
 
-    IERC20(asset.underlying).safeTransferFrom(from, address(this), amount);
+    /// @dev Check the received amount instead of trusting Spokes, prevents malicious Spokes to transfer less.
```
yan-man: normal comments // outside of natspec. Maybe just: `enforces spoke transfers funds from user to hub`

src/hub/Hub.sol
```diff
@@ -282,7 +287,11 @@ contract Hub is IHub, AccessManaged {
 
     asset.updateDrawnRate(assetId);
 
-    IERC20(asset.underlying).safeTransferFrom(from, address(this), totalAmount);
+    /// @dev Check the received amount instead of trusting Spokes, prevents malicious Spokes to transfer less.
```
yan-man: comment convention, maybe can shorten like above

tests/gas/Hub.Operations.gas.t.sol
```diff
@@ -14,13 +14,15 @@ contract HubOperations_Gas_Tests is Base {
   }
 
   function test_add() public {
-    vm.prank(address(spoke1));
+    vm.startPrank(address(spoke1));
+    tokenList.usdx.transferFrom(alice, address(hub1), 1000e6);
     hub1.add(usdxAssetId, 1000e6, alice);
```
yan-man: add a stopPrank for completeness

tests/gas/Hub.Operations.gas.t.sol
```diff
@@ -14,13 +14,15 @@ contract HubOperations_Gas_Tests is Base {
...
     vm.snapshotGasLastCall('Hub.Operations', 'add');
   }
 
   function test_remove() public {
     vm.startPrank(address(spoke1));
+    tokenList.usdx.transferFrom(alice, address(hub1), 1000e6);
     hub1.add(usdxAssetId, 1000e6, alice);
     hub1.remove(usdxAssetId, 500e6, alice);
```
yan-man, line 28: stopPrank

src/spoke/Spoke.sol
```diff
@@ -201,6 +204,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
     _validateSupply(reserve);
 
+    IERC20(reserve.underlying).safeTransferFrom(msg.sender, address(reserve.hub), amount);
     uint256 suppliedShares = reserve.hub.add(reserve.assetId, amount, msg.sender);
```
DhairyaSethi: we can avoid passing msg.sender now here and in repay

src/hub/Hub.sol
```diff
@@ -204,7 +204,11 @@ contract Hub is IHub, AccessManaged {
 
     asset.updateDrawnRate(assetId);
 
-    IERC20(asset.underlying).safeTransferFrom(from, address(this), amount);
+    // Enforces spoke transfers the correct funds from user to hub.
```
DhairyaSethi: not needed here i think, in any case lowercase without full stop for internal comments 

also add dev comment in natspec for this (in interface for both add & restore) AND another dev comment to pls mention any donated liquidity can be skimmed by any spoke (ie its intended behaviour) 


also pls evaluate the entire hub contract, that all liquidity decrease leads to an asset transfer out & all liquidity increase leads to an asset transfer in (dont forget about sweep, reclaim) - the only exception to this would be eliminate & report deficit where liquidity is decrease/increase is ok - pls validate basically this change is ok 

i think a graph about all asset transfers & change in asset.liquidity would be cool

src/hub/Hub.sol
avniculae, line 1: should we add comments before `InsufficientLiquidity()` errors to mention that these are mandatory now given we perform unchecked sub below?

src/hub/interfaces/IHub.sol
```diff
@@ -198,6 +198,9 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev The `irData` must be empty if the interest rate strategy is not updated.
   error InvalidInterestRateStrategy();
 
+  /// @notice Thrown when the amount of underlying tokens received is less than expected.
```
avniculae: tokens or assets?

src/hub/interfaces/IHubBase.sol
```diff
@@ -91,11 +91,12 @@ interface IHubBase {
 
   /// @notice Add asset on behalf of user.
   /// @dev Only callable by active spokes.
+  /// @dev Must receive the funds prior to calling, this hub balance is verified against the recorded liquidity.
```
avniculae: wdyt of
```suggestion
  /// @dev Must receive the funds prior to calling, hub's balance is verified against the stored liquidity.
```

avniculae: same below

src/spoke/Spoke.sol
```diff
@@ -201,7 +204,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
     _validateSupply(reserve);
 
-    uint256 suppliedShares = reserve.hub.add(reserve.assetId, amount, msg.sender);
+    IERC20(reserve.underlying).safeTransferFrom(msg.sender, address(reserve.hub), amount);
```
avniculae: should we cache reserve.hub here? (optimiser does the job anyways)

src/spoke/Spoke.sol
```diff
@@ -292,12 +296,16 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       offsetDelta: -userPosition.premiumOffset.toInt256(),
       realizedDelta: accruedPremium.toInt256() - premiumDebtRestored.toInt256()
     });
+    IERC20(reserve.underlying).safeTransferFrom(
```
avniculae: should we store underlying as IERC20 now?

Kogaroshi: don't mind keeping it as `address`, wdyt @DhairyaSethi @miguelmtzinf ?

src/spoke/Spoke.sol
Kogaroshi: tested the change, the gas cost increases, so left as is.

src/hub/Hub.sol
```diff
@@ -204,10 +204,7 @@ contract Hub is IHub, AccessManaged {
     asset.updateDrawnRate(assetId);
 
     // enforces spoke transfers the correct funds from user to hub
-    require(
-      IERC20(asset.underlying).balanceOf(address(this)) >= asset.liquidity,
-      InvalidAmountReceived()
-    );
+    require(asset.underlying.balanceOf(address(this)) >= asset.liquidity, InvalidAmountReceived());
```
DhairyaSethi, line 8: i think we can now cache liquidity no?

src/position-manager/SignatureGateway.sol
```diff
@@ -103,14 +103,14 @@ contract SignatureGateway is ISignatureGateway, GatewayBase, NoncesKeyed, Multic
     require(SignatureChecker.isValidSignatureNow(user, digest, signature), InvalidSignature());
     _useCheckedNonce(user, params.nonce);
 
-    (address underlying, address hub) = _getReserveData(spoke, reserveId);
+    (address underlying, ) = _getReserveData(spoke, reserveId);
```
DhairyaSethi: does that method even need to return hub now? can this just become getReserveUnderlying?

Kogaroshi: yeah I considered simplifying it too, do we consider other Gateways might need to get the Hub alongside the underlying ?

DhairyaSethi: idts tbh, if needed they can always implement

Kogaroshi: okay, I'll simplify then

src/hub/Hub.sol
```diff
@@ -191,22 +191,24 @@ contract Hub is IHub, AccessManaged {
...
     asset.addedShares += shares;
     spoke.addedShares += shares;
-    asset.liquidity += amount.toUint128();
+    asset.liquidity = newLiquidity.toUint128();
 
     asset.updateDrawnRate(assetId);
 
-    asset.underlying.safeTransferFrom(from, address(this), amount);
+    // enforces spoke transfers the correct funds from user to hub
+    require(asset.underlying.balanceOf(address(this)) >= newLiquidity, InvalidAmountReceived());
```
miguelmtzinf: `>` is needed because:
i. spoke could have transferred more than `amount`
ii. there could have been some airdrops, so `asset.liquidity < balanceOf`

Wondering if we should do `==` instead and do `asset.liquidity = balanceOf` instead

src/hub/Hub.sol
miguelmtzinf, line 252: kinda asymetric now. We could give approval to spoke, so it can send funds to whoever they want.

src/position-manager/GatewayBase.sol
```diff
@@ -47,13 +47,9 @@ abstract contract GatewayBase is IGatewayBase, Rescuable, Ownable2Step {
...
 
   /// @return The underlying asset for `reserveId` on the specified spoke.
-  /// @return The corresponding hub address.
-  function _getReserveData(
-    address spoke,
-    uint256 reserveId
-  ) internal view returns (address, address) {
+  function _getReserveUnderlying(address spoke, uint256 reserveId) internal view returns (address) {
     ISpoke.Reserve memory reserveData = ISpoke(spoke).getReserve(reserveId);
-    return (reserveData.underlying, address(reserveData.hub));
```
miguelmtzinf, line 56: 🥇

src/hub/Hub.sol
Kogaroshi: I didn't put `==` to prevent griefing attacks by frontrunning and donating to the Hub to block an user deposit/repay.

But doing `asset.liquidity = balanceOf` after the check could absorb the donations without minting any shares, dunno if it's the path we want to go with ?

src/hub/Hub.sol
Kogaroshi: Gas wise it would also make those methods more expensive for all Spokes. If a Spoke want's to send to specific addresses, you can have the `to = spoke` when calling the Hub, and then it's up to the Spoke do transfer to whatever, so only those exotic Spokes have increased gas costs, while the current one stays cheaper.

src/hub/Hub.sol
DhairyaSethi: i think `>=` makes the most sense here

> But doing asset.liquidity = balanceOf after the check could absorb the donations without minting any shares, dunno if it's the path we want to go with 

would require more changes on the spoke, no? to always check if it's there's donated liquidity available to be absorbed for each user action to not prevent them from going through

miguelmtzinf: correct, `asset.liquidity = balanceOf` would make supply index absorb any airdrop/donation. If we wanna avoid any donation/airdrop, we can use `=` and calculate `balanceOf` difference for the input amount in asset terms (`balanceOf` over `asset.liquidity` for the starting balance).

src/hub/Hub.sol
miguelmtzinf: they can transfer to whoever they want now, by using the `to` no? They can do it with my suggestion as well.
What I'm wondering is if makes sense to give allowance to spoke and remove the `to` input.

Kogaroshi: consistency wise, it would make sense, to remove `to` and either make the Spoke do the transfer, but gas wise we'd increase for no real addition there.

DhairyaSethi: it's slightly cheaper this way (no allowance consumed on token) & we don't have to play around (ie max approve spoke on listed) & worry about allowance of tokens internally 

think the asymmetry makes sense

src/hub/Hub.sol
miguelmtzinf: not an option because this is executed once the transfer happens.

yan-man: should we move this up to like L201 to fail early before accounting (as it stands is this reentrancy risk?) 

We could also add this to validateAdd

src/hub/Hub.sol
```diff
@@ -268,25 +270,25 @@ contract Hub is IHub, AccessManaged {
...
-    uint256 totalAmount = drawnAmount + premiumAmount;
-    asset.liquidity += totalAmount.toUint128();
+    uint256 newLiquidity = asset.liquidity + drawnAmount + premiumAmount;
+    asset.liquidity = newLiquidity.toUint128();
 
     asset.updateDrawnRate(assetId);
 
-    asset.underlying.safeTransferFrom(from, address(this), totalAmount);
+    // enforces spoke transfers the correct funds from user to hub
+    require(asset.underlying.balanceOf(address(this)) >= newLiquidity, InvalidAmountReceived());
```
yan-man: same move up to L286?

src/hub/Hub.sol
Kogaroshi: We can move it right after the `newLiquidity` is calculated indeed, to revert early.

But I'm not sure reentrancy risk would arise here, the transfer is done before calling `Hub.add()`, and the only external call here is a `balanceOf`.

src/spoke/TreasurySpoke.sol
```diff
@@ -59,7 +60,7 @@ contract TreasurySpoke is ITreasurySpoke, Ownable2Step {
 
   /// @inheritdoc ITreasurySpoke
   function transfer(address token, address to, uint256 amount) external onlyOwner {
```
DhairyaSethi, line 62: let's just inherit rescuable on this at this point, wdyt?

Kogaroshi: on the needed methods yeah, this `transfer` can be replaced by `Rescuable.rescueToken()`. But on the naming side, here it's Treasury making transfer, not rescuing lost assets, so idk.

tests/unit/Hub/Hub.Add.t.sol
```diff
@@ -47,37 +47,48 @@ contract HubAddTest is HubBase {
...
+        collateralFactor: 78_00,
+        maxLiquidationBonus: 100_00,
+        liquidationFee: 0
+      })
+    );
     hub1.addSpoke(minDecimalAssetId, address(spoke1), spokeConfig);
     vm.stopPrank();
   }
 
-  function test_add_revertsWith_TransferFromFailed() public {
```
DhairyaSethi, line 54: we still want this test no?

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -8,6 +8,39 @@ contract SpokeRepayTest is SpokeBase {
   using PercentageMath for uint256;
   using SafeCast for uint256;
 
+  function test_supply_revertsWith_TransferFromFailed() public {
```
DhairyaSethi:
```suggestion
  function test_repay_revertsWith_TransferFromFailed() public {
```
pls add back in supply also

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -8,6 +8,39 @@ contract SpokeRepayTest is SpokeBase {
...
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+
+    vm.startPrank(bob);
+    tokenList.dai.approve(address(spoke1), approvalAmount);
+    vm.expectRevert(SafeTransferLib.TransferFromFailed.selector);
+    spoke1.repay(_daiReserveId(spoke1), daiRepayAmount, bob);
+    vm.stopPrank();
+  }
+
+  function test_supply_fuzz_revertsWith_TransferFromFailed(uint256 daiRepayAmount) public {
```
DhairyaSethi: same

tests/unit/Hub/Hub.Add.t.sol
Kogaroshi: was moved here https://github.com/aave/aave-v4/blob/60847ab7b8df93f0d75aa2cc79200f074b936589/tests/unit/Spoke/Spoke.Supply.t.sol#L41 since it's the Spoke doing the transfer now

tests/unit/Spoke/Spoke.Repay.t.sol
```diff
@@ -8,6 +8,39 @@ contract SpokeRepayTest is SpokeBase {
...
+    uint256 approvalAmount = daiRepayAmount - 1;
+
+    Utils.supplyCollateral(spoke1, _wethReserveId(spoke1), bob, wethSupplyAmount, bob);
+    Utils.supply(spoke1, _daiReserveId(spoke1), alice, daiSupplyAmount, alice);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, daiBorrowAmount, bob);
+
+    vm.startPrank(bob);
+    tokenList.dai.approve(address(spoke1), approvalAmount);
+    vm.expectRevert(SafeTransferLib.TransferFromFailed.selector);
+    spoke1.repay(_daiReserveId(spoke1), daiRepayAmount, bob);
```
DhairyaSethi, line 25: let's also do an expectCall on the token with the transferFrom params and correct caller since the error is not clear bc of lib? wdyt? can avoid if overkill but would love once for supply & repay for sanity

let's also add a test where this fails bc of insufficient allowance (error thrown by lib is the same, it doesn't bubble up errors)

tests/Base.t.sol
```diff
@@ -388,11 +388,28 @@ abstract contract Base is Test {
...
-      tokenList.usdy.approve(address(hub1), UINT256_MAX);
+      tokenList.weth.approve(address(spoke1), UINT256_MAX);
+      tokenList.usdx.approve(address(spoke1), UINT256_MAX);
+      tokenList.dai.approve(address(spoke1), UINT256_MAX);
+      tokenList.wbtc.approve(address(spoke1), UINT256_MAX);
+      tokenList.usdy.approve(address(spoke1), UINT256_MAX);
+
+      tokenList.weth.approve(address(spoke2), UINT256_MAX);
+      tokenList.usdx.approve(address(spoke2), UINT256_MAX);
+      tokenList.dai.approve(address(spoke2), UINT256_MAX);
```
DhairyaSethi: run these in another for loop pls

DhairyaSethi: with an if cond when usdy is not listed to skip it

tests/Base.t.sol
```diff
@@ -2332,6 +2349,16 @@ abstract contract Base is Test {
     );
   }
 
+  function assertHubLiquidity(IHub targetHub, uint256 assetId, string memory label) internal view {
```
DhairyaSethi, line 2350: comment that at any point token.balanceOf(hub) == hub.getAssetLiquidity() + "any external donations"

we've run this invariant in most scenarios against all hub fn's right

tests/Utils.sol
```diff
@@ -202,4 +215,44 @@ library Utils {
...
+    ISpoke spoke,
+    uint256 reserveId,
+    address caller,
+    address from,
+    address to,
+    uint256 amount
+  ) internal {
+    IHub hub = IHub(address(spoke.getReserve(reserveId).hub));
+    _transferFrom(
+      IERC20(hub.getAsset(spoke.getReserve(reserveId).assetId).underlying),
```
DhairyaSethi:
```suggestion
      IERC20(spoke.getReserve(reserveId).underlying),
```

tests/Base.t.sol
DhairyaSethi: let's do it after spoke functions also pls im paranoid

src/hub/Hub.sol
DhairyaSethi: exactly, there. is still spoke state changes after the transfer (was this way before also bc hub transferred assets, returned shares, and then spoke does its internal state changes) 
-> we do not support erc777 or callback tokens

tests/unit/Spoke/Spoke.Repay.t.sol
Kogaroshi: I used the non-fuzz test for insufficient allowance, and the fuzz one for no allowance at all.
But can keep the current non-fuzz for insufficient allowance, and change the  fuzz one to be insufficient balance test, wdyt ?

tests/Base.t.sol
Kogaroshi: Added it after Spoke tests for : supply, withdraw, borrow, repay.
Other target functions we want to add it ?

DhairyaSethi: liquidation call (should be very ez, just add it to the base test contract which does _checkedLiquidationCall

update user rp & dyn config also lol

src/hub/Hub.sol
DhairyaSethi: we cannot do this as long as hub allows relisting the same underlying
we need to track liquidity per underlying (not assetId) for the balance check to fully reduce trust assumption

src/hub/Hub.sol
```diff
@@ -191,22 +191,24 @@ contract Hub is IHub, AccessManaged {
...
-  function add(uint256 assetId, uint256 amount, address from) external returns (uint256) {
+  function add(uint256 assetId, uint256 amount) external returns (uint256) {
     Asset storage asset = _assets[assetId];
     SpokeData storage spoke = _spokes[assetId][msg.sender];
 
     asset.accrue(_spokes, assetId);
-    _validateAdd(asset, spoke, amount, from);
+    _validateAdd(asset, spoke, amount);
 
+    uint256 newLiquidity = asset.liquidity + amount;
```
miguelmtzinf: if we apply #933 we must add `asset.feeAmount` too (which contains unrealized already)

DhairyaSethi: no we don't? it's not part of available liquidity

miguelmtzinf: it must be part of the balance of the contract, since `feeAmount` is held here. no?

Kogaroshi:

PR #776, opened 2025-09-17
test: POC liquidation with zero debt shares via donation

yan-man:
see test `test_liq_donation`
scenario:
- bob has a liquidatable position of HF=0.95
- debt index has been inflated to 10
- liquidator attacker liquidates some debt amt < 1 full share multiple times. The repay donation causes < 1 debt share being reduced, with > 1 coll share reduced each time
- by the end of the chained liqs, attacker has earned DAI (spending only gas) from LiqBonus, to put bob's user position in a more vulnerable position, by reducing collateral shares without affecting debt shares, reducing HF and putting user at higher risk of bad debt.
- however, this can only be done while position > $1k, ie dust threshold 

(took 100k liqCalls:)
Screenshot 2025-09-17 at 17 51 35

### DIFF
--- a/tests/unit/Spoke/Liquidations/Spoke.Liq.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.Liq.t.sol
@@ -47,8 +47,8 @@ contract SpokeLiqTest is SpokeBase {
   function test_liq_donation() public {
     uint256 desiredHF = 0.95e18;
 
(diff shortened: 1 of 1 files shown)


avniculae:
thanks for this @yan-man . In practice, debt index will be under 2, making this attack less feasible right? (since donation is less)

yan-man:
sorry missed your comment @avniculae - yes, exploitable with higher debt index, less feasible in practice. Yeah donation would always be < index

PR #777, opened 2025-09-18
fix: change `onlyPositionManager` modifier to internal function

yan-man:
- can save around 110B in spoke size, but adds a few gas per op

Screenshot 2025-09-17 at 21 33 52
Screenshot 2025-09-17 at 21 31 58


yan-man:
didnt like this change; didnt think contract size is critical factor atm

PR #789, opened 2025-09-22
fix: `_getReserveData` consistency

### LINKED ISSUE #780: inconsistent _getReserveData return for NativeTokenGateway / SignatureGateway
NativeTokenGateway
- _getReserveData returns address / address

SignatureGateway
- _getReserveData returns IERC20 / address

yan-man:
- update `_getReserveData` in `NativeTokenGateway` contract to have a consistent return type `(IERC20, address)` with that of `SignatureGateway`

closes #780

### DIFF
--- a/src/position-manager/NativeTokenGateway.sol
+++ b/src/position-manager/NativeTokenGateway.sol
@@ -47,7 +47,7 @@ contract NativeTokenGateway is
   /// @inheritdoc INativeTokenGateway
   function supplyNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
     (IERC20 underlying, address hub) = _getReserveData(reserveId);
-    _validateParams(address(underlying), amount);
+    _validateParams(underlying, amount);
     require(msg.value == amount, NativeAmountMismatch());
 
     _nativeWrapper.deposit{value: amount}();
@@ -58,7 +58,7 @@ contract NativeTokenGateway is
   /// @inheritdoc INativeTokenGateway
   function withdrawNative(uint256 reserveId, uint256 amount, address receiver) external {
     (IERC20 underlying, ) = _getReserveData(reserveId);
-    _validateParams(address(underlying), amount);
+    _validateParams(underlying, amount);
     require(receiver != address(0), InvalidAddress());
 
     uint256 withdrawAmount = MathUtils.min(
@@ -74,7 +74,7 @@ contract NativeTokenGateway is
   /// @inheritdoc INativeTokenGateway
   function borrowNative(uint256 reserveId, uint256 amount, address receiver) external {
     (IERC20 underlying, ) = _getReserveData(reserveId);
-    _validateParams(address(underlying), amount);
+    _validateParams(underlying, amount);
     require(receiver != address(0), InvalidAddress());
 
     _spoke.borrow(reserveId, amount, msg.sender);
@@ -85,7 +85,7 @@ contract NativeTokenGateway is
   /// @inheritdoc INativeTokenGateway
   function repayNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
     (IERC20 underlying, address hub) = _getReserveData(reserveId);
-    _validateParams(address(underlying), amount);
+    _validateParams(underlying, amount);
     require(msg.value == amount, NativeAmountMismatch());
 
     uint256 userDebtAmount = _spoke.getUserTotalDebt(reserveId, msg.sender);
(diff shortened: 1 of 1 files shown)


src/position-manager/SignatureGateway.sol
```diff
@@ -92,9 +92,9 @@ contract SignatureGateway is ISignatureGateway, Multicall, Rescuable, Ownable2St
     );
     require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
 
-    (IERC20 underlying, address hub) = _getReserveData(reserveId);
-    underlying.safeTransferFrom(onBehalfOf, address(this), amount);
-    underlying.forceApprove(hub, amount);
+    (address underlying, address hub) = _getReserveData(reserveId);
```
DhairyaSethi: let's have it return IERC20 instead on both

src/position-manager/SignatureGateway.sol
```diff
@@ -291,9 +291,9 @@ contract SignatureGateway is ISignatureGateway, Multicall, Rescuable, Ownable2St
     bytes32 r,
     bytes32 s
   ) external {
-    (IERC20 underlying, ) = _getReserveData(reserveId);
+    (address underlying, ) = _getReserveData(reserveId);
     try
-      IERC20Permit(address(underlying)).permit({
```
DhairyaSethi, line 296: using SafeErc20 for * would fix the double cast

src/position-manager/SignatureGateway.sol
yan-man: was considering that initially; any particular reason?

DhairyaSethi: it's much cleaner to not repeatedly cast constants

yan-man: https://github.com/aave/aave-v4/pull/789/commits/25ba950c9e3fe83d92246b9a7ad9fe36e115aaf2

src/position-manager/NativeTokenGateway.sol
```diff
@@ -46,8 +46,8 @@ contract NativeTokenGateway is
 
   /// @inheritdoc INativeTokenGateway
   function supplyNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
-    (address underlying, address hub) = _getReserveData(reserveId);
-    _validateParams(underlying, amount);
+    (IERC20 underlying, address hub) = _getReserveData(reserveId);
+    _validateParams(address(underlying), amount);
```
DhairyaSethi: pls cast down inside the validate fn

src/position-manager/SignatureGateway.sol
yan-man: can you clarify - still need to cast to IERC20Permit to call permit right?

src/position-manager/NativeTokenGateway.sol
yan-man: https://github.com/aave/aave-v4/pull/789/commits/1de988abc95a6f18a929ce7e1db07179e4ecf8b6

src/position-manager/NativeTokenGateway.sol
```diff
@@ -119,14 +119,14 @@ contract NativeTokenGateway is
...
+  function _validateParams(IERC20 underlying, uint256 amount) internal view {
+    require(address(underlying) == address(_nativeWrapper), InvalidReserveId());
     require(amount > 0, InvalidAmount());
   }
 
-  function _getReserveData(uint256 reserveId) internal view returns (address, address) {
+  function _getReserveData(uint256 reserveId) internal view returns (IERC20, address) {
     ISpoke.Reserve memory reserveData = _spoke.getReserve(reserveId);
-    return (reserveData.underlying, address(reserveData.hub));
+    return (IERC20(reserveData.underlying), address(reserveData.hub));
```
DhairyaSethi, line 129: now do you see why it was omitted bc it's casted up & down in this contract w/o any intermediate usage in casted up form 🙃

DhairyaSethi: hence i would close this PR but fine w merging if you prefer this for strict consistency

yan-man: i mean it does make sense, but just could be confusing given the internal methods are the same name and almost exactly the same except for this part

DhairyaSethi: on different type of position managers, they don't both inherit the same contract (yet)

Kogaroshi: I think it makes sense in the direction of having a common base for PositionManagers, in this specific NativeGateway it's more unnecessary casting than before but this should become an exception compared to future PositionManagers.

DhairyaSethi: exactly, that's why i would do this change then bc it mostly naturally comes out from the consolidation. it's ugly rn

PR #790, opened 2025-09-23
feat: Skip notifying risk premium when constant

avniculae:


src/spoke/Spoke.sol
```diff
@@ -382,7 +389,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     } else {
       // If unsetting, check HF and update user rp
       uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
-      _notifyRiskPremiumUpdate(onBehalfOf, newUserRiskPremium);
+      _notifyOrRefreshRiskPremium(onBehalfOf, newUserRiskPremium, reserveId);
```
yan-man: hm, does this one change the logic then - previously in setUsingAsCollateral it refreshes RP on all reserves, but if RP doesn't change, do we need to refresh on reserveId? Should this instead just be INVALID_RESERVE_ID

avniculae:
Consider skipping risk premium refresh if change is within a threshold as well, use #792 as inspiration

PR #792, opened 2025-09-23
feat: add rp threshold

avniculae:


avniculae:
This will act as a reference for future release https://github.com/aave/aave-v4/pull/790#issuecomment-3457570035

PR #793, opened 2025-09-24
[DRAFT] feat: Remove castToView

avniculae:


src/spoke/Spoke.sol
```diff
@@ -744,49 +744,94 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
 
       UserPosition storage userPosition = _userPositions[user][reserveId];
       Reserve storage reserve = _reserves[reserveId];
 
-      uint256 assetPrice = IAaveOracle(ORACLE).getReservePrice(reserveId);
-      uint256 assetUnit = uint256(10).uncheckedExp(reserve.decimals);
+      uint256 assetPrice;
+      uint256 assetUnit;
+
+      if (borrowing) {
```
miguelmtzinf, line 801: why you had to move the borrowing part up?

avniculae: small optimisation to avoid reading oracle price if reserve is not borrowed and does not have CF > 0

avniculae:
Closing, this does not scale well with number of reserves

PR #796, opened 2025-09-25
feat: Prevent collateral dust in liquidations

avniculae:
- prevent collateral dust in liquidations
- remove debtToCollateral interim variable in liquidations
- checks that collateral reserve was supplied by user in liquidation
- targeted tests for user account data
- scenario-based test for liquidations
- more fuzzing variations
- enables liquidation fee in gas tests for better gas reporting
- z3 prover script to make sure debtToLiquidate <= debtReserveBalance

prevents the accumulation of collateral dust in liquidations, which can pose a greater risk for bad debt accumulation by disincentivizing liquidations of remaining dust collateral that report deficit.

### DIFF
--- a/src/spoke/libraries/LiquidationLogic.sol
+++ b/src/spoke/libraries/LiquidationLogic.sol
@@ -244,10 +244,7 @@ library LiquidationLogic {
   }
 
   /// @notice Calculates the liquidation amounts.
-  /// @dev Invoked by `liquidateUser` method. To prevent accumulation of dust, it enforces one of the following conditions:
-  /// 1. liquidate all debt
-  /// 2. liquidate all collateral
-  /// 3. leave at least `DUST_LIQUIDATION_THRESHOLD` of collateral and debt (in value terms)
+  /// @dev Invoked by `liquidateUser` method.
   /// @return The collateral to liquidate.
   /// @return The collateral to transfer to liquidator.
   /// @return The debt to liquidate.
@@ -261,6 +258,10 @@ library LiquidationLogic {
       maxLiquidationBonus: params.maxLiquidationBonus
     });
 
+    // To prevent accumulation of dust, one of the following conditions is enforced:
+    // 1. liquidate all debt
+    // 2. liquidate all collateral
+    // 3. leave at least `DUST_LIQUIDATION_THRESHOLD` of collateral and debt (in value terms)
     uint256 debtToLiquidate = _calculateDebtToLiquidate(
       CalculateDebtToLiquidateParams({
         debtReserveBalance: params.debtReserveBalance,
@@ -347,7 +348,7 @@ library LiquidationLogic {
 
   /// @notice Calculates the debt that should be liquidated.
   /// @dev Generally, it returns the minimum of `debtToCover`, `debtReserveBalance` and `debtToTarget`.
-  /// If dust debt would be left behind, it returns `debtReserveBalance` to ensure the debt is fully cleared and no dust is left.
+  /// If debt dust would be left behind, it returns `debtReserveBalance` to ensure the debt is fully cleared and no dust is left.
   function _calculateDebtToLiquidate(
     CalculateDebtToLiquidateParams memory params
   ) internal pure returns (uint256) {
--- a/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Dust.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Dust.t.sol
@@ -108,7 +108,7 @@ contract SpokeLiquidationCallDustTest is SpokeLiquidationCallBaseTest {
     assertEq(spoke.getUserSuppliedAssets(_daiReserveId(spoke), alice), 0);
   }
 
-  /// @dev debtToCover would theoretically results in dust debt, but is allowed to proceed because collateral reserve was fully liquidated
+  /// @dev debtToCover would theoretically results in debt dust, but is allowed to proceed because collateral reserve was fully liquidated
   function test_debtToCover_exceeds_collateralValue() public {
     uint256 collateralFactor = 80_00;
     uint256 liquidationBonus = 124_00;
@@ -154,7 +154,7 @@ contract SpokeLiquidationCallDustTest is SpokeLiquidationCallBaseTest {
 
     // debtToTarget > debtToCover, so debtToTarget doesn't come into play
     assertGt(debtToTarget, debtToCover);
-    // debtToCover would result in dust debt
+    // debtToCover would result in debt dust
     uint256 theoreticalRemainingDebt = spoke.getUserTotalDebt(_usdxReserveId(spoke), alice) -
       debtToCover;
     assertLt(
@@ -175,7 +175,7 @@ contract SpokeLiquidationCallDustTest is SpokeLiquidationCallBaseTest {
     assertGt(spoke.getUserTotalDebt(_usdxReserveId(spoke), alice), theoreticalRemainingDebt);
   }
 
-  /// @dev dust debt allowed if all collateral is liquidated
+  /// @dev debt dust allowed if all collateral is liquidated
   function test_dustDebt_allowed() public {
     uint256 collateralFactor = 80_00;
     uint256 liquidationBonus = 124_00;


tests/mocks/MockSpoke.sol
```diff
@@ -3,11 +3,26 @@
 pragma solidity ^0.8.0;
 
 import {Spoke, ISpoke, IHubBase, SafeCast, PositionStatusMap} from 'src/spoke/Spoke.sol';
+import {Test} from 'forge-std/Test.sol';
 
-contract MockSpoke is Spoke {
+/// @dev inherit from Test to exclude contract from forge size check
```
DhairyaSethi, line 8: lol nice one, i think we can also just add 
bool public constant IS_TEST = true; 
bc i think that's the signal to forge that this is a test contract

tests/misc/debtToLiquidate.py
```diff
@@ -0,0 +1,65 @@
...
+def mulDivUp(a, num, den):
+    return (a * num + den - 1) / den
+
+s = Solver()
+
+debtAssetPrice = Int('debtAssetPrice')
+s.add(1 <= debtAssetPrice, debtAssetPrice <= 10**30)
+debtAssetDecimals = Int('debtAssetDecimals')
+s.add(1 <= debtAssetDecimals, debtAssetDecimals <= 18)
+debtAssetUnit = Int('debtAssetUnit')
```
DhairyaSethi:
```suggestion
debtAssetUnit = 10**debtAssetDecimals 
```
same with `collateralAssetUnit`, they dont need to be symbolic, two less vars to solve, nbd since this runs in time

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -118,19 +119,23 @@ library LiquidationLogic {
     ISpoke.DynamicReserveConfig storage collateralDynConfig,
     LiquidateUserParams memory params
   ) external returns (bool) {
-    IHubBase collateralHub = collateralReserve.hub;
```
avniculae, line 134: removed this bc of stack too deep. I think it's fine to read it twice bc compiler is optimising it, wdyt @DhairyaSethi ?

src/spoke/Spoke.sol
```diff
@@ -43,8 +43,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     LiquidationLogic.HEALTH_FACTOR_LIQUIDATION_THRESHOLD;
 
   /// @inheritdoc ISpoke
-  uint256 public constant DUST_DEBT_LIQUIDATION_THRESHOLD =
-    LiquidationLogic.DUST_DEBT_LIQUIDATION_THRESHOLD;
+  uint256 public constant DUST_LIQUIDATION_THRESHOLD = LiquidationLogic.DUST_LIQUIDATION_THRESHOLD;
```
yan-man, line 50: hm wondering if we need this in spoke?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -118,19 +119,23 @@ library LiquidationLogic {
...
+    uint256 collateralReserveBalance = collateralReserve.hub.previewRemoveByShares(
+      collateralReserve.assetId,
+      collateralPosition.suppliedShares
+    );
     _validateLiquidationCall(
```
yan-man, line 139: ah maybe it's a bigger change and not sure if it's worth it, but maybe we can split into 2 validations so that we can exit earlier. 1 set of validation is for non-HF related (ie flags, CF, balances), then validate HF afterwards, so that we can save a calcUserAccountData if the first validation fails

yan-man: something like this: https://github.com/aave/aave-v4/pull/859/files#diff-c2b8fc3739c2b0dd34b9513f5f01b4c4a225b524c9b34a4a1fa355105f994c47

ahh but maybe too late in the game and cost benefit too low given the breaking tests and edge case

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -249,25 +252,46 @@ library LiquidationLogic {
...
+      DUST_LIQUIDATION_THRESHOLD;
+
+    if (
+      collateralToLiquidate > params.collateralReserveBalance ||
+      (leavesCollateralDust && debtToLiquidate < params.debtReserveBalance)
+    ) {
       collateralToLiquidate = params.collateralReserveBalance;
-      debtToCollateral = collateralToLiquidate.percentDivUp(liquidationBonus);
+
+      // decreases if collateralToLiquidate > params.collateralReserveBalance
```
yan-man:
```suggestion
      // debtToLiquidate is decreased if collateralToLiquidate > params.collateralReserveBalance
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -249,25 +252,46 @@ library LiquidationLogic {
...
+      //  - if it decreases, it might bypass the debt dust condition
```
yan-man:
```suggestion
      //  - if so, debt dust could remain
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -249,25 +252,46 @@ library LiquidationLogic {
...
+      // increases if leavesCollateralDust && debtToLiquidate < params.debtReserveBalance
```
yan-man:
```suggestion
      // debtToLiquidate is increased if (leavesCollateralDust && debtToLiquidate < params.debtReserveBalance), ensuring collateral reserve is fully liquidated
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -249,25 +252,46 @@ library LiquidationLogic {
...
+      //  - if it increases, it will increase by at most DUST_LIQUIDATION_THRESHOLD (in base currency).
+      //    Since debtToLiquidate < params.debtReserveBalance and debt dust condition was enforced, it
+      //    is guaranteed that the increase will not make debtToLiquidate exceed params.debtReserveBalance
```
yan-man: `- can only increase by at most DUST_LIQUIDATION_THRESHOLD (in base currency)`
` Since debt dust condition was enforced, it
  is guaranteed that debtToLiquidate will never exceed params.debtReserveBalance`

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -249,25 +252,46 @@ library LiquidationLogic {
...
+      //  - target health factor is ignored to prevent leaving dust
```
yan-man:
```suggestion
      //  - target health factor is bypassed to prevent leaving dust
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -249,25 +252,46 @@ library LiquidationLogic {
...
       debtToLiquidate = collateralToLiquidate.mulDivUp(
         params.collateralAssetPrice * params.debtAssetUnit * PercentageMath.PERCENTAGE_FACTOR,
         params.debtAssetPrice * params.collateralAssetUnit * liquidationBonus
       );
     }
 
+    // make sure the liquidator intends to cover the necessary debt
+    // in case the amount was adjusted due to dust conditions
```
yan-man:
```
revert if the liquidator does not cover the necessary debt to prevent dust from remaining
```

src/spoke/libraries/LiquidationLogic.sol
yan-man: or rm/edit below as it's the same comment

src/spoke/libraries/LiquidationLogic.sol
avniculae: I agree, it would be great, but there are a few challenges including spoke size (which is exceeding max size in the PR you linked)

tests/misc/debtToLiquidate.py
avniculae: [ffabc76](https://github.com/aave/aave-v4/pull/796/commits/ffabc7672da3e9a23b0169f3f07b01d2811083fe)

src/spoke/Spoke.sol
avniculae: I think it's good to expose it

src/spoke/libraries/LiquidationLogic.sol
avniculae: [e0ffb03](https://github.com/aave/aave-v4/pull/796/commits/e0ffb032d96381aa72d859e3b1ebd6a38294ec46)

src/spoke/libraries/LiquidationLogic.sol
avniculae: [e0ffb03](https://github.com/aave/aave-v4/pull/796/commits/e0ffb032d96381aa72d859e3b1ebd6a38294ec46)

src/spoke/libraries/LiquidationLogic.sol
avniculae: [e0ffb03](https://github.com/aave/aave-v4/pull/796/commits/e0ffb032d96381aa72d859e3b1ebd6a38294ec46)

src/spoke/libraries/LiquidationLogic.sol
avniculae: [e0ffb03](https://github.com/aave/aave-v4/pull/796/commits/e0ffb032d96381aa72d859e3b1ebd6a38294ec46)

src/spoke/libraries/LiquidationLogic.sol
avniculae: [e0ffb03](https://github.com/aave/aave-v4/pull/796/commits/e0ffb032d96381aa72d859e3b1ebd6a38294ec46)

src/spoke/libraries/LiquidationLogic.sol
avniculae: [e0ffb03](https://github.com/aave/aave-v4/pull/796/commits/e0ffb032d96381aa72d859e3b1ebd6a38294ec46)

src/spoke/Spoke.sol
DhairyaSethi: yes definitely good to expose, it's still accessible from Spoke.getLiquidationLogic().DUST_THRESHOLD() but it's fine as is. if we have code size issues we can consider removing it here from since it's still accessible from elsewhere

src/spoke/interfaces/ISpoke.sol
```diff
@@ -473,8 +476,9 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   function HEALTH_FACTOR_LIQUIDATION_THRESHOLD() external view returns (uint64);
 
   /// @notice Returns the minimum required remaining base currency amount after a partial liquidation.
-  /// @return The minimum debt amount considered as dust, denominated in USD with 26 decimals.
-  function DUST_DEBT_LIQUIDATION_THRESHOLD() external view returns (uint256);
+  /// @dev Applicable to both collateral and debt.
+  /// @return The minimum amount considered as dust, denominated in USD with 26 decimals.
```
yan-man: actually i think this is "max" amt considered as dust

src/spoke/libraries/LiquidationLogic.sol
yan-man: yeah just an opt too, so not really necessary

src/spoke/interfaces/ISpoke.sol
avniculae: [945b6a9](https://github.com/aave/aave-v4/pull/796/commits/945b6a9c66fcb4dae50facbbbb7998823ddac5b1)

tests/misc/debtToLiquidate.py
```diff
@@ -0,0 +1,63 @@
...
+s.add(1 <= collateralAssetPrice, collateralAssetPrice <= 10**30)
+collateralAssetDecimals = Int('collateralAssetDecimals')
+s.add(1 <= collateralAssetDecimals, collateralAssetDecimals <= 18)
+collateralAssetUnit = 10**collateralAssetDecimals
+
+liquidationBonus = Int('liquidationBonus')
+s.add(PERCENTAGE_FACTOR <= liquidationBonus, liquidationBonus < PERCENTAGE_FACTOR * PERCENTAGE_FACTOR)
+
+debtReserveBalance = Int('debtReserveBalance')
+s.add(0 <= debtReserveBalance, debtReserveBalance <= 10**30)
```
yan-man, line 33: should be strictly < debtReserveBalance?

tests/misc/debtToLiquidate.py
```diff
@@ -0,0 +1,63 @@
...
+debtToLiquidate = Int('debtToLiquidate')
+s.add(0 <= debtToLiquidate, debtToLiquidate <= debtReserveBalance)
+
+collateralReserveBalance = Int('collateralReserveBalance')
+s.add(0 <= collateralReserveBalance, collateralReserveBalance <= 10**30)
```
yan-man, line 38: strictly < ?

tests/misc/debtToLiquidate.py
```diff
@@ -0,0 +1,63 @@
...
+debtAssetPrice = Int('debtAssetPrice')
+s.add(1 <= debtAssetPrice, debtAssetPrice <= 10**30)
+debtAssetDecimals = Int('debtAssetDecimals')
+s.add(1 <= debtAssetDecimals, debtAssetDecimals <= 18)
+debtAssetUnit = 10**debtAssetDecimals
+
+collateralAssetPrice = Int('collateralAssetPrice')
+s.add(1 <= collateralAssetPrice, collateralAssetPrice <= 10**30)
+collateralAssetDecimals = Int('collateralAssetDecimals')
+s.add(1 <= collateralAssetDecimals, collateralAssetDecimals <= 18)
```
yan-man, line 26: nit: can tighten up decimals bounds now to 6? the min now in place

tests/misc/debtToLiquidate.py
avniculae: does not matter right?

tests/misc/debtToLiquidate.py
avniculae: same as above, does not matter right?

tests/misc/debtToLiquidate.py
avniculae: sure, we can, but there's no harm in having a bigger range here

src/spoke/interfaces/ISpoke.sol
```diff
@@ -472,9 +475,10 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @return The minimum health factor considered healthy, expressed in WAD (18 decimals) (e.g. 1e18 is 1.00).
   function HEALTH_FACTOR_LIQUIDATION_THRESHOLD() external view returns (uint64);
 
-  /// @notice Returns the minimum required remaining base currency amount after a partial liquidation.
-  /// @return The minimum debt amount considered as dust, denominated in USD with 26 decimals.
-  function DUST_DEBT_LIQUIDATION_THRESHOLD() external view returns (uint256);
+  /// @notice Returns the maximum amount considered as dust after a liquidation.
+  /// @dev Applicable to both collateral and debt.
```
```suggestion
  /// @notice Returns the maximum amount considered as dust for a user's collateral and debt balances after a liquidation.
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -269,25 +272,42 @@ library LiquidationLogic {
...
+      collateralToLiquidate > params.collateralReserveBalance ||
+      (leavesCollateralDust && debtToLiquidate < params.debtReserveBalance)
+    ) {
       collateralToLiquidate = params.collateralReserveBalance;
-      debtToCollateral = collateralToLiquidate.percentDivUp(liquidationBonus);
+
+      // - debtToLiquidate is decreased if collateralToLiquidate > params.collateralReserveBalance (if so, debt dust could remain).
+      // - debtToLiquidate is increased if (leavesCollateralDust && debtToLiquidate < params.debtReserveBalance), ensuring collateral reserve
+      //   is fully liquidated (potentially bypassing the target health factor). Can only increase by at most DUST_LIQUIDATION_THRESHOLD (in
+      //   value terms). Since debt dust condition was enforced, it is guaranteed that debtToLiquidate will never exceed params.debtReserveBalance.
```
```suggestion
      // - `debtToLiquidate` is decreased if `collateralToLiquidate > params.collateralReserveBalance` (if so, debt dust could remain).
      // - `debtToLiquidate` is increased if `(leavesCollateralDust && debtToLiquidate < params.debtReserveBalance)`, ensuring collateral reserve
      //   is fully liquidated (potentially bypassing the target health factor). Can only increase by at most `DUST_LIQUIDATION_THRESHOLD` (in
      //   value terms). Since debt dust condition was enforced, it is guaranteed that `debtToLiquidate` will never exceed `params.debtReserveBalance`.
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -269,25 +272,42 @@ library LiquidationLogic {
...
+      //   is fully liquidated (potentially bypassing the target health factor). Can only increase by at most `DUST_LIQUIDATION_THRESHOLD` (in
+      //   value terms). Since debt dust condition was enforced, it is guaranteed that `debtToLiquidate` will never exceed `params.debtReserveBalance`.
       debtToLiquidate = collateralToLiquidate.mulDivUp(
         params.collateralAssetPrice * params.debtAssetUnit * PercentageMath.PERCENTAGE_FACTOR,
         params.debtAssetPrice * params.collateralAssetUnit * liquidationBonus
       );
     }
 
+    // revert if the liquidator does not cover the necessary debt to prevent dust from remaining
+    require(params.debtToCover >= debtToLiquidate, ISpoke.MustNotLeaveDust());
```
miguelmtzinf, line 308: why is this always needed? in a simple liquidation (no dust in any side) we don't need to check this holds?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -347,14 +367,15 @@ library LiquidationLogic {
       maxDebtToLiquidate = debtToTarget;
     }
 
-    uint256 remainingDebtValue = (params.debtReserveBalance - maxDebtToLiquidate).mulDivDown(
-      params.debtAssetPrice.toWad(),
-      params.debtAssetUnit
-    );
+    bool leavesDebtDust = maxDebtToLiquidate < params.debtReserveBalance &&
+      (params.debtReserveBalance - maxDebtToLiquidate).mulDivDown(
```
miguelmtzinf: rounding up here and rounding down in `debtToLiquidate` re-calculation later would be safer, so rounding issues don't make the `debtToLiquidate` be greater that `debtReserveBalance`

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -322,14 +345,15 @@ library LiquidationLogic {
       );
   }
 
-  /// @notice Calculates the maximum debt that can be liquidated.
-  /// @dev Dust debt less than the `DUST_DEBT_LIQUIDATION_THRESHOLD` cannot be left behind, unless the collateral reserve is fully liquidated.
-  function _calculateMaxDebtToLiquidate(
-    CalculateMaxDebtToLiquidateParams memory params
+  /// @notice Calculates the debt that should be liquidated.
+  /// @dev Generally, it returns the minimum of `debtToCover`, `debtReserveBalance` and `debtToTarget`.
+  /// If it would result in dust debt being left behind, the function returns `debtReserveBalance`.
```
```suggestion
  /// If dust debt would be left behind, it returns `debtReserveBalance` to ensure the debt is fully cleared and no dust is left.
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -237,11 +239,15 @@ library LiquidationLogic {
       params.isUsingAsCollateral && params.collateralFactor > 0,
       ISpoke.CollateralCannotBeLiquidated()
     );
+    require(params.collateralReserveBalance > 0, ISpoke.ReserveNotSupplied());
     require(params.debtReserveBalance > 0, ISpoke.ReserveNotBorrowed());
   }
 
   /// @notice Calculates the liquidation amounts.
-  /// @dev Invoked by `liquidateUser` method.
+  /// @dev Invoked by `liquidateUser` method. To prevent accumulation of dust, it enforces one of the following conditions:
```
miguelmtzinf: I'm fine moving the comment about dust down, right before `_calculateDebtToLiquidate`

src/spoke/libraries/LiquidationLogic.sol
avniculae: I agree it would fee safer bc we round down `debtToLiquidate`, but what I don't like about it is the fact that we would round `debtToLiquidate` differently is we enter the if branch or not. We want to round up the `debtToLiquidate` to help the user. Also, when assessing dust you want to round down to increase the chances of hitting the dust condition (which would also be different than collateral dust assessment rounding)

Personally I am fine with the current rounding given the confidence brought by the z3 prover

src/spoke/libraries/LiquidationLogic.sol
avniculae: [d2e1836](https://github.com/aave/aave-v4/pull/796/commits/d2e18363539d5ba38aa18c0c942dca3c0c83bcf0)

src/spoke/libraries/LiquidationLogic.sol
avniculae: yes, but at this point we don't know if debt to liquidate increased due to debt or collateral dust

StErMi: without this check it's possible (if we are in the dust debt scenario) that the liquidator would pay more than it is willing to pay (only possible if the allowance approved by the liquidator is `> params.debtToCover`)

PR #801, opened 2025-09-27
fix: Notify after updating dynamic config

CheyenneAtapour:
Calculate, update, and broadcast user risk premium after the user updates their dynamic config


src/spoke/Spoke.sol
```diff
@@ -401,6 +401,8 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
       _checkCanCall(msg.sender, msg.data);
     }
     _refreshDynamicConfig(onBehalfOf);
+    uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
+    _notifyRiskPremiumUpdate(onBehalfOf, newUserRiskPremium);
```
DhairyaSethi, line 400:
```suggestion
    _notifyRiskPremiumUpdate(onBehalfOf, _calculateUserAccountData(onBehalfOf).userRiskPremium);
```

hf validation isn't needed
also pls add some tests for this

CheyenneAtapour: we're ok to make a user put themselves in a liquidatable position by updating their own dynamic config?

DhairyaSethi: yes we are, we don't want to block this action

CheyenneAtapour: Decided it's fine as is

src/spoke/Spoke.sol
```diff
@@ -396,6 +396,8 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
       _checkCanCall(msg.sender, msg.data);
     }
     _refreshDynamicConfig(onBehalfOf);
+    uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
```
DhairyaSethi, line 399: 398 is not needed, 399 already refreshes dynamic config 

we can now rm the fn 398 completely too i think

CheyenneAtapour: Done

.github/workflows/dependency-review.yml
DhairyaSethi, line 1: can you pls do `git reset --soft origin/dev` and then commit + force push pls? nbd either way

narnona:
Would this be inconsistent with the idea in https://github.com/aave/aave-v4/pull/1037? The user just wants to update the dynamic config, but in this way, they would need to consider whether the premium has improved.

miguelmtzinf:
> Would this be inconsistent with the idea in #1037? The user just wants to update the dynamic config, but in this way, they would need to consider whether the premium has improved.

Good q @narnona ,

This function directly affects the user configuration, so it makes sense to apply the latest risk premium configuration to the user’s position when updating the dynamic config to the latest version (also, the set of assets with a given collateral category can change).

The core idea of refreshing the premium only when it clearly worsens is the strategy applied across all user actions that do not explicitly trigger configuration updates (such as `updateUserRiskPremium` and `updateUserDynamicConfig`).

PR #802, opened 2025-09-27
fix: prevent CF=0 on historical config key

### LINKED ISSUE #800: Prevent historical config keys from being set with CF=0
- doing so will lead to not being able to liquidate the debt positions attached to that historical key, as it reverts with `CollateralCannotBeLiquidated`
- instead, block historical config keys from being set to CF=0
- offboarding will have to be gradual for prior config keys, and CF gradually decreased

yan-man:
- prevent `collateralFactor` of historical config keys to be set to 0, as CF=0 reserves cannot be liquidated (reverts on `_validateLiquidationCall`)

closes #800

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -177,7 +177,6 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
   ) external restricted {
     require(reserveId < _reserveCount, ReserveNotListed());
     _validateHistoricDynamicReserveConfig(_dynamicConfig[reserveId][configKey], dynamicConfig);
-    _validateDynamicReserveConfig(dynamicConfig);
     _dynamicConfig[reserveId][configKey] = dynamicConfig;
     emit UpdateDynamicReserveConfig(reserveId, configKey, dynamicConfig);
   }
@@ -696,6 +695,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     // @dev sufficient check since maxLiquidationBonus is always >= 100_00
     require(currentConfig.maxLiquidationBonus != 0, ConfigKeyUninitialized());
     require(newConfig.collateralFactor > 0, InvalidCollateralFactor());
+    _validateDynamicReserveConfig(newConfig);
   }
 
   function _validateDynamicReserveConfig(DynamicReserveConfig calldata config) internal pure {


src/spoke/Spoke.sol
```diff
@@ -181,6 +181,8 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
       _dynamicConfig[reserveId][configKey].maxLiquidationBonus != 0,
       ConfigKeyUninitialized()
     );
+    // @dev CF of historical config keys cannot be 0, otherwise liquidations of prior positions will revert
+    require(dynamicConfig.collateralFactor > 0, InvalidCollateralFactor());
```
DhairyaSethi: pls merge this and the above require in a helper _validate method. also i would remove liquidations will revert line on the comment for this

src/spoke/Spoke.sol
```diff
@@ -689,6 +691,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
   }
 
   function _validateDynamicReserveConfig(DynamicReserveConfig calldata config) internal pure {
+    // @dev sufficient check since maxLiquidationBonus is always >= 100_00
     // Enforce that at moment loan is taken, there should be enough collateral to cover liquidation
```
DhairyaSethi, line 702: merge comments

tests/unit/Spoke/Spoke.Borrow.Scenario.t.sol
```diff
@@ -598,16 +599,6 @@ contract SpokeBorrowScenarioTest is SpokeBase {
 
     ISpoke.UserAccountData memory userAccountData = spoke1.getUserAccountData(alice);
     assertEq(_calculateExpectedUserRP(alice, spoke1), userAccountData.userRiskPremium);
-    assertEq(
```
DhairyaSethi, line 601: why rmed?

src/spoke/Spoke.sol
miguelmtzinf: please move the comment to the function natspec. (and add it to the helper validate method, if added)

src/spoke/interfaces/ISpoke.sol
```diff
@@ -186,6 +186,11 @@ interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
   error MustNotLeaveDust();
   error InvalidDebtToCover();
 
+  /**
+   * @notice Emitted when the collateral factor is invalid during update of historical dynamic risk config.
```
```suggestion
   * @dev Emitted when the collateral factor is invalid during update of historical dynamic risk config.
```

tests/unit/Spoke/Spoke.Borrow.Scenario.t.sol
yan-man: rm bc in this test, it starts with coll1 as non zero, does these assertions, and then changes coll1 CF=0, and checks the assertions again but now with coll1 not contributing.

but now, we can't set coll1 to CF=0 historically, so can't change it mid-way through. 

so instead altered the test to just immediately set new CF=0 and from the get-go coll1 does not contribute

src/spoke/interfaces/ISpoke.sol
yan-man: https://github.com/aave/aave-v4/pull/802/commits/7664b8ff9505fdfea232e5c9977134a4a1c27941

src/spoke/Spoke.sol
yan-man: https://github.com/aave/aave-v4/pull/802/commits/7664b8ff9505fdfea232e5c9977134a4a1c27941

src/spoke/Spoke.sol
yan-man: actually this was not meant to be added here, rm'ed it

src/spoke/Spoke.sol
```diff
@@ -176,11 +176,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     DynamicReserveConfig calldata dynamicConfig
   ) external restricted {
     require(reserveId < _reserveCount, ReserveNotListed());
-    // @dev sufficient check since maxLiquidationBonus is always >= 100_00
-    require(
-      _dynamicConfig[reserveId][configKey].maxLiquidationBonus != 0,
-      ConfigKeyUninitialized()
-    );
+    _validateHistoricDynamicReserveConfig(_dynamicConfig[reserveId][configKey], dynamicConfig);
```
DhairyaSethi, line 179: make dyn config storage pointer and reuse for assignment on 181

src/spoke/Spoke.sol
```diff
@@ -688,6 +684,20 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     require(config.collateralRisk <= MAX_ALLOWED_COLLATERAL_RISK, InvalidCollateralRisk());
   }
 
+  /**
+   * @dev CollateralFactor of historical config keys cannot be 0.
+   * @param currentConfig The current dynamic reserve config.
+   * @param newConfig The dynamic reserve config to validate.
```
DhairyaSethi, line 689: param comments not really needed, not useful here. the dev comment can be inside. 
we can decide on consistency on location of dev comment in the natspec pr

src/spoke/interfaces/ISpoke.sol
```diff
@@ -186,6 +186,11 @@ interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
   error MustNotLeaveDust();
   error InvalidDebtToCover();
 
+  /**
+   * @dev Emitted when the collateral factor is invalid during update of historical dynamic risk config.
```
DhairyaSethi:
```suggestion
   * @dev Thrown when trying to set zero collateralFactor on historic dynamic configuration keys.
```
reverts thrown, events emitted

src/spoke/Spoke.sol
yan-man: Miguel wanted to add it to function natspec: https://github.com/aave/aave-v4/pull/802#discussion_r2387939858

src/spoke/Spoke.sol
yan-man: wanted to do this initially but cannot set the storage pointer to `dynamicConfig` as it's a mapping, fails on compilation

DhairyaSethi: _dynamicConfig[reserveId][configKey] is the dynamic config struct, so can be referenced w storage pointer, no?

src/spoke/interfaces/ISpoke.sol
yan-man: yeah just mixed it up here: https://github.com/aave/aave-v4/pull/802/commits/db365e60f59af1a94a1e0827b0554f32a2742b52

src/spoke/Spoke.sol
yan-man: this fails (if this was your suggestion):
```
DynamicReserveConfig storage currentConfig = _dynamicConfig[reserveId][configKey];
 currentConfig = dynamicConfig;
```

DhairyaSethi: it shouldn't? ('dynamicConfig' is just a normal simple struct in calldata tho, _dynamicConfig is a mapping)

if _dynamicConfig[reserveId][configKey] = dynamicConfig works

DhairyaSethi: ah ok it wouldn't work

src/spoke/Spoke.sol
DhairyaSethi, line 689:
```suggestion
```

yan-man: would like to keep this for consistency for now and can rm from internal methods as needed when we settle on the convention, in the natspec PR

src/spoke/Spoke.sol
```diff
@@ -176,11 +176,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     DynamicReserveConfig calldata dynamicConfig
   ) external restricted {
     require(reserveId < _reserveCount, ReserveNotListed());
-    // @dev sufficient check since maxLiquidationBonus is always >= 100_00
-    require(
-      _dynamicConfig[reserveId][configKey].maxLiquidationBonus != 0,
-      ConfigKeyUninitialized()
-    );
+    _validateHistoricDynamicReserveConfig(_dynamicConfig[reserveId][configKey], dynamicConfig);
     _validateDynamicReserveConfig(dynamicConfig);
```
DhairyaSethi: i think 179 can just internally invoke 180

yan-man: did this bc i figured we wanted to minimize internal chained calls but sure np
https://github.com/aave/aave-v4/pull/802/commits/eae23b56841543c93fb0500a1166f64acbcef773

PR #813, opened 2025-09-30
fix: Use uncheckedExp

CheyenneAtapour:
* Makes use of uncheckedExp for gas savings and simplicity where appropriate


src/hub/Hub.sol
```diff
@@ -688,7 +688,8 @@ contract Hub is IHub, AccessManaged {
     uint256 addCap = spoke.addCap;
     require(
       addCap == MAX_ALLOWED_SPOKE_CAP ||
-        addCap * 10 ** asset.decimals >= previewAddByShares(assetId, spoke.addedShares) + amount,
+        addCap * MathUtils.uncheckedExp(10, asset.decimals) >=
```
DhairyaSethi, line 691: not consistent with how we're doing in spoke

PR #823, opened 2025-09-30
feat: Cap debt amount to liquidate after recomputation

### LINKED ISSUE #822: The amount of liquidated debt is not explicitly capped after recomputation
If collateral reserve balance is not sufficient to cover the liquidated debt, then we seize all collateral and recompute the amount of debt that needs to be repaid. Due to roundings, there might be a possibility that the new `debtToLiquidate` exceeds the debt reserve balance. If this happens, liquidation might revert with underflow when updating user position's drawn shares. Note that due to donations on the hub side when restoring debt, liquidation might succeed even though `debtToLiquidate` exceeds the debt reserve balance (restored drawn shares are rounded down).

https://github.com/aave/aave-v4/blob/f2f0b66b0a918093327dcd2ecdc16648892984ba/src/spoke/libraries/LiquidationLogic.sol#L257-L264

While a set of numbers to trigger this scenario was not yet identified, we can simply protect against this by making sure the recomputed `debtToLiquidate` amount does not exceed the debt reserve balance (if it does, we can just scale it down to the debt reserve balance)

avniculae:
Closes #822

### DIFF
--- a/src/spoke/libraries/LiquidationLogic.sol
+++ b/src/spoke/libraries/LiquidationLogic.sol
@@ -257,10 +257,12 @@ library LiquidationLogic {
     if (collateralToLiquidate > params.collateralReserveBalance) {
       collateralToLiquidate = params.collateralReserveBalance;
       debtToCollateral = collateralToLiquidate.percentDivUp(liquidationBonus);
-      debtToLiquidate = debtToCollateral.mulDivUp(
-        params.collateralAssetPrice * params.debtAssetUnit,
-        params.debtAssetPrice * params.collateralAssetUnit
-      );
+      debtToLiquidate = debtToCollateral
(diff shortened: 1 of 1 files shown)


yan-man:
impl lgtm, probably ideal but maybe difficult to get to a scenario to write a test where this value has been exceeded and is adjusted back to debtReserveBalance

miguelmtzinf:
This seems not neccesary if https://github.com/aave/aave-v4/pull/821 is applied. We plan to prove that it's impossible to end up with more than `debtReserveBalance` so we can discard this PR.

avniculae:
Proved using z3 in #796 that `debtToLiquidate` cannot exceed `debtReserveBalance`

PR #827, opened 2025-09-30
fix: Change drawn function

CheyenneAtapour:
Note: This does add gas, so consider not adding


DhairyaSethi:
decided against this to save the minor gas

PR #830, opened 2025-10-01
Fix: Skip hf check on withdraw non-collateral

CheyenneAtapour:
When withdrawing an asset that is not set as collateral, do not check health factor, and do not update and broadcast risk premium

### DIFF
--- a/tests/gas/Spoke.Operations.gas.t.sol
+++ b/tests/gas/Spoke.Operations.gas.t.sol
@@ -103,7 +103,7 @@ contract SpokeOperations_Gas_Tests is SpokeBase {
     skip(100);
 
     spoke.withdraw(reserveId.weth, UINT256_MAX, alice);
-    vm.snapshotGasLastCall(NAMESPACE, 'withdraw: non collateral, full');
+    vm.snapshotGasLastCall(NAMESPACE, 'withdraw: non collateral');
     vm.stopPrank();
   }
 
--- a/snapshots/Spoke.Operations.ZeroRiskPremium.json
+++ b/snapshots/Spoke.Operations.ZeroRiskPremium.json
@@ -28,5 +28,5 @@
   "withdraw: 0 borrows, partial": "138153",
   "withdraw: 1 borrow, partial": "226946",
   "withdraw: 2 borrows, partial": "310296",
-  "withdraw: non collateral, full": "135610"
+  "withdraw: non collateral": "135610"
 }
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -28,5 +28,5 @@
   "withdraw: 0 borrows, partial": "138153",
   "withdraw: 1 borrow, partial": "240946",
   "withdraw: 2 borrows, partial": "338296",
-  "withdraw: non collateral, full": "135610"
+  "withdraw: non collateral": "135610"
 }


src/spoke/Spoke.sol
```diff
@@ -227,8 +227,13 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
 
     userPosition.suppliedShares -= withdrawnShares.toUint128();
 
-    uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
-    _notifyRiskPremiumUpdate(onBehalfOf, newUserRiskPremium);
+    if (
+      this.isUsingAsCollateral(reserveId, onBehalfOf) &&
```
DhairyaSethi: pls use position status mapping directly instead of making self call

src/spoke/Spoke.sol
```diff
@@ -227,8 +227,13 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
+      _dynamicConfig[reserveId][userPosition.configKey].collateralFactor > 0
+    ) {
+      uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
```
DhairyaSethi: this is checking prev dynamic config, but on withdraw we're supposed to use latest dynamic config

and anyway, i don't think cf > 0 check is needed

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
```diff
@@ -296,6 +296,65 @@ contract SpokeWithdrawScenarioTest is SpokeBase {
...
+
+    skip(3560 days);
+
+    // Position is underwater
+    ISpoke.UserAccountData memory userData = spoke1.getUserAccountData(bob);
+    assertLt(userData.healthFactor, 1e18, 'hf below 1');
+
+    vm.startPrank(bob);
+    // Can still withdraw usdx because CF = 0
+    spoke1.withdraw(usdxReserveId, type(uint256).max, bob);
```
DhairyaSethi: foundry constant for uint max

src/spoke/Spoke.sol
CheyenneAtapour: But if user is using as collateral and CF == 0, then can't we skip hf check?

src/spoke/Spoke.sol
CheyenneAtapour: I made this function public, no reason for it not to be

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
CheyenneAtapour: Done

src/spoke/Spoke.sol
DhairyaSethi: no, bc earlier it could have been > 0 but now isn't. if it's 0 both times then yes, but then anyway the hf check should pass (barring stuff due to time like price & yield) 
it's too narrow a case and we can avoid

src/spoke/Spoke.sol
```diff
@@ -467,6 +472,10 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     {} catch {}
   }
 
+  function isUsingAsCollateral(uint256 reserveId, address user) public view returns (bool) {
```
DhairyaSethi: pls revert let's put it back with isBorrowing

src/spoke/Spoke.sol
```diff
@@ -228,8 +228,13 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
 
     userPosition.suppliedShares -= withdrawnShares.toUint128();
 
-    uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
-    _notifyRiskPremiumUpdate(onBehalfOf, newUserRiskPremium);
+    if (
+      isUsingAsCollateral(reserveId, onBehalfOf) &&
+      _dynamicConfig[reserveId][reserve.dynamicConfigKey].collateralFactor > 0
```
```suggestion
      _positionStatus[user].isUsingAsCollateral(reserveId) 
```
or can use pub method, but cf check is incorrect

yan-man: you're saying to rm cf check completely?

DhairyaSethi: yes

src/spoke/Spoke.sol
```diff
@@ -228,8 +228,13 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
+    ) {
+      uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
+      _notifyRiskPremiumUpdate(onBehalfOf, newUserRiskPremium);
```
yan-man, line 233: was the plan to also merge _refreshAndValidateUserPosition and _notifyRiskPremiumUpdate into a single func in this PR, or at all?

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -116,12 +116,16 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
...
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 1);
     vm.expectRevert(ISpoke.HealthFactorBelowThreshold.selector);
     vm.prank(alice);
     spoke1.withdraw(_usdxReserveId(spoke1), 500e6, alice);
 
-    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), _randomBps());
+    updateCollateralFactor(
+      spoke1,
+      _usdxReserveId(spoke1),
+      vm.randomUint(1, PercentageMath.PERCENTAGE_FACTOR)
```
yan-man: i think should be strictly < PERCENTAGE_FACTOR

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
```diff
@@ -296,6 +296,65 @@ contract SpokeWithdrawScenarioTest is SpokeBase {
...
+    // Bob borrows weth
+    Utils.borrow({
+      spoke: spoke1,
+      reserveId: wethReserveId,
+      caller: bob,
+      amount: 2e18,
+      onBehalfOf: bob
+    });
+
+    skip(3560 days);
```
yan-man, line 332: 10 years on purpose?

src/spoke/Spoke.sol
CheyenneAtapour: Not in this PR, I think the goal should still be to keep these PRs minimal to show auditors, but we can do that in a separate refactor pr

tests/unit/Spoke/Spoke.Withdraw.Scenario.t.sol
CheyenneAtapour: Yeah, 1 year not enough to make position underwater, 10 years felt cleaner as a round number

src/spoke/Spoke.sol
```diff
@@ -228,8 +228,10 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
 
     userPosition.suppliedShares -= withdrawnShares.toUint128();
 
-    uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
-    _notifyRiskPremiumUpdate(onBehalfOf, newUserRiskPremium);
+    if (_positionStatus[onBehalfOf].isUsingAsCollateral(reserveId)) {
```
DhairyaSethi, line 231: can you add snapshot scenario for withdraw non collateral - gas should be close to withdraw 0 borrows

tests/gas/Spoke.Operations.gas.t.sol
```diff
@@ -98,6 +98,12 @@ contract SpokeOperations_Gas_Tests is SpokeBase {
 
     spoke.withdraw(reserveId.usdx, 1e6, alice);
     vm.snapshotGasLastCall(NAMESPACE, 'withdraw: 2 borrows, partial');
+    spoke.supply(reserveId.weth, 1000e18, alice);
+
+    skip(100);
+
+    spoke.withdraw(reserveId.weth, UINT256_MAX, alice);
+    vm.snapshotGasLastCall(NAMESPACE, 'withdraw: non collateral, full');
```
DhairyaSethi:
```suggestion
    vm.snapshotGasLastCall(NAMESPACE, 'withdraw: non collateral');
```

PR #831, opened 2025-10-01
fix: If cf 0 enforce remaining dyn config 0

CheyenneAtapour:


tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -5,18 +5,28 @@ pragma solidity ^0.8.0;
...
+
   function test_supply_does_not_trigger_dynamicConfigUpdate() public {
     DynamicConfig[] memory configs = _getUserDynConfigKeys(spoke1, alice);
 
     Utils.supplyCollateral(spoke1, _usdxReserveId(spoke1), alice, 1000e6, alice);
-    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), _randomBps());
+    updateCollateralFactor(
+      spoke1,
+      _usdxReserveId(spoke1),
+      vm.randomUint(50_00, PercentageMath.PERCENTAGE_FACTOR).toUint16()
```
CheyenneAtapour, line 17: This is future proofing because when you borrow, dynamic config is updated, and there is a hf check (this is being enforced in another PR)

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -71,7 +88,11 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
 
     skip(123 days);
 
-    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 80_00);
+    config.collateralFactor = 80_00;
+    config.maxLiquidationBonus = 105_00;
+    config.liquidationFee = 50;
```
miguelmtzinf, line 93: let's store previous values in memory so we avoid hardcoding pls

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
miguelmtzinf: why `50_00`? we are adding DAI collateral later so we can still borrow no?

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -90,12 +111,16 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
     configs = _getUserDynConfigKeys(spoke1, alice);
     skip(322 days);
 
-    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 0);
+    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 1);
```
miguelmtzinf, line 114: why

tests/unit/Spoke/Spoke.DynamicConfig.Triggers.t.sol
```diff
@@ -55,7 +65,14 @@ contract SpokeDynamicConfigTriggersTest is SpokeBase {
...
 
     // usdx (user coll) is offboarded
-    updateCollateralFactor(spoke1, _usdxReserveId(spoke1), 0);
+    ISpoke.DynamicReserveConfig memory config = ISpoke.DynamicReserveConfig({
+      collateralFactor: 0,
+      maxLiquidationBonus: 0,
+      liquidationFee: 0
+    });
+    vm.prank(SPOKE_ADMIN);
+    spoke1.addDynamicReserveConfig(_usdxReserveId(spoke1), config);
```
miguelmtzinf, line 74:
```suggestion
    vm.prank(SPOKE_ADMIN);
    spoke1.addDynamicReserveConfig(_usdxReserveId(spoke1), ISpoke.DynamicReserveConfig({
      collateralFactor: 0,
      maxLiquidationBonus: 0,
      liquidationFee: 0
    }));
```
we can also create a helper if needed

miguelmtzinf:
I was initially leaning towards adding this check, since overriding a historical dynamic config key (exceeding the limit) could potentially result in mixing values (using both old and new at the same time). However, this assumption is incorrect, because we explicitly assign values to all parameters of the dynamic config, which makes it impossible for old values to persist.
Therefore, this change is not completely necessary (we can add anyway, nbd).

CheyenneAtapour:
Decided against this, too confusing for any of the potential value add

PR #832, opened 2025-10-01
feat: Implement getRoundData in UnitPriceFeed

CheyenneAtapour:
* Implements the `getRoundData` function in the `UnitPriceFeed` contract

### DIFF
--- a/tests/unit/UnitPriceFeed.t.sol
+++ b/tests/unit/UnitPriceFeed.t.sol
@@ -30,7 +30,7 @@ contract UnitPriceFeedTest is Base {
   }
 
   function test_getRoundData() public {
-    uint80 _roundId = uint80(vm.randomUint(0, UINT256_MAX));
+    uint80 _roundId = uint80(vm.randomUint());
     (
       uint80 roundId,
       int256 answer,


tests/unit/UnitPriceFeed.t.sol
```diff
@@ -29,9 +29,26 @@ contract UnitPriceFeedTest is Base {
...
+    ) = unitPriceFeed.getRoundData(roundId);
+    assertEq(_roundId, roundId);
+    assertEq(answer, int256(10 ** _decimals));
+    assertEq(startedAt, roundId);
+    assertEq(updatedAt, roundId);
+    assertEq(answeredInRound, roundId);
+  }
+
+  function test_getRoundData_revertsWith_InvalidRoundId() public {
+    uint80 roundId = uint80(vm.randomUint(block.timestamp + 1, type(uint80).max));
```
yan-man:
```suggestion
    uint80 roundId = uint80(vm.randomUint(vm.getBlockTimestamp() + 1, type(uint80).max));
```

tests/unit/UnitPriceFeed.t.sol
```diff
@@ -29,9 +29,26 @@ contract UnitPriceFeedTest is Base {
     assertEq(unitPriceFeed.version(), 1);
   }
 
-  function test_getRoundData_revertsWith_OperationNotSupported() public {
-    vm.expectRevert(UnitPriceFeed.OperationNotSupported.selector);
-    unitPriceFeed.getRoundData(0);
+  function test_getRoundData() public {
+    uint80 roundId = uint80(vm.randomUint(0, block.timestamp));
```
yan-man: also vm getBlockTimestamp

tests/unit/UnitPriceFeed.t.sol
```diff
@@ -29,9 +29,26 @@ contract UnitPriceFeedTest is Base {
...
+    (
+      uint80 _roundId,
+      int256 answer,
+      uint256 startedAt,
+      uint256 updatedAt,
+      uint80 answeredInRound
+    ) = unitPriceFeed.getRoundData(roundId);
```
yan-man: nit: set the getRoundData input as _roundId instead of the output, to match the impl

tests/unit/UnitPriceFeed.t.sol
CheyenneAtapour: Done

tests/unit/UnitPriceFeed.t.sol
CheyenneAtapour: Done

tests/unit/UnitPriceFeed.t.sol
CheyenneAtapour: Done

tests/unit/UnitPriceFeed.t.sol
```diff
@@ -29,9 +29,20 @@ contract UnitPriceFeedTest is Base {
     assertEq(unitPriceFeed.version(), 1);
   }
 
-  function test_getRoundData_revertsWith_OperationNotSupported() public {
-    vm.expectRevert(UnitPriceFeed.OperationNotSupported.selector);
-    unitPriceFeed.getRoundData(0);
+  function test_getRoundData() public {
+    uint80 _roundId = uint80(vm.randomUint(0, UINT256_MAX));
```
DhairyaSethi:
```suggestion
    uint80 _roundId = uint80(vm.randomUint());
```
exactly the same as well, it doesn't need ranges

src/misc/UnitPriceFeed.sol
```diff
@@ -41,9 +41,23 @@ contract UnitPriceFeed is AggregatorV3Interface {
...
+    view
+    returns (
+      uint80 roundId,
+      int256 answer,
+      uint256 startedAt,
+      uint256 updatedAt,
+      uint80 answeredInRound
+    )
+  {
+    roundId = _roundId;
```
DhairyaSethi, line 56: i still think this should revert when round id is in the future

CheyenneAtapour: How do we enforce that though? I thought my previous require statement was incorrect

yan-man: haha yeah I was thinking that as well... but ig doesnt matter either way, it WILL be still `_units` in the future so not really wrong. This one is kind of subjective, I prefer reverting but ok either way

DhairyaSethi: it's more about properties for price feeds

CheyenneAtapour: I mean, what is the line of code I should add here to make it revert?

yan-man: lol it would literally be the same revert line you removed earlier, which Miguel thought we didnt need, so dont know if we should be bringing that back

miguelmtzinf: this is a constant price feed, which will return 1 unit in the future as well. I think we do not have a proper way to check if round id is in the future, and even if we had it, why do we care since we know it's constant? I'm okay with how it is now.

DhairyaSethi: bc someone could be relying on price feed reverts for future queries property 


that being said, i'm ok moving forward here as is

PR #833, opened 2025-10-01
feat: 2d nonce

DhairyaSethi:
Adds keyed-nonces (2d nonces) to support parallel sig-based actions

### DIFF
--- a/snapshots/Spoke.Operations.ZeroRiskPremium.json
+++ b/snapshots/Spoke.Operations.ZeroRiskPremium.json
@@ -1,8 +1,8 @@
 {
   "borrow: first": "223061",
   "borrow: second action, same reserve": "203633",
-  "liquidationCall: full": "310782",
-  "liquidationCall: partial": "331244",
+  "liquidationCall: full": "311217",
+  "liquidationCall: partial": "331679",
   "permitReserve + repay (multicall)": "263718",
   "permitReserve + supply (multicall)": "145146",
   "permitReserve + supply + enable collateral (multicall)": "179124",
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -1,8 +1,8 @@
 {
   "borrow: first": "248561",
   "borrow: second action, same reserve": "212033",
-  "liquidationCall: full": "319398",
-  "liquidationCall: partial": "339860",
+  "liquidationCall: full": "319833",
+  "liquidationCall: partial": "340295",
   "permitReserve + repay (multicall)": "272334",
   "permitReserve + supply (multicall)": "145146",
   "permitReserve + supply + enable collateral (multicall)": "179124",
--- a/src/spoke/libraries/LiquidationLogic.sol
+++ b/src/spoke/libraries/LiquidationLogic.sol
@@ -253,13 +253,16 @@ library LiquidationLogic {
       params.debtAssetPrice * params.collateralAssetUnit,
       params.debtAssetUnit * params.collateralAssetPrice
     );
-    uint256 collateralToLiquidate = debtToCollateral.percentMulDown(liquidationBonus);
+    uint256 collateralToLiquidate = debtToLiquidate.mulDivDown(
+      params.debtAssetPrice * params.collateralAssetUnit * liquidationBonus,
+      params.debtAssetUnit * params.collateralAssetPrice * PercentageMath.PERCENTAGE_FACTOR
+    );
     if (collateralToLiquidate > params.collateralReserveBalance) {
       collateralToLiquidate = params.collateralReserveBalance;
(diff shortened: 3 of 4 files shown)


src/utils/NoncesKeyed.sol
```diff
@@ -0,0 +1,51 @@
...
+  /// @inheritdoc INoncesKeyed
+  function useNonce(uint192 key) external {
+    _useNonce(msg.sender, key);
+  }
+
+  /// @inheritdoc INoncesKeyed
+  function nonces(address owner, uint192 key) external view returns (uint256) {
+    return _pack(key, _nonces[owner][key]);
+  }
+
```
Kogaroshi, line 22: Since we are not keeping the exact OZ version, might be good to also have function `useNonce()` & function `nonces(address owner)` where `key = 0` not to break compatibility with some contracts not managing KeyedNonces ?
Or we consider it a non-issue ?

DhairyaSethi: not needed, no contract is doing it, i think it's tech debt on OZ's part also

src/utils/NoncesKeyed.sol
```diff
@@ -0,0 +1,51 @@
...
+  /// @notice Consumes the next unused nonce for an address and key.
+  /// @dev Returns the current value without the key prefix. Consumed nonce is increased, so calling this function twice
```
miguelmtzinf: is this wrong? as we are returning with prefix

src/utils/NoncesKeyed.sol
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /// @dev Same as `_useNonce` but checking that `nonce` is the next valid for `owner`.
+  function _useCheckedNonce(address owner, uint256 keyNonce) internal {
+    (uint192 key, ) = _unpack(keyNonce);
+    uint256 current = _useNonce(owner, key);
+    require(keyNonce == current, InvalidAccountNonce(owner, current));
+  }
+
+  /// @dev Pack key and nonce into a keyNonce.
```
miguelmtzinf, line 42: does it make sense to say we are using ony one word, using the first 192 bits for key, next 64 for sequential nonce

src/interfaces/INoncesKeyed.sol
```diff
@@ -0,0 +1,18 @@
...
+
+  /// @notice Returns the next unused nonce for an address and key. Result contains the key prefix.
+  /// @param owner The address of the nonce over.
+  /// @param key The nonceKey, specifies lifetime of the nonce.
+  /// @return keyNonce The first 24 bytes are for the key, & the last 8 bytes for the nonce.
+  function nonces(address owner, uint192 key) external view returns (uint256 keyNonce);
+
+  /// @notice Allows caller to revoke their next sequential nonce at specified `key`.
+  /// @param key The nonceKey, specifies lifetime of the nonce.
+  function useNonce(uint192 key) external;
```
miguelmtzinf: move up

src/spoke/interfaces/ISpoke.sol
```diff
@@ -12,7 +13,7 @@ import {ISpokeBase} from 'src/spoke/interfaces/ISpokeBase.sol';
  * @author Aave Labs
  * @notice Full interface for Spoke
  */
-interface ISpoke is ISpokeBase, IMulticall, IAccessManaged {
+interface ISpoke is ISpokeBase, IMulticall, IAccessManaged, INoncesKeyed {
```
```suggestion
interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
```
just to follow same order as implementation

src/utils/NoncesKeyed.sol
DhairyaSethi: ah damn missed this, good call, tysm

src/utils/NoncesKeyed.sol
DhairyaSethi: yep feels a bit misleading in the sense we still use a double mapping to store, so we don't have an `owner => uint256 keyNonce` mapping but instead have `owner => uint192 key => uint64 nonce` so it's never stored as a word
i didn't remove the double mapping to not change logic from OZ but it would be cheaper gas wise as well

src/interfaces/INoncesKeyed.sol
```diff
@@ -7,8 +7,9 @@ interface INoncesKeyed {
   error InvalidAccountNonce(address account, uint256 currentNonce);
 
   /// @notice Allows caller to revoke their next sequential nonce at specified `key`.
+  /// @dev This does not invalidate nonce at other `key` sessions.
   /// @param key The key which specifies lifetime of the nonce.
```
miguelmtzinf: i d suggest to use one name for all (lifetime, session and namespace). I'm good with namespace personally but no strong opinion

DhairyaSethi: sounds good, namespace it is

PR #834, opened 2025-10-01
fix: Check oracle decimals in spoke

### LINKED ISSUE #824: Make sure spoke oracle uses 8 decimals
The spoke assumes an 8-decimal oracle:
https://github.com/aave/aave-v4/blob/f2f0b66b0a918093327dcd2ecdc16648892984ba/src/spoke/libraries/LiquidationLogic.sol#L109

We need to make sure the oracle always uses 8 decimals on every new upgrade of the spoke

avniculae:
Enforces on the spoke that the oracle provides prices with 8 decimals

Closes #824

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -70,6 +70,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
    */
   constructor(address oracle_) {
     require(oracle_ != address(0), InvalidAddress());
+    require(IAaveOracle(oracle_).DECIMALS() == 8, InvalidOracleDecimals());
     ORACLE = oracle_;
   }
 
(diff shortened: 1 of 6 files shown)


src/spoke/Spoke.sol
```diff
@@ -70,6 +70,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
    */
   constructor(address oracle_) {
     require(oracle_ != address(0), InvalidAddress());
+    require(IAaveOracle(oracle_).DECIMALS() == 8, InvalidOracleDecimals());
```
DhairyaSethi, line 73: zero address check not needed anymore since we call on it

StErMi: I agree with that unless you want to have more precise errors.

PR #836, opened 2025-10-01
fix: spoke cleanup

DhairyaSethi:
- update `!=0` checks to `>0` for consistency
- check Spoke oracle decimals matches defined constant during intialization
- safecast
- renaming, comments for clarity
- validate non-zero hub address in AssetInterestRateStrategy contract
- update `Spoke.getReserveTotalDebt` logic with call to the `Hub.getSpokeTotalOwed` 
- rename `list` local variable used in `_calculateAndPotentiallyRefreshUserAccountData` to `collateralInfo`
- rename `drawnDebtRestored`, `premiumDebtRestored` variables in the `_reportDeficit` function


src/spoke/Spoke.sol
```diff
@@ -824,9 +820,8 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
           reserve.assetId,
           userPosition
         );
-        userAccountData.totalDebtInBaseCurrency +=
-          (drawnDebt * assetPrice).wadDivUp(assetUnit) +
-          (premiumDebt * assetPrice).wadDivUp(assetUnit);
+        userAccountData.totalDebtInBaseCurrency += ((drawnDebt + premiumDebt) * assetPrice)
+          .wadDivUp(assetUnit);
```
DhairyaSethi: there is no precision loss here as long as we are 18 scaled currency, hence this is okay, & this saves 200 gas

src/hub/Hub.sol
```diff
@@ -436,9 +436,9 @@ contract Hub is IHub, AccessManaged {
   /// @inheritdoc IHub
   function getSpokeConfig(
     uint256 assetId,
-    address spoke
+    address spokeAddr
   ) external view returns (SpokeConfig memory) {
-    SpokeData storage spoke = _spokes[assetId][spoke];
+    SpokeData storage spoke = _spokes[assetId][spokeAddr];
```
yan-man: nit: would prob prefer arg address as spoke and var as spokeData or something, given the arg addr is external facing

src/hub/interfaces/IAssetInterestRateStrategy.sol
```diff
@@ -47,6 +47,11 @@ interface IAssetInterestRateStrategy is IBasicInterestRateStrategy {
     uint32 variableRateSlope2;
   }
 
+  /**
+   * @notice Thrown when the given address is invalid.
```
yan-man, line 51: should we specify when hub address is invalid zero addr or something more specific

src/hub/Hub.sol
DhairyaSethi: we use `spoke` var everywhere, think it's better this way for consistency

src/hub/interfaces/IAssetInterestRateStrategy.sol
DhairyaSethi: this is only for constructor, not end user facing so fine is as. we could fashion existing error, `OnlyHub` to something like `InvalidHub` and reuse at both places

src/spoke/Spoke.sol
```diff
@@ -64,13 +67,10 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     _;
   }
 
-  /**
-   * @dev Constructor.
-   * @param oracle_ The address of the AaveOracle contract.
-   */
+  /// @dev Constructor.
+  /// @param oracle_ The address of `AaveOracle` which supports existing reserves on contract upgrades.
```
yan-man: wondering why to specify the upgrades part

src/spoke/Spoke.sol
```diff
@@ -645,27 +641,14 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
     require(address(reserve.hub) != address(0), ReserveNotListed());
     require(!reserve.paused, ReservePaused());
   }
 
-  /**
-   * @dev Calculates the user's premium debt offset in assets amount from a given share amount.
-   * @dev Rounds down to the nearest assets amount. Uses the opposite rounding direction of the
-   * debt shares-to-assets conversion to prevent underflow in premium debt.
-   */
-  function _previewPremiumOffset(
```
yan-man, line 663: just rm in favor of calling previewDrawByShares directly, was that the rationale?

src/hub/Hub.sol
miguelmtzinf: i d prefer the opposite and name this one spokeConfig

src/spoke/Spoke.sol
DhairyaSethi: bc of a really good inline comment, it's much cleaner that way. this never needed such verbose natspec

src/spoke/Spoke.sol
DhairyaSethi: stermi keeps insisting

src/hub/Hub.sol
DhairyaSethi: we use `spoke` var everywhere, think it's better this way for consistency

src/hub/interfaces/IAssetInterestRateStrategy.sol
yan-man: for natspec, error fine as is

src/spoke/Spoke.sol
miguelmtzinf: don't think this comment belongs to here or the contract... more to docs/instructions for upgrades. So i d just say: The address of the AaveOracle

src/spoke/Spoke.sol
```diff
@@ -856,12 +841,12 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     uint256 debtCounterInBaseCurrency = userAccountData.totalDebtInBaseCurrency;
     uint256 collateralCounterInBaseCurrency = 0;
 
-    list.sortByKey(); // sort by collateral risk
+    collateralInfo.sortByKey(); // sort by collateral risk
```
miguelmtzinf: sort by collateral risk in ASC order, collateral value in DESC order

src/spoke/Spoke.sol
DhairyaSethi: stermi keeps insisting on this point, @yan-man let's find a place for it in the natspec then if possible? rming from here since you also didn't like it (right?)

src/spoke/Spoke.sol
```diff
@@ -366,17 +368,16 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
 
-    _validateSetUsingAsCollateral(_reserves[reserveId], usingAsCollateral);
-
     positionStatus.setUsingAsCollateral(reserveId, usingAsCollateral);
 
     if (usingAsCollateral) {
       _refreshDynamicConfig(onBehalfOf, reserveId);
     } else {
-      // If unsetting, check HF and update user rp
       uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates HF
```
yan-man:
```suggestion
      uint256 newUserRiskPremium = _refreshAndValidateUserPosition(onBehalfOf); // validates health factor
```

src/spoke/Spoke.sol
yan-man: a bit confused here, I was just talking about the `_previewPremiumOffset` function being removed completely vs the natspec part

edit: i see you mean we can have an inline comment to explain the rounding rather than keeping this separate method

src/spoke/Spoke.sol
```diff
@@ -691,18 +675,18 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
     // @dev sufficient check since maxLiquidationBonus is always >= 100_00
-    require(currentConfig.maxLiquidationBonus != 0, ConfigKeyUninitialized());
+    require(currentConfig.maxLiquidationBonus > 0, ConfigKeyUninitialized());
     require(newConfig.collateralFactor > 0, InvalidCollateralFactor());
     _validateDynamicReserveConfig(newConfig);
   }
 
   function _validateDynamicReserveConfig(DynamicReserveConfig calldata config) internal pure {
-    // Enforce that at moment loan is taken, there should be enough collateral to cover liquidation
+    // there should be enough collateral to cover liquidation at moment a loan is taken
```
yan-man: maybe we can re-write to "debt is taken" as we dont use loan anywhere else

src/spoke/Spoke.sol
```diff
@@ -729,7 +713,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     require(!usingAsCollateral || !reserve.frozen, ReserveFrozen());
   }
 
-  // @dev allows donation on drawn debt
+  /// @dev Allows donation on drawn debt.
```
yan-man: did we want to rm references to donation?

DhairyaSethi: can rm, or we can do consistently in natspec pr

src/spoke/Spoke.sol
DhairyaSethi: it just did one thing, our idea before about having it was to showcase the explicit rounding with shitty natspec, i think the debt owed to protocol analogy is good and we don't need this one fn abstraction anymore

src/spoke/Spoke.sol
```diff
@@ -905,28 +889,26 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
-        assetId,
-        userPosition.premiumShares
-      ).toUint128());
+      Reserve storage reserve = _reserves[reserveId];
+      uint256 assetId = reserve.assetId;
+      IHubBase hub = reserve.hub;
+
+      uint256 oldPremiumShares = userPosition.premiumShares;
+      uint256 oldPremiumOffset = userPosition.premiumOffset;
+      uint256 accruedUserPremium = hub.previewRestoreByShares(assetId, oldPremiumShares) -
```
yan-man: for consistency if rm "user" from some of these, then this should also just be accruedPremium

src/spoke/Spoke.sol
```diff
@@ -948,30 +930,27 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
-        drawnDebtRestored,
-        premiumDebtRestored,
+        drawnDebtReported,
+        premiumDebtReported,
         premiumDelta
       );
       _settlePremiumDebt(userPosition, premiumDelta.realizedDelta);
       userPosition.drawnShares -= deficitShares.toUint128();
-      // newUserRiskPremium is 0 due to no collateral remaining
-      // non-zero deficit means user ends up with zero total debt
```
yan-man, line 977: can rm first comment, but 2nd line was to emphasize that any deficit in a single reserve cleans out their debt as deficit across all remaining debt reserves, perhaps can keep

src/spoke/Spoke.sol
DhairyaSethi: removed instead since it should be intuitive from `validateUserPosition`, lmk wdyt

src/spoke/Spoke.sol
yan-man: ok will see if we can add to natspec PR somewhere, yeah correct didnt like it

src/spoke/Spoke.sol
DhairyaSethi: yes good catch, tysm

src/spoke/Spoke.sol
```diff
@@ -856,15 +842,13 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     uint256 debtCounterInBaseCurrency = userAccountData.totalDebtInBaseCurrency;
     uint256 collateralCounterInBaseCurrency = 0;
 
-    list.sortByKey(); // sort by collateral risk
+    collateralInfo.sortByKey(); // sort by collateralRisk in asc, collateralValue in desc
```
yan-man: dont think `collateralValue` should be combined here, implying it's a variable or term. perhaps either `userCollateralInBaseCurrency` or `collateral amount` or `collateral value` (but imo dont really like value) or something else

src/spoke/Spoke.sol
yan-man: also works

src/spoke/Spoke.sol
```diff
@@ -366,18 +368,17 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     address onBehalfOf
   ) external onlyPositionManager(onBehalfOf) {
     PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
+    _validateSetUsingAsCollateral(_reserves[reserveId], usingAsCollateral);
+
```
```suggestion
    _validateSetUsingAsCollateral(_reserves[reserveId], usingAsCollateral);

    PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
```

src/spoke/Spoke.sol
DhairyaSethi: personally think it's fine, don't see confusion with collateral amount here specifically 

can make them collateral risk & collateral value with space to signal it's not a specific var 

open to suggestions, full var name for base currency is tediously long

PR #840, opened 2025-10-01
feat: improve packing

DhairyaSethi:
re-order asset struct to group more commonly read vars together, saves ~2k gas on average (scales linearly with notify, ie borrowed reserves)

has to reduce timestamp bits from 40 to 32 -> protocol can work until ~2106-02-07, ie 80 years


src/hub/interfaces/IHub.sol
```diff
@@ -25,18 +25,18 @@ interface IHub is IHubBase, IAccessManaged {
...
     address underlying;
     //
     address irStrategy;
-    uint96 drawnRate;
     //
     address reinvestmentController;
     //
     address feeReceiver;
+    uint16 liquidityFee;
+    uint8 decimals;
```
miguelmtzinf, line 39: Tried with other alternatives (packing these two with `underlying` or `irStrategy`) and this seems to be most efficient one

PR #841, opened 2025-10-02
fix: Change premium function

CheyenneAtapour:


CheyenneAtapour:
We ultimately decided against this change as it adds gas, and doesn't provide much perceived benefit for code readability

PR #842, opened 2025-10-02
fix: Enforce minimum decimals on asset

CheyenneAtapour:
* Enforces a hardcoded minimum asset decimals (5) allowed to be listed on the hub

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -61,7 +61,7 @@ contract Hub is IHub, AccessManaged {
       InvalidAddress()
     );
     require(
-      decimals >= MIN_ALLOWED_UNDERLYING_DECIMALS && decimals <= MAX_ALLOWED_UNDERLYING_DECIMALS,
+      MIN_ALLOWED_UNDERLYING_DECIMALS <= decimals && decimals <= MAX_ALLOWED_UNDERLYING_DECIMALS,
       InvalidAssetDecimals()
     );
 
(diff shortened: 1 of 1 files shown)


src/hub/Hub.sol
```diff
@@ -372,7 +378,8 @@ contract Hub is IHub, AccessManaged {
     asset.accrue(assetId, _spokes[assetId][asset.feeReceiver]);
     _validateSweep(asset, msg.sender, amount);
 
-    asset.liquidity -= amount.toUint128();
+    uint256 liquidity = asset.liquidity;
+    asset.liquidity = liquidity.uncheckedSub(amount).toUint128();
```
DhairyaSethi: pointless & very bad for readability. i had mentioned caching it only if it was being used twice, here there's no point to caching it 


also we don't care about gas on sweep and reclaim, let's revert this change

src/hub/Hub.sol
```diff
@@ -57,7 +60,10 @@ contract Hub is IHub, AccessManaged {
       underlying != address(0) && feeReceiver != address(0) && irStrategy != address(0),
       InvalidAddress()
     );
-    require(decimals <= MAX_ALLOWED_UNDERLYING_DECIMALS, InvalidAssetDecimals());
+    require(
+      decimals >= MIN_ALLOWED_UNDERLYING_DECIMALS && decimals <= MAX_ALLOWED_UNDERLYING_DECIMALS,
+      InvalidAssetDecimals()
+    );
```
DhairyaSethi, line 66:
```suggestion
    require(
      MIN_ALLOWED_UNDERLYING_DECIMALS <= decimals && decimals <= MAX_ALLOWED_UNDERLYING_DECIMALS,
      InvalidAssetDecimals()
    );
```
👻

PR #844, opened 2025-10-02
feat: block refreshPremium when spoke is frozen

DhairyaSethi:


DhairyaSethi:
needs skip notify on repay AND skip notify when rp is same optimisations

miguelmtzinf:
This implementation restricts even more the usage of `Hub.refreshPremium`, which negatively impacts the user experience by increasing the likelihood of breaking user positions (when one of the assets in the user’s position is deactivated or has a zero drawCap).
Closing in favor of implementing a less restrictive approach: https://github.com/aave/aave-v4/pull/882

PR #845, opened 2025-10-02
fix : Spoke instance correct init check

Kogaroshi:


src/spoke/instances/SpokeInstance.sol
```diff
@@ -17,12 +17,13 @@ contract SpokeInstance is Spoke {
...
   /// @inheritdoc Spoke
-  function initialize(address _authority) external override reinitializer(SPOKE_REVISION) {
+  function initialize(address _authority) public override reinitializer(SPOKE_REVISION) {
     require(_authority != address(0), InvalidAddress());
     __AccessManaged_init(_authority);
     if (_liquidationConfig.targetHealthFactor == 0) {
       _liquidationConfig.targetHealthFactor = HEALTH_FACTOR_LIQUIDATION_THRESHOLD;
       emit UpdateLiquidationConfig(_liquidationConfig);
     }
+    super.initialize(_authority);
```
miguelmtzinf, line 27: I do not have a strong opinion but I think the whole thing is a bit overkill. I d remove the check and the if condition.
This is because Base and Instance is the same contract, and whenever there is a new version of the Spoke we are checking what's initialized before and what's not, and explicity do it here. Every upgrade would be a different thing / have a different init code.

src/spoke/instances/SpokeInstance.sol
```diff
@@ -17,12 +17,13 @@ contract SpokeInstance is Spoke {
   }
 
   /// @inheritdoc Spoke
-  function initialize(address _authority) external override reinitializer(SPOKE_REVISION) {
+  function initialize(address _authority) public override reinitializer(SPOKE_REVISION) {
```
miguelmtzinf, line 20: realizing we don't have a getter for SPOKE_REVISION in any interface. Should we create it or overkill?

Kogaroshi: It would be in a new ISpokeInstance.sol, not sure it's really needed for just a getter ?

src/spoke/instances/SpokeInstance.sol
Kogaroshi: Yeah, it's not really enforceable there, new init logic could remove the `super` call and the check is then irrelevant.

miguelmtzinf:
The sole purpose of `SpokeInstance` is to handle initialization logic (primarily setting the version and assigning values to existing and newly introduced variables).
Each deployment of a `SpokeInstance` may include different logic, which makes it impractical to introduce complex conditions or interactions between `SpokeInstance` and `Spoke`.

PR #846, opened 2025-10-02
Add reserve and asset existence checks in Spoke and Hub

miguelmtzinf:
- Add `reserveId` existence check to all functions in Spoke.
- Add `assetId` existence check only to write functions in Hub for better developer experience.
- The `assetId` existence check is not included in all Hub functions, because they are not strictly necessary (returned values are consistent) and increase gas costs. They are considered not required since the functions in question are not user-facing but are instead intended for integration by Spokes.

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -492,8 +492,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
   /// @inheritdoc ISpokeBase
   function getUserDebt(uint256 reserveId, address user) external view returns (uint256, uint256) {
-    UserPosition storage userPosition = _userPositions[user][reserveId];
     Reserve storage reserve = _getReserve(reserveId);
+    UserPosition storage userPosition = _userPositions[user][reserveId];
     (uint256 drawnDebt, uint256 premiumDebt, ) = _getUserDebt(
       reserve.hub,
       reserve.assetId,
@@ -504,8 +504,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
   /// @inheritdoc ISpokeBase
   function getUserTotalDebt(uint256 reserveId, address user) external view returns (uint256) {
-    UserPosition storage userPosition = _userPositions[user][reserveId];
     Reserve storage reserve = _getReserve(reserveId);
+    UserPosition storage userPosition = _userPositions[user][reserveId];
     (uint256 drawnDebt, uint256 premiumDebt, ) = _getUserDebt(
       reserve.hub,
       reserve.assetId,


tests/unit/misc/SignatureGateway/SignatureGateway.Reverts.ReserveNotListed.t.sol
miguelmtzinf, line 1: Fine removing the entire file as this is now handled by the Spoke itself.

src/spoke/Spoke.sol
```diff
@@ -452,7 +452,7 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
     bytes32 s
   ) external {
     Reserve storage reserve = _reserves[reserveId];
-    address underlying = reserve.underlying;
+    address underlying = address(reserve.underlying);
```
yan-man: do we need to cast? i think `reserve.underlying` is already an address

tests/unit/Hub/Hub.Config.t.sol
```diff
@@ -66,6 +66,14 @@ contract HubConfigTest is HubBase {
...
 
+  function test_updateSpokeConfig_revertsWith_AssetNotListed() public {
+    uint256 assetId = _randomInvalidAssetId(hub1);
+    address spoke = vm.randomAddress();
+    IHub.SpokeConfig memory spokeConfig;
+    vm.expectRevert(IHub.AssetNotListed.selector);
+    Utils.updateSpokeConfig(hub1, ADMIN, assetId, spoke, spokeConfig);
+  }
+
   function test_updateSpokeConfig_fuzz_revertsWith_SpokeNotListed(
```
yan-man, line 77: getting a failing test on this: test_updateSpokeConfig_fuzz_revertsWith_SpokeNotListed, like in CI

miguelmtzinf: resolved now

src/position-manager/interfaces/INativeTokenGateway.sol
```diff
@@ -16,9 +16,9 @@ interface INativeTokenGateway is IRescuable {
   error InvalidAmount();
 
   /**
-   * @notice Thrown when the given reserveId does not match the wrapped native asset reserveId on the Spoke.
+   * @notice Thrown when the underlying asset is not the wrapped native asset.
    */
-  error InvalidReserveId();
+  error NotNativeWrapper();
```
CheyenneAtapour: Should it be something like "NotNativeWrappedAsset"? This makes it sound like an incorrect wrapper contract

Or "NotNativeWrapped" for shorter

src/hub/Hub.sol
```diff
@@ -367,6 +369,7 @@ contract Hub is IHub, AccessManaged {
 
   /// @inheritdoc IHub
   function sweep(uint256 assetId, uint256 amount) external {
+    require(assetId < _assetCount, AssetNotListed());
```
DhairyaSethi, line 390: not sure if it's consistent to have this on sweep and reclaim, since it's not on other liquidity action functions (like add, remove etc)

src/spoke/Spoke.sol
```diff
@@ -601,13 +614,15 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
   function getDynamicReserveConfig(
     uint256 reserveId
   ) external view returns (DynamicReserveConfig memory) {
+    require(address(_reserves[reserveId].hub) != address(0), ReserveNotListed());
     return _dynamicConfig[reserveId][_reserves[reserveId].dynamicConfigKey];
```
DhairyaSethi: here we can create reserve storage pointer as we're using dynamicConfigKey

src/hub/Hub.sol
miguelmtzinf: The other liquidity functions are relaying on the `spoke.active` check for existence, which is fine. `sweep` and `reclaim` are kinda special as they are not called by all spokes (we couldn't use spoke's flags).
At the moment they are relying on the asset's `investmentController` check (zero for non-existent asset). I d like to add this more explicit check for better dev experience, because i do not foresee this being called often.If you think otherwise, i'm happy removing and keep as is (relying on controller auth check).

src/position-manager/interfaces/INativeTokenGateway.sol
miguelmtzinf: agree

src/hub/Hub.sol
DhairyaSethi: ah gotcha makes sense

DhairyaSethi:
| SpokeInstance | <sup>↑4% (+837)</sup> 24,492 | <sup>↑3% (+837)</sup> 25,259 | <sup>↓91% (-837)</sup> 84 | <sup>↓3% (-837)</sup> 23,893 |

runtime margin only 84 with this 😔

src/hub/Hub.sol
```diff
@@ -167,12 +167,14 @@ contract Hub is IHub, AccessManaged {
     address spoke,
     SpokeConfig calldata config
   ) external restricted {
+    require(assetId < _assetCount, AssetNotListed());
```
DhairyaSethi, line 173: maybe we also use helper for consistency here? nbd

src/spoke/Spoke.sol
```diff
@@ -482,17 +482,20 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
   function isBorrowing(uint256 reserveId, address user) external view returns (bool) {
+    _validateListedReserve(_reserves[reserveId]);
     return _positionStatus[user].isBorrowing(reserveId);
   }
 
   /// @inheritdoc ISpokeBase
   function getUserDebt(uint256 reserveId, address user) external view returns (uint256, uint256) {
     UserPosition storage userPosition = _userPositions[user][reserveId];
     Reserve storage reserve = _reserves[reserveId];
+    _validateListedReserve(reserve);
```
```suggestion
    Reserve storage reserve = _getReserve(reserveId);
```
could have also done something like this, is shorter & probably cleaner

src/position-manager/interfaces/INativeTokenGateway.sol
```diff
@@ -15,8 +15,10 @@ interface INativeTokenGateway is IRescuable {
   /// @notice Thrown when the given amount is invalid.
   error InvalidAmount();
 
-  /// @notice Thrown when the given reserveId does not match the wrapped native asset reserveId on the Spoke.
-  error InvalidReserveId();
+  /**
+   * @notice Thrown when the underlying asset is not the wrapped native asset.
+   */
```
DhairyaSethi:
```suggestion
  /// @notice Thrown when the underlying asset is not the wrapped native asset.
```

src/spoke/Spoke.sol
```diff
@@ -481,18 +480,20 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
   function isBorrowing(uint256 reserveId, address user) external view returns (bool) {
+    _getReserve(reserveId);
     return _positionStatus[user].isBorrowing(reserveId);
   }
 
   /// @inheritdoc ISpokeBase
   function getUserDebt(uint256 reserveId, address user) external view returns (uint256, uint256) {
     UserPosition storage userPosition = _userPositions[user][reserveId];
-    Reserve storage reserve = _reserves[reserveId];
+    Reserve storage reserve = _getReserve(reserveId);
```
DhairyaSethi, line 495: would reorder to keep reserve first and then user position, it's what we're doing on functions above as well 

same for the getter below

PR #847, opened 2025-10-02
opt : update OZ dependencies to v5.4

Kogaroshi:
Updates OpenZeppelin dependencies to version 5.4


src/dependencies/openzeppelin/AccessManaged.sol
DhairyaSethi, line 1: think this is unused now bc we only use upgradeable variant, let's clean up any remaining which aren't used either

Kogaroshi: this one is used for the Hub, but I'll go over the rest again

PR #848, opened 2025-10-02
feat: remove named return params for consistency

### LINKED ISSUE #803: fix: rm return named params
- in certain methods, they return named params. Based on conventions, update those to explicitly return values

yan-man:
Removes named return parameters for consistency across the codebase

closes #803


src/libraries/math/MathUtils.sol
```diff
@@ -38,10 +38,12 @@ library MathUtils {
    * @param b The second value to compare.
    * @return result The minimum of the two values.
    */
-  function min(uint256 a, uint256 b) internal pure returns (uint256 result) {
+  function min(uint256 a, uint256 b) internal pure returns (uint256) {
+    uint256 result;
     assembly ('memory-safe') {
       result := xor(b, mul(xor(a, b), lt(a, b)))
     }
+    return result;
```
DhairyaSethi: would make these an exception pls, for fully assembly stuff. this is very ugly 😢

src/interfaces/IMulticall.sol
```diff
@@ -8,5 +8,5 @@ interface IMulticall {
    * @param data The encoded function data for each of the calls to make to this contract.
    * @return results The results from each of the calls passed in via data.
    */
-  function multicall(bytes[] calldata data) external returns (bytes[] memory results);
+  function multicall(bytes[] calldata data) external returns (bytes[] memory);
```
DhairyaSethi, line 11: exception on interfaces too pls!

src/spoke/SpokeConfigurator.sol
```diff
@@ -80,7 +80,7 @@ contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
     address priceSource,
     ISpoke.ReserveConfig calldata config,
     ISpoke.DynamicReserveConfig calldata dynamicConfig
-  ) external onlyOwner returns (uint256 reserveId) {
+  ) external onlyOwner returns (uint256) {
     return ISpoke(spoke).addReserve(hub, assetId, priceSource, config, dynamicConfig);
   }
```
DhairyaSethi, line 85: this fine, the idea is that we should always return and never rely on assigning the return var and move on

src/spoke/libraries/PositionStatusMap.sol
DhairyaSethi, line 1: revert all changes here pls

src/interfaces/IMulticall.sol
yan-man: we were only having it on some interfaces too, so if we want to do this, should add everywhere

src/libraries/math/MathUtils.sol
yan-man: let's get a group decision on this today

DhairyaSethi: even v3 which embracing this design makes an exception for pure assembly, strongly against doing this here

src/interfaces/IMulticall.sol
yan-man: decided not on interfaces

src/libraries/math/MathUtils.sol
yan-man: decided to keep named params on libs w assembly functions

src/spoke/Spoke.sol
```diff
@@ -763,9 +763,13 @@ abstract contract Spoke is ISpoke, Multicall, AccessManagedUpgradeable, EIP712 {
...
-  ) internal returns (UserAccountData memory userAccountData) {
-    userAccountData = _calculateAndPotentiallyRefreshUserAccountData(user, true);
+  ) internal returns (UserAccountData memory) {
+    UserAccountData memory userAccountData = _calculateAndPotentiallyRefreshUserAccountData(
+      user,
+      true
+    );
     emit RefreshAllUserDynamicConfig(user);
+    return userAccountData;
   }
```
DhairyaSethi: here is the _one place_ i would make an exception bc this reads much much worse

DhairyaSethi: this was intentionally done against style

StErMi:
FYI there are still functions that are using "named returns"

- `MathUtils.min`
- `MathUtils.mulDivDown`
- `MathUtils.mulDivUp`
- `PercentageMath.percentMulDown`
- `PercentageMath.percentMulUp`
- `PercentageMath.percentDivDown`
- `PercentageMath.percentDivUp`
- Plenty of functions in `WadRayMath`
- `UnitPriceFeed.getRoundData`
- `UnitPriceFeed.latestRoundData`
- `Spoke._calculateAndPotentiallyRefreshUserAccountData`
- Plenty of functions in `PositionStatusMap`
- `NoncesKeyed._unpack`

And probably I could have missed others.

👉 At this point I would suggest you not only explicitly return a value but also remove the return variable name.

DhairyaSethi:
@StErMi it's an exception with fn in pure yul bc it's cleaner to assign vars & with struct returns

PR #852, opened 2025-10-03
rft: remove debtToCollateral

avniculae:


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -290,7 +285,10 @@ library LiquidationLogic {
     require(params.debtToCover >= debtToLiquidate, ISpoke.MustNotLeaveDust());
 
     uint256 collateralToLiquidator = collateralToLiquidate -
-      (collateralToLiquidate - debtToCollateral).percentMulDown(params.liquidationFee);
+      collateralToLiquidate.mulDivDown(
+        params.liquidationFee * (liquidationBonus - PercentageMath.PERCENTAGE_FACTOR),
+        liquidationBonus * PercentageMath.PERCENTAGE_FACTOR
```
DhairyaSethi, line 290: can add toBps() similar to toWad() we already have

avniculae: I wouldn't add it here because liquidationBonus is already bps. We need it bc numerator multiplies 2 bps. We can add `toBps` in the other places, wdyt?

avniculae: although I am fine keeping PERCENTAGE_FACTOR in the other places as well, as `toBps()` can confuse readers

avniculae: toWad() makes sense bc we divide 2 wad values, but when we use PERCENTAGE_FACTOR we only have one bps value

PR #854, opened 2025-10-03
fix: consistent rounding on convertToDrawnShares

DhairyaSethi:
`convertTo*` methods have no specific standard necessarily and are opinionated. We are just making the rounding direction consistent for `convertToDrawnShares`. convertTo can be removed as well since all rounding directions are already exposed through preview methods

todo:
- symbolic tests proving convertTo* & corresponding preview* equivalence

### DIFF
--- a/src/hub/interfaces/IHub.sol
+++ b/src/hub/interfaces/IHub.sol
@@ -269,7 +269,7 @@ interface IHub is IHubBase, IAccessManaged {
 
   /**
    * @notice Converts the specified amount of drawn assets to shares amount.
-   * @dev Rounds down to the nearest shares amount.
+   * @dev Rounds up to the nearest shares amount.
    * @param assetId The identifier of the asset.
    * @param assets The amount of drawn assets to convert to shares amount.
    * @return The amount of drawn shares converted from assets amount.


tests/unit/Hub/Hub.Draw.t.sol
```diff
@@ -89,6 +89,10 @@ contract HubDrawTest is HubBase {
     assertEq(underlying.balanceOf(bob), MAX_SUPPLY_AMOUNT - amount, 'bob asset final balance');
     assertEq(underlying.balanceOf(address(spoke1)), 0, 'spoke1 asset final balance');
     assertEq(underlying.balanceOf(address(spoke2)), 0, 'spoke2 asset final balance');
+    assertEq(
+      hub1.convertToDrawnShares(assetId, amount),
+      hub1.previewRestoreByShares(assetId, amount)
+    );
```
DhairyaSethi, line 95: just done for coverage, todo equivalence checks for the convert functions and use only preview* in test helpers

src/hub/Hub.sol
```diff
@@ -499,7 +499,7 @@ contract Hub is IHub, AccessManaged {
 
   /// @inheritdoc IHub
   function convertToDrawnShares(uint256 assetId, uint256 assets) external view returns (uint256) {
-    return _assets[assetId].toDrawnSharesDown(assets);
+    return _assets[assetId].toDrawnSharesUp(assets);
```
miguelmtzinf, line 502: is the rounding direction mentioned in IHub docs?

DhairyaSethi: F good catch forgot

PR #856, opened 2025-10-03
fix: aave oracle constructor param check

DhairyaSethi:
Ensures that the address of the spoke contract passed to the oracle constructor is not `address(0)`

### DIFF
--- a/src/spoke/AaveOracle.sol
+++ b/src/spoke/AaveOracle.sol
@@ -30,7 +30,8 @@ contract AaveOracle is IAaveOracle {
   constructor(address spoke_, uint8 decimals_, string memory description_) {
     require(spoke_ != address(0), InvalidAddress());
     SPOKE = spoke_;
-    DECIMALS = decimals_; // compatibility should be validated on the spoke
+    // decimals compatibility should be validated on the spoke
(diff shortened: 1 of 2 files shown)


src/spoke/AaveOracle.sol
```diff
@@ -28,8 +28,9 @@ contract AaveOracle is IAaveOracle {
    * @param description_ The description of the oracle.
    */
   constructor(address spoke_, uint8 decimals_, string memory description_) {
+    require(spoke_ != address(0), InvalidAddress());
     SPOKE = spoke_;
-    DECIMALS = decimals_;
+    DECIMALS = decimals_; // compatibility should be validated on the spoke
```
yan-man: maybe put in line above

src/spoke/interfaces/IAaveOracle.sol
```diff
@@ -33,6 +33,11 @@ interface IAaveOracle is IPriceOracle {
    */
   error InvalidPrice(uint256 reserveId);
 
+  /**
+   * @dev Thrown when the given address is invalid.
```
yan-man: let's just adopt /// natspec on new comments

PR #857, opened 2025-10-03
feat: Emit fee receiver in AccrueFees event

CheyenneAtapour:


src/hub/Hub.sol
```diff
@@ -121,7 +121,8 @@ contract Hub is IHub, AccessManaged {
   ) external restricted {
     require(assetId < _assetCount, AssetNotListed());
     Asset storage asset = _assets[assetId];
-    asset.accrue(assetId, _spokes[assetId][asset.feeReceiver]);
+    address feeReceiver = asset.feeReceiver;
```
yan-man, line 125: below L141 should also be feeReceiver then instead of asset.feeReceiver

src/hub/Hub.sol
```diff
@@ -174,7 +175,8 @@ contract Hub is IHub, AccessManaged {
   /// @inheritdoc IHub
   function setInterestRateData(uint256 assetId, bytes calldata irData) external restricted {
     Asset storage asset = _assets[assetId];
-    asset.accrue(assetId, _spokes[assetId][asset.feeReceiver]);
+    address feeReceiver = asset.feeReceiver;
+    asset.accrue(assetId, _spokes[assetId][feeReceiver], feeReceiver);
```
yan-man, line 184: why not just pass asset.feeReceiver directly

src/hub/Hub.sol
```diff
@@ -183,8 +185,9 @@ contract Hub is IHub, AccessManaged {
   function add(uint256 assetId, uint256 amount, address from) external returns (uint256) {
     Asset storage asset = _assets[assetId];
     SpokeData storage spoke = _spokes[assetId][msg.sender];
+    address feeReceiver = asset.feeReceiver;
 
-    asset.accrue(assetId, _spokes[assetId][asset.feeReceiver]);
+    asset.accrue(assetId, _spokes[assetId][feeReceiver], feeReceiver);
```
yan-man, line 195: same, pass directly?

src/hub/interfaces/IHub.sol
```diff
@@ -85,7 +85,7 @@ interface IHub is IHubBase, IAccessManaged {
     uint256 drawnAmount,
     uint256 premiumAmount
   );
-  event AccrueFees(uint256 indexed assetId, uint256 shares);
+  event AccrueFees(uint256 indexed assetId, uint256 shares, address feeReceiver);
```
yan-man: hm do we want to move it to 2nd arg and indexed?

src/hub/libraries/AssetLogic.sol
```diff
@@ -119,7 +119,8 @@ library AssetLogic {
   function accrue(
     IHub.Asset storage asset,
     uint256 assetId,
-    IHub.SpokeData storage feeReceiver
+    IHub.SpokeData storage feeReceiver,
+    address feeReceiverAddress
```
yan-man: nit: feeReceiverSpoke and feeReceiver

src/hub/Hub.sol
CheyenneAtapour: Just cause it'd have to be accessed twice here

src/hub/interfaces/IHub.sol
CheyenneAtapour: I'm agnostic

src/hub/libraries/AssetLogic.sol
CheyenneAtapour: hmm, I'm not sure what the convention should be here. It's not clear to me that `feeReceiver` should be the address

src/hub/Hub.sol
CheyenneAtapour: Done

src/hub/Hub.sol
DhairyaSethi: thing is `_spokes[assetId][feeReceiver]` is still looked up always even when it's not needed (ie fees is zero)

yan-man: i think we can assume that fees > 0 in most cases and the zero will be kind of rare edge case right?

ok you're right @CheyenneAtapour not sure what I was seeing before... it was late

miguelmtzinf:
Closing this in favor of #866

PR #858, opened 2025-10-04
feat: Validate underlying decimals

CheyenneAtapour:


src/hub/Hub.sol
```diff
@@ -818,6 +816,20 @@ contract Hub is IHub, AccessManaged {
     require(amount > 0 && amount <= asset.swept, InvalidAmount());
   }
 
+  function _validateAssetDecimals(address underlying, uint8 givenDecimals) internal view {
```
yan-man: i think we can just do decimals instead of givenDecimals

src/hub/Hub.sol
```diff
@@ -818,6 +816,20 @@ contract Hub is IHub, AccessManaged {
...
+    (bool success, bytes memory returnData) = address(underlying).staticcall(
+      abi.encodeCall(IERC20Metadata.decimals, ())
+    );
+    if (success) {
+      require(abi.decode(returnData, (uint8)) == givenDecimals, InvalidAssetDecimals());
```
yan-man: can we just do `require(success && ...)`, or do we want this to proceed even if staticcall fails

src/hub/Hub.sol
CheyenneAtapour: I just didn't want that to be confused from the fetched decimals, but yeah sure

src/hub/Hub.sol
CheyenneAtapour: The suggestion was to fallback to given decimals if it fails, but I'm open to whatever

CheyenneAtapour:
Won't add this check

PR #861, opened 2025-10-04
feat: optimise math fn

DhairyaSethi:
- Updates `bpsToRay` & `bpsToWad` methods in assembly, consistent with rest of WadRayMath; functions are 1-1 equivalent to before 
- Updates `calculateLinearInterest` to be in assembly. two functional changes
  - retains checkedSub(timestamp, lastUpdateTimestamp) with explicit revert on underflow (we can rm this check for another 30 gas, but check seems worth to keep in)
  - removes checkedAdd(rate, RAY)
    -  for this, makes rate explicitly uint96, this ensures that this addition can never overflow: uint96.max + RAY ~ 96 + 90 bits < 256 bits
    
max current timestamp for overflow is 1e48

note:
```solidity
unchecked { result = result / SECONDS_PER_YEAR; }
```
was non functional before. this does not skip zero denominator check


src/libraries/math/MathUtils.sol
```diff
@@ -2,31 +2,29 @@
...
 /// @title MathUtils library
 /// @author Aave Labs
 library MathUtils {
-  using WadRayMath for uint256;
-
+  uint256 internal constant RAY = 1e27;
   /// @dev Ignoring leap years
   uint256 internal constant SECONDS_PER_YEAR = 365 days;
 
   /// @notice Function to calculate the interest accumulated using a linear interest rate formula.
```
miguelmtzinf:
```suggestion
  /// @notice Calculates the interest accumulated using a linear interest rate formula.
```

src/libraries/math/WadRayMath.sol
```diff
@@ -143,22 +143,38 @@ library WadRayMath {
...
     }
   }
 
   /// @notice Converts value from basis points to Wad, rounding down.
   /// @dev Reverts if intermediate multiplication overflows.
   /// @return c = floor(a * WAD / PERCENTAGE_FACTOR) in Wad units.
-  function bpsToWad(uint256 a) internal pure returns (uint256) {
-    return (a * WAD) / PERCENTAGE_FACTOR;
+  function bpsToWad(uint256 a) internal pure returns (uint256 c) {
+    // to avoid overflow, b/WAD == a
```
avniculae, line 155: there is no b here (same below)

DhairyaSethi: ty fixed

PR #863, opened 2025-10-04
chore: rename `InBaseCurrency` to `Value`

DhairyaSethi:
- rename 'InBaseCurrency' to 'Value' to avoid confusion with v3 meaning (where base currency means scaled by oracle ie 8 decimals, here value means scaled by 18 + 8 decimals, ie 1 usd = 1e26)
- rft risk premium computation ==> should be equivalent to before

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -439,9 +439,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     if (!_positionManager[msg.sender].approval[onBehalfOf]) {
       return;
     }
-
     _positionManager[msg.sender].approval[onBehalfOf] = false;
-
     emit SetUserPositionManager(onBehalfOf, msg.sender, false);
   }
 


src/spoke/Spoke.sol
```diff
@@ -325,7 +325,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       drawnDebt: 0, // populated below
       premiumDebt: 0, // populated below
       accruedPremium: 0, // populated below
-      totalDebtInBaseCurrency: userAccountData.totalDebtInBaseCurrency,
+      totalDebtValue: userAccountData.totalDebtValue,
       suppliedCollateralsCount: userAccountData.suppliedCollateralsCount,
```
```suggestion
      collateralReservesCount: userAccountData.collateralReservesCount,
```
this feels more consistent/aligned with `borrowedReservesCount`

src/spoke/Spoke.sol
```diff
@@ -533,7 +534,6 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     uint16 configKey
   ) external view returns (DynamicReserveConfig memory) {
     _getReserve(reserveId);
-    // @dev we do not revert if key is unset
```
miguelmtzinf, line 536: worth to add to ISpoke?

src/spoke/Spoke.sol
```diff
@@ -728,48 +728,46 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    accountData.healthFactor = accountData.totalDebtInBaseCurrency == 0
-      ? type(uint256).max
-      : accountData
+    if (accountData.totalDebtValue > 0) {
+      accountData.healthFactor = accountData
         .avgCollateralFactor
-        .wadDivDown(accountData.totalDebtInBaseCurrency)
+        .wadDivDown(accountData.totalDebtValue)
         .fromBpsDown();
+    } else accountData.healthFactor = type(uint256).max;
```
```suggestion
    } else {
        accountData.healthFactor = type(uint256).max;
    }
```

src/spoke/Spoke.sol
DhairyaSethi: yes! was removed from here to add to ISpoke natspec only

src/spoke/Spoke.sol
DhairyaSethi: we said if it stays in a single line then we can avoid `{ .. }`, no? happy to make the change tho

src/spoke/Spoke.sol
DhairyaSethi: it was written this way because we have collateralCount and this count is specifically for collateral's which are non zero supplied (bc we can have enabled as collateral but not supplied -> these wouldn't be included in this counter)

src/spoke/Spoke.sol
```diff
@@ -728,48 +728,46 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-          assetUnit
-        );
+        accountData.totalDebtValue += ((drawnDebt + premiumDebt) * assetPrice).wadDivUp(assetUnit);
         accountData.borrowedReservesCount = accountData.borrowedReservesCount.uncheckedAdd(1);
       }
     }
 
     // at this point avgCollateralFactor is the weighted sum of collateral scaled by collateralFactor
     // (avgCollateralFactor / totalCollateral) * totalCollateral can be simplified to avgCollateralFactor
     // strip BPS factor from result, because running avgCollateralFactor sum has been scaled by collateralFactor (in BPS) above
```
miguelmtzinf: this comment is a bit confusing to me. I d change to this, and placed it right before the hf calc (L739)
```suggestion
    // at this point, avgCollateralFactor is the collateral-weighted sum (scaled by collateralFactor in BPS)
    // health factor uses this directly for simplicity, while the division by totalCollateral to compute the weighted average is done later.
```

src/spoke/interfaces/ISpoke.sol
```diff
@@ -65,11 +65,11 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
...
   struct UserAccountData {
-    uint256 userRiskPremium;
+    uint256 riskPremium;
     uint256 avgCollateralFactor;
     uint256 healthFactor;
-    uint256 totalCollateralInBaseCurrency;
-    uint256 totalDebtInBaseCurrency;
+    uint256 totalCollateralValue;
+    uint256 totalDebtValue;
     uint256 suppliedCollateralsCount; // number of reserves with collateral factor > 0, enabled as collateral and strictly positive supplied amount
```
miguelmtzinf: same

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -31,7 +31,7 @@ library LiquidationLogic {
     uint256 drawnDebt;
     uint256 premiumDebt;
     uint256 accruedPremium;
-    uint256 totalDebtInBaseCurrency;
+    uint256 totalDebtValue;
     address liquidator;
     uint256 suppliedCollateralsCount;
```
miguelmtzinf: same

src/spoke/Spoke.sol
DhairyaSethi: [d23f8b6](https://github.com/aave/aave-v4/pull/863/commits/d23f8b6972f887a05d99dd24145e38c4365f9e17)

tests/Base.t.sol
```diff
@@ -1356,7 +1356,7 @@ abstract contract Base is Test {
   }
 
   /// returns the USD value of the reserve normalized by it's decimals, in terms of WAD
-  function _getValueInBaseCurrency(
+  function _getValue(
```
miguelmtzinf, line 1359: this and `_getDebtValue` are exactly the same.
We could rename to `getReserveAmountValue`

src/spoke/Spoke.sol
DhairyaSethi: we can rename these to: `borrowedCount`, `suppliedCollateralCount` to make it shorter cc @avniculae

src/spoke/Spoke.sol
```diff
@@ -728,48 +728,46 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
-    // at this point avgCollateralFactor is the weighted sum of collateral scaled by collateralFactor
-    // (avgCollateralFactor / totalCollateral) * totalCollateral can be simplified to avgCollateralFactor
-    // strip BPS factor from result, because running avgCollateralFactor sum has been scaled by collateralFactor (in BPS) above
-    accountData.healthFactor = accountData.totalDebtInBaseCurrency == 0
-      ? type(uint256).max
-      : accountData
+    if (accountData.totalDebtValue > 0) {
+      // at this point, `avgCollateralFactor` is the collateral-weighted sum (scaled by collateralFactor in BPS)
+      // health factor uses this directly for simplicity, while the division by totalCollateral to compute the weighted average is done later
```
```suggestion
      // at this point, `avgCollateralFactor` is the collateral-weighted sum (scaled by `collateralFactor` in BPS)
      // health factor uses this directly for simplicity, while the division by `totalCollateral` to compute the weighted average is done later
```

tests/Base.t.sol
DhairyaSethi: image
debt roundsUp

miguelmtzinf: my bad, ty for clarifying

src/spoke/Spoke.sol
avniculae: I'm fine with `borrowedCount` and `suppliedCollateralCount`. other suggestion: `activeCollateralCount`

DhairyaSethi: [f655711](https://github.com/aave/aave-v4/pull/863/commits/f6557110e8f9f34af5028217279490b65d8e36c7)

src/position-manager/interfaces/ISignatureGateway.sol
```diff
@@ -160,9 +160,9 @@ interface ISignatureGateway is IMulticall, INoncesKeyed, IRescuable {
     address onBehalfOf,
     uint256 value,
     uint256 deadline,
-    uint8 v,
-    bytes32 r,
-    bytes32 s
```
StErMi, line 165: The natspec must be updated for the params `permitV`, `permitR` and `permitS`

DhairyaSethi: yeah we'd omitted natspec for these three in the interface

src/spoke/interfaces/ISpoke.sol
```diff
@@ -362,9 +362,9 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     address onBehalfOf,
     uint256 value,
     uint256 deadline,
-    uint8 v,
-    bytes32 r,
-    bytes32 s
```
StErMi, line 367: The natspec must be updated for the params `permitV`, `permitR` and `permitS`

src/spoke/Spoke.sol
```diff
@@ -728,48 +731,47 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
         .fromBpsDown();
+    } else {
+      accountData.healthFactor = type(uint256).max;
+    }
 
-    // divide by total collateral to get avg collateral factor in wad
-    accountData.avgCollateralFactor = accountData.totalCollateralInBaseCurrency == 0
-      ? 0
-      : accountData
+    if (accountData.totalCollateralValue > 0) {
```
StErMi, line 751: when `accountData.totalCollateralValue` is zero, `accountData.avgCollateralFactor` will be zero anyway but I would have liked to see an **explicit** reset in the `lese` branch that does not exist right now.

PR #864, opened 2025-10-05
fix: Reorder functions

miguelmtzinf:
- Functions are grouped by category (asset, spoke, user) and target user (DAO, user).
- Getters of constants/immutable variables are places at the bottom.
- Reorder of functions based on [Solidity style guidelines.](https://docs.soliditylang.org/en/latest/style-guide.html#order-of-functions)

Changes:
- `_isPositionManager` internal function modifier changed from `private` to `internal`.
- A handful of getter functions were removed from `IHub`, because they were part of `IHubBase` already (dup).

### DIFF
--- a/src/hub/interfaces/IHub.sol
+++ b/src/hub/interfaces/IHub.sol
@@ -65,33 +65,33 @@ interface IHub is IHubBase, IAccessManaged {
     uint56 drawCap;
   }
 
-  /// @notice Emitted when a spoke is added.
-  /// @param assetId The identifier of the asset.
-  /// @param spoke The address of the spoke.
-  event AddSpoke(uint256 indexed assetId, address indexed spoke);
-
   /// @notice Emitted when an asset is added.
   /// @param assetId The identifier of the asset.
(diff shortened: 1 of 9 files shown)


src/hub/interfaces/IHub.sol
```diff
@@ -332,74 +313,54 @@ interface IHub is IHubBase, IAccessManaged {
...
   /// @param assetId The identifier of the asset.
-  /// @return The amount of deficit.
-  function getAssetDeficit(uint256 assetId) external view returns (uint256);
+  /// @return The current drawn index of the asset.
+  function getAssetDrawnIndex(uint256 assetId) external view returns (uint256);
+
+  /// @notice Returns the current drawn rate for the specified asset.
+  /// @param assetId The identifier of the asset.
+  /// @return The current drawn rate of the asset.
+  function getAssetDrawnRate(uint256 assetId) external view returns (uint256);
```
DhairyaSethi, line 339: these don't return the 'current' but rather last stored index and rate. let's correct the natspec or impl (since last stored values are given by getAsset())

PR #865, opened 2025-10-06
fix: Rename getters and move to IHubBase, rename error

miguelmtzinf:
- Rename `getLiquidity` to `getAssetLiquidity` and move to `IHubBase`.
- Rename `getDeficit` to `getAssetDeficit` and move to `IHubBase`.
- Rename `getSwept` to `getAssetSwept`.
- Move `getSpokeDeficit` getter to `IHubBase`.
- Rename `SpecifiedCurrencyNotBorrowedByUser` error to `SpecifiedReserveNotBorrowedByUser`
- Add in `_validateAdd` function a quick natspec doc about cap max value (similar to `_validateDraw /// @dev Spoke with maximum cap have unlimited draw capacity.`)
- Remove any unnecessary `virtual` and `override` modifier.
- Remove zero address check on `SignatureGateway` contract for owner given that `Ownable` already does it.

### DIFF
--- a/src/position-manager/SignatureGateway.sol
+++ b/src/position-manager/SignatureGateway.sol
@@ -72,7 +72,7 @@ contract SignatureGateway is
   /// @param spoke_ The address of the connected spoke.
   /// @param initialOwner_ The address of the initial owner.
   constructor(address spoke_, address initialOwner_) Ownable(initialOwner_) {
-    require(spoke_ != address(0) && initialOwner_ != address(0), InvalidAddress());
+    require(spoke_ != address(0), InvalidAddress());
     _spoke = ISpoke(spoke_);
   }
 


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -237,7 +237,7 @@ library LiquidationLogic {
       params.isUsingAsCollateral && params.collateralFactor > 0,
       ISpoke.CollateralCannotBeLiquidated()
     );
-    require(params.debtReserveBalance > 0, ISpoke.SpecifiedCurrencyNotBorrowedByUser());
+    require(params.debtReserveBalance > 0, ISpoke.SpecifiedReserveNotBorrowedByUser());
```
DhairyaSethi: do we need `specified` prefix here lol, feels inconsistent with rest, for ex ReserveNotListed bc that's also referring to a "specified" reserve

miguelmtzinf: renamed to `ReserveNotBorrowed`

avniculae: agreed, I have the same renaming in #796

PR #866, opened 2025-10-06
fix: emit fee receiver on accrue

DhairyaSethi:
Emit the address of the `feeReceiver` for the event emitted on the `accrue` function
- facilitates offchain indexing


src/hub/libraries/AssetLogic.sol
```diff
@@ -130,10 +130,14 @@ library AssetLogic {
...
-  ) internal {
+    mapping(uint256 => IHub.Asset) storage assets,
+    mapping(uint256 => mapping(address => IHub.SpokeData)) storage spokes,
+    uint256 assetId
+  ) internal returns (IHub.Asset storage) {
+    IHub.Asset storage asset = assets[assetId];
+
+    if (asset.lastUpdateTimestamp == block.timestamp) return asset;
+
     uint256 drawnIndex = asset.getDrawnIndex();
```
```suggestion
    uint256 newDrawnIndex = asset.getDrawnIndex();
```
just noticing now

src/hub/Hub.sol
```diff
@@ -209,10 +207,9 @@ contract Hub is IHub, AccessManaged {
 
   /// @inheritdoc IHubBase
   function remove(uint256 assetId, uint256 amount, address to) external returns (uint256) {
-    Asset storage asset = _assets[assetId];
+    Asset storage asset = _assets.accrue(_spokes, assetId);
```
miguelmtzinf: not a fan of returning the asset... 
```suggestion
    Asset storage asset = _assets[assetId];
    _assets.accrue(_spokes, assetId);
```

it this good also in terms of gas costs?

src/hub/Hub.sol
```diff
@@ -25,7 +25,7 @@ contract Hub is IHub, AccessManaged {
   using WadRayMath for uint256;
   using SharesMath for uint256;
   using PercentageMath for uint256;
-  using AssetLogic for Asset;
+  using AssetLogic for *;
```
miguelmtzinf: this makes the code a bit more ugly / less intuitive imho...
wondering if we should revert this change and call `AssetLogic.accrue(assets, spokes)` instead

src/hub/libraries/AssetLogic.sol
yan-man: or "next" to match the natspec on `indexDelta`

src/hub/libraries/AssetLogic.sol
```diff
@@ -131,9 +131,13 @@ library AssetLogic {
   /// @notice Accrues interest and fees for the specified asset.
   function accrue(
     IHub.Asset storage asset,
-    uint256 assetId,
-    IHub.SpokeData storage feeReceiver
+    mapping(address => IHub.SpokeData) storage assetSpokes,
```
miguelmtzinf: Passing the asset's spokes or all spokes (as below) is very similar in terms of gas costs.
```suggestion
    mapping(uint256 => mapping(address => IHub.SpokeData)) storage spokes,
```

DhairyaSethi: yep it's just the additional hash and lookup which is 'wasted' if we don't end up in the feesAmount > 0 branch

DhairyaSethi: since it's just passed there i think just passing _spokes is fine too

miguelmtzinf: agree, changed

PR #867, opened 2025-10-06
rft: risk premium calc

avniculae:


src/spoke/Spoke.sol
```diff
@@ -750,22 +750,20 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+    collateralInfo.sortByKey(); // sort by collateral risk in ASC, collateral value in DES
+    for (uint256 index = 0; debtValueLeftToCover > 0 && index < collateralInfo.length(); ) {
+      (uint256 collateralRisk, uint256 userCollateralValue) = collateralInfo.get(index);
+      userCollateralValue = userCollateralValue.min(debtValueLeftToCover);
       accountData.riskPremium += userCollateralValue * collateralRisk;
-      runningCollateralValue += userCollateralValue;
-      runningDebtValue -= userCollateralValue;
-      runningIndex = runningIndex.uncheckedAdd(1);
+      debtValueLeftToCover -= userCollateralValue;
+      index = index.uncheckedAdd(1);
```
DhairyaSethi: this is also not needed anymore, solidity will already do this add unchecked in the for loop

src/spoke/Spoke.sol
```diff
@@ -750,22 +750,20 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    uint256 runningDebtValue = accountData.totalDebtValue;
-    uint256 runningCollateralValue = 0;
-    uint256 runningIndex = 0;
-    collateralInfo.sortByKey(); // sort by collateral risk in ASC, collateral value in DESC
-
-    while (runningIndex < collateralInfo.length() && runningDebtValue > 0) {
-      (uint256 collateralRisk, uint256 userCollateralValue) = collateralInfo.get(runningIndex);
-      userCollateralValue = userCollateralValue.min(runningDebtValue);
+    uint256 debtValueLeftToCover = accountData.totalDebtValue;
+    collateralInfo.sortByKey(); // sort by collateral risk in ASC, collateral value in DES
```
DhairyaSethi: desc

avniculae: [405f0ab](https://github.com/aave/aave-v4/pull/867/commits/405f0ab1e3ac1ece033c5d2a6af56741b7df29af)

src/spoke/Spoke.sol
avniculae: [405f0ab](https://github.com/aave/aave-v4/pull/867/commits/405f0ab1e3ac1ece033c5d2a6af56741b7df29af)

src/spoke/Spoke.sol
```diff
@@ -750,22 +750,19 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-      (uint256 collateralRisk, uint256 userCollateralValue) = collateralInfo.get(runningIndex);
-      userCollateralValue = userCollateralValue.min(runningDebtValue);
+    for (uint256 index = 0; debtValueLeftToCover > 0 && index < collateralInfo.length(); ++index) {
+      (uint256 collateralRisk, uint256 userCollateralValue) = collateralInfo.get(index);
+      userCollateralValue = userCollateralValue.min(debtValueLeftToCover);
       accountData.riskPremium += userCollateralValue * collateralRisk;
-      runningCollateralValue += userCollateralValue;
-      runningDebtValue -= userCollateralValue;
-      runningIndex = runningIndex.uncheckedAdd(1);
+      debtValueLeftToCover -= userCollateralValue;
```
avniculae: should we do uncheckedsub here? (sinc userCollateralValue is capped above, so it is at most debtValueLeftToCover)

src/spoke/Spoke.sol
DhairyaSethi: why did gas increase? 👀

src/spoke/Spoke.sol
avniculae: [808afd2](https://github.com/aave/aave-v4/pull/867/commits/808afd2fc2d05f54001e82b6cdeb53acdcc838d2)

src/spoke/Spoke.sol
avniculae:
```
The precise conditions under which the new optimization avoids the overflow check are the following:
- The loop condition is a comparison of the form i < ... for a local variable i (called "loop counter" from now on).
- This comparison must be performed on the same type as the loop counter, i.e. the type of the right-hand-side must be implicitly convertible to the type of the loop counter, such that the loop counter is not implicitly widened before comparing.
- The loop counter must be a local variable of a built-in integer type.
- The loop expression must be a prefix or postfix increment of the loop counter, i.e. i++ or ++i.
- The loop counter may not be modified in the loop condition or the loop body.
```

I think bc we have 2 conditions in the for loop

DhairyaSethi: damn, it'll probably be cheaper if revert this change over moving if the second condition inside the loop body with a if (<) break

avniculae: you prefer it this way [a74db3e](https://github.com/aave/aave-v4/pull/867/commits/a74db3e38fb22f2dde46cc582b20c078d4839d4f) ?

DhairyaSethi: no how about 

for (index = 0; index < len; ++i)
  (..) = list.get()
  if debtRemaining >= collBalance { 
    ...
    break;
  }
  ...

so we break .min into an explicit branch instead like before and break instead

avniculae: I don't like the code duplication in this scenario, what about this [c2b7a54](https://github.com/aave/aave-v4/pull/867/commits/c2b7a549ae8b174c2298133aa1b1ab6c849a01e0) ?

src/spoke/Spoke.sol
```diff
@@ -750,22 +750,23 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
-    uint256 runningDebtValue = accountData.totalDebtValue;
-    uint256 runningCollateralValue = 0;
-    uint256 runningIndex = 0;
+    uint256 debtValueLeftToCover = accountData.totalDebtValue;
     collateralInfo.sortByKey(); // sort by collateral risk in ASC, collateral value in DESC
+    for (uint256 index = 0; index < collateralInfo.length(); ++index) {
+      if (debtValueLeftToCover == 0) {
+        break;
+      }
```
DhairyaSethi, line 758: if we move this towards the end we save a length lookup and comparison in over collateralised cases no?

PR #869, opened 2025-10-07
fix: Validate hub actions via underflow

CheyenneAtapour:
- Removes Hub's checks on spokes’ accounting, relying instead on math underflow for validation.
- UX is not affected, as these checks apply only to functions that are not user-facing but are invoked by Spokes.
  - Spoke checks max withdrawal limit before calling remove (max withdraw if input amount is greater than max)
  - Spoke checks max repayable amount before calling (max repay if input amount is greater than total debt)
- ~4k gas reduction on `remove` and <1k gas on `restore`.
- Removing checks of total `liquidity` on `remove`, `restore` and `transferShares` does not bring significant savings.
- Adds check for available liquidity on `sweep` for consistency

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -330,6 +330,7 @@ contract Hub is IHub, AccessManaged {
 
     asset.accrue(_spokes, assetId);
     _validateEliminateDeficit(callerSpoke, amount);
+
     uint256 deficit = coveredSpoke.deficit;
     require(amount <= deficit, InvalidAmount());
 


src/hub/Hub.sol
```diff
@@ -787,10 +778,6 @@ contract Hub is IHub, AccessManaged {
     require(from != address(this), InvalidAddress());
     require(drawnAmount + premiumAmount > 0, InvalidAmount());
     require(spoke.active, SpokeNotActive());
-    uint256 drawn = _getSpokeDrawn(spoke, assetId);
-    uint256 premium = _getSpokePremium(spoke, assetId);
-    require(drawnAmount <= drawn, SurplusAmountRestored(drawn));
```
CheyenneAtapour, line 778: Are we ok allowing donations up to a share on restore?

CheyenneAtapour: Keep the drawn one

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
```diff
@@ -111,7 +111,7 @@ contract SpokeBorrowValidationTest is SpokeBase {
     Utils.supply(spoke1, daiReserveId, alice, daiAmount, alice);
 
     // Bob draw more than supplied dai amount
-    vm.expectRevert(abi.encodeWithSelector(IHub.InsufficientLiquidity.selector, daiAmount));
+    vm.expectRevert();
```
Kogaroshi: Why not `vm.expectRevert(stdError.arithmeticError);` here ? (and in some other tests)

CheyenneAtapour: It reverts without data. I think it's because it's not directly on the hub?

src/hub/Hub.sol
```diff
@@ -789,10 +782,6 @@ contract Hub is IHub, AccessManaged {
     require(spoke.active, SpokeNotActive());
     require(!spoke.paused, SpokePaused());
     require(drawnAmount + premiumAmount > 0, InvalidAmount());
-    uint256 drawn = _getSpokeDrawn(asset, spoke);
-    uint256 premium = _getSpokePremium(asset, spoke);
-    require(drawnAmount <= drawn, SurplusDeficitReported(drawn));
```
miguelmtzinf, line 793: why removing this one here but not in `restore`? we need it because we could report more drawn deficit than the spoke's drawn amount.
ex: sharePrice =1.28, spoke.drawnShares = 3, spoke.premium = 0, spoke.totalOwed = ceil(5*1.28) = 4, input.drawnAmount = 5, calculated drawnShares = floor(5/.1.28) = 3, so we report more than drawn (5) that there exists (4)

miguelmtzinf: I d like to have a test case for the above

src/hub/Hub.sol
```diff
@@ -242,13 +242,12 @@ contract Hub is IHub, AccessManaged {
 
     asset.accrue(_spokes, assetId);
     _validateDraw(asset, spoke, amount, to);
-    uint256 liquidity = asset.liquidity;
-    require(amount <= liquidity, InsufficientLiquidity(liquidity));
```
miguelmtzinf, line 245: i d add this check back (in `remove` too), as it's about available liquidity and not about spokes' accounting. No big diff in terms of gas costs.

src/hub/Hub.sol
```diff
@@ -827,13 +816,13 @@ contract Hub is IHub, AccessManaged {
   function _validateSweep(Asset storage asset, address caller, uint256 amount) internal view {
```
miguelmtzinf, line 827: let's add liquidity check in `sweep` pls

src/hub/Hub.sol
```diff
@@ -775,9 +770,7 @@ contract Hub is IHub, AccessManaged {
     require(spoke.active, SpokeNotActive());
     require(!spoke.paused, SpokePaused());
     uint256 drawn = _getSpokeDrawn(asset, spoke);
-    uint256 premium = _getSpokePremium(asset, spoke);
     require(drawnAmount <= drawn, SurplusAmountRestored(drawn));
-    require(premiumAmount <= premium, SurplusAmountRestored(premium));
```
DhairyaSethi, line 779: i don't think we can remove this check

src/hub/Hub.sol
```diff
@@ -218,13 +219,12 @@ contract Hub is IHub, AccessManaged {
...
     _validateRemove(asset, spoke, amount, to);
-    uint256 liquidity = asset.liquidity;
-    require(amount <= liquidity, InsufficientLiquidity(liquidity));
 
+    uint128 liquidity = asset.liquidity;
     uint128 shares = asset.toAddedSharesUp(amount).toUint128();
     asset.addedShares -= shares;
     spoke.addedShares -= shares;
-    asset.liquidity = liquidity.uncheckedSub(amount).toUint128();
+    asset.liquidity = liquidity - amount.toUint128();
```
DhairyaSethi: would revert this change, `liquidity` check can be shown to user also in normal paths

src/hub/Hub.sol
```diff
@@ -273,12 +272,13 @@ contract Hub is IHub, AccessManaged {
...
     _validateRestore(asset, spoke, drawnAmount, premiumAmount, from);
 
+    uint128 liquidity = asset.liquidity;
     uint128 drawnShares = asset.toDrawnSharesDown(drawnAmount).toUint128();
     asset.drawnShares -= drawnShares;
     spoke.drawnShares -= drawnShares;
     _applyPremiumDelta(asset, spoke, premiumDelta, premiumAmount);
     uint256 totalAmount = drawnAmount + premiumAmount;
-    asset.liquidity += totalAmount.toUint128();
+    asset.liquidity = liquidity + totalAmount.toUint128();
```
DhairyaSethi: no point in caching if it's only read once, pls revert

src/hub/Hub.sol
miguelmtzinf: why?

src/hub/Hub.sol
CheyenneAtapour: Reverting actually does add gas it seems

src/hub/Hub.sol
CheyenneAtapour: Done

src/hub/Hub.sol
CheyenneAtapour: Done

src/hub/Hub.sol
CheyenneAtapour: Done

src/hub/Hub.sol
DhairyaSethi: how much? it looks very bad, and kinda pointless. if it's less gas i would eat it to make the code cleaner

src/hub/Hub.sol
DhairyaSethi: it's required by repay in a sense, it restricts the maximum value for 'premiumAmount' which is used in _applyPremiumDelta

src/hub/Hub.sol
```diff
@@ -775,9 +777,7 @@ contract Hub is IHub, AccessManaged {
     require(spoke.active, SpokeNotActive());
     require(!spoke.paused, SpokePaused());
     uint256 drawn = _getSpokeDrawn(asset, spoke);
-    uint256 premium = _getSpokePremium(asset, spoke);
     require(drawnAmount <= drawn, SurplusAmountRestored(drawn));
```
DhairyaSethi, line 779: we can remove this check i think, why are we keeping this?

CheyenneAtapour: depends if we wanna restrict donations here, similar reasoning for this: https://github.com/aave/aave-v4/pull/869#discussion_r2443864852

src/hub/Hub.sol
CheyenneAtapour: like 60 gas. It's just 1 line and so not bad imo

DhairyaSethi: i would remove tbh

src/hub/Hub.sol
CheyenneAtapour: but isn't it already restricted with this? `require(premiumAfter + premiumAmount - premiumBefore <= 2, InvalidPremiumChange());`

Say everything is repaid, so `premiumAfter = 0` `premiumBefore = 10` and `premiumAmount = 13`, this will revert

src/hub/Hub.sol
CheyenneAtapour: @miguelmtzinf tiebreaker :)

src/hub/Hub.sol

src/hub/Hub.sol
DhairyaSethi: no it's not restricted with that, you can pass arbitrary deltas such that your premium after is actually very high, very high premiumAmount will bypass this check

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
yan-man: can you `expectRevert` with an address arg

src/hub/Hub.sol
```diff
@@ -654,10 +661,7 @@ contract Hub is IHub, AccessManaged {
     SpokeData storage receiver,
     uint256 shares
   ) internal {
-    uint256 addedShares = sender.addedShares;
-    require(shares <= addedShares, AddedSharesExceeded(addedShares));
-
-    sender.addedShares = addedShares.uncheckedSub(shares).toUint128();
+    sender.addedShares -= shares.toUint128();
```
yan-man, line 663: consistency, ig it was also mentioned on a Valentin pr before:
in add, it is just "shares" but in other places it is like "drawnShares" (ie draw, restore), and reportDeficit vs eliminateDeficit are inconsistent too. Maybe just "shares" everywhere? not a huge deal

CheyenneAtapour: I personally like the specifying prefix on shares, but it does get confusing to have `addedShares` twice.

But in any case, I'm not touching that convention here, can address outside this pr

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
CheyenneAtapour: Actually, the revert reason has changed now, updated this

src/hub/Hub.sol
miguelmtzinf: removed for code simplicity

src/hub/Hub.sol
miguelmtzinf: added back

src/hub/Hub.sol
```diff
@@ -738,8 +741,6 @@ contract Hub is IHub, AccessManaged {
     require(amount > 0, InvalidAmount());
     require(spoke.active, SpokeNotActive());
     require(!spoke.paused, SpokePaused());
-    uint256 removable = asset.toAddedAssetsDown(spoke.addedShares);
-    require(amount <= removable, AddedAmountExceeded(removable));
   }
 
   /// @dev Spoke with maximum cap have unlimited draw capacity.
```
DhairyaSethi, line 746: we can inline the base & premium debt calc on draw & restore 
& we can save on the index calc

miguelmtzinf: that d make it uglier, worse readability imo

Issue #782, opened 2025-09-22
Malicious Spoke Can Lock Assets In Hub


LHerskind:
likelihood: low
impact: high

# Description
A `Spoke` that is registered for a given `assetId` can call the `refreshPremium` function with almost free reign on the `PremiumDelta` values. 

This can be abused by the `Spoke` to update the `premiumShares` and `premiumOffset` such that they will overflow if any increase is made.

Since honest implementations of the `Spoke` will apply `_notifyRiskPremiumUpdate` on most function calls, which in turn execute `refreshPremium` for assets borrowed of the user. It can be made such that most calls made on behalf of  users borrowing the `assetId` will revert with an `SafeCastOverflowedUintDowncast` error during execution in the `Hub`.

While some actions can still be executed if they lower the the values sufficiently, the interest accrued makes the required value large and larger. 

The attack can potentially be undone by adding a "non-standard" `Spoke` that undoes the refresh by refreshing the opposite direction. But in the meantime, it can potentially be impossible to liquidate or withdraw.
# Proof of Concept

```solidity
// SPDX-License-Identifier: UNLICENSED
// Copyright (c) 2025 Aave Labs
pragma solidity ^0.8.0;

import "tests/unit/Spoke/SpokeBase.t.sol";
import {IHubBase} from "src/hub/interfaces/IHubBase.sol";
import {SafeCast} from "src/dependencies/openzeppelin/SafeCast.sol";

contract PremiumOffsetUnderflowTest is SpokeBase {
    using SharesMath for uint256;
    using WadRayMath for uint256;
    using PercentageMath for *;
    using SafeCast for uint256;

    struct GeneralLocalVars {
        uint256 usdxSupplyAmount;
        uint256 wethSupplyAmount;
        uint256 daiBorrowAmount;
        uint40 lastUpdateTimestamp;
        uint256 delay;
        uint256 expectedPremiumDebt;
        uint256 expectedpremiumShares;
        uint256 expectedUserRiskPremium;
    }

    function test_riskPremiumOverflow() public {
        GeneralLocalVars memory vars;
        // Start with collateral that gives us a specific risk premium
        vars.usdxSupplyAmount = 200e6; // 200 usd, 50% collateralRisk
        vars.daiBorrowAmount = 100e18; // 180 usd - close to max with USDX collateral

        ReserveIds memory reservesIds;
        reservesIds.usdx = _usdxReserveId(spoke1);
        reservesIds.weth = _wethReserveId(spoke1);
[...]

PR #870, opened 2025-10-07
feat: Limit premium on refresh premium

CheyenneAtapour:
- create a configurable risk premium cap limiting the amount of premium shares that a spoke can update when invoking `hub.refreshPremium`
- can be set to a `MAX_RISK_PREMIUM_THRESHOLD` sentinel value to ignore the restriction 
- prevent malicious spokes from being able to DOS other spokes by calling refreshPremium, such that the values of asset.premiumShares / asset.premiumOffset trigger an overflow when another spoke tries to increase them

closes https://github.com/aave/aave-v4/issues/782

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -624,9 +624,9 @@ contract Hub is IHub, AccessManaged {
     SpokeData storage spokeData = _spokes[assetId][spoke];
     return
       SpokeConfig({
-        riskPremiumCap: spokeData.riskPremiumCap,
         addCap: spokeData.addCap,
         drawCap: spokeData.drawCap,
+        riskPremiumCap: spokeData.riskPremiumCap,
         active: spokeData.active,
         paused: spokeData.paused
       });


src/hub/Hub.sol
```diff
@@ -697,6 +697,9 @@ contract Hub is IHub, AccessManaged {
     spoke.premiumOffset = spoke.premiumOffset.add(premium.offsetDelta).toUint128();
     spoke.realizedPremium = spoke.realizedPremium.add(premium.realizedDelta).toUint128();
 
+    // Asset premium shares cannot exceed max RP times drawn shares
+    require(asset.premiumShares <= asset.drawnShares.percentMulUp(1000_00), InvalidPremiumChange());
```
CheyenneAtapour: Would be nice to define this constant on the hub, instead of relying on pulling the spoke one, or having this `1000_00` as a magic number

DhairyaSethi: yep that's the idea; pls define as constant on the hub

DhairyaSethi: and this should be checked on spoke, not asset

tests/gas/Hub.Operations.gas.t.sol
```diff
@@ -79,13 +79,11 @@ contract HubOperations_Gas_Tests is Base {
...
-    int256 premiumOffset = hub1
-      .previewRestoreByShares(daiAssetId, uint256(premiumShares))
-      .toInt256();
-
+    Utils.supplyCollateral(spoke1, _daiReserveId(spoke1), alice, 1000e18, alice);
+    Utils.borrow(spoke1, _daiReserveId(spoke1), alice, 500e18, alice);
+    skip(100);
     vm.prank(address(spoke1));
-    hub1.refreshPremium(daiAssetId, IHubBase.PremiumDelta(premiumShares, premiumOffset, 0));
+    hub1.refreshPremium(daiAssetId, IHubBase.PremiumDelta(-1, -1, 1));
```
miguelmtzinf: why

src/hub/Hub.sol
```diff
@@ -24,7 +24,7 @@ contract Hub is IHub, AccessManaged {
   using SafeCast for uint256;
   using WadRayMath for uint256;
   using SharesMath for uint256;
-  using PercentageMath for uint256;
```
miguelmtzinf, line 26: where we were using this in the Hub?

tests/gas/Hub.Operations.gas.t.sol
CheyenneAtapour: I like this because it's a simple and valid case of `refreshPremium` which has nonzero for all parameters.

But this gas test has definitely been a contentious issue. Open to suggestions here!

src/hub/Hub.sol
CheyenneAtapour: Only usage I see is comparison with `PERCENTAGE_FACTOR`

src/hub/Hub.sol

src/hub/Hub.sol
```diff
@@ -37,6 +37,9 @@ contract Hub is IHub, AccessManaged {
   /// @inheritdoc IHub
   uint56 public constant MAX_ALLOWED_SPOKE_CAP = type(uint56).max;
 
+  /// @inheritdoc IHub
+  uint24 public constant MAX_ALLOWED_COLLATERAL_RISK = 1000_00; // 1000.00%
```
DhairyaSethi: this should be called risk premium i think; it's more consistent because spoke's report their risk premium with premium shares & not collateral risk (in any case it's more like an averaged collateral risk backing spoke's debt? but hub does not offer loans, only credit lines)

CheyenneAtapour: Alright, I'm not 100% sold, but I do like the name `risk premium` better in general, and I see your perspective. I changed, let's see if others complain

src/hub/Hub.sol
```diff
@@ -676,8 +679,9 @@ contract Hub is IHub, AccessManaged {
     receiver.addedShares += shares.toUint128();
   }
 
-  /// @dev Applies premium deltas on asset & spoke premium owed, also checks premium
-  /// does not increase by more than `premiumAmount`.
+  /// @dev Applies premium deltas on asset & spoke premium owed.
+  /// @dev Checks premium does not increase by more than `premiumAmount`.
+  /// @dev Checks spoke premium shares do not exceed max collateral risk times drawn shares.
```
```suggestion
  /// @dev Checks updated risk premium is within allowed limit. 
```

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -66,8 +71,12 @@ contract HubRefreshPremiumTest is HubBase {
     PremiumDataLocal memory premiumDataBefore = _loadAssetPremiumData(hub1, assetId);
     (, uint256 premiumBefore) = hub1.getAssetOwed(daiAssetId);
     bool reverting;
+    IHub.Asset memory asset = hub1.getAsset(assetId);
 
-    if (offsetDelta > sharesDelta) {
+    if ((asset.drawnShares).percentMulUp(1000_00) < asset.premiumShares + sharesDelta.toUint256()) {
```
DhairyaSethi:
```suggestion
    if (asset.drawnShares.percentMulUp(Constants.MAX_ALLOWED_RISK_PREMIUM) < asset.premiumShares + sharesDelta.toUint256()) {
```

src/hub/Hub.sol
```diff
@@ -37,6 +37,9 @@ contract Hub is IHub, AccessManaged {
   /// @inheritdoc IHub
   uint56 public constant MAX_ALLOWED_SPOKE_CAP = type(uint56).max;
 
+  /// @inheritdoc IHub
+  uint24 public constant MAX_ALLOWED_RISK_PREMIUM = 1000_00; // 1000.00%
```
DhairyaSethi: add this Constants.sol also and add test the two match

tests/gas/Hub.Operations.gas.t.sol
DhairyaSethi: existing test should work, no? it's expected spoke behaviour

CheyenneAtapour: Another idea here is maybe to move all the premium debt into realized?

CheyenneAtapour: Before, we were arbitrarily introducing premium shares and offset into `refreshPremium` without having anything drawn, which is certainly not expected behavior

src/hub/Hub.sol

tests/unit/Hub/Hub.RefreshPremium.t.sol
CheyenneAtapour: Done

src/hub/Hub.sol

DhairyaSethi:
fv properties passing: see #910

src/hub/interfaces/IHub.sol
```diff
@@ -378,4 +378,9 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev The value is not inclusive; using the maximum value indicates no cap.
   /// @return The maximum cap value, expressed in asset units.
   function MAX_ALLOWED_SPOKE_CAP() external view returns (uint56);
+
+  /// @notice Returns the maximum allowed risk premium.
+  /// @dev This is used to validate and limit premium values reported by spokes.
+  /// @return The maximum risk premium value, expressed in bps (e.g. 100_00 is 100.00%).
```
yan-man: nit: dont think we need the e.g. part, given we use BPS elsewhere

src/hub/Hub.sol
yan-man: ok kind of an issue here re: deps... so now MAX_ALLOWED_RISK_PREMIUM is dependent on the max collateral risk from spoke, as this max RP is based on assumption that user would have all their colls as max CR. 

so should we call this instead MAX_ALLOWED_COLLATERAL_RISK and instead have spoke read this from hub? and in spoke it is no longer a constant val that defines the max CR, but rather dependent on the hub that a given reserve is connected to, as this could vary cross hubs

tests/unit/Hub/Hub.Restore.t.sol
```diff
@@ -507,11 +507,12 @@ contract HubRestoreTest is HubBase {
     IHubBase.PremiumDelta memory premiumDelta = _getExpectedPremiumDelta({
       spoke: spoke1,
       user: alice,
-      reserveId: daiAssetId,
+      reserveId: _daiReserveId(spoke1),
       premiumRestored: premium
     });
-    premiumDelta.sharesDelta += vm.randomUint(0, 2).toInt256();
+    premiumDelta.sharesDelta += 2.toInt256();
```
yan-man: why setting to max here?

src/hub/Hub.sol
CheyenneAtapour: haha yeah @DhairyaSethi wanted me to change from `MAX_ALLOWED_COLLATERAL_RISK` to `MAX_ALLOWED_RISK_PREMIUM` here: https://github.com/aave/aave-v4/pull/870#discussion_r2422026684

Seems like too much work to have spoke read it from Hub, and nice to have the spokes self-sufficient. They should ofc be configured properly so we never have this issue anyway

tests/unit/Hub/Hub.Restore.t.sol
CheyenneAtapour: It reverts with 1 share as well, just did this to match the test name

src/hub/interfaces/IHub.sol
CheyenneAtapour: Done

tests/gas/Hub.Operations.gas.t.sol
DhairyaSethi: > we were arbitrarily introducing premium shares and offset into refreshPremium without having anything drawn, which is certainly not expected behavior

definitely is ser, spoke arbitrarily decides it's premium data; and it can be without anything drawn also

CheyenneAtapour: With this change, we cannot have premium shares without drawn shares

tests/unit/Hub/Hub.Restore.t.sol
yan-man: hm maybe should just be 1, as this is a shares delta rather than asset amt delta, so this is independent of that "2 wei" premium issue right, or do I misunderstand? why test w 2 wei shares specifically in that case

tests/unit/Hub/Hub.Restore.t.sol
```diff
@@ -545,7 +546,7 @@ contract HubRestoreTest is HubBase {
     IHubBase.PremiumDelta memory premiumDelta = _getExpectedPremiumDelta({
       spoke: spoke1,
       user: alice,
-      reserveId: daiAssetId,
+      reserveId: _daiReserveId(spoke1),
```
yan-man, line 549: hm this seems bad... is this saved just bc dai reserveId matched the dai asset id coincidentally initially and everything still worked?

tests/unit/Hub/Hub.Restore.t.sol
CheyenneAtapour: yeah, I'm not sure why this test was originally written, maybe it was meant to do 2 wei assets. But yeah I can change to 1 share to show smallest failing case

tests/unit/Hub/Hub.Restore.t.sol
CheyenneAtapour: Yes, that's exactly right. It should have been `reserveId` all this time

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -79,10 +85,23 @@ contract HubRefreshPremiumTest is HubBase {
     PremiumDataLocal memory premiumDataBefore = _loadAssetPremiumData(hub1, assetId);
```
yan-man, line 192: is this always 0 so it never goes into any of the test branches properly bc there is no asset premium data being loaded thru a borrow?

edit: should be about scenarios w/ drawnShares > 0 instead of prem data

CheyenneAtapour: Premium data does indeed start at 0. It's about the combination of the premium deltas that we are testing here

CheyenneAtapour: Actually, you were right, some checks were skipped because of no drawn, causing any premiumShares to revert in the MAX_ALLOWED_RISK_PREMIUM case. Fixed

src/hub/Hub.sol
```diff
@@ -690,6 +695,10 @@ contract Hub is IHub, AccessManaged {
       premium,
       premiumAmount
     );
+    require(
```
avniculae, line 714: what is preventing a malicious spoke from passing a very high premiumAmount to bypass this check?

tests/unit/Hub/Hub.RefreshPremium.t.sol
```diff
@@ -246,7 +273,14 @@ contract HubRefreshPremiumTest is HubBase {
...
     });
 
-    if (
+    uint256 expectedPremiumShares = premiumDelta.sharesDelta >= 0
+      ? asset.premiumShares + premiumDelta.sharesDelta.toUint256()
+      : asset.premiumShares - (-premiumDelta.sharesDelta).toUint256();
+
+    if (asset.drawnShares.percentMulUp(1000_00) < expectedPremiumShares) {
+      reverting = true;
+      vm.expectRevert(IHub.InvalidPremiumChange.selector);
```
avniculae, line 382: are there concrete (non-fuzzed) tests that test this new flow?

src/hub/Hub.sol
DhairyaSethi: they have to have that much premium debt to begin with, it's checked in _validateRestore against their premium debt

src/hub/Hub.sol
```diff
@@ -689,6 +709,13 @@ contract Hub is IHub, AccessManaged {
       premium,
       premiumAmount
     );
+
+    uint24 riskPremiumCap = spoke.riskPremiumCap;
+    require(
+      riskPremiumCap == MAX_ALLOWED_RISK_PREMIUM_CAP ||
+        spoke.premiumShares <= spoke.drawnShares.percentMulUp(riskPremiumCap),
```
miguelmtzinf, line 716: now we can partially block `refreshPremium` by setting `riskPremiumCap` to 0. :1st_place_medal:

DhairyaSethi: yes ✅

src/hub/Hub.sol
```diff
@@ -619,7 +622,14 @@ contract Hub is IHub, AccessManaged {
     address spoke
   ) external view returns (SpokeConfig memory) {
     SpokeData storage spokeData = _spokes[assetId][spoke];
-    return SpokeConfig(spokeData.active, spokeData.paused, spokeData.addCap, spokeData.drawCap);
+    return
+      SpokeConfig({
+        riskPremiumCap: spokeData.riskPremiumCap,
+        addCap: spokeData.addCap,
+        drawCap: spokeData.drawCap,
```
```suggestion
        addCap: spokeData.addCap,
        drawCap: spokeData.drawCap,
        riskPremiumCap: spokeData.riskPremiumCap,
```

src/hub/Hub.sol
```diff
@@ -628,7 +638,13 @@ contract Hub is IHub, AccessManaged {
     _updateSpokeConfig(
       assetId,
       feeReceiver,
-      SpokeConfig({active: true, paused: false, addCap: MAX_ALLOWED_SPOKE_CAP, drawCap: 0})
+      SpokeConfig({
+        active: true,
+        paused: false,
+        addCap: MAX_ALLOWED_SPOKE_CAP,
+        drawCap: 0,
+        riskPremiumCap: 0
```
```suggestion
        addCap: MAX_ALLOWED_SPOKE_CAP,
        drawCap: 0,
        riskPremiumCap: 0,
        active: true,
        paused: false
```

PR #872, opened 2025-10-07
fix : gateways spoke agnostic

Kogaroshi:
- Refactor Gateways PositionManagers to have a base for common functionalities, and make the Gateways spoke-agnostic.
- Allows for flexibility in being able to re-use the same gateway contracts across multiple spokes


src/position-manager/GatewayBase.sol
```diff
@@ -0,0 +1,48 @@
...
+    ISpoke(spoke).renouncePositionManagerRole(user);
+  }
+
+  /// @inheritdoc IGatewayBase
+  function isSpokeRegistered(address spoke) external view returns (bool) {
+    return _registeredSpokes[spoke];
+  }
+
+  function _validateSpoke(address spoke) internal view {
+    require(spoke != address(0), InvalidAddress());
```
DhairyaSethi: not needed, second require covers this

src/position-manager/GatewayBase.sol
```diff
@@ -0,0 +1,48 @@
...
+    require(user != address(0), InvalidAddress());
+    ISpoke(spoke).renouncePositionManagerRole(user);
+  }
+
+  /// @inheritdoc IGatewayBase
+  function isSpokeRegistered(address spoke) external view returns (bool) {
+    return _registeredSpokes[spoke];
+  }
+
+  function _validateSpoke(address spoke) internal view {
```
DhairyaSethi: make this virtual? 
and let's add a onlySpoke modifier instead which calls this, wdyt?

Kogaroshi: agree with virtual, and I'd say not modifier, to keep consistency with the rest of the codebase

src/position-manager/SignatureGateway.sol
```diff
@@ -69,207 +60,243 @@ contract SignatureGateway is
...
+        abi.encode(
+          SUPPLY_TYPEHASH,
+          params.spoke,
+          params.reserveId,
+          params.amount,
+          params.onBehalfOf,
+          params.nonce,
+          params.deadline
+        )
       )
```
DhairyaSethi: maybe let's abstract this in a library, with extends EIP712 types and does this hashing
such that if we ever want to optimise the hash, then we can do it in assembly which doesn't expand memory like https://github.com/Uniswap/v4-periphery/blob/60cd93803ac2b7fa65fd6cd351fd5fd4cc8c9db5/src/libraries/ERC721PermitHash.sol#L11-L39

src/position-manager/SignatureGateway.sol
```diff
@@ -69,207 +60,243 @@ contract SignatureGateway is
...
-    uint256 reserveId,
-    uint256 amount,
-    address onBehalfOf,
-    uint256 nonce,
-    uint256 deadline,
-    bytes calldata signature
-  ) external {
-    require(block.timestamp <= deadline, InvalidSignature());
+  function supplyWithSig(EIP712Types.Supply memory params, bytes calldata signature) external {
+    _validateSpoke(params.spoke);
```
DhairyaSethi: modifier makes more sense for this?

src/position-manager/SignatureGateway.sol
```diff
@@ -69,207 +60,243 @@ contract SignatureGateway is
...
     try
-      _spoke.setUserPositionManagerWithSig(address(this), user, approve, nonce, deadline, signature)
+      ISpoke(spoke).setUserPositionManagerWithSig(
+        address(this),
+        params.user,
+        params.approve,
+        params.nonce,
+        params.deadline,
+        signature
+      )
```
DhairyaSethi, line 172: this change also should be made on the spoke for consistency where it just takes the struct?

src/position-manager/SignatureGateway.sol
```diff
@@ -69,207 +60,243 @@ contract SignatureGateway is
...
     {} catch {}
   }
 
   /// @inheritdoc ISignatureGateway
   function permitReserve(
+    address spoke,
     uint256 reserveId,
```
DhairyaSethi, line 179: at this point we may as well just input underlying but current is safer ig

src/position-manager/interfaces/INativeTokenGateway.sol
```diff
@@ -26,37 +20,43 @@ interface INativeTokenGateway is IRescuable {
...
   /// @notice Wraps the native asset and supplies to the Spoke.
   /// @dev Contract must be an active & approved user position manager of the caller.
+  /// @param spoke The address of the spoke.
+  /// @param reserveId The identifier of the reserve for the wrapped asset.
+  /// @param amount Amount to wrap and supply.
+  function supplyNative(address spoke, uint256 reserveId, uint256 amount) external payable;
+
+  /// @notice Wraps the native asset,supplies to the Spoke and sets it as collateral.
+  /// @dev Contract must be an active & approved user position manager of the caller.
+  /// @param spoke The address of the spoke.
```
DhairyaSethi:
```suggestion
  /// @param spoke The address of a registered spoke.
```

src/position-manager/interfaces/INativeTokenGateway.sol
```diff
@@ -26,37 +20,43 @@ interface INativeTokenGateway is IRescuable {
 
   /// @notice Wraps the native asset and supplies to the Spoke.
   /// @dev Contract must be an active & approved user position manager of the caller.
+  /// @param spoke The address of the spoke.
+  /// @param reserveId The identifier of the reserve for the wrapped asset.
+  /// @param amount Amount to wrap and supply.
+  function supplyNative(address spoke, uint256 reserveId, uint256 amount) external payable;
+
+  /// @notice Wraps the native asset,supplies to the Spoke and sets it as collateral.
+  /// @dev Contract must be an active & approved user position manager of the caller.
```
DhairyaSethi, line 29: also mention on registered `spoke`?

src/position-manager/interfaces/ISignatureGateway.sol
```diff
@@ -4,158 +4,94 @@ pragma solidity ^0.8.0;
...
-interface ISignatureGateway is IMulticall, INoncesKeyed, IRescuable {
-  /// @notice Thrown when the given address is invalid.
-  error InvalidAddress();
-
+interface ISignatureGateway is IMulticall, INoncesKeyed, IGatewayBase {
   /// @notice Thrown when signature deadline has passed or signer is not `onBehalfOf`.
   error InvalidSignature();
 
-  /// @notice Facilitates `supply` action on connected SPOKE() with a typed signature from `onBehalfOf`.
+  /// @notice Facilitates `supply` action on the given spoke with a typed signature from `onBehalfOf`.
```
DhairyaSethi:
```suggestion
  /// @notice Facilitates `supply` action on the specified `spoke` with a typed signature from `onBehalfOf`.
```
that's the proposition we've been using so far

tests/unit/misc/GatewayBase.t.sol
```diff
@@ -0,0 +1,127 @@
...
+    assertEq(gateway.owner(), address(ADMIN));
+    assertEq(gateway.pendingOwner(), address(0));
+
+    assertEq(gateway.rescueGuardian(), address(ADMIN));
+  }
+
+  function test_registerSpoke_fuzz(address newSpoke) public {
+    assertFalse(gateway.isSpokeRegistered(newSpoke));
+
+    vm.expectEmit(true, true, true, true);
```
DhairyaSethi:
```suggestion
    vm.expectEmit();
```

tests/unit/misc/SignatureGateway/SignatureGateway.Reverts.InvalidSignature.t.sol
```diff
@@ -5,13 +5,20 @@ pragma solidity ^0.8.0;
 import 'tests/unit/misc/SignatureGateway/SignatureGateway.Base.t.sol';
 
 contract SignatureGatewayInvalidSignatureTest is SignatureGatewayBaseTest {
+  function setUp() public virtual override {
+    super.setUp();
+
+    vm.prank(address(ADMIN));
+    gateway.registerSpoke(address(spoke1), true);
+  }
```
DhairyaSethi: instead let's register in BaseSetup only and disable it on the revertsWith InvalidSpoke file

tests/unit/misc/SignatureGateway/SignatureGateway.Reverts.SpokeNotRegistered.t.sol
```diff
@@ -0,0 +1,95 @@
...
+
+    assertTrue(spoke1.isPositionManagerActive(address(gateway)));
+    assertTrue(spoke1.isPositionManager(alice, address(gateway)));
+  }
+
+  function test_supplyWithSig_revertsWith_SpokeNotRegistered() public {
+    EIP712Types.Supply memory p = _supplyData(spoke1, alice, _warpBeforeRandomDeadline());
+    bytes memory signature = _sign(alicePk, _getTypedDataHash(gateway, p));
+
+    vm.expectRevert(IGatewayBase.SpokeNotRegistered.selector);
```
DhairyaSethi: in these tests the signature can just be random bytes no? doesn't need to be correct; nbd either way

src/position-manager/GatewayBase.sol
DhairyaSethi: we have `onlyPositionManager` on spoke which is sort of similar

DhairyaSethi: function x(..) external {
  _validate(..);
  ...
}
looks very ugly, it's prime modifier candidate imo

src/position-manager/SignatureGateway.sol
DhairyaSethi: we should do some gas snapshots as well, but not needed for now; we can log in a ticket & do over the week after this gets merged

src/position-manager/GatewayBase.sol
Kogaroshi: okay, let's make it a modifier there

src/position-manager/GatewayBase.sol
```diff
@@ -0,0 +1,48 @@
...
+
+import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
+import {Rescuable} from 'src/utils/Rescuable.sol';
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+import {IGatewayBase} from 'src/position-manager/interfaces/IGatewayBase.sol';
+
+/// @title GatewayBase
+/// @author Aave Labs
+/// @notice Base implementation for gateway common functionalities.
+abstract contract GatewayBase is IGatewayBase, Rescuable, Ownable2Step {
```
DhairyaSethi, line 13: so we have `Rescuable` as a dep because the two position manager's we've built should not hold funds at any point, and any sent must be by accident or malicious. but could we have `gateways` (less likely) or `position managers` (more likely)  which inherit this who may hold funds and we don't want rescuable functionality?

for this what we can do is make the functionality optional, by exposing an internal method `_enableRescuable()` which returns true or false; and makes Rescuable ops conditional OR override them here to disable

DhairyaSethi: some examples of the idea in the wild
https://github.com/Vectorized/solady/blob/main/src/utils/EIP712.sol#L62
https://github.com/Vectorized/solady/blob/main/src/tokens/ERC20.sol#L187
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC4626.sol#L130

Kogaroshi: I made this with the assumption for a base purely for gateways, so contracts that should not hold funds, and that PositionManagers would have a different Base contract with potentially different dependencies.

DhairyaSethi: yeah it doesn't seem likely gateways should ever be holding left over funds

src/position-manager/SignatureGateway.sol
Kogaroshi: https://github.com/aave/aave-v4/issues/880

src/position-manager/SignatureGateway.sol
Kogaroshi: We'd have to import `EIP712Types` in the Spoke, and add a check in this method to ensure `params.positionManager` is the SigGateway instead of just replacing the parameter, do we want to or not worth it ?

DhairyaSethi: it should be consistent across both imo, fine importing EIP712Types on the spoke

src/libraries/cryptography/EIP712Hash.sol
```diff
@@ -0,0 +1,145 @@
...
+
+  bytes32 public constant UPDATE_USER_RISK_PREMIUM_TYPEHASH =
+    // keccak256('UpdateUserRiskPremium(address spoke,address user,uint256 nonce,uint256 deadline)')
+    0xb41e132023782c9b02febf1b9b7fe98c4a73f57ebc63ba44cd71f6365ea09eaf;
+
+  bytes32 public constant UPDATE_USER_DYNAMIC_CONFIG_TYPEHASH =
+    // keccak256('UpdateUserDynamicConfig(address spoke,address user,uint256 nonce,uint256 deadline)')
+    0xba177b1f5b5e1e709f62c19f03c97988c57752ba561de58f383ebee4e8d0a71c;
+
+  function hashSupply(EIP712Types.Supply memory params) internal pure returns (bytes32) {
```
DhairyaSethi: would rename all to just `hash`

src/position-manager/NativeTokenGateway.sol
```diff
@@ -75,23 +74,32 @@ contract NativeTokenGateway is
...
 
   /// @inheritdoc INativeTokenGateway
-  function borrowNative(uint256 reserveId, uint256 amount, address receiver) external {
-    (IERC20 underlying, ) = _getReserveData(reserveId);
+  function borrowNative(
+    address spoke,
+    uint256 reserveId,
+    uint256 amount
+  ) external onlyRegisteredSpoke(spoke) {
+    ISpoke _spoke = ISpoke(spoke);
```
DhairyaSethi: internal var with underscore is vv weird :(

src/position-manager/SignatureGateway.sol
```diff
@@ -2,283 +2,199 @@
...
     bytes calldata signature
-  ) external {
-    require(block.timestamp <= deadline, InvalidSignature());
-    bytes32 hash = _hashTypedData(
-      keccak256(
-        abi.encode(SUPPLY_TYPEHASH, address(_spoke), reserveId, amount, onBehalfOf, nonce, deadline)
-      )
+  ) external onlyRegisteredSpoke(params.spoke) {
+    require(block.timestamp <= params.deadline, InvalidSignature());
+    bytes32 hash = _hashTypedData(EIP712Hash.hashSupply(params));
```
DhairyaSethi:
```suggestion
    bytes32 hash = _hashTypedData(params.hash());
```

Kogaroshi: Yeah, wanted to discuss on this syntax instead, gonna make the changes there

src/position-manager/SignatureGateway.sol
```diff
@@ -2,283 +2,199 @@
...
+  constructor(address initialOwner_) GatewayBase(initialOwner_) {}
 
   /// @inheritdoc ISignatureGateway
   function supplyWithSig(
-    uint256 reserveId,
-    uint256 amount,
-    address onBehalfOf,
-    uint256 nonce,
-    uint256 deadline,
+    EIP712Types.Supply memory params,
```
DhairyaSethi:
```suggestion
    EIP712Types.Supply calldata params,
```
even tho at some point we're pulling this whole thing in memory anyway, looks better for now and once we optimise the hashing we would like to keep this in calldata and bring to memory on our terms avoiding expansion

src/position-manager/GatewayBase.sol
```diff
@@ -0,0 +1,63 @@
...
+  /// @return The corresponding hub address.
+  function _getReserveData(
+    ISpoke spoke,
+    uint256 reserveId
+  ) internal view returns (IERC20, address) {
+    ISpoke.Reserve memory reserveData = spoke.getReserve(reserveId);
+    return (IERC20(reserveData.underlying), address(reserveData.hub));
+  }
+
+  /// @dev RescueGuardian is the owner of the contract.
```
miguelmtzinf: what does this mean

src/spoke/Spoke.sol
```diff
@@ -409,29 +407,18 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
   function setUserPositionManagerWithSig(
-    address positionManager,
-    address user,
-    bool approve,
-    uint256 nonce,
-    uint256 deadline,
+    EIP712Types.SetUserPositionManager calldata params,
     bytes calldata signature
   ) external {
-    require(block.timestamp <= deadline, InvalidSignature());
```
miguelmtzinf, line 425: I d prefer to keep this independent from posmanagers

DhairyaSethi: it's a change i suggested to start using the structs in the fn call as well for the 712 typed intents since now nonce is also passed as calldata and checked later, that's why the same convention is followed here for consistency

src/position-manager/NativeTokenGateway.sol
```diff
@@ -52,95 +40,114 @@ contract NativeTokenGateway is
...
+    _validateParams(underlying, amount_);
+    require(msg.value == amount_, NativeAmountMismatch());
+
+    _nativeWrapper.deposit{value: amount_}();
+    _nativeWrapper.forceApprove(hub, amount_);
+    spoke.supply(reserveId_, amount_, user_);
+
+    if (enableCollateral_) {
+      spoke.setUsingAsCollateral(reserveId_, true, user_);
+    }
```
miguelmtzinf: this could be placed out of the func and be called directly when needed

src/position-manager/NativeTokenGateway.sol
```diff
@@ -52,95 +40,114 @@ contract NativeTokenGateway is
...
-  function supplyNative(uint256 reserveId, uint256 amount) external payable nonReentrant {
-    (IERC20 underlying, address hub) = _getReserveData(reserveId);
-    _validateParams(underlying, amount);
-    require(msg.value == amount, NativeAmountMismatch());
-
-    _nativeWrapper.deposit{value: amount}();
-    _nativeWrapper.forceApprove(hub, amount);
-    _spoke.supply(reserveId, amount, msg.sender);
+  function supplyNative(
+    address spoke_,
```
miguelmtzinf: we only use this convention for constructor input params

src/libraries/cryptography/EIP712Hash.sol
```diff
@@ -0,0 +1,157 @@
...
+
+  bytes32 public constant UPDATE_USER_DYNAMIC_CONFIG_TYPEHASH =
+    // keccak256('UpdateUserDynamicConfig(address spoke,address user,uint256 nonce,uint256 deadline)')
+    0xba177b1f5b5e1e709f62c19f03c97988c57752ba561de58f383ebee4e8d0a71c;
+
+  bytes32 public constant SET_USER_POSITION_MANAGER_TYPEHASH =
+    // keccak256('SetUserPositionManager(address positionManager,address user,bool approve,uint256 nonce,uint256 deadline)')
+    0x758d23a3c07218b7ea0b4f7f63903c4e9d5cbde72d3bcfe3e9896639025a0214;
+
+  function hash(EIP712Types.Supply memory params) internal pure returns (bytes32) {
```
DhairyaSethi: all of these need to be calldata too when we start optimising

src/position-manager/NativeTokenGateway.sol
```diff
@@ -7,38 +7,26 @@ import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.
 import {SafeERC20, IERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
```
miguelmtzinf, line 6: rm ownable2step

src/libraries/cryptography/EIP712Hash.sol
DhairyaSethi: rn i think it's double mem alloc, once for this & once for abi.encode

src/position-manager/GatewayBase.sol
```diff
@@ -0,0 +1,63 @@
...
+  /// @inheritdoc IGatewayBase
+  function registerSpoke(address spoke, bool active) external onlyOwner {
+    require(spoke != address(0), InvalidAddress());
+    _registeredSpokes[spoke] = active;
+    emit SpokeRegistered(spoke, active);
+  }
+
+  /// @inheritdoc IGatewayBase
+  function renouncePositionManagerRole(address spoke, address user) external onlyOwner {
+    _isSpokeValid(spoke);
```
DhairyaSethi: modifier

src/spoke/Spoke.sol
miguelmtzinf: to be clear, I was referring to the Spoke. I d like him to stay independent from position managers and this new library, which the majority of fns are related to SignatureGateway

src/position-manager/GatewayBase.sol
Kogaroshi: that the role of RescueGuardian is given to the contract `owner`, I'll reword this comment

src/position-manager/SignatureGateway.sol
```diff
@@ -2,283 +2,200 @@
...
   }
 
   /// @inheritdoc ISignatureGateway
   function setSelfAsUserPositionManagerWithSig(
-    address user,
-    bool approve,
-    uint256 nonce,
-    uint256 deadline,
+    address spoke,
+    EIP712Types.SetUserPositionManager calldata params,
```
miguelmtzinf, line 161: should `spoke` be part of the sig so intent cannot be tampered (spoke replaced)?

miguelmtzinf: Oh, it's fine because the vefiring contract is the Spoke itself and we are just passing the intent data

src/position-manager/NativeTokenGateway.sol
```diff
@@ -105,42 +113,40 @@ contract NativeTokenGateway is
...
-  /// @return The corresponding hub address.
-  function _getReserveData(uint256 reserveId) internal view returns (IERC20, address) {
-    ISpoke.Reserve memory reserveData = _spoke.getReserve(reserveId);
-    return (IERC20(reserveData.underlying), address(reserveData.hub));
+  function _supplyNative(
+    address spokeAddress,
+    uint256 reserveId,
+    uint256 amount,
+    address user
+  ) internal {
```
miguelmtzinf: these 2 internal functions must be above any internal view (`_validateParams`)

src/position-manager/NativeTokenGateway.sol
```diff
@@ -105,42 +113,40 @@ contract NativeTokenGateway is
...
     require(amount > 0, InvalidAmount());
   }
 
-  /// @return The underlying asset for `reserveId` on connected spoke.
-  /// @return The corresponding hub address.
-  function _getReserveData(uint256 reserveId) internal view returns (IERC20, address) {
-    ISpoke.Reserve memory reserveData = _spoke.getReserve(reserveId);
-    return (IERC20(reserveData.underlying), address(reserveData.hub));
+  function _supplyNative(
+    address spokeAddress,
```
miguelmtzinf:
```suggestion
    address spoke,
```
we tend to use `spoke` for spoke's address and `spokeData` for spoke's struct

src/position-manager/NativeTokenGateway.sol
```diff
@@ -105,42 +113,40 @@ contract NativeTokenGateway is
...
+    (IERC20 underlying, address hub) = _getReserveData(spoke, reserveId);
+    _validateParams(underlying, amount);
+    require(msg.value == amount, NativeAmountMismatch());
+
+    _nativeWrapper.deposit{value: amount}();
+    _nativeWrapper.forceApprove(hub, amount);
+    spoke.supply(reserveId, amount, user);
+  }
+
+  function _setUsingAsCollateral(address spokeAddress, uint256 reserveId, address user) internal {
```
miguelmtzinf: this is only used once so i d rm it

src/position-manager/NativeTokenGateway.sol
```diff
@@ -105,42 +113,40 @@ contract NativeTokenGateway is
...
+  function _supplyNative(
+    address spokeAddress,
+    uint256 reserveId,
+    uint256 amount,
+    address user
+  ) internal {
+    ISpoke spoke = ISpoke(spokeAddress);
+    (IERC20 underlying, address hub) = _getReserveData(spoke, reserveId);
+    _validateParams(underlying, amount);
+    require(msg.value == amount, NativeAmountMismatch());
```
miguelmtzinf: personal preference is not having msg.value checks on internal fns... I d move this check to the v begining of the fn, same on repayment.

src/position-manager/NativeTokenGateway.sol
```diff
@@ -105,42 +113,40 @@ contract NativeTokenGateway is
...
-  /// @return The underlying asset for `reserveId` on connected spoke.
-  /// @return The corresponding hub address.
-  function _getReserveData(uint256 reserveId) internal view returns (IERC20, address) {
-    ISpoke.Reserve memory reserveData = _spoke.getReserve(reserveId);
-    return (IERC20(reserveData.underlying), address(reserveData.hub));
+  function _supplyNative(
+    address spokeAddress,
+    uint256 reserveId,
+    uint256 amount,
+    address user
```
miguelmtzinf:
```suggestion,
    address user,
    uint256 amount
```
reads better imho

src/libraries/cryptography/EIP712Hash.sol
```diff
@@ -0,0 +1,161 @@
...
+          params.spoke,
+          params.user,
+          params.nonce,
+          params.deadline
+        )
+      );
+  }
+
+  function hash(
+    EIP712Types.SetUserPositionManager calldata params
```
miguelmtzinf: not used anymore in src. Also, should we move this lib to `position-manager/libraries`? Do we foresee this lib being used by multiple posm contracts?

src/position-manager/NativeTokenGateway.sol
```diff
@@ -52,50 +39,71 @@ contract NativeTokenGateway is
...
   /// @inheritdoc INativeTokenGateway
-  function borrowNative(uint256 reserveId, uint256 amount, address receiver) external {
-    (IERC20 underlying, ) = _getReserveData(reserveId);
+  function borrowNative(
+    address spokeAddress,
+    uint256 reserveId,
+    uint256 amount
+  ) external onlyRegisteredSpoke(spokeAddress) {
+    ISpoke spoke = ISpoke(spokeAddress);
+    (IERC20 underlying, address hub) = _getReserveData(spoke, reserveId);
```
miguelmtzinf:
```suggestion
    (IERC20 underlying, ) = _getReserveData(spoke, reserveId);
```

src/position-manager/NativeTokenGateway.sol
```diff
@@ -52,50 +39,71 @@ contract NativeTokenGateway is
...
   /// @inheritdoc INativeTokenGateway
-  function withdrawNative(uint256 reserveId, uint256 amount, address receiver) external {
-    (IERC20 underlying, ) = _getReserveData(reserveId);
+  function withdrawNative(
+    address spokeAddress,
+    uint256 reserveId,
+    uint256 amount
+  ) external onlyRegisteredSpoke(spokeAddress) {
+    ISpoke spoke = ISpoke(spokeAddress);
+    (IERC20 underlying, address hub) = _getReserveData(spoke, reserveId);
```
miguelmtzinf:
```suggestion
    (IERC20 underlying, ) = _getReserveData(spoke, reserveId);
```

src/position-manager/NativeTokenGateway.sol
```diff
@@ -52,50 +39,71 @@ contract NativeTokenGateway is
...
+  ) external payable nonReentrant onlyRegisteredSpoke(spokeAddress) {
+    _supplyNative(spokeAddress, reserveId, amount, msg.sender);
+  }
 
-    _nativeWrapper.deposit{value: amount}();
-    _nativeWrapper.forceApprove(hub, amount);
-    _spoke.supply(reserveId, amount, msg.sender);
+  /// @inheritdoc INativeTokenGateway
+  function supplyAndCollateralNative(
+    address spokeAddress,
```
miguelmtzinf:
```suggestion
    address spoke,
```

snapshots/NativeTokenGateway.Operations.json
```diff
@@ -0,0 +1,9 @@
+{
+  "borrowNative": "333087",
+  "repayNative": "262109",
+  "supplyAndCollateralNative: initial": "206398",
+  "supplyNative: initial": "176665",
+  "supplyNative: second time": "132132",
```
DhairyaSethi: need only one type of this
also pls make all actions the second one, ie do a normal borrow on user's account such that its storage is warm and then snapshot with gateway

DhairyaSethi: for all tests

src/position-manager/NativeTokenGateway.sol
```diff
@@ -3,42 +3,29 @@
 pragma solidity 0.8.28;
 
 import {ReentrancyGuardTransient} from 'src/dependencies/openzeppelin/ReentrancyGuardTransient.sol';
-import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
 import {SafeERC20, IERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
 import {Address} from 'src/dependencies/openzeppelin/Address.sol';
 import {MathUtils} from 'src/libraries/math/MathUtils.sol';
-import {Rescuable} from 'src/utils/Rescuable.sol';
 import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+import {GatewayBase} from 'src/position-manager/GatewayBase.sol';
```
yan-man: maybe can move this up to group interfaces together

src/position-manager/interfaces/INativeTokenGateway.sol
```diff
@@ -24,39 +18,45 @@ interface INativeTokenGateway is IRescuable {
...
+  /// @param amount Amount to wrap and supply.
+  function supplyNative(address spokeAddress, uint256 reserveId, uint256 amount) external payable;
+
+  /// @notice Wraps the native asset,supplies to a specified registered `spoke` and sets it as collateral.
   /// @dev Contract must be an active & approved user position manager of the caller.
+  /// @param spokeAddress The address of the registered `spoke`.
   /// @param reserveId The identifier of the reserve for the wrapped asset.
   /// @param amount Amount to wrap and supply.
-  function supplyNative(uint256 reserveId, uint256 amount) external payable;
+  function supplyAndCollateralNative(
```
yan-man: maybe `supplyAsCollateralNative`

src/position-manager/NativeTokenGateway.sol
Kogaroshi: but we also want to use `spoke` to the ISpoke casted var, what name should we use then ?

src/libraries/cryptography/EIP712Hash.sol
Kogaroshi: It was kept to cover all structs in EIP712Types, for later usecases. But we can remove for now if we prefer.

src/position-manager/GatewayBase.sol
```diff
@@ -0,0 +1,67 @@
...
+    require(spoke != address(0), InvalidAddress());
+    _registeredSpokes[spoke] = active;
+    emit SpokeRegistered(spoke, active);
+  }
+
+  /// @inheritdoc IGatewayBase
+  function renouncePositionManagerRole(
+    address spoke,
+    address user
+  ) external onlyOwner onlyRegisteredSpoke(spoke) {
```
DhairyaSethi:
```suggestion
  ) external onlyOwner {
```
this probably doesn't need the modifier since its admin, can see a scenario that someone approved on an unregistered (could be previously registered) and admin wants to cleanup

src/position-manager/NativeTokenGateway.sol
```diff
@@ -105,42 +112,29 @@ contract NativeTokenGateway is
...
   /// @inheritdoc INativeTokenGateway
   function NATIVE_WRAPPER() external view returns (address) {
     return address(_nativeWrapper);
   }
 
-  /// @inheritdoc INativeTokenGateway
-  function SPOKE() external view returns (address) {
-    return address(_spoke);
-  }
+  function _supplyNative(address spoke, uint256 reserveId, address user, uint256 amount) internal {
```
DhairyaSethi, line 128: pls add dev comment above that msg.value == amount check to be done before

DhairyaSethi: especially since this only does part of the validations (ie param)

src/position-manager/SignatureGateway.sol
```diff
@@ -2,283 +2,196 @@
...
 
-    (IERC20 underlying, address hub) = _getReserveData(reserveId);
-    underlying.safeTransferFrom(onBehalfOf, address(this), amount);
-    underlying.forceApprove(hub, amount);
+    (IERC20 underlying, address hub) = _getReserveData(params.spoke, params.reserveId);
+    underlying.safeTransferFrom(params.onBehalfOf, address(this), params.amount);
+    underlying.forceApprove(hub, params.amount);
 
-    _spoke.supply(reserveId, amount, onBehalfOf);
+    ISpoke(params.spoke).supply(params.reserveId, params.amount, params.onBehalfOf);
```
DhairyaSethi: caching spoke on the stack might be cheaper for all of these ops? since it's used to often

DhairyaSethi: dup is cheaper than push calldataload

src/position-manager/SignatureGateway.sol
```diff
@@ -2,283 +2,196 @@
...
+  ) external onlyRegisteredSpoke(params.spoke) {
+    require(block.timestamp <= params.deadline, InvalidSignature());
+    bytes32 hash = _hashTypedData(params.hash());
+    require(
+      SignatureChecker.isValidSignatureNow(params.onBehalfOf, hash, signature),
+      InvalidSignature()
     );
-    require(SignatureChecker.isValidSignatureNow(onBehalfOf, hash, signature), InvalidSignature());
-    _useCheckedNonce(onBehalfOf, nonce);
+    _useCheckedNonce(params.onBehalfOf, params.nonce);
```
DhairyaSethi: i realised now permit2 uses checked nonce before sig validation, 
note this can be relevant bc sig validation can potently do an external call when it's 1271 case 

i had internally put it after, but lmk your thoughts @Kogaroshi @miguelmtzinf

tests/unit/misc/GatewayBase.t.sol
```diff
@@ -0,0 +1,123 @@
...
+    vm.prank(ADMIN);
+    gateway.registerSpoke(address(spoke1), true);
+
+    vm.expectRevert(IGatewayBase.InvalidAddress.selector);
+    vm.prank(ADMIN);
+    gateway.renouncePositionManagerRole(address(spoke1), address(0));
+  }
+
+  function test_renouncePositionManagerRole_revertsWith_SpokeNotRegistered() public {
+    (address user, uint256 userPk) = makeAddrAndKey(string(vm.randomBytes(32)));
```
DhairyaSethi: user pk not needed 
```suggestion
    address user = vm.randomAddress();
```

tests/unit/misc/GatewayBase.t.sol
```diff
@@ -0,0 +1,123 @@
...
+    assertEq(gateway.owner(), address(ADMIN));
+    assertEq(gateway.pendingOwner(), address(0));
+
+    assertEq(gateway.rescueGuardian(), address(ADMIN));
+  }
+
+  function test_registerSpoke_fuzz(address newSpoke) public {
+    assertFalse(gateway.isSpokeRegistered(newSpoke));
+
+    vm.expectEmit();
```
DhairyaSethi:
```suggestion
    vm.expectEmit(address(gateway));
```

DhairyaSethi: for all of these

tests/unit/misc/GatewayBase.t.sol
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  function test_registerSpoke_revertsWith_InvalidAddress() public {
+    vm.expectRevert(IGatewayBase.InvalidAddress.selector);
+    vm.prank(ADMIN);
+    gateway.registerSpoke(address(0), true);
+  }
+
+  function test_renouncePositionManagerRole() public {
+    (address user, uint256 userPk) = makeAddrAndKey(string(vm.randomBytes(32)));
```
DhairyaSethi: userPk not needed in any of them, would revert to randomAddress

tests/unit/NativeTokenGateway.t.sol
```diff
@@ -41,36 +35,8 @@ contract NativeTokenGatewayTest is Base {
...
-
-  function test_renouncePositionManagerRole_revertsWith_OwnableUnauthorizedAccount() public {
-    (address user, ) = makeAddrAndKey(string(vm.randomBytes(32)));
-
-    vm.prank(user);
-    spoke1.setUserPositionManager(address(nativeTokenGateway), true);
-
-    vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, user));
-    vm.prank(user);
-    nativeTokenGateway.renouncePositionManagerRole(user);
```
DhairyaSethi, line 73: we could have still kept all of these tests even though they're already tested with gateway wrapper just to confirm that it inherits all functionality correctly (doesn't override for ex) since we already had these written but nbd

tests/unit/misc/NativeTokenGateway.t.sol
```diff
@@ -101,29 +67,83 @@ contract NativeTokenGatewayTest is Base {
...
+
+  function test_supplyNative_revertsWith_SpokeNotRegistered() public {
+    uint256 amount = 100e18;
+    vm.expectRevert(IGatewayBase.SpokeNotRegistered.selector);
+    vm.prank(bob);
+    nativeTokenGateway.supplyNative{value: amount}(address(spoke2), _wethReserveId(spoke1), amount);
+
+    vm.expectRevert(IGatewayBase.SpokeNotRegistered.selector);
+    vm.prank(bob);
+    nativeTokenGateway.supplyNative{value: amount}(address(0), _wethReserveId(spoke1), amount);
```
DhairyaSethi, line 82: instead of two could make a helper 
fn _unregisteredSpoke(gateway) 
addr spoke = randomAddress()
while (gateway.isRegistered(spoke)) spoke = randomAddress() 


and you can make this test symbolic (ie add setArbitraryStorage at the start for gateway) 
(when proving with halmos we'll have to replace while loop with vm.assume())


but very overkill lol

tests/unit/misc/SignatureGateway/SignatureGateway.Reverts.SpokeNotRegistered.t.sol
```diff
@@ -0,0 +1,90 @@
...
+      abi.encodeWithSelector(IGatewayBase.SpokeNotRegistered.selector, address(gateway))
+    );
+    vm.prank(vm.randomAddress());
+    gateway.updateUserRiskPremiumWithSig(p, signature);
+  }
+
+  function test_updateUserDynamicConfigWithSig_revertsWith_SpokeNotRegistered(
+    EIP712Types.UpdateUserDynamicConfig memory p
+  ) public {
+    bytes memory signature = vm.randomBytes(vm.randomUint(1, 10));
```
DhairyaSethi:
```suggestion
    bytes memory signature = vm.randomBytes(vm.randomUint(72));
```

DhairyaSethi: don't like limiting to 1-10

src/position-manager/SignatureGateway.sol
Kogaroshi: We could move it before the sig validation to cover those cases yeah, no strong preference

DhairyaSethi: i mean i intentionally put it after bc i thought the smart wallets, if needed to query back into this contract, should receive the last USED nonce. this current nonce hasn't been used bc they haven't confirmed it yet

Kogaroshi: Yeah that makes sense to only use the nonce once the sig is confirmed. What's the impact of EIP-1271 using contract to make another external call where the nonce is used during this context ?
Even with multiple signature that use the same nonce, `_useCheckedNonce` would fail after the 1st validation.

src/position-manager/GatewayBase.sol
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  /// @return The underlying asset for `reserveId` on the specified spoke.
+  /// @return The corresponding hub address.
+  function _getReserveData(
+    address spoke,
+    uint256 reserveId
+  ) internal view returns (IERC20, address) {
+    ISpoke.Reserve memory reserveData = ISpoke(spoke).getReserve(reserveId);
+    return (IERC20(reserveData.underlying), address(reserveData.hub));
```
miguelmtzinf: feels simpler to return underlying as address here (rm ierc20 import)

src/position-manager/NativeTokenGateway.sol
```diff
@@ -105,42 +112,30 @@ contract NativeTokenGateway is
...
-  /// @dev RescueGuardian is the owner of the contract.
-  function _rescueGuardian() internal view override returns (address) {
-    return owner();
+    _nativeWrapper.deposit{value: amount}();
+    _nativeWrapper.forceApprove(hub, amount);
+    ISpoke(spoke).supply(reserveId, amount, user);
   }
 
   function _validateParams(IERC20 underlying, uint256 amount) internal view {
     require(address(underlying) == address(_nativeWrapper), NotNativeWrappedAsset());
```
miguelmtzinf:
```suggestion
    require(address(_nativeWrapper) == address(underlying), NotNativeWrappedAsset());
```
stored vs input, i think it's what we have been doing across the code

src/position-manager/NativeTokenGateway.sol
```diff
@@ -52,50 +39,70 @@ contract NativeTokenGateway is
...
+    address spoke,
+    uint256 reserveId,
+    uint256 amount
+  ) external payable nonReentrant onlyRegisteredSpoke(spoke) {
     require(msg.value == amount, NativeAmountMismatch());
+    (IERC20 underlying, address hub) = _getReserveData(spoke, reserveId);
+    _validateParams(underlying, amount);
 
-    uint256 userDebtAmount = _spoke.getUserTotalDebt(reserveId, msg.sender);
+    uint256 userDebtAmount = ISpoke(spoke).getUserTotalDebt(reserveId, msg.sender);
```
miguelmtzinf:
```suggestion
    uint256 userTotalDebt = ISpoke(spoke).getUserTotalDebt(reserveId, msg.sender);
```
or `userDebt`

src/position-manager/interfaces/INativeTokenGateway.sol
```diff
@@ -24,39 +18,45 @@ interface INativeTokenGateway is IRescuable {
...
+  /// @param spoke The address of the registered `spoke`.
   /// @param reserveId The identifier of the reserve for the wrapped asset.
   /// @param amount Amount to borrow and unwrap.
-  /// @param receiver Address that will receive the unwrapped native asset.
-  function borrowNative(uint256 reserveId, uint256 amount, address receiver) external;
+  function borrowNative(address spoke, uint256 reserveId, uint256 amount) external;
 
-  /// @notice Wraps the native asset and repays debt on the Spoke.
+  /// @notice Wraps the native asset and repays debt on a specified registered `spoke`.
   /// @dev Contract must be an active & approved user position manager of the caller.
```
miguelmtzinf, line 55:
```suggestion
  /// @dev It refunds any excess funds sent beyond the required debt repayment.
  /// @dev Contract must be an active & approved user position manager of the caller.
```

src/position-manager/SignatureGateway.sol
```diff
@@ -2,283 +2,190 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity 0.8.28;
 
-import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
 import {SignatureChecker} from 'src/dependencies/openzeppelin/SignatureChecker.sol';
 import {IERC20Permit} from 'src/dependencies/openzeppelin/IERC20Permit.sol';
 import {SafeERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
 import {IERC20} from 'src/dependencies/openzeppelin/IERC20.sol';
```
miguelmtzinf:
```suggestion
import {SafeERC20, IERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
```

src/position-manager/SignatureGateway.sol
```diff
@@ -2,283 +2,190 @@
...
 /// multicall can be executed independently in order of signed nonce & deadline; does not guarantee batch atomicity.
-contract SignatureGateway is
-  ISignatureGateway,
-  NoncesKeyed,
-  Multicall,
-  Rescuable,
-  Ownable2Step,
-  EIP712
-{
+contract SignatureGateway is ISignatureGateway, NoncesKeyed, Multicall, GatewayBase, EIP712 {
```
miguelmtzinf: we can move `GatewayBase` at the very beggining in both posmanagers so they have same storageLayout
```suggestion
contract SignatureGateway is ISignatureGateway, GatewayBase, NoncesKeyed, Multicall, EIP712 {
```

src/position-manager/NativeTokenGateway.sol
```diff
@@ -3,11 +3,10 @@
...
 import {ReentrancyGuardTransient} from 'src/dependencies/openzeppelin/ReentrancyGuardTransient.sol';
-import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
 import {SafeERC20, IERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
 import {Address} from 'src/dependencies/openzeppelin/Address.sol';
 import {MathUtils} from 'src/libraries/math/MathUtils.sol';
-import {Rescuable} from 'src/utils/Rescuable.sol';
+import {GatewayBase} from 'src/position-manager/GatewayBase.sol';
 import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
 import {INativeWrapper} from 'src/position-manager/interfaces/INativeWrapper.sol';
 import {INativeTokenGateway} from 'src/position-manager/interfaces/INativeTokenGateway.sol';
```
miguelmtzinf:
```suggestion
import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
import {INativeWrapper} from 'src/position-manager/interfaces/INativeWrapper.sol';
import {GatewayBase} from 'src/position-manager/GatewayBase.sol';
import {INativeTokenGateway} from 'src/position-manager/interfaces/INativeTokenGateway.sol';
```

same in `SignatureGateway`, base and interface at the end pls

src/position-manager/interfaces/ISignatureGateway.sol
```diff
@@ -4,158 +4,94 @@ pragma solidity ^0.8.0;
...
-    bool approve,
-    uint256 nonce,
-    uint256 deadline,
+    address spoke,
+    EIP712Types.SetUserPositionManager calldata params,
     bytes calldata signature
   ) external;
 
-  /// @notice Allows consuming a permit for the given reserve's underlying asset on connected SPOKE().
+  /// @notice Allows consuming a permit for the given reserve's underlying asset on the specified registered `spoke`.
```
miguelmtzinf: Rest of functions are using the form "Facilitates...". Apart from that, should we change natspec of this one and `setSelfAsUserPositionManagerWithSig` to reflect they are just passing the data to the specific spoke so it's the spoke validating the sig instead of posm

src/libraries/cryptography/EIP712Hash.sol
miguelmtzinf: I prefer to rm as it's not used, and move to posm dir

src/position-manager/GatewayBase.sol
Kogaroshi: made the changes there.
Also we should consider once #898 is merged to also replace SafeERC20, it should reduce the gas costs for the gateways and simplify the code (no more casting `IERC20`)

Kogaroshi: It was decided to keep OZ's SafeERC20 for now, as it's more battletested, and to have 2 distinct libraries for ERC20 transfers and native transfers as to not mix those.

PR #875, opened 2025-10-07
refactor: add immutable keyword to description

AGMASO:
## General Changes

- one liner add immutable keyword missing in AaveOracle.sol

### DIFF
--- a/src/spoke/AaveOracle.sol
+++ b/src/spoke/AaveOracle.sol
@@ -17,7 +17,7 @@ contract AaveOracle is IAaveOracle {
   uint8 public immutable DECIMALS;
 
   /// @inheritdoc IAaveOracle
-  string public DESCRIPTION;
+  string public immutable DESCRIPTION;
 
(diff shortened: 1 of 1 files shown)


src/spoke/AaveOracle.sol
```diff
@@ -17,7 +17,7 @@ contract AaveOracle is IAaveOracle {
   uint8 public immutable DECIMALS;
 
   /// @inheritdoc IAaveOracle
-  string public DESCRIPTION;
+  string public immutable DESCRIPTION;
```
yan-man, line 20: i think string as it is dynamically sized cannot be an immutable

AGMASO: Makes sense. My bad, I was confused by the highcasing, thought something was missing.

yan-man: good catch though, bc typically we do use the upper case for constants/immutables

PR #877, opened 2025-10-08
fix: Add validation on eliminate deficit

CheyenneAtapour:


src/hub/Hub.sol
```diff
@@ -330,7 +330,9 @@ contract Hub is IHub, AccessManaged {
 
     uint128 shares = previewRemoveByAssets(assetId, amount).toUint128();
     asset.addedShares -= shares;
-    callerSpoke.addedShares -= shares;
+    uint128 callerAddedShares = callerSpoke.addedShares;
```
yan-man: nit: think we can just call this `addedShares`

src/hub/Hub.sol
yan-man, line 1: is there a test we can also add here, if a spoke attempts to eliminate deficit too many shares?

src/hub/Hub.sol
CheyenneAtapour: I refactored slightly, and now have spoke and asset `addedShares` (cheaper due to caching)

src/hub/Hub.sol
CheyenneAtapour: Added

miguelmtzinf:
Closing in favor of https://github.com/aave/aave-v4/pull/869. We prefer to rely on math underflow checks instead of explicit validations for greater efficiency. UX is not impacted as this functions is not user-facing.

PR #878, opened 2025-10-08
feat: skip notify when risk premium change is 0 - 0

DhairyaSethi:
Skips broadcasting a change in the user's risk premium when their risk premium is both zero before and after the action
- gas savings because refresh premium is a costly action. These redundant operations can be skipped for efficiency, as all premium data values remain 0 before/after the action if risk premium is 0.

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -786,10 +786,10 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   function _notifyRiskPremiumUpdate(address user, uint256 newRiskPremium) internal {
     PositionStatus storage positionStatus = _positionStatus[user];
 
-    if (newRiskPremium == 0 && !positionStatus.hasPositivePremium) {
+    if (newRiskPremium == 0 && !positionStatus.hasPositiveRiskPremium) {
       return;
     }
-    positionStatus.hasPositivePremium = newRiskPremium > 0;
+    positionStatus.hasPositiveRiskPremium = newRiskPremium > 0;
 
     uint256 reserveId = _reserveCount;
     while ((reserveId = positionStatus.nextBorrowing(reserveId)) != PositionStatusMap.NOT_FOUND) {
--- a/src/spoke/interfaces/ISpoke.sol
+++ b/src/spoke/interfaces/ISpoke.sol
@@ -62,7 +62,7 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
 
   struct PositionStatus {
     mapping(uint256 slot => uint256) map;
-    bool hasPositivePremium; // premiumShares > 0
+    bool hasPositiveRiskPremium; // premiumShares > 0
   }
 
   struct UserAccountData {


tests/unit/Spoke/Spoke.AccrueInterest.Scenario.t.sol
```diff
@@ -52,10 +52,11 @@ contract SpokeAccrueInterestScenarioTest is SpokeBase {
   }
 
   /// Second accrual after an action - which should update the user rp
-  function test_accrueInterest_fuzz_RPBorrowAndskipTime_twoActions(
+  function test_accrueInterest_fuzz_RPBorrowAndSkipTime_twoActions(
     TestAmounts memory amounts,
     uint32 skipTime
   ) public {
+    vm.skip(true, 'pending rft');
```
CheyenneAtapour, line 59: why are we skipping?

CheyenneAtapour: Oh, there must have been risk premium changes that relied on notify? Maybe worth double checking then that we didn't break any expected behaviors with this opt?

src/spoke/interfaces/ISpoke.sol
```diff
@@ -56,12 +56,13 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
...
 
   struct PositionManagerConfig {
-    bool active;
     mapping(address user => bool) approval;
+    bool active;
   }
 
   struct PositionStatus {
     mapping(uint256 slot => uint256) map;
+    bool hasRiskPremium;
```
miguelmtzinf: you prefer this over `nonZeroRiskPremium`?

tests/unit/Spoke/Spoke.AccrueInterest.Scenario.t.sol
DhairyaSethi: it's very difficult to fix this test, it needs to be rewritten

src/spoke/interfaces/ISpoke.sol
DhairyaSethi: either works yes

src/spoke/Spoke.sol
```diff
@@ -786,6 +786,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   function _notifyRiskPremiumUpdate(address user, uint256 newRiskPremium) internal {
```
miguelmtzinf, line 786:
```suggestion
  /// @dev Skips the refresh if the user risk premium remains zero."
  function _notifyRiskPremiumUpdate(address user, uint256 newRiskPremium) internal {
```

src/spoke/interfaces/ISpoke.sol
avniculae: I personally prefer `nonZeroRiskPremium` over `hasRiskPremium`

DhairyaSethi: preference to avoid negated var names for simplicity

avniculae: I think `hasRiskPremium` can be a bit confusing: all users have a risk premium, it's just that it can be 0

miguelmtzinf: alternatives: `isRiskPremiumCharged`, `isRiskPremiumOn`, `isRiskPremiumActive`, `isRiskPremiumEnabled`, `isRiskPremiumPowered`, `isRiskPremiumApplied`. 
I like the first two

DhairyaSethi: i think if riskPremium is zero you don't 'have' risk premium, you're not being charged anything extra

Kogaroshi: `hasActiveRiskPremium` ? Pref to use `has` instead of `is` in this case since it's user related

avniculae: > i think if riskPremium is zero you don't 'have' risk premium, you're not being charged anything extra

from the perspective of the user I agree, but I think from the perspective of a developer/auditor everybody has risk premium (or at least this is the way I'm seeing it)

PR #883, opened 2025-10-08
opt: refreshPremium

DhairyaSethi:
- remove asset premium 2 wei check since it's a subset of spoke check
- reduce ~12 unnecessary warm sloads by caching


src/hub/Hub.sol
```diff
@@ -686,20 +686,26 @@ contract Hub is IHub, AccessManaged {
...
+    realizedPremium = realizedPremium.add(premium.realizedDelta);
+    spoke.premiumShares = premiumShares.toUint128();
+    spoke.premiumOffset = premiumOffset.toUint128();
+    spoke.realizedPremium = realizedPremium.toUint128();
 
     // can increase due to precision loss on premium (drawn unchanged)
-    require(asset.premium() + premiumAmount - assetPremiumBefore <= 2, InvalidPremiumChange());
-    uint256 spokePremiumAfter = _getSpokePremium(spoke, assetId);
+    uint256 spokePremiumAfter = previewRestoreByShares(assetId, premiumShares) - premiumOffset;
+    spokePremiumAfter += realizedPremium;
```
yan-man, line 707: can we just add this line to L707 directly, ie `previewRestoreByShares(assetId, premiumShares) - premiumOffset + realizedPremium`

DhairyaSethi: no we cannot, we want to utilise the underflow check; it's very intentional 😉

yan-man: would `(previewRestoreByShares(assetId, premiumShares) - premiumOffset) + realizedPremium` still use the underflow check

DhairyaSethi: it does not. that is exactly why on every premium calculation (AssetLogic.premium(), Hub._getSpokeDebt, Spoke._getUserDebt) we first isolate the subtraction in an accruedPremium var, to get underflow check, and then add it separately to realisedPremium

CheyenneAtapour:
Should this target dev branch?

src/hub/Hub.sol
DhairyaSethi: @yan-man actually i was slightly off here, @avniculae correctly pointed out as solidity evaluated expressions from left to right that order: `shares - offset + realized` will revert; but the opposite order `realised + shares - offset` may not; so it's safer to not rely on this and store independently to force the check

src/hub/Hub.sol
```diff
@@ -685,20 +685,26 @@ contract Hub is IHub, AccessManaged {
     PremiumDelta calldata premium,
     uint256 premiumAmount
   ) internal {
-    uint256 assetPremiumBefore = asset.premium();
-    uint256 spokePremiumBefore = _getSpokePremium(spoke, assetId);
+    uint256 premiumShares = spoke.premiumShares;
+    uint256 premiumOffset = spoke.premiumOffset;
+    uint256 realizedPremium = spoke.realizedPremium;
+    uint256 spokePremiumBefore = previewRestoreByShares(assetId, premiumShares) - premiumOffset;
```
avniculae, line 691: I think this optimisation makes the code a bit ugly. shall we create a small helper function
```
getPremium(assetId, premiumShares, premiumOffset, realizedPremium)
   accruedPremium = previewRestoreByShares(assetId, premiumShares) - premiumOffset;
   return realizedPremium + accruedPremium 
```

and use this function here, in `_getSpokePremium` and in AssetLogic's `premium` function?

DhairyaSethi: yeah agree, thought about this but this caching is only really useful here and nowhere else; filling out args for one time use will look uglier elsewhere

DhairyaSethi: i was to also not use `previewRestoreByShares` the full helpers inside the code also bc it internally does an avoidable lookup

avniculae: > filling out args for one time use will look uglier elsewhere

maybe a bit uglier, but I think it also helps with code duplication, especially around this checked subtraction point we discusses about

miguelmtzinf: agree, the opt made the code uglier... but I'm fine with it as it brings significant gas savings. Also, at this point, I think better to use `toDrawnAssetsUp` directly, because i) it's more efficient and ii) even more evident that we round up in opposite direction than `premiumOffset` (one could wonder why we are using `previewRestore` instead of `previewAdd`)
```suggestion
    // rounds up, in the opposite direction to premium offset calculation
    uint256 spokePremiumBefore = asset.toDrawnAssetsUp(premiumShares) - premiumOffset;
```

DhairyaSethi: i don't think we need to mention premium offset calculation comment here, it's on the spoke and this is hub 

def good with directly using toDrawnAssetsUp -> would be my preference as well -> didn't do it bc of consistency with other internal methods in the hub, are we good with doing the optimal explicit rounding everywhere?

DhairyaSethi:
premiumOffset integrity upheld https://prover.certora.com/output/6838972/ea06a19d350f4418bb255225a72f4dcb

src/hub/Hub.sol
miguelmtzinf: re comment: good pt. does it make sense to say this is equivalent to `getSpokePremium` or smt?
re opt: again, at this point we can just optimize as much as we can (we can say it somewhere)

miguelmtzinf:
This change was reverted because the asset premium check is actually needed.
FV property [premiumOffsetIntegirty](https://prover.certora.com/output/40726/fe07377d7075416f8bfd238aae52129e/?anonymousKey=58b4358cabe15f9d2ecdbb412ad78e0ea8ed0cf2) was not passing.

PR #884, opened 2025-10-09
feat: liquidate to receive shares

yan-man:
- allow liquidation to receive `addedShares` rather than underlying collateral asset
- not possible to receive shares of a collateral reserve that is frozen
- allow for liquidations to occur even in low liquidity environment by providing collateral to liquidators in the form of supplied shares


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -456,11 +460,13 @@ library LiquidationLogic {
...
 
-    uint256 sharesToLiquidator = hub.remove(
-      assetId,
-      params.collateralToLiquidator,
-      params.liquidator
-    );
+    uint256 sharesToLiquidator;
+    if (params.receiveShares) {
+      sharesToLiquidator = hub.previewRemoveByAssets(assetId, params.collateralToLiquidator);
+      position.suppliedShares += sharesToLiquidator.toUint128();
```
avniculae: this should be liquidator's UserPosition, not user's

yan-man: yeah i know, WIP bc thinking about best way to pass liq position in atm

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -125,16 +126,16 @@ library LiquidationLogic {
   function liquidateUser(
     ISpoke.Reserve storage collateralReserve,
     ISpoke.Reserve storage debtReserve,
-    ISpoke.UserPosition storage collateralPosition,
-    ISpoke.UserPosition storage debtPosition,
+    mapping(address user => mapping(uint256 reserveId => ISpoke.UserPosition))
+      storage userPositions,
```
yan-man: passed in whole mapping to combat stack overflow issues

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -456,11 +461,14 @@ library LiquidationLogic {
...
 
-    uint256 sharesToLiquidator = hub.remove(
-      assetId,
-      params.collateralToLiquidator,
-      params.liquidator
-    );
+    uint256 sharesToLiquidator;
+    if (params.receiveShares) {
+      sharesToLiquidator = hub.previewRemoveByAssets(assetId, params.collateralToLiquidator);
+      liquidatorPosition.suppliedShares += sharesToLiquidator.toUint128();
```
yan-man: decide: 
- whether we need to notify (ie from consistency PR https://github.com/aave/aave-v4/pull/888
- it is allowed if frozen? prob ok bc it's shifting internal accounting rather than adding new shares

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -182,17 +183,19 @@ library LiquidationLogic {
 
     bool isCollateralPositionEmpty = _liquidateCollateral(
       collateralReserve,
-      collateralPosition,
+      userPositions[params.user][params.collateralReserveId],
+      userPositions[params.liquidator][params.collateralReserveId],
```
DhairyaSethi: let's also pass userPositions here? we don't need to lookup the slot for liquidator user pos if receive shares is false

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -65,13 +65,15 @@ interface ISpokeBase {
   /// @param liquidatedDebt The debt amount of borrowed asset to be liquidated.
   /// @param liquidatedCollateral The amount of collateral received by the liquidator.
   /// @param liquidator The address of the liquidator.
+  /// @param receiveShares Whether the liquidator receives collateral in assets or supplied shares.
```
miguelmtzinf:
```suggestion
  /// @param receiveShares Whether the liquidator receives collateral in supplied shares, otherwise in underlying.
```

tests/gas/Spoke.Operations.gas.t.sol
```diff
@@ -164,11 +164,37 @@ contract SpokeOperations_Gas_Tests is SpokeBase {
     skip(100);
 
     vm.startPrank(bob);
-    spoke.liquidationCall(reserveId.usdx, reserveId.dai, alice, 100_000e18);
-    vm.snapshotGasLastCall(NAMESPACE, 'liquidationCall: partial');
+    spoke.liquidationCall(reserveId.usdx, reserveId.dai, alice, 100_000e18, false);
+    vm.snapshotGasLastCall(NAMESPACE, 'liquidationCall (receive assets): partial');
```
miguelmtzinf:
```suggestion
    vm.snapshotGasLastCall(NAMESPACE, 'liquidationCall: partial');
```
i d just keep them as before

src/spoke/Spoke.sol
```diff
@@ -343,8 +345,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     bool isUserInDeficit = LiquidationLogic.liquidateUser(
       _reserves[collateralReserveId],
       _reserves[debtReserveId],
```
miguelmtzinf, line 355: does passing `_reserves` save anything? i d make sense to do it for consistency also

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: does `positions` name makes the line go above the limit / jump into a new one?

src/spoke/Spoke.sol
DhairyaSethi: would declare storage pointers for both pos & reserves like before but at the top of this fn in the library, such that we don't do repeated lookups for storage slots -> hopefully stack too deep is feeling kind

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: - it's fine we don't notify as it's consistent with `supply`
- liquidations are a good way to offboard an asset of the system... so this could be problamatic. If DAO decides to lower the CF of the collateral, we would probably want to get rid of it completely...  I d rather check that `receiveShares` is set to `false` when `asset.frozen` is `true`.

tests/mocks/LiquidationLogicWrapper.sol
```diff
@@ -2,6 +2,8 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity ^0.8.0;
 
+import {console2 as console} from 'forge-std/console2.sol';
```
miguelmtzinf: rm

src/spoke/interfaces/ISpokeBase.sol
CheyenneAtapour: What about just extending your suggestion to "underlying assets" to be clear?

tests/gas/Spoke.Operations.gas.t.sol
yan-man: as in leaving this as is but still adding the receive shares snapshots tests separately right? 

https://github.com/aave/aave-v4/commit/ca5722e0beff683bc3f4866b6d3172e8bd609e9a

src/spoke/libraries/LiquidationLogic.sol
yan-man: ha good call, keeps it same line

https://github.com/aave/aave-v4/pull/884/commits/ca5722e0beff683bc3f4866b6d3172e8bd609e9a

src/spoke/Spoke.sol
yan-man: > hopefully stack too deep is feeling kind

it was not feeling kind...I passed in `reserves` for consistency anyway. It costs ~2k more, basically i think bc we do the repeated lookups

https://github.com/aave/aave-v4/pull/884/commits/ca5722e0beff683bc3f4866b6d3172e8bd609e9a

tests/mocks/LiquidationLogicWrapper.sol
yan-man: https://github.com/aave/aave-v4/pull/884/commits/ca5722e0beff683bc3f4866b6d3172e8bd609e9a

src/spoke/interfaces/ISpokeBase.sol
yan-man: https://github.com/aave/aave-v4/commit/ca5722e0beff683bc3f4866b6d3172e8bd609e9a

src/spoke/libraries/LiquidationLogic.sol
yan-man: https://github.com/aave/aave-v4/commit/ca5722e0beff683bc3f4866b6d3172e8bd609e9a

src/spoke/libraries/LiquidationLogic.sol
yan-man: https://github.com/aave/aave-v4/pull/884/commits/d86685d1f0794300321407a2c7608a22d8d9a9dd

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -113,43 +120,43 @@ library LiquidationLogic {
...
   ) external returns (bool) {
-    uint256 collateralReserveBalance = collateralReserve.hub.previewRemoveByShares(
-      collateralReserve.assetId,
-      collateralPosition.suppliedShares
-    );
+    uint256 collateralReserveBalance = reserves[params.collateralReserveId]
+      .hub
+      .previewRemoveByShares(
+        reserves[params.collateralReserveId].assetId,
+        positions[params.user][params.collateralReserveId].suppliedShares
```
DhairyaSethi: this and all functions below are doing so much duplicate and unnecessary work on the hashing to get this slot each time!

do we not have stack space to declare collateralReserve & debtReserve storage pointers at the start of this function instead & pass exactly as before 

we just don't load the storage pointer for liquidator, for him we pass the mapping and load if needed (ie receive shares is true)

yan-man: best i could do is cache one of the reserves: https://github.com/aave/aave-v4/pull/884/commits/a27e1496cde1ce77b10f3f5c80542af4de486d37

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -113,93 +120,95 @@ library LiquidationLogic {
...
         user: params.user,
         liquidator: params.liquidator,
         debtToCover: params.debtToCover,
         collateralReserveHub: address(collateralReserve.hub),
-        debtReserveHub: address(debtReserve.hub),
+        debtReserveHub: address(reserves[params.debtReserveId].hub),
         collateralReservePaused: collateralReserve.paused,
-        debtReservePaused: debtReserve.paused,
+        collateralReserveFrozen: collateralReserve.frozen,
+        debtReservePaused: reserves[params.debtReserveId].paused,
```
DhairyaSethi: i think this can just take in the storage pointers for collateral reserve & debt reserve, to avoid needing to do this which is what's causing us to be so close to stack too deep 

i think it was originally written this way for testing? but i believe we can do the above to have more leeway on stack too deep & make the code look a little nicer too

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -113,93 +120,95 @@ library LiquidationLogic {
...
+          debtReserveBalance: params.drawnDebt + params.premiumDebt,
+          collateralReserveBalance: collateralReserveBalance,
+          debtToCover: params.debtToCover,
+          totalDebtValue: params.totalDebtValue,
+          healthFactor: params.healthFactor,
+          maxLiquidationBonus: collateralDynConfig.maxLiquidationBonus,
+          collateralFactor: collateralDynConfig.collateralFactor,
+          liquidationFee: collateralDynConfig.liquidationFee,
+          debtAssetPrice: IAaveOracle(params.oracle).getReservePrice(params.debtReserveId),
+          debtAssetUnit: uint256(10).uncheckedExp(reserves[params.debtReserveId].decimals),
```
DhairyaSethi:
```suggestion
          debtAssetUnit: MathUtils.uncheckedExp(10, debtReserve.decimals),
```

DhairyaSethi: we will have space to cache debt reserve storage pointer if we apply the above suggestion 
in either case, we need to make the unchecked exp syntax change for consistency

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -113,93 +120,95 @@ library LiquidationLogic {
...
+            params.collateralReserveId
+          ),
+          collateralAssetUnit: uint256(10).uncheckedExp(collateralReserve.decimals)
+        })
+      );
 
     bool isCollateralPositionEmpty = _liquidateCollateral(
-      collateralReserve,
-      collateralPosition,
+      reserves,
```
DhairyaSethi: this can take reserves & collateral reserve again such that we don't redeclare it no? reserves is only needed for opening liquidator user pos when flag is true

src/spoke/libraries/LiquidationLogic.sol
yan-man: do you mean like this? https://github.com/aave/aave-v4/pull/903/files

had tried it already and stack issues

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -144,62 +149,67 @@ library LiquidationLogic {
...
+          healthFactor: params.healthFactor,
+          maxLiquidationBonus: collateralDynConfig.maxLiquidationBonus,
+          collateralFactor: collateralDynConfig.collateralFactor,
+          liquidationFee: collateralDynConfig.liquidationFee,
+          debtAssetPrice: IAaveOracle(params.oracle).getReservePrice(params.debtReserveId),
+          debtAssetUnit: uint256(10).uncheckedExp(debtReserve.decimals),
+          collateralAssetPrice: IAaveOracle(params.oracle).getReservePrice(
+            params.collateralReserveId
+          ),
+          collateralAssetUnit: uint256(10).uncheckedExp(collateralReserve.decimals)
```
miguelmtzinf: to avoid stack too deep we can pass decimals instead of units. Then, use local vars for units in `_calculateLiquidationAmounts`.
Apart from that, this struct is huge. We can consider passing the `liquidationConfig` and `collateralDynConfig` storage pointers.

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -418,24 +430,24 @@ library LiquidationLogic {
...
-        offsetDelta: -position.premiumOffset.toInt256(),
+        sharesDelta: -debtPosition.premiumShares.toInt256(),
+        offsetDelta: -debtPosition.premiumOffset.toInt256(),
         realizedDelta: params.accruedPremium.toInt256() - premiumDebtToLiquidate.toInt256()
       });
 
-      uint256 drawnSharesLiquidated = reserve.hub.restore(
-        reserve.assetId,
+      uint256 drawnSharesLiquidated = debtReserve.hub.restore(
+        debtReserve.assetId,
```
```suggestion
        params.debtReserveId,
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -445,28 +457,32 @@ library LiquidationLogic {
...
-    ISpoke.Reserve storage reserve,
-    ISpoke.UserPosition storage position,
+    ISpoke.Reserve storage collateralReserve,
+    mapping(address user => mapping(uint256 reserveId => ISpoke.UserPosition)) storage positions,
     LiquidateCollateralParams memory params
   ) internal returns (bool) {
-    IHubBase hub = reserve.hub;
-    uint256 assetId = reserve.assetId;
+    IHubBase hub = collateralReserve.hub;
+    uint256 assetId = collateralReserve.assetId;
```
miguelmtzinf: we can use `params.collateralReserveId` instead

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -445,28 +457,32 @@ library LiquidationLogic {
...
   function _liquidateCollateral(
-    ISpoke.Reserve storage reserve,
-    ISpoke.UserPosition storage position,
+    ISpoke.Reserve storage collateralReserve,
+    mapping(address user => mapping(uint256 reserveId => ISpoke.UserPosition)) storage positions,
     LiquidateCollateralParams memory params
   ) internal returns (bool) {
-    IHubBase hub = reserve.hub;
-    uint256 assetId = reserve.assetId;
+    IHubBase hub = collateralReserve.hub;
```
miguelmtzinf: we could just pass the hub address here

src/spoke/libraries/LiquidationLogic.sol
DhairyaSethi: this is hub call, need to pass asset id only, no?

yan-man: yes correct need assetId for hub call

src/spoke/libraries/LiquidationLogic.sol
yan-man: hm we need assetId instead of reserveId right?

src/spoke/libraries/LiquidationLogic.sol
yan-man: to clarify, you mean include it in LiquidateCollateralParams right? (and we should pass it as IHub instead of address I think)

src/spoke/libraries/LiquidationLogic.sol
yan-man: > to avoid stack too deep we can pass decimals instead of units

this works, will do this so we can do the MathUtils.uncheckedExp update: https://github.com/aave/aave-v4/pull/884/commits/6f888b6092b96c8394e45517defa58cddf932ff7

> We can consider passing the liquidationConfig and collateralDynConfig storage pointers

pairing with this change causes stack issues: https://github.com/aave/aave-v4/compare/feat/liquidate-receive-shares...opt/poc-liqLogic-pointers6?expand=1

src/spoke/libraries/LiquidationLogic.sol
yan-man: trying to pass hub as a param caused stack issues

src/hub/interfaces/IHubBase.sol
```diff
@@ -151,7 +151,7 @@ interface IHubBase {
   /// @param premiumDelta The change in premium.
   function refreshPremium(uint256 assetId, PremiumDelta calldata premiumDelta) external;
 
-  /// @notice Transfers existing `addedShares` of caller spoke to `feeReceiver`.
+  /// @notice Transfers `shares` amount of existing `addedShares` of caller spoke to `feeReceiver`.
```
yan-man, line 154: also updated this comment for clarity according to stermi suggestion

src/hub/Hub.sol
```diff
@@ -878,4 +855,27 @@ contract Hub is IHub, AccessManaged {
     require(caller == asset.reinvestmentController, OnlyReinvestmentController());
     require(amount > 0 && amount <= asset.swept, InvalidAmount());
   }
+
+  /// @dev Validates applied premium delta for given premium data and returns updated premium data.
+  function _validateApplyPremiumDelta(
```
DhairyaSethi, line 860: this was much better where it was, right below _applyPremiumDelta
why was this moved down?

miguelmtzinf: it's pure and based on solidity style guideline must be at the end

src/spoke/Spoke.sol
```diff
@@ -294,7 +294,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       msg.sender
     );
 
-    _settlePremiumDebt(userPosition, premiumDelta.realizedDelta);
+    LiquidationLogic.settlePremiumDebt(userPosition, premiumDelta.realizedDelta);
```
DhairyaSethi: not a fan, would much rather duplicate
a soln where this is hidden is, add an innocent
```solidity
using LiquidationLogic for *;
```
 at the top and make this
```solidity
userPosition.settlePremiumDebt(premiumDelta.realizedDelta);
```
the same can be done in liquidation logic usage so nobody will ever know 😉

src/spoke/Spoke.sol
```diff
@@ -988,4 +978,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       fnOut := fnIn
     }
   }
+
+  function _validateReserveConfig(ReserveConfig calldata config) internal pure {
```
DhairyaSethi: you're moving around based on `pure` & `view` but now is very very awkwardly placed at the end, it was fine before logically clubbed with other `_validate*` functions, and right after `_validateDynamicConfig` pls revert this change

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -65,13 +65,15 @@ interface ISpokeBase {
   /// @param liquidatedDebt The debt amount of borrowed asset to be liquidated.
   /// @param liquidatedCollateral The amount of collateral received by the liquidator.
   /// @param liquidator The address of the liquidator.
+  /// @param receiveShares Whether the liquidator receives collateral in supplied shares, otherwise in underlying assets.
```
DhairyaSethi:
```suggestion
  /// @param receiveShares True to receive collateral in `suppliedShares`, false to receiver in underlying.
```

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -118,11 +120,13 @@ interface ISpokeBase {
   /// @param debtReserveId The reserveId of the underlying asset borrowed by the liquidated user, to be repaid by Liquidator.
   /// @param user The address of the user to liquidate.
   /// @param debtToCover The desired amount of debt to cover.
+  /// @param receiveShares Whether the liquidator receives collateral in supplied shares, otherwise in underlying assets.
```
DhairyaSethi: the tense of this natspec and on comment need to be consistent w the rest

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -221,6 +232,114 @@ library LiquidationLogic {
...
+    uint256 sharesToLiquidator;
+    if (params.receiveShares) {
+      sharesToLiquidator = hub.previewRemoveByAssets(assetId, params.collateralToLiquidator);
+      positions[params.liquidator][params.collateralReserveId].suppliedShares += sharesToLiquidator
+        .toUint128();
+    } else {
+      sharesToLiquidator = hub.remove(assetId, params.collateralToLiquidator, params.liquidator);
+    }
+
+    if (sharesToLiquidate > sharesToLiquidator) {
```
miguelmtzinf, line 302: per @DhairyaSethi, if `receiveShares = true` and `sharesToLiquidate <= sharesToLiquidator` (no fees), collateral reserve does not accrue.

src/hub/Hub.sol
DhairyaSethi: it's very awkwardly placed now, was much better before; let's make an exception instead of being sticklers pls

src/spoke/Spoke.sol
miguelmtzinf: moved this up, right before `_validateReserveDynamicConfig`

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -77,64 +81,66 @@ library LiquidationLogic {
...
   /// @param collateralDynConfig The collateral dynamic config.
   /// @param params The liquidate user params.
-  /// @return True if the liquidation results in deficit, false otherwise.
+  /// @return True if the liquidation results in deficit.
   function liquidateUser(
     ISpoke.Reserve storage collateralReserve,
     ISpoke.Reserve storage debtReserve,
-    ISpoke.UserPosition storage collateralPosition,
-    ISpoke.UserPosition storage debtPosition,
+    mapping(address user => mapping(uint256 reserveId => ISpoke.UserPosition)) storage positions,
```
avniculae, line 135: personally I'd prefer to pass the 3 positions that we are interested in (coll, debt and liq), but nbd

src/spoke/libraries/LiquidationLogic.sol
avniculae: a very good point, are there any concerns? (apart from the current issue with the supply share price)

src/spoke/libraries/LiquidationLogic.sol
yan-man: I think I tried that initially and ran into some stack issues. Lots of stack issues I encountered trying to tweak things

PR #885, opened 2025-10-09
feat : return shares amount

Kogaroshi:
Return the shares amounts in user actions on the Spoke (supply, withdraw, borrow, repay)


src/spoke/Spoke.sol
```diff
@@ -256,23 +260,24 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
     Reserve storage reserve = _getReserve(reserveId);
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
     _validateRepay(reserve);
     IHubBase hub = reserve.hub;
-    uint256 assetId = reserve.assetId;
 
     (uint256 drawnDebtRestored, uint256 premiumDebtRestored, uint256 accruedPremium) = _getUserDebt(
       hub,
-      assetId,
+      reserve.assetId,
```
avniculae, line 280: is this bc of stack too deep? have we validated that optimiser will optimise this?

Kogaroshi: Yeah, needed to prevent StackTooDeep

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -21,15 +21,21 @@ interface ITreasurySpoke is ISpokeBase {
   /// @param reserveId The identifier of the reserve.
   /// @param amount The amount of asset to supply.
   /// @param onBehalfOf Unused parameter for this spoke.
-  function supply(uint256 reserveId, uint256 amount, address onBehalfOf) external;
+  /// @return The amount of shares minted.
```
avniculae: any reason to use minted here instead of supplied?

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -21,15 +21,21 @@ interface ITreasurySpoke is ISpokeBase {
...
+  function supply(uint256 reserveId, uint256 amount, address onBehalfOf) external returns (uint256);
 
   /// @notice Withdraws a specified amount of underlying asset from the given reserve.
   /// @dev Providing an amount greater than the maximum withdrawable value signals a full withdrawal.
   /// @dev The reserve identifier must match the asset identifier in the hub.
   /// @param reserveId The identifier of the reserve.
   /// @param amount The amount of asset to withdraw.
   /// @param onBehalfOf Unused parameter for this spoke.
-  function withdraw(uint256 reserveId, uint256 amount, address onBehalfOf) external;
+  /// @return The amount of shares burned.
```
avniculae: same q as above

src/spoke/Spoke.sol
avniculae: looking at the gas snapshots it looks like it optimises this, all good

src/spoke/Spoke.sol
```diff
@@ -256,23 +260,24 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     _notifyRiskPremiumUpdate(onBehalfOf, newRiskPremium);
 
     emit Borrow(reserveId, msg.sender, onBehalfOf, drawnShares);
+
+    return drawnShares;
```
avniculae, line 264: @DhairyaSethi @miguelmtzinf can you pls remind me, why are we calling them drawnShares and not borrowedShares?

avniculae: same q for repay: we call them restoredShares

yan-man: just bc they're drawn from the hub, so hub pov it's drawnShares, but borrowedShares would work too bc it's in spoke then, and does prob make more sense now that we return it 

and to be consistent with supply/withdraw

src/spoke/interfaces/ITreasurySpoke.sol
yan-man: yeah same q, as it's a bit inconsistent with spokebase

src/spoke/Spoke.sol
miguelmtzinf: we have `drawnShares` also as part of the `UserPosition` struct... We can rename this one, but would look weird anyway:
`userPosition.drawnShares += borrowedShares.toUint128();`

DhairyaSethi:
think we can avoid for `liquidationCall` right now

PR #888, opened 2025-10-09
Poc/notify consistency

yan-man:
- add notify to `supply` (for consistency w other actions) to check gas changes


src/spoke/Spoke.sol
```diff
@@ -226,10 +229,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
     userPosition.suppliedShares -= withdrawnShares.toUint128();
 
-    if (_positionStatus[onBehalfOf].isUsingAsCollateral(reserveId)) {
-      uint256 newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
-      _notifyRiskPremiumUpdate(onBehalfOf, newRiskPremium);
-    }
+    uint256 newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
```
DhairyaSethi: not validate always, the condition still holds here; we need to do: if enabledAsCollateral: rp = _refreshAndValidate else rp = _calcUserAccountData().rp (note no refresh or validate in else branch)

src/spoke/Spoke.sol
```diff
@@ -375,11 +375,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
     if (usingAsCollateral) {
       _refreshDynamicConfig(onBehalfOf, reserveId);
```
DhairyaSethi: this is not needed now bc you're refreshing below

src/spoke/Spoke.sol
```diff
@@ -375,11 +375,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    } else {
-      uint256 newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
-      _notifyRiskPremiumUpdate(onBehalfOf, newRiskPremium);
     }
 
+    uint256 newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
```
DhairyaSethi: we should avoid validate part in the if branch (ie usingAsCollateral=true)

src/spoke/Spoke.sol
```diff
@@ -226,10 +229,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
+    uint256 newRiskPremium;
     if (_positionStatus[onBehalfOf].isUsingAsCollateral(reserveId)) {
-      uint256 newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
-      _notifyRiskPremiumUpdate(onBehalfOf, newRiskPremium);
+      newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
+    } else {
+      newRiskPremium = _calculateUserAccountData(onBehalfOf).riskPremium;
     }
+    _notifyRiskPremiumUpdate(onBehalfOf, newRiskPremium);
```
DhairyaSethi, line 237: on supply we don't want to refresh dynamic config (& hence shouldn't need to check hf) 
here in either case it should just be calc & notify 
like repay

src/spoke/Spoke.sol
```diff
@@ -373,12 +378,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
+    uint256 newRiskPremium;
     if (usingAsCollateral) {
-      _refreshDynamicConfig(onBehalfOf, reserveId);
+      newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
     } else {
-      uint256 newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
-      _notifyRiskPremiumUpdate(onBehalfOf, newRiskPremium);
+      newRiskPremium = _calculateUserAccountData(onBehalfOf).riskPremium;
     }
```
DhairyaSethi, line 385: we don't want to refresh & validate on the if branch here, but rather on else

yan-man:
supply is anyway going to be refreshed when set as collateral due to this change: https://github.com/aave/aave-v4/pull/939; otherwise want to keep supply action as lean as possible gas wise

PR #890, opened 2025-10-09
opt: Caching on add and restore

CheyenneAtapour:


src/hub/Hub.sol
yan-man, line 1: other minor opts we can add here, for spoke (where stack doesnt overflow):
- similar caching on suppliedShares accounting before updating
- passing _calculateUserAccountData(onBehalfOf).riskPremium directly to _notifyRiskPremiumUpdate instead of caching first
- passing _refreshAndValidateUserPosition directly to _notifyRiskPremiumUpdate (but this one is i think a bit harder to read after impl)

not sure if it should go here

miguelmtzinf:
Closing, because the gas cost benefits are negligible (~200 across operations) and this change reduces readability.

PR #893, opened 2025-10-10
feat : compiler profiles to use via-IR

Kogaroshi:
New compiler profiles to use via-IR for Spoke & Hub, pushing as much as possible the compiler runs to stay under the size limit while reducing the execution gas costs.


Kogaroshi:
Compilation time comparaison : 
`main` => `Solc 0.8.28 finished in 67.75s`
`viaIR` => `Solc 0.8.28 finished in 69.95s` (Spoke & Hub are compiled viaIR in parallel of the main compilation thread)

tests/Deploy.sol
```diff
@@ -0,0 +1,72 @@
...
+    return computeAddress(salt, initCodeHash, address(this));
+  }
+
+  function computeAddress(
+    bytes32 salt,
+    bytes32 bytecodeHash,
+    address deployer
+  ) internal pure returns (address addr) {
+    /// @solidity memory-safe-assembly
+    assembly {
```
DhairyaSethi:
```suggestion
    assembly ('memory-safe') {
```
we're not using older compiler versions, dont need to specify using comment, and this is being deprecated in solc 0.8.31

tests/DeployUtils.sol
```diff
@@ -0,0 +1,72 @@
...
+    return deployHub(authority, '');
+  }
+
+  function deployHub(address authority, bytes32 salt) internal returns (IHub hub) {
+    bytes memory initCode = abi.encodePacked(
+      vm.getCode('src/hub/Hub.sol:Hub'),
+      abi.encode(authority)
+    );
+    assembly {
+      hub := create2(0, add(initCode, 0x20), mload(initCode), salt)
```
DhairyaSethi: i think these should deploy using a common factory that we enforce on all chains 
just the deterministic deployer that foundry also internally uses is fine https://etherscan.io/address/0x4e59b44847b379578588920ca78fbf26c0b4956c

we can do vanity shenanigans easily on hub

tests/Deploy.sol
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {Vm} from 'forge-std/Vm.sol';
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+
+library Deploy {
```
DhairyaSethi: `Deployer` instead?

tests/DeployUtils.sol
```diff
@@ -0,0 +1,72 @@
...
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+
+library Deploy {
+  Vm internal constant vm = Vm(address(uint160(uint256(keccak256('hevm cheat code')))));
+
+  function deploySpokeInstance(address oracle) internal returns (ISpoke) {
+    return deploySpokeInstance(oracle, '');
+  }
+
+  function deploySpokeInstance(address oracle, bytes32 salt) internal returns (ISpoke spoke) {
```
DhairyaSethi: we can add proxify on this lib as well, wdyt?

foundry.toml
```diff
@@ -7,11 +7,23 @@ fs_permissions = [{ access = "read", path = "tests/mocks/JsonBindings.sol" }]
...
-optimizer_runs = 200
+optimizer_runs = 444444444444
 bytecode_hash = "none"
 gas_snapshot_check = false
 gas_limit = 1099511627776
 
+additional_compiler_profiles = [
+  { name = "hub", optimizer = true, via_ir = true, optimizer_runs = 6404 },
+  { name = "spoke", optimizer = true, via_ir = true, optimizer_runs = 830 },
+  { name = "tests", optimizer = true, via_ir = false, optimizer_runs = 444444444444 },
```
DhairyaSethi, line 18: why can't tests be with standard-fast-enough 200, again? if this is not meaningfully slower to compile tests its fine

Kogaroshi: The tests need to have the same profile as the general `src/` profile, to prevent import issues.
Those [values](https://github.com/aave/aave-v4/pull/893#issuecomment-3547278802) are with the tests under such high runs number, doesn't seem to impact tests compile time.

tests/Deploy.sol
Kogaroshi: idk, not really a deployer contract since its a lib, maybe `DeployUtils` ?

foundry.toml
DhairyaSethi: ah ok we're maximising opt from the general src also by compiling at higher runs? does it really make a difference? i think it's only gateways & oracle right

Kogaroshi: Oracle, Posm/Gateways, Configurators yeah, marginal changes, just to squeeze as much gas opt as we can.
Can be good for some Posm in the future.

tests/DeployUtils.sol
Kogaroshi: faa6e8b992037542497e30471cad3ca02907a402

tests/Deploy.sol
Kogaroshi: faa6e8b992037542497e30471cad3ca02907a402

tests/Deploy.sol
Kogaroshi: faa6e8b992037542497e30471cad3ca02907a402

tests/DeployUtils.sol
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  function _proxify(
+    address deployer,
+    address impl,
+    address proxyAdminOwner,
+    bytes memory initData
+  ) internal returns (address) {
+    vm.prank(deployer);
```
DhairyaSethi: i meant proxify here to be used in production and also create2-ed for vanity address xD 

but we can revert this change lol

Kogaroshi: Ah, mb misunderstanding.
But regarding this & https://github.com/aave/aave-v4/pull/893#discussion_r2667452025, here those are just test deployment utils, not the actual deployment scripts, this should be part of the Deploy Payload Engine update we'd need to apply once we merge this one (if we do).

src/spoke/interfaces/ISpoke.sol
```diff
@@ -439,6 +439,8 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     bytes32 permitS
   ) external;
 
+  function initialize(address authority) external;
```
yan-man: why needed here?

tests/DeployUtils.sol
```diff
@@ -0,0 +1,97 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {Vm} from 'forge-std/Vm.sol';
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+import {TransparentUpgradeableProxy, ITransparentUpgradeableProxy} from 'src/dependencies/openzeppelin/TransparentUpgradeableProxy.sol';
```
yan-man: maybe can mv these external deps up for consistency

tests/unit/Hub/Hub.Config.t.sol
```diff
@@ -806,7 +806,8 @@ contract HubConfigTest is HubBase {
   function _assumeValidAssetConfig(IHub.AssetConfig memory newConfig) internal pure {
     newConfig.liquidityFee = bound(newConfig.liquidityFee, 0, PercentageMath.PERCENTAGE_FACTOR)
       .toUint16();
-    vm.assume(address(newConfig.feeReceiver) != address(0) || newConfig.liquidityFee == 0);
+    //vm.assume(address(newConfig.feeReceiver) != address(0) || newConfig.liquidityFee == 0);
```
yan-man: rm

src/spoke/interfaces/ISpoke.sol
Kogaroshi: Because we do not import `Spoke.sol` in the `Base.t.sol` anymore (due to compiler profile different settings), so solution was to go through the interface.

tests/DeployUtils.sol
```diff
@@ -0,0 +1,97 @@
...
+  function getDeterministicSpokeInstanceAddress(
+    address oracle,
+    bytes32 salt
+  ) internal view returns (address) {
+    bytes memory initCode = abi.encodePacked(
+      vm.getCode('src/spoke/instances/SpokeInstance.sol:SpokeInstance'),
+      abi.encode(oracle)
+    );
+    bytes32 initCodeHash = keccak256(initCode);
+    return computeAddress(salt, initCodeHash, address(this));
```
DhairyaSethi: lets have the deployer configurable

src/spoke/interfaces/ISpoke.sol
```diff
@@ -439,6 +439,11 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     bytes32 permitS
   ) external;
 
+  /// @notice Initializes the spoke contract.
+  /// @dev To be overridden by the inheriting Spoke instance contract.
```
yan-man: should we rm this dev comment as it's in Spoke.sol

tests/DeployUtils.sol
Kogaroshi: 0e1cb20c36645d9361b7c1d5df9449f869313371

tests/DeployUtils.sol
Kogaroshi: 0e1cb20c36645d9361b7c1d5df9449f869313371

src/spoke/interfaces/ISpoke.sol
```diff
@@ -439,6 +439,11 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
...
+  /// @param authority The address of the AccessManager contract.
+  function initialize(address authority) external;
```
miguelmtzinf: we left it out in purpose, because this can easily change in future versions

tests/DeployUtils.sol
```diff
@@ -0,0 +1,149 @@
...
+library DeployUtils {
+  // https://github.com/safe-global/safe-singleton-factory
+  address public constant CREATE2_FACTORY = 0x914d7Fec6aaC8cd542e72Bca78B30650d45643d7;
+  bytes internal constant CREATE2_FACTORY_BYTECODE =
+    hex'7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3';
+
+  Vm internal constant vm = Vm(address(uint160(uint256(keccak256('hevm cheat code')))));
+
+  error NoCreate2Factory();
+  error Create2DeploymentFailed();
```
miguelmtzinf: move up

tests/DeployUtils.sol
```diff
@@ -0,0 +1,149 @@
...
+      (, returnData) = CREATE2_FACTORY.call(creationBytecode);
+
+      address deployedAt = address(uint160(bytes20(returnData)));
+      require(deployedAt == computed, Create2DeploymentFailed());
+
+      return deployedAt;
+    }
+  }
+
+  function isContractDeployed(address _addr) internal view returns (bool isContract) {
```
miguelmtzinf:
```suggestion
  function _isContractDeployed(address instance) internal view returns (bool) {
```

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -9,29 +9,30 @@ contract SpokeConfigTest is SpokeBase {
...
     assertNotEq(instance.getLiquidationLogic(), address(0));
   }
 
   function test_spoke_deploy_reverts_on_InvalidConstructorInput() public {
+    DeployWrapper deployer = new DeployWrapper();
+
     vm.expectRevert();
-    new SpokeInstance(address(0));
+    deployer.deploySpokeInstance(address(0));
   }
```
DhairyaSethi, line 25: for this we should use the lib and not the ext wrapper imo

ext wrapper only used for its specific tests

expect_internal_revert = true is would be needed for forge to catch internal errors

src/spoke/interfaces/ISpoke.sol
Kogaroshi: will remove

src/spoke/interfaces/ISpoke.sol
Kogaroshi: okay will remove

tests/unit/Spoke/Spoke.Config.t.sol
Kogaroshi: Tried with `expect_internal_revert = true`, wasn't catching the revert, I think due to compute then deploy in the internal method.

tests/DeployUtils.sol
```diff
@@ -0,0 +1,149 @@
...
+  ) internal view returns (address) {
+    bytes memory initCode = abi.encodePacked(
+      vm.getCode('src/hub/Hub.sol:Hub'),
+      abi.encode(authority)
+    );
+    bytes32 initCodeHash = keccak256(initCode);
+    return computeCreate2Address(salt, initCodeHash);
+  }
+
+  function setCreate2Factory() internal {
```
miguelmtzinf: should we call this fn within this lib's functions themselves? instead of assuming it will be called before any call to the factory through this lib

tests/DeployUtils.sol
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  function deployHub(address authority, bytes32 salt) internal returns (IHub hub) {
+    bytes memory initCode = abi.encodePacked(
+      vm.getCode('src/hub/Hub.sol:Hub'),
+      abi.encode(authority)
+    );
+    return IHub(_create2Deploy(salt, initCode));
+  }
+
```
miguelmtzinf, line 57: I d like to split this lib into 2: one for factory stuff, another for deploy helper

tests/DeployUtils.sol
```diff
@@ -0,0 +1,149 @@
...
+  function deploySpokeInstance(address oracle, bytes32 salt) internal returns (ISpoke spoke) {
+    bytes memory initCode = abi.encodePacked(
+      vm.getCode('src/spoke/instances/SpokeInstance.sol:SpokeInstance'),
+      abi.encode(oracle)
+    );
+    return ISpoke(_create2Deploy(salt, initCode));
+  }
+
+  function deployProxifiedSpokeInstance(
```
miguelmtzinf: I think we can just call ir `deploySpokeInstance` or just `deploySpoke`
if we want to expose a deploy helper for the implenentation only, we can use `deploySpokeImplementation`?

Kogaroshi: 9b7125a225af5ed08531d8718db3c8e1eb860d41

tests/DeployUtils.sol
Kogaroshi: 9b7125a225af5ed08531d8718db3c8e1eb860d41

tests/DeployUtils.sol
Kogaroshi: 9b7125a225af5ed08531d8718db3c8e1eb860d41

tests/DeployUtils.sol
Kogaroshi: 9b7125a225af5ed08531d8718db3c8e1eb860d41

tests/DeployUtils.sol
Kogaroshi: 9b7125a225af5ed08531d8718db3c8e1eb860d41

src/spoke/interfaces/ISpoke.sol
Kogaroshi: 9b7125a225af5ed08531d8718db3c8e1eb860d41

src/spoke/interfaces/ISpoke.sol
```diff
@@ -436,6 +436,8 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IMulticall {
     bytes32 permitS
   ) external;
 
+  function initialize(address authority) external;
```
DhairyaSethi: still here?

tests/Create2Utils.sol
```diff
@@ -0,0 +1,68 @@
...
+    if (_isContractDeployed(CREATE2_FACTORY)) {
+      return;
+    }
+    vm.etch(CREATE2_FACTORY, CREATE2_FACTORY_BYTECODE);
+  }
+
+  function create2Deploy(bytes32 salt, bytes memory bytecode) internal returns (address) {
+    if (_isContractDeployed(CREATE2_FACTORY) == false) {
+      revert NoCreate2Factory();
+    }
```
DhairyaSethi: require style revert

src/spoke/interfaces/ISpoke.sol
Kogaroshi: needed for test to initialize(), since we don't import the `Spoke.sol` contract anymore.
Other solution is a `ISpokeInstance` (either in src or in mocks) just for that function, wdyt ?

DhairyaSethi: we can manually create this interface temp in the test file, i think thats fine

tests/Create2Utils.sol
Kogaroshi: b9fcd227652ab1342d4e58c95bbf9dc4c87532d8

src/spoke/interfaces/ISpoke.sol
Kogaroshi: b9fcd227652ab1342d4e58c95bbf9dc4c87532d8

foundry.toml
```diff
@@ -7,11 +7,23 @@ fs_permissions = [{ access = "read", path = "tests/mocks/JsonBindings.sol" }]
 solc_version = "0.8.28"
 evm_version = "cancun"
 optimizer = true
-optimizer_runs = 200
+optimizer_runs = 444444444444
 bytecode_hash = "none"
 gas_snapshot_check = false
 gas_limit = 1099511627776
 
+additional_compiler_profiles = [
```
avniculae, line 15: where do we use these profiles? isn't `compilation_restrictions` below sufficient?

tests/mocks/DeployWrapper.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {DeployUtils} from 'tests/DeployUtils.sol';
+
+contract DeployWrapper {
```
avniculae, line 7: this wrapper is just casting to `address` right? do we need it?

tests/mocks/ISpokeInstance.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.20;
+
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+
+interface ISpokeInstance is ISpoke {
```
avniculae, line 7: why is the interface needed in test? can't we just use `SpokeInstance`?

foundry.toml
Kogaroshi: this defines the profiles, and the `compilation_restrictions` applies them to specific contracts/dir, as long as parameters are matching one the defined profile

tests/mocks/DeployWrapper.sol
Kogaroshi: It is used for `Hub.config.t.sol` & `Spoke.config.t.sol` for the deploy revert tests, as the internal expect revert cheatcode does not work without this wrapper.

tests/mocks/ISpokeInstance.sol
Kogaroshi: We cannot import `Spoke`, `SpokeInstance` and `Hub` in the tests, are the are now compiled under the `restrictions` defined in the config file, which differs from the tests compiler profile, and causes import to fail.
So instead we rely on the interfaces, and this is also why we have new deploy methods for those via-IR built contracts in the test helper libs.

Kogaroshi: Note that if via-IR was applied codebase wide (on src/ & tests/ with the same configuration), this issue would not exist, and the tests would be the same as the current branch.
But we prefer not to build via-IR the contracts that do not need it, and absolutely don't want to compile the test files with via-IR, and keep the current behaviors we have in the base branch.

tests/unit/Hub/Hub.Config.t.sol
```diff
@@ -23,9 +23,11 @@ contract HubConfigTest is HubBase {
     );
   }
 
-  function test_hub_deploy_revertsWith_InvalidAddress() public {
-    vm.expectRevert(IHub.InvalidAddress.selector, address(hub1));
-    new Hub(address(0));
+  function test_hub_deploy_reverts_on_InvalidConstructorInput() public {
```
avniculae, line 26:
```suggestion
  function test_deployHub_revertsWith_InvalidConstructorInput() public {
```

I know that this convention was not fully followed in the past months, but I think we should still aim to follow it

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -9,29 +9,30 @@ contract SpokeConfigTest is SpokeBase {
   using PercentageMath for uint256;
 
   function test_spoke_deploy() public {
-    address predictedSpokeAddress = vm.computeCreateAddress(
-      address(this),
-      vm.getNonce(address(this))
-    );
     address oracle = makeAddr('AaveOracle');
+    address predictedSpokeAddress = DeployUtils.getDeterministicSpokeInstanceAddress(oracle);
```
avniculae: I think we didn't need this predictedSpokeAddress in the first place, we can remove I think

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -9,29 +9,30 @@ contract SpokeConfigTest is SpokeBase {
...
   function test_spoke_deploy_reverts_on_InvalidConstructorInput() public {
+    DeployWrapper deployer = new DeployWrapper();
+
     vm.expectRevert();
-    new SpokeInstance(address(0));
+    deployer.deploySpokeImplementation(address(0));
   }
 
-  function test_spoke_deploy_revertsWith_InvalidOracleDecimals() public {
+  function test_spoke_deploy_reverts_on_InvalidOracleDecimals() public {
```
avniculae, line 27:
```suggestion
  function test_deploySpoke_revertsWith_InvalidOracleDecimals() public {
```

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -9,29 +9,30 @@ contract SpokeConfigTest is SpokeBase {
...
+    DeployWrapper deployer = new DeployWrapper();
     address oracle = makeAddr('AaveOracle');
+
     vm.mockCall(oracle, abi.encodeCall(IPriceOracle.DECIMALS, ()), abi.encode(7));
-    vm.expectRevert(ISpoke.InvalidOracleDecimals.selector);
-    new SpokeInstance(oracle);
+    vm.expectRevert();
```
avniculae, line 32: why can't we check the error selector anymore?

tests/unit/Hub/Hub.Config.t.sol
Kogaroshi: this test was renamed to match same tests on the Spoke : https://github.com/aave/aave-v4/blob/4a0efac9b4afc01642266f9081169a506a971b48/tests/unit/Spoke/Spoke.Config.t.sol#L25
We either should keep both tests as is, or rename both then.
I went with current naming because in this case we do not expect the actual custom Error, but only a revert, because with the current deploy logic for those contract we cannot bubble up the deploy revert issue.

DhairyaSethi: it's correct as is yes, revertsWith is when you specify the exact err

tests/unit/Spoke/Spoke.Config.t.sol
Kogaroshi: same as https://github.com/aave/aave-v4/pull/893#discussion_r2691222472

tests/unit/Spoke/Spoke.Config.t.sol
Kogaroshi: see https://github.com/aave/aave-v4/pull/893#discussion_r2691222472

Due to how Create2Factory works, when we do deploy (https://github.com/aave/aave-v4/blob/13ae97663feb5afee47e2b2a7eb2b4b9e4fee634/tests/Create2Utils.sol#L35) and experience a revert, no error message is bubbled up by the Factory, and we instead need to verify the deployed address and revert with a new error if needed.

tests/DeployUtils.sol
```diff
@@ -0,0 +1,104 @@
...
+import {Create2Utils} from 'tests/Create2Utils.sol';
+
+library DeployUtils {
+  Vm internal constant vm = Vm(address(uint160(uint256(keccak256('hevm cheat code')))));
+
+  function deploySpokeImplementation(address oracle) internal returns (ISpoke) {
+    return deploySpokeImplementation(oracle, '');
+  }
+
+  function deploySpokeImplementation(address oracle, bytes32 salt) internal returns (ISpoke spoke) {
```
avniculae: should we return ISpokeInstance here?

tests/Base.t.sol
```diff
@@ -2239,16 +2241,12 @@ abstract contract Base is Test {
...
 
-    address spokeImpl = address(new SpokeInstance(address(oracle)));
-    ISpoke spoke = ISpoke(
-      _proxify(
-        deployer,
-        spokeImpl,
-        proxyAdminOwner,
-        abi.encodeCall(Spoke.initialize, (_accessManager))
-      )
+    ISpoke spoke = DeployUtils.deploySpoke(
```
avniculae: should we use named args here?

tests/DeployUtils.sol
Kogaroshi: `ISpokeInstance` is only used for the proxyfy `initData`, I think returning as `ISpoke` is better here.

tests/Create2Utils.sol
```diff
@@ -0,0 +1,66 @@
...
+
+      return deployedAt;
+    }
+  }
+
+  function _isContractDeployed(address instance) internal view returns (bool) {
+    return (instance.code.length > 0);
+  }
+
+  function computeCreate2Address(
```
avniculae, line 48: why not use cheatcodes?

```
    function computeCreate2Address(bytes32 salt, bytes32 initCodeHash, address deployer) external pure returns (address);
    function computeCreate2Address(bytes32 salt, bytes32 initCodeHash) external pure returns (address);
```

we can also change `create2_deployer` in fundry.toml

tests/Create2Utils.sol
```diff
@@ -0,0 +1,66 @@
...
+
+  function setCreate2Factory() internal {
+    if (_isContractDeployed(CREATE2_FACTORY)) {
+      return;
+    }
+    vm.etch(CREATE2_FACTORY, CREATE2_FACTORY_BYTECODE);
+  }
+
+  function create2Deploy(bytes32 salt, bytes memory bytecode) internal returns (address) {
+    require(_isContractDeployed(CREATE2_FACTORY), NoCreate2Factory());
```
avniculae, line 26: why can't we just populate code here on demand, and we never revert. this way we don't need `setCreate2Factory` anymore

tests/Base.t.sol
Kogaroshi: 2122b7039b493602f96a0f51c3846f7f81eb14b8

tests/unit/Spoke/Spoke.Config.t.sol
Kogaroshi: 2122b7039b493602f96a0f51c3846f7f81eb14b8

tests/Create2Utils.sol
Kogaroshi: this helpers were wrote with the optics of being used for deploy scripts as well if needed, where `setCreate2Factory` wouldn't be allowed.

foundry.toml
avniculae:
```
Once we have a set of restrictions we need to construct settings objects to satisfy them. This is done by adding additional_compiler_profiles, which is a mapping from profile name to settings overrides.
```

lol didn't know this, interesting, thanks

tests/Create2Utils.sol
Kogaroshi: we were using the cheatcodes before, @DhairyaSethi prefered we have scripts with the Factory.

foundry.toml
avniculae: they could just take in a profile name in compilation restrictions...

tests/Create2Utils.sol
DhairyaSethi: the address prediction stuff is resolved & merged on main now bw spoke & oracle. we can now just do normal create for tests here & do create2 stuff in the deployment engine PR 

(im ok w how this PR sets up the tests btw, but we can avoid to reduce footprint)

tests/DeployUtils.sol
```diff
@@ -0,0 +1,104 @@
...
+  function getDeterministicHubAddress(address authority) internal returns (address) {
+    return getDeterministicHubAddress(authority, '');
+  }
+
+  function getDeterministicHubAddress(address authority, bytes32 salt) internal returns (address) {
+    bytes memory initCode = abi.encodePacked(
+      vm.getCode('src/hub/Hub.sol:Hub'),
+      abi.encode(authority)
+    );
+    bytes32 initCodeHash = keccak256(initCode);
```
avniculae: should we have 2 helper fns? getHubInitCodeHash and getSpokeInstanceInitiCodeHash?

tests/unit/Hub/Hub.Config.t.sol
avniculae: I see, okay. We can define better conventions when we refactor tests e.g., revertsWith and revertsOn

tests/DeployUtils.sol
avniculae: we cast to it in a lot of tests

tests/Create2Utils.sol
avniculae: yeah we can move these utils outside of tests once we have deployment scripts, sg

tests/DeployUtils.sol
Kogaroshi: ec7f7b56eb205b48070987ccf585236f9307afc5

tests/DeployUtils.sol
Kogaroshi: ec7f7b56eb205b48070987ccf585236f9307afc5

tests/mocks/ISpokeInstance.sol
miguelmtzinf, line 1: why mock? we ll be importing a mock from deploy scripts later?

miguelmtzinf: same with `DeployWrapper`, but I think we intend to use a different thing for actuald eployment, correct?

tests/DeployUtils.sol
```diff
@@ -0,0 +1,99 @@
...
+  function deploySpokeImplementation(
+    address oracle,
+    bytes32 salt
+  ) internal returns (ISpokeInstance spoke) {
+    Create2Utils.setCreate2Factory();
+    return ISpokeInstance(Create2Utils.create2Deploy(salt, _getSpokeInstanceInitCode(oracle)));
+  }
+
+  function deploySpoke(
+    address deployer,
```
miguelmtzinf: why it needs `deployer`? not needed in proxy-based contracts

tests/mocks/ISpokeInstance.sol
Kogaroshi: It's only used in tests currently.
We can change its location in the Deploy Engine PR

tests/DeployUtils.sol
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  function _proxify(
+    address deployer,
+    address impl,
+    address proxyAdminOwner,
+    bytes memory initData
+  ) internal returns (address) {
+    vm.prank(deployer);
+    TransparentUpgradeableProxy proxy = new TransparentUpgradeableProxy(
```
miguelmtzinf, line 74: which compilation settings are we using for this contract?

tests/DeployUtils.sol
Kogaroshi: it's for `_proxify()`, leftovers from previous logic where we predicted addresses. Will remove

tests/DeployUtils.sol
Kogaroshi: The default config, so `optimizer = true, via_ir = false, optimizer_runs = 444444444444`

tests/Create2Utils.sol
```diff
@@ -0,0 +1,66 @@
...
+  error Create2DeploymentFailed();
+
+  // https://github.com/safe-global/safe-singleton-factory
+  address public constant CREATE2_FACTORY = 0x914d7Fec6aaC8cd542e72Bca78B30650d45643d7;
+  bytes internal constant CREATE2_FACTORY_BYTECODE =
+    hex'7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffe03601600081602082378035828234f58015156039578182fd5b8082525050506014600cf3';
+
+  Vm internal constant vm = Vm(address(uint160(uint256(keccak256('hevm cheat code')))));
+
+  function setCreate2Factory() internal {
```
miguelmtzinf: `loadCreate2Factory` makes more sense to me, but nbd.
apart from that, why not calling this in `create2Deploy`itself?

Kogaroshi: https://github.com/aave/aave-v4/pull/893#discussion_r2691307293

tests/DeployUtils.sol
Kogaroshi: 5913401517c47e0071d5ff794cfe3d140b2760bd

PR #895, opened 2025-10-10
opt: inline all hub.preview internal calls

DhairyaSethi:


yan-man:
actually @DhairyaSethi also there are some in liquidation logic that should be updated too

PR #901, opened 2025-10-13
fix : prevent key collision on KeyValueList add()

### LINKED ISSUE #900: Prevent encoding collision on KeyValueList
In `KeyValueList`:
- if `key` is `_MAX_KEY`, and `value` is zero, `pack` will return `0`
- The encoding here is the same as for the "empty" value, where unpack returns (0,0)
So if the list has the MAX_KEY, 0  element, it will be considered like an empty entry.

Proposed solution : 
`add()` should require `key < _MAX_KEY` (instead of `<=`)

Note : 
This issue should not be present in the current Spoke version, because we are using `uint24` for the key (and max `_MAX_KEY` is `2**32-1`).

Kogaroshi:
- Prevent key collision on `KeyValueList` add() function by requiring `key < _MAX_KEY` (instead of <=)
- This issue anyway should not be present in the current Spoke version, because we are using `uint24` for the key (and max _MAX_KEY is 2**32-1).

Fixes #900


tests/unit/KeyValueList.t.sol
```diff
@@ -5,9 +5,27 @@ pragma solidity ^0.8.0;
 import {Test} from 'forge-std/Test.sol';
 import {KeyValueList} from 'src/spoke/libraries/KeyValueList.sol';
 
+import {console2 as console} from 'forge-std/console2.sol';
```
DhairyaSethi: pls rm

tests/unit/KeyValueList.t.sol
```diff
@@ -119,3 +137,18 @@ contract KeyValueListTest is Test {
     return value % KeyValueList._MAX_VALUE;
   }
 }
+
+/// @dev needed to catch the emitted error from the library
```
DhairyaSethi: can add setting (we also use it for math lib tests) to avoid needing this
```
/// forge-config: default.allow_internal_expect_revert = true
```

tests/unit/KeyValueList.t.sol
Kogaroshi: oops mb

src/spoke/libraries/KeyValueList.sol
```diff
@@ -36,7 +36,7 @@ library KeyValueList {
 
   /// @notice Inserts packed `key`, `value` at `idx`. Reverts if data exceeds maximum allowed size.
   function add(List memory self, uint256 idx, uint256 key, uint256 value) internal pure {
-    require(key <= _MAX_KEY && value <= _MAX_VALUE, MaxDataSizeExceeded());
+    require(key < _MAX_KEY && value <= _MAX_VALUE, MaxDataSizeExceeded());
```
DhairyaSethi: let's limit value to also be strictly less than MAX_VALUE for consistency? doesn't really matter, maybe we rename constants?

tests/unit/KeyValueList.t.sol
```diff
@@ -5,9 +5,25 @@ pragma solidity ^0.8.0;
 import {Test} from 'forge-std/Test.sol';
 import {KeyValueList} from 'src/spoke/libraries/KeyValueList.sol';
 
+/// forge-config: default.allow_internal_expect_revert = true
 contract KeyValueListTest is Test {
   using KeyValueList for KeyValueList.List;
 
+  function test_fuzz_add(uint256 key, uint256 value) public {
```
DhairyaSethi, line 87: let's write a test where we `add` unique values only and assert only unique values are received on `get`?

src/spoke/libraries/KeyValueList.sol
Kogaroshi: I'm okay with current constant names

src/spoke/libraries/KeyValueList.sol
```diff
@@ -36,7 +36,7 @@ library KeyValueList {
 
   /// @notice Inserts packed `key`, `value` at `idx`. Reverts if data exceeds maximum allowed size.
   function add(List memory self, uint256 idx, uint256 key, uint256 value) internal pure {
-    require(key <= _MAX_KEY && value <= _MAX_VALUE, MaxDataSizeExceeded());
+    require(key < _MAX_KEY && value < _MAX_VALUE, MaxDataSizeExceeded());
```
miguelmtzinf, line 40: should this revert in `pack` instead? also, let's add a dev docs above

tests/unit/KeyValueList.t.sol
```diff
@@ -5,9 +5,79 @@ pragma solidity ^0.8.0;
...
+  function test_add_unique() public {
+    KeyValueList.List memory list = KeyValueList.init(9);
+
+    list.add(0, 1, 1);
+    list.add(1, 100, 1e15);
+    list.add(2, 100, 5e20);
+    list.add(3, 5e4, 5e20);
+    list.add(4, 5e4, 1e12);
+    list.add(5, 45e6, 2.5e50);
+    list.add(6, 45e6, 10);
```
DhairyaSethi: pls add edge cases in the explicit case, ie type(uint32).max - 1; type(uint224).max - 1 etc

tests/unit/KeyValueList.t.sol
```diff
@@ -5,9 +5,79 @@ pragma solidity ^0.8.0;
 import {Test} from 'forge-std/Test.sol';
 import {KeyValueList} from 'src/spoke/libraries/KeyValueList.sol';
 
+/// forge-config: default.allow_internal_expect_revert = true
 contract KeyValueListTest is Test {
   using KeyValueList for KeyValueList.List;
 
+  function test_add_unique() public {
```
DhairyaSethi, line 12: let's have a fuzz test for this as well?
fuzzed input can just be seed or array whichever is better wrt performance

src/spoke/libraries/KeyValueList.sol
```diff
@@ -35,8 +35,9 @@ library KeyValueList {
   }
 
   /// @notice Inserts packed `key`, `value` at `idx`. Reverts if data exceeds maximum allowed size.
```
yan-man, line 37: should we update the natspec here or move below into @dev too?

tests/unit/KeyValueList.t.sol
```diff
@@ -5,9 +5,106 @@ pragma solidity ^0.8.0;
...
+    list.add(3, 5e4, 5e20);
+    list.add(4, 5e4, 1e12);
+    list.add(5, 45e6, 2.5e50);
+    list.add(6, 45e6, 10);
+    vm.expectRevert(KeyValueList.MaxDataSizeExceeded.selector);
+    list.add(7, 4294967295, 10000000000);
+    vm.expectRevert(KeyValueList.MaxDataSizeExceeded.selector);
+    list.add(7, 45e8, 10000000000);
+    vm.expectRevert(KeyValueList.MaxDataSizeExceeded.selector);
+    list.add(type(uint32).max - 1, 75e9, 10000000000);
```
DhairyaSethi: uint32 max for index? 

i thought the suggestion was uint32.max at KEY would revert, and we test the edge case when key is uint32.max - 1 which doesn't revert

Kogaroshi: indeed, shouldn't fail is key is `uint32.max - 1`, here it fails because I put it as the index ...

PR #902, opened 2025-10-13
opt: inline all hub.preview internal calls

miguelmtzinf:
- Inline all hub conversions, and refactor `_applyPremiumDelta` function
- optimization so that there is more explicit rounding clarity and code simplification

mirroring #895

### DIFF
--- a/snapshots/Hub.Operations.json
+++ b/snapshots/Hub.Operations.json
@@ -1,6 +1,6 @@
 {
   "add": "113749",
-  "draw": "121860",
+  "draw": "121855",
   "eliminateDeficit: full": "57840",
   "eliminateDeficit: partial": "67416",
   "payFee": "73573",
--- a/snapshots/Spoke.Operations.ZeroRiskPremium.json
+++ b/snapshots/Spoke.Operations.ZeroRiskPremium.json
@@ -1,6 +1,6 @@
 {
-  "borrow: first": "198799",
-  "borrow: second action, same reserve": "179320",
+  "borrow: first": "198794",
+  "borrow: second action, same reserve": "179315",
   "liquidationCall: full": "290492",
   "liquidationCall: partial": "314586",
   "permitReserve + repay (multicall)": "235271",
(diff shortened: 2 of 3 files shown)


src/hub/Hub.sol
```diff
@@ -679,52 +681,71 @@ contract Hub is IHub, AccessManaged {
...
+    asset.premiumShares = premiumShares.toUint128();
+    asset.premiumOffset = premiumOffset.toUint128();
+    asset.realizedPremium = realizedPremium.toUint128();
+
+    uint256 premiumAfter = asset.toDrawnAssetsUp(premiumShares) - premiumOffset;
+    premiumAfter += realizedPremium;
+    // can increase due to precision loss on premium (drawn unchanged)
+    require(premiumAfter + premiumAmount - premiumBefore <= 2, InvalidPremiumChange());
+
+    // spoke premium change
```
DhairyaSethi, line 700: the apply & validate logic can be abstracted in a helper, no?

DhairyaSethi: something like https://github.com/aave/aave-v4/commit/caae863b75fae75dc9582e58a370a7ccaed8dd10 
can be made cleaner

src/hub/Hub.sol
```diff
@@ -765,43 +792,40 @@ contract Hub is IHub, AccessManaged {
     require(amount > 0, InvalidAmount());
     require(spoke.active, SpokeNotActive());
     uint256 drawCap = spoke.drawCap;
-    uint256 drawn = _getSpokeDrawn(spoke, assetId);
-    uint256 premium = _getSpokePremium(spoke, assetId);
-
+    uint256 owed = _getSpokeDrawn(asset, spoke) + _getSpokePremium(asset, spoke);
```
DhairyaSethi, line 794: remove assetId param here in _validateDraw params, it's unused

src/hub/Hub.sol
DhairyaSethi: done with [889e36b](https://github.com/aave/aave-v4/pull/902/commits/889e36bad8e29caa4a4af93204ba605d22a3c253)

PR #905, opened 2025-10-13
fix: reorder liquidation logic functions

miguelmtzinf:
Reorders `LiquidationLogic` functions to follow Solidity style guidelines.
Library functions that are used outside do not have `_` preffix and are placed first. 
Oversight from #864 .

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -294,7 +294,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       msg.sender
     );
 
-    _settlePremiumDebt(userPosition, premiumDelta.realizedDelta);
+    LiquidationLogic.settlePremiumDebt(userPosition, premiumDelta.realizedDelta);
     userPosition.drawnShares -= restoredShares.toUint128();
     if (userPosition.drawnShares == 0) {
       _positionStatus[onBehalfOf].setBorrowing(reserveId, false);
@@ -470,11 +470,6 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   }
 
   /// @inheritdoc ISpoke
-  function getLiquidationLogic() external pure returns (address) {
-    return address(LiquidationLogic);
-  }
-
-  /// @inheritdoc ISpoke
   function getLiquidationConfig() external view returns (LiquidationConfig memory) {
     return _liquidationConfig;
   }
@@ -638,6 +633,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   /// @inheritdoc ISpoke
   function DOMAIN_SEPARATOR() external view returns (bytes32) {
     return _domainSeparator();
+  }
+
+  /// @inheritdoc ISpoke
+  function getLiquidationLogic() external pure returns (address) {
+    return address(LiquidationLogic);
   }
 
   function _updateReservePriceSource(uint256 reserveId, address priceSource) internal {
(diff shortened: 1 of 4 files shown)


src/spoke/Spoke.sol
```diff
@@ -860,7 +860,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   }
 
   /// @notice Settles the premium debt by realizing change in premium and resetting premium shares and offset.
-  function _settlePremiumDebt(UserPosition storage userPosition, int256 realizedDelta) internal {
+  function settlePremiumDebt(UserPosition storage userPosition, int256 realizedDelta) internal {
```
DhairyaSethi: why this change here? internal methods should be prefixed with underscore

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -221,6 +221,52 @@ library LiquidationLogic {
...
+      });
+
+      uint256 drawnSharesLiquidated = reserve.hub.restore(
+        reserve.assetId,
+        drawnDebtToLiquidate,
+        premiumDebtToLiquidate,
+        premiumDelta,
+        params.liquidator
+      );
+      settlePremiumDebt(position, premiumDelta.realizedDelta);
```
DhairyaSethi: here we can syntax sugar but nbd
```solidity
position.settlePremiumDebt(premiumDelta.realizedDelta);
```

src/spoke/libraries/LiquidationLogic.sol
CheyenneAtapour, line 1: I see that `calculateLiquidationBonus` is sandwiched between two functions which are pre-fixed with `_`. I know we are ok with using no `_` prefix inside libraries, but it does make sense to me to keep this on the ones which are used as helpers inside larger functions (as it seems is the case). However, do we want a convention to separate these functions prefixed with `_` from the ones not prefixed with `_`?

CheyenneAtapour: In addition, it seems `internal` and `internal pure` are mixed together. One should show up before the other as a block

yan-man: yeah think `_liquidateCollateral` should go above the pure ones

`calculateLiquidationBonus` was an exception on the _ part bc we call it from spoke (stylistically)

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -221,6 +221,52 @@ library LiquidationLogic {
       });
   }
 
+  /// @dev Invoked by `liquidateUser` method.
+  /// @return True if the debt position becomes zero after restoring, false otherwise.
```
yan-man: this one's on me but perhaps we should just rm "false otherwise" on all these natspec, feels unnecessary and redundant

src/spoke/libraries/LiquidationLogic.sol
CheyenneAtapour: All library functions are allowed to not have the `_` prefix. I just mean in terms of ordering, perhaps it's nice to keep not mixed with `_` prefixed ones

miguelmtzinf: yeah, reordered so functions that are externally used are placed first.

src/spoke/Spoke.sol
miguelmtzinf: do we wanna use the same convention for Libraries than for contracts?
Majority of libraries we have are internal, and we are not using underscore preffix. I was leaning towards using preffix when it's internally used in the library, and no preffix when it's used out of the lib.
It can be confusing so open to just remove preffix in all libs' functions.

DhairyaSethi: libraries can have separate convention
we can do underscore on private functions there, ie stull only to be used inside the library & not in it's usage (external, internal)

so in the lib for consistency you can make this fn private if you don't like removing underscore, but for contracts would strictly enforce underscore on all internal & private methods pls bc their usage is by user, and not another contract

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -221,6 +221,82 @@ library LiquidationLogic {
       });
   }
 
+  function settlePremiumDebt(
```
miguelmtzinf: rm

yan-man: https://github.com/aave/aave-v4/pull/905/commits/668c2867a169ce4ee3813572b7fbdbede196e141

src/spoke/libraries/LiquidationLogic.sol
yan-man: https://github.com/aave/aave-v4/pull/905/commits/668c2867a169ce4ee3813572b7fbdbede196e141

src/spoke/libraries/LiquidationLogic.sol
yan-man: i also moved the pure ones down, and regular internal:

external

internal, in order of it being called:
- liquidateCollateral
- liquidateDebt
- settlePremDebt 

internal pure

src/spoke/Spoke.sol
yan-man: reverted this to include _

src/spoke/libraries/LiquidationLogic.sol
CheyenneAtapour: Why do we have an external one? And also, I thought we wanted to have the ones that are called not by other functions `private`

miguelmtzinf: External one is needed because of Spoke code size.
We cannot make them private in the library because of tests.

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: removed Spoke fn instead of lib one.

PR #906, opened 2025-10-13
Remove 2d map for PositionStatus

DhairyaSethi:
The 2D map was introduced to support more than 128 assets, which is a limitation in V3 today. This comes from using a single bitmap (256 bits) shared between collateral and debt flags. Since it’s not clear that supporting more than 128 reserves per Spoke is actually needed, we could revert to a single bitmap and cap support at 128 reserves, as this would reduce gas costs and simplify the code.

### DIFF
--- a/snapshots/NativeTokenGateway.Operations.json
+++ b/snapshots/NativeTokenGateway.Operations.json
@@ -1,8 +1,8 @@
 {
-  "borrowNative": "229604",
+  "borrowNative": "228104",
   "repayNative": "168312",
-  "supplyAsCollateralNative": "160373",
+  "supplyAsCollateralNative": "160081",
   "supplyNative": "136476",
-  "withdrawNative: full": "125620",
-  "withdrawNative: partial": "136825"
+  "withdrawNative: full": "125469",
+  "withdrawNative: partial": "136636"
 }
--- a/snapshots/SignatureGateway.Operations.json
+++ b/snapshots/SignatureGateway.Operations.json
@@ -1,10 +1,10 @@
 {
-  "borrowWithSig": "215893",
+  "borrowWithSig": "214647",
   "repayWithSig": "189160",
   "setSelfAsUserPositionManagerWithSig": "74858",
-  "setUsingAsCollateralWithSig": "85053",
+  "setUsingAsCollateralWithSig": "84761",
   "supplyWithSig": "153205",
-  "updateUserDynamicConfigWithSig": "62769",
-  "updateUserRiskPremiumWithSig": "61579",
-  "withdrawWithSig": "131713"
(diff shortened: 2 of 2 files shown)


src/spoke/interfaces/ISpoke.sol
```diff
@@ -97,7 +97,7 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @dev map The map of bitmap buckets for the position status.
```
miguelmtzinf, line 97:
```suggestion
  /// @dev map The bitmap of the position status.
```

src/spoke/libraries/PositionStatusMap.sol
```diff
@@ -104,25 +89,16 @@ library PositionStatusMap {
   /// @dev Ignores dirty bits beyond the configured `reserveCount` within the last bucket.
   /// @param fromReserveId The identifier of the reserve to start searching from.
   /// @return reserveId The reserve identifier for the next reserve that is borrowed or used as collateral.
```
miguelmtzinf, line 91:
```suggestion
  /// @return reserveId The previous reserveId that is borrowed or used as collateral, or `NOT_FOUND` if none is found.
```

src/spoke/libraries/PositionStatusMap.sol
miguelmtzinf, line 86:
```suggestion
  /// @notice Finds the previous borrowing or collateral reserve strictly before `fromReserveId`.
```

src/spoke/libraries/PositionStatusMap.sol
```diff
@@ -133,60 +109,18 @@ library PositionStatusMap {
...
-    }
-  }
-
-  /// @notice Finds the previous collateral reserve strictly before `fromReserveId`.
-  /// @dev The search starts at `fromReserveId` (exclusive) and scans backward across buckets.
-  /// @dev Returns `NOT_FOUND` if no collateral reserve exists before the bound.
-  /// @dev Ignores dirty bits beyond the configured `reserveCount` within the last bucket.
-  /// @param fromReserveId The exclusive upper bound to start from (this reserveId is not considered).
-  /// @return The previous collateral reserveId, or `NOT_FOUND` if none is found.
-  function nextCollateral(
```
miguelmtzinf, line 156: wasn't this used before?

miguelmtzinf:
Given that the added complexity is low and the gas overhead is not significant, we prefer to keep the 2D map in case a Spoke ends up listing more than 128 reserves. While this is unlikely, we do see potential use cases that could lead to a large number of listings, such as maturity-based assets, RWAs, or strategy vaults.

PR #914, opened 2025-10-15
rft: consistent `calculateRestoreAmount`

yan-man:
- uses a consistent `calculateRestoreAmount` calc between spoke/liquidation logic lib, similar to settlePremiumDebt
- also implements cap for safety from https://github.com/aave/aave-v4/pull/823/files


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -315,8 +317,9 @@ library LiquidationLogic {
     LiquidateDebtParams memory params
   ) internal returns (bool) {
     {
-      uint256 premiumDebtToLiquidate = params.premiumDebt.min(params.debtToLiquidate);
-      uint256 drawnDebtToLiquidate = params.debtToLiquidate - premiumDebtToLiquidate;
+      (uint256 drawnDebtToLiquidate, uint256 premiumDebtToLiquidate) = params
+        .debtToLiquidate
+        .calculateRestoreAmount(params.drawnDebt, params.premiumDebt);
```
miguelmtzinf, line 322: it's easy to miss that we are capping drawn debt here... as it's expected to be done in `_calculateLiquidationAmounts`.
We could do the following here, and cap the drawn debt there if we want to.
```suggestion
        .calculateRestoreAmount(params.debtToLiquidate, params.premiumDebt);
```

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -513,4 +516,18 @@ library LiquidationLogic {
     }
     return !isDebtPositionEmpty || borrowedCount > 1;
   }
+
+  function calculateRestoreAmount(
```
miguelmtzinf, line 520: i d move it up, so it's the first internal pure of the contract

yan-man:
decided against doing this

PR #915, opened 2025-10-16
Fix: Remaining Misc Items

CheyenneAtapour:
Note: Please don't merge yet, going through everything and can see if it's even worth merging.

### DIFF
--- a/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
@@ -336,7 +336,7 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
       newAvgCollateralFactor += collateralFactor * userSuppliedAmount;
     }
 
-    if (totalCollateralValue != 0) {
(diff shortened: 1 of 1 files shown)


tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
```diff
@@ -336,7 +336,7 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
       newAvgCollateralFactor += collateralFactor * userSuppliedAmount;
     }
 
-    if (totalCollateralValue != 0) {
+    if (totalCollateralValue > 0) {
```
yan-man, line 339: tbh dont think this really matters but nbd. Was this brought up somewhere as a suggestion?

CheyenneAtapour: This was a Stermi finding, but ofc applies to src more so to be consistent in style

CheyenneAtapour:
Not needed, initially started to address remaining small outstanding issues, but irrelevant

PR #918, opened 2025-10-16
feat: `liquidityFee` input for `addAsset` in `hubConfigurator`

yan-man:
- add `liquidityFee` as an input arg for `addAsset` method in `hubConfigurator` so that a liquidity fee can be set when a new asset is added. Optimizes the flow, as it no longer requires calling `updateLiquidityFee` immediately after listing a new asset to set a liquidity fee. 
- prevents potential misconfiguration and simplifies onboarding, as typical assets will include a liquidityFee > 0


src/hub/HubConfigurator.sol
```diff
@@ -24,17 +24,20 @@ contract HubConfigurator is Ownable2Step, IHubConfigurator {
     address hub,
     address underlying,
     address feeReceiver,
+    uint256 liquidityFee,
```
DhairyaSethi, line 27: is there an `addAsset` variant w/o `liquidityFee`? bc it reverts if passed as zero, do we not think zero liq fee will ever be the default case

yan-man: hm where does it revert on LF = 0? (just wrote a test for it w LF=0 but it passes fine) but yeah we were discussing having a separate variant or not today

DhairyaSethi: ah no i misread a test it's fine

tests/unit/HubConfigurator.t.sol
```diff
@@ -44,8 +44,14 @@ contract HubConfiguratorTest is HubBase {
     _addAsset({
       fetchErc20Decimals: vm.randomBool(),
       underlying: vm.randomAddress(),
-      decimals: bound(vm.randomUint(), 0, Constants.MAX_ALLOWED_UNDERLYING_DECIMALS).toUint8(),
```
DhairyaSethi, line 47: sadge this wasn't caught before

src/hub/HubConfigurator.sol
yan-man: i also typo'd liquidationFee so maybe was confusing

PR #921, opened 2025-10-16
fix: Return 0 when future round requested

CheyenneAtapour:
* Update unit price feed so that when a future roundId is requested from the UnitPriceFeed's `getRoundData`, the function will return all 0 values.
* this properly implements `getRoundData` to be fully compatible with the Chainlink `AggregatorV3Interface`


tests/unit/UnitPriceFeed.t.sol
```diff
@@ -30,7 +30,9 @@ contract UnitPriceFeedTest is Base {
   }
 
   function test_getRoundData() public {
-    uint80 _roundId = uint80(vm.randomUint());
+    uint80 skipTime = uint80(vm.randomUint());
```
DhairyaSethi:
```suggestion
    uint80 skipTime = vm.randomUint(80).toUint80();
```

tests/unit/UnitPriceFeed.t.sol
```diff
@@ -45,6 +47,24 @@ contract UnitPriceFeedTest is Base {
     assertEq(answeredInRound, roundId);
   }
 
+  function test_getRoundData_futureRound() public {
+    uint80 skipTime = uint80(vm.randomUint(0, type(uint80).max - 1));
+    skip(skipTime);
+    uint80 _roundId = uint80(vm.randomUint(skipTime + 1, type(uint80).max));
```
DhairyaSethi:
```suggestion
    uint80 skipTime = vm.randomUint(0, type(uint80).max - 1).toUint80();
    skip(skipTime);
    uint80 _roundId = uint80(vm.randomUint(skipTime + 1, type(uint80).max));
```

CheyenneAtapour: last line, did you mean this?

 vm.randomUint(skipTime + 1, type(uint80).max).toUint80();

tests/unit/UnitPriceFeed.t.sol
CheyenneAtapour: Done

src/misc/UnitPriceFeed.sol
```diff
@@ -48,11 +48,13 @@ contract UnitPriceFeed is AggregatorV3Interface {
       uint80 answeredInRound
     )
   {
-    roundId = _roundId;
-    answer = _units;
-    startedAt = _roundId;
-    updatedAt = _roundId;
-    answeredInRound = _roundId;
+    if (_roundId <= uint80(block.timestamp)) {
```
DhairyaSethi, line 51: would rather it be more explicit with if in future return (0, 0, 0, ) like chainlink oracle but nbd

PR #927, opened 2025-10-17
[WIP]: fix: Liquidity growth calculation in fees logic & fee cache in transient storage

avniculae:
- https://prover.certora.com/output/3973608/d1918d257dee446590bfd764ecd4f651/?anonymousKey=a59f76058509f8901d57a8d94830bc2aa8472ee7


src/hub/libraries/AssetLogic.sol
```diff
@@ -128,29 +146,38 @@ library AssetLogic {
...
+      spokes[assetId][feeReceiver].addedShares += feeShares;
+      FEE_SHARES_SLOT.tstore(0);
+      emit IHub.AccrueFees(assetId, feeReceiver, feeShares);
+    }
+    FEE_AMOUNT_SLOT.tstore(0);
+  }
+
+  /// @notice Accrues interest and fees for the specified asset.
+  function accrue(IHub.Asset storage asset) internal {
     if (asset.lastUpdateTimestamp == block.timestamp) {
```
avniculae, line 167: add comment -- this cannot be removed now

src/hub/libraries/AssetLogic.sol
```diff
@@ -20,6 +21,18 @@ library AssetLogic {
   using WadRayMath for *;
   using MathUtils for uint256;
   using SafeCast for uint256;
+  using TransientSlot for *;
+
+  /// bytes32(uint256(keccak256("FEE_AMOUNT")) - 1)
+  TransientSlot.Uint256Slot internal constant FEE_AMOUNT_SLOT =
```
avniculae, line 27: pack fee amount and fee shares in the same slot

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: UNLICENSED
 // Copyright (c) 2025 Aave Labs
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.28;
```
avniculae, line 3: undo this (this was done to make certora rule compile)

src/hub/Hub.sol
```diff
@@ -575,15 +581,19 @@ contract Hub is IHub, AccessManaged {
   /// @inheritdoc IHubBase
   function getSpokeAddedAssets(uint256 assetId, address spoke) external view returns (uint256) {
     Asset storage asset = _assets[assetId];
-    uint256 unrealized = spoke == asset.feeReceiver ? asset.unrealizedFeeShares() : 0;
-    return asset.toAddedAssetsDown(_spokes[assetId][spoke].addedShares + unrealized);
+    uint256 feeShares = spoke == asset.feeReceiver
+      ? asset.getUnrealizedFeeShares(asset.getUnrealizedFeeAmount(asset.getDrawnIndex()))
```
avniculae, line 585: if this is queried after accrue but before fee shares are minted (in the same block), this will be wrong. however, I don't think this scenario exists (need to double check). If such scenario does not exist, should we fix it (by looking at the transient storage as well) just to be safer, or leave as is?

src/hub/Hub.sol
```diff
@@ -203,6 +206,7 @@ contract Hub is IHub, AccessManaged {
     asset.liquidity += amount.toUint128();
 
     asset.updateDrawnRate(assetId);
+    asset.mintFeeShares(_spokes, assetId);
```
avniculae, line 209: is there a way to ensure we always mint fee shares at the end of actions that accrue? current impl is a bit error-prone

src/hub/libraries/AssetLogic.sol
```diff
@@ -128,29 +146,38 @@ library AssetLogic {
...
+      asset.addedShares += feeShares;
+      spokes[assetId][feeReceiver].addedShares += feeShares;
+      FEE_SHARES_SLOT.tstore(0);
+      emit IHub.AccrueFees(assetId, feeReceiver, feeShares);
+    }
+    FEE_AMOUNT_SLOT.tstore(0);
+  }
+
+  /// @notice Accrues interest and fees for the specified asset.
+  function accrue(IHub.Asset storage asset) internal {
```
yan-man, line 166: let's make sure to clearly document this and the need to couple this with the `mintFeeShares`

avniculae:
This was addressed in #933

PR #930, opened 2025-10-18
test: Gas snapshot worst case liquidation

CheyenneAtapour:
Please don't merge in, only for debugging and informational purposes

This PR examines the gas cost of the worst case scenario for liquidations and borrowing. As a user enables more collaterals, this increases the gas cost of actions which calculate user risk premium, as the algorithm for risk premium must loop through all collateral assets. Currently, OpenZeppelin's implementation of quicksort is being used, which has worst-case performance when the array is sorted (either in ascending or descending order). The purpose of this PR is to find the limit on the number of collaterals where actions on a user position succeed, and to avoid the scenario of a user's position being unable to be liquidated due to out-of-gas exceptions. This work was foundational to motivate the implementation in #974 

Currently, on the v0.5.6 release (commit: https://github.com/aave/aave-v4/commit/6959e3219b5506bf2acae18551cbb2a68a5b8fba), this test succeeds with 138 collaterals, and fails with 139.
The cost of calling the mock oracle for price is `6054` while the cost for a chainlink oracle is `15642`. The code here is currently run with calling a mock oracle. We are assuming a transaction gas limit of 16.78m gas, as per the Fusaka Ethereum upgrade. We also are running tests using `--isolate` flag (by enforcing this behavior at the top level of both test contracts).

The additional cost to borrow another reserve is non-linear, and so ranges from approximately 86k to 106k gas (the more reserves a user is borrowing, the more the cost of the additional collateral is).
The additional cost to liquidate a user with an additional collateral also ranges from approximately 102k to 128k gas.

Additionally, please note that the liquidation function fails at 167 collaterals due to StackOverflow (not gas). We believe this is due to the recursion within OpenZeppelin's quicksort implementation.


tests/unit/Spoke/Spoke.MaxReserves.t.sol
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    console.log('exited loop');
+
+    skip(1000 days);
+    // Alice can be liquidated
+    assertLe(
+      spoke1.getUserAccountData(alice).healthFactor,
+      spoke1.HEALTH_FACTOR_LIQUIDATION_THRESHOLD()
+    );
```
DhairyaSethi, line 46: run test with isolate and snapshot gas taken by this pls

tests/unit/Spoke/Spoke.MaxReserves.t.sol
```diff
@@ -0,0 +1,172 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Spoke/SpokeBase.t.sol';
+
+/// forge-config: default.isolate = true
```
miguelmtzinf, line 7: not sure this can be applied to the entire test contract. Ive seen it only applied to functions. Can you double check

CheyenneAtapour: Yes, I just tested it. I added a brand new test in the same file, declaring and not declaring this config at the top of the contract. When I have this line, the test behaves the same as if I explicitly use --isolate flag on the compiler and not having this line on the contract

PR #933, opened 2025-10-19
fix: Supply share price

### LINKED ISSUE #932: Supply share price can decrease between a read and a write operation
While share price does not decrease between 2 write operations (i.e., between 2 accrue operations), it might decrease between a read and a write operation (i.e., between preview and accrue operations).This happens because of the underestimation of the liquidity growth when computing the fee amount, as well as the minting of the shares in the accrue operation (or for accounting its effect in a preview operation).

avniculae:
- restores the invariant that supply exchange rate is monotonically increasing, even between write/read operations
- fixes the underestimation of the liquidity growth when computing the fee amount, which was causing supply exchange rate to decrease in certain edge cases between an accrual, and then a view operation, and then another accrual
- separates the minting of the shares from the accrue operation. Minting fee shares becomes a permissioned action. This also reduces gas consumption during accruals. 
- addresses an issue where liquidityFee can be minimized due to rounding through repeated accrual actions

closes #932

### DIFF
--- a/tests/unit/Hub/Hub.Config.t.sol
+++ b/tests/unit/Hub/Hub.Config.t.sol
@@ -565,7 +565,6 @@ contract HubConfigTest is HubBase {
   function test_updateAssetConfig_oldFeeReceiver_flags() public {
     _test_updateAssetConfig_oldFeeReceiver_flags({active: true, paused: true});
     _test_updateAssetConfig_oldFeeReceiver_flags({active: true, paused: false});
-
     _test_updateAssetConfig_oldFeeReceiver_flags({active: false, paused: true});
     _test_updateAssetConfig_oldFeeReceiver_flags({active: false, paused: false});
   }


src/hub/libraries/AssetLogic.sol
```diff
@@ -170,24 +160,21 @@ library AssetLogic {
...
-    return feeAmount.toSharesDown(asset.totalAddedAssets() - feeAmount, asset.addedShares);
-  }
-
-  /// @notice Calculates the amount of unrealized fee shares since last accrual.
-  function unrealizedFeeShares(IHub.Asset storage asset) internal view returns (uint256) {
-    return asset.getFeeShares(asset.getDrawnIndex().uncheckedSub(asset.drawnIndex));
+    uint256 liquidityGrowth = asset.drawnShares.rayMulUp(drawnIndex) -
+      asset.drawnShares.rayMulUp(lastDrawnIndex) +
+      asset.premiumShares.rayMulUp(drawnIndex) -
+      asset.premiumShares.rayMulUp(lastDrawnIndex);
```
DhairyaSethi: we can cache drawn & premium shares also here

src/hub/Hub.sol
```diff
@@ -555,15 +566,15 @@ contract Hub is IHub, AccessManaged {
...
+    uint256 feeShares = spoke == asset.feeReceiver ? asset.toAddedSharesDown(asset.feeAmount) : 0;
+    return asset.toAddedAssetsDown(_spokes[assetId][spoke].addedShares + feeShares);
   }
 
   /// @inheritdoc IHubBase
   function getSpokeAddedShares(uint256 assetId, address spoke) external view returns (uint256) {
     Asset storage asset = _assets[assetId];
-    uint256 unrealized = spoke == asset.feeReceiver ? asset.unrealizedFeeShares() : 0;
-    return _spokes[assetId][spoke].addedShares + unrealized;
+    uint256 feeShares = spoke == asset.feeReceiver ? asset.toAddedSharesDown(asset.feeAmount) : 0;
```
avniculae: this feels inconsistent -- we should either not consider unminted fee shares at all, or consider asset.feeAmount + unrealized fees, wdyt? @DhairyaSethi @yan-man @miguelmtzinf

avniculae: I assume we should not cnsider unminted fee shares, since they are not part of the "system" yet, right? (not part of total added shares and assets)

DhairyaSethi: yes agree, with this change we should disregard this special condition for `feeReceiver`; it's a little tricky for `totalAddedAssets` bc the `feeAmount` does belong to the `feeReceiver` but it's not supplied/earning yield -> so would avoid there as well

avniculae: [3daae9b](https://github.com/aave/aave-v4/pull/933/commits/3daae9bdd64811e43a132e905087e56fbdddb64b)

src/hub/libraries/AssetLogic.sol
avniculae: [4c78a13](https://github.com/aave/aave-v4/pull/933/commits/4c78a1328f557dab00b099e69d8e2995e807df6e)

src/hub/Hub.sol
```diff
@@ -385,20 +387,27 @@ contract Hub is IHub, AccessManaged {
...
     asset.updateDrawnRate(assetId);
 
     emit TransferShares(assetId, msg.sender, toSpoke, shares);
   }
 
+  /// @inheritdoc IHub
+  function mintFeeShares(uint256 assetId) external returns (uint256) {
+    Asset storage asset = _assets[assetId];
+    asset.accrue();
+    return _mintFeeShares(asset, assetId);
```
DhairyaSethi: update drawn rate missing; new liquidity is added in the system

src/hub/Hub.sol
```diff
@@ -721,6 +728,21 @@ contract Hub is IHub, AccessManaged {
...
+  function _mintFeeShares(Asset storage asset, uint256 assetId) internal returns (uint256) {
+    uint256 feeAmount = asset.feeAmount;
+    uint128 feeShares;
+    if (feeAmount > 0) {
+      feeShares = asset.toAddedSharesDown(feeAmount).toUint128();
+      address feeReceiver = asset.feeReceiver;
+      asset.addedShares += feeShares;
+      _spokes[assetId][feeReceiver].addedShares += feeShares;
+      asset.feeAmount = 0;
+      emit AccrueFees(assetId, feeReceiver, feeShares);
```
DhairyaSethi: let's also emit feeAmount so backend knows exactly how much to reduce from that asset property 
it should not add any extra cost to user facing methods 
also let's add gas test for mint fee shares method

src/hub/Hub.sol
```diff
@@ -721,6 +728,21 @@ contract Hub is IHub, AccessManaged {
...
 
+  function _mintFeeShares(Asset storage asset, uint256 assetId) internal returns (uint256) {
+    uint256 feeAmount = asset.feeAmount;
+    uint128 feeShares;
+    if (feeAmount > 0) {
+      feeShares = asset.toAddedSharesDown(feeAmount).toUint128();
+      address feeReceiver = asset.feeReceiver;
+      asset.addedShares += feeShares;
+      _spokes[assetId][feeReceiver].addedShares += feeShares;
+      asset.feeAmount = 0;
```
DhairyaSethi: we dont need to add to liquidity right

avniculae: no, liquidity is correct, but we are subtracting feeamount from it in total added assets

src/hub/Hub.sol
avniculae: [42ed5f8](https://github.com/aave/aave-v4/pull/933/commits/42ed5f8016ceebcf24fff7519abecff8ebb209f4)

src/hub/Hub.sol
```diff
@@ -385,20 +387,28 @@ contract Hub is IHub, AccessManaged {
...
     emit TransferShares(assetId, msg.sender, toSpoke, shares);
   }
 
+  /// @inheritdoc IHub
+  function mintFeeShares(uint256 assetId) external returns (uint256) {
+    Asset storage asset = _assets[assetId];
+    asset.accrue();
+    asset.updateDrawnRate(assetId);
+    return _mintFeeShares(asset, assetId);
+  }
```
```suggestion
    uint256 feeShares = _mintFeeShares(asset, assetId);
    asset.updateDrawnRate(assetId);
    return feeShares;
  }
```

DhairyaSethi: this doesn't matter (bc affected state, ie supply share price, doesn't contribute to the ir params) but it's just for consistency, lmk wdyt

src/hub/Hub.sol
avniculae: also renamed event: [b382cc6](https://github.com/aave/aave-v4/pull/933/commits/b382cc66dfbe0cb6821a730203d794b04d5eb66c)

src/hub/Hub.sol
avniculae: it's not needed right?

src/hub/Hub.sol
avniculae: also gas test here [42ed5f8](https://github.com/aave/aave-v4/pull/933/commits/42ed5f8016ceebcf24fff7519abecff8ebb209f4)

src/hub/interfaces/IHub.sol
```diff
@@ -33,6 +33,8 @@ interface IHub is IHubBase, IAccessManaged {
     address reinvestmentController;
     //
     address feeReceiver;
+    //
+    uint128 feeAmount;
     uint16 liquidityFee;
```
DhairyaSethi: ok an opt which can save reading a slot for supply share price 
if we make all amounts uint112 (instead of 128) and we pack liquidity, feeAmount together, pack liquidity fee with premiumShares & drawnShares i think that would be optimal

DhairyaSethi: liquidity fee is much more optimally packed with drawnIndex or drawnShares, premiumShares

src/hub/libraries/AssetLogic.sol
```diff
@@ -54,74 +54,75 @@ library AssetLogic {
...
   ) internal view returns (uint256) {
-    return assets.toSharesDown(asset.totalAddedAssets(), asset.totalAddedShares());
+    return assets.toSharesDown(asset.totalAddedAssets(), asset.addedShares);
   }
 
   /// @notice Updates the drawn rate of a specified asset.
   /// @dev Premium debt is not used in the interest rate calculation.
   /// @dev Uses last stored index, asset accrual should have already occurred.
   function updateDrawnRate(IHub.Asset storage asset, uint256 assetId) internal {
+    // asset accrual should have already occurred
```
DhairyaSethi: add as a dev comment pls

src/hub/Hub.sol
yan-man: ha i think just consistency for:
accrue
action
updateDrawnRate

tests/Base.t.sol
```diff
@@ -2607,4 +2609,35 @@ abstract contract Base is Test {
...
+  ) internal view returns (uint256) {
+    uint256 expectedFeeAmount = hub.getAsset(assetId).feeAmount +
+      _calcUnrealizedFeeAmount(hub, assetId);
+    return hub.getSpokeAddedAssets(assetId, hub.getAsset(assetId).feeReceiver) + expectedFeeAmount;
+  }
+
+  function _getAddedAssetsWithFees(IHub hub, uint256 assetId) internal view returns (uint256) {
+    return
+      hub.getAddedAssets(assetId) +
+      IHub(address(hub)).getAsset(assetId).feeAmount +
```
yan-man: shouldnt need this casting

src/hub/Hub.sol
DhairyaSethi: yeah not needed just consistency

src/hub/interfaces/IHub.sol
```diff
@@ -258,6 +266,11 @@ interface IHub is IHubBase, IAccessManaged {
   /// @param toSpoke The address of the recipient spoke.
   function transferShares(uint256 assetId, uint256 shares, address toSpoke) external;
 
+  /// @notice Mints shares for the accrued fees to the current fee receiver of the specified asset.
+  /// @param assetId The identifier of the asset.
+  /// @return The amount of shares minted.
+  function mintFeeShares(uint256 assetId) external returns (uint256);
```
yan-man: not user facing, but should we also return asset amt to be consistent w supply, withdraw, etc.

src/hub/Hub.sol
```diff
@@ -488,18 +498,20 @@ contract Hub is IHub, AccessManaged {
 
   /// @inheritdoc IHubBase
   function getAddedShares(uint256 assetId) external view returns (uint256) {
-    return _assets[assetId].totalAddedShares();
+    return _assets[assetId].addedShares;
   }
```
DhairyaSethi, line 509: new getter would be good getAssetFeeAmount()

src/hub/interfaces/IHub.sol
DhairyaSethi: not really, hub only returns shares. spoke returns both

src/hub/libraries/AssetLogic.sol
```diff
@@ -172,24 +160,24 @@ library AssetLogic {
```
yan-man, line 159: on L157, `lastUpdateTimestamp.toUint32()` for consistency

yan-man:
hm is there a test for the scenario that caused this whole issue
accrue
view
accrue

showing that the supply rate doesnt decrease now

src/hub/libraries/AssetLogic.sol
yan-man: and maybe "must" have already occurred, otherwise drawn will be incorrect and out of sync

CheyenneAtapour: This is already part of the last dev comment, so can just remove. @yan-man I think `must` implies it will revert / it cannot continue otherwise. `Should` seems appropriate to me just to ensure things are in sync and working properly

tests/Base.t.sol
```diff
@@ -1931,8 +1931,10 @@ abstract contract Base is Test {
   }
 
   /// @dev Helper function to withdraw fees from the treasury spoke
-  function withdrawLiquidityFees(uint256 assetId, uint256 amount) internal {
-    uint256 fees = hub1.getSpokeAddedAssets(assetId, address(treasurySpoke));
+  function _withdrawLiquidityFees(IHub hub, uint256 assetId, uint256 amount) internal {
```
CheyenneAtapour, line 1938: thank you for underscore prefix :)

tests/Utils.sol
```diff
@@ -18,7 +18,8 @@ library Utils {
     uint256 amount,
     address user
   ) internal returns (uint256) {
-    approve(IHub(address(hub)), assetId, user, amount);
+    IHub hub = IHub(address(hub));
```
CheyenneAtapour, line 21: should we just pass `IHub` to this function in the first place to simplify and remove this casting?

src/hub/libraries/AssetLogic.sol
DhairyaSethi: i mean we can read the last update timestamp in 32 bits also

src/hub/Hub.sol
```diff
@@ -721,6 +729,21 @@ contract Hub is IHub, AccessManaged {
     );
   }
 
+  function _mintFeeShares(Asset storage asset, uint256 assetId) internal returns (uint256) {
+    uint256 feeAmount = asset.feeAmount;
+    uint128 feeShares;
+    if (feeAmount > 0) {
```
DhairyaSethi: this is very important, nice. to make this no op we've usually been doing 
if feeAmount == 0: return 0

src/hub/libraries/AssetLogic.sol
avniculae: yeah I would rather make lastUpdateTimestamp uint32 directly

src/hub/Hub.sol
avniculae: should we include unrealized fees as well?

src/hub/Hub.sol
avniculae: [6f75828](https://github.com/aave/aave-v4/pull/933/commits/6f7582872968069188dcf1515c54d468d7c0a639)

tests/Utils.sol
avniculae: I wouldn't change utils fns in this PR

src/hub/libraries/AssetLogic.sol
avniculae: [6f75828](https://github.com/aave/aave-v4/pull/933/commits/6f7582872968069188dcf1515c54d468d7c0a639)

src/hub/libraries/AssetLogic.sol
avniculae: [6f75828](https://github.com/aave/aave-v4/pull/933/commits/6f7582872968069188dcf1515c54d468d7c0a639)

tests/Base.t.sol
avniculae: [6f75828](https://github.com/aave/aave-v4/pull/933/commits/6f7582872968069188dcf1515c54d468d7c0a639)

src/hub/Hub.sol
avniculae: [6f75828](https://github.com/aave/aave-v4/pull/933/commits/6f7582872968069188dcf1515c54d468d7c0a639)

src/hub/Hub.sol
avniculae: added it here [6f75828](https://github.com/aave/aave-v4/pull/933/commits/6f7582872968069188dcf1515c54d468d7c0a639), but question above still stands

src/hub/Hub.sol
```diff
@@ -731,16 +736,23 @@ contract Hub is IHub, AccessManaged {
...
-    if (feeAmount > 0) {
-      feeShares = asset.toAddedSharesDown(feeAmount).toUint128();
-      address feeReceiver = asset.feeReceiver;
-      asset.addedShares += feeShares;
-      _spokes[assetId][feeReceiver].addedShares += feeShares;
-      asset.feeAmount = 0;
-      emit MintFeeShares(assetId, feeReceiver, feeShares, feeAmount);
+    if (feeAmount == 0) {
+      return 0;
+    }
```
DhairyaSethi: not needed

src/hub/Hub.sol
DhairyaSethi: can you a quick test for the noop case also pls

src/hub/Hub.sol
avniculae: not needed, it's just to avoid conversion from assets into shares

src/hub/Hub.sol
avniculae: see `tests/unit/Hub/Hub.MintFeeShares.t.sol`

src/hub/Hub.sol
DhairyaSethi: i think `getAssetFeeAmount` should extrapolate and get the final value only?

for the stored value we have hub.getAsset().feeAmount

avniculae: agreed

src/hub/Hub.sol
```diff
@@ -141,6 +142,7 @@ contract Hub is IHub, AccessManaged {
     }
 
     if (asset.feeReceiver != config.feeReceiver) {
+      _mintFeeShares(asset, assetId);
```
DhairyaSethi, line 146: this skips & gives less than one share worth of fees of the old receiver to the new one, i think that's ok; let's document bc it will be raised at audit?

src/hub/Hub.sol
```diff
@@ -540,6 +553,10 @@ contract Hub is IHub, AccessManaged {
       });
   }
 
+  function getAssetFeeAmount(uint256 assetId) external view returns (uint256) {
+    return _assets[assetId].feeAmount;
```
```suggestion
    Asset storage asset = _assets[assetId];
    return asset.feeAmount + asset.getUnrealizedFeeAmount(asset.getDrawnIndex());
```

DhairyaSethi: similar to `getAssetIndex()` which returns the extrapolated value

avniculae: this is duplicate of this https://github.com/aave/aave-v4/pull/933#discussion_r2454413043 no?

src/hub/libraries/AssetLogic.sol
```diff
@@ -131,65 +131,52 @@ library AssetLogic {
...
-        MathUtils.calculateLinearInterest(asset.drawnRate, uint32(lastUpdateTimestamp))
+        MathUtils.calculateLinearInterest(asset.drawnRate, lastUpdateTimestamp)
       );
   }
 
   /// @notice Calculates the amount of fee shares derived from the index growth due to interest accrual.
   /// @dev The true liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee.
-  /// @param indexDelta The delta between the current and next drawn index.
-  function getFeeShares(
+  /// @param drawnIndex The current drawn index.
```
miguelmtzinf, line 161: a bit weird, because the `current` one is not supposed to be the stored one?
so this in practice is `newDrawnIndex`

src/hub/libraries/AssetLogic.sol
```diff
@@ -131,65 +131,52 @@ library AssetLogic {
...
+  function getUnrealizedFeeAmount(
     IHub.Asset storage asset,
-    uint256 indexDelta
+    uint256 drawnIndex
   ) internal view returns (uint256) {
-    if (indexDelta == 0) return 0;
+    uint256 lastDrawnIndex = asset.drawnIndex;
+    if (drawnIndex == lastDrawnIndex) return 0;
```
miguelmtzinf: i d flip it, and add brackets. same in the condition below

src/hub/interfaces/IHub.sol
```diff
@@ -97,11 +99,17 @@ interface IHub is IHubBase, IAccessManaged {
   /// @param config The new spoke configuration struct.
   event UpdateSpokeConfig(uint256 indexed assetId, address indexed spoke, SpokeConfig config);
 
-  /// @notice Emitted when fees are accrued to `feeReceiver`.
+  /// @notice Emitted when fees are minted to `feeReceiver`.
   /// @param assetId The identifier of the asset.
   /// @param spoke The address of the current feeReceiver.
-  /// @param shares The amount of shares accrued.
-  event AccrueFees(uint256 indexed assetId, address indexed spoke, uint256 shares);
```
miguelmtzinf, line 136: we could use this for signaling the accrual of fees in asset terms.

src/hub/libraries/AssetLogic.sol
avniculae: stored is not current, right? same as docs for `getAssetDrawnIndex`

src/hub/interfaces/IHub.sol
avniculae: Do you think it's a good idea to signal the accrual of fees when they are not part of the system (ie totalAddedAssets/Shares) yet?

src/hub/Hub.sol
```diff
@@ -385,20 +387,29 @@ contract Hub is IHub, AccessManaged {
...
+    asset.accrue();
     _validateTransferShares(asset, sender, receiver, shares);
     _transferShares(sender, receiver, shares);
     asset.updateDrawnRate(assetId);
 
     emit TransferShares(assetId, msg.sender, toSpoke, shares);
   }
 
+  /// @inheritdoc IHub
+  function mintFeeShares(uint256 assetId) external returns (uint256) {
```
miguelmtzinf: this new fn opens the door to questioning the need of fee receiver as an spoke, because it requires to be periodically called.

This fn can be a `claim` function instead so:
1. only callable by the admin (we can re-use `feeReceiver` or just use `restricted`)
2. called periodically to harvest or to reinvest
3. to reinvest, DAO needs to figure which Spoke to use (existing one v custom v 4626 lp vault)
4. simpler code: `addAsset` and `updateAsset` is simpler because no `feeReceiver` stuff

In contrast, keeping the `feeReceiver` logic fix the 3rd point now.

src/hub/libraries/AssetLogic.sol
avniculae: [9c654b6](https://github.com/aave/aave-v4/pull/933/commits/9c654b6f8a6b5b2555ccfa8dff1e499029355b31)

src/hub/Hub.sol
avniculae: [9c654b6](https://github.com/aave/aave-v4/pull/933/commits/9c654b6f8a6b5b2555ccfa8dff1e499029355b31)

src/hub/Hub.sol
miguelmtzinf: ah, `payFeeShares` would be impacted, and most probably liquidation fee would be keep in `asset.feeAmount` too.

src/hub/Hub.sol
avniculae: [cc6b014](https://github.com/aave/aave-v4/pull/933/commits/cc6b0144cf5b62b3481c5224a48409e943fc0a1d)

src/hub/Hub.sol
avniculae: [109cd88](https://github.com/aave/aave-v4/pull/933/commits/109cd884d0bd832bb275a9e44ee2050a1ec07bc2)

src/hub/Hub.sol
```diff
@@ -721,6 +737,24 @@ contract Hub is IHub, AccessManaged {
     );
   }
 
+  function _mintFeeShares(Asset storage asset, uint256 assetId) internal returns (uint256) {
+    uint256 feeAmount = asset.feeAmount;
+    uint128 feeShares = asset.toAddedSharesDown(feeAmount).toUint128();
+    if (feeShares == 0) {
```
miguelmtzinf: we could do perfect shares instead of donating

tests/misc/liquidity_growth.py
```diff
@@ -16,11 +17,11 @@
 rayMulUp = lambda a, b: (a * b + RAY - 1) / RAY
 rayMulDown = lambda a, b: (a * b) / RAY
 
-expected = rayMulUp(base, index2) - rayMulUp(base, index1) + rayMulUp(premium, index2) - rayMulUp(premium, index1)
-actual = rayMulDown(base, index2 - index1) + rayMulDown(premium, index2 - index1)
-# actual = rayMulDown(base + premium, index2 - index1) # incorrect
+trueLiquidityGrowth = rayMulUp(base, index2) - rayMulUp(base, index1) + rayMulUp(premium, index2) - rayMulUp(premium, index1)
+x = rayMulDown(base, index2 - index1) + rayMulDown(premium, index2 - index1) # incorrect
+# x = rayMulDown(base + premium, index2 - index1) # incorrect
```
DhairyaSethi: no i mean the main point is line 21 underestimates and 22 can overestimate -> pls clarify this too in the comment, or the rework the script to show

avniculae: under/over estimation is not relevant anymore. What we want to show is that these are not equivalent, since we are now interested in the true liquidity growth

DhairyaSethi: yes agree, id still add a comment that one is under & the other is over bc it's interesting

src/hub/Hub.sol
```diff
@@ -385,20 +388,29 @@ contract Hub is IHub, AccessManaged {
...
     asset.updateDrawnRate(assetId);
 
     emit TransferShares(assetId, msg.sender, toSpoke, shares);
   }
 
+  /// @inheritdoc IHub
+  function mintFeeShares(uint256 assetId) external returns (uint256) {
+    Asset storage asset = _assets[assetId];
+    asset.accrue();
+    uint256 feeShares = _mintFeeShares(asset, assetId);
```
avniculae: should we revert is asset is not active or paused? cc @miguelmtzinf @DhairyaSethi

src/hub/interfaces/IHub.sol
```diff
@@ -258,6 +266,11 @@ interface IHub is IHubBase, IAccessManaged {
   /// @param toSpoke The address of the recipient spoke.
   function transferShares(uint256 assetId, uint256 shares, address toSpoke) external;
 
+  /// @notice Mints shares for the accrued fees to the current fee receiver of the specified asset.
```
```suggestion
  /// @notice Mints shares to the fee receiver using accrued asset fees.
```

src/hub/interfaces/IHub.sol
```diff
@@ -33,6 +33,8 @@ interface IHub is IHubBase, IAccessManaged {
     address reinvestmentController;
     //
     address feeReceiver;
+    //
+    uint128 feeAmount;
```
miguelmtzinf: should we rename to `accruedFees` or `accruedFeeAmount`?

src/hub/libraries/AssetLogic.sol
```diff
@@ -131,65 +131,56 @@ library AssetLogic {
...
       return previousIndex;
     }
     return
       previousIndex.rayMulUp(
-        MathUtils.calculateLinearInterest(asset.drawnRate, uint32(lastUpdateTimestamp))
+        MathUtils.calculateLinearInterest(asset.drawnRate, lastUpdateTimestamp)
       );
   }
 
   /// @notice Calculates the amount of fee shares derived from the index growth due to interest accrual.
```
miguelmtzinf: fix docs please

src/hub/libraries/AssetLogic.sol
miguelmtzinf: yeah, so i d use `asset.drawnIndex` as `drawnIndex` and `newDrawnIndex` as input param?

src/hub/interfaces/IHub.sol
miguelmtzinf: missing docs too

src/hub/Hub.sol
```diff
@@ -721,6 +737,24 @@ contract Hub is IHub, AccessManaged {
...
+      return 0;
+    }
+
+    address feeReceiver = asset.feeReceiver;
+
+    asset.addedShares += feeShares;
+    _spokes[assetId][feeReceiver].addedShares += feeShares;
+
+    asset.feeAmount = 0;
+    emit MintFeeShares(assetId, feeReceiver, feeShares, feeAmount);
```
miguelmtzinf: should we signal with an `Add` too? there is change on liquidity accounting (different from before, where liquidity was already part of `totalAddedAssets`)

DhairyaSethi: we weren't before also, i think it's fine 

we don't signal with Add on transfer shares for a similar reason

src/hub/libraries/AssetLogic.sol
```diff
@@ -131,65 +131,56 @@ library AssetLogic {
...
   /// @dev The true liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee.
-  /// @param indexDelta The delta between the current and next drawn index.
-  function getFeeShares(
+  /// @param drawnIndex The current drawn index.
+  function getUnrealizedFeeAmount(
```
miguelmtzinf: wonder if this should return `feesAmount + unrealized` instead.

src/hub/Hub.sol
```diff
@@ -141,6 +142,8 @@ contract Hub is IHub, AccessManaged {
     }
 
     if (asset.feeReceiver != config.feeReceiver) {
+      // if fees are worth less than one share, no shares are minted and existing fees will be minted to the new fee receiver
```
```suggestion
      // if fees are worth less than one share, no shares are minted and remaining fees are left to be minted for the new fee receiver
```

src/hub/Hub.sol
```diff
@@ -385,20 +388,29 @@ contract Hub is IHub, AccessManaged {
...
     _transferShares(sender, receiver, shares);
     asset.updateDrawnRate(assetId);
 
     emit TransferShares(assetId, msg.sender, toSpoke, shares);
   }
 
+  /// @inheritdoc IHub
+  function mintFeeShares(uint256 assetId) external returns (uint256) {
+    Asset storage asset = _assets[assetId];
+    asset.accrue();
```
```suggestion
    require(assetId < _assetCount, AssetNotListed());
    Asset storage asset = _assets[assetId];
    
    asset.accrue();

```

src/hub/Hub.sol
miguelmtzinf: we must check active flag

src/hub/Hub.sol
miguelmtzinf: anyway, i d move this comment to a `@dev` in the function itself

src/hub/libraries/AssetLogic.sol
avniculae: call asset.drawnIndex previousIndex

src/hub/interfaces/IHub.sol
avniculae: add in UpdateAsset

src/hub/Hub.sol
avniculae: make fn restricted

src/hub/Hub.sol
avniculae: don't check flags of receiver spoke

src/hub/Hub.sol
avniculae: fix this https://github.com/aave/aave-v4-stermi-audit/issues/75

src/hub/libraries/AssetLogic.sol
avniculae: feesAmount are realized, but not minted

src/hub/interfaces/IHub.sol
avniculae: realizedFees, getUnrealizedFees

src/hub/Hub.sol
miguelmtzinf: not the same, in both cases before the liquidity was part of the accounting.
now, liquidity accounting is updated: assets from `feeAmount` to `liquidity`, addedShares increasing

src/hub/interfaces/IHub.sol
avniculae: this will be addressed separately in #953

src/hub/libraries/AssetLogic.sol
```diff
@@ -131,65 +131,56 @@ library AssetLogic {
...
     asset.drawnIndex = newDrawnIndex.toUint128();
     asset.lastUpdateTimestamp = block.timestamp.toUint32();
-
-    uint128 feeShares = asset.getFeeShares(indexDelta).toUint128();
-    if (feeShares > 0) {
-      address feeReceiver = asset.feeReceiver;
-      asset.addedShares += feeShares;
-      spokes[assetId][feeReceiver].addedShares += feeShares;
-      emit IHub.AccrueFees(assetId, feeReceiver, feeShares);
-    }
```
yan-man, line 155: before we were emitting on accrued feeShares, now we calc feeAmount but dont emit new fee amount, wondering if in `UpdateAsset` we should include this. And perhaps `lastUpdateTimestamp` as it gets updated too

src/hub/Hub.sol
avniculae: to be fixed in a separate PR

src/hub/libraries/AssetLogic.sol
```diff
@@ -115,7 +115,7 @@ library AssetLogic {
 
   /// @notice Updates the drawn rate of a specified asset.
   /// @dev Premium debt is not used in the interest rate calculation.
-  /// @dev Uses last stored index, asset accrual should have already occurred.
+  /// @dev Uses last stored index and realized fees, asset accrual should have already occurred.
```
yan-man: nit: realized fees is not used, just emitted

src/hub/libraries/AssetLogic.sol
yan-man: nvm see this is done in latest commit

src/hub/Hub.sol
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

src/hub/Hub.sol
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

tests/unit/Hub/Hub.MintFeeShares.t.sol
```diff
@@ -0,0 +1,138 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Hub/HubBase.t.sol';
+
+contract HubMintFeeSharesTest is HubBase {
+  function setUp() public override {
+    super.setUp();
+  }
```
yan-man: nit: i think unneeded?

src/hub/libraries/AssetLogic.sol
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

src/hub/interfaces/IHub.sol
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

src/hub/interfaces/IHub.sol
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

tests/misc/liquidity_growth.py
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

src/hub/Hub.sol
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

src/hub/interfaces/IHub.sol
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

tests/unit/Hub/Hub.MintFeeShares.t.sol
```diff
@@ -0,0 +1,138 @@
...
+    _addAndDrawLiquidity(
+      hub1,
+      daiAssetId,
+      bob,
+      address(spoke1),
+      3,
+      bob,
+      address(spoke1),
+      1,
+      365 days
```
yan-man: nit: unclear on what the values of 3 and 1 correspond to, perhaps could use named params

src/hub/libraries/AssetLogic.sol
avniculae: [edd7243](https://github.com/aave/aave-v4/pull/933/commits/edd7243bd655cb2575f1f9e68a0525f5945f0af8)

tests/unit/Hub/Hub.MintFeeShares.t.sol
```diff
@@ -0,0 +1,138 @@
...
+
+    // drawn index is 2.000...001
+    assertEq(hub1.getAssetDrawnIndex(daiAssetId), 2e27 + 1);
+
+    vm.recordLogs();
+    Utils.mintFeeShares(hub1, daiAssetId, ADMIN);
+    Vm.Log[] memory entries = vm.getRecordedLogs();
+    _assertEventNotEmitted(IHub.MintFeeShares.selector);
+
+    assertEq(hub1.getAsset(daiAssetId).realizedFees, 1, 'realized fees after');
```
yan-man, line 148: so here, given minting didnt happen, this `realizedFees` val matches the value before calling mintFeeShares right?

avniculae: yeah, this is the only scenario where realizedFees can be > 0 after a call to mintFeeShares

tests/unit/Hub/Hub.MintFeeShares.t.sol
```diff
@@ -0,0 +1,138 @@
...
+        asset.liquidity,
+        hub1.previewRestoreByShares(daiAssetId, hub1.getAssetDrawnShares(daiAssetId)),
+        asset.deficit,
+        asset.swept
+      )
+    );
+    uint256 mockRate = 0.3e27;
+    vm.mockCall(address(irStrategy), irCalldata, abi.encode(mockRate));
+
+    // after mintFeeShares, the fee shares should be the amount of the fees
```
yan-man, line 73: this is bc share price is 1 right

src/hub/libraries/AssetLogic.sol
avniculae: so still used no?

tests/unit/Hub/Hub.MintFeeShares.t.sol
```diff
@@ -0,0 +1,138 @@
...
+      365 days
+    );
+
+    // drawn index is 1.0000...002
+    assertEq(hub1.getAssetDrawnIndex(daiAssetId), 1e27 + 2);
+
+    _mockInterestRateRay(1e27 - 3);
+    updateLiquidityFee(hub1, daiAssetId, PercentageMath.PERCENTAGE_FACTOR);
+
+    // mint fee shares just to accrue (liquidity fee is 0, so no fees are minted)
```
yan-man, line 136: hm liquidityFee is set to 100% right?

can you add an assert after the mintFeeShares that the feeShares are indeed 0

also bc this test is testing mintFeeShares would ideally prefer to use another way of accruing to just isolate the action, maybe add 1 wei ?

src/hub/Hub.sol
avniculae: here it's from `feeAmount` to `addedShares`, not to `liquidity`. Fees that are not yet minted are still part of `liquidity`. I don't think we should emit an `Add` event here since it does not have the same effect as an add, wdyt?

tests/unit/Hub/Hub.MintFeeShares.t.sol
avniculae: I set it to 0 at the beginning of this test

src/hub/Hub.sol
```diff
@@ -720,6 +736,24 @@ contract Hub is IHub, AccessManaged {
...
+    address feeReceiver = asset.feeReceiver;
+
+    asset.addedShares += shares;
+    _spokes[assetId][feeReceiver].addedShares += shares;
+
+    asset.realizedFees = 0;
+    emit MintFeeShares(assetId, feeReceiver, shares, fees);
+
+    return shares;
+  }
```
DhairyaSethi, line 755: pls kill whitespaces

tests/unit/Hub/Hub.MintFeeShares.t.sol
avniculae: no, it's not 1 actually. this was added in https://github.com/aave/aave-v4/pull/937 , will remove it

src/hub/Hub.sol
avniculae: [b3f3ac7](https://github.com/aave/aave-v4/pull/933/commits/b3f3ac7b69ebc6b72376d25944233c807c64531c)

tests/unit/Hub/Hub.MintFeeShares.t.sol
avniculae: [b3f3ac7](https://github.com/aave/aave-v4/pull/933/commits/b3f3ac7b69ebc6b72376d25944233c807c64531c)

tests/unit/Hub/Hub.MintFeeShares.t.sol
avniculae: [b3f3ac7](https://github.com/aave/aave-v4/pull/933/commits/b3f3ac7b69ebc6b72376d25944233c807c64531c)

src/hub/Hub.sol
```diff
@@ -560,9 +560,9 @@ contract Hub is IHub, AccessManaged {
   }
 
   /// @inheritdoc IHub
-  function getAssetFeeAmount(uint256 assetId) external view returns (uint256) {
+  function getAssetAccruedFees(uint256 assetId) external view returns (uint256) {
```
miguelmtzinf, line 562: `getAssetFees` instead?

avniculae: [f6ec448](https://github.com/aave/aave-v4/pull/933/commits/f6ec448b2a4ebcbd5f36abdb4cf739e4a2037cb9)

src/hub/libraries/AssetLogic.sol
yan-man: ig just how I am thinking of `used` as being incorporated into a calculation or smth rather than showing its value

miguelmtzinf:
```suggestion
  /// @dev Uses last stored index; asset accrual should have already occurred.
```
agree with Yan

src/hub/libraries/AssetLogic.sol
```diff
@@ -54,142 +54,132 @@ library AssetLogic {
...
       liquidity: asset.liquidity,
-      drawn: asset.drawnShares.rayMulUp(drawnIndex),
+      drawn: asset.drawn(drawnIndex),
       deficit: asset.deficit,
       swept: asset.swept
     });
     asset.drawnRate = newDrawnRate.toUint96();
 
-    emit IHub.UpdateAsset(assetId, drawnIndex, newDrawnRate);
+    emit IHub.UpdateAsset(assetId, drawnIndex, newDrawnRate, asset.realizedFees);
```
miguelmtzinf, line 130: does this value contain `realized + unrealized`? only if this is called after `accrue` in the same tx right

avniculae: exactly, which is always happening. See dev comment at the top of the function

src/hub/interfaces/IHub.sol
```diff
@@ -111,9 +113,16 @@ interface IHub is IHubBase, IAccessManaged {
   /// @param assetId The identifier of the asset.
   /// @param drawnIndex The new drawn index of the asset.
   /// @param drawnRate The new drawn rate of the asset.
-  event UpdateAsset(uint256 indexed assetId, uint256 drawnIndex, uint256 drawnRate);
+  /// @param realizedFees The new realized fees of the asset (not yet minted).
```
```suggestion
  /// @param accruedFees The accrued fees of the asset since the last action.
```

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
```diff
@@ -148,7 +148,7 @@ contract SpokeAccrueLiquidityFeeEdgeCasesTest is SpokeBase {
       assertApproxEqAbs(totalOwedAfter, hub1.getAssetTotalOwed(assetId), 1);
 
       feesAccrued += totalOwedAfter - totalOwedBefore;
-      totalOwedBefore = totalOwedAfter;
+      totalOwedBefore = hub1.getAssetTotalOwed(assetId);
```
DhairyaSethi, line 151: lets apply the fix to the test below as well

src/hub/interfaces/IHub.sol
```diff
@@ -111,9 +113,16 @@ interface IHub is IHubBase, IAccessManaged {
...
+  event UpdateAsset(
+    uint256 indexed assetId,
+    uint256 drawnIndex,
+    uint256 drawnRate,
+    uint256 realizedFees
+  );
 
   /// @notice Emitted when an asset configuration is updated.
+  /// @dev If fee receiver is updated and fees are worth less than one share, no shares are minted and remaining fees are left to be minted for the new fee receiver.
```
miguelmtzinf: moves this to `mintFeeShares`, does not belong to here

src/hub/interfaces/IHub.sol
```diff
@@ -129,11 +138,17 @@ interface IHub is IHubBase, IAccessManaged {
   /// @param config The new spoke configuration struct.
   event UpdateSpokeConfig(uint256 indexed assetId, address indexed spoke, SpokeConfig config);
 
-  /// @notice Emitted when fees are accrued to `feeReceiver`.
+  /// @notice Emitted when fees are minted to `feeReceiver`.
```
```suggestion
  /// @notice Emitted when fees are minted to the fee receiver spoke.
```

src/hub/interfaces/IHub.sol
```diff
@@ -129,11 +138,17 @@ interface IHub is IHubBase, IAccessManaged {
...
   /// @param assetId The identifier of the asset.
   /// @param spoke The address of the current feeReceiver.
```
```suggestion
  /// @param feeReceiver The address of the current fee receiver spoke.
```

src/hub/interfaces/IHub.sol
```diff
@@ -290,6 +305,11 @@ interface IHub is IHubBase, IAccessManaged {
   /// @param toSpoke The address of the recipient spoke.
   function transferShares(uint256 assetId, uint256 shares, address toSpoke) external;
 
+  /// @notice Mints shares to the fee receiver using realized asset fees.
```
```suggestion
  /// @notice Mints shares to the fee receiver from accrued fees.
```

src/hub/interfaces/IHub.sol
```diff
@@ -317,6 +337,11 @@ interface IHub is IHubBase, IAccessManaged {
   /// @return The asset configuration struct.
   function getAssetConfig(uint256 assetId) external view returns (AssetConfig memory);
 
+  /// @notice Returns the amount of fees accrued but not yet minted for the specified asset.
+  /// @param assetId The identifier of the asset.
+  /// @return The amount of fees accrued but not yet minted.
+  function getAssetFees(uint256 assetId) external view returns (uint256);
```
miguelmtzinf: we could use `getAssetAccruedFees` also if feels more consistent with the rest

src/hub/interfaces/IHub.sol
```diff
@@ -51,6 +51,8 @@ interface IHub is IHubBase, IAccessManaged {
     address reinvestmentController;
     //
     address feeReceiver;
+    //
+    uint128 realizedFees;
```
miguelmtzinf, line 56: add docs above, explicity saying it's in asset units.

src/hub/interfaces/IHub.sol
```diff
@@ -317,6 +337,11 @@ interface IHub is IHubBase, IAccessManaged {
   /// @return The asset configuration struct.
   function getAssetConfig(uint256 assetId) external view returns (AssetConfig memory);
 
+  /// @notice Returns the amount of fees accrued but not yet minted for the specified asset.
+  /// @param assetId The identifier of the asset.
+  /// @return The amount of fees accrued but not yet minted.
```
```suggestion
  /// @notice Returns the accrued fees for the asset, expressed in asset units.
  /// @dev Accrued fees are excluded from total added assets.
  /// @param assetId The identifier of the asset.
  /// @return The amount of accrued fees.
```

src/hub/Hub.sol
```diff
@@ -385,20 +387,30 @@ contract Hub is IHub, AccessManaged {
...
+    asset.accrue();
     _validateTransferShares(asset, sender, receiver, shares);
     _transferShares(sender, receiver, shares);
     asset.updateDrawnRate(assetId);
 
     emit TransferShares(assetId, msg.sender, toSpoke, shares);
   }
 
+  /// @inheritdoc IHub
+  function mintFeeShares(uint256 assetId) external restricted returns (uint256) {
```
miguelmtzinf: does it make sense to move this up right below `setInterestRateData`? so all restricted functions are together.

src/hub/Hub.sol
```diff
@@ -720,6 +736,22 @@ contract Hub is IHub, AccessManaged {
     );
   }
 
+  function _mintFeeShares(Asset storage asset, uint256 assetId) internal returns (uint256) {
```
miguelmtzinf, line 738: add `@dev` saying it's no-op when shares is zero

src/hub/interfaces/IHub.sol
```diff
@@ -129,11 +138,17 @@ interface IHub is IHubBase, IAccessManaged {
...
+  /// @notice Emitted when fees are minted to `feeReceiver`.
   /// @param assetId The identifier of the asset.
   /// @param spoke The address of the current feeReceiver.
-  /// @param shares The amount of shares accrued.
-  event AccrueFees(uint256 indexed assetId, address indexed spoke, uint256 shares);
+  /// @param shares The amount of shares minted.
+  /// @param assets The amount of assets used to mint the shares.
+  event MintFeeShares(
+    uint256 indexed assetId,
+    address indexed spoke,
```
```suggestion
    address indexed feeReceiver,
```

src/hub/Hub.sol
miguelmtzinf: imo it's the same effect but not feeling strong about emitting the `Add` event

src/hub/Hub.sol
avniculae: yeah I think now it makes sense

src/hub/interfaces/IHub.sol
avniculae: > since the last action.

this is not correct, it's since last mint

src/hub/interfaces/IHub.sol
avniculae: to confirm, you want to revert this https://github.com/aave/aave-v4/pull/933#discussion_r2466157207 ?

src/hub/interfaces/IHub.sol
avniculae: 76b5105

src/hub/Hub.sol
avniculae: 76b5105

src/hub/Hub.sol
avniculae: 76b5105

src/hub/interfaces/IHub.sol
avniculae: 76b5105

src/hub/interfaces/IHub.sol
avniculae: 76b5105

src/hub/interfaces/IHub.sol
avniculae: [c245e10](https://github.com/aave/aave-v4/pull/933/commits/c245e10ad2b00c684aed007f5e758a3e53344d02)

src/hub/interfaces/IHub.sol
avniculae: [76b5105](https://github.com/aave/aave-v4/pull/933/commits/76b510571c736c7ceea0d78ee08d31f9d3c6e75b)

src/hub/interfaces/IHub.sol
avniculae: [76b5105](https://github.com/aave/aave-v4/pull/933/commits/76b510571c736c7ceea0d78ee08d31f9d3c6e75b)

src/hub/interfaces/IHub.sol
avniculae: [76b5105](https://github.com/aave/aave-v4/pull/933/commits/76b510571c736c7ceea0d78ee08d31f9d3c6e75b)

src/hub/interfaces/IHub.sol
avniculae: [76b5105](https://github.com/aave/aave-v4/pull/933/commits/76b510571c736c7ceea0d78ee08d31f9d3c6e75b)

tests/unit/Spoke/Spoke.AccrueLiquidityFee.EdgeCases.t.sol
avniculae: [76b5105](https://github.com/aave/aave-v4/pull/933/commits/76b510571c736c7ceea0d78ee08d31f9d3c6e75b)

src/hub/interfaces/IHub.sol
avniculae: [76b5105](https://github.com/aave/aave-v4/pull/933/commits/76b510571c736c7ceea0d78ee08d31f9d3c6e75b)

src/hub/interfaces/IHub.sol
```diff
@@ -290,6 +305,11 @@ interface IHub is IHubBase, IAccessManaged {
   /// @param toSpoke The address of the recipient spoke.
   function transferShares(uint256 assetId, uint256 shares, address toSpoke) external;
 
+  /// @notice Mints shares to the fee receiver from accrued fees.
+  /// @param assetId The identifier of the asset.
+  /// @return The amount of shares minted.
+  function mintFeeShares(uint256 assetId) external returns (uint256);
```
miguelmtzinf: move this up pls, same order as in `Hub`

src/hub/Hub.sol
```diff
@@ -720,6 +736,24 @@ contract Hub is IHub, AccessManaged {
     );
   }
 
+  /// @dev If fee receiver is updated and fees are worth less than one share, no shares are minted and remaining fees are left to be minted for the new fee receiver.
```
miguelmtzinf: this is now kinda up. I d put it in `updateAssetConfig` or just rm

src/hub/Hub.sol
```diff
@@ -720,6 +736,24 @@ contract Hub is IHub, AccessManaged {
...
+  /// @dev No op when fees are worth less than one share.
```
miguelmtzinf: this can also be added to mintFeeShares external fn

src/hub/interfaces/IHub.sol
miguelmtzinf: Let's change to the follwing
`/// @param accruedFees The accrued fees of the asset since the last mint.`
because what's "new"?

src/hub/Hub.sol
avniculae: [8481363](https://github.com/aave/aave-v4/pull/933/commits/8481363a3d318661052e8e05cc07988e1498d7de)

src/hub/Hub.sol
avniculae: [8481363](https://github.com/aave/aave-v4/pull/933/commits/8481363a3d318661052e8e05cc07988e1498d7de)

src/hub/interfaces/IHub.sol
avniculae: [8481363](https://github.com/aave/aave-v4/pull/933/commits/8481363a3d318661052e8e05cc07988e1498d7de)

src/hub/interfaces/IHub.sol
avniculae: [8481363](https://github.com/aave/aave-v4/pull/933/commits/8481363a3d318661052e8e05cc07988e1498d7de)

src/hub/Hub.sol
yan-man: agree with Alex - actually feeAmount goes completely into liquidity. Part thru minted shares, remaining as donation, so total liquidity remains unchanged

src/hub/interfaces/IHub.sol
miguelmtzinf, line 260:
```suggestion
  /// @dev If the fee receiver is updated, adds it as a new spoke with maximum add cap and zero draw cap, and sets old fee receiver caps to zero.
  /// @dev If the fee receiver is updated, accrued fees are minted as shares before the update if their value exceeds one share.
```

src/hub/Hub.sol
```diff
@@ -192,6 +192,16 @@ contract Hub is IHub, AccessManaged {
     asset.updateDrawnRate(assetId);
   }
 
+  /// @inheritdoc IHub
+  function mintFeeShares(uint256 assetId) external restricted returns (uint256) {
+    Asset storage asset = _assets[assetId];
+    asset.accrue();
+    require(_spokes[assetId][asset.feeReceiver].active, SpokeNotActive());
```
yan-man: so we specifically want to validate active flag on feeReceiver spoke on mintFeeShares, how about `updateAssetConfig`? 

seems a bit inconsistent to not do it there

src/hub/Hub.sol
```diff
@@ -192,6 +192,16 @@ contract Hub is IHub, AccessManaged {
...
+    uint256 feeShares = _mintFeeShares(asset, assetId);
```
yan-man, line 199: also maybe we can pass in spoke storage pointer, so that we're not having `_spokes[assetId][asset.feeReceiver]` here and also in `_mintFeeShares`

avniculae: agreed, we could optimise further, not sure if worth it though

DhairyaSethi: will be immutable, it should be the cleanest code we can deliver

DhairyaSethi: but i just want to move forward at this time

avniculae: imo clean != gas optimised. I find this cleaner tbh

yan-man: "find this cleaner" @avniculae you mean current version as is right

src/hub/interfaces/IHub.sol
```diff
@@ -258,6 +258,7 @@ interface IHub is IHubBase, IAccessManaged {
 
   /// @notice Updates the configuration of an asset.
   /// @dev If the fee receiver is updated, adds it as a new spoke with maximum add cap and zero draw cap, and sets old fee receiver caps to zero.
+  /// @dev If the fee receiver is updated, accrued fees are minted as shares before the update if their value exceeds one share.
```
yan-man, line 261: nit but not just exceeds 1 share, gte 1 share

yan-man:
```suggestion
  /// @dev If the fee receiver is updated, accrued fees are minted as shares before the update if its value exceeds one share.
```

src/hub/Hub.sol
```diff
@@ -720,6 +736,22 @@ contract Hub is IHub, AccessManaged {
...
+  function _mintFeeShares(Asset storage asset, uint256 assetId) internal returns (uint256) {
+    uint256 fees = asset.realizedFees;
+    uint128 shares = asset.toAddedSharesDown(fees).toUint128();
+    if (shares == 0) {
+      return 0;
+    }
+
+    address feeReceiver = asset.feeReceiver;
+    asset.addedShares += shares;
+    _spokes[assetId][feeReceiver].addedShares += shares;
```
CheyenneAtapour: could cache asset and spoke `addedShares` to potentially make this slightly cheaper, but not a big deal

src/hub/interfaces/IHub.sol
CheyenneAtapour: haha isn't "their" correct here, because it's talking about `accrued fees` plural?

yan-man: ig i'm thinking of "accrued fees" as a singular consolidated value

CheyenneAtapour: same, but to me it just sounds better with "their" the way it's currently written

src/hub/Hub.sol
avniculae: yes, but no strong opinion

src/hub/Hub.sol
```diff
@@ -195,10 +194,27 @@ contract Hub is IHub, AccessManaged {
...
+      return 0;
+    }
+
+    asset.addedShares += feeShares;
+    feeReceiverSpoke.addedShares += feeShares;
+    asset.realizedFees = 0;
+
     asset.updateDrawnRate(assetId);
+
+    emit MintFeeShares(assetId, feeReceiver, feeShares, fees);
```
avniculae: tests fail now bc we emit this after UpdateAsset even from `updateDrawnRate`

PR #937, opened 2025-10-20
test: resolve failing caused by new `liquidityFee` impl

yan-man:
- resolve failing tests from mintFeeShares implementation
- add new basic test for mintFeeShares

### DIFF
--- a/tests/unit/Hub/Hub.Config.t.sol
+++ b/tests/unit/Hub/Hub.Config.t.sol
@@ -424,6 +424,17 @@ contract HubConfigTest is HubBase {
       newConfig.feeReceiver != _getFeeReceiver(hub1, assetId) &&
       !hub1.isSpokeListed(assetId, newConfig.feeReceiver)
     ) {
+      if (_calcUnrealizedFeeAmount(hub1, assetId) > 0) {
+        vm.expectEmit(address(hub1));
+        emit IHub.AccrueFees(
+          assetId,
+          _getFeeReceiver(hub1, assetId),
+          hub1.previewAddByAssets(
+            assetId,
+            hub1.getAsset(assetId).feeAmount + _calcUnrealizedFeeAmount(hub1, assetId)
+          )
+        );
+      }
       vm.expectEmit(address(hub1));
       emit IHub.AddSpoke(assetId, newConfig.feeReceiver);
       vm.expectEmit(address(hub1));
@@ -504,6 +515,8 @@ contract HubConfigTest is HubBase {
     uint256 amount = 1000e18;
     _addLiquidity(assetId, amount);
     _drawLiquidity(assetId, amount, true);
+
+    skip(365 days);
 
     IHub.AssetConfig memory config = hub1.getAssetConfig(assetId);
     address oldFeeReceiver = config.feeReceiver;
--- a/tests/unit/Hub/Hub.MintFeeShares.t.sol
+++ b/tests/unit/Hub/Hub.MintFeeShares.t.sol
@@ -5,7 +5,8 @@ pragma solidity ^0.8.0;
 import 'tests/unit/Hub/HubBase.t.sol';
 
 contract HubMintFeeSharesTest is HubBase {
-  function test_mintFeeShares() public {
+  function setUp() public override {
+    super.setUp();
     // Create debt to build up fees on the existing treasury spoke
     _addAndDrawLiquidity(
       hub1,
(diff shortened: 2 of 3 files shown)


tests/Base.t.sol
```diff
@@ -2590,4 +2592,23 @@ abstract contract Base is Test {
   function _bpsToRay(uint256 bps) internal pure returns (uint256) {
     return (bps * WadRayMath.RAY) / PercentageMath.PERCENTAGE_FACTOR;
   }
+
+  /// @dev Calculate expected fee amount based on previous drawn index
+  function _calcExpectedFeeAmount(IHub hub, uint256 assetId) internal view returns (uint256) {
```
avniculae: I'd call this `_calcUnrealizedFeeAmount`

tests/unit/Hub/Hub.Config.t.sol
```diff
@@ -534,6 +534,7 @@ contract HubConfigTest is HubBase {
     uint256 oldFees = hub1.getSpokeAddedShares(assetId, oldFeeReceiver);
 
     skip(365 days);
+    hub1.mintFeeShares(assetId);
```
avniculae, line 554: why do we need this here?

tests/unit/Hub/Hub.MintFeeShares.t.sol
```diff
@@ -0,0 +1,51 @@
...
+
+    uint256 mintedShares = hub1.mintFeeShares(daiAssetId);
+
+    assertEq(mintedShares, expectedMintedShares, 'minted shares');
+    assertEq(hub1.getAsset(daiAssetId).feeAmount, 0, 'fee amount after');
+    assertEq(
+      hub1.getSpokeAddedShares(daiAssetId, feeReceiver),
+      expectedMintedShares,
+      'added shares'
+    );
```
avniculae, line 53: I'd also check asset added shares here, and also the share price not decreasing

avniculae: I'd also include a test to make sure accrual takes place before mint fee shares

tests/unit/Hub/Hub.Remove.t.sol
```diff
@@ -144,23 +144,11 @@ contract HubRemoveTest is HubBase {
...
-    );
-    assertEq(
-      assetData.addedShares,
-      hub1.getSpokeAddedShares(assetId, feeReceiver),
-      'asset addedShares after'
-    );
     assertEq(
       assetData.liquidity,
-      hub1.getSpokeAddedAssets(assetId, feeReceiver) + _calculateBurntInterest(hub1, assetId),
+      hub1.getAsset(assetId).feeAmount + _calculateBurntInterest(hub1, assetId),
```
avniculae, line 151: here we don't need to take into consideration unrealized fees bc no time has passed since last action right?

tests/unit/Hub/HubAccrueInterest.t.sol
```diff
@@ -112,7 +112,11 @@ contract HubAccrueInterestTest is Base {
 
     assertEq(elapsed, daiInfo.lastUpdateTimestamp - startTime);
     assertEq(daiInfo.drawnIndex, expectedDrawnIndex1, 'drawnIndex');
-    assertEq(hub1.getAddedAssets(daiAssetId), addAmount + addAmount2 + interest, 'addAmount');
+    assertEq(
+      _getAddedAssetsWithFee(hub1, daiAssetId),
```
avniculae: We should include unrealized fees here as well, even though it's not needed in this case bc no time has passed, wdyt?

I think we should test total added assets with unrealized fees as well

tests/unit/HubConfigurator.t.sol
```diff
@@ -407,6 +408,7 @@ contract HubConfiguratorTest is HubBase {
       100e18,
       365 days
     );
+    hub1.mintFeeShares(daiAssetId);
```
avniculae, line 411: can we remove this from here?

tests/unit/Hub/Hub.MintFeeShares.t.sol
avniculae: I'd also include a test where we change fee receiver and make sure the new mint is taking place for the new receiver (I know we have a similar test in updateAssetConfig)

tests/unit/Hub/Hub.Remove.t.sol
yan-man: yes that's right, since last remove action

PR #938, opened 2025-10-21
fix: refresh premium in same block

avniculae:


src/spoke/Spoke.sol
```diff
@@ -56,6 +58,10 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   /// @inheritdoc ISpoke
   address public immutable ORACLE;
 
+  /// keccak256('REFRESHED_IN_SAME_BLOCK')
+  bytes32 internal constant _REFRESHED_IN_SAME_BLOCK =
```
avniculae, line 62: we actually need block, not transaction

PR #939, opened 2025-10-21
feat: user risk premium is set to zero when underwater

DhairyaSethi:
- extend risk premium definition: if position is unhealthy, risk premium is reset to 0 
- implement notify (risk premium update) when supplying collateral & when enabling as collateral, so that a user position under water cannot enjoy a 0 risk premium by adding collateral to become healthy
- do not allow liquidator to have reserve enabled as collateral when receiving shares, for same reason as above 
- this feature was later reverted in #1005, as it was no longer needed once premium accounting was resolved in #991 

By setting risk premium to 0 when a user position is underwater (HF < 1), this ensures that malicious users cannot inflate the supply exchange rate when their position is unhealthy (in essence benefitting at the expense of protocol deficit), because the _notifyRiskPremiumUpdate in the updateUserRiskPremium method will be ignored when RP = 0 (which is the source of the 2 wei generation of premium debt that can occur). In cases where a position has a HF >= 1, the attack is not profitable as the user position will have to retain sufficient collateral, and will only damage their own position by generating instantaneous premium debt from repeatedly invoking updateUserRiskPremium.

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -384,7 +384,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     _validateSetUsingAsCollateral(_reserves[reserveId], usingAsCollateral);
     PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
 
-    // no op also ensures only new collateral without health factor check
+    // no op also ensures only new collateral is enabled and refreshed without health factor check
     if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) {
       return;
     }


src/spoke/Spoke.sol
```diff
@@ -204,6 +204,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     uint256 suppliedShares = reserve.hub.add(reserve.assetId, amount, msg.sender);
     userPosition.suppliedShares += suppliedShares.toUint128();
 
+    if (_positionStatus[onBehalfOf].isUsingAsCollateral(reserveId)) {
```
DhairyaSethi, line 208: we need to notify on enableAsCollateral or notify always on supply - still thinking about this

Kogaroshi: I'd say on `enableAsCollateral`, so we never notify for users never intending to draw debt (like tokenization vaults for exemple)

DhairyaSethi: yeah forgot i already added the notify on enable as coll. in any case notify would be short circruited for those users bc their risk premium will be zero 

and tokenalisation vaults are directly interacting with the hub, they don't connect with spokes

src/spoke/Spoke.sol
```diff
@@ -204,6 +204,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+      UserAccountData memory accountData = _calculateUserAccountData(onBehalfOf);
+      _notifyRiskPremiumUpdate(onBehalfOf, accountData.riskPremium);
```
miguelmtzinf: we can inline

DhairyaSethi: idt it matters tbh but yep

src/spoke/Spoke.sol
```diff
@@ -688,8 +695,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     bool refreshConfig
   ) internal returns (UserAccountData memory accountData) {
     PositionStatus storage positionStatus = _positionStatus[user];
-
     uint256 reserveId = _reserveCount;
+
+    if (positionStatus.isEmpty(reserveId)) {
+      accountData.healthFactor = type(uint256).max;
+      return accountData;
```
miguelmtzinf: we canot do this, because we are returning zero values in the rest of params

src/spoke/Spoke.sol
miguelmtzinf: we can move this at the end now

src/spoke/Spoke.sol
DhairyaSethi: that's the point, when position is empty (ie no borrowed or supplied) we return early bc all other values are zero

src/spoke/Spoke.sol
```diff
@@ -763,6 +775,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     }
 
     uint256 debtValueLeftToCover = accountData.totalDebtValue;
+    // riskPremium is 0 when position is underwater or has no debt
+    if (
+      accountData.healthFactor < HEALTH_FACTOR_LIQUIDATION_THRESHOLD || debtValueLeftToCover == 0
```
```suggestion
     debtValueLeftToCover == 0 ||  accountData.healthFactor < HEALTH_FACTOR_LIQUIDATION_THRESHOLD
```
i d favor the second leg, which is more likely

src/spoke/Spoke.sol
DhairyaSethi: saves 5 gas [5707d09](https://github.com/aave/aave-v4/pull/939/commits/5707d09ff669ff1ad1b2b886a6b678f95128812a)

src/spoke/Spoke.sol
miguelmtzinf: ah, you are right. Anyway, what's the use case you are thinking about this would be needed for?
I think it's a bit pointless to add logic that does not add value to the majority of cases, and just makes the flow more efficient for non-users of the system.

src/spoke/Spoke.sol
```diff
@@ -763,22 +775,24 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
     for (uint256 index = 0; index < collateralInfo.length(); ++index) {
-      if (debtValueLeftToCover == 0) {
-        break;
-      }
-
       (uint256 collateralRisk, uint256 userCollateralValue) = collateralInfo.get(index);
       userCollateralValue = userCollateralValue.min(debtValueLeftToCover);
       accountData.riskPremium += userCollateralValue * collateralRisk;
       debtValueLeftToCover = debtValueLeftToCover.uncheckedSub(userCollateralValue);
+      if (debtValueLeftToCover == 0) break;
```
```suggestion
      if (debtValueLeftToCover == 0) {
        break;
      }
```
sorry mate

src/spoke/Spoke.sol
```diff
@@ -305,8 +310,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       _positionStatus[onBehalfOf].setBorrowing(reserveId, false);
     }
 
-    UserAccountData memory userAccountData = _calculateUserAccountData(onBehalfOf);
-    _notifyRiskPremiumUpdate(onBehalfOf, userAccountData.riskPremium);
+    uint256 newRiskPremium = _calculateUserAccountData(onBehalfOf).riskPremium;
+    _notifyRiskPremiumUpdate(onBehalfOf, newRiskPremium);
```
yan-man, line 315: liquidationCall does `_notifyRiskPremiumUpdate(onBehalfOf, _calculateUserAccountData(onBehalfOf).riskPremium);` can you update here or there for consistency

src/spoke/Spoke.sol
```diff
@@ -688,8 +695,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     bool refreshConfig
   ) internal returns (UserAccountData memory accountData) {
     PositionStatus storage positionStatus = _positionStatus[user];
-
     uint256 reserveId = _reserveCount;
+
+    if (positionStatus.isEmpty(reserveId)) {
+      accountData.healthFactor = type(uint256).max;
```
yan-man: i think perhaps we could declare this at the top, as the "default" case is HF being max. And can then rm this line and the else branch from L766

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
```diff
@@ -1279,4 +1292,18 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
     _checkSpokeBalances(params, accountsInfoBefore, accountsInfoAfter, liquidationMetadata);
     _checkHubBalances(params, accountsInfoBefore, accountsInfoAfter, liquidationMetadata);
   }
+
+  function _hasPositiveRiskPremium(ISpoke spoke, address user) internal returns (bool) {
```
yan-man, line 1297: maybe can add some comments above here as it's a lot going on in this helper, just to read the `hasPositiveRiskPremium` var which is unreachable

src/spoke/Spoke.sol
DhairyaSethi: it's just to optimise the non enabled as collateral supplier to exit early and avoid the memory allocation for collateralInfo

src/spoke/libraries/PositionStatusMap.sol
```diff
@@ -81,6 +81,22 @@ library PositionStatusMap {
...
+  /// @notice Returns true if a user has no collateral & no borrowed reserves.
+  /// @dev Does not disregard potential dirty bits set after `reserveCount`.
+  function isEmpty(
+    ISpoke.PositionStatus storage self,
+    uint256 reserveCount
+  ) internal view returns (bool) {
+    unchecked {
+      uint256 bucket = reserveCount.bucketId();
+      bool empty = self.map[bucket] == 0;
+      while (!empty && bucket != 0) {
```
avniculae: shouldn't it be 
```
while (empty && bucket != 0) {
```

src/spoke/Spoke.sol
```diff
@@ -763,22 +776,26 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
     for (uint256 index = 0; index < collateralInfo.length(); ++index) {
-      if (debtValueLeftToCover == 0) {
-        break;
-      }
-
       (uint256 collateralRisk, uint256 userCollateralValue) = collateralInfo.get(index);
       userCollateralValue = userCollateralValue.min(debtValueLeftToCover);
       accountData.riskPremium += userCollateralValue * collateralRisk;
       debtValueLeftToCover = debtValueLeftToCover.uncheckedSub(userCollateralValue);
+      if (debtValueLeftToCover == 0) {
```
avniculae, line 793: fair point to move this here now 👍🏻

src/spoke/Spoke.sol
```diff
@@ -863,7 +880,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       userPosition.drawnShares -= deficitShares.toUint128();
       positionStatus.setBorrowing(reserveId, false);
     }
-    // non-zero deficit means user ends up with zero total debt
+
+    positionStatus.hasPositiveRiskPremium = false;
```
avniculae: good catch

tests/unit/Spoke/Spoke.RiskPremium.EdgeCases.t.sol
```diff
@@ -610,15 +611,16 @@ contract SpokeRiskPremiumEdgeCasesTest is SpokeBase {
...
 
-    // Now Bob's RP should be greater than collateral risk of dai, since debt is not fully covered by it
-    assertGt(
-      _getUserRiskPremium(spoke2, bob),
-      _getCollateralRisk(spoke2, _daiReserveId(spoke2)),
-      'Bob user risk premium after collateral accrual'
-    );
+    uint256 bobRiskPremium = _getUserRiskPremium(spoke2, bob);
+    if (_isHealthy(spoke2, bob)) {
+      // Now Bob's RP should be greater than collateral risk of dai, since debt is not fully covered by it
```
CheyenneAtapour, line 616: Can you make an assert that the debt is not fully covered by the total value of the collateral in the `else` case? For some reason, when I was messing around, I was getting HF not healthy, BUT the debt was not greater than total value of collaterals. Just would be a nice sanity check, because something seemed fishy while I was investigating there

tests/unit/Spoke/Spoke.RiskPremium.t.sol
```diff
@@ -179,20 +179,9 @@ contract SpokeRiskPremiumTest is SpokeBase {
...
-      _getCollateralRisk(spoke2, _wbtcReserveId(spoke2)) +
-      _getValue(spoke2, _wethReserveId(spoke2), wethSupplyAmount) *
-      _getCollateralRisk(spoke2, _wethReserveId(spoke2))) / collateralValue;
-    assertEq(
-      _getUserRiskPremium(spoke2, bob),
-      expectedRiskPremium,
-      'user risk premium matches weighted sum of collaterals'
-    );
+    assertFalse(_isHealthy(spoke2, bob));
+    // risk premium should be zero when position is underwater, the above proves with worst case of 100% cf
```
CheyenneAtapour, line 183: what do you mean 100% cf here? The collaterals are not 100% cf

CheyenneAtapour, line 183: what do you mean 100% cf here? The collaterals are not 100% cf

DhairyaSethi: worst case; because risk premium is collValue * CR / debtValue and hf is collValue * CF / debtValue; consider the CF margin which means for 100% CR hf is always less than risk premium (if all coll is used to cover debt)

src/spoke/libraries/PositionStatusMap.sol
DhairyaSethi: removed bc not super useful anyway

src/spoke/Spoke.sol
DhairyaSethi: removed bc it's not super useful anyway

src/spoke/Spoke.sol
DhairyaSethi: removed this check entirely

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -274,7 +274,7 @@ library LiquidationLogic {
   }
 
   /// @dev Invoked by `liquidateUser` method.
-  /// @return True if the collateral position is empty, false otherwise.
+  /// @return True if the user collateral position becomes empty after removing.
   function _liquidateCollateral(
```
DhairyaSethi, line 279: todo: when `receiveShares` is true, enforce that `liquidator` has not enabled reserve as collateral

DhairyaSethi: #966

snapshots/SignatureGateway.Operations.json
```diff
@@ -1,9 +1,9 @@
 {
-  "borrowWithSig": "211725",
-  "repayWithSig": "237828",
+  "borrowWithSig": "211704",
+  "repayWithSig": "237802",
   "setSelfAsUserPositionManagerWithSig": "75492",
-  "setUsingAsCollateralWithSig": "85259",
-  "supplyWithSig": "147957",
+  "setUsingAsCollateralWithSig": "123522",
```
avniculae: why is it a 40k increase here?

snapshots/Spoke.Operations.ZeroRiskPremium.json
```diff
@@ -1,34 +1,34 @@
...
-  "repay: partial": "173691",
+  "borrow: first": "196380",
+  "borrow: second action, same reserve": "176618",
+  "liquidationCall (receiveShares): full": "261175",
+  "liquidationCall (receiveShares): partial": "300558",
+  "liquidationCall: full": "282699",
+  "liquidationCall: partial": "304416",
+  "permitReserve + repay (multicall)": "231949",
+  "permitReserve + supply (multicall)": "168376",
+  "permitReserve + supply + enable collateral (multicall)": "253068",
```
avniculae: this one also has a big increase

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -358,12 +362,20 @@ library LiquidationLogic {
       ISpoke.HealthFactorNotBelowThreshold()
     );
     require(
-      params.isUsingAsCollateral && params.collateralFactor > 0,
+      params.collateralFactor > 0 &&
+        positionStatus[params.user].isUsingAsCollateral(params.collateralReserveId),
       ISpoke.CollateralCannotBeLiquidated()
     );
     require(params.collateralReserveBalance > 0, ISpoke.ReserveNotSupplied());
     require(params.debtReserveBalance > 0, ISpoke.ReserveNotBorrowed());
```
miguelmtzinf: wondering if these 2 should be moved up, right before hf check

snapshots/SignatureGateway.Operations.json
DhairyaSethi: with we have to notify on enabling as collateral as well now

snapshots/Spoke.Operations.ZeroRiskPremium.json
DhairyaSethi: same bc notify on enable as coll

src/spoke/Spoke.sol
```diff
@@ -383,12 +389,15 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
     positionStatus.setUsingAsCollateral(reserveId, usingAsCollateral);
 
+    uint256 newRiskPremium;
     if (usingAsCollateral) {
       _refreshDynamicConfig(onBehalfOf, reserveId);
+      newRiskPremium = _calculateUserAccountData(onBehalfOf).riskPremium;
     } else {
-      uint256 newRiskPremium = _refreshAndValidateUserAccountData(onBehalfOf).riskPremium;
+      newRiskPremium = _refreshAndValidateUserAccountData(onBehalfOf).riskPremium;
       _notifyRiskPremiumUpdate(onBehalfOf, newRiskPremium);
```
miguelmtzinf: bad merge @DhairyaSethi

src/spoke/Spoke.sol
```diff
@@ -766,24 +775,28 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
     for (uint256 index = 0; index < collateralInfo.length(); ++index) {
-      if (debtValueLeftToCover == 0) {
-        break;
-      }
-
       (uint256 collateralRisk, uint256 userCollateralValue) = collateralInfo.get(index);
       userCollateralValue = userCollateralValue.min(debtValueLeftToCover);
       accountData.riskPremium += userCollateralValue * collateralRisk;
       debtValueLeftToCover = debtValueLeftToCover.uncheckedSub(userCollateralValue);
+      if (debtValueLeftToCover == 0) {
```
miguelmtzinf, line 793:
```suggestion

      if (debtValueLeftToCover == 0) {
```
just a linebreak

src/spoke/Spoke.sol
DhairyaSethi: damn tysm

snapshots/SignatureGateway.Operations.json
avniculae: there are 2 reserves there?

src/spoke/Spoke.sol
DhairyaSethi: intentionally removed it looked worse haha but will add

src/spoke/Spoke.sol
avniculae, line 388: I think we should leave a comment here to mention that this check is now very important, it's not just an opt. This is bc if `usingAsCollateral` is true, we refresh dynamic config without checking HF

snapshots/SignatureGateway.Operations.json
DhairyaSethi: nope just the one; it's mainly from the new calcUserAccData loop

src/spoke/Spoke.sol
miguelmtzinf: i d place it before the loop so it's not missed when reading

src/spoke/Spoke.sol
```diff
@@ -766,24 +774,28 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     }
 
     uint256 debtValueLeftToCover = accountData.totalDebtValue;
+    if (
+      debtValueLeftToCover == 0 || accountData.healthFactor < HEALTH_FACTOR_LIQUIDATION_THRESHOLD
+    ) {
+      // riskPremium is 0 when position is underwater or has no debt
```
avniculae:
```suggestion
      // riskPremium is 0 when user has no debt or is unhealthy
```

src/spoke/Spoke.sol
avniculae, line 298: signedSub

src/spoke/Spoke.sol
miguelmtzinf: what about adding a dev comment
`/// @dev It clears the user position, setting drawn, premium, and risk premium to zero.`

snapshots/Spoke.Operations.json
```diff
@@ -1,34 +1,34 @@
...
+  "supply: second action, same reserve": "129161",
+  "updateUserDynamicConfig: 1 collateral": "73487",
+  "updateUserDynamicConfig: 2 collaterals": "87983",
+  "updateUserRiskPremium: 1 borrow": "173093",
+  "updateUserRiskPremium: 2 borrows": "253679",
+  "usingAsCollateral: 0 borrows, enable": "78231",
+  "usingAsCollateral: 1 borrow, disable": "153416",
   "usingAsCollateral: 1 borrow, enable": "32364",
-  "usingAsCollateral: 2 borrows, disable": "279528",
-  "usingAsCollateral: 2 borrows, enable": "41942",
```
avniculae, line 28: this is weird no?

src/spoke/Spoke.sol
DhairyaSethi: i think it's fine here at the end with the event emission, thats the order of actions of notify (wrt event, the storage write is at the top & we we don't expect reentry)
im considering the gas of 

```suggestion
    if (positionStatus.hasPositiveRiskPremium) {
      positionStatus.hasPositiveRiskPremium = false;
      emit UpdateUserRiskPremium(user, 0);
    }
```

src/spoke/Spoke.sol
DhairyaSethi: it does unchecked `toInt256()`, let's do this in a separate PR; apparently it does ~100 gas

snapshots/Spoke.Operations.json
DhairyaSethi: which part

src/spoke/Spoke.sol
DhairyaSethi: i think it's fine w/o but yep lmk wdyt [9e37adb](https://github.com/aave/aave-v4/pull/939/commits/9e37adbff843d9a53f87279a60d64f54ec685171)

src/spoke/Spoke.sol
```diff
@@ -379,16 +385,18 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
 
     if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) {
-      return;
+      return; // no op also ensures only new collateral is refreshed once
```
```suggestion
// no op if unchanged; ensures only new collateral is refreshed once, avoiding unnecessary user risk premium updates
if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) {
      return;
```

DhairyaSethi: the main point is not avoid updates but refresh once is done w/o health factor check which is only ok if it's a NEW collateral

miguelmtzinf:
```suggestion
// no op if unchanged; ensures only new collateral is refreshed once so full dynamic config refresh cannot be skipped
if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) {
      return;
```

DhairyaSethi: > avoiding unnecessary user risk premium updates

is totally fine, we allow unnecessary rp updates through `updateUserRiskPremium`; when underwater rp becomes zero so this is ok

snapshots/Spoke.Operations.json
avniculae: only 41k?

src/spoke/Spoke.sol
avniculae: to me this is not clear enough, would like us to mention that this cannot be removed (otherwise it will be v bad)

PR #940, opened 2025-10-21
feat: skip risk premium refresh when user is underwater

avniculae:


src/spoke/Spoke.sol
```diff
@@ -307,7 +307,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     }
 
     UserAccountData memory userAccountData = _calculateUserAccountData(onBehalfOf);
-    _notifyRiskPremiumUpdate(onBehalfOf, userAccountData.riskPremium);
+    if (userAccountData.healthFactor >= HEALTH_FACTOR_LIQUIDATION_THRESHOLD) {
```
avniculae, line 310: if we don't notify bc of this condition, then user will have 0 rp for the repaid reserve, while non-zero for the other reserves (it will make the rp out of sync)

src/spoke/Spoke.sol
```diff
@@ -364,7 +366,10 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       _reportDeficit(user);
     } else {
       // new risk premium only needs to be propagated if no deficit exists
-      _notifyRiskPremiumUpdate(user, _calculateUserAccountData(user).riskPremium);
+      userAccountData = _calculateUserAccountData(user);
+      if (userAccountData.healthFactor >= HEALTH_FACTOR_LIQUIDATION_THRESHOLD) {
```
avniculae, line 370: if we don't refresh rp here, user can supply after liquidation to avoid further liquidations and keep a better rp

src/spoke/Spoke.sol
miguelmtzinf: we could set it to 0 instead.

avniculae: at which point it's better to go with #939 right?

avniculae:
Closing in favour of #939

PR #949, opened 2025-10-22
feat: emit asset amount in spoke events

yan-man:
- emit asset amount in spoke user action events (supply, borrow, etc.)
- this provides more information to simplify offchain indexing


tests/unit/misc/NativeTokenGateway.t.sol
```diff
@@ -483,6 +499,11 @@ contract NativeTokenGatewayTest is SpokeBase {
     _checkFinalBalances();
   }
 
+  // struct PrevBalance {
+  //   uint256 userBalance;
+  //   uint256 hubBalance;
+  // }
+
```
DhairyaSethi:
```suggestion
```

PR #950, opened 2025-10-23
feat: Treasury spoke compatible with multiple hubs

CheyenneAtapour:
This implementation removes the inherited interface `ISpoke`, because it changes the interface for the TreasurySpoke functions to include a hub. 

The alternative to this approach would be to keep the interface, but keep a mapping in storage to determine the correct hub to call, similar to the current Spoke implementation.

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -336,8 +336,7 @@ contract Hub is IHub, AccessManaged {
     spoke.drawnShares -= drawnShares;
     _applyPremiumDelta(asset, spoke, premiumDelta);
 
-    uint256 deficitAmountRay = uint256(drawnShares) *
-      asset.drawnIndex +
+    uint256 deficitAmountRay = uint256(drawnShares) * asset.drawnIndex +
       premiumDelta.restoredPremiumRay;
     asset.deficitRay += deficitAmountRay.toUint200();
     spoke.deficitRay += deficitAmountRay.toUint200();
@@ -842,7 +841,7 @@ contract Hub is IHub, AccessManaged {
     require(
       addCap == MAX_ALLOWED_SPOKE_CAP ||
         addCap * MathUtils.uncheckedExp(10, asset.decimals) >=
-        asset.toAddedAssetsUp(spoke.addedShares) + amount,
+          asset.toAddedAssetsUp(spoke.addedShares) + amount,
       AddCapExceeded(addCap)
     );
   }
@@ -871,7 +870,7 @@ contract Hub is IHub, AccessManaged {
     require(
       drawCap == MAX_ALLOWED_SPOKE_CAP ||
         drawCap * MathUtils.uncheckedExp(10, asset.decimals) >=
-        owed + amount + uint256(spoke.deficitRay).fromRayUp(),
+          owed + amount + uint256(spoke.deficitRay).fromRayUp(),
       DrawCapExceeded(drawCap)
     );
   }
@@ -931,7 +930,7 @@ contract Hub is IHub, AccessManaged {
     require(
       addCap == MAX_ALLOWED_SPOKE_CAP ||
         addCap * MathUtils.uncheckedExp(10, asset.decimals) >=
-        asset.toAddedAssetsUp(receiver.addedShares + shares),
+          asset.toAddedAssetsUp(receiver.addedShares + shares),
       AddCapExceeded(addCap)
     );
   }


src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -3,12 +3,11 @@
 pragma solidity ^0.8.0;
 
 import {IHubBase} from 'src/hub/interfaces/IHubBase.sol';
```
Kogaroshi: not needed anymore ?

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -17,28 +16,32 @@ interface ITreasurySpoke is ISpokeBase {
 
   /// @notice Supplies a specified amount of the underlying asset to a given reserve.
   /// @dev The Spoke pulls the underlying asset from the caller, so prior approval is required.
-  /// @dev The reserve identifier must match the asset identifier in the Hub.
-  /// @param reserveId The identifier of the reserve.
+  /// @dev The given asset identifier must match the asset identifier in the given hub.
```
Kogaroshi: We use "Hub" in the natspec of `ISpoke`.

tests/Base.t.sol
```diff
@@ -2337,7 +2337,7 @@ abstract contract Base is Test {
       return; // nothing to withdraw
     }
     vm.prank(TREASURY_ADMIN);
-    treasurySpoke.withdraw(assetId, amount, address(treasurySpoke));
+    treasurySpoke.withdraw(address(hub1), assetId, amount, address(treasurySpoke));
```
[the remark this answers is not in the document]
Kogaroshi: correct comment

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -13,18 +13,14 @@ contract TreasurySpokeTest is SpokeBase {
   }
```
Kogaroshi, line 49: let's add tests with multiple Hubs, for example with same asset in multiple Hubs

CheyenneAtapour: Done

src/spoke/interfaces/ITreasurySpoke.sol
CheyenneAtapour: Done

src/spoke/interfaces/ITreasurySpoke.sol
CheyenneAtapour: Removed

tests/Base.t.sol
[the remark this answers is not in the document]

src/spoke/TreasurySpoke.sol
```diff
@@ -6,54 +6,49 @@ import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.
...
     uint256 amount,
     address
   ) external onlyOwner returns (uint256, uint256) {
-    (address underlying, ) = HUB.getAssetUnderlyingAndDecimals(reserveId);
-    IERC20(underlying).safeTransferFrom(msg.sender, address(HUB), amount);
-    uint256 shares = HUB.add(reserveId, amount);
+    IHub hubContract = IHub(hub);
+    address underlying = hubContract.getAsset(assetId).underlying;
+    IERC20(underlying).safeTransferFrom(msg.sender, hub, amount);
+    uint256 shares = hubContract.add(assetId, amount);
```
Kogaroshi: we might want to add a non zero check on shares, so we don't just donate treasury funds there ?

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -2,13 +2,10 @@
...
-import {IHubBase} from 'src/hub/interfaces/IHubBase.sol';
-import {ISpokeBase} from 'src/spoke/interfaces/ISpokeBase.sol';
-
 /// @title ITreasurySpoke
 /// @author Aave Labs
 /// @notice Interface for the TreasurySpoke.
-interface ITreasurySpoke is ISpokeBase {
+interface ITreasurySpoke {
   /// @notice Thrown when an unsupported action is attempted.
   error UnsupportedAction();
```
Kogaroshi: still needed ?

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -17,28 +14,32 @@ interface ITreasurySpoke is ISpokeBase {
 
   /// @notice Supplies a specified amount of the underlying asset to a given reserve.
   /// @dev The Spoke pulls the underlying asset from the caller, so prior approval is required.
-  /// @dev The reserve identifier must match the asset identifier in the Hub.
-  /// @param reserveId The identifier of the reserve.
+  /// @dev The given asset identifier must match the asset identifier in the given Hub.
```
Kogaroshi: is this comment still needed ? now we can consider the hub + assetId is an informed input, there is no need to match reserveId with an assetId there.

src/spoke/TreasurySpoke.sol
CheyenneAtapour: There already is a nonzero check for this on the hub, right?

src/spoke/interfaces/ITreasurySpoke.sol
CheyenneAtapour: Removed, good catch

src/spoke/interfaces/ITreasurySpoke.sol
CheyenneAtapour: Removed

src/spoke/TreasurySpoke.sol
Kogaroshi: true, mb there

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -2,43 +2,39 @@
...
-
+interface ITreasurySpoke {
   /// @notice Thrown when the given address is invalid.
   error InvalidAddress();
 
   /// @notice Supplies a specified amount of the underlying asset to a given reserve.
   /// @dev The Spoke pulls the underlying asset from the caller, so prior approval is required.
-  /// @dev The reserve identifier must match the asset identifier in the Hub.
-  /// @param reserveId The identifier of the reserve.
+  /// @param hub The address of the Hub contract.
```
yan-man:
```suggestion
  /// @param hub The address of the Hub.
```

src/spoke/TreasurySpoke.sol
```diff
@@ -64,58 +59,12 @@ contract TreasurySpoke is ITreasurySpoke, Ownable2Step {
   }
 
   /// @inheritdoc ITreasurySpoke
-  function getSuppliedAmount(uint256 reserveId) external view returns (uint256) {
-    return HUB.getSpokeAddedAssets(reserveId, address(this));
+  function getSuppliedAmount(address hub, uint256 assetId) external view returns (uint256) {
```
yan-man: feel like this should be `getSpokeSuppliedAssets` and the other `getSpokeSuppliedShares` for clarity, given that it's not for this given owner only (ie if the owner changes, would still return from all this spoke's supply)

src/spoke/TreasurySpoke.sol
```diff
@@ -64,58 +59,12 @@ contract TreasurySpoke is ITreasurySpoke, Ownable2Step {
   }
```
yan-man, line 57: would it be too complicated if we had another supply/withdraw method here, that allowed for a spoke input, so that this spoke could directly supply to the tokenization spoke? or is spoke -> spoke supply too much inception

src/spoke/TreasurySpoke.sol
```diff
@@ -6,54 +6,49 @@ import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.
...
-  /// @param hub_ The address of the Hub.
-  constructor(address owner_, address hub_) Ownable(owner_) {
-    require(hub_ != address(0), InvalidAddress());
-
-    HUB = IHubBase(hub_);
-  }
+  constructor(address owner_) Ownable(owner_) {}
 
   /// @inheritdoc ITreasurySpoke
   function supply(
```
yan-man, line 23: feel like we were initially constrained here bc it was trying to follow ISpokeBase (before), but rather I kind of think this should be `add` and `remove` instead of supply/withdraw, as it's doing it for the spoke -> hub rather than user -> spoke. ie if owner changes, they dont keep their supplied amount, it's owned by the spoke actually and withdraw just goes to whoever is owner at that moment. So it's like owner acts on behalf of the spoke

yan-man: and given that we're straying from the ISpokeBase, either way dont think we need that last address blank input, we can do whatever

src/spoke/TreasurySpoke.sol
CheyenneAtapour: hmm, @DhairyaSethi is this helpful to have?

src/spoke/TreasurySpoke.sol
CheyenneAtapour: I feel like the naming is more aligned how it is, but open to hear what others think as well, cc @miguelmtzinf @DhairyaSethi @Kogaroshi

src/spoke/TreasurySpoke.sol
CheyenneAtapour: Done

src/spoke/interfaces/ITreasurySpoke.sol
CheyenneAtapour: Done

src/spoke/TreasurySpoke.sol
DhairyaSethi: idts, we can add stuff like this later when there's concrete plans since this contract should be upgradeable now

src/spoke/TreasurySpoke.sol
DhairyaSethi: yeah i get the idea. we should rm the last unused import for sure, given we've already deviated from ISpoke.supply signature 

re naming of the function, im agnostic. maybe its better to have different names from both (hub,spoke) to avoid confusion

src/spoke/TreasurySpoke.sol
```diff
@@ -6,54 +6,49 @@ import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.
...
-    uint256 reserveId,
+    address hub,
+    uint256 assetId,
     uint256 amount,
     address
   ) external onlyOwner returns (uint256, uint256) {
```
DhairyaSethi, line 27: this can also be hub, underlying, amount only

src/spoke/TreasurySpoke.sol
```diff
@@ -6,54 +6,49 @@ import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.
...
-    (address underlying, ) = HUB.getAssetUnderlyingAndDecimals(reserveId);
-    IERC20(underlying).safeTransferFrom(msg.sender, address(HUB), amount);
-    uint256 shares = HUB.add(reserveId, amount);
+    IHub hubContract = IHub(hub);
+    address underlying = hubContract.getAsset(assetId).underlying;
```
DhairyaSethi: you can use getUnderlyingAndDecimals such that this uses IHubBase only

src/spoke/TreasurySpoke.sol
DhairyaSethi, line 27: perhaps we should validate that hub is valid somehow - it's kinda scary it's arbitrary input. acceptable design since it's onlyOwner always but alas. 

in posms, we additional redundant spoke registration check for this reason

CheyenneAtapour: is underlying better than assetId?

src/spoke/TreasurySpoke.sol
CheyenneAtapour: Done

src/spoke/TreasurySpoke.sol
CheyenneAtapour: Call to getAssetUnderlyingAndDecimals will fail now, so should be good right? EDIT: Ok, that doesn't change anything actually. How do you propose we validate it?

src/spoke/TreasurySpoke.sol
Kogaroshi: I agree, new method names is better

src/spoke/TreasurySpoke.sol
Kogaroshi: Underlying might be easier for admin/payloads, no need to fetch `assetId` for the underlying and then the TreasurySpoke will fetch back that underlying, instead just pass underlying and let it fetch `assetId` + it ensures the methods here and the same as Spoke in terms of params types, so less confusion possible

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -2,43 +2,39 @@
...
 /// @title ITreasurySpoke
 /// @author Aave Labs
 /// @notice Interface for the TreasurySpoke.
-interface ITreasurySpoke is ISpokeBase {
-  /// @notice Thrown when an unsupported action is attempted.
-  error UnsupportedAction();
-
+interface ITreasurySpoke {
   /// @notice Thrown when the given address is invalid.
   error InvalidAddress();
```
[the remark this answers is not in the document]
miguelmtzinf: unused

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -2,43 +2,39 @@
...
   /// @param amount The amount of asset to supply.
   /// @param onBehalfOf Unused parameter for this spoke.
   /// @return The amount of shares supplied.
   /// @return The amount of assets supplied.
   function supply(
-    uint256 reserveId,
+    address hub,
+    uint256 assetId,
     uint256 amount,
     address onBehalfOf
```
miguelmtzinf: remove

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -2,43 +2,39 @@
...
   ) external returns (uint256, uint256);
 
   /// @notice Withdraws a specified amount of underlying asset from the given reserve.
   /// @dev Providing an amount greater than the maximum withdrawable value signals a full withdrawal.
-  /// @dev The reserve identifier must match the asset identifier in the Hub.
-  /// @param reserveId The identifier of the reserve.
+  /// @param hub The address of the Hub.
+  /// @param assetId The identifier of the asset.
   /// @param amount The amount of asset to withdraw.
   /// @param onBehalfOf Unused parameter for this spoke.
```
miguelmtzinf: remove

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -49,20 +45,16 @@ interface ITreasurySpoke is ISpokeBase {
...
 
-  /// @notice Returns the amount of assets supplied.
-  /// @dev The reserve identifier must match the asset identifier in the Hub.
-  /// @param reserveId The identifier of the reserve.
+  /// @notice Returns the amount of assets supplied by this spoke.
+  /// @param hub The address of the Hub.
+  /// @param assetId The identifier of the asset.
   /// @return The amount of assets supplied.
-  function getSuppliedAmount(uint256 reserveId) external view returns (uint256);
+  function getSpokeSuppliedAssets(address hub, uint256 assetId) external view returns (uint256);
```
miguelmtzinf: it may be enough `getSuppliedAssets` and `getSuppliedShares`

src/spoke/TreasurySpoke.sol
DhairyaSethi: exactly that's why i prefer underlying input

src/spoke/TreasurySpoke.sol
```diff
@@ -2,58 +2,51 @@
...
+import {ITreasurySpoke} from 'src/spoke/interfaces/ITreasurySpoke.sol';
 
 /// @title TreasurySpoke
 /// @author Aave Labs
 /// @notice Spoke contract used as a treasury where accumulated fees are treated as supplied assets.
 /// @dev Dedicated to a single user, controlled exclusively by the owner.
-/// @dev Utilizes all assets from the Hub without restrictions, making reserve and asset identifiers aligned.
-/// @dev Allows withdraw to claim fees and supply to invest back into the Hub via this dedicated spoke.
-contract TreasurySpoke is ITreasurySpoke, Ownable2Step {
+/// @dev Allows withdraw to claim fees and supply to invest back into the corresponding hub via this dedicated spoke.
```
miguelmtzinf:
```suggestion
/// @dev Allows withdraw to claim fees and supply to invest back into any Hub asset.
```

src/spoke/TreasurySpoke.sol
miguelmtzinf: the `getAssetId(underlying)` is a validation itself

src/spoke/TreasurySpoke.sol
```diff
@@ -2,58 +2,51 @@
...
     uint256 amount,
     address
   ) external onlyOwner returns (uint256, uint256) {
-    (address underlying, ) = HUB.getAssetUnderlyingAndDecimals(reserveId);
-    IERC20(underlying).safeTransferFrom(msg.sender, address(HUB), amount);
-    uint256 shares = HUB.add(reserveId, amount);
+    IHubBase hubContract = IHubBase(hub);
+    (address underlying, ) = hubContract.getAssetUnderlyingAndDecimals(assetId);
+    IERC20(underlying).safeTransferFrom(msg.sender, hub, amount);
+    uint256 shares = hubContract.add(assetId, amount);
```
miguelmtzinf: we return this directly

src/spoke/TreasurySpoke.sol
DhairyaSethi: it's not. that assumes hub is correct, i'm saying if admin is tricked or something into calling malicious hub since it is user input

DhairyaSethi: a validation would be hub registration like we have for spoke registration on the posms 

again, it's likely redundant 

and we don't need it here more so bc these are all admin functions. it's just doing a call on a user input address is very scary regardless

tests/unit/HubConfigurator.t.sol
```diff
@@ -367,15 +367,22 @@ contract HubConfiguratorTest is HubBase {
       365 days
     );
 
-    assertGe(treasurySpoke.getSuppliedShares(daiAssetId), 0);
+    assertGe(treasurySpoke.getSpokeSuppliedShares(address(hub1), daiAssetId), 0);
```
miguelmtzinf: i think it's probably better to decouple here and prank feeReceiver?

src/spoke/TreasurySpoke.sol
CheyenneAtapour: @Kogaroshi Do you mean new names from both hub and spoke, or add and remove?

src/spoke/TreasurySpoke.sol
CheyenneAtapour: We need to move getAssetId to IHubBase in this case, right?

src/spoke/interfaces/ITreasurySpoke.sol
CheyenneAtapour: @yan-man had suggested it like that: https://github.com/aave/aave-v4/pull/950#discussion_r2898241975
For now I can change to that, and we can discuss here

src/spoke/TreasurySpoke.sol
CheyenneAtapour: We don't want to return both amount and shares?

src/spoke/interfaces/ITreasurySpoke.sol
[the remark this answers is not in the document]
CheyenneAtapour: Removed

src/spoke/TreasurySpoke.sol
CheyenneAtapour: I moved getAssetId to IHubBase and did this

src/spoke/interfaces/ITreasurySpoke.sol
CheyenneAtapour: Done

src/spoke/interfaces/ITreasurySpoke.sol
CheyenneAtapour: Done

src/spoke/TreasurySpoke.sol
CheyenneAtapour: Done

src/spoke/TreasurySpoke.sol
CheyenneAtapour: Done tho

tests/unit/HubConfigurator.t.sol
CheyenneAtapour: Done: decoupled this test file from treasury spoke's logic

src/spoke/TreasurySpoke.sol
miguelmtzinf: ah sorry, got confused and thought hub was returning both

src/hub/Hub.sol
```diff
@@ -465,7 +465,7 @@ contract Hub is IHub, AccessManaged {
     return _assets[_underlyingToAssetId[underlying]].underlying == underlying;
   }
 
-  /// @inheritdoc IHub
+  /// @inheritdoc IHubBase
```
miguelmtzinf: move fn right after  `getAssetCount`

tests/unit/Spoke/Spoke.MultipleHub.Base.t.sol
```diff
@@ -39,7 +39,16 @@ contract SpokeMultipleHubBase is SpokeBase {
     // Canonical hub and spoke
     hub1 = DeployUtils.deployHub(address(accessManager), hex'01');
     (spoke1, oracle1) = _deploySpokeWithOracle(ADMIN, address(accessManager));
-    treasurySpoke = new TreasurySpoke(ADMIN, address(hub1));
+    TreasurySpokeInstance treasurySpokeImpl = new TreasurySpokeInstance();
+    treasurySpoke = ITreasurySpoke(
+      address(
+        new TransparentUpgradeableProxy(
```
DhairyaSethi: we use Utils.proxify but nbd

src/spoke/TreasurySpoke.sol
```diff
@@ -2,60 +2,51 @@
...
-    (address underlying, ) = HUB.getAssetUnderlyingAndDecimals(reserveId);
-    IERC20(underlying).safeTransferFrom(msg.sender, address(HUB), amount);
-    uint256 shares = HUB.add(reserveId, amount);
-
-    return (shares, amount);
+    address hub,
+    address underlying,
+    uint256 amount
+  ) external onlyOwner returns (uint256) {
+    IHubBase hubContract = IHubBase(hub);
```
miguelmtzinf:
```suggestion
    IHubBase targetHub = IHubBase(hub);
```
let's use `targetHub` instead, across fns

src/spoke/TreasurySpoke.sol
miguelmtzinf: let's revert and return both please

src/spoke/TreasurySpoke.sol
miguelmtzinf: don't think the function would be understood as dependent on the owner, given that we are not passing the address of the owner and what not.

src/spoke/TreasurySpoke.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.MultipleHub.Base.t.sol
CheyenneAtapour: Done

src/hub/Hub.sol

src/spoke/TreasurySpoke.sol
CheyenneAtapour: Done

src/spoke/TreasurySpoke.sol
miguelmtzinf: true, we assume these calls are safe given that are permissioned.

PR #951, opened 2025-10-23
rft: calculateUserAccountData option 1

miguelmtzinf:
Refactors all execution flows that require calculation of the user position, consolidating them into a single function with explicit input parameters that clearly define the requirements of each call.

Alternate approaches:
- option 2: #956 
- option 3: #960 
- option 4: #962


src/spoke/Spoke.sol
```diff
@@ -921,9 +954,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
-  function _calculateUserAccountData(address user) internal view returns (UserAccountData memory) {
-    // SAFETY: function does not modify state when refreshConfig is false.
-    return _castToView(_calculateAndPotentiallyRefreshUserAccountData)(user, false);
+  /// @notice View version of `_calculateUserAccountData`.
+  /// @dev It does not refresh dynamic config before calculating.
+  function _viewCalculateUserAccountData(
+    address user
+  ) internal view returns (UserAccountData memory) {
+    return _castToView(_calculateUserAccountData)(user, false);
```
miguelmtzinf, line 962: I d like to merge these 2 functions into 1.

avniculae: how?

src/spoke/Spoke.sol
```diff
@@ -405,7 +434,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     if (!_isPositionManager({user: onBehalfOf, manager: msg.sender})) {
       _checkCanCall(msg.sender, msg.data);
     }
-    uint256 newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
+    uint256 newRiskPremium = _validateUserPosition({
```
avniculae, line 437: should we merge _validateUserPosition when refresh and validate are true with notify?

src/spoke/Spoke.sol
```diff
@@ -305,7 +313,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       _positionStatus[onBehalfOf].setBorrowing(reserveId, false);
     }
 
-    UserAccountData memory userAccountData = _calculateUserAccountData(onBehalfOf);
+    UserAccountData memory userAccountData = _validateUserPosition({
```
avniculae, line 316: I think `validate` name for this fn is confusing since we pass a `validate` flag that can be 0. Should we call it something else? eg `_calculateUserAccountData`? (and name the current `_calculateUserAccountData` something else?

miguelmtzinf: yeah, open to suggestions.
- I like `validate` because it's kinda similar to all other validations we have for each operation, but this time for the user position.
- I think the current `_calculateUserAccountData` is fine.
- we can rename `validateHealthFactor` flag to `enforceHealth` or `enforceHealthFactor`

src/spoke/Spoke.sol
miguelmtzinf: Had that thought too. I think it's too much and code becomes a bit opaque.

src/spoke/Spoke.sol
miguelmtzinf: this `_viewCalculateUserAccountData` would need to static call (via low-level call or assembly) `_calculateUserAccountData` directly, with no intermediate fn

src/spoke/Spoke.sol
DhairyaSethi: agree with the above, 
 like enforceHealth for the flag

src/spoke/Spoke.sol
```diff
@@ -257,7 +261,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       positionStatus.setBorrowing(reserveId, true);
     }
 
-    uint256 newRiskPremium = _refreshAndValidateUserPosition(onBehalfOf);
+    uint256 newRiskPremium = _validateUserPosition({
+      user: onBehalfOf,
+      refreshConfig: true,
+      validateHealthFactor: true
```
DhairyaSethi, line 267: i think this is now more error prone, we ALWAYS want to validate health when refreshing config

miguelmtzinf: we could merge flags, under the `refreshConfigAndEnforceHealth`?

DhairyaSethi: thats kinda what we had before with 'refreshAndValidateUserPosition'

DhairyaSethi: if we only have one flag like before then this change is just renaming functions no

src/spoke/Spoke.sol
```diff
@@ -624,7 +657,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
   /// @inheritdoc ISpoke
   function getUserAccountData(address user) external view returns (UserAccountData memory) {
-    return _calculateUserAccountData(user);
+    return _viewCalculateUserAccountData(user);
```
miguelmtzinf, line 660: I think we should create a `getUserAccountDataWithLatestDynamicConfigKey`
```
  function _viewCalculateUserAccountData(
    address user,
    bool refreshConfig
  ) internal view returns (UserAccountData memory) {
    return _castToView(_calculateUserAccountData)(user, refreshConfig);
  }
```

src/spoke/Spoke.sol
miguelmtzinf: before:
`_calculateAndRefreshUserAccountData`, `_calculateAndPotentiallyRefreshUserAccountData`, `_refreshAndValidateUserPosition`, `_calculateUserAccountData`,
now:
- `_validateUserPosition`: general one, with explicity instructions via named input params
- `_calculateUserAccountData`: just for the calc
- `_viewCalculateUserAccountData`: only view, just for external use.

which is way clearer imho

src/spoke/Spoke.sol
DhairyaSethi: it's not safe & error prone, bc the function `_calculateUserAccountData` changes state when `refreshConfig` is true; hence it needs to be very carefully used 

an idea i have is we expose this information through revert data instead of return data; in that case it's definite there were no state changes

src/spoke/Spoke.sol
miguelmtzinf: anyway, i think keeping both flags separate is better readability (even if they are usually called together)

src/spoke/Spoke.sol
DhairyaSethi:
```solidity
  function getUserAccountDataWithRefreshConfig(
    address user
  ) external view returns (UserAccountData memory) {
    (, bytes memory ret) = address(this).staticcall(
      abi.encodeCall(this._calculateUserAccountDataWithRefresh, (user))
    );
    return abi.decode(ret, (UserAccountData));
  }

  function _calculateUserAccountDataWithRefresh(address user) external {
    assert(msg.sender == address(this));
    UserAccountData memory accountData = _calculateAndPotentiallyRefreshUserAccountData(user, true);
    assembly ('memory-safe') {
      revert(add(accountData, 20), mload(accountData))
    }
  }
```

src/spoke/Spoke.sol
avniculae: solidity will not allow this no?

src/spoke/Spoke.sol
miguelmtzinf: i d rather create a separate contract (most probably with dup code) than adding code that is hard to reason about

src/spoke/Spoke.sol
yan-man: > no intermediate fn

ie goal to rm `_castToView` completely yes?

src/spoke/Spoke.sol
miguelmtzinf: not sure i like the term `enforce` because it could mean that we applying any change to make it healthy

yan-man: agree w Alex that validate is a bit confusing here, as other validation methods don't write to storage but this would if refreshConfig is true

src/spoke/Spoke.sol
DhairyaSethi, line 660: let's inline the view fn's `_castToView` logic and i would bring back the SAFETY comment i liked it

miguelmtzinf:
Closing in favor of #962

PR #953, opened 2025-10-23
feat: reorder and tighter packing

DhairyaSethi:
- reduce `amounts` fields in storage from 128 to 120 bits; max possible value goes from ~1e38 to ~1e36
- reduce `index` field in storage from 128 to 120 bits; max index is ~340282366921 RAY
- increase `timestamp` field in storage from 32 to 40 bits

reorder hub vars to pack more efficiently, optimizing gas consumption


snapshots/Hub.Operations.json
```diff
@@ -1,15 +1,15 @@
...
-  "mintFeeShares": "84828",
-  "payFee": "71140",
-  "refreshPremium": "71913",
-  "remove: full": "73767",
-  "remove: partial": "78496",
-  "reportDeficit": "109874",
-  "restore: full": "91042",
-  "restore: partial": "104456",
-  "transferShares": "69941"
+  "add": "118313",
```
yan-man, line 2: interesting, why does add incr

DhairyaSethi: bc added shares is moved to an independent slot by replacing it with fee amount 

this causes supply, withdraw repay to go up by 3k but everything else down

src/spoke/interfaces/ISpoke.sol
```diff
@@ -71,13 +71,13 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @dev suppliedShares The supplied shares of the user position.
   /// @dev dynamicConfigKey The key of the user position dynamic config.
   struct UserPosition {
-    uint128 drawnShares;
-    uint128 realizedPremium;
+    uint120 drawnShares;
```
avniculae, line 74: what about targetHealthFactor? should we change it to uint120 as well?

DhairyaSethi: intentionally kept it 128 but doesn't really matter no?

PR #955, opened 2025-10-23
feat: update liquidity transfer flow and prevent asset re-listing in Hub

### LINKED ISSUE #764: Spoke caps can be bypassed by compromised spokes

yan-man:
- update the transfer flow between users and the hub, by enforcing that Spokes transferFrom funds from user to hub 
- requires users to approve specific spokes rather than approving the hub prior to a user action 
- as a result, assets cannot be relisted on the hub (due to logic around sufficient liquidity checks on hub)
- addresses a potential issue where a compromised spoke can control an arbitrary amount of funds from users who approved their tokens to the same hub, even if the user intended to interact with a different spoke.

closes #764

### DIFF
--- a/tests/gas/Hub.Operations.gas.t.sol
+++ b/tests/gas/Hub.Operations.gas.t.sol
@@ -18,6 +18,11 @@ contract HubOperations_Gas_Tests is Base {
     tokenList.usdx.transferFrom(alice, address(hub1), 1000e6);
     hub1.add(usdxAssetId, 1000e6);
     vm.snapshotGasLastCall('Hub.Operations', 'add');
+
+    vm.startSnapshotGas('Hub.Operations', 'add: with transfer');
+    tokenList.usdx.transferFrom(alice, address(hub1), 1000e6);
+    hub1.add(usdxAssetId, 1000e6);
+    vm.stopSnapshotGas();
     vm.stopPrank();
   }
 
@@ -86,6 +91,47 @@ contract HubOperations_Gas_Tests is Base {
     );
     hub1.restore(daiAssetId, drawnRemaining, premiumRemaining, premiumDelta);
     vm.snapshotGasLastCall('Hub.Operations', 'restore: full');
+    vm.stopPrank();
+  }
+
+  function test_restore_with_transfer() public {
+    uint256 drawnRemaining;
+    uint256 premiumRemaining;
+    vm.startPrank(address(spoke2));
+    tokenList.dai.transferFrom(alice, address(hub1), 1000e18);
+    hub1.add(daiAssetId, 1000e18);
+    vm.stopPrank();
+
+    vm.startPrank(address(spoke1));
+    tokenList.usdx.transferFrom(alice, address(hub1), 1000e6);
+    hub1.add(usdxAssetId, 1000e6);
+    hub1.draw(daiAssetId, 500e18, alice);
+    int256 premiumShares = hub1.previewDrawByAssets(daiAssetId, 500e18).toInt256();
+    int256 premiumOffset = hub1
+      .previewRestoreByShares(daiAssetId, uint256(premiumShares))
+      .toInt256();
+    hub1.refreshPremium(daiAssetId, IHubBase.PremiumDelta(premiumShares, premiumOffset, 0));
+
+    skip(1000);
+
+    (drawnRemaining, premiumRemaining) = hub1.getSpokeOwed(daiAssetId, address(spoke1));
+    vm.startSnapshotGas('Hub.Operations', 'restore: partial - with transfer');
+    tokenList.dai.transferFrom(alice, address(hub1), drawnRemaining / 2);
+    hub1.restore(daiAssetId, drawnRemaining / 2, 0, IHubBase.PremiumDelta(0, 0, 0));
+    vm.stopSnapshotGas();
+
+    skip(100);
+
+    (drawnRemaining, premiumRemaining) = hub1.getSpokeOwed(daiAssetId, address(spoke1));
+    IHubBase.PremiumDelta memory premiumDelta = IHubBase.PremiumDelta(
+      -premiumShares,
+      -premiumOffset,
+      0
+    );
+    vm.startSnapshotGas('Hub.Operations', 'restore: full - with transfer');
+    tokenList.dai.transferFrom(alice, address(hub1), drawnRemaining + premiumRemaining);
+    hub1.restore(daiAssetId, drawnRemaining, premiumRemaining, premiumDelta);
+    vm.stopSnapshotGas();
     vm.stopPrank();
   }
 
--- a/snapshots/Hub.Operations.json
+++ b/snapshots/Hub.Operations.json
@@ -1,5 +1,6 @@
 {
   "add": "87764",
+  "add: with transfer": "109371",
   "draw": "120584",
   "eliminateDeficit: full": "59756",
   "eliminateDeficit: partial": "69332",
@@ -10,6 +11,8 @@
   "remove: partial": "81308",
   "reportDeficit": "112658",
   "restore: full": "77662",
+  "restore: full - with transfer": "145600",
   "restore: partial": "86292",
+  "restore: partial - with transfer": "135925",
   "transferShares": "67922"
 }


src/hub/Hub.sol
```diff
@@ -622,6 +627,10 @@ contract Hub is IHub, AccessManaged {
     return _spokes[assetId][spoke];
   }
 
+  function isAssetListed(address underlying) external view returns (bool) {
```
Kogaroshi: maybe `isUnderlyingListed()` makes more sense here ?

tests/unit/Hub/Hub.Config.t.sol
```diff
@@ -227,7 +227,26 @@ contract HubConfigTest is HubBase {
     );
   }
 
+  function test_addAsset_fuzz_reverts_AssetAlreadyListed() public {
```
Kogaroshi: not `fuzz` here

DhairyaSethi:
@miguelmtzinf i think you meant to close https://github.com/aave/aave-v4/pull/964? 

this one is still very much on the table 😄

miguelmtzinf:
> @miguelmtzinf i think you meant to close https://github.com/aave/aave-v4/pull/964? 
> 
> this one is still very much on the table 😄 

correct, apologies

src/hub/Hub.sol
```diff
@@ -202,22 +205,24 @@ contract Hub is IHub, AccessManaged {
...
     asset.addedShares += shares;
     spoke.addedShares += shares;
-    asset.liquidity += amount.toUint120();
+    asset.liquidity = newLiquidity.toUint120();
 
     asset.updateDrawnRate(assetId);
 
-    asset.underlying.safeTransferFrom(from, address(this), amount);
+    // enforces spoke transfers the correct funds from user to hub
+    require(asset.underlying.balanceOf(address(this)) >= newLiquidity, InvalidAmountReceived());
```
miguelmtzinf: let's add a rescue function please

src/spoke/Spoke.sol
```diff
@@ -202,7 +204,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
     _validateSupply(reserve);
 
-    uint256 suppliedShares = reserve.hub.add(reserve.assetId, amount, msg.sender);
+    reserve.underlying.safeTransferFrom(msg.sender, address(reserve.hub), amount);
+    uint256 suppliedShares = reserve.hub.add(reserve.assetId, amount);
```
DhairyaSethi, line 210: let's cache reserve.hub?

src/hub/Hub.sol
DhairyaSethi: not needed, can happen by registering a new spoke, with addCap and add & remove

src/spoke/Spoke.sol
Kogaroshi: It actually increases the gas slightly (+8). Do we want to cache it for readability purposes then ?

src/spoke/Spoke.sol
```diff
@@ -3,6 +3,7 @@
 pragma solidity 0.8.28;
 
 import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
+import {SafeTransferLib} from 'src/dependencies/solady/SafeTransferLib.sol';
 import {IERC20Permit} from 'src/dependencies/openzeppelin/IERC20Permit.sol';
```
yan-man, line 6: nit: ordering of interfaces/contracts; also keeping openzeppelin ones together grouped and separate from the solady one

src/spoke/Spoke.sol
yan-man: in favor of caching where possible, in spite of slight gas incr

src/spoke/Spoke.sol
```diff
@@ -202,7 +204,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
     _validateSupply(reserve);
 
-    uint256 suppliedShares = reserve.hub.add(reserve.assetId, amount, msg.sender);
+    reserve.underlying.safeTransferFrom(msg.sender, address(reserve.hub), amount);
+    uint256 suppliedShares = reserve.hub.add(reserve.assetId, amount);
     userPosition.suppliedShares += suppliedShares.toUint120();
```
yan-man, line 211: should we check allowance is 0 after the transfer, or is this overkill

src/spoke/Spoke.sol
yan-man: hm nvm seems this isnt the "proper" style

src/spoke/Spoke.sol
Kogaroshi: We might not want that, some contracts built on top of a specific Spoke might give it unlimited allowance to reduce gas cost.

src/hub/interfaces/IHub.sol
```diff
@@ -239,6 +242,9 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev The `irData` must be empty if the interest rate strategy is not updated.
   error InvalidInterestRateStrategy();
 
+  /// @notice Thrown when the amount of underlying assets received is less than expected.
+  error InvalidAmountReceived();
+
   /// @notice Adds a new asset to the Hub.
   /// @dev The same underlying asset address can be added as an asset multiple times.
```
```suggestion
  /// @dev The same underlying asset address cannot be added as an asset multiple times.
```

src/hub/interfaces/IHub.sol
```diff
@@ -372,6 +378,9 @@ interface IHub is IHubBase, IAccessManaged {
   /// @return The address of the spoke.
   function getSpokeAddress(uint256 assetId, uint256 index) external view returns (address);
 
+  /// @notice Returns true when `underlying` is currently listed as an asset.
```
```suggestion
  /// @notice Returns whether the underlying is listed as an asset.
  /// @param underlying The address of the underlying asset.
  /// @return True if the underlying asset is listed.
```

src/hub/interfaces/IHubBase.sol
```diff
@@ -91,11 +91,12 @@ interface IHubBase {
 
   /// @notice Add asset on behalf of user.
   /// @dev Only callable by active spokes.
+  /// @dev Must receive the funds prior to calling, hub's balance is verified against the stored liquidity.
+  /// @dev Donated liquidity can be skimmed by any spoke.
```
miguelmtzinf:
```suggestion
  /// @dev Underlying assets must be transferred to the Hub before invocation.
```
I think the rest can be explained in docs

src/hub/interfaces/IHubBase.sol
```diff
@@ -116,18 +117,18 @@ interface IHubBase {
   /// @notice Restore assets on behalf of user.
   /// @dev Only callable by active spokes.
   /// @dev Interest is always paid off first from premium, then from drawn.
+  /// @dev Must receive the funds prior to calling, hub's balance is verified against the stored liquidity.
+  /// @dev Donated liquidity can be skimmed by any spoke.
```
miguelmtzinf:
```suggestion
  /// @dev Underlying assets must be transferred to the Hub before invocation.
```

src/hub/interfaces/IHubBase.sol
DhairyaSethi: think it's important to mention explicitly here that donated liquidity being skimmed is an expected feature 

```suggestion
  /// @dev Underlying assets must be transferred to the Hub before invocation.
  /// @dev Donated liquidity can be skimmed by any spoke.
```

src/hub/Hub.sol
miguelmtzinf: we can move this after `newLiquidity` definition. let's rm comment pls

miguelmtzinf: not sure about the error name tbh

src/hub/Hub.sol
```diff
@@ -279,25 +284,25 @@ contract Hub is IHub, AccessManaged {
...
-    uint256 totalAmount = drawnAmount + premiumAmount;
-    asset.liquidity += totalAmount.toUint120();
+    uint256 newLiquidity = asset.liquidity + drawnAmount + premiumAmount;
+    asset.liquidity = newLiquidity.toUint120();
 
     asset.updateDrawnRate(assetId);
 
-    asset.underlying.safeTransferFrom(from, address(this), totalAmount);
+    // enforces spoke transfers the correct funds from user to hub
+    require(asset.underlying.balanceOf(address(this)) >= newLiquidity, InvalidAmountReceived());
```
miguelmtzinf: should we follow same order as in `add`?

src/hub/Hub.sol
```diff
@@ -637,6 +642,11 @@ contract Hub is IHub, AccessManaged {
     return _spokes[assetId][spoke];
   }
 
+  /// @inheritdoc IHub
+  function isUnderlyingListed(address underlying) external view returns (bool) {
```
miguelmtzinf: i d suggest to move it up, right before `getAssetCount`
in IHubBase too.
Also, do we wanna expose functions to list all underlying addreses?

src/position-manager/GatewayBase.sol
```diff
@@ -47,13 +47,9 @@ abstract contract GatewayBase is IGatewayBase, Rescuable, Ownable2Step {
...
   /// @return The underlying asset for `reserveId` on the specified spoke.
-  /// @return The corresponding Hub address.
-  function _getReserveData(
-    address spoke,
-    uint256 reserveId
-  ) internal view returns (address, address) {
+  function _getReserveUnderlying(address spoke, uint256 reserveId) internal view returns (address) {
     ISpoke.Reserve memory reserveData = ISpoke(spoke).getReserve(reserveId);
-    return (reserveData.underlying, address(reserveData.hub));
+    return reserveData.underlying;
```
miguelmtzinf:
```suggestion
    return ISpoke(spoke).getReserve(reserveId).underlying;
```

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -86,7 +86,7 @@ interface ISpokeBase {
 
   /// @notice Supplies an amount of underlying asset of the specified reserve.
   /// @dev It reverts if the reserve associated with the given reserve identifier is not listed.
-  /// @dev The Hub pulls the underlying asset from the caller, so prior token approval is required.
+  /// @dev The spoke pulls the underlying asset from the caller, so prior token approval is required.
```
miguelmtzinf:
```suggestion
  /// @dev The Spoke pulls the underlying asset from the caller, so prior token approval is required.
```

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -132,7 +132,7 @@ interface ISpokeBase {
 
   /// @notice Repays a specified amount of underlying asset to a given reserve.
   /// @dev It reverts if the reserve associated with the given reserve identifier is not listed.
-  /// @dev The Hub pulls the underlying asset from the caller, so prior approval is required.
+  /// @dev The spoke pulls the underlying asset from the caller, so prior approval is required.
```
miguelmtzinf:
```suggestion
  /// @dev The Spoke pulls the underlying asset from the caller, so prior approval is required.
```

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -147,7 +147,7 @@ interface ISpokeBase {
 
   /// @notice Liquidates a user position.
   /// @dev It reverts if the reserves associated with any of the given reserve identifiers are not listed.
-  /// @dev Invokes Hub `restore`, and pulls underlying repaid debt assets from caller (Liquidator), hence it needs prior approval.
+  /// @dev The spoke pulls underlying repaid debt assets from caller (Liquidator), hence it needs prior approval.
```
miguelmtzinf:
```suggestion
  /// @dev The Spoke pulls underlying repaid debt assets from caller (Liquidator), hence it needs prior approval.
```

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -16,8 +16,8 @@ interface ITreasurySpoke is ISpokeBase {
   error InvalidAddress();
 
   /// @notice Supplies a specified amount of the underlying asset to a given reserve.
-  /// @dev The Hub pulls the underlying asset from the caller, so prior approval is required.
-  /// @dev The reserve identifier must match the asset identifier in the Hub.
+  /// @dev The spoke pulls the underlying asset from the caller, so prior approval is required.
```
miguelmtzinf:
```suggestion
  /// @dev The Spoke pulls the underlying asset from the caller, so prior approval is required.
```

src/spoke/interfaces/ITreasurySpoke.sol
```diff
@@ -16,8 +16,8 @@ interface ITreasurySpoke is ISpokeBase {
...
+  /// @dev The reserve identifier must match the asset identifier in the hub.
```
miguelmtzinf:
```suggestion
  /// @dev The reserve identifier must match the asset identifier in the Hub.
```

src/spoke/Spoke.sol
miguelmtzinf: let's cache, in repay too

Kogaroshi: can't cache in repay, we hit Stack Too Deep

DhairyaSethi: against caching here since the code is shorter (& cleaner) & apparently cheaper

src/hub/Hub.sol
Kogaroshi: we could, would that be necessary tho ?

src/hub/Hub.sol
```diff
@@ -43,6 +43,7 @@ contract Hub is IHub, AccessManaged {
   mapping(uint256 assetId => Asset) internal _assets;
   mapping(uint256 assetId => mapping(address spoke => SpokeData)) internal _spokes;
   mapping(uint256 assetId => EnumerableSet.AddressSet) internal _assetToSpokes;
+  EnumerableSet.AddressSet internal _underlyingSet;
```
miguelmtzinf: `_underlyingAssets` looks better to me

src/hub/interfaces/IHubBase.sol
miguelmtzinf: Do not think it's neccesary, but okay with adding it. Don't you think that it's more a general aspect of the Hub? maybe better to document in a `@dev` at contract level instead.

src/hub/Hub.sol
```diff
@@ -574,6 +576,11 @@ contract Hub is IHub, AccessManaged {
     return _assets[assetId].drawnRate;
   }
 
+  /// @inheritdoc IHub
+  function isUnderlyingListed(address underlying) external view returns (bool) {
```
miguelmtzinf: I d move it up before `getAssetCount`

src/hub/interfaces/IHubBase.sol
DhairyaSethi: i think it makes sense to immediately add here bc the first thought that comes to mind after reading assets must be transferred to hub is what happens to isolated donations

Kogaroshi: Yeah I'm also on the side of keeping it here, easier to notice when inspecting transfer flows too.

src/hub/Hub.sol
Kogaroshi: mb, didn't paste at the correct line before

tests/unit/Hub/Hub.Recover.t.sol
```diff
@@ -0,0 +1,103 @@
...
+      paused: false,
+      addCap: Constants.MAX_ALLOWED_SPOKE_CAP,
+      drawCap: Constants.MAX_ALLOWED_SPOKE_CAP,
+      riskPremiumThreshold: Constants.MAX_ALLOWED_COLLATERAL_RISK
+    });
+    vm.prank(ADMIN);
+    hub1.addSpoke(daiAssetId, recoverSpoke, spokeConfig);
+  }
+
+  function test_recovery_scenario_fuzz(uint256 lostAmount) public {
```
yan-man: can you add some comments above describing this test

src/hub/interfaces/IHub.sol
```diff
@@ -239,8 +242,11 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev The `irData` must be empty if the interest rate strategy is not updated.
   error InvalidInterestRateStrategy();
 
+  /// @notice Thrown when the amount of underlying assets received is less than expected.
+  error InvalidAmountReceived();
```
miguelmtzinf: I like `InsufficientLiquidity`

src/hub/interfaces/IHubBase.sol
miguelmtzinf: ok, how about rephrasing to this: `Extra underlying liquidity can be used by any Spoke through this action.`

DhairyaSethi: current statement seems perfectly succinct. problem is with? 
donated - we can rephrase it to 'unclaimed'/'extra liquidity'
skimmed - this is a very legitimate defi term for this, uni v2, curve, balancer extensively use it for this concept exactly & would prefer to retain this

tests/unit/Hub/Hub.Rescue.t.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    // remove all
+    Utils.remove({
+      hub: hub1,
+      assetId: daiAssetId,
+      caller: address(spoke1),
+      amount: 5e20,
+      to: alice
+    });
+    Utils.remove({hub: hub1, assetId: daiAssetId, caller: address(spoke2), amount: 10e22, to: bob});
```
yan-man, line 84: why are these removes here without any assertions after?

src/hub/interfaces/IHub.sol
DhairyaSethi: that also signals the revert scenario when trying to withdraw/borrow during liquidity crunch

tests/unit/Hub/Hub.Rescue.t.sol
Kogaroshi: to ensure all users can remove and there is enough liquidity (so the recover didn't pull more then expected). I'll add a comment there

src/hub/interfaces/IHub.sol
miguelmtzinf: I'm fine using this error for all, since fundamentally the revert reason is the same: there is no enough liquidity in the Hub to allow the action.

DhairyaSethi: i think it's a misleading for spoke integrations but sure

src/hub/interfaces/IHub.sol
```diff
@@ -179,6 +179,9 @@ interface IHub is IHubBase, IAccessManaged {
   /// @notice Thrown when an asset is not listed.
   error AssetNotListed();
 
+  /// @notice Thrown when an underlying asset is already listed.
+  error AssetAlreadyListed();
```
```suggestion
  /// @notice Thrown when an underlying asset is already listed.
  error UnderlyingAlreadyListed();
  
  /// @notice Thrown when an asset is not listed.
  error AssetNotListed();

```

src/hub/interfaces/IHubBase.sol
miguelmtzinf: Ok, let's go with the following then:
`Extra underlying liquidity retained in the Hub can be skimmed by any Spoke through this action.`

tests/gas/Hub.Operations.gas.t.sol
```diff
@@ -14,14 +14,17 @@ contract HubOperations_Gas_Tests is Base {
   }
 
   function test_add() public {
-    vm.prank(address(spoke1));
-    hub1.add(usdxAssetId, 1000e6, alice);
+    vm.startPrank(address(spoke1));
+    tokenList.usdx.transferFrom(alice, address(hub1), 1000e6);
+    hub1.add(usdxAssetId, 1000e6);
     vm.snapshotGasLastCall('Hub.Operations', 'add');
```
miguelmtzinf, line 20: this is now a bit counter-intuitive. Should we account for the transfer in the gas snapshot? same for restore

Kogaroshi: The transfer is not part of the Hub action now, so imo to should be included in the gas cost. It's part of the Spoke action and costs.

tests/unit/Hub/Hub.Recover.t.sol
```diff
@@ -0,0 +1,106 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Hub/HubBase.t.sol';
+
+contract HubRecoverTest is HubBase {
```
miguelmtzinf: i d suggest to rename to `HubRescueTest` or `HubRescueFundsTest` so it's clearer

miguelmtzinf: i d change all `recover` to `rescue` too

tests/Base.t.sol
```diff
@@ -533,10 +554,10 @@ abstract contract Base is Test {
       }),
       new bytes(0)
     );
-    // add DAI again
+    // add usdz
```
```suggestion
    // add USDZ
```

tests/Base.t.sol
```diff
@@ -2163,7 +2184,7 @@ abstract contract Base is Test {
     uint256 indexDelta
   ) internal pure returns (uint256 feesAmount) {
     return
-      indexDelta.rayMulDown(initialDrawnShares + initialPremiumShares).percentMulDown(liquidityFee);
+      indexDelta.rayMulUp(initialDrawnShares + initialPremiumShares).percentMulDown(liquidityFee);
```
miguelmtzinf, line 2187: why this change is needed now?

tests/Base.t.sol
```diff
@@ -2361,6 +2382,16 @@ abstract contract Base is Test {
     );
   }
 
+  function assertHubLiquidity(IHub targetHub, uint256 assetId, string memory label) internal view {
```
```suggestion
  function _assertHubLiquidity(IHub targetHub, uint256 assetId, string memory label) internal view {
```
same above `_assertBorrowRateSynced`

tests/Utils.sol
```diff
@@ -18,10 +18,10 @@ library Utils {
     uint256 amount,
     address user
   ) internal returns (uint256) {
-    IHub hub = IHub(address(hub));
-    approve(hub, assetId, user, amount);
+    approve(IHub(address(hub)), assetId, caller, user, amount);
```
miguelmtzinf: can we fix these so many castings? here and in helper functions

tests/Utils.sol
```diff
@@ -18,10 +18,10 @@ library Utils {
...
+    transferFrom(IHub(address(hub)), assetId, caller, user, address(hub), amount);
```
miguelmtzinf: can just the `caller` transfer the funds?

tests/unit/Hub/Hub.Add.t.sol
```diff
@@ -757,4 +810,234 @@ contract HubAddTest is HubBase {
...
+      'hub token balance post-add'
+    );
+  }
+
+  function test_add_fuzz_skimDonations_wrongSpokeTransfer(
+    uint256 assetId,
+    uint256 amount,
+    uint256 donationAmount
+  ) public {
+    assetId = bound(assetId, 0, hub1.getAssetCount() - 3); // Exclude duplicated DAI and usdy
```
miguelmtzinf: no duplicated DAI anymore, can you search and fix across tests?

tests/unit/Hub/Hub.Add.t.sol
```diff
@@ -757,4 +810,234 @@ contract HubAddTest is HubBase {
...
+
+    // normal deposit
+    vm.startPrank(address(spoke1));
+    underlying.transferFrom(alice, address(hub1), amount);
+
+    vm.expectEmit(address(hub1));
+    emit IHubBase.Add(assetId, address(spoke1), shares, amount);
+
+    uint256 addedShares = hub1.add(assetId, amount);
+    vm.stopPrank();
```
miguelmtzinf: should we use utils helper for brevity

tests/Base.t.sol
Kogaroshi: Used in `tests/unit/Spoke/Spoke.AccrueLiquidityFee.t.sol`, seems like it wasn't updated when the feeShare calculations were, and it failed during the CI fuzz runs when the expected amount was 0 instead of 2 wei.
Weirdly enough this failing case didn't appear before.

tests/Utils.sol
Kogaroshi: Here the `caller` is the Spoke, and we want it to transfer the funds from the `user` to the Hub, to mock the real Spoke behavior

tests/unit/Hub/Hub.Add.t.sol
Kogaroshi: true, no duplicate dai. Do we also want to include USDY in those potential tests here ? 
Because `hub1` current order is `[...,USDY,USDZ]`, so if we want. to keep excluding USDY better to either keep the current reduction, or change the order.
Otherwise yeah we can remove the exclusion all together.

tests/unit/Hub/Hub.Add.t.sol
Kogaroshi: utils helpers do not return the `shares` value in the current version, if we want to replace we'll need to make them return the value now

tests/gas/Hub.Operations.gas.t.sol
miguelmtzinf: not part on add/restore but part on remove/draw?

Kogaroshi: on remove/draw it's the Hub doing the transfers still, so it's part of the specific action gas consumption. but on add/restore yeah it's not up to the Hub to execute the transfer, it can be done however the Spoke implements it, so it's more of a variable cost there, including the raw transferFrom we do right now would not reflect all cases.

tests/unit/Hub/Hub.Add.t.sol
```diff
@@ -13,6 +13,9 @@ contract HubAddTest is HubBase {
   function setUp() public override {
     super.setUp();
 
+    TestnetERC20 usda = new TestnetERC20('USDA', 'USDA', Constants.MIN_ALLOWED_UNDERLYING_DECIMALS);
```
miguelmtzinf, line 16: why do we need this?

Kogaroshi: it's the ERC20 used for `minDecimalAssetId`

tests/gas/Hub.Operations.gas.t.sol
miguelmtzinf: because the transfer is included in remove/draw, I think it makes sense to count for the transfer in add/restore operations, so we can compare apples to apples. Also, an add/restore operation cannot succeed without a transfer beforehand.

tests/unit/Hub/Hub.Skim.t.sol
```diff
@@ -0,0 +1,318 @@
...
+      'hub token balance post-add'
+    );
+  }
+
+  function test_skimAdd_fuzz_donationBeforeAdd(
+    uint256 assetId,
+    uint256 amount,
+    uint256 donationAmount
+  ) public {
+    assetId = bound(assetId, 0, hub1.getAssetCount() - 3); // Exclude duplicated DAI and usdy
```
miguelmtzinf:
```suggestion
    assetId = bound(assetId, 0, hub1.getAssetCount() - 3); // Exclude usdy
```
can you fix across files pls?

tests/unit/Hub/Hub.Rescue.t.sol
```diff
@@ -86,21 +86,138 @@ contract HubRescueTest is HubBase {
     assertEq(underlying.balanceOf(address(hub1)), 0, 'final hub amount');
   }
 
+  /// @dev Recovery of funds directly transferred to the hub including interest accrual
```
miguelmtzinf: Rescue*

tests/unit/Hub/Hub.Rescue.t.sol
```diff
@@ -86,21 +86,138 @@ contract HubRescueTest is HubBase {
...
+      hub1.previewAddByShares(daiAssetId, 1),
+      'recover amount'
+    ); // can differ by up to 1 share worth of assets due to remove donation rounding
+    assertEq(recoverAddedShares, recoverWithdrawnShares, 'recover shares');
+    assertEq(finalHubBalance, prevHubBalance - recoverAmount, 'hub balance');
+    assertEq(finalRecoveryBalance, prevRecoveryBalance + recoverAmount, 'recovery balance');
+    _assertHubLiquidity(hub1, daiAssetId, 'hub1.recover');
+  }
+
+  /// @dev Another spoke cannot improperly recover liquidity fee without transferring underlying tokens
```
miguelmtzinf: recue*

tests/gas/Hub.Operations.gas.t.sol
DhairyaSethi: the transfer strategy can be anything, spoke can utilise previously directly transferred assets, permit2 transfer (which consumes approval sig as well), expect a transfer in the life cycle before where it does not initiate the transfer

DhairyaSethi: it's similar reasoning to why we weren't including approval cost in the gas tests before

miguelmtzinf: don't agree. While I see your point that transfer strat can be anything (also transfer gas costs depends on underlying token), this change makes the snapshots a bit confusing because now we don't clearly know how much an effective add operation costs and we cannot compare an add/restore with the rest.
Id revert, or create a separate snapshot for an add/restore without transfer.

DhairyaSethi: see before this PR also when add was pulling tokens you require prior approval - w/o that ALSO the add operation would fail and we were not considering the approval txn cost

PR #956, opened 2025-10-23
rft: calculateUserAccountData option 2

miguelmtzinf:


tests/mocks/MockSpoke.sol
```diff
@@ -44,11 +44,7 @@ contract MockSpoke is Spoke, Test {
     userPosition.drawnShares += drawnShares.toUint128();
     positionStatus.setBorrowing(reserveId, true);
 
-    ISpoke.UserAccountData memory userAccountData = _validateUserPosition({
-      user: onBehalfOf,
-      refreshConfig: true,
-      validateHealthFactor: false
-    });
+    ISpoke.UserAccountData memory userAccountData = _calculateAndValidateUserPosition(onBehalfOf);
```
miguelmtzinf: This breaks some tests because we shouldn't check HF here... we were using the "potentially" before
any idea?

yan-man: can we not call `_calculateUserAccountData(onBehalfOf, true);` ? doing so fixes the tests for me locally

src/spoke/Spoke.sol
```diff
@@ -313,11 +305,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       _positionStatus[onBehalfOf].setBorrowing(reserveId, false);
     }
 
-    UserAccountData memory userAccountData = _validateUserPosition({
-      user: onBehalfOf,
-      refreshConfig: false,
-      validateHealthFactor: false
-    });
+    UserAccountData memory userAccountData = _calculateUserPosition(onBehalfOf);
```
avniculae, line 308: the reason I don't like `calculateUserPosition` is because `UserPosition` is also used to name the struct we store per user and per reserve, whereas here it's just per user

wdyt

src/spoke/Spoke.sol
```diff
@@ -261,11 +257,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       positionStatus.setBorrowing(reserveId, true);
     }
 
-    uint256 newRiskPremium = _validateUserPosition({
-      user: onBehalfOf,
-      refreshConfig: true,
-      validateHealthFactor: true
-    }).riskPremium;
+    uint256 newRiskPremium = _calculateAndValidateUserPosition(onBehalfOf).riskPremium;
```
DhairyaSethi, line 260: it's not immediately obvious from this fn name that it refreshes dynamic config 

you require already knowing an invariant that any time we validate, it's always with latest config

DhairyaSethi: this function changes state, that's not intuitive from the name as well

src/spoke/Spoke.sol
yan-man: I like some flavor of the old *UserAccountData name, given it returns a UserAccountData struct

src/spoke/Spoke.sol
```diff
@@ -956,10 +924,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
   /// @notice View version of `_calculateUserAccountData`.
   /// @dev It does not refresh dynamic config before calculating.
-  function _viewCalculateUserAccountData(
-    address user
-  ) internal view returns (UserAccountData memory) {
-    return _castToView(_calculateUserAccountData)(user, false);
+  function _viewCalculateUserPosition(address user) internal view returns (UserAccountData memory) {
```
yan-man, line 927: prefer the old name

miguelmtzinf:
Closing in favor of #962

PR #957, opened 2025-10-24
feat: spoke approvals on hub - option 4

yan-man:
- add spoke approvals to hub to prevent malicious spokes from utilizing hanging token-hub approvals
- todo: approval-specific tests


Kogaroshi:
If going that route, it might be worth adding a permitSpoke on the Hub to for the user to sign so the Spoke can serve it to the Hub, so even if there is an extra step needed, batching with multicall would not worsen the UX too much ?

yan-man:
closed in favor of https://github.com/aave/aave-v4/pull/955

PR #958, opened 2025-10-24
test: liquidation helper for gas snapshot

yan-man:
- update liquidation snapshot test so that user pos becomes liquidatable without an action that causes rp=0 immediately. Thus the liquidation gas test does not skip notify when rp is 0 -> 0, giving a clearer picture of gas changes due to feature setting rp=0 when underwater


yan-man:
see separate PR; updated to merge directly into main: https://github.com/aave/aave-v4/pull/974

PR #941, opened 2025-10-21
Fix: Validate hub actions via underflow

CheyenneAtapour:
- Removes Hub's validation checks on spokes’ accounting, relying instead on math underflow for validation.
- UX is not affected, as these checks apply only to functions that are not user-facing but are invoked by Spokes.
  - Spoke checks max withdrawal limit before calling `remove` (max withdraw if input amount is greater than max)
  - Spoke checks max repayable amount before calling `repay` (max repay if input amount is greater than total debt)
- Adds check for available liquidity on `sweep` for consistency
- ~4k gas reduction on `remove` and <1k gas on `restore`

PR #959, opened 2025-10-24
opt: cache index

DhairyaSethi:
- move `getAssetDrawnIndex()` to hub base because the latest drawn index is not visible with `getAsset` method
- use this getter to cache index & compute debt on spoke (we evaluated approach for hub to expose method to compute debt owing to the fact that a spoke can connect to different hubs and it needs to respect each hub's internal accounting mechanism -> this can be dealt with later as & when need arises since spoke is upgradable)
- fixes compiler warning on `_validateRemove` to rm unused param

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -482,6 +482,11 @@ contract Hub is IHub, AccessManaged {
   }
 
   /// @inheritdoc IHubBase
+  function getAssetDrawnIndex(uint256 assetId) external view returns (uint256) {
+    return _assets[assetId].getDrawnIndex();
+  }
+
+  /// @inheritdoc IHubBase
   function getAddedAssets(uint256 assetId) external view returns (uint256) {
     return _assets[assetId].totalAddedAssets();
   }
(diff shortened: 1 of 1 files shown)


src/spoke/Spoke.sol
```diff
@@ -806,15 +806,15 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       Reserve storage reserve = _reserves[reserveId];
       uint256 assetId = reserve.assetId;
       IHubBase hub = reserve.hub;
+      uint256 drawnIndex = hub.getAssetDrawnIndex(assetId);
 
       uint256 oldPremiumShares = userPosition.premiumShares;
       uint256 oldPremiumOffset = userPosition.premiumOffset;
-      uint256 accruedPremium = hub.previewRestoreByShares(assetId, oldPremiumShares) -
-        oldPremiumOffset;
+      uint256 accruedPremium = oldPremiumShares.rayMulUp(drawnIndex) - oldPremiumOffset;
```
yan-man, line 813: i quite like this bc the rayMulUp and rayMulDown are explicitly clear and easy to follow now

src/hub/Hub.sol
```diff
@@ -758,12 +758,7 @@ contract Hub is IHub, AccessManaged {
     );
   }
 
-  function _validateRemove(
-    Asset storage asset,
```
yan-man, line 762: lol why did we have this before if we didnt use asset, just consistency w other validate methods?

DhairyaSethi: we were calculating removable in this and removed it in https://github.com/aave/aave-v4/pull/941 and was missed

PR #960, opened 2025-10-24
rft: calculateUserAccountData option 3

miguelmtzinf:


src/spoke/Spoke.sol
```diff
@@ -662,29 +686,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    emit RefreshAllUserDynamicConfig(user);
-    return accountData;
-  }
-
-  /// @notice Refreshes the dynamic config and calculates the user account data if `refreshConfig` is true.
-  /// @dev User RiskPremium calc runs until the first of either debt or collateral is exhausted.
-  function _calculateAndPotentiallyRefreshUserAccountData(
+  /// @notice Calculates the user account data.
+  /// @dev It runs user risk calculations until the first of either collateral or debt is exhausted.
+  function _calculateUserAccountData(
```
DhairyaSethi: not obvious this can modify state; i think the idea is it's inferred from the named arg (refreshConfig) usage of this?

src/spoke/Spoke.sol
```diff
@@ -922,9 +930,10 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     return config.active && config.approval[user];
   }
 
-  function _calculateUserAccountData(address user) internal view returns (UserAccountData memory) {
-    // SAFETY: function does not modify state when refreshConfig is false.
-    return _castToView(_calculateAndPotentiallyRefreshUserAccountData)(user, false);
+  /// @notice View version of `_calculateUserAccountData`.
+  /// @dev It does not refresh dynamic config before calculating.
+  function _viewCalculateUserPosition(address user) internal view returns (UserAccountData memory) {
+    return _castToView(_calculateUserAccountData)(user, false);
```
DhairyaSethi: would retain the SAFETY comment pls, and we can just inline this in the getter, it's only used once in the code so doesn't need internal method abstraction

src/spoke/Spoke.sol
miguelmtzinf: in favor of renaming and not creating wrapper functions.
Suggestions: `computeUserAccountData`, `calculateUserAccountDataWithRefresh`, `computeUserAccountDataWithRefresh`, `processUserAccountData`
I think the last one would be fine + adding dev docs here.

src/spoke/Spoke.sol
```diff
@@ -617,15 +642,14 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+        maxLiquidationBonus: _dynamicConfig[reserveId][_userPositions[user][reserveId].configKey]
+          .maxLiquidationBonus
       });
   }
 
   /// @inheritdoc ISpoke
   function getUserAccountData(address user) external view returns (UserAccountData memory) {
-    return _calculateUserAccountData(user);
+    return _viewCalculateUserPosition(user);
   }
```
DhairyaSethi, line 654:
```suggestion
  function getUserAccountData(address user) external view returns (UserAccountData memory) {
    // SAFETY: function does not modify state when refreshConfig is false.
    return _castToView(_calculateUserAccountData)(user, false);
  }
```

src/spoke/Spoke.sol
yan-man: in favor of: calculateUserAccountDataWithRefresh

miguelmtzinf:
Closing in favor of #962

PR #962, opened 2025-10-24
rft: calculateUserAccountData option 4

miguelmtzinf:
Refactors all execution flows that require calculation of the user position.

- No functional change, just renaming of the internal methods
- `_castToView` trick is now only applied for the external getter (were used in `repay` too)
- Optimization to add clarity to execution flows involving calculateUserAccountData

### DIFF
--- a/src/spoke/interfaces/ISpoke.sol
+++ b/src/spoke/interfaces/ISpoke.sol
@@ -90,10 +90,10 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   }
 
   /// @notice User position status data.
-  /// @dev map The bitmap of the position status.
+  /// @dev map The map of bitmap buckets for the position status.
   /// @dev hasPositiveRiskPremium True if the user position has a risk premium strictly greater than 0.
   struct PositionStatus {
-    mapping(uint256 slot => uint256) map;
+    mapping(uint256 bucket => uint256) map;
     bool hasPositiveRiskPremium;
   }
 


src/spoke/Spoke.sol
```diff
@@ -662,29 +663,30 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+  function _calculateUserAccountData(address user) internal returns (UserAccountData memory) {
+    return _processUserAccountData(user, false);
   }
 
-  /// @notice Refreshes the dynamic config and calculates the user account data if `refreshConfig` is true.
-  /// @dev User RiskPremium calc runs until the first of either debt or collateral is exhausted.
-  function _calculateAndPotentiallyRefreshUserAccountData(
+  /// @notice Process the user account data.
+  /// @dev It updates user dynamic config if `refreshConfig` is true.
+  /// @dev It runs user risk calculations until the first of either collateral or debt is exhausted.
```
DhairyaSethi: i think its fine to remove this comment from here, it's kinda pointless and add it below during the risk premium calculations somewhere

src/spoke/Spoke.sol
```diff
@@ -662,29 +663,30 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+  /// @notice Calculates the user account data with the current user dynamic config.
+  function _calculateUserAccountData(address user) internal returns (UserAccountData memory) {
+    return _processUserAccountData(user, false);
   }
 
-  /// @notice Refreshes the dynamic config and calculates the user account data if `refreshConfig` is true.
-  /// @dev User RiskPremium calc runs until the first of either debt or collateral is exhausted.
-  function _calculateAndPotentiallyRefreshUserAccountData(
+  /// @notice Process the user account data.
+  /// @dev It updates user dynamic config if `refreshConfig` is true.
```
```suggestion
  /// @notice Process the user account data. Updates user dynamic config if `refreshConfig` is true.
```
i think the state change is imp to mention in notice

src/spoke/Spoke.sol
```diff
@@ -662,29 +663,30 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-  /// @notice Refreshes user dynamic configuration and checks the position is healthy.
-  /// @return The user's new risk premium.
-  function _refreshAndValidateUserPosition(address user) internal returns (uint256) {
-    UserAccountData memory accountData = _calculateAndRefreshUserAccountData(user);
+  /// @notice Calculates and validates the user account data.
+  /// @dev It refreshes the dynamic config before calculation.
+  /// @dev It checks that the health factor is above the liquidation threshold.
+  function _refreshAndValidateUserAccountData(
+    address user
+  ) internal returns (UserAccountData memory) {
```
DhairyaSethi, line 673: so in all the usage of this function, it only uses risk premium; so i think its fine if it returns just that

currently it looks like the optimiser already does this so nbd

src/spoke/Spoke.sol
```diff
@@ -662,29 +663,30 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-  /// @notice Refreshes the dynamic config and calculates the user account data.
-  function _calculateAndRefreshUserAccountData(
-    address user
-  ) internal returns (UserAccountData memory) {
-    UserAccountData memory accountData = _calculateAndPotentiallyRefreshUserAccountData(user, true);
-    emit RefreshAllUserDynamicConfig(user);
-    return accountData;
+  /// @notice Calculates the user account data with the current user dynamic config.
+  function _calculateUserAccountData(address user) internal returns (UserAccountData memory) {
+    return _processUserAccountData(user, false);
```
```suggestion
    return _processUserAccountData(user, false); // does not modify state
```
optional

src/spoke/Spoke.sol
```diff
@@ -617,15 +618,15 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+        maxLiquidationBonus: _dynamicConfig[reserveId][_userPositions[user][reserveId].configKey]
+          .maxLiquidationBonus
       });
   }
 
   /// @inheritdoc ISpoke
   function getUserAccountData(address user) external view returns (UserAccountData memory) {
-    return _calculateUserAccountData(user);
+    // SAFETY: function does not modify state when refreshConfig is false.
+    return _castToView(_calculateUserAccountData)(user);
```
```suggestion
    // SAFETY: function does not modify state when `refreshConfig` is false.
    return _castToView(_processUserAccountData)(user, false);
```

DhairyaSethi: is more accurate

src/spoke/Spoke.sol
```diff
@@ -765,6 +766,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
     uint256 debtValueLeftToCover = accountData.totalDebtValue;
     collateralInfo.sortByKey(); // sort by collateral risk in ASC, collateral value in DESC
+
+    // it runs until the first of either collateral or debt is exhausted
```
```suggestion
    // runs until either the collateral or debt is exhausted
```

PR #964, opened 2025-10-24
fix: prevent `receiveShares` if liquidator is `usingAsCollateral`

yan-man:
- if liquidator is using the collateral reserve as collateral, revert if they choose to receive shares, due to skipped notifyRiskPremium

### DIFF
--- a/src/spoke/libraries/LiquidationLogic.sol
+++ b/src/spoke/libraries/LiquidationLogic.sol
@@ -366,15 +366,12 @@ library LiquidationLogic {
     );
     require(params.collateralReserveBalance > 0, ISpoke.ReserveNotSupplied());
     require(params.debtReserveBalance > 0, ISpoke.ReserveNotBorrowed());
-    // if the liquidator is using the collateral reserve as collateral, they cannot receive shares
-    require(
-      !(params.receiveShares && params.liquidatorUsingAsCollateral),
-      ISpoke.CannotReceiveShares()
-    );
-    require(
-      !(params.receiveShares && params.collateralReserveFrozen),
-      ISpoke.CannotReceiveShares()
-    );
+    if (params.receiveShares) {
+      require(
+        !params.liquidatorUsingAsCollateral && !params.collateralReserveFrozen,
+        ISpoke.CannotReceiveShares()
+      );
+    }
   }
 
   /// @notice Calculates the liquidation amounts.


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -363,7 +366,12 @@ library LiquidationLogic {
     );
     require(params.collateralReserveBalance > 0, ISpoke.ReserveNotSupplied());
     require(params.debtReserveBalance > 0, ISpoke.ReserveNotBorrowed());
-    require(!params.receiveShares || !params.collateralReserveFrozen, ISpoke.CannotReceiveShares());
+    // if the liquidator is using the collateral reserve as collateral, they cannot receive shares
+    require(
```
miguelmtzinf: a bit hard to read, i d split i think

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -363,7 +366,12 @@ library LiquidationLogic {
...
+      !(params.receiveShares &&
+        (params.liquidatorUsingAsCollateral || params.collateralReserveFrozen)),
```
miguelmtzinf: we dont check cf > 0 right, for simplicity

yan-man: it's checked above in L364

src/spoke/libraries/LiquidationLogic.sol
yan-man: https://github.com/aave/aave-v4/pull/964/commits/6016d915a34c030fd2b47f51d4acf9062259536b

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -363,7 +366,15 @@ library LiquidationLogic {
...
-    require(!params.receiveShares || !params.collateralReserveFrozen, ISpoke.CannotReceiveShares());
+    // if the liquidator is using the collateral reserve as collateral, they cannot receive shares
+    require(
+      !(params.receiveShares && params.liquidatorUsingAsCollateral),
+      ISpoke.CannotReceiveShares()
+    );
+    require(
+      !(params.receiveShares && params.collateralReserveFrozen),
+      ISpoke.CannotReceiveShares()
+    );
```
DhairyaSethi:
```suggestion
    if (params.receiveShares) {
      require(!params.liquidatorUsingAsCollateral && !params.collateralReserveFrozen), ISpoke.CannotReceiveShares());
    }
```

src/spoke/Spoke.sol
```diff
@@ -341,7 +341,10 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       activeCollateralCount: userAccountData.activeCollateralCount,
       borrowedCount: userAccountData.borrowedCount,
       liquidator: msg.sender,
-      receiveShares: receiveShares
+      receiveShares: receiveShares,
+      liquidatorUsingAsCollateral: receiveShares
+        ? _positionStatus[msg.sender].isUsingAsCollateral(collateralReserveId)
+        : false
```
DhairyaSethi, line 347: feels a bit scary to pass a conditional value here, i understand why you did it but what i would do is pass additional param _positionStatus storage pointer and then read inside the validate fn

miguelmtzinf: it's also weird that we are setting to `false` because `receiveShares` is `false`, even if it's actually `true`

miguelmtzinf: passing the storage pointer would need a lot of changes (tests too) and we d need to pass `collateralReserveId` to validate fn too.
We could read even if neccesary, increasing 2k for regular liquidation... not good.
Leaning towards passing `receiveSharesNotCollateral` flag here instead (right now we check `receiveShares` twice).

```
      receiveShares: receiveShares,
      liquidatorUsingAsCollateral: receiveShares
        ? _positionStatus[msg.sender].isUsingAsCollateral(collateralReserveId)
        : false
```

```
      receiveSharesNotCollateral: receiveShares
        ? _positionStatus[msg.sender].isUsingAsCollateral(collateralReserveId)
        : false
```

DhairyaSethi:
closing in favour of #966

PR #965, opened 2025-10-25
rft: split refreshAndValidate

miguelmtzinf:
Splits the `refreshAndValidate` function so:
- `refreshUserAccountData` calls `calculateUserAccountData` with `refreshConfig: true`. Now it has only 1 responsability.
- `validateHealthFactor` is called explictly. Now it's clear when HF is checked on each action.


miguelmtzinf:
Closing, as having both the refresh and health factor check in the same function is considered a precautionary measure to avoid footguns.

PR #966, opened 2025-10-27
fix: revert when liquidator enabled collateral on receive shares

DhairyaSethi:


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -124,7 +124,7 @@ library LiquidationLogic {
   /// @param collateralReserve The collateral reserve to seize during liquidation.
   /// @param debtReserve The debt reserve to repay during liquidation.
   /// @param positions The user positions mapping in storage.
-  /// @param positionStatus The user's position status.
+  /// @param positionStatuses The user position statuses mapping in storage.
```
```suggestion
  /// @param positionStatus The mapping of user position statutes
```
i know it's changed now but i d keep it `positionStatus` tbh

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -358,12 +363,19 @@ library LiquidationLogic {
       ISpoke.HealthFactorNotBelowThreshold()
     );
     require(
-      params.isUsingAsCollateral && params.collateralFactor > 0,
+      positionStatuses[params.user].isUsingAsCollateral(params.collateralReserveId) &&
+        params.collateralFactor > 0,
```
miguelmtzinf: do we wanna change the order so cf gets validated first?

DhairyaSethi: doesn't matter we have to check both always but yep

src/spoke/libraries/LiquidationLogic.sol
DhairyaSethi: it will be confusing bc we refer to the mapped var (after mapping resolution) as `positionStatus` also (see line 136 here, and every invocation on spoke)

avniculae: I also think it's fine to keep `positionStatus`

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -358,12 +363,19 @@ library LiquidationLogic {
...
       ISpoke.CollateralCannotBeLiquidated()
     );
     require(params.collateralReserveBalance > 0, ISpoke.ReserveNotSupplied());
     require(params.debtReserveBalance > 0, ISpoke.ReserveNotBorrowed());
-    require(!params.receiveShares || !params.collateralReserveFrozen, ISpoke.CannotReceiveShares());
```
avniculae, line 366: can you remind me why coll reserve can be frozen if liquidator does not receive shares?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -124,7 +124,7 @@ library LiquidationLogic {
   /// @param collateralReserve The collateral reserve to seize during liquidation.
   /// @param debtReserve The debt reserve to repay during liquidation.
   /// @param positions The user positions mapping in storage.
-  /// @param positionStatus The user's position status.
+  /// @param positionStatus The mapping of user position status in storage.
```
miguelmtzinf: do not like the `in storage`. It's the only place where we are using this term.

DhairyaSethi: i mean it's not the only place, see above, but rewording

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -358,12 +362,19 @@ library LiquidationLogic {
...
     require(
-      params.isUsingAsCollateral && params.collateralFactor > 0,
+      params.collateralFactor > 0 &&
+        positionStatus[params.user].isUsingAsCollateral(params.collateralReserveId),
       ISpoke.CollateralCannotBeLiquidated()
     );
     require(params.collateralReserveBalance > 0, ISpoke.ReserveNotSupplied());
     require(params.debtReserveBalance > 0, ISpoke.ReserveNotBorrowed());
-    require(!params.receiveShares || !params.collateralReserveFrozen, ISpoke.CannotReceiveShares());
+    if (params.receiveShares) {
```
miguelmtzinf, line 372:
```suggestion
    // liquidator cannot receive collateral shares if the reserve is frozen or if it is being used as collateral.
    if (params.receiveShares) {
```

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: let's add the comment below so it's clearer

src/spoke/Spoke.sol
```diff
@@ -344,10 +344,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       liquidator: msg.sender,
       receiveShares: receiveShares
     });
+    Reserve storage debtReserve = _reserves[debtReserveId];
```
avniculae: should we declare these before params above? I'd like us to keep `_getUserDebt` as close as possible to params, since it populates some fields (debts and accrued premium)

src/spoke/Spoke.sol
```diff
@@ -356,10 +358,10 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     ];
 
     bool isUserInDeficit = LiquidationLogic.liquidateUser(
-      _reserves[collateralReserveId],
-      _reserves[debtReserveId],
+      collateralReserve,
+      debtReserve,
       _userPositions,
-      _positionStatus[user],
+      _positionStatus,
```
avniculae, line 363: alternatively, should we pass 2 positions status inputs? for user and liquidator

avniculae: I had a similar comment to Yan's PR when we started passing `_userPositions` as well

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: thanks!

src/spoke/Spoke.sol
DhairyaSethi: i think it's fine, liquidation position status is loaded on demand and saves us a keccak hash

right now, to the keen observer there is a duplicate lookup for user position status (which is always needed) but opt catches it

src/spoke/Spoke.sol
```diff
@@ -327,7 +327,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     uint256 debtToCover,
     bool receiveShares
   ) external {
+    Reserve storage debtReserve = _reserves[debtReserveId];
+    Reserve storage collateralReserve = _reserves[collateralReserveId];
```
```suggestion
    Reserve storage collateralReserve = _reserves[collateralReserveId];
    Reserve storage debtReserve = _reserves[debtReserveId];
```
lol

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -358,12 +362,20 @@ library LiquidationLogic {
...
     require(
-      params.isUsingAsCollateral && params.collateralFactor > 0,
+      params.collateralFactor > 0 &&
+        positionStatus[params.user].isUsingAsCollateral(params.collateralReserveId),
       ISpoke.CollateralCannotBeLiquidated()
     );
     require(params.collateralReserveBalance > 0, ISpoke.ReserveNotSupplied());
     require(params.debtReserveBalance > 0, ISpoke.ReserveNotBorrowed());
-    require(!params.receiveShares || !params.collateralReserveFrozen, ISpoke.CannotReceiveShares());
+    // liquidator cannot receive collateral shares if the reserve is frozen or if it is being used as collateral.
```
```suggestion
    // liquidator cannot receive collateral shares if the reserve is frozen or if it is being used as collateral
```

src/spoke/libraries/LiquidationLogic.sol
yan-man: @avniculae just consistency with frozen behavior, where we dont allow `supply` and this is a form of that for liquidator

PR #968, opened 2025-10-28
rft : Refactor Hub to remove `assetId` and use `underlying`

DhairyaSethi:


Kogaroshi:
Closed because decision is to go with https://github.com/aave/aave-v4/pull/955

PR #969, opened 2025-10-28
rft : Refactor Hub to remove `assetId` and use `asset`

DhairyaSethi:
Refactors the Hub contract to remove `assetId`, using the asset underlying address as key for mappings instead.
This refactor also includes the changes made in #955 

TODO : 
- fix currently broken unit tests
- find good value for `MAX_ALLOWED_ASSET_COUNT`


Kogaroshi:

Issue #929, opened 2025-10-17
Flip `collateralToLiquidate` rounding in liquidations


avniculae:
Currently, `collateralToLiquidate` is rounded down when being computed

https://github.com/aave/aave-v4/blob/886cbce636e9edaa0d6798dc6d1c67312e9f8ff8/src/spoke/libraries/LiquidationLogic.sol#L405-L408

This can round down to 0, which will make liquidation revert (hub reverts if 0 amount is removed)

We should round up instead to avoid this scenario. If we do this, and `debtToLiquidate` is recomputed, then `debtToLiquidate` might exceed `debtReserveBalance`. Hence, we need to add a cap on `debtToLiquidate` if recomputed

PR #970, opened 2025-10-28
fix: Rollback liquidation rounding and skip collateral removal when `collateralToLiquidator` is 0

### LINKED ISSUE #972: Liquidators might be incentivised to perform tiny liquidation due to rounding
- `collateralToLiquidate` is rounded up when computed liquidations (initial issue #929) 
- this however opens up possibilities for liquidators to profit from liquidating very small amounts, where the additional 1 wei of collateral received due to the rounding direction makes the attack profitable
- instead, consider reverting the rounding (to round it down) and skipping the `hub.remove()` call if `collateralToLiquidator` is 0

avniculae:
- collateralToLiquidate is rounded up when computed liquidations (initial issue https://github.com/aave/aave-v4/issues/929)
- this however opens up possibilities for liquidators to profit from liquidating very small amounts, where the additional 1 wei of collateral received due to the rounding direction makes the attack profitable
- instead, revert the rounding (to round it down) and instead skip the `hub.remove()` call if `collateralToLiquidator` is 0

Closes #972

### DIFF
--- a/src/spoke/SpokeConfigurator.sol
+++ b/src/spoke/SpokeConfigurator.sol
@@ -14,6 +14,8 @@ import {ISpokeConfigurator} from 'src/spoke/interfaces/ISpokeConfigurator.sol';
 contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
   using SafeCast for uint256;
 
+  mapping(address spoke => uint256) internal _maxReserves;
+
   /// @dev Constructor.
   /// @param owner_ The address of the owner.
   constructor(address owner_) Ownable(owner_) {}
@@ -69,6 +71,12 @@ contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
   }
 
   /// @inheritdoc ISpokeConfigurator
+  function updateMaxReserves(address spoke, uint256 maxReserves) external onlyOwner {
+    _maxReserves[spoke] = maxReserves;
+    emit UpdateMaxReserves(spoke, maxReserves);
+  }
+
+  /// @inheritdoc ISpokeConfigurator
   function addReserve(
     address spoke,
     address hub,
@@ -77,6 +85,10 @@ contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
     ISpoke.ReserveConfig calldata config,
     ISpoke.DynamicReserveConfig calldata dynamicConfig
   ) external onlyOwner returns (uint256) {
+    require(
+      ISpoke(spoke).getReserveCount() < _maxReserves[spoke],
+      MaximumReservesReached(spoke, _maxReserves[spoke])
+    );
     return ISpoke(spoke).addReserve(hub, assetId, priceSource, config, dynamicConfig);
   }
 
@@ -254,5 +266,10 @@ contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
     bool active
   ) external onlyOwner {
     ISpoke(spoke).updatePositionManager(positionManager, active);
+  }
+
+  /// @inheritdoc ISpokeConfigurator
+  function getMaxReserves(address spoke) external view returns (uint256) {
+    return _maxReserves[spoke];
   }
 }
--- a/src/spoke/interfaces/ISpokeConfigurator.sol
+++ b/src/spoke/interfaces/ISpokeConfigurator.sol
@@ -8,6 +8,16 @@ import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
 /// @author Aave Labs
 /// @notice Interface for the SpokeConfigurator.
 interface ISpokeConfigurator {
+  /// @notice Emitted when the maximum allowed number of reserves for a spoke is updated.
+  /// @param spoke The address of the spoke.
+  /// @param maxReserves The new maximum number of reserves.
+  event UpdateMaxReserves(address spoke, uint256 maxReserves);
+
+  /// @dev Thrown upon adding a reserve when the maximum allowed number of reserves is already reached.
+  /// @param spoke The address of the spoke.
+  /// @param maxReserves The maximum allowed number of reserves.
+  error MaximumReservesReached(address spoke, uint256 maxReserves);
+
   /// @notice Updates the price source of a reserve.
   /// @dev The price source must implement the AggregatorV3Interface.
   /// @param spoke The address of the spoke.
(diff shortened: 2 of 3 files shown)


snapshots/Spoke.Operations.json
```diff
@@ -1,10 +1,10 @@
 {
   "borrow: first": "257816",
   "borrow: second action, same reserve": "200788",
-  "liquidationCall (receiveShares): full": "262664",
-  "liquidationCall (receiveShares): partial": "290260",
-  "liquidationCall: full": "287411",
-  "liquidationCall: partial": "297607",
+  "liquidationCall (receiveShares): full": "351249",
```
avniculae: what happened here, also below?

avniculae: ah, maybe bc of the 0 rp opt?

yan-man: hm why does the 0 rp opt cause this jump in gas in this PR

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
```diff
@@ -94,7 +94,7 @@ contract LiquidationLogicLiquidateCollateralTest is LiquidationLogicBaseTest {
   }
 
   // hub reverts on remove when collateralToLiquidator is 0
```
yan-man: maybe can expand this to clarify that the hub remove call is skipped then here

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -292,12 +292,14 @@ library LiquidationLogic {
...
     uint256 sharesToLiquidator;
-    if (params.receiveShares) {
-      sharesToLiquidator = hub.previewRemoveByAssets(assetId, params.collateralToLiquidator);
-      positions[params.liquidator][params.collateralReserveId].suppliedShares += sharesToLiquidator
-        .toUint120();
-    } else {
-      sharesToLiquidator = hub.remove(assetId, params.collateralToLiquidator, params.liquidator);
+    if (params.collateralToLiquidator > 0) {
+      if (params.receiveShares) {
+        sharesToLiquidator = hub.previewRemoveByAssets(assetId, params.collateralToLiquidator);
```
yan-man: ah damnit I am thinking we should maybe instead do a round down here instead of round up. otherwise the liquidator is incentivized to do small liquidations with collateral amt < 1 share, essentially taking shares from the liquidation fee.

yan-man: tackled in separate PR: https://github.com/aave/aave-v4/pull/975

snapshots/Spoke.Operations.json
avniculae: got to the bottom of this, it turns out we were never setting liquidation config in spokes in the base setup. this means the target health factor was set to 1, which made the older implementation fall just below under 1 due to the rounding (which we now changed); hence, user was still liquidatable and rp notification was skipped due to 0-0 opt

[8b6ae11](https://github.com/aave/aave-v4/pull/970/commits/8b6ae1199bbb519398232d7787b2f614e3316220)

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
avniculae: [8b6ae11](https://github.com/aave/aave-v4/pull/970/commits/8b6ae1199bbb519398232d7787b2f614e3316220)

tests/unit/Spoke/Spoke.Getters.t.sol
```diff
@@ -10,8 +10,128 @@ contract SpokeGettersTest is SpokeBase {
...
+    );
+    spokeInfo[spoke].usdy.reserveId = spoke.addReserve(
+      address(hub1),
+      usdyAssetId,
+      _deployMockPriceFeed(spoke, 1e8),
+      spokeInfo[spoke].usdy.reserveConfig,
+      spokeInfo[spoke].usdy.dynReserveConfig
+    );
+
+    hub1.addSpoke(wethAssetId, address(spoke), spokeConfig);
```
yan-man, line 124: maybe can do the same thing you did with generalized spoke and set an internal hub var for this test file rather than hub1

snapshots/Spoke.Operations.json
```diff
@@ -1,10 +1,10 @@
 {
   "borrow: first": "258055",
   "borrow: second action, same reserve": "201027",
-  "liquidationCall (receiveShares): full": "312676",
-  "liquidationCall (receiveShares): partial": "312394",
-  "liquidationCall: full": "320005",
-  "liquidationCall: partial": "319723",
+  "liquidationCall (receiveShares): full": "325554",
```
avniculae, line 4: this is full in the sense that it liquidates the maximum allowed. However, we might want to also add a scenario where the entire debt reserve is liquidated, since that also sets flag to 0. should we do it? @miguelmtzinf @DhairyaSethi @yan-man

snapshots/Spoke.Operations.json
```diff
@@ -1,10 +1,10 @@
...
+  "liquidationCall (receiveShares): partial": "312396",
```
avniculae, line 5: user is still liquidatable in this case, which keeps its rp at 0 hence rp 0-0 optimisation kicks in, this is why we have a difference between partial and full

snapshots/Spoke.Operations.json
miguelmtzinf: i though this was fully liquidating the entire debt tbh

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -292,14 +292,16 @@ library LiquidationLogic {
     collateralPosition.suppliedShares = suppliedShares;
 
     uint256 sharesToLiquidator;
-    if (params.receiveShares) {
-      sharesToLiquidator = hub.previewAddByAssets(assetId, params.collateralToLiquidator);
-      if (sharesToLiquidator > 0) {
-        positions[params.liquidator][params.collateralReserveId]
-          .suppliedShares += sharesToLiquidator.toUint120();
+    if (params.collateralToLiquidator > 0) {
```
miguelmtzinf, line 295: this is not similar to repay on behalf @avniculae 
it's just that bonus for the liquidator is zero.

avniculae: not only the bonus is 0, but rather the liquidator does not receive any collateral at all. I still think it's similar to repay on behalf, wdyt?

tests/unit/Spoke/Spoke.Getters.t.sol
avniculae: why? I did that to have the uninitialised liq config in the spoke

snapshots/Spoke.Operations.json
avniculae: let's address this later wdyt https://github.com/aave/aave-v4/issues/983

tests/unit/Spoke/Spoke.Getters.t.sol
yan-man: nbd, just thinking to make it easier to future proof for updating tests later with other hubs

src/spoke/libraries/LiquidationLogic.sol
yan-man: yeah similar, except you can only repay like 1 wei of debt ig

PR #973, opened 2025-10-29
fix: impose max limit on dynamicConfig keys instead of overriding config keys

Kogaroshi:
- Add check to block `dynamicConfigKey` overflow in `addDynamicReserveConfig` & expand `dynamicConfigKey` to `uint24`, allowing up to 16,777,215 unique keys.
- Use `_getReserve()` to verify reserve is listed in spoke for `setUsingAsCollateral()` & `liquidateCall()`

When the new maximum of `dynamicConfigKey` is reached, the 1st ones are not replaced via the overflow logic anymore, but instead Governance is able to update old configKeys, but those historic updates do not allow to set the collateralFactor to 0.

This fix simplifies logic around asset offboarding, so that collateralFactor of 0 cannot be set for config keys that have existing user positions (possible before due to wrapping of config key values), which would block liquidations from occurring.


src/spoke/Spoke.sol
```diff
@@ -164,11 +164,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   ) external restricted returns (uint16) {
     require(reserveId < _reserveCount, ReserveNotListed());
     _validateDynamicReserveConfig(dynamicConfig);
-    uint16 dynamicConfigKey;
+    uint16 dynamicConfigKey = _reserves[reserveId].dynamicConfigKey;
+    require(dynamicConfigKey < type(uint16).max, DynamicConfigKeyOverflow());
     // overflow is desired, we implicitly invalidate & override stale config
     unchecked {
-      dynamicConfigKey = ++_reserves[reserveId].dynamicConfigKey;
+      dynamicConfigKey += 1;
```
miguelmtzinf: let's use `uncheckedAdd`

src/spoke/Spoke.sol
```diff
@@ -164,11 +164,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   ) external restricted returns (uint16) {
     require(reserveId < _reserveCount, ReserveNotListed());
     _validateDynamicReserveConfig(dynamicConfig);
-    uint16 dynamicConfigKey;
+    uint16 dynamicConfigKey = _reserves[reserveId].dynamicConfigKey;
+    require(dynamicConfigKey < type(uint16).max, DynamicConfigKeyOverflow());
```
miguelmtzinf: better to use a constant `MAX_ALLOWED_DYNAMIC_CONFIG_KEY`, and `InvalidDynamicConfigKey` or `MaximumAllowedDynamicConfig` error

src/spoke/Spoke.sol
```diff
@@ -333,8 +335,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     uint256 debtToCover,
     bool receiveShares
   ) external {
-    Reserve storage collateralReserve = _reserves[collateralReserveId];
-    Reserve storage debtReserve = _reserves[debtReserveId];
+    Reserve storage collateralReserve = _getReserve(collateralReserveId);
+    Reserve storage debtReserve = _getReserve(debtReserveId);
```
miguelmtzinf, line 339: remove these checks from LiquidationLogic.validate fn

DhairyaSethi: these checks were already redundant bc at the start of liquidation we do previewRemove & previewRestore on both these hubs

src/spoke/Spoke.sol
```diff
@@ -38,6 +38,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   /// @inheritdoc ISpoke
   uint24 public constant MAX_ALLOWED_COLLATERAL_RISK = 1000_00; // 1000.00%
 
+  /// @inheritdoc ISpoke
+  uint256 public constant MAX_ALLOWED_DYNAMIC_CONFIG_KEY = type(uint24).max;
```
DhairyaSethi, line 42: these constants now just eating up the spoke code size, might have to just eventually make these internal to save on public getter code

src/spoke/Spoke.sol
DhairyaSethi, line 339: not a fan of these checks bc they are redundant (since we already fetch balances at the start of liquidation flow through both hubs) and we can save the gas here

src/spoke/Spoke.sol
Kogaroshi: agreed, tho right now we can afford the space, they should be made internal if we need too later (for via-IR opt for exemple).

src/spoke/interfaces/ISpoke.sol
```suggestion
  /// @dev Appends dynamic config to the next available key; reverts if `MAX_ALLOWED_DYNAMIC_CONFIG_KEY` is reached.
```

src/spoke/Spoke.sol
```diff
@@ -164,12 +164,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   function addDynamicReserveConfig(
     uint256 reserveId,
     DynamicReserveConfig calldata dynamicConfig
-  ) external restricted returns (uint16) {
+  ) external restricted returns (uint24) {
     require(reserveId < _reserveCount, ReserveNotListed());
     _validateDynamicReserveConfig(dynamicConfig);
-    uint16 dynamicConfigKey = _reserves[reserveId].dynamicConfigKey;
+    uint24 dynamicConfigKey = _reserves[reserveId].dynamicConfigKey;
     require(dynamicConfigKey < MAX_ALLOWED_DYNAMIC_CONFIG_KEY, MaximumAllowedDynamicConfig());
```
miguelmtzinf: move this one line up

src/spoke/interfaces/ISpoke.sol
miguelmtzinf: check if there is any comment about overriding existing config keys anywhere else pls

src/spoke/Spoke.sol
miguelmtzinf: how much we d save by removing?

Kogaroshi: -370 gas on liquidationCall

src/spoke/interfaces/ISpoke.sol
```diff
@@ -287,6 +286,9 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen.
   error CannotReceiveShares();
 
+  /// @notice Thrown when trying to add a dynamic config but the key would overflow.
+  error MaximumAllowedDynamicConfig();
```
```suggestion
  /// @notice Thrown when the maximum number of dynamic config keys is reached.
  error MaximumDynamicConfigKeyReached();
```

src/spoke/Spoke.sol
miguelmtzinf: let's keep them, then

src/spoke/Spoke.sol
miguelmtzinf: @Kogaroshi let's move this up before `_validate` pls

PR #974, opened 2025-10-29
feat: Limit number of spoke reserves via spoke configurator

CheyenneAtapour:
* Enforces a limit on the spoke configurator such that when a reserve is added to a spoke, the operation reverts if that additional reserve would exceed the allowed limit on the spoke
* The owner of the spoke configurator can configure the limit of reserves per each spoke

The motivation for this is that if a reserve has an arbitrary number of spokes (and we do not enforce limits on numbers of collateral on the user level), a user can enable so many collaterals that it causes any operations on their position to fail because they exceed gas limits. The main cause of this is the sorting algorithm involved in calculating a user's risk premium. Indeed, analysis shows that 166 collaterals is the limit where calculating risk premium succeeds, but 167 fails. The other operations such as supplying, borrowing, or liquidating will still succeed and are only in the realm of 17m gas, vs the limit of 45m. The analysis was performed for the worst case of the quicksort algorithm, in which the collaterals are already in sorted order (of collateral risk).

Such analysis should be performed on each spoke implementation to determine the appropriate reserve limit for each spoke

closes https://github.com/aave/aave-v4/issues/734

### DIFF
--- a/src/spoke/SpokeConfigurator.sol
+++ b/src/spoke/SpokeConfigurator.sol
@@ -73,7 +73,7 @@ contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
   /// @inheritdoc ISpokeConfigurator
   function updateMaxReserves(address spoke, uint256 maxReserves) external onlyOwner {
     _maxReserves[spoke] = maxReserves;
-    emit MaxReservesUpdated(spoke, maxReserves);
+    emit UpdateMaxReserves(spoke, maxReserves);
   }
 
   /// @inheritdoc ISpokeConfigurator
--- a/src/spoke/interfaces/ISpokeConfigurator.sol
+++ b/src/spoke/interfaces/ISpokeConfigurator.sol
@@ -11,7 +11,7 @@ interface ISpokeConfigurator {
   /// @notice Emitted when the maximum allowed number of reserves for a spoke is updated.
   /// @param spoke The address of the spoke.
   /// @param maxReserves The new maximum number of reserves.
-  event MaxReservesUpdated(address spoke, uint256 maxReserves);
+  event UpdateMaxReserves(address spoke, uint256 maxReserves);
 
   /// @dev Thrown upon adding a reserve when the maximum allowed number of reserves is already reached.
   /// @param spoke The address of the spoke.
--- a/tests/unit/SpokeConfigurator.t.sol
+++ b/tests/unit/SpokeConfigurator.t.sol
@@ -167,7 +167,7 @@ contract SpokeConfiguratorTest is SpokeBase {
   function test_updateMaxReserves() public {
     uint256 newMaxReserves = vm.randomUint();
     vm.expectEmit(address(spokeConfigurator));
-    emit ISpokeConfigurator.MaxReservesUpdated(spokeAddr, newMaxReserves);
+    emit ISpokeConfigurator.UpdateMaxReserves(spokeAddr, newMaxReserves);
     vm.prank(SPOKE_CONFIGURATOR_ADMIN);
     spokeConfigurator.updateMaxReserves(spokeAddr, newMaxReserves);
 


src/spoke/SpokeConfigurator.sol
```diff
@@ -14,6 +14,8 @@ import {ISpokeConfigurator} from 'src/spoke/interfaces/ISpokeConfigurator.sol';
 contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
   using SafeCast for uint256;
 
+  mapping(address => uint256) internal _reserveLimit;
```
yan-man: if it's internal perhaps we should also create a getter for this?

src/spoke/SpokeConfigurator.sol
```diff
@@ -68,6 +70,11 @@ contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
     ISpoke(spoke).updateLiquidationConfig(liquidationConfig);
   }
 
+  /// @inheritdoc ISpokeConfigurator
+  function updateReserveLimit(address spoke, uint256 limit) external onlyOwner {
+    _reserveLimit[spoke] = limit;
```
yan-man: maybe also emit an event here?

should we check limit > 0? hm ig limit=0 is valid 

and shouldnt be an issue, but also then we're ok with setting limit < current reserveCount too

src/spoke/interfaces/ISpokeConfigurator.sol
```diff
@@ -8,6 +8,11 @@ import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
 /// @author Aave Labs
 /// @notice Interface for the SpokeConfigurator.
 interface ISpokeConfigurator {
+  /// @dev Thrown when the reserve limit for a spoke is reached.
+  /// @param spoke The address of the spoke.
+  /// @param limit The reserve limit.
```
yan-man: maybe can clarify here that limit is inclusive

should we rename to cap vs limit for consistency as well?

tests/unit/SpokeConfigurator.t.sol
```diff
@@ -180,7 +180,40 @@ contract SpokeConfiguratorTest is SpokeBase {
     });
   }
 
+  function test_addReserve_revertsWith_ReserveLimitReached() public {
+    vm.prank(SPOKE_CONFIGURATOR_ADMIN);
+    spokeConfigurator.updateReserveLimit(spokeAddr, 0);
```
yan-man: can you add another test scenario for a non-zero value

src/spoke/interfaces/ISpokeConfigurator.sol
CheyenneAtapour: We don't specify inclusive for the other caps.

I like limit because cap can be confusing with supply or draw cap

src/spoke/SpokeConfigurator.sol
CheyenneAtapour: Emitted event

tests/unit/SpokeConfigurator.t.sol
CheyenneAtapour: Updated this test to use potentially non-zero values (using random uint)

src/spoke/SpokeConfigurator.sol
CheyenneAtapour: Added

miguelmtzinf:
```suggestion
  mapping(address spoke => uint256) internal _maxNumberReserves;
```

CheyenneAtapour: Updated, but @miguelmtzinf did you want the getters and everything to change names as well? I kinda like the name as is for the external functions etc.

Reason being `maxNumberReserves` is not proper grammar, because it should be `maxNumberOfReserves`, or we could just do `maxReserves`?

src/spoke/SpokeConfigurator.sol
```diff
@@ -68,6 +70,12 @@ contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
     ISpoke(spoke).updateLiquidationConfig(liquidationConfig);
   }
 
+  /// @inheritdoc ISpokeConfigurator
+  function updateReserveLimit(address spoke, uint256 limit) external onlyOwner {
```
yan-man: maybe these names should be updated to reflect new var name, ie updateMaxNumberReserves or something

src/spoke/interfaces/ISpokeConfigurator.sol
yan-man: maybe also update the natspec to reflect new var naming too

yan-man:
btw @CheyenneAtapour also closes this one: https://github.com/aave/aave-v4/issues/734

src/spoke/SpokeConfigurator.sol
CheyenneAtapour: Done, I went with just `maxReserves`

src/spoke/interfaces/ISpokeConfigurator.sol
CheyenneAtapour: Done, I went with just `maxReserves`

CheyenneAtapour:
I kinda like `limit` better personally, but nbd

src/spoke/interfaces/ISpokeConfigurator.sol
```diff
@@ -38,6 +48,11 @@ interface ISpokeConfigurator {
     ISpoke.LiquidationConfig calldata liquidationConfig
   ) external;
 
+  /// @notice Updates the maximum number of reserves allowed to exist on a spoke.
+  /// @param spoke The address of the spoke.
```
miguelmtzinf, line 53:
```suggestion
  /// @dev It allows to set the maximum below the amount of reserves that currently exist.
  /// @param spoke The address of the spoke.
```

src/spoke/interfaces/ISpokeConfigurator.sol
```diff
@@ -8,6 +8,16 @@ import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
...
 interface ISpokeConfigurator {
+  /// @notice Emitted when the maximum allowed number of reserves for a spoke is updated.
+  /// @param spoke The address of the spoke.
+  /// @param maxReserves The new maximum number of reserves.
+  event MaxReservesUpdated(address spoke, uint256 maxReserves);
+
+  /// @dev Thrown upon adding a reserve when the maximum allowed number of reserves is already reached.
+  /// @param spoke The address of the spoke.
+  /// @param maxReserves The maximum allowed number of reserves.
+  error MaxReservesReached(address spoke, uint256 maxReserves);
```
miguelmtzinf:
```suggestion
  error MaximumReservesReached(address spoke, uint256 maxReserves);
```

CheyenneAtapour: How about the name of the event? Is that fine?

src/spoke/interfaces/ISpokeConfigurator.sol
CheyenneAtapour: Changed, I worded it slightly differently

src/spoke/interfaces/ISpokeConfigurator.sol
```diff
@@ -8,6 +8,16 @@ import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
 /// @author Aave Labs
 /// @notice Interface for the SpokeConfigurator.
 interface ISpokeConfigurator {
+  /// @notice Emitted when the maximum allowed number of reserves for a spoke is updated.
+  /// @param spoke The address of the spoke.
+  /// @param maxReserves The new maximum number of reserves.
+  event MaxReservesUpdated(address spoke, uint256 maxReserves);
```
miguelmtzinf:
```suggestion
  event UpdateMaxReserves(address spoke, uint256 maxReserves);
```
we are doing present tense verb + noum

CheyenneAtapour: Done

PR #975, opened 2025-10-30
fix: round down on `sharesToLiquidator`

yan-man:
- when liquidator chooses to `receiveShares`, add fix to round down `sharesToLiquidator` to be in favor of feeReceiver/protocol. In `receiveShares` case, remove is skipped and all shares accounting remains in Spoke, as shares are directly transferred from user to liquidator. This fixed rounding in essence will favor the liquidationFee shares which are calculated afterwards as `(sharesToLiquidate - sharesToLiquidator)` at the expense of the liquidator
- by rounding up (as is), liquidator is incentivized to split liquidations into multiple small amounts and receiveShares - at its most profitable, liquidating 1 wei of collateral each time to round up to 1 share, unfairly rewarding the liquidator at the expense of liquidationFee.

### DIFF
--- a/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
@@ -994,20 +994,11 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
     );
 
     // Fee Receivers
-    if (!params.receiveShares) {
-      assertEq(
-        accountsInfoAfter.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
-        accountsInfoBefore.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
-        'collateral fee receiver: collateral supplied'
-      );
-    } else {
-      assertApproxEqAbs(
-        accountsInfoAfter.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
-        accountsInfoBefore.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
-        1,
-        'collateral fee receiver: collateral supplied (receivedShares)'
-      );
-    }
+    assertEq(
+      accountsInfoAfter.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
+      accountsInfoBefore.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
+      'collateral fee receiver: collateral supplied'
+    );
 
     assertEq(
       accountsInfoAfter.collateralFeeReceiverBalanceInfo.borrowedFromSpoke,


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -293,9 +293,11 @@ library LiquidationLogic {
...
     uint256 sharesToLiquidator;
     if (params.receiveShares) {
-      sharesToLiquidator = hub.previewRemoveByAssets(assetId, params.collateralToLiquidator);
-      positions[params.liquidator][params.collateralReserveId].suppliedShares += sharesToLiquidator
-        .toUint120();
+      sharesToLiquidator = hub.previewAddByAssets(assetId, params.collateralToLiquidator);
+      if (sharesToLiquidator > 0) {
+        positions[params.liquidator][params.collateralReserveId]
+          .suppliedShares += sharesToLiquidator.toUint120();
+      }
```
miguelmtzinf, line 300:
```
user.suppliedShares = 10
liquidator.suppliedShares = 0
sharePrice 5

collateralToLiquidate 14
collateralToLiquidator 7

- sharesToLiquidate = (14/5).up = 3
- user.suppliedShares = 10 - 3 = 7

--- 2 paths:

-- A) remove
- remove 7 assets, 2 shares
- sharesToLiquidator = 2
- hub: hub.liquidity -= 7, spoke.addedShares -= 2
- liquidator gets 7 assets
- payFeeShares sharesToLiquidate - sharesToLiquidator = 3-2 = 1
- hub: spoke.addedShares -=1, feeReceiver.addedShares += 1

so:
- user.suppliedShares = 7
- liquidator.suppliedShares = 0
- liquidator earns 7 assets -> if supplied: 1 shares
- feeReceiver.addedShares += 1 -> if withdrawn: 5 assets
- spoke.addedShares -= 2
- hub.liquidity -= 7
- if fees withdrawn: hub.liquidity -= 12, feeReceiver.addedShares = 0, spoke.addedShares -= 3
	so 3 assets are in Hub (3*5 -12 = 3), with no spoke/user assigned -> donation.


--B) receiveShares
- sharesToLiquidator = (7/5).down = 1
- liquidator.addedShares = 1
- hub: no change
- payFeeShares sharesToLiquidate - sharesToLiquidator = 3-1 = 2
- hub: spoke.addedShares -=2, feeReceiver.addedShares += 2

so:
- user.suppliedShares = 7
- liquidator.addedShares = 1
- liquidator earns 5 assets if withdrawn
- hub: spoke.addedShares -=2, feeReceiver.addedShares += 2
- hub.liquidity no change
- if liquidator withdraws: liquidator.addedShares = 0, spoke.addedShares -= 3, feeReceiver.addedShares += 2, hub.liquidity -= 5
- if liquidator and fees withdraws:  liquidator.addedShares = 0, spoke.addedShares -= 3, feeReceiver.addedShares = 0, hub.liquidity -= 15
```

The remove action does not make the most of the shares (donation given to suppliers) as receiveShares does, but seems fine.
```
remove: liquidator gets 7 assets, fees are 5 assets (12 assets, 3 shares)
receiveShares: liquidator gets 5 assets, fees are 10 assets (15 assets, 3 shares)
```

yan-man: for ex 1)
> 15 assets (and we only had 14, so 1 d be minted out of thin air)

i think I'd slightly look at this differently; rather, that 15 assets is the entitled amount to withdraw from hub. This to me is the "correct" default value, and giving the 14 is instead short-changing the liquidator by 1 asset, which gets implicitly donated to total assets in hub by increasing the ex rate from 5 -> (50-14 assets)/(10-3 shares) -> 36/7 -> 5.14 new share price. ie we are allowed to remove up to 15 assets without messing up accounting, but we just round in favor of protocol to end up at 14. 

so in the receive shares case, the protocol simply doesnt benefit from the 1 wei donation (and stays at share price 5)

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -293,9 +293,11 @@ library LiquidationLogic {
...
     } else {
       sharesToLiquidator = hub.remove(assetId, params.collateralToLiquidator, params.liquidator);
```
avniculae, line 302: should we add a comment here to mention that these are not necessarily shares sent to the liquidator? these are rather shares burnt for the asset amount sent to the liquidator

src/spoke/libraries/LiquidationLogic.sol
avniculae: > (and we only had 14, so 1 d be minted out of thin air).

I don't think these would be "minted" out of thin air. We decrease user's supplies shares by 3 shares anyways (worth 15 assets), so the 3 shares we give to the liquidator are fully backed. Rather, I think the problem here is the incentivisation of liquidators to split liquidations into multiple small ones, since the additional 1 share received by liquidators will make liquidators steal treasury fees.

One more argument why rounding up would never end up "minting" shares out of thin air is the fact that converting user's supplied shares into withdrawable assets, and then back to removed shares will always be equal to the supplies shares of the user. See this script here (I think it might be worth adding it in this PR, wdyt?)


```
# Highlights the fact that supplies shares are always equal to removed shares (after doing the conversion to assets and back to shares).
from z3 import *

WAD = IntVal(10**18)
VIRTUAL_SHARES = IntVal(10**6)
VIRTUAL_ASSETS = IntVal(10**6)

def mulDivDown(a, num, den):
    return (a * num) / den

def mulDivUp(a, num, den):
    return (a * num + den - 1) / den

def previewRemoveByShares(shares, totalAddedAssets, totalAddedShares):
    return mulDivDown(shares, totalAddedAssets + VIRTUAL_ASSETS, totalAddedShares + VIRTUAL_SHARES)

def previewRemoveByAssets(assets, totalAddedAssets, totalAddedShares):
    return mulDivUp(assets, totalAddedShares + VIRTUAL_SHARES, totalAddedAssets + VIRTUAL_ASSETS)

def check(propertyDescription):
    print(f"\n-- {propertyDescription} --")
    result = s.check()
    if result == sat:
        print("Counterexample found:")
        print(s.model())
    elif result == unsat:
        print(f"Property holds.")
    elif result == unknown:
        print("Timed out or unknown.")

s = Solver()

totalAddedAssets = Int('totalAddedAssets')
s.add(0 <= totalAddedAssets, totalAddedAssets <= 10**30)
totalAddedShares = Int('totalAddedShares')
s.add(totalAddedAssets >= totalAddedShares, totalAddedAssets + VIRTUAL_ASSETS < (totalAddedShares + VIRTUAL_SHARES) * 100)
suppliedShares = Int('suppliedShares')
s.add(0 <= suppliedShares, suppliedShares <= totalAddedShares)

withdrawableAssets = previewRemoveByShares(suppliedShares, totalAddedAssets, totalAddedShares)
removedShares = previewRemoveByAssets(withdrawableAssets, totalAddedAssets, totalAddedShares)

s.add(removedShares != suppliedShares)
check("Supplies shares are always equal to removed shares (after doing the conversion to assets and back to shares).")
```

tests/mocks/LiquidationLogicWrapper.sol
avniculae, line 27: I think we can remove this params from here right?

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
avniculae: should we replace this params assignments with an explicit args
```
= Params({...})
```
to showcase that receiveShares is false here?

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
```diff
@@ -93,6 +96,71 @@ contract LiquidationLogicLiquidateCollateralTest is LiquidationLogicBaseTest {
...
+  /// on receiveShares, sharesToLiquidator should round down
+  function test_liquidateCollateral_receiveShares_sharesToLiquidatorIsZero() public {
+    // increase reserve index to ensure sharesToLiquidator rounds to 0 while feeShares rounds up to 1
+    _increaseReserveIndex(spoke1, reserveId);
+
+    // supply ex rate is between 1 and 2
+    assertGt(hub.previewAddByShares(assetId, WadRayMath.RAY), WadRayMath.RAY);
+    assertLt(hub.previewAddByShares(assetId, WadRayMath.RAY), 2 * WadRayMath.RAY);
+
+    params.collateralToLiquidate = 1;
```
avniculae: same comment as above (use explicit args)

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
```diff
@@ -93,6 +96,71 @@ contract LiquidationLogicLiquidateCollateralTest is LiquidationLogicBaseTest {
...
+
+    assertEq(sharesToLiquidate, 1);
+    assertEq(sharesToLiquidator, 0);
+    assertEq(feeShares, 1);
+
+    vm.expectCall(address(hub), abi.encodeCall(IHubBase.payFeeShares, (assetId, feeShares)), 1);
+    liquidationLogicWrapper.liquidateCollateral(params);
+
+    // sharesToLiquidator should round to 0 and remain unchanged
+    assertPosition(
```
avniculae, line 136: let's also assert user position here, and treasury spoke shares

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
```diff
@@ -93,6 +96,71 @@ contract LiquidationLogicLiquidateCollateralTest is LiquidationLogicBaseTest {
...
+
+    uint256 sharesToLiquidate = hub.previewRemoveByAssets(assetId, params.collateralToLiquidate);
+    uint256 sharesToLiquidator = hub.previewAddByAssets(assetId, params.collateralToLiquidator);
+    uint256 feeShares = sharesToLiquidate - sharesToLiquidator;
+
+    vm.expectCall(address(hub), abi.encodeCall(IHubBase.payFeeShares, (assetId, feeShares)), 1);
+    liquidationLogicWrapper.liquidateCollateral(params);
+
+    // sharesToLiquidator should round to 0 and remain unchanged
+    assertPosition(
```
avniculae, line 183: same comment as above (assert user position and treasury spoke shares)

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
```diff
@@ -93,6 +96,71 @@ contract LiquidationLogicLiquidateCollateralTest is LiquidationLogicBaseTest {
...
+    params.receiveShares = true;
+    params.user = borrower;
+
+    uint256 sharesToLiquidate = hub.previewRemoveByAssets(assetId, params.collateralToLiquidate);
+    uint256 sharesToLiquidator = hub.previewAddByAssets(assetId, params.collateralToLiquidator);
+    uint256 feeShares = sharesToLiquidate - sharesToLiquidator;
+
+    assertEq(sharesToLiquidate, 1);
+    assertEq(sharesToLiquidator, 0);
+    assertEq(feeShares, 1);
```
avniculae, line 130: I'd remove these given https://github.com/aave/aave-v4/pull/975/files#r2486464364, wdyt?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
```diff
@@ -992,11 +994,21 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
...
+      assertEq(
+        accountsInfoAfter.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
+        accountsInfoBefore.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
+        'collateral fee receiver: collateral supplied'
+      );
+    } else {
+      assertApproxEqAbs(
+        accountsInfoAfter.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
+        accountsInfoBefore.collateralFeeReceiverBalanceInfo.suppliedInSpoke,
+        1,
```
avniculae: why? this should not happen right? the fee receiver does not supply anything in the liquidated spoke correct?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
```diff
@@ -1079,19 +1091,30 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
...
+        accountsInfoAfter.collateralFeeReceiverBalanceInfo.addedInHub,
+        accountsInfoBefore.collateralFeeReceiverBalanceInfo.addedInHub +
+          liquidationMetadata.collateralToLiquidate -
+          liquidationMetadata.collateralToLiquidator,
+        _approxRelFromBps(1),
+        'collateral fee receiver: added'
+      );
+    } else {
+      // collateral rounded down on receiveShares, can differ by 1 share worth of assets
+      uint256 absDiff = _hub(params.spoke, params.collateralReserveId).previewRemoveByAssets(
```
avniculae: this will always be 1 no? 1 asset into shares (up) is always equal to 1, no matter the share price

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: yeah, you are right, updated the example. Most important part here is that share price is respected, and updates are consistent with spoke side (same number of shares).

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
avniculae: also, why do we need this if else here? can we not adjust the discrepancy to work for both cases?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
```diff
@@ -1079,19 +1091,30 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
...
     // Fee Receivers
-    assertApproxEqRel(
-      accountsInfoAfter.collateralFeeReceiverBalanceInfo.addedInHub,
-      accountsInfoBefore.collateralFeeReceiverBalanceInfo.addedInHub +
-        liquidationMetadata.collateralToLiquidate -
-        liquidationMetadata.collateralToLiquidator,
-      _approxRelFromBps(1),
-      'collateral fee receiver: added'
-    );
-    assertEq(
```
avniculae, line 1090: why was this removed?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
avniculae: also, absDiff is in share terms but comparison below is in asset terms, this is not right I think

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: idk it's needed.. it's a donation, like on any other action

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
yan-man: via payFeeShares feeReceiver should receive supplied shares?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
yan-man: this got lost when adding the conditional assertion, added back

https://github.com/aave/aave-v4/pull/975/commits/686e4514d5e771118db16bc3a65fa25304347c78

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
yan-man: good catch, updated. 

> can we not adjust the discrepancy to work for both cases?

perhaps but would prefer to split out this assertion to 1) explicitly account for differences in the 2 situations, 2) it would need to expand the approxEqRel bounds in an annoying way, ex. small values like ex. 1233 vs 1232 which are off by 1 wei in abs terms will fail due to the bound not being within rel(1) (ie 0.01%)

tests/mocks/LiquidationLogicWrapper.sol
yan-man: could but i like that it gives a way to init the wrapper w the borrower/liquidator without needing to call the setters after

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
yan-man: prefer to keep them to be explicit for this scenario

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
yan-man: https://github.com/aave/aave-v4/pull/975/commits/686e4514d5e771118db16bc3a65fa25304347c78

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
yan-man: https://github.com/aave/aave-v4/pull/975/commits/686e4514d5e771118db16bc3a65fa25304347c78

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
yan-man: https://github.com/aave/aave-v4/pull/975/commits/686e4514d5e771118db16bc3a65fa25304347c78

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
yan-man: https://github.com/aave/aave-v4/pull/975/commits/686e4514d5e771118db16bc3a65fa25304347c78

src/spoke/libraries/LiquidationLogic.sol
avniculae: yes, it's just that the naming (`sharesToLiquidator`) is a bit confusing in this context, nbd

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
avniculae: that is on the hub level no?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
```diff
@@ -1101,17 +1106,12 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
...
-        _assetId(params.spoke, params.collateralReserveId),
-        1
-      );
       assertApproxEqAbs(
         accountsInfoAfter.collateralFeeReceiverBalanceInfo.addedInHub,
         accountsInfoBefore.collateralFeeReceiverBalanceInfo.addedInHub +
           liquidationMetadata.collateralToLiquidate -
           liquidationMetadata.collateralToLiquidator,
-        absDiff,
+        1,
```
avniculae, line 1105: can we maybe use an inline if for the discrepancy?

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
```diff
@@ -65,14 +67,19 @@ contract LiquidationLogicLiquidateCollateralTest is LiquidationLogicBaseTest {
...
-      1,
-      hub.previewRemoveByShares(assetId, suppliedShares)
-    );
+    params = LiquidationLogic.LiquidateCollateralParams({
+      collateralToLiquidate: bound(
+        collateralToLiquidate,
+        1,
+        hub.previewRemoveByShares(assetId, userSuppliedShares)
+      ),
+      collateralToLiquidator: 0, // populated below
```
avniculae, line 76: why not here?

tests/unit/Spoke/Spoke.ReceiveShares.t.sol
```diff
@@ -0,0 +1,48 @@
...
+    _openDebtPosition(spoke1, _wbtcReserveId(spoke1), MAX_SUPPLY_AMOUNT / 2, true);
+    _openDebtPosition(spoke1, _usdxReserveId(spoke1), MAX_SUPPLY_AMOUNT / 2, true);
+    _openDebtPosition(spoke1, _usdyReserveId(spoke1), MAX_SUPPLY_AMOUNT / 2, true);
+    _openDebtPosition(spoke1, _daiReserveId(spoke1), MAX_SUPPLY_AMOUNT / 2, true);
+
+    skip(1);
+  }
+
+  function test_spoke_receive_shares() public {
+    console.log('supply ex rate %e', hub1.previewRemoveByShares(wbtcAssetId, MAX_SUPPLY_AMOUNT));
```
avniculae: is this WIP?

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
avniculae: I was thinking to hardcode these numbers in the assertPosition checks below, nbd though

tests/unit/Spoke/Spoke.ReceiveShares.t.sol
yan-man: ah damn didnt mean to push this, was just a test file I was messing with

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidateCollateral.t.sol
yan-man: bc it depends on params.collateralToLiquidate, the bounded value which isn't set till after this line

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
yan-man: do you mean like `receiveShares?1:0` or similar?

can't bc one is approxEqAbs vs assertApproxEqRel

avniculae: right, thanks!

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
yan-man: gotcha, good catch! yes you're right didn't need to change this assertion 

https://github.com/aave/aave-v4/pull/975/commits/d3c1dd71607c90f0f56eb7328822b267c3470355

PR #976, opened 2025-10-30
feat: Emit oracle address in Spoke initialize function

CheyenneAtapour:
Emits the address of the oracle when the spoke is initialized

### DIFF
--- a/src/spoke/interfaces/ISpoke.sol
+++ b/src/spoke/interfaces/ISpoke.sol
@@ -115,8 +115,8 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     uint256 borrowedCount;
   }
 
-  /// @notice Emitted when a spoke is initialized and the oracle is set.
-  /// @param oracle The address of the oracle.
+  /// @notice Emitted when the oracle address of the spoke is updated.
+  /// @param oracle The new address of the oracle.
   event UpdateOracle(address oracle);
 
   /// @notice Emitted when a liquidation config is updated.


src/spoke/Spoke.sol
```diff
@@ -80,7 +80,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   }
 
   /// @dev To be overridden by the inheriting Spoke instance contract.
-  function initialize(address authority) external virtual;
+  function initialize(address authority) public virtual {
+    emit OracleSet(ORACLE);
```
miguelmtzinf: let's revert this change and emit in the initialize. Having all logic in one single place vs 2 is better imho

CheyenneAtapour: You mean just in the spoke instance?

src/spoke/interfaces/ISpoke.sol
```diff
@@ -115,6 +115,10 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     uint256 borrowedCount;
   }
 
+  /// @notice Emitted when a spoke is initialized and the oracle is set.
```
yan-man: imo don't need to specify it is when spoke is initialized

src/spoke/instances/SpokeInstance.sol
```diff
@@ -21,6 +21,7 @@ contract SpokeInstance is Spoke {
   /// @dev The authority contract must implement the `AccessManaged` interface for access control.
   /// @param authority The address of the authority contract which manages permissions.
   function initialize(address authority) external override reinitializer(SPOKE_REVISION) {
+    emit OracleSet(ORACLE);
```
miguelmtzinf: what if we do `InitializeSpoke` insteadd at the end?

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: mmm, but I like it since that's the only place where this is emitted

src/spoke/instances/SpokeInstance.sol
CheyenneAtapour: Decided to leave as is

src/spoke/interfaces/ISpoke.sol
```suggestion
  /// @notice Emitted when the oracle address of the spoke is updated.
```

src/spoke/interfaces/ISpoke.sol
```diff
@@ -115,6 +115,10 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
...
+  /// @param oracle The address of the oracle.
```
```suggestion
  /// @param oracle The new address of the oracle.
```

src/spoke/interfaces/ISpoke.sol

src/spoke/interfaces/ISpoke.sol

PR #977, opened 2025-10-30
docs: Protocol's tech docs

CanonicalJP:
Adds the protocol tech docs

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -84,8 +84,6 @@ This parameter is configurable and part of the Spoke's risk parameters. This mea
 
 $CR_i$ is the Collateral Risk of the asset $i$
 
-> 💡 This parameter only applies if an asset is utilized as collateral.
-
 ### User Risk Premium
 
 The User Risk Premium $RP_u$ represents the quality of assets used as collateral by the user $u$ to borrow against. It depends on multiple dynamic parameters:
@@ -273,7 +271,7 @@ The lower the HF of a position becomes, the larger the bonus the liquidators rec
 
 Once a user becomes liquidatable, the protocol offers a minimum liquidation bonus equal to
 
-$$ minLB = (maxLB - 100\%) \times lbFactor + 100\% $$
+$$ minLB = (maxLB - 100\\%) \times lbFactor + 100\\% $$
 
 where
 
@@ -292,44 +290,5 @@ $$
 where
 
 - $HF\_LIQ\_THRESHOLD$: per Spoke. Represents the health factor threshold under which the user becomes liquidatable. Equals 1.
-- $hf_{beforeLiq}$: per user. Represents the user’s health factor before liquidation. Exact formula detailed below.
+- $hf_{beforeLiq}$: per user. Represents the user’s health factor before liquidation.
 - $hfForMaxBonus$: per Spoke. Represents the health factor threshold under which the protocol awards the maximum liquidation bonus.
-
-### Implications of the Dutch-auction mechanism
-
-Awarding a higher liquidation bonus as the health factor decreases introduces a potential issue: if a user’s health factor decreases due to liquidation, liquidators are incentivized to split the liquidation into multiple smaller parts to obtain higher liquidation bonuses. This increases the chances of the user creating bad debt to the system and affects the overall user experience.
-
-### When does health factor decrease after liquidation?
-
-$$
-\begin{align*}
-& hf_{beforeLiq} = \frac{C}{D} = \frac{\sum_{i=1}^n c_i \times cf_i}{D} \\
-& hf_{afterLiq} = \frac{\left(\sum_{i=1,i\neq k}^n c_i \times cf_i\right) + (c_k - l \times lb_k) \times cf_k}{D - l}
-\end{align*}
-$$
-
-$$
-\begin{align*}
-& hf_{afterLiq} < hf_{beforeLiq} \\
-& \equiv \frac{\left(\sum_{i=1,i\neq k}^n c_i \times cf_i\right) + (c_k - l \times lb_k) \times cf_k}{D - l} < \frac{C}{D} \\
-& \equiv \left[\left(\sum_{i=1,i\neq k}^n c_i \times cf_i\right) + (c_k - l \times lb_k) \times cf_k\right] \times D < C \times (D - l) \\
-& \equiv \left[\left(\sum_{i=1,i\neq k}^n c_i \times cf_i\right) + c_k \times cf_k - l \times lb_k \times cf_k\right] \times D < D \times C - l \times C \\
-& \equiv \left[\left(\sum_{i=1}^n c_i \times cf_i\right) - l \times lb_k \times cf_k\right] \times D < D \times C - l \times C \\
-& \equiv D \times C - D \times l \times lb_k \times cf_k < D \times C - l \times C \\
-& \equiv D \times l \times lb_k \times cf_k > l \times C \\
-& \equiv D \times lb_k \times cf_k > C, \text{since } l > 0 \\
-& \equiv lb_k \times cf_k > \frac{C}{D}, \text{since } D > 0 \\
-& \equiv lb_k \times cf_k > hf_{beforeLiq}
-\end{align*}
-$$
-
-where
-
-- $C$ is the user’s total collateral value in USD, weighted by collateral factors.
-- $c_i$ is the user’s $i$-th collateral value in USD: the supplied amount of the user’s $i$-th collateral multiplied by the collateral’s price.
-- $cf_i$ is the user’s $i$-th collateral factor. Note that this can differ from the latest collateral factor of that specific collateral, given that liquidations do not refresh users’ dynamic configurations.
-- $D$ is the user’s total debt value in USD.
-- $l$ is the amount of debt liquidated in USD.
-- $lb_k$ is the liquidation bonus determined by the Dutch-auction algorithm described above, based on user’s dynamic configuration key for the $k$-th collateral.
-- $hf_{beforeLiq}$ is the user’s health factor before the liquidation.
-- $hf_{afterLiq}$ is the user’s health factor after the liquidation.


docs/overview.md
```diff
@@ -0,0 +1,295 @@
+# Untitled
```
yan-man: why untitled

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+# Overview
+
+Aave V4 introduces an architectural redesign centered around the Liquidity Hub, enabling protocol flexibility and efficiency. This innovative architecture allows the Governor (e.g. the Aave DAO) to dynamically manage Spokes, adding new borrowing capabilities and removing outdated ones without requiring costly liquidity migrations.
+
+The protocol implements sophisticated risk management through its Liquidity Premiums system, where each asset receives a dynamic risk factor (called collateral risk) ranging from 0.00 to 1000.00 (BPS) based on asset's implied volatility, market conditions, liquidity, risk, etc. This granular pricing mechanism introduces base rates for high-quality collaterals (such as ETH) while adjusting borrowing costs proportionally to risk profiles.
```
yan-man: - maybe capitalize Collateral Risk
- as it stands the `0.00 to 1000.00` is in %, it should be `0 to 100000` for BPS
- `proportionally to risk profiles` -> maybe can clarify that this is collateral risk profile (but i see its mentioned in the next paragraph)

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+By providing preferential rates for stronger collaterals and optimizing capital efficiency, Aave V4 creates a more robust lending environment that accurately prices risk and rewards. Consequently, attracting higher-quality collateral while offering improved yields for suppliers and lower fees for borrowers utilizing safer assets.
```
yan-man: `Consequently, attracting higher-quality collateral while offering improved yields for suppliers and lower fees for borrowers utilizing safer assets.`
- grammar, this is not a full sentence. Maybe "... this attracts higher-quality..."
- maybe `utilizing safer collateral assets`

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+  S1 --> HUB
+  S2 --> HUB
+  S3 --> HUB
+
+  %% Styling (optional)
+  classDef box fill:#0b0b0d,stroke:#e5e5e5,stroke-width:1px,color:#ffffff;
+  class LP,BR,S1,S2,S3,HUB box;
+```
+
+Spokes are individual modules that can connect to one or multiple LHs. They route user actions (supply/withdraw and borrow/repay) to the appropriate LH based on reserve configuration and available caps. Whenever liquidity is restored, Spokes pay a base interest (determined by an interest rate strategy at the LH level) and a risk premium that determines how much borrowers will pay on top of the base rate, depending on the collateral composition of each user.
```
yan-man: > `Whenever liquidity is restored`

- maybe `restored on the hub...`

this also talks about Spokes paying a base interest, but later mentions borrowers paying, maybe can be confusing. Can stick to 1 POV perhaps (imo should be borrower pov)

> `a risk premium that determines how much borrowers will pay on top of the base rate, depending on the collateral composition of each user.`

- also uses borrower and user interchangeably, can be confusing. Maybe `...depending on their collateral composition`

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+A LH can have an unspecified number of Spokes, each one contributing to the total outstanding debt and to the interest generated. The LH manages the basic accounting (total liquidity vs available), the interest rates, the draw and supply caps, among others.
```
yan-man: > the interest rates, the draw and supply caps, among others.

- `the interest rates,  and the draw and supply caps, among other parameters.`

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Liquidity Hub (LH)
+
+The LH is immutable and serves as the central coordinator for liquidity management in Aave V4. The design allows for multiple LHs to exist, with each LH maintaining oversight of its own set of Spokes. Each LH sets the supply/borrow caps for its Spokes and enforces crucial accounting invariants. The design objective is to make the LH as simple as possible.
```
yan-man: > The design objective is to make

- `The design objective was to make...`

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+The key aspects of the LH include
```
yan-man:
```suggestion
The key aspects of the LH include:
```

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+- Maintaining a registry of authorized Spokes for each supported asset.
+- Setting liquidity caps to limit Spoke drawing.
```
yan-man:
```suggestion
- Setting liquidity caps to limit Spoke drawing and adding.
```

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+- Enforcing accounting invariants:
+  - Total borrowed assets <= total supplied assets
+  - Total borrowed shares <= total minted shares
```
yan-man:
```suggestion
  - Total borrowed shares <= total minted debt shares
```

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+  - LH assets <= sum of Spoke assets (converted from shares)
```
yan-man: should this be == ?

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Spokes
+
+The Spokes are upgradeable and the primary components responsible for facilitating lending and borrowing functionalities for specific assets within the Aave V4 ecosystem. They can register into LHs and are allowed to draw (borrow) liquidity from them. The nature of the spoke is not specific and can be anything, crypto based, RWA based, DEX LPs based, and so on.
```
yan-man: > `The nature of the spoke is not specific and can be anything, crypto based, RWA based, DEX LPs based, and so on.`

`The nature of the spoke is not specific and can be anything - crypto based, RWA based, DEX LPs based, and so on.`

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+Users interact with the Spokes, which interact directly with the LHs, managing the following aspects:
+
+- Handling lending and borrowing functionality.
+- Managing user data structures and configurations.
+- Providing emergency stop functionality to halt operations, if needed.
```
yan-man, line 68: these emergency ops also exist on hub too

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+- Utilizing share-based accounting internally to optimize gas costs and ensure assets remain within designated caps as interest accrues.
```
yan-man: i dont think share accounting was primarily to optimize gas costs, but rather simplify accounting

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+- Having a distinct `reserveId` in each Spoke, different from the `assetId` in the LH, to allow for Spoke-specific configurations.
+- Managing Oracle interactions.
+
+# Risk Premium
+
+Interest paid as a result of a user's debt is directly impacted by the quality of the assets used as collateral. Likewise, risk level (quality) of the collateral assets of the user determines the additional charge for borrowing, on top of the asset's drawn interest rate.
```
yan-man: > Interest paid

- `Interest charged`

I think need to also somewhere earlier explain that base rate is drawn interest rate, or specify that here before using it interchangeably

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Premium
+
+### Collateral Risk
+
+The collateral risk $CR_i$ is specified by the asset’s quality, which is a BPS number, ranging from 0_00 to 1000_00. A value of 0% means highest quality and risk free, while a value of 1000% signifies the lowest quality and maximum risk possible within the system.
```
yan-man:
```suggestion
The collateral risk $CR_i$ is specified by the quality of an asset `i`, which is a BPS value ranging from 0_00 to 1000_00. A value of 0% means highest quality and risk free, while a value of 1000% signifies the lowest quality and maximum risk possible within the system.
```

- or just use CR instead of CR_i
- also think we should stick to BPS or % here and not jump between the 2

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+This parameter is configurable and part of the Spoke's risk parameters. This means the same asset can have a different collateral risk value across Spokes.
```
yan-man: I think `Collateral Risk` should be capitalized

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+$CR_i$ is the collateral risk of the asset $i$
+
+> 💡 This parameter is only a factor if an asset is utilized as collateral.
+
+### User Risk Premium
+
+The user risk premium $RP_u$ represents the quality of assets used as collateral to borrow against. It depends on multiple dynamic parameters:
```
yan-man: either explain `RP_u` is per user, or just use `RP`

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+- Collateral amount ($C_{u, i}$): Liquidity supplied as collateral, monotonically increasing due to interest it generates if borrowed.
```
yan-man: - explain if you mean collateral asset amount, or $ value
- explain the underscores

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+- Asset price ($P_i$): Prices are continuously fluctuating, with some assets being less volatile than others.
+- Collateral Risk ($CR_i$): Risk parameter configured and updated on a regular basis by the Governor.
```
yan-man: explain underscores

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+Ideally, the user risk premium would be updated continuously to reflect its dynamic nature, ensuring it is always up-to-date and aligned with the last state of the user’s position. However, this is technically infeasible because of the limitations of EVM blockchains, requiring constant updates by means of onchain transactions. Instead, the user risk premium is updated only when the user performs certain actions which affect collateralization (withdraw, borrow, repay); or an update is permissionlessly triggered (updateUserRiskPremium) by the same user, a position manager or governance, when it is beneficial for a given user position. If the user remains inactive, their user risk premium remains constant. Exceptionally the Governor retains the ability to forcibly update the user risk premium of a given user to match the most recent risk parameters of its collateral assets, even in the absence of user interaction. This is particularly relevant in scenarios where a specific user position has accumulated additional risk between interactions.
```
yan-man: > `or an update is permissionlessly triggered (updateUserRiskPremium) by the same user, a position manager or governance, when it is beneficial for a given user position`

- need to clarify that permissionless is only for a user on their own position. For posm and gov, it is restricted. And it doesnt need to be beneficial for a user position anymore

docs/overview.md
```diff
@@ -0,0 +1,295 @@
...
+
+$RP_u$ is the risk premium of the user $u$
+
+### Risk Premium Algorithm
+
+The algorithm used to calculate it for a given user position follows these steps, with the purpose of first finding the collateral assets and its corresponding amounts which are sufficient to cover the position's debts, in base currency. The weighted average of the Collateral Risk of these collateral assets is then calculated to yield the User Risk Premium:
```
yan-man: > first finding the collateral assets and its corresponding amounts which are sufficient to cover the position's debts

- finding the corresponding amounts of collateral assets which are sufficient to cover the position's debts

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+- $C_{u, i}$ is the amount of asset $i$ supplied as collateral by the user $u$, which sufficiently covers the user's debt, as specified in the above algorithm
+- $P_i$ is the price of asset $i$
+
+**Example 1:** the value of the first collateral asset matches the value of the user’s total debt:
+
+$RP_u = f(CR_0, C_{u,0}, P_0) = CR_0$
+
+**Example 2:** the total value of the first and second collateral assets matches the user’s total debt.
+
+$RP_u = f(CR_i, C_{u, i}, P_i) = \frac{CR_0C_{u,0}P_0 + CR_1C_{u,1}P_1}{C_{u,0}P_0+C_{u,1}P_1}$
```
yan-man, line 129: maybe overkill but would be nice to have an ex. with some numbers too

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Premium Offset
+
+Operationally, the premium is accounted for by assigning additional (virtual) debt shares (premium debt shares) that exist solely to increase interest accrual and are not repayable principal. To separate this premium component from principal interest, we track a premium offset. Premium drawn shares are recorded in shares units, while the premium offset is tracked in asset units. A user’s accrued premium debt at any time equals the assets value of their premium drawn shares minus the premium offset. On user actions, this accrued amount is moved to a realized‑premium variable and both the premium drawn shares and premium offset are reset, since the Risk Premium of a user may change and the accounting must be recalibrated.
```
yan-man: maybe can clarify here that offset intuitively is the principal amt at time of a borrow, while premium shares grow with interest. Thus why subtracting yields the premium debt

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+# Interest Accrual
+
+Interest on every borrow position is split into two concurrent streams: **drawn debt** accrues at the LH’s Drawn Rate $R_{s,i}$, reflecting utilization of LH liquidity, while a separate **premium debt** stream accrues at the incremental rate $R_{sbase,i}RP_u$, where $RP_u$ is the user-specific Risk Premium derived from the quality mix of their collateral.
```
yan-man: no incorrect - both grow at the same drawn rate, but premium just grows based on a different principal (the offset)

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+$D_{u,i}$ is the debt of a user $u$ for asset $i$
+
+$D_{u,i} = D_{u,ibase} + D_{u,premium}$
```
yan-man:
```suggestion
$D_{u,i} = D_{u,ibase} + D_{u,ipremium}$
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+- $D_{u,ibase}$ is the base debt of a user $u$ for the asset $i$
+- $D_{u,premium}$ is the premium debt of a user $u$
```
yan-man: should this also be for asset i?

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+The sum of these gives the expected total interest accumulation such that the user’s debt $D_{u,i}$ grows at rate $R_{u,i}$. This separation is purely internal and separate from the user’s perspective; users simply see their total owed amount increase at the higher rate $R_{u,i}$.
+
+## Base Debt
+
+Base Debt refers to the core portion of a user’s outstanding borrow position that is tied to the actual liquidity drawn from the LH. When a user in Spoke $s$ borrows an asset $i$, the system records this borrowed amount as the user’s base debt.
+
+$D_{u,ibase}$ is the borrowed amount of asset $i$ by a user $u$
+
+This represents the principal liquidity provided by the LH's reserve to the Spoke on the user’s behalf. At the moment of borrowing, the user’s base debt equals the amount borrowed.
```
yan-man: maybe good to use another word besides reserve here as that's used for spoke reserve

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+$D_{u,premium}= D_{u,premium} + R_{sbase,i}RP_uD_{u,ibase}$
+
+$R_{sbase,i}RP_uD_{u,ibase} = ΔD_{u,premium}$
+
+# Dynamic Risk Configuration
+
+One of the major risk‑side limitations of V3 lies in its single, global risk configuration per asset. This design creates significant governance overhead and potential user harm through unexpected liquidations, as any parameter change, in particular lowering the liquidation threshold, immediately affects every open position.
+
+V4 makes it possible for multiple risk configurations to exist side‑by‑side. Whenever the Governor adjusts collateralization parameters (currently the Collateral Factor (CF) or Liquidation Bonus (LB), the protocol adds a new configuration instead of replacing the old one. Earlier configurations continue to govern positions opened under them while updated parameters apply to new positions. In special cases defined by the Governor, existing positions may be modified.
```
yan-man:
```suggestion
V4 makes it possible for multiple risk configurations to exist side‑by‑side. Whenever the Governor adjusts collateralization parameters (currently the Collateral Factor (CF) or Liquidation Bonus (LB)), the protocol adds a new configuration instead of replacing the old one. Earlier configurations continue to govern positions opened under them while updated parameters apply to new positions. In special cases defined by the Governor, existing positions may be modified BY UPDATING THEIR RISK CONFIG TO THE LATEST PARAMETERS. -> clarify what modified means
```

yan-man: also maybe examples that fall under the special cases so it isn't as arbitrary (ie threatening protocol health or extreme emergency cases)

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+Every time the Governor adjusts the Collateral Factor (CF) or Liquidation Bonus (LB), it corresponds to a new configuration. These configurations are stored in a bounded dictionary of up to 65k entries (2^16) identified by incremental keys, with each reserve holding the key that points to the current active configuration.
+
+Each user position also stores the key corresponding to the active configuration when that position became risk-bearing. This key is refreshed whenever the user performs specific actions, and may continue to reference a prior configuration when there are changes on the dynamic risk configuration between user interactions.
```
yan-man:
```suggestion
Each user position also stores the key corresponding to the active configuration when that position became risk-bearing. This key is refreshed whenever the user performs specific actions, but may continue to reference a prior configuration even if there are changes on the dynamic risk configuration between user interactions.
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Governance Impact
+
+Risk‑Config IDs allow parameter updates without affecting existing open positions. The Governor retains the ability to update parameters of old keys. However, during normal operations the system updates upon user interaction without requiring governance intervention.
```
yan-man: > Risk‑Config IDs

rename this (ie to dynamic config keys) to be consistent throughout

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Design
+
+Dynamic configuration extends the reserve model with a per‑reserve mapping that holds every historic configuration, referenced by a `configKey`. Collateralization parameters now reside inside the dynamic mapping rather than the static reserve record; for the moment this set comprises the CF and the LB.
```
yan-man: now this set includes CF, LB, and liquidationFee

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+Each reserve stores the latest configKey, which represents the current up-to-date risk configuration. In contrast, every user position retains a snapshot of the active configKey corresponding to the configuration in effect at the time of its last risk-increasing event. This snapshot is refreshed across all assets of a user position only when the user performs an action which elevates the risk posed to the system, by disabling an asset as collateral, withdrawing, or borrowing. When a user designates a new asset as collateral, only the configKey snapshot of the asset in play is refreshed.
```
yan-man: > by disabling an

such as disabling an

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Automatic Rebinding and Hard Safety Guard
+
+When a user attempts a health‑decreasing action, the engine checks the latest configuration for each collateral in the position. If the position remains sustainable under this configuration, the engine rebinds the snapshot to this latest key and allows the action to proceed. However, if the latest configuration would leave the position under‑collateralized, the engine blocks the action.
```
yan-man:
```suggestion
When a user attempts a health‑decreasing action, the engine checks the latest configuration for each collateral in the position. If the position remains sustainable under this configuration, the engine rebinds the snapshot to this latest key and allows the action to proceed. However, if the latest configuration would leave the position under‑collateralized, the engine blocks the action by reverting.
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+### Feature Notes
+
+The architecture of dynamic configuration comes with several practical constraints and behaviors that integrators and the Governor should note. The points that follow detail some of those mechanics.
+
+1. The `configKey` is currently defined as a `uint16`(65k max active configurations).
```
yan-man: note: could change

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+2. If the system exceeds the maximum number of configurations, the key space wraps and the earliest configuration is overwritten, which can impact users bound to that configuration. Although unlikely given the large available capacity, this behavior is expected and desired.
```
yan-man: note: could change

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+   2. `updateDynamicReserveConfig` updates a prior configuration, affecting existing positions bound to that configKey.
+6. Users can refresh their Dynamic Risk Configuration:
+   1. `updateUserDynamicConfig` updates their snapshots to the latest configKey for all collateral reserves
+
+# Liquidation Engine
+
+Aave V4 introduces a redesigned liquidation mechanism that replaces the fixed close‑factor logic used in V3. Instead of always seizing a fixed percentage of a user’s debt and collateral, V4 allows liquidators to repay just enough debt and seize just enough collateral to bring the borrower’s health factor (HF) back to a configurable Target Health Factor (`TargetHealthFactor >= HEALTH_FACTOR_LIQUIDATION_THRESHOLD`). The mechanism includes safeguards against “dust” debt and collateral, adopts a dutch‑auction style variable liquidation bonus, while ensuring that liquidations do not result in remaining dust collateral or debt unless the respective corresponding debt or collateral reserves are fully liquidated. These changes aim to improve user experience and reduce the chance of protocol‑level bad debt.
```
yan-man:
```suggestion
Aave V4 introduces a redesigned liquidation mechanism that replaces the fixed close‑factor logic used in V3. Instead of always seizing a fixed percentage of a user’s debt and collateral, V4 allows liquidators to repay just enough debt and seize just enough collateral to bring the borrower’s health factor (HF) back to a configurable Target Health Factor (`TargetHealthFactor >= HEALTH_FACTOR_LIQUIDATION_THRESHOLD`). The mechanism adopts a dutch‑auction style variable liquidation bonus, but also includes safeguards against “dust” which ensures that liquidations do not result in remaining dust collateral or debt unless the respective corresponding debt or collateral reserves are fully liquidated. These changes aim to improve user experience and reduce the chance of protocol‑level bad debt.
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Key Differences from Aave v3
+
+- **Target Health Factor vs Close Factor:** In V3, the default close factor is 50% (with a 100% close factor when HF < 0.95). Liquidators would always repay half of a borrower’s debt and seize half of their collateral. V4 removes the default close‑factor: a liquidator only repays the debt required to bring the borrower back to the Target Health Factor determined by the Governor.
```
yan-man:
```suggestion
- **Target Health Factor vs Close Factor:** In V3, the default close factor is 50% (with a 100% close factor when HF < 0.95 or when liquidation amounts are under a given base currency threshold). Liquidators would typically repay half of a borrower’s debt and seize half of their collateral. V4 removes the default close‑factor: a liquidator only repays the debt required to bring the borrower back to the Target Health Factor determined by the Governor.
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+- **Dynamic Dust Debt Handling**: To avoid dust collateral or debt in user positions, V3 reverts when the remaining amount is below a hard‑coded threshold. V4 instead dynamically adjusts the maximum debt that can be liquidated (`maxDebtToLiquidate`) and, if the liquidator opts to fully repay, allows full repayment to prevent dust. Dust may still remain if either the collateral or debt reserve is fully liquidated.
```
yan-man: not really true, bc in V3 it also "dynamically" adjusts. It reverts if the debt remaining is under $1k, but you are still able to liq the whole position in that case. 

Instead V4 is just more "clean" in dust handling.

ex. $2500 debt remaining, $3500 coll remaining (HF = 0.96) 
- in v3, you can liquidate only 50% bc HF > 0.95 and starting amount is > $2000
- you would need to liquidate 50% debt in first go to get to $1250 remaining, then must liquidate 100% of $1250 on 2nd round to liquidate remaining 
- in v4 you can liquidate up to $1500 to end up with debt >= $1000, or $2500 to get the whole position

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+- **Dutch‑Auction Style Liquidation Bonus:** V3 applies a static liquidation bonus that does not depend on the borrower’s health factor. V4 introduces a variable liquidation bonus that increases linearly as the health factor decreases. Governance can specify two spoke‑wide parameters that shape the bonus curve: `healthFactorForMaxBonus` and `liquidationBonusFactor`.
```
yan-man: nit: bonus curve not really a curve, it's linearly interpolating

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+| `TargetHealthFactor`                          | A spoke‑wide value set by the Governor representing the HF to which a borrower should be restored after liquidation. Liquidators repay only enough debt to reach this HF under normal circumstances that do not result in dust collateral or debt remaining.                                  | Must be ≥ the `HEALTH_FACTOR_LIQUIDATION_THRESHOLD` constant.      |
+| `DUST_LIQUIDATION_THRESHOLD`                  | Hard‑coded threshold used to prevent extremely small leftover debt. The maximum debt that can be liquidated is increased to ensure that debt or collateral dust less than this threshold does not remain unless the corresponding respective collateral or debt reserve is fully liquidated.  | Hard‑coded constant set to 1_000 USD in base units.                |
```
yan-man, line 238: is this meant to be technical-facing or purely user-facing, should we specify 1e26 is the base units?

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+| `maxLiquidationBonus`                         | Per reserve defined maximum liquidation bonus for a collateral, expressed in basis points (BPS). A value of 105.00 means there is 5.00 extra seized collateral over the amount of debt repaid in base currency.                                                                               | Must be ≥ 100.00                                                   |
```
yan-man: i think should be 105_00, 5_00 for bps, or deal in percentages (as is, 105.00, 5.00 are %)

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+| `healthFactorForMaxBonus`                     | Spoke‑wide value expressed in WAD units defining the HF below which the max bonus applies. It must be less than or equal to `HEALTH_FACTOR_LIQUIDATION_THRESHOLD` to avoid division‑by‑zero.                                                                                                  | `healthFactorForMaxBonus` < `HEALTH_FACTOR_LIQUIDATION_THRESHOLD`. |
```
yan-man, line 240: should we specify that division by zero is not in calculating liq bonus, but rather only in debtToTargetHF calcs during liquidation?

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+| `liquidationBonusFactor`                      | Spoke‑wide percentage (expressed in BPS) specifying the fraction of the max bonus earned at the threshold `HEALTH_FACTOR_LIQUIDATION_THRESHOLD`. It defines the minimum bonus; e.g., a factor of 80.00 yields a bonus equal to 80% of the max bonus when HF equals the liquidation threshold. | liquidationBonusFactor must be ≤ 100.00                            |
```
yan-man: if bps, should be 80_00, 100_00

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Liquidation Process in V4
+
+The following high‑level steps outline the V4 liquidation flow:
+
+1. **Check Eligibility:** When a borrower’s HF drops below the `HEALTH_FACTOR_LIQUIDATION_THRESHOLD`, anyone can trigger a liquidation; however, accounts are not allowed to liquidate their own positions. The protocol retrieves the borrower’s total debt value, current HF, and total collateral value, counting only reserves with `usingAsCollateral` enabled, CF > 0, and the reserve not paused; freezing is allowed.
```
yan-man: > freezing is allowed

liquidations in which reserves are frozen are allowed. 

but also - now with the receiveShares=true, we dont allow collateral reserve to be frozen. So can mention this caveat

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+2. **Determine Debt to Repay:** Based on the Target Health Factor `TargetHealthFactor`, the protocol computes the debt that must be repaid to restore the borrower’s HF to `TargetHealthFactor`. The required repayment amount depends on the borrower’s current debt and collateral (CF, LB, HF).
+3. **Handle Dust Debt:** If the borrower’s remaining debt after a standard liquidation would be below the `DUST_LIQUIDATION_THRESHOLD`, and the liquidator intends to fully repay the debt, the protocol increases the allowable debt that can be liquidated. So that the entire debt can be covered. However, dust may still remain if the liquidator targets debt equal to the full amount of the collateral reserve $C_i$ being seized (i.e., $Δ C_i = C_i$), then a residual debt $D_{dust} > 0$ can remain when there are multiple collateral reserves ($N_{coll} > 1$); if there is a single collateral reserve ($N_{coll} = 1$), the residual is recorded as a protocol deficit.
```
yan-man: > liquidated. So that

liquidated, so that

> the residual is recorded as a protocol deficit.

the residual, along with any other existing debt across all reserves, are recorded as protocol deficit.

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+4. **Calculate Collateral to Seize**: Convert the debt to be repaid into the collateral asset’s value and apply the liquidation bonus for this specific liquidation. By this point the bonus is fixed (not variable during execution) based on the position’s HF at the start of liquidation and the reserve’s `maxLiquidationBonus`. The formula in this step just computes that liquidation bonus and the resulting collateral to transfer. If the chosen collateral is not sufficient, all of that collateral is seized and the repaid debt is recomputed; per V4 rules, collateral dust may remain.
```
yan-man: next, we then account for collateral dust remaining (above we only accounted for dust debt remaining)

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+5. **Apply Debt Repayment & Transfer Collateral**: Reduce the borrower’s debt amount by the repaid amount. Transfer the corresponding collateral to the liquidator with the liquidation bonus applied, minus the protocol fee (as in V3). The fee portion is sent to the protocol/fee receiver via the LH as shares, accrues yield there, and the shares are assigned directly via LH accounting.
+6. **Emit Events and Update State:** A `LiquidationCall` event is emitted containing details of the liquidation. The borrower’s and reserve’s interest indices are updated. If the borrower still has debt outstanding and no remaining collateral, the system will record a protocol deficit.
```
yan-man, line 252: single event so far but we might be adding more TBD

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+- **Max‑bonus region**: When HF ≤ `healthFactorForMaxBonus`, the liquidator earns the maximum bonus (`maxLiquidationBonus`) minus a portion (`liquidationFee`) that is taken as protocol fees. Example: if `maxLiquidationBonus = 105%` and `liquidationFee = 10%`, the gross bonus is 5% of the repaid debt; 10% of that bonus is taken as fees, so the liquidator receives a net 4.5% collateral bonus (both in base currency).
+- **Threshold region:** At HF = `HEALTH_FACTOR_LIQUIDATION_THRESHOLD`, the bonus equals `liquidationBonusFactor × maxLiquidationBonus`. This ensures that even the safest possible liquidation (just below the threshold) still yields a non‑zero bonus.
+- **Linear interpolation:** For HF between `healthFactorForMaxBonus` and the liquidation threshold, the bonus increases linearly from `liquidationBonusFactor × maxLiquidationBonus` to `maxLiquidationBonus`.
+
+The lower the HF of a position becomes, the larger the bonus the liquidators receive.
+
+Once a user becomes liquidatable, the protocol offers a minimum liquidation bonus equal to
+
+$$ minLB = (maxLB - 100\%) \times lbFactor + 100\% $$
```
Kogaroshi: display of the formula broken

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+- Maintaining a registry of authorized Spokes for each supported asset.
+- Setting liquidity caps to limit Spoke drawing and adding.
+- Providing emergency stop functionality to halt operations, if needed.
+- Enforcing accounting invariants:
+  - Total borrowed assets <= total supplied assets
+  - Total borrowed shares <= total minted debt shares
+  - LH assets <= sum of Spoke assets (converted from shares)
+
+## Spokes
```
Kogaroshi, line 59: Might be interesting (especially after we merge the Hub transfer refactors), to also specify how the funds are transferred from user to Hub, and to which contracts user must give allowance.

Also just personal preference, but might be better to use "Hub" instead of "LB", to match contract name.

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+Due to rounding effects and the creation of negligible interest premiums during liquidations, the borrower’s final health factor after liquidation may not exactly match the `TargetHealthFactor`. In rare cases the final HF may be slightly above or below the target.
+
+A deficit is only reported if, after liquidation, the borrower has no more collateral left across any of his reserves and debt still remains.
+
+## Dutch‑Auction Style Liquidation Bonus
+
+The liquidation bonus in V4 varies linearly between a minimum and maximum value based on the borrower’s health factor:
+
+- **Max‑bonus region**: When HF ≤ `healthFactorForMaxBonus`, the liquidator earns the maximum bonus (`maxLiquidationBonus`) minus a portion (`liquidationFee`) that is taken as protocol fees. Example: if `maxLiquidationBonus = 105%` and `liquidationFee = 10%`, the gross bonus is 5% of the repaid debt; 10% of that bonus is taken as fees, so the liquidator receives a net 4.5% collateral bonus (both in base currency).
```
yan-man: i dont think base currency needs mentioning here, as it is 4.5% of asset amount regardless of base currency. also now with receiveShares input, can be in shares instead of collateral asset

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+The lower the HF of a position becomes, the larger the bonus the liquidators receive.
+
+Once a user becomes liquidatable, the protocol offers a minimum liquidation bonus equal to
+
+$$ minLB = (maxLB - 100\%) \times lbFactor + 100\% $$
+
+where
+
+- $maxLB$: per collateral. Represents the maximum liquidation bonus that can be awarded. Must be greater than 100%. For example, 103% means a 3% effective liquidation bonus.
```
yan-man:
```suggestion
- $maxLB$: per collateral. Represents the maximum liquidation bonus that can be awarded. Represented as a value greater than 100%. For example, 103% means a 3% effective liquidation bonus.
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+- $lbFactor$: per collateral. Represents the minimum percentage of the effective liquidation bonus that can be awarded. Must be at most 100%. For example, 50% while $maxLB = 103\%$ means $minLB = 101.5\%$.
```
yan-man:
```suggestion
- $liquidationBonusFactor$: per collateral. Represents the proportion of the max liquidation bonus that can be awarded. Must be at most 100%. For example, 50% while $maxLB = 103\%$ means $minLB = 101.5\%$.
```

think we should use `liquidationBonusFactor` everywhere if we're quoting it as a var/term bc it appears that way in code

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+A deficit is only reported if, after liquidation, the borrower has no more collateral left across any of his reserves and debt still remains.
+
+## Dutch‑Auction Style Liquidation Bonus
+
+The liquidation bonus in V4 varies linearly between a minimum and maximum value based on the borrower’s health factor:
+
+- **Max‑bonus region**: When HF ≤ `healthFactorForMaxBonus`, the liquidator earns the maximum bonus (`maxLiquidationBonus`) minus a portion (`liquidationFee`) that is taken as protocol fees. Example: if `maxLiquidationBonus = 105%` and `liquidationFee = 10%`, the gross bonus is 5% of the repaid debt; 10% of that bonus is taken as fees, so the liquidator receives a net 4.5% collateral bonus (both in base currency).
+- **Threshold region:** At HF = `HEALTH_FACTOR_LIQUIDATION_THRESHOLD`, the bonus equals `liquidationBonusFactor × maxLiquidationBonus`. This ensures that even the safest possible liquidation (just below the threshold) still yields a non‑zero bonus.
+- **Linear interpolation:** For HF between `healthFactorForMaxBonus` and the liquidation threshold, the bonus increases linearly from `liquidationBonusFactor × maxLiquidationBonus` to `maxLiquidationBonus`.
```
yan-man, line 268: I think should add another region below this for a separate threshold region, which is under which the bonus is == maxLB

docs/overview.md
```diff
@@ -0,0 +1,294 @@
+# Overview
+
+Aave V4 introduces an architectural redesign centered around the Liquidity Hub, enabling protocol flexibility and efficiency. This innovative architecture allows the Governor (e.g., the Aave DAO) to dynamically manage Spokes, adding new borrowing capabilities and removing outdated ones without requiring costly liquidity migrations.
```
CheyenneAtapour: Both are true, so fine either way, but I'm more excited about capital efficiency than protocol efficiency :)

```suggestion
Aave V4 introduces an architectural redesign centered around the Liquidity Hub, enabling protocol flexibility and capital efficiency. This innovative architecture allows the Governor (e.g., the Aave DAO) to dynamically manage Spokes, adding new borrowing capabilities and removing outdated ones without requiring costly liquidity migrations.
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+lb = \begin{cases}
+maxLB & \text{if } hf_{beforeLiq} \le hfForMaxBonus \\
+minLB + (maxLB - minLB) \times \frac{HF\_LIQ\_THRESHOLD - hf_{beforeLiq}}{HF\_LIQ\_THRESHOLD - hfForMaxBonus} & \text{if } hf_{beforeLiq} > hfForMaxBonus
+\end{cases}
+$$
+
+where
+
+- $HF\_LIQ\_THRESHOLD$: per Spoke. Represents the health factor threshold under which the user becomes liquidatable. Equals 1.
+- $hf_{beforeLiq}$: per user. Represents the user’s health factor before liquidation. Exact formula detailed below.
```
yan-man: should there be a formula below

docs/overview.md
```diff
@@ -0,0 +1,294 @@
+# Overview
+
+Aave V4 introduces an architectural redesign centered around the Liquidity Hub, enabling protocol flexibility and efficiency. This innovative architecture allows the Governor (e.g., the Aave DAO) to dynamically manage Spokes, adding new borrowing capabilities and removing outdated ones without requiring costly liquidity migrations.
+
+The protocol implements sophisticated risk management through its Liquidity Premiums system, where each asset receives a dynamic risk factor (called Collateral Risk) ranging from 0 to 100_000 (BPS) based on the asset's implied volatility, market conditions, liquidity, risk, etc. This granular pricing mechanism introduces base rates for high-quality collateral (such as ETH) while adjusting borrowing costs proportionally to Collateral Risk profiles.
```
CheyenneAtapour:
```suggestion
The protocol implements sophisticated risk management through its Liquidity Premiums system, where each asset receives a dynamic risk factor (called Collateral Risk) ranging from 0 to 1000_00 (BPS) based on the asset's implied volatility, market conditions, liquidity, risk, etc. This granular pricing mechanism introduces base rates for high-quality collateral (such as ETH) while adjusting borrowing costs proportionally to Collateral Risk profiles.
```

Or we could also just say 0-1000%?

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+  S3 --> HUB
+
+  %% Styling (optional)
+  classDef box fill:#0b0b0d,stroke:#e5e5e5,stroke-width:1px,color:#ffffff;
+  class LP,BR,S1,S2,S3,HUB box;
+```
+
+Spokes are individual modules that can connect to one or more LHs. They route user actions (supply/withdraw and borrow/repay) to the appropriate LH based on reserve configuration and available caps. Whenever liquidity is restored on the LH, Spokes have to pay a base interest (determined by an interest rate strategy at the LH level) and a Risk Premium (determined by the collateral composition of the user that triggered the action).
+
+An LH can have an unspecified number of Spokes, each one contributing to the total outstanding debt and to the interest generated. The LH manages the basic accounting (total liquidity vs available), the interest rates, and the draw and supply caps, among other parameters.
```
CheyenneAtapour:
```suggestion
A LH can have an unspecified number of Spokes, each one contributing to the total outstanding debt and to the interest generated. The LH manages the basic accounting (total liquidity vs available), the interest rates, and the draw and supply caps, among other parameters.
```

docs/overview.md
CheyenneAtapour: It's correct because ceil(a) + ceil(b) >= ceil(a + b) (due to rounding up each time vs the single rounding)

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+The LH is immutable and serves as the central coordinator for liquidity management in Aave V4. The design allows for multiple LHs to exist, with each LH maintaining oversight of its own set of Spokes. Each LH sets the supply/borrow caps for its Spokes and enforces crucial accounting invariants. The design objective was to make the LH as simple as possible.
+
+The key aspects of the LH include:
+
+- Maintaining a registry of authorized Spokes for each supported asset.
+- Setting liquidity caps to limit Spoke drawing and adding.
+- Providing emergency stop functionality to halt operations, if needed.
+- Enforcing accounting invariants:
+  - Total borrowed assets <= total supplied assets
+  - Total borrowed shares <= total minted debt shares
```
CheyenneAtapour: Shouldn't this one be `==`?

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+## Spokes
+
+The Spokes are upgradeable and the primary components responsible for facilitating lending and borrowing functionalities for specific assets within the Aave V4 ecosystem. They can register into LHs and are allowed to draw (borrow) liquidity from them. The nature of the Spoke is not specific and can be anything: crypto-based, RWA-based, DEX-LPs-based, and so on.
+
+Users interact with the Spokes, which then interact directly with the LHs. The Spokes manage the following aspects:
+
+- Handling lending and borrowing functionality.
+- Managing user data structures and configurations.
+- Providing emergency stop functionality to halt operations, if needed.
+- Utilizing share-based accounting internally to simplify bookkeeping and ensure assets remain within designated caps as interest accrues.
```
CheyenneAtapour: Caps are enforced at the hub level, not spoke

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+- Having a distinct `reserveId` in each Spoke, different from the `assetId` in the LH, to allow for Spoke-specific configurations.
+- Managing oracle interactions.
+
+# Risk Premium
+
+Interest charged as a result of a user's debt is directly impacted by the quality of the assets used as collateral. Likewise, risk level (quality) of the collateral assets of the user determines the additional charge for borrowing, on top of the asset's base rate (i.e., the asset's drawn rate).
```
CheyenneAtapour:
```suggestion
Each user's debt interest is directly impacted by the quality of the assets used as collateral. The risk level (quality) of the collateral assets of the user determines the additional charge for borrowing, on top of the asset's base rate (i.e., the asset's drawn rate).
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Premium
+
+### Collateral Risk
+
+The Collateral Risk $CR_i$ is specified by the quality of the asset $i$, which is a BPS value, ranging from 0 to 100_000. A value of 0 means highest quality and risk-free, while a value of 1_000 signifies the lowest quality and maximum risk possible within the system.
```
CheyenneAtapour: Again, maybe do %s. But also, here there is an inconsistency. Using both 100_000 and 1_000 when you mean the same number

CheyenneAtapour: ....`maximum risk possible for a collateral`

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+This parameter is configurable and part of the Spoke's risk parameters. This means the same asset can have a different Collateral Risk value across Spokes.
+
+$CR_i$ is the Collateral Risk of the asset $i$
+
+> 💡 This parameter is only a factor if an asset is utilized as collateral.
```
CheyenneAtapour:
```suggestion
> 💡 This parameter only applies if an asset is utilized as collateral.
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+Ideally, the User Risk Premium would be updated continuously to reflect its dynamic nature, ensuring it is always up-to-date and aligned with the last state of the user’s position. However, this is technically infeasible because of the limitations of EVM blockchains, requiring constant updates by means of onchain transactions. Instead, the User Risk Premium is updated only when the user performs certain actions which affect collateralization (withdraw, borrow, repay). An update can also be permissionlessly triggered (updateUserRiskPremium) by a user on their own positions, when it is beneficial for a given position. Additionally, a position manager or the Governor can trigger permissioned updates in particular cases. If the user remains inactive, their User Risk Premium remains constant. Exceptionally, the Governor retains the ability to forcibly update the User Risk Premium of a given user to match the most recent risk parameters of its collateral assets, even in the absence of user interaction. This is particularly relevant in scenarios where a specific user position has accumulated additional risk between interactions.
```
CheyenneAtapour: Risk premium can also be updated when supplying an asset used as collateral, so can perhaps emit the `(withdraw, borrow, repay)` portion

CheyenneAtapour: A user can trigger `updateUserRiskPremium` regardless if it goes up or down

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+$RP_u$ is the Risk Premium of the user $u$
+
+### Risk Premium Algorithm
+
+The algorithm used to calculate it for a given user position follows these steps, with the purpose of finding the corresponding amounts of collateral assets which are sufficient to cover the position's debts, in base currency. The weighted average of the Collateral Risk of these collateral assets is then calculated to yield the User Risk Premium:
```
CheyenneAtapour:
```suggestion
The algorithm used to calculate risk premium for a given user position follows these steps, with the purpose of finding the corresponding amounts of collateral assets which are sufficient to cover the position's debts, in base currency. The weighted average of the Collateral Risk of these collateral assets is then calculated to yield the User Risk Premium:
```

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Base Debt
+
+Base Debt refers to the core portion of a user’s outstanding borrow position that is tied to the actual liquidity drawn from the LH. When a user in Spoke $s$ borrows an asset $i$, the system records this borrowed amount as the user’s base debt.
+
+$D_{u,ibase}$ is the borrowed amount of asset $i$ by a user $u$
+
+This represents the principal liquidity provided by the LH's reserve to the Spoke on the user’s behalf. At the moment of borrowing, the user’s base debt equals the amount borrowed.
+
+Over time, the base debt accrues interest at the LH’s base borrow rate strategy $R_{sbase,i}$, in accordance with the interest rate strategy for that Spoke. This means that as time progresses, the accrued base interest is added to the user’s base debt, increasing the amount the user owes to the protocol’s liquidity providers for that particular asset.
```
CheyenneAtapour: base debt will just accrue with the base borrow rate strategy, the stuff on spoke is premium, which is separate

docs/overview.md
CheyenneAtapour: Could also split this up into 2 sentences, one about the dutch-auction with variable liquidation bonus, and then mention separately the safeguards against dust. Because the sentence is quite long, and contains a lot of info

docs/overview.md
CanonicalJP: Agree that Hubs is better than LHs, I'll review the first topic

docs/overview.md
CanonicalJP: updated sentence since that was not the intended meaning

docs/overview.md
```diff
@@ -0,0 +1,294 @@
...
+
+# Dynamic Risk Configuration
+
+One of the major risk‑side limitations of V3 lies in its single, global risk configuration per asset. This design creates significant governance overhead and potential user harm through unexpected liquidations, as any parameter change, in particular lowering the liquidation threshold, immediately affects every open position.
+
+V4 makes it possible for multiple risk configurations to exist side‑by‑side. Whenever the Governor adjusts collateralization parameters (currently the Collateral Factor (CF) or Liquidation Bonus (LB), the protocol adds a new configuration instead of replacing the old one. Earlier configurations continue to govern positions opened under them while updated parameters apply to new positions. In special cases defined by the Governor, existing positions may be modified.
+
+Every time the Governor adjusts the Collateral Factor (CF) or Liquidation Bonus (LB), it corresponds to a new configuration. These configurations are stored in a bounded dictionary of up to 65k entries (2^16) identified by incremental keys, with each reserve holding the key that points to the current active configuration.
```
Kogaroshi: yup to 16M entries now (2^24)

docs/overview.md
Kogaroshi: changed, uint24 now

docs/overview.md
Kogaroshi: changed, does not wrap anymore, no overwrite.

README.md
```diff
@@ -1,5 +1,9 @@
 # Aave V4
 
+## Documentation
+
+- [Aave V4 Overview](./docs/overview.md)
+
```
yan-man: can you rm the changes to this file just to minimize diffs in my PR, and i'll add this into mine

docs/overview.md
CheyenneAtapour: Confirmed today in sync that it should be

docs/overview.md
```diff
@@ -0,0 +1,335 @@
...
+
+$RP_u = f(CR_0, C_{u,0}, P_0) = CR_0$
+
+**Example 2:** the total value of the first and second collateral assets matches the user’s total debt.
+
+$RP_u = f(CR_i, C_{u, i}, P_i) = \frac{CR_0C_{u,0}P_0 + CR_1C_{u,1}P_1}{C_{u,0}P_0+C_{u,1}P_1}$
+
+### Premium Offset
+
+Operationally, the premium is accounted for by assigning additional (virtual) debt shares (premium debt shares) that exist solely to increase interest accrual and are not repayable principal. To separate this premium component from principal interest, we track a premium offset, which is equivalent to the principal when the borrowing call is executed. Premium drawn shares are recorded in shares units, while the premium offset is tracked in asset units.
```
CheyenneAtapour: This is not entirely accurate (but this is def tricky to write). The premium offset is equivalent to the value of the premium shares converted to asset terms, at the time the borrowing call is executed. 

Can say something along the lines of: The premium component is separated from the principal interest by tracking a premium offset. At the time of borrowing, the value of the offset exactly matches the value of the premium shares in asset terms. As time goes on, interest accumulates to grow the value of the premium shares beyond the offset. This interest is referred to as premium.

docs/overview.md
```diff
@@ -0,0 +1,335 @@
...
+
+A user’s accrued premium debt at any time equals the assets value of their premium drawn shares minus the premium offset. On user actions, this accrued amount is moved to a realized‑premium variable and both the premium drawn shares and premium offset are reset, since the Risk Premium of a user may change and the accounting must be recalibrated.
+
+# Interest Accrual
+
+Interest on every borrow position is split into two concurrent streams accruing at the Hub’s Drawn Rate: the **drawn debt** stream accrues based on the principal (reflects utilization of Hub liquidity), resulting in the base rate $R_{sbase,i}$ for the asset $i$. The **premium debt** stream accrues based on the offset (reflects the quality mix of the position's collateral), resulting in the Risk Premium $R_{sbase,i}RP_u$, where $RP_u$ is the premium of the user $u$.
```
CheyenneAtapour: Kinda is mixing together concepts. I don't exactly like my suggestion, but a start:
```suggestion
Interest on every borrow position is split into two concurrent streams accruing at the Hub’s Drawn Rate: the **drawn debt** stream accrues based on the principal (reflects utilization of Hub liquidity), resulting in the base rate $R_{sbase,i}$ for the asset $i$. The **premium debt** stream accrues based on the premium shares minus the offset. The premium debt stream reflects the quality mix of the position's collateral, determined by the Risk Premium $R_{sbase,i}RP_u$, where $RP_u$ is the premium of the user $u$.
```

docs/overview.md
```diff
@@ -0,0 +1,335 @@
...
+
+**Example 2:** the total value of the first and second collateral assets matches the user’s total debt.
+
+$RP_u = f(CR_i, C_{u, i}, P_i) = \frac{CR_0C_{u,0}P_0 + CR_1C_{u,1}P_1}{C_{u,0}P_0+C_{u,1}P_1}$
+
+### Premium Offset
+
+Operationally, the premium is implemented via additional virtual debt shares (“premium shares”) that increase interest accrual but are never repayable principal. We separate this component from principal interest by tracking a premium offset in asset units. At borrow time, the offset is set so that, in asset terms, it exactly equals the value of the premium shares. As time elapses, interest accrues on the premium shares causing their asset value to exceed the offset; the excess is the premium. Premium shares are recorded in share units. The premium offset is recorded in asset units.
+
+A user’s accrued premium debt at any time equals the assets value of their premium drawn shares minus the premium offset. On user actions, this accrued amount is moved to a realized‑premium variable and both the premium drawn shares and premium offset are reset, since the Risk Premium of a user may change and the accounting must be recalibrated.
```
CheyenneAtapour, line 135: realized premium variable potentially being removed with pending changes

yan-man: https://github.com/aave/aave-v4/pull/993

docs/overview.md
```diff
@@ -0,0 +1,335 @@
...
+
+### Collateral Risk
+
+The Collateral Risk $CR_i$ is specified by the quality of the asset $i$, which is a BPS value, ranging from 0 to 1000_00. A value of 0 means highest quality and risk-free, while a value of 1000_00 signifies the lowest quality and maximum risk possible for a collateral.
+
+This parameter is configurable and part of the Spoke's risk parameters. This means the same asset can have a different Collateral Risk value across Spokes.
+
+$CR_i$ is the Collateral Risk of the asset $i$
+
+> 💡 This parameter only applies if an asset is utilized as collateral.
```
miguelmtzinf: feels out of place. we can rm imo

docs/overview.md
```diff
@@ -0,0 +1,335 @@
...
+
+- **Max‑bonus region**: When HF ≤ `healthFactorForMaxBonus`, the liquidator earns the maximum bonus (`maxLiquidationBonus`) minus a portion (`liquidationFee`) that is taken as protocol fees. Example: if `maxLiquidationBonus = 105_00` and `liquidationFee = 10_00`, the gross bonus is 5% of the repaid debt; 10% of that bonus is taken as fees, so the liquidator receives a net 4.5% collateral bonus.
+- **Threshold region:** At HF = `HEALTH_FACTOR_LIQUIDATION_THRESHOLD`, the bonus equals `liquidationBonusFactor × maxLiquidationBonus`. This ensures that even the safest possible liquidation (just below the threshold) still yields a non‑zero bonus.
+- **Linear interpolation:** For HF between `healthFactorForMaxBonus` and the liquidation threshold, the bonus increases linearly from `liquidationBonusFactor × maxLiquidationBonus` to `maxLiquidationBonus`.
+
+The lower the HF of a position becomes, the larger the bonus the liquidators receive.
+
+Once a user becomes liquidatable, the protocol offers a minimum liquidation bonus equal to
+
+$$ minLB = (maxLB - 100\%) \times lbFactor + 100\% $$
```
miguelmtzinf: ) not shown properly.

PR #978, opened 2025-10-31
feat: Add 2 wei buffer for hf check

CheyenneAtapour:
This change splits the Health Factor calculation. While preserving the standard computation (collateral value divided by debt value), it introduces an additional Health Factor based on an overestimated debt value (adding 2 wei per borrowed reserve). This adjustment accounts for the potential impact of the 2 wei imprecision in premium debt calculations on the user’s hf.

Depending on the specific action and logic involved, the system will use one of the two Health Factor variants.


src/spoke/Spoke.sol
```diff
@@ -767,8 +767,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       // the division by `totalCollateralValue` to compute the weighted average is done later
       accountData.healthFactor = accountData
         .avgCollateralFactor
-        .wadDivDown(accountData.totalDebtValue)
-        .fromBpsDown();
+        .wadDivDown(accountData.totalDebtValue + (2 * accountData.borrowedCount))
```
Kogaroshi: Let's maybe use an internal constant instead of magic number ?

yan-man: also, this value should not be on the total debt value level, but rather 2 wei on each individual debt asset amount level right (ie L759, 2 added to prem+drawn debt)? that's the annoying part (if we split out into returning 2 diff values), need to convert 2 diff asset amounts to value - 1 asset amt (true) and 1 asset amt (pessimistic)

yan-man: i imagine that will break quite a few tests

CheyenneAtapour: Yes, you're right @yan-man, this needs to be in units of each debt asset. I think what I can do is perhaps calculate that number during the looping, store into a variable, and then subtract out of the final debt value that is returned. And keep this idea of using just in hf calc

yan-man: i think that should work, but would still need to convert twice from asset amt -> value for each asset in the loop (1 for regular, 1 pessimistic), is that right?

CheyenneAtapour: Yeah, we decided in sync better to just have another variable that has the pessimistic total debt value, exactly due to rounding

yan-man: ah might have missed that during sync

src/spoke/Spoke.sol
```diff
@@ -757,6 +757,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
         );
         // we can simplify since there is no precision loss due to the division here
         accountData.totalDebtValue += ((drawnDebt + premiumDebt) * assetPrice).wadDivUp(assetUnit);
+        accountData.overEstimatedDebtValue += ((drawnDebt + premiumDebt + 2) * assetPrice).wadDivUp(
```
miguelmtzinf: are we using this for `targetHealthFactor`?

avniculae: At the moment this PR still uses `totalDebtValue` instead of `overEstimatedDebtValue` when calculating `targetHealthFactor`, while utilising the new health factor calculation.

Existing formula:
```
      params.totalDebtValue.mulDivUp(
        params.debtAssetUnit * (params.targetHealthFactor - params.healthFactor),
        (params.targetHealthFactor - liquidationPenalty) * params.debtAssetPrice.toWad()
      );
```

Instead, I think the new formula should be:
```
      params.overEstimatedDebtValue.mulDivUp(
        params.debtAssetUnit * (params.targetHealthFactor - params.healthFactor),
        (params.targetHealthFactor - liquidationPenalty) * params.debtAssetPrice.toWad()
      );
```

In general, I think health factor is not great because it loses precision when dividing the CF-weighted collateral value by debt value. I think we could remove health factor from the user account data, and replace it with `weightedTotalCollateralValue` (or some better name).

At this point, we can replace the formula from above with
```
    (
      params.overEstimatedDebtValue * params.targetHealthFactor - 
      params.weightedTotalCollateralValue
    ).mulDivUp(
      params.debtAssetUnit, 
      (params.targetHealthFactor - liquidationPenalty) * params.debtAssetPrice.toWad()
    );
```

Of course, when validating user positions, we will need to replace the HF check with
```
userAccountData.weightedTotalCollateralValue >= userAccountData.overEstimatedDebtValue * HEALTH_FACTOR_LIQUIDATION_THRESHOLD
```

wdyt?

src/spoke/Spoke.sol
avniculae, line 779: I assume this is fine bc `accountData.totalDebtValue > 0` is equivalent to `accountData.overEstimatedDebtValue > 0` (bc we only update both of these if borrowing is true, which should always mean >0 debt). However, I would use the latter, wdyt?

src/spoke/Spoke.sol
```diff
@@ -769,8 +772,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       // the division by `totalCollateralValue` to compute the weighted average is done later
       accountData.healthFactor = accountData
         .avgCollateralFactor
-        .wadDivDown(accountData.totalDebtValue)
-        .fromBpsDown();
+        .wadDivDown(accountData.overEstimatedDebtValue)
+        .fromBpsDown(); // Use overestimated debt for hf check since hf is checked before broadcasting premium updates
```
yan-man: in favor of adding this to natspec @dev as it's important

tests/Base.t.sol
```diff
@@ -173,6 +173,9 @@ abstract contract Base is Test {
 
   Decimals internal decimals = Decimals({usdx: 6, usdy: 18, dai: 18, wbtc: 8, weth: 18});
 
+  uint256 internal debtPremiumBuffer = 2;
+  uint256 internal collateralPremiumBuffer = 4;
```
yan-man: why does this collateralPremiumBuffer exist?

tests/Base.t.sol
```diff
@@ -2194,6 +2198,13 @@ abstract contract Base is Test {
     return (finalHf, requiredDebtAmount);
   }
 
+  /// @dev Returns the price of 2 wei for the given reserve on the spoke
+  function _priceOf2Wei(ISpoke spoke, uint256 reserveId) internal view returns (uint256) {
+    uint256 assetPrice = IAaveOracle(spoke.ORACLE()).getReservePrice(reserveId);
+    uint256 assetUnit = MathUtils.uncheckedExp(10, spoke.getReserve(reserveId).decimals);
```
yan-man: nit: i think can re-use a test helper we have for _convertAmountToValue

also this is not price, so much as value right?

tests/Base.t.sol
CheyenneAtapour: Idea is (with the assumption of cf=50% within the tests) that if you add 2 wei of debt, then you need 4 wei of collateral

src/spoke/Spoke.sol
CheyenneAtapour: I like using the former because I only want to use that over-estimated debt in exactly one place: namely the hf check

tests/Base.t.sol
yan-man: don't we have a variety of CF values though, not just 50%? and this is also presuming that the debt asset is same as coll asset only right. can you add some comments here for this. tbh not really a fan of this as it seems a narrow case

CheyenneAtapour: The code usually uses the assumption that CF is 50% or greater, which does hold true. It is true that it just adds wei to whatever asset is there. I think the alternative is to use 2 variables, and then a helper to ensure we are precise with how much buffer we add to account for the 2 wei. But this did seem clean enough. I can revisit to see what works cleanest when all the logic is resolved in this PR

yan-man: yeah helper is ideal but more annoying to write to be exact, sg

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -70,6 +72,7 @@ library LiquidationLogic {
     uint256 debtReserveBalance;
     uint256 debtToCover;
     uint256 totalDebtValue;
+    uint256 overEstimatedDebtValue;
```
miguelmtzinf: we do not need both. I suggest to keep only  `overEstimatedDebtValue`.
Best we can do is to keep same name for params and values so we know what we are using (so we don't pass overestimatedDebtValue as totalDebtValue for example).

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -405,6 +410,7 @@ library LiquidationLogic {
         debtReserveBalance: params.debtReserveBalance,
         debtToCover: params.debtToCover,
         totalDebtValue: params.totalDebtValue,
+        overEstimatedDebtValue: params.overEstimatedDebtValue,
```
```suggestion
totalDebtValue: params.overEstimatedDebtValue,
```

src/spoke/Spoke.sol
miguelmtzinf: re targetHealthFactor, now it's fixed.
re HF precision, don't think it makes a great difference, are you thinking on any edge case scenario?

src/spoke/interfaces/ISpoke.sol
```diff
@@ -109,8 +109,10 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     uint256 riskPremium;
     uint256 avgCollateralFactor;
     uint256 healthFactor;
+    uint256 healthFactorForLiqCheck;
```
miguelmtzinf, line 112:
```suggestion
    uint256 healthFactorForLiquidation;
```
add to docs

src/spoke/interfaces/ISpoke.sol
```diff
@@ -109,8 +109,10 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
...
     uint256 totalCollateralValue;
     uint256 totalDebtValue;
+    uint256 overEstimatedDebtValue;
```
```suggestion
    uint256 overestimatedTotalDebtValue;
```
I like `inflatedTotalDebtValue` because it's shorter but no strong opinion

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf, line 406: are we over-estimating the bonus because we are passing the under-estimated HF? is this needed because we are also using the under-estimated HF on `_calculateDebtToTargetHealthFactor`?

src/spoke/interfaces/ISpoke.sol
miguelmtzinf: I was thinking to use `healthFactor` (for the true hf) and `healthFactorForLiquidation` (for the underestimated hf), but maybe better to be more explicit and use `healthFactor` and `healthFactorWithInflatedDebt`/`healthFactorWithExtraDebt`.

avniculae: it's actually the other way around, liquidations use the true hf, and the rest the underestimate hf

avniculae: I would call them `healthFactor` for the true hf, and `minHealthFactor` for the underestimated hf, wdyt?

miguelmtzinf: We can use `healthFactorWithExtraDebt` and `totalDebtValueWithExtra`
or `healthFactorWithInflatedDebt` and `inflatedTotalDebtValue`

src/spoke/interfaces/ISpoke.sol
avniculae: what about `maxTotalDebtValue`? (in alignment with this https://github.com/aave/aave-v4/pull/978/files#r2498928385 )

src/spoke/libraries/LiquidationLogic.sol
avniculae: yeah, we should use true HF here. Liq bonus is awarded based on the HF before liquidation

src/spoke/Spoke.sol
avniculae: I would prefer higher precision, but I'm not aware of any edge cases atm with the current solution

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -153,7 +157,7 @@ library LiquidationLogic {
         collateralReservePaused: collateralReserve.paused,
         collateralReserveFrozen: collateralReserve.frozen,
         debtReservePaused: debtReserve.paused,
-        healthFactor: params.healthFactor,
+        healthFactor: params.healthFactorForLiqCheck,
```
avniculae, line 162: I would like us to rename fields in all structs to be as clear as possible, wdyt?

src/spoke/Spoke.sol
miguelmtzinf: We gotta use `totalDebtValue` here as we are dividing by that.

miguelmtzinf, line 805: which value to use in `debtValueLeftToCover` is not impt

src/spoke/Spoke.sol
```diff
@@ -759,6 +761,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
         );
         // we can simplify since there is no precision loss due to the division here
         accountData.totalDebtValue += ((drawnDebt + premiumDebt) * assetPrice).wadDivUp(assetUnit);
+        accountData.overEstimatedDebtValue += ((drawnDebt + premiumDebt + 2) * assetPrice).wadDivUp(
```
miguelmtzinf: let's move the 2 to a internal constant

tests/Base.t.sol
miguelmtzinf: do not get this. please make `debtPremiumBuffer` a constant and rename to `PREMIUM_DEBT_BUFFER` or delta

tests/unit/Spoke/Spoke.AccrueLiquidityFee.t.sol
```diff
@@ -145,16 +145,16 @@ contract SpokeAccrueLiquidityFeeTest is SpokeBase {
...
       })
     );
 
     Utils.mintFeeShares(hub1, assetId, ADMIN);
     assertApproxEqAbs(
       hub1.getSpokeAddedShares(assetId, address(treasurySpoke)),
       expectedFeeShares,
-      1,
-      'treasury shares'
+      2,
```
CheyenneAtapour, line 160: This was previously failing at 20k runs, I think due to the precision loss compounding twice (once from above, and then here)

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: Out of the names here, I do prefer `healthFactorWithInflatedDebt` and `inflatedTotalDebtValue`. However, I purposely went with this because the underestimated `healthFactor` is mostly used everywhere, except `healthFactorForLiqCheck` is used to determine only if a position is liquidatable

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: I like `inflatedTotalDebtValue` too - updated

src/spoke/Spoke.sol

src/spoke/Spoke.sol
CheyenneAtapour: So this is fine as is, right?

tests/Base.t.sol
CheyenneAtapour: Ok, I replaced all instances of it with `PREMIUM_DEBT_BUFFER`, but will need to go through and perhaps refactor where it's used

src/spoke/Spoke.sol
CheyenneAtapour: Added to natspec on `_processUserAccountData`

tests/Base.t.sol

src/spoke/libraries/LiquidationLogic.sol
CheyenneAtapour: you mean it should be `overEstimatedDebtValue: params.overEstimatedDebtValue` right? to match params

src/spoke/libraries/LiquidationLogic.sol
CheyenneAtapour: ah dang, ok if this is wrong, then breaks my naming convention mentioned here: https://github.com/aave/aave-v4/pull/978#discussion_r2501708379, so can revisit

src/spoke/libraries/LiquidationLogic.sol
CheyenneAtapour: Yeah, @miguelmtzinf is aligned with that too: https://github.com/aave/aave-v4/pull/978#discussion_r2498777763

src/spoke/libraries/LiquidationLogic.sol
CheyenneAtapour: Just noting here that this change breaks 6 tests, causing underflow on all

miguelmtzinf:
Closing this, as we consider the proposed fix difficult to reason about. We prefer to find a better solution that properly addresses the rounding issue occurring in premium debt calculations across all three levels (Hub, Spoke, and User).

PR #979, opened 2025-10-31
chore: signal user position updates on liq

DhairyaSethi:


yan-man:
see https://github.com/aave/aave-v4/pull/990

PR #986, opened 2025-11-05
feat : deploy payload engine

Kogaroshi:
[WIP]

Add deploy payload engine and scripts for : 
- full Aave V4 deploy
- unique Hub deploy
- unique Spoke deploy

### DIFF
--- a/src/deployments/orchestration/AaveV4TestOrchestration.sol
+++ b/src/deployments/orchestration/AaveV4TestOrchestration.sol
@@ -138,12 +138,15 @@ library AaveV4TestOrchestration {
     // Set Hub Roles
     AaveV4HubRolesProcedure.setHubRoles(accessManagerAddress, report.hubAddress);
 
+    // Configure Hub Assets
+    configureHubsAssets(paramsList);
+
     return (accessManagerAddress, report);
   }
 
   function configureHubsAssets(
     ConfigData.AddAssetParams[] memory paramsList
-  ) external returns (uint256[] memory) {
+  ) public returns (uint256[] memory) {
     uint256[] memory assetIds = new uint256[](paramsList.length);
     for (uint256 i; i < paramsList.length; ++i) {
       assetIds[i] = AaveV4HubConfigProcedures.addAsset(paramsList[i]);


src/deployments/batches/AaveV4AccessBatch.sol
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {BatchReports} from 'src/deployments/types/BatchReports.sol';
+import {AaveV4AccessManagerDeployProcedure} from 'src/deployments/procedures/AaveV4AccessManagerDeployProcedure.sol';
+
+contract AaveV4AccessBatch is AaveV4AccessManagerDeployProcedure {
```
DhairyaSethi: all of these contracts don't need `AaveV4` prefix imo, its redundant

src/deployments/batches/AaveV4SpokeInstanceBatch.sol
```diff
@@ -0,0 +1,60 @@
...
+contract AaveV4SpokeInstanceBatch is
+  AaveV4SpokeInstanceDeployProcedure,
+  AaveV4TransparentUpgradeableProxyDeployProcedure,
+  AaveV4AaveOracleDeployProcedure,
+  AaveV4SpokeConfiguratorDeployProcedure
+{
+  BatchReports.SpokeInstanceBatchReport internal _report;
+
+  constructor(
+    Vm vm,
```
DhairyaSethi, line 20: vm is always constant and can be inherited from a common contract like `DeployUtils`

src/deployments/procedures/deploy/AaveV4AaveOracleDeployProcedure.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {AaveOracle} from 'src/spoke/AaveOracle.sol';
+
+contract AaveV4AaveOracleDeployProcedure {
+  function _deployAaveOracle(
```
DhairyaSethi, line 8: idea would be to reuse these functions, or directly orchestration for test setup as well

src/deployments/procedures/AaveV4TransparentUpgradeableProxyDeployProcedure.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import {TransparentUpgradeableProxy} from 'src/dependencies/openzeppelin/TransparentUpgradeableProxy.sol';
+
+contract AaveV4TransparentUpgradeableProxyDeployProcedure {
+  function _deployTransparentUpgradeableProxy(
```
DhairyaSethi: would simply call this `_proxify`

src/deployments/batches/AaveV4SpokeInstanceBatch.sol
DhairyaSethi: all procedures could inherit this common contract with `vm` and a `bool public constant IS_TEST = true` to skip on contract size check

src/deployments/batches/AaveV4AccessBatch.sol
Kogaroshi: named them using a convention similar to the v3-origin one. But we can decide to use a simpler one instead, no strong opinion.

src/deployments/batches/AaveV4SpokeInstanceBatch.sol
Kogaroshi: I do not plan to make them inherit, but to deploy those to execute the batch, so the idea was to pass the context there directly, but we can change the behavior.

src/deployments/procedures/deploy/AaveV4AaveOracleDeployProcedure.sol
Kogaroshi: for test setup, I though of reusing the procedures but with a different batch logic, alongside some for mock tokens & setting up the parameters.

src/deployments/batches/AaveV4SpokeInstanceBatch.sol
```diff
@@ -0,0 +1,56 @@
...
+    address deployer,
+    address admin_,
+    address accessManagerAddress_,
+    uint8 oracleDecimals_,
+    string memory oracleDescription_
+  ) {
+    address predictedSpokeInstanceAddress = vm.computeCreateAddress(
+      deployer,
+      vm.getNonce(deployer) + 2
```
yan-man, line 29: why +2 here?

Kogaroshi: we compute the address for the Spoke proxy, but between the moment we compute and the actual deployment, we have 2 other deployment to be executed (Oracle & Spoke Implementation), which means 2 txs to be broadcasted, and increasing the deployed nonce twice.

yan-man: hm weird I am hitting the `SpokeInstance address mismatch` error when running this script locally, were you getting the same?

Kogaroshi: when running the script, make sure to use a Foundry account and specify the `sender` so `msg.sender `and `deployer` matches during the whole script, using : 
`--account ${ACCOUNT} --sender ${SENDER}`

yan-man: hm we cant use vm cheatcodes in actual deployments, so dont think we can use this approach for predicting address

was there a reason you opted not to use create2 for computing predicted address like in v3?

src/deployments/batches/AaveV4SpokeInstanceBatch.sol
yan-man: > skip on contract size check

@DhairyaSethi where were you intending this part would be checked, was there anything else IS_TEST was meant to change in this file

src/deployments/batches/AaveV4SpokeInstanceBatch.sol
DhairyaSethi: we can, the script which deploys the contracts one by one isn't deployed onchain. only the config part for payload & config engine

src/deployments/orchestration/AaveV4DeployOrchestration.sol
```diff
@@ -0,0 +1,232 @@
...
+
+library AaveV4DeployOrchestration {
+  bool public constant IS_TEST = true;
+  Vm private constant vm = Vm(address(bytes20(uint160(uint256(keccak256('hevm cheat code'))))));
+
+  uint8 private constant ORACLE_DECIMALS = 8;
+  string private constant ORACLE_SUFFIX = ' (USD)';
+
+  function deployAaveV4(
+    Logger logger,
```
yan-man, line 29: was the intention that Logger contract would also be deployed as part of the typical onchain deployment, or merely used outside the broadcast? if latter, we cant pass the logger into this method, no? @Kogaroshi

src/deployments/batches/AaveV4SpokeInstanceBatch.sol
Kogaroshi: Compiler will check those contract size since they are located in `src/`.

src/deployments/batches/AaveV4SpokeInstanceBatch.sol
Kogaroshi: The `broadcast` logic needs to be reworked to ensure only the actual contracts to deploy would be broadcasted onchain, and not the helper ones.

src/deployments/orchestration/AaveV4DeployOrchestration.sol
Kogaroshi: No, the Logger is not meant to be actually deployed, and yes to be used outside actually broadcasted methods.

src/deployments/orchestration/AaveV4DeployOrchestration.sol
```diff
@@ -0,0 +1,232 @@
...
+    address admin,
+    address nativeWrapper,
+    string[] memory hubLabels,
+    string[] memory spokeLabels,
+    bool setRoles
+  ) internal returns (OrchestrationReports.FullDeploymentReport memory) {
+    OrchestrationReports.FullDeploymentReport memory report;
+
+    // Deploy Access Batch
+    address accessManagerAdmin = setRoles ? deployer : admin;
```
miguelmtzinf, line 40: `setRoles` _here_ actually signifies whether to use `deployer` or `admin` as admin of access manager contract right? but setting up roles anyway

src/deployments/orchestration/AaveV4DeployOrchestration.sol
```diff
@@ -0,0 +1,232 @@
...
+    }
+    logger.writeGroup('SpokeInstances', spokeEntries);
+
+    // Deploy Gateways Batch
+    report.gatewaysBatchReport = _deployGatewaysBatch(admin, nativeWrapper);
+    logger.log('NativeTokenGateway', report.gatewaysBatchReport.nativeGatewayAddress);
+    logger.log('SignatureGateway', report.gatewaysBatchReport.signatureGatewayAddress);
+
+    // Set Roles if needed
+    if (setRoles) {
```
miguelmtzinf, line 112: what d be the reason for `setRoles = false`,  setting roles by the accessManagerAdmin later?

src/deployments/batches/AaveV4HubBatch.sol
```diff
@@ -0,0 +1,32 @@
...
+  AaveV4HubDeployProcedure,
+  AaveV4InterestRateStrategyDeployProcedure,
+  AaveV4TreasurySpokeDeployProcedure
+{
+  BatchReports.HubBatchReport internal _report;
+
+  constructor(address admin_, address accessManagerAddress_) {
+    address hubAddress = _deployHub(accessManagerAddress_);
+    address irStrategyAddress = _deployInterestRateStrategy(hubAddress);
+    address treasurySpokeAddress = _deployTreasurySpoke(admin_, hubAddress);
```
miguelmtzinf, line 20: hmm wondering if this `admin_` should be configurable too

src/deployments/procedures/deploy/AaveV4AaveOracleDeployProcedure.sol
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {AaveOracle} from 'src/spoke/AaveOracle.sol';
+
+contract AaveV4AaveOracleDeployProcedure {
+  function _deployAaveOracle(
+    address spoke_,
+    uint8 decimals_,
+    string memory description_
```
miguelmtzinf, line 11: using `_` suffix everywhere, while convention is only for constructor params

src/deployments/procedures/deploy/AaveV4SpokeInstanceDeployProcedure.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {SpokeInstance} from 'src/spoke/instances/SpokeInstance.sol';
+
+contract AaveV4SpokeInstanceDeployProcedure {
```
miguelmtzinf, line 7: better to have the proxy deployment here no? and return the proxy address

src/deployments/procedures/roles/AaveV4AdminRolesProcedure.sol
```diff
@@ -0,0 +1,52 @@
...
+
+import {Roles} from 'src/libraries/types/Roles.sol';
+
+import {IAccessManager} from 'src/dependencies/openzeppelin/IAccessManager.sol';
+
+library AaveV4AdminRolesProcedure {
+  function setConfiguratorAdminRoles(
+    address accessManagerAddress,
+    address spokeConfiguratorAddress,
+    address hubConfiguratorAddress
```
miguelmtzinf, line 13: i think no need to have `address` suffix

src/deployments/procedures/roles/AaveV4AdminRolesProcedure.sol
```diff
@@ -0,0 +1,52 @@
...
+    address spokeConfiguratorAddress
+  ) internal {
+    IAccessManager(accessManagerAddress).grantRole(
+      Roles.SPOKE_ADMIN_ROLE,
+      spokeConfiguratorAddress,
+      0
+    );
+  }
+
+  function setNewAdminRole(
```
miguelmtzinf, line 41: i d rename to smt like `changeAdminRole`, because above we have `set` functions that do not remove current ones...

src/deployments/procedures/roles/AaveV4AdminRolesProcedure.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {Roles} from 'src/libraries/types/Roles.sol';
```
miguelmtzinf, line 5: using src Roles as default, let's revisit all roles and `USER_POSITION_UPDATER_ROLE`

src/deployments/procedures/roles/AaveV4HubRolesProcedure.sol
```diff
@@ -0,0 +1,24 @@
...
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+
+library AaveV4HubRolesProcedure {
+  function setHubRoles(address accessManagerAddress, address hubAddress) internal {
+    bytes4[] memory selectors = new bytes4[](5);
+    selectors[0] = IHub.addAsset.selector;
+    selectors[1] = IHub.updateAssetConfig.selector;
+    selectors[2] = IHub.addSpoke.selector;
+    selectors[3] = IHub.updateSpokeConfig.selector;
+    selectors[4] = IHub.setInterestRateData.selector;
```
miguelmtzinf, line 17: easy to miss any, so let's double check. It's fine because all restricted fns are permissioned anyway

src/deployments/procedures/config/AaveV4HubConfigProcedures.sol
```diff
@@ -0,0 +1,81 @@
...
+          irStrategy: params.irStrategy,
+          reinvestmentController: address(0)
+        }),
+        new bytes(0)
+      );
+    }
+    return assetId;
+  }
+
+  function addAssetViaConfigurator(
```
miguelmtzinf, line 34: how do we plan to use configurator fns?

src/deployments/procedures/roles/AaveV4AdminRolesProcedure.sol
yan-man: see latest. Now we have:
spoke:
- position updater
- spoke configurator

hub:
- fee minter
- hub configurator

spoke admin has both; hub admin has both

src/deployments/procedures/roles/AaveV4AdminRolesProcedure.sol
yan-man: no longer relevant

src/deployments/batches/AaveV4HubBatch.sol
yan-man: see latest, this has been implemented

src/deployments/orchestration/AaveV4DeployOrchestration.sol
yan-man: changed this to grantRoles, as I am thinking we can set the roles by default and simply not grant them till later.

src/deployments/orchestration/AaveV4DeployOrchestration.sol
yan-man: no longer relevant. now setRoles by default, grantRoles is configurable bool

src/deployments/procedures/config/AaveV4HubConfigProcedures.sol
yan-man: can be used to write scripts to update/add assets post-deployment

src/deployments/procedures/roles/AaveV4HubRolesProcedure.sol
yan-man: will also add some more tests for these procedures, see latest

src/deployments/procedures/deploy/AaveV4SpokeInstanceDeployProcedure.sol
yan-man: see latest

src/deployments/procedures/deploy/AaveV4AaveOracleDeployProcedure.sol
yan-man: see latest

src/deployments/procedures/roles/AaveV4AdminRolesProcedure.sol
yan-man: see latest

Kogaroshi:
Closed in favor of #1047

PR #990, opened 2025-11-06
feat: liquidation events for user position updates

yan-man:
- add liquidation events to signal user position updates

### DIFF
--- a/src/spoke/libraries/LiquidationLogic.sol
+++ b/src/spoke/libraries/LiquidationLogic.sol
@@ -290,7 +290,7 @@ library LiquidationLogic {
     uint256 assetId = collateralReserve.assetId;
 
     uint256 sharesToLiquidate = hub.previewRemoveByAssets(assetId, params.collateralToLiquidate);
-    uint120 newSuppliedShares = collateralPosition.suppliedShares - sharesToLiquidate.toUint120();
+    uint120 userSuppliedShares = collateralPosition.suppliedShares - sharesToLiquidate.toUint120();
     uint256 sharesToLiquidator;
     if (params.collateralToLiquidator > 0) {
       if (params.receiveShares) {
@@ -312,7 +312,7 @@ library LiquidationLogic {
       }
     }
 
-    collateralPosition.suppliedShares = newSuppliedShares;
+    collateralPosition.suppliedShares = userSuppliedShares;
 
     emit ISpokeBase.Withdraw(
       params.collateralReserveId,
@@ -326,7 +326,7 @@ library LiquidationLogic {
       hub.payFeeShares(assetId, sharesToLiquidate.uncheckedSub(sharesToLiquidator));
     }
 
-    return newSuppliedShares == 0;
+    return userSuppliedShares == 0;
   }
 
   /// @dev Invoked by `liquidateUser` method.


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -288,27 +288,43 @@ library LiquidationLogic {
...
+            sharesToLiquidator,
+            params.collateralToLiquidator
+          );
         }
       } else {
         sharesToLiquidator = hub.remove(assetId, params.collateralToLiquidator, params.liquidator);
       }
     }
 
+    collateralPosition.suppliedShares = newSuppliedShares;
```
yan-man: moved this down to be consistent with typical spoke actions: 
- hub action (remove)
- update user position accounting

miguelmtzinf: "issue" is that we are doing a supply before a withdraw here, and RPC calls between these 2 cant lead to bad accounting? withdraw before supply is better in that sense? not sure if this happens somewher else in the code

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -288,27 +288,43 @@ library LiquidationLogic {
...
+      sharesToLiquidate,
+      params.collateralToLiquidate
+    );
+
     if (sharesToLiquidate > sharesToLiquidator) {
       hub.payFeeShares(assetId, sharesToLiquidate.uncheckedSub(sharesToLiquidator));
     }
 
-    return suppliedShares == 0;
+    return newSuppliedShares == 0;
```
miguelmtzinf: wondering if better to rename to `userSuppliedShares`

yan-man: agnostic

src/spoke/libraries/LiquidationLogic.sol
yan-man: you'd be in favor of moving just the event above, or all of it?
just event:
- Withdraw event
- Supply event (if receiveShares) or Hub.remove (ie Remove event)
- userPosition accounting in spoke

all of it:
- userPosition accounting in spoke
- Withdraw event
- Supply event (if receiveShares) or Hub.remove (ie Remove event)

I'm more in favor of "just event" so the Hub action happening before the userPosition accounting remains consistent w other Spoke actions but that decouples the user action from event, so not sure. wdyt?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -290,27 +290,43 @@ library LiquidationLogic {
...
         if (sharesToLiquidator > 0) {
           positions[params.liquidator][params.collateralReserveId]
             .suppliedShares += sharesToLiquidator.toUint120();
+
+          emit ISpokeBase.Supply(
+            params.collateralReserveId,
+            params.liquidator,
+            params.liquidator,
+            sharesToLiquidator,
+            params.collateralToLiquidator
```
yan-man, line 307: another thing is here - if backend is wanting to use emitted amount to keep up to date internal share price, I think this is incorrect or misleading, bc this operation by definition wont change share price as it occurs only on spoke internal accounting. 

ex. share price is 1.5
- collateralToLiquidator is 2
- sharesToLiquidator is 1 share

in other supply actions, donation happens on hub based on rounding, but here the asset amt is not doing anything. wdyt?

src/spoke/libraries/LiquidationLogic.sol
yan-man: https://github.com/aave/aave-v4/pull/990/commits/148e49e281d5e4fdf213ae54b979425d3837f97d

src/spoke/libraries/LiquidationLogic.sol
DhairyaSethi: gas increase is significant and this is a good pt, we can consider just emitting one new event or add to liquidationCall event as well; that way order is does not lead to bad indexer accounting & gas delta likely smaller hit

src/spoke/libraries/LiquidationLogic.sol
DhairyaSethi: indexer is not supposed to touch share price based on spoke events, it is only touching that when reading hub events

yan-man: discussed, so it is fine as is

miguelmtzinf:
Closing this in favor of #992, given that emitting more events increases gas costs and liquidation is critical operation of the system.

PR #991, opened 2025-11-07
feat: New premium accounting

avniculae:
- Due to precision errors induced by `rayMul` and due to the fact that we track premium at 3 different levels (spoke.userPosition, hub.asset, hub.spoke), we are required to round in different directions when computing accrued premium to avoid underflows.
- This PR removes any precision errors in premium accounting by replacing `rayMul(Up/Down)` operations with raw multiplication. This means that premium is using (token.decimals + 27) decimals precision, as well as deficit accounting, in order to prevent misreporting of deficit by spokes. 
- With this fix, there is no longer any potential 2 wei instantaneous premium debt created as a result of precision loss due to rounding.

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -48,8 +48,8 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   /// @dev The maximum allowed value for an asset identifier (inclusive).
   uint256 internal constant MAX_ALLOWED_ASSET_ID = type(uint16).max;
 
-  /// @dev The maximum allowed collateral risk value for a reserve, expressed in bps (e.g. 100_00 is 100.00%).
-  uint24 internal constant MAX_ALLOWED_COLLATERAL_RISK = 1000_00; // 1000.00%
+  /// @dev The maximum allowed collateral risk value for a reserve, expressed in BPS (e.g. 100_00 is 100.00%).
+  uint24 internal constant MAX_ALLOWED_COLLATERAL_RISK = 1000_00;
 
   /// @dev The maximum allowed value for a dynamic configuration key (inclusive).
   uint256 internal constant MAX_ALLOWED_DYNAMIC_CONFIG_KEY = type(uint24).max;
@@ -831,7 +831,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       // riskPremium is 0 when user has no debt or is unhealthy
       return accountData;
     }
-    collateralInfo.sortByKey(); // sort by collateral risk in ASC, collateral value in DESC
+
+    // sort by collateral risk in ASC, collateral value in DESC
+    collateralInfo.sortByKey();
 
     // runs until either the collateral or debt is exhausted
     for (uint256 index = 0; index < collateralInfo.length(); ++index) {
--- a/src/spoke/AaveOracle.sol
+++ b/src/spoke/AaveOracle.sol
@@ -39,7 +39,7 @@ contract AaveOracle is IAaveOracle {
     AggregatorV3Interface targetSource = AggregatorV3Interface(source);
     require(targetSource.decimals() == DECIMALS, InvalidSourceDecimals(reserveId));
     _sources[reserveId] = targetSource;
-    _getSourcePrice(reserveId); // check if the source is valid
+    _getSourcePrice(reserveId);
     emit UpdateReserveSource(reserveId, source);
   }
 


src/spoke/Spoke.sol
```diff
@@ -283,30 +284,40 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    (uint256 drawnDebtRestored, uint256 premiumDebtRestored, uint256 accruedPremium) = _getUserDebt(
-      reserve.hub,
-      reserve.assetId,
-      userPosition
-    );
+    IHubBase.PremiumDelta memory premiumDelta = IHubBase.PremiumDelta({
+      sharesDelta: -userPosition.premiumShares.toInt256(),
+      offsetDeltaRay: -userPosition.premiumOffsetRay.toInt256(),
+      accruedPremiumRay: 0, // populated below
+      restoredPremiumRay: 0 // populated below
```
avniculae, line 314: due to stack too deep 😑

src/hub/libraries/Premium.sol
```diff
@@ -0,0 +1,30 @@
...
+    uint256 accruedPremiumRay = premiumShares * drawnIndex - premiumOffsetRay;
+    return accruedPremiumRay;
+  }
+
+  function calculatePremiumDebt(
+    uint256 realizedPremiumRay,
+    uint256 accruedPremiumRay
+  ) internal pure returns (uint256) {
+    uint256 premiumRay = realizedPremiumRay + accruedPremiumRay;
+    return premiumRay.divUp(WadRayMath.RAY);
```
miguelmtzinf: this reminds me to `fromBps`

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -327,22 +329,24 @@ library LiquidationLogic {
 
       IHubBase.PremiumDelta memory premiumDelta = IHubBase.PremiumDelta({
         sharesDelta: -debtPosition.premiumShares.toInt256(),
-        offsetDelta: -debtPosition.premiumOffset.toInt256(),
-        realizedDelta: params.accruedPremium.toInt256() - premiumDebtToLiquidate.toInt256()
+        offsetDeltaRay: -debtPosition.premiumOffsetRay.toInt256(),
+        accruedPremiumRay: params.accruedPremiumRay,
+        // todo premium-fix: we could pass realizedPremiumRay from the hub to avoid a warm sload
```
avniculae: highlighting this TODO here

src/hub/libraries/Premium.sol
avniculae: true, we can introduce `fromRayUp`, definitely

src/spoke/libraries/LiquidationLogic.sol
avniculae: is it worth it @miguelmtzinf @DhairyaSethi ? I am personally fine with the current impl

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -327,8 +331,12 @@ library LiquidationLogic {
...
+        restoredPremiumRay: (premiumDebtToLiquidate * WadRayMath.RAY).min(
```
avniculae: this `min` function applied here means that the true full-precision premium that is reduced from the user's position can be a bit less then `premiumDebtToLiquidate` (in asset terms). This means that the user health post liquidation can be a bit worse than the target health factor (if liquidation restores health in full) --> this is fine bc our target health factor restoration is not accurate anyways. 

Also, if `min` function has any effect here, it means that user's true premium is not perfectly divisible by RAY, hence restored premium in asset terms (which is transferred from the user) will be rounded up in the hub and will equal `premiumDebtToLiquidate` --> this means liquidator cannot benefit from rounding errors by perfoming many small liquidations.

is my thinking correct @miguelmtzinf @DhairyaSethi @yan-man ?

src/hub/libraries/Premium.sol
avniculae: [249e145](https://github.com/aave/aave-v4/pull/991/commits/249e1452a355747220207d259536af5ed8eef1aa)

src/hub/Hub.sol
```diff
@@ -834,30 +854,30 @@ contract Hub is IHub, AccessManaged {
...
-    require(drawnAmount + premiumAmount > 0, InvalidAmount());
+    require(drawnAmount > 0 || premiumAmountRay > 0, InvalidAmount());
     require(spoke.active, SpokeNotActive());
     require(!spoke.paused, SpokePaused());
     uint256 drawn = _getSpokeDrawn(asset, spoke);
-    uint256 premium = _getSpokePremium(asset, spoke);
+    uint256 premiumRay = _getSpokePremiumRay(asset, spoke);
     require(drawnAmount <= drawn, SurplusAmountRestored(drawn));
-    require(premiumAmount <= premium, SurplusAmountRestored(premium));
+    require(premiumAmountRay <= premiumRay, SurplusAmountRestored(premiumRay.fromRayUp()));
```
avniculae: todo: add test where premiumAmountRay is > premiumRay, but equal in asset terms (when rounded up). same for reportDeficit

src/hub/interfaces/IHub.sol
```diff
@@ -15,8 +15,8 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev decimals The number of decimals of the underlying asset.
   /// @dev deficit The amount of outstanding bad debt across all spokes, expressed in asset units.
   /// @dev swept The outstanding liquidity which has been invested by the reinvestment controller, expressed in asset units.
-  /// @dev realizedPremium The interest-free premium debt already accrued across all spokes, expressed in asset units.
-  /// @dev premiumOffset The total premium offset across all spokes, used to calculate the premium, expressed in asset units.
+  /// @dev realizedPremiumRay The interest-free premium debt already accrued across all spokes, expressed in asset units + RAY.
+  /// @dev premiumOffsetRay The total premium offset across all spokes, used to calculate the premium, expressed in asset units + RAY.
```
```suggestion
  /// @dev realizedPremiumRay The interest-free premium debt already accrued across all spokes, expressed in RAY.
  /// @dev premiumOffsetRay The total premium offset across all spokes, used to calculate the premium, expressed in RAY.
```
this is not enough? could someone think that we are doing whole assets units * ray? don't think so

miguelmtzinf: Can you rename `Ray` to `RAY` in `MathUtils` pls

src/hub/interfaces/IHubBase.sol
```diff
@@ -9,8 +9,9 @@ interface IHubBase {
   /// @notice Changes to premium owed accounting.
   struct PremiumDelta {
```
miguelmtzinf, line 14: whoops, i think we should prob document this struct here

src/hub/libraries/AssetLogic.sol
```diff
@@ -174,12 +179,25 @@ library AssetLogic {
...
+    uint256 liquidityGrowthDrawn = drawnShares.rayMulUp(drawnIndex) -
+      drawnShares.rayMulUp(previousIndex);
 
-    uint256 liquidityGrowth = drawnShares.rayMulUp(drawnIndex) -
-      drawnShares.rayMulUp(previousIndex) +
-      premiumShares.rayMulUp(drawnIndex) -
-      premiumShares.rayMulUp(previousIndex);
-    return liquidityGrowth.percentMulDown(liquidityFee);
+    uint120 premiumShares = asset.premiumShares;
+    uint256 accruedPremiumBeforeRay = Premium.calculateAccruedPremiumRay(
```
```suggestion
    uint256 accruedPremiumRayBefore = Premium.calculateAccruedPremiumRay(
```

miguelmtzinf: we can also change the order (current first, previous afterwards)

src/hub/libraries/AssetLogic.sol
```diff
@@ -174,12 +179,25 @@ library AssetLogic {
...
+    );
+    uint256 accruedPremiumRay = Premium.calculateAccruedPremiumRay(
+      premiumShares,
+      drawnIndex,
+      asset.premiumOffsetRay
+    );
+    uint256 liquidityGrowthPremium = Premium.calculatePremiumDebt(
+      asset.realizedPremiumRay,
+      accruedPremiumRay
+    ) - Premium.calculatePremiumDebt(asset.realizedPremiumRay, accruedPremiumBeforeRay);
```
miguelmtzinf: why shouldn't we substract in RAY terms?

avniculae: bc this is the true increase in premium from the perpective of `totalAddedAssets`

src/hub/Hub.sol
miguelmtzinf: the error can be misleading as is? The user could have tried to repay premiumAmountRay = 1.2, but premiumRay is 1.1, so we are emitting `SurplusAmountRestored(2)` (2 is 1.1 ray round up too).

Si, i guess we need to emit the value in ray
```suggestion
    require(premiumAmountRay <= premiumRay, SurplusAmountRestored(premiumRay));
```

src/hub/Hub.sol
```diff
@@ -905,22 +925,35 @@ contract Hub is IHub, AccessManaged {
...
+    uint256 premiumOffsetRay,
+    uint256 realizedPremiumRay,
+    PremiumDelta calldata premiumDelta
+  ) internal pure returns (uint120, uint256, uint256) {
+    uint256 premiumBeforeRay = Premium.calculateAccruedPremiumRay(
+      premiumShares,
+      drawnIndex,
+      premiumOffsetRay
+    );
+    premiumBeforeRay += realizedPremiumRay;
```
miguelmtzinf: i think putting it together reads well too: realized + calc

src/hub/Hub.sol
```diff
@@ -905,22 +925,35 @@ contract Hub is IHub, AccessManaged {
...
+
+    premiumShares = premiumShares.add(premiumDelta.sharesDelta);
+    premiumOffsetRay = premiumOffsetRay.add(premiumDelta.offsetDeltaRay);
+    realizedPremiumRay =
```
miguelmtzinf: how much gas we are adding by using new local vars? updating input params is not good practice

src/hub/Hub.sol
```diff
@@ -905,22 +925,35 @@ contract Hub is IHub, AccessManaged {
...
+      realizedPremiumRay +
+      premiumDelta.accruedPremiumRay -
+      premiumDelta.restoredPremiumRay;
+
+    uint256 premiumAfterRay = Premium.calculateAccruedPremiumRay(
```
```suggestion
    uint256 premiumRayAfter = Premium.calculateAccruedPremiumRay(
```

src/spoke/Spoke.sol
miguelmtzinf: I think I d like more, but no strong opinion. Current approach gives more visibility about the full struct values
```
IHubBase.PremiumDelta memory premiumDelta;
premiumDelta.sharesDelta = -userPosition.premiumShares.toInt256();
premiumDelta.offsetDeltaRay = -userPosition.premiumOffsetRay.toInt256();
```

src/spoke/Spoke.sol
```diff
@@ -286,34 +287,39 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
       premiumDebtRestored,
       amount
     );
-    IHubBase.PremiumDelta memory premiumDelta = IHubBase.PremiumDelta({
-      sharesDelta: -userPosition.premiumShares.toInt256(),
-      offsetDelta: -userPosition.premiumOffset.toInt256(),
-      realizedDelta: accruedPremium.toInt256() - premiumDebtRestored.toInt256()
-    });
+
+    premiumDelta.restoredPremiumRay = (premiumDebtRestored * WadRayMath.RAY).min(
```
miguelmtzinf: should we add a `toRay` in `WadRayMath`, as we do with `toWad` and check overflow?

miguelmtzinf: also wondering if `_calculateRestoreAmount` should return `premiumDebtRestoredRay` also

src/hub/Hub.sol
DhairyaSethi: the user cannot input in fractional rays, only whole rays, bc the user input is always in asset units; so this is never have any loss

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: couldn't we enhance precision of `processUserAccountData` by using premium in full precision for totalDebtValue and HF calc?

avniculae: we could, we'll need to scale up drawn (base) debt. not sure if we want that.

src/hub/Hub.sol
avniculae: I think this change can be avoided, as it is not needed in the current spoke. another (potentially incorrect) vault could pass restoredPremiumRay = 1.2, while spoke (assume it's the last user) has 1.1 premium left. Without this change, this would revert with an underflow in applyPremiumDelta. With this change, it would fail with SurplusAmountRestored(2), even though user/spoke premium debt is 2 rounded up. 

Let me know if this makes sense, and which option to go with. I am fine reverting this change, as this is just an UX error imo, which should not be triggered by a correct spoke implementation. wdyt?

src/hub/Hub.sol
avniculae: I prefer Ray at the very end, it can be hard to track the precision the var with this suggestion, wdyt?

src/spoke/Spoke.sol
```diff
@@ -286,34 +287,39 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+      realizedPremiumRay + premiumDelta.accruedPremiumRay
+    );
     reserve.underlying.safeTransferFrom(
       msg.sender,
       address(reserve.hub),
       drawnDebtRestored + premiumDebtRestored
```
avniculae, line 335: this is probably fine, bc if min has any effect above, premiumDelta.restoredPremiumRay.fromRayUp() will be equal to premiumDebtRestored anyways. however, I think it's better to update premiumDebtRestore in between

src/hub/Hub.sol
yan-man: agree w Ray at the end for all vars (easier to be consistent)

src/hub/interfaces/IHub.sol
```diff
@@ -68,8 +69,8 @@ interface IHub is IHubBase, IAccessManaged {
 
   /// @notice Spoke position and configuration data.
   /// @dev premiumShares The premium shares of a spoke for a given asset.
-  /// @dev premiumOffset The premium offset of a spoke for a given asset, used to calculate the premium, expressed in asset units.
-  /// @dev realizedPremium The interest-free premium debt already accrued for a spoke for a given asset, expressed in asset units.
+  /// @dev premiumOffsetRay The premium offset of a spoke for a given asset, used to calculate the premium, expressed in asset units + RAY.
+  /// @dev realizedPremiumRay The interest-free premium debt already accrued for a spoke for a given asset, expressed in asset units + RAY.
```
yan-man: if making the change above that Miguel suggested, should do the same here to just say "expressed in RAY"

src/hub/interfaces/IHubBase.sol
```diff
@@ -305,8 +302,8 @@ interface IHubBase {
   /// @param assetId The identifier of the asset.
   /// @param spoke The address of the spoke.
   /// @return The amount of premium shares.
-  /// @return The premium offset.
-  /// @return The realized premium.
+  /// @return The premium offset, in asset units + RAY.
+  /// @return The realized premium, in asset units + RAY.
```
yan-man: same in this file natspec for consistency

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: do not think it's worth

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: any issue there? I think it's worth digging into this to make the target hf mechanism work more accurately

src/hub/Hub.sol
miguelmtzinf: It feels like `premiumRay` is the thing, and we wanna compare it before and after (we use `Before`/`After` as suffixes). Because of this, I prefer to keep that together (premium and ray). (Same with the rest)

Similarly, we do this for things like balances: `userBalanceBefore` and `userBalanceAfter`

src/hub/Hub.sol
miguelmtzinf: `premiumRay.fromRayUp()` is not accurate. We either ack, emit with full precision or rm.
If emitting in full precision, we can use `SurplusDrawnRestored` and `SurplusPremiumRestored`, and `SurplusDrawnDeficitReported` and `SurplusPremiumDeficitReported` for `reportDeficit`.

src/spoke/libraries/LiquidationLogic.sol
avniculae: I think we should consider this in a future PR, as this is a breaking change that would also affect tests I suspect.
Tracking this here https://github.com/aave/aave-v4/issues/999

src/spoke/Spoke.sol
avniculae: [1df1b5f](https://github.com/aave/aave-v4/pull/991/commits/1df1b5f0af3e3ea49b8a1b4a1145de6b5650eb52)

src/spoke/Spoke.sol
avniculae: [1df1b5f](https://github.com/aave/aave-v4/pull/991/commits/1df1b5f0af3e3ea49b8a1b4a1145de6b5650eb52)

src/hub/Hub.sol
avniculae: [1df1b5f](https://github.com/aave/aave-v4/pull/991/commits/1df1b5f0af3e3ea49b8a1b4a1145de6b5650eb52)

src/hub/Hub.sol
avniculae: [1df1b5f](https://github.com/aave/aave-v4/pull/991/commits/1df1b5f0af3e3ea49b8a1b4a1145de6b5650eb52)

src/hub/Hub.sol
avniculae: [1df1b5f](https://github.com/aave/aave-v4/pull/991/commits/1df1b5f0af3e3ea49b8a1b4a1145de6b5650eb52)

src/hub/Hub.sol
avniculae: [1df1b5f](https://github.com/aave/aave-v4/pull/991/commits/1df1b5f0af3e3ea49b8a1b4a1145de6b5650eb52)

src/hub/libraries/AssetLogic.sol
avniculae: [1df1b5f](https://github.com/aave/aave-v4/pull/991/commits/1df1b5f0af3e3ea49b8a1b4a1145de6b5650eb52)

src/hub/Hub.sol
avniculae: [1df1b5f](https://github.com/aave/aave-v4/pull/991/commits/1df1b5f0af3e3ea49b8a1b4a1145de6b5650eb52)

src/hub/interfaces/IHubBase.sol
avniculae: [e35d83f](https://github.com/aave/aave-v4/pull/991/commits/e35d83f927dce6ee87fe9f0c35e0c9686bfd555f)

src/hub/interfaces/IHub.sol
miguelmtzinf: we could say `denominated in asset units and scaled by RAY`

src/hub/interfaces/IHub.sol
```diff
@@ -194,16 +196,24 @@ interface IHub is IHubBase, IAccessManaged {
...
+  error SurplusPremiumRestored(uint256 maxAllowedRestoreRay);
 
   /// @notice Thrown when the premium change is invalid.
   error InvalidPremiumChange();
 
-  /// @notice Thrown when a surplus on existing deficit is reported.
-  /// @param amount The amount of surplus on existing deficit assets.
-  error SurplusDeficitReported(uint256 amount);
+  /// @notice Thrown when a drawb surplus on existing deficit is reported.
+  /// @param amount The amount of drawn surplus on existing deficit assets.
```
miguelmtzinf: what's this? is not the maximum allowed drawn amount to report?

src/hub/interfaces/IHub.sol
```diff
@@ -194,16 +196,24 @@ interface IHub is IHubBase, IAccessManaged {
...
-  /// @notice Thrown when a surplus amount is restored.
-  /// @param maxAllowedRestore The maximum allowed restore amount.
-  error SurplusAmountRestored(uint256 maxAllowedRestore);
+  /// @notice Thrown when a surplus drawn amount is restored.
+  /// @param maxAllowedRestore The maximum allowed drawn amount to restore.
+  error SurplusDrawnRestored(uint256 maxAllowedRestore);
+
+  /// @notice Thrown when a surplus premium amount is restored.
+  /// @param maxAllowedRestoreRay The maximum allowed premium amount to restore, expressed in asset units + RAY.
+  error SurplusPremiumRestored(uint256 maxAllowedRestoreRay);
```
miguelmtzinf: should we do `SurplusPremiumRayRestored`?

src/hub/Hub.sol
```diff
@@ -776,13 +784,25 @@ contract Hub is IHub, AccessManaged {
     return asset.toDrawnAssetsUp(spoke.drawnShares);
   }
 
+  function _getSpokePremiumRay(
```
avniculae, line 821: add external fn on this

avniculae: also maybe on the asset and user (in the spoke)?

src/hub/interfaces/IHub.sol
avniculae: typo above as well

src/hub/Hub.sol
avniculae: [a59b446](https://github.com/aave/aave-v4/pull/991/commits/a59b446902d97ab0fa7fb7f0f7dc992ef377d949)

src/hub/interfaces/IHub.sol
avniculae: [a59b446](https://github.com/aave/aave-v4/pull/991/commits/a59b446902d97ab0fa7fb7f0f7dc992ef377d949)

src/hub/interfaces/IHub.sol
avniculae: [a59b446](https://github.com/aave/aave-v4/pull/991/commits/a59b446902d97ab0fa7fb7f0f7dc992ef377d949)

src/hub/interfaces/IHubBase.sol
avniculae: [a59b446](https://github.com/aave/aave-v4/pull/991/commits/a59b446902d97ab0fa7fb7f0f7dc992ef377d949)

src/hub/interfaces/IHub.sol
avniculae: [a59b446](https://github.com/aave/aave-v4/pull/991/commits/a59b446902d97ab0fa7fb7f0f7dc992ef377d949)

src/hub/interfaces/IHub.sol
avniculae: [a59b446](https://github.com/aave/aave-v4/pull/991/commits/a59b446902d97ab0fa7fb7f0f7dc992ef377d949)

src/hub/interfaces/IHub.sol
```diff
@@ -79,11 +80,12 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev paused True if the spoke is prevented from performing actions that instantly update the liquidity.
   /// @dev deficit The deficit reported by a spoke for a given asset, expressed in asset units.
   struct SpokeData {
+    uint120 drawnShares;
```
miguelmtzinf, line 83: fix order of docs pls

src/hub/interfaces/IHubBase.sol
```diff
@@ -246,6 +247,11 @@ interface IHubBase {
   /// @return The total amount of the assets owed.
   function getAssetTotalOwed(uint256 assetId) external view returns (uint256);
 
+  /// @notice Returns the full precision premium of the specified asset owed to the Hub.
+  /// @param assetId The identifier of the asset.
+  /// @return The amount of premium owed, expressed in asset units and scaled by RAY.
+  function getAssetPremiumOwedRay(uint256 assetId) external view returns (uint256);
```
miguelmtzinf: I think `getAssetPremiumRay` is fine?

src/hub/interfaces/IHubBase.sol
```diff
@@ -246,6 +247,11 @@ interface IHubBase {
   /// @return The total amount of the assets owed.
   function getAssetTotalOwed(uint256 assetId) external view returns (uint256);
 
+  /// @notice Returns the full precision premium of the specified asset owed to the Hub.
+  /// @param assetId The identifier of the asset.
+  /// @return The amount of premium owed, expressed in asset units and scaled by RAY.
```
miguelmtzinf, line 250:
```suggestion
  /// @return The amount of owed premium assets, expressed in asset units and scaled by RAY.
```

src/hub/interfaces/IHubBase.sol
```diff
@@ -246,6 +247,11 @@ interface IHubBase {
   /// @return The total amount of the assets owed.
   function getAssetTotalOwed(uint256 assetId) external view returns (uint256);
 
+  /// @notice Returns the full precision premium of the specified asset owed to the Hub.
```
miguelmtzinf:
```suggestion
  /// @notice Returns the amount of owed premium with full precision for specified asset.
```

src/hub/interfaces/IHubBase.sol
```diff
@@ -295,6 +301,12 @@ interface IHubBase {
   /// @return The total amount of the asset owed.
   function getSpokeTotalOwed(uint256 assetId, address spoke) external view returns (uint256);
 
+  /// @notice Returns the full precision premium of the specified asset owed to the Hub by the specified spoke.
+  /// @param assetId The identifier of the asset.
+  /// @param spoke The address of the spoke.
+  /// @return The amount of premium owed, expressed in asset units and scaled by RAY.
```
miguelmtzinf:
```suggestion
  /// @notice Returns the amount of owed premium with full precision for specified asset ans spoke.
  /// @param assetId The identifier of the asset.
  /// @param spoke The address of the spoke.
  /// @return The amount of owed premium assets, expressed in asset units and scaled by RAY.
```
don't think we need `owed to the Hub` or `added to the Hub`... I see that in the other functions.

avniculae: [2e5a290](https://github.com/aave/aave-v4/pull/991/commits/2e5a2904fe82f3fb196c430005ebf5fd347b5dea)

src/hub/interfaces/IHubBase.sol
avniculae: [2e5a290](https://github.com/aave/aave-v4/pull/991/commits/2e5a2904fe82f3fb196c430005ebf5fd347b5dea)

src/hub/interfaces/IHubBase.sol
avniculae: [2e5a290](https://github.com/aave/aave-v4/pull/991/commits/2e5a2904fe82f3fb196c430005ebf5fd347b5dea)

src/hub/interfaces/IHubBase.sol
avniculae: [2e5a290](https://github.com/aave/aave-v4/pull/991/commits/2e5a2904fe82f3fb196c430005ebf5fd347b5dea)

src/hub/interfaces/IHub.sol
avniculae: [2e5a290](https://github.com/aave/aave-v4/pull/991/commits/2e5a2904fe82f3fb196c430005ebf5fd347b5dea)

src/hub/libraries/Premium.sol
```diff
@@ -0,0 +1,30 @@
...
+
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
+
+/// @title Premium library
+/// @author Aave Labs
+/// @notice Implements the premium calculations.
+library Premium {
+  using WadRayMath for uint256;
+
+  function calculateAccruedPremiumRay(
```
miguelmtzinf, line 14: add docs to these two pls

src/hub/libraries/Premium.sol
```diff
@@ -0,0 +1,30 @@
...
+    uint256 premiumShares,
+    uint256 drawnIndex,
+    uint256 premiumOffsetRay
+  ) internal pure returns (uint256) {
+    return premiumShares * drawnIndex - premiumOffsetRay;
+  }
+
+  function calculatePremiumDebtRay(
```
miguelmtzinf:
```suggestion
  function calculatePremiumRay(
```

src/hub/libraries/Premium.sol
```diff
@@ -0,0 +1,30 @@
...
+    uint256 realizedPremiumRay,
```
miguelmtzinf: better to put it first or last?

src/hub/libraries/Premium.sol
```diff
@@ -0,0 +1,30 @@
...
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.20;
+
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
+
+/// @title Premium library
+/// @author Aave Labs
+/// @notice Implements the premium calculations.
+library Premium {
+  using WadRayMath for uint256;
```
miguelmtzinf: not needed, rm import

src/hub/libraries/AssetLogic.sol
```diff
@@ -174,12 +182,27 @@ library AssetLogic {
...
 
-    uint256 liquidityGrowth = drawnShares.rayMulUp(drawnIndex) -
-      drawnShares.rayMulUp(previousIndex) +
-      premiumShares.rayMulUp(drawnIndex) -
-      premiumShares.rayMulUp(previousIndex);
-    return liquidityGrowth.percentMulDown(liquidityFee);
+    uint256 realizedPremiumRay = asset.realizedPremiumRay;
+    uint120 premiumShares = asset.premiumShares;
+    uint256 premiumOffsetRay = asset.premiumOffsetRay;
+    uint256 premiumDebtRayAfter = Premium.calculatePremiumDebtRay(
```
miguelmtzinf: `premiumRayAfter` instead? or `totalPremiumRayAfter`

src/hub/Hub.sol
miguelmtzinf, line 316: this event remains accurate wrt to `premiumAmount`, but Spoke is not because we are emitting the rounded up value? well, it's the amount that the user transfers anyway so it's correct?

src/hub/Hub.sol
```diff
@@ -316,19 +317,20 @@ contract Hub is IHub, AccessManaged {
...
-    _validateReportDeficit(asset, spoke, drawnAmount, premiumAmount);
+    _validateReportDeficit(asset, spoke, drawnAmount, premiumDelta.restoredPremiumRay);
 
     uint120 drawnShares = asset.toDrawnSharesDown(drawnAmount).toUint120();
     asset.drawnShares -= drawnShares;
     spoke.drawnShares -= drawnShares;
-    _applyPremiumDelta(asset, spoke, premiumDelta, premiumAmount);
+    _applyPremiumDelta(asset, spoke, premiumDelta);
+
+    uint256 premiumAmount = premiumDelta.restoredPremiumRay.fromRayUp();
```
miguelmtzinf: we are over-reporting here, which goes against the protocol. After some time it can happen that X was reported but only Y << X was true bad debt (delta due to rounding-up operations across time).

src/hub/Hub.sol
```diff
@@ -776,13 +809,25 @@ contract Hub is IHub, AccessManaged {
     return asset.toDrawnAssetsUp(spoke.drawnShares);
   }
 
+  function _getSpokePremiumRay(
```
miguelmtzinf, line 821:
```suggestion
  /// @dev Returns the spoke's premium amount with full precision for a specified asset.
  function _getSpokePremiumRay(
```
maybe place it after `_getSpokePremium`

src/hub/libraries/Premium.sol
avniculae: I prefer first, since realized is from the past

src/hub/Hub.sol
avniculae: spoke is emitting both premiumDebtRestored and the ray version. yes, I think it's still correct since that is the asset amount that is trasferred

src/hub/Hub.sol
avniculae: [9c25e4c](https://github.com/aave/aave-v4/pull/991/commits/9c25e4ce9236f6e412382b316ca8e0b05b1205de)

src/hub/Hub.sol
avniculae: [9c25e4c](https://github.com/aave/aave-v4/pull/991/commits/9c25e4ce9236f6e412382b316ca8e0b05b1205de)

src/hub/libraries/AssetLogic.sol
avniculae: [9c25e4c](https://github.com/aave/aave-v4/pull/991/commits/9c25e4ce9236f6e412382b316ca8e0b05b1205de)

src/hub/libraries/Premium.sol
avniculae: [9c25e4c](https://github.com/aave/aave-v4/pull/991/commits/9c25e4ce9236f6e412382b316ca8e0b05b1205de)

src/hub/libraries/Premium.sol
avniculae: [9c25e4c](https://github.com/aave/aave-v4/pull/991/commits/9c25e4ce9236f6e412382b316ca8e0b05b1205de)

src/hub/libraries/Premium.sol
avniculae: [9c25e4c](https://github.com/aave/aave-v4/pull/991/commits/9c25e4ce9236f6e412382b316ca8e0b05b1205de)

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -225,4 +225,11 @@ interface ISpokeBase {
   /// @param user The address of the user.
   /// @return The total debt amount.
   function getUserTotalDebt(uint256 reserveId, address user) external view returns (uint256);
+
+  /// @notice Returns the full precision premium debt of a specific user for a given reserve.
+  /// @dev It reverts if the reserve associated with the given reserve identifier is not listed.
+  /// @param reserveId The identifier of the reserve.
+  /// @param user The address of the user.
+  /// @return The amount of premium debt, expressed in asset units and scaled by RAY.
+  function getUserPremiumDebtRay(uint256 reserveId, address user) external view returns (uint256);
```
miguelmtzinf, line 234: we don't have `getUserPremiumData`?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -413,12 +409,10 @@ library LiquidationLogic {
 
   /// @notice Calculates the liquidation amounts.
   /// @dev Invoked by `liquidateUser` method.
-  /// @return The collateral to liquidate.
-  /// @return The collateral to transfer to liquidator.
-  /// @return The debt to liquidate.
+  /// @return The liquidation amounts.
```
miguelmtzinf: we can probably rm

src/spoke/Spoke.sol
```diff
@@ -992,16 +1017,18 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
   function _calculateRestoreAmount(
     uint256 drawnDebt,
-    uint256 premiumDebt,
+    uint256 premiumDebtRay,
     uint256 amount
   ) internal pure returns (uint256, uint256) {
+    uint256 premiumDebt = premiumDebtRay.fromRayUp();
     if (amount >= drawnDebt + premiumDebt) {
```
miguelmtzinf, line 1050:
```suggestion
    if (amount >= drawnDebt + premiumDebtRay.fromRayUp()) {
```
we can do this so it's clear we only need this for comparing and detect a max repay. (we can also leave a quick comment)

src/spoke/Spoke.sol
```diff
@@ -992,16 +1017,18 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-      return (drawnDebt, premiumDebt);
+      return (drawnDebt, premiumDebtRay);
     }
-    if (amount <= premiumDebt) {
-      return (0, amount);
+    uint256 amountRay = amount.toRay();
```
```suggestion

    uint256 amountRay = amount.toRay();
```
just a breakline here. If not max repay, we do other stuff

src/hub/libraries/Premium.sol
miguelmtzinf: i thought it could be error-prone we have only 2 functions which are similar wrt to premium, and easy to pass params in the wrong order. If we put `realizedPremiumRay` at the end, both will start with `premiumShares -  drawnIndex -  offset`

src/spoke/interfaces/ISpokeBase.sol
avniculae: we have `getUserPosition`, which contains the premium data

src/spoke/Spoke.sol
avniculae: we use it below as well

src/hub/Hub.sol
DhairyaSethi: backend will require full precision amounts

src/spoke/Spoke.sol
avniculae: [9fdd713](https://github.com/aave/aave-v4/pull/991/commits/9fdd7138dd86fd194c55d6cde40459dc9bc3cc98)

src/spoke/Spoke.sol
avniculae: refactored fn a bit [9fdd713](https://github.com/aave/aave-v4/pull/991/commits/9fdd7138dd86fd194c55d6cde40459dc9bc3cc98)

src/spoke/libraries/LiquidationLogic.sol
avniculae: [9fdd713](https://github.com/aave/aave-v4/pull/991/commits/9fdd7138dd86fd194c55d6cde40459dc9bc3cc98)

src/hub/libraries/Premium.sol
avniculae: [9fdd713](https://github.com/aave/aave-v4/pull/991/commits/9fdd7138dd86fd194c55d6cde40459dc9bc3cc98)

src/hub/interfaces/IHub.sol
```diff
@@ -36,8 +36,9 @@ interface IHub is IHubBase, IAccessManaged {
     uint120 deficit;
     uint120 swept;
     //
-    uint120 realizedPremium;
-    uint120 premiumOffset;
+    uint256 realizedPremiumRay;
+    //
+    uint256 premiumOffsetRay;
```
DhairyaSethi: these can be smaller, very conservatively in 196 bits otherwise slightly higher; let's pack them to the nearest uint variant available. it doesn't really matter here but might be useful for future spoke implementations to pack stuff in the remaining couple of words

src/hub/libraries/AssetLogic.sol
```diff
@@ -174,12 +182,26 @@ library AssetLogic {
...
+      premiumOffsetRay: premiumOffsetRay,
+      realizedPremiumRay: realizedPremiumRay
+    });
+    uint256 premiumRayBefore = Premium.calculatePremiumRay({
+      premiumShares: premiumShares,
+      drawnIndex: previousIndex,
+      premiumOffsetRay: premiumOffsetRay,
+      realizedPremiumRay: realizedPremiumRay
+    });
+    uint256 liquidityGrowthPremium = premiumRayAfter.fromRayUp() - premiumRayBefore.fromRayUp();
```
DhairyaSethi, line 204: can you try in the z3 script if we over estimate if we are to divUp(ray) after the subtraction? i think what you have here is correct but just for good measure

src/libraries/math/WadRayMath.sol
```diff
@@ -140,14 +140,37 @@ library WadRayMath {
     }
   }
 
+  /// @notice Casts value to RAY, adding 27 digits of precision.
+  /// @dev Reverts if intermediate multiplication overflows.
+  /// @return b = a * RAY in RAY units.
+  function toRay(uint256 a) internal pure returns (uint256 b) {
+    // to avoid overflow, b/RAY == a
```
DhairyaSethi: this comment needs to go below where we div(b, ray)

src/spoke/Spoke.sol
```diff
@@ -286,34 +287,36 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    uint256 restoredShares = reserve.hub.restore(
-      reserve.assetId,
-      drawnDebtRestored,
-      premiumDebtRestored,
-      premiumDelta
-    );
+    uint256 restoredShares = reserve.hub.restore(reserve.assetId, drawnDebtRestored, premiumDelta);
 
-    userPosition.settlePremiumDebt(premiumDelta.realizedDelta);
+    userPosition.settlePremiumDebt(premiumDelta.accruedPremiumRay, premiumDelta.restoredPremiumRay);
```
DhairyaSethi: i think it's fine now to just pass the struct even if we use half of it to make this cleaner, we didnt earlier bc we only used one third

src/spoke/Spoke.sol
```diff
@@ -992,16 +1017,20 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
   function _calculateRestoreAmount(
```
DhairyaSethi, line 1044: let add natspec above it returns base amount to restore & premium amount in ray to restore

tests/Base.t.sol
```diff
@@ -1514,6 +1514,28 @@ abstract contract Base is Test {
...
+  function _calculateRestoreAmounts(
+    uint256 restoreAmount,
+    uint256 drawn,
+    uint256 premium
+  ) internal view returns (uint256 baseAmount, uint256 premiumAmount) {
+    if (restoreAmount <= premium) {
+      return (0, restoreAmount);
+    }
+
+    return (drawn.min(restoreAmount - premium), premium);
```
DhairyaSethi, line 1522: is this logic any cheaper in general? it avoids one branch at the expense of an extra safe sub

avniculae: `min` also has a branch inside, so it's the same number of branches no?

src/libraries/math/WadRayMath.sol
avniculae: this follows toWad(). should I change both?

src/hub/Hub.sol
avniculae: we have full precision amounts in premiumDelta in both Spoke.repay and Hub.restore

src/spoke/Spoke.sol
avniculae: [5367c3e](https://github.com/aave/aave-v4/pull/991/commits/5367c3e5f9f7af9dc583cc3bc15c9c3f5b3038e8)

src/hub/libraries/AssetLogic.sol
avniculae: [7dbdaf2](https://github.com/aave/aave-v4/pull/991/commits/7dbdaf2f273231daec29472479c33d81423bbbbf)

src/spoke/Spoke.sol
avniculae: at that point I'd call it applyPremiumDelta, and use all fields, wdyt?

src/libraries/math/WadRayMath.sol
DhairyaSethi: yes

tests/Base.t.sol
DhairyaSethi: min is branchless through the xor trick

avniculae: ah correct. I'd still keep function as is in src/, there are edge cases due to mixing asset and ray values.

src/spoke/Spoke.sol
avniculae: using all fields exceeds spoke contract size

DhairyaSethi: with apply naming we'll have to apply the shares & offset deltas as well and it's inefficient bc of two additional warm sloads 


let's see it trimming nonces keyed helps w code size

avniculae: I think it's error-prone to pass PremiumDelta and not use the share and offset deltas

DhairyaSethi: disagree its fine imo & cleaner

src/hub/interfaces/IHub.sol
avniculae: [f1f547a](https://github.com/aave/aave-v4/pull/991/commits/f1f547a147bd3c44b3375d31c1d353a22ec6dceb)

src/libraries/math/WadRayMath.sol
avniculae: [fb4cbbf](https://github.com/aave/aave-v4/pull/991/commits/fb4cbbf076ba8bff0204096c06b725eb41016d2c)

src/spoke/Spoke.sol
avniculae: lmk if you like this [ff113f2](https://github.com/aave/aave-v4/pull/991/commits/ff113f2a29678d92385ce826af6eabc896a1a32d)

src/hub/interfaces/IHub.sol
```diff
@@ -36,8 +36,9 @@ interface IHub is IHubBase, IAccessManaged {
...
     uint120 swept;
     //
-    uint120 realizedPremium;
-    uint120 premiumOffset;
+    uint200 realizedPremiumRay;
+    //
+    uint200 premiumOffsetRay;
     //
     uint16 liquidityFee;
     uint120 drawnShares;
```
DhairyaSethi, line 43: this separation isn't true anymore, will need to put drawnShares first

avniculae: [ae2554b](https://github.com/aave/aave-v4/pull/991/commits/ae2554b0c3475267326f9e63368a4356e29727d3)

DhairyaSethi: @avniculae pls fix order on natspec also

avniculae: [6f529bb](https://github.com/aave/aave-v4/pull/991/commits/6f529bbd4f1350a093e602c568a7c72c5fec48d6)

src/spoke/Spoke.sol
miguelmtzinf: alinged with @avniculae here, passing smt and not using it is weird and raises questions about why it's passed and not used

DhairyaSethi: it's implicitly passed here, the commit looks cleaner but is 1k more expensive bc the additional sloads i mentioned above. i'm agnostic

src/spoke/Spoke.sol
```diff
@@ -990,18 +1003,28 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     require(config.liquidationFee <= PercentageMath.PERCENTAGE_FACTOR, InvalidLiquidationFee());
   }
 
+  /// @dev Calculates the drawn debt and premium debt to restore for the given amount.
+  /// @param drawnDebt The maximum amount of drawn debt that can be restored.
+  /// @param premiumDebtRay The maximum amount of premium debt that can be restored, expressed in asset units and scaled by RAY.
+  /// @param amount The amount to restore.
+  /// @return The amount of drawn debt to restore.
```
```suggestion
  /// @return The amount of drawn debt to restore, expressed in asset units.
```

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -70,26 +70,26 @@ interface ISpokeBase {
...
   /// @param collateralSharesToLiquidate The total amount of collateral shares to liquidate.
   /// @param collateralSharesToLiquidator The amount of collateral shares that the liquidator received.
-  /// @param liquidator The address of the liquidator.
-  /// @param receiveShares True if the liquidator received collateral in supplied shares rather than underlying assets.
   event LiquidationCall(
     uint256 indexed collateralReserveId,
     uint256 indexed debtReserveId,
     address indexed user,
+    address liquidator,
+    bool receiveShares,
```
miguelmtzinf, line 86: would we like to move it at the end instead?

miguelmtzinf: not possible due to stack too deep.

PR #993, opened 2025-11-10
feat: Compound premium

avniculae:


src/spoke/Spoke.sol
```diff
@@ -287,39 +287,34 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-      realizedPremiumRay,
-      premiumDelta.accruedPremiumRay
-    ) = _getUserDebt(reserve.hub, reserve.assetId, userPosition);
+    uint256 premiumDebtRay;
+    (drawnDebtRestored, premiumDebtRestored, premiumDebtRay) = _getUserDebt(
+      reserve.hub,
+      reserve.assetId,
+      userPosition
+    );
     (drawnDebtRestored, premiumDebtRestored) = _calculateRestoreAmount(
```
avniculae, line 293: maybe we can refactor this to accept premiumDebtRay, similar to liquidations

avniculae: which approach is better?

DhairyaSethi: yeah we should

DhairyaSethi: `restoredPremiumRay: (premiumDebtRestored * WadRayMath.RAY).min(premiumDebtRay)`
currently this logic should happen inside `_calculateRestoreAmount`, no?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -267,17 +264,9 @@ library LiquidationLogic {
   }
 
   /// @notice Settles the premium debt by realizing change in premium and resetting premium shares and offset.
-  function settlePremiumDebt(
```
Kogaroshi: should this still be called "settle" ? now it's just a reset

avniculae:
Closing in favour of #1019

PR #994, opened 2025-11-10
opt: Move realized premium into premium offset

avniculae:


avniculae:
Closing in favour of #1019

PR #1000, opened 2025-11-12
refactor: liquidateUser

avniculae:


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -238,23 +247,18 @@ library LiquidationLogic {
...
     );
 
-    return
-      _evaluateDeficit({
-        isCollateralPositionEmpty: vars.isCollateralPositionEmpty,
-        isDebtPositionEmpty: vars.isDebtPositionEmpty,
-        activeCollateralCount: params.activeCollateralCount,
-        borrowedCount: params.borrowedCount
-      });
+    return _evaluateDeficit(evaluateDeficitParams);
```
DhairyaSethi: if we're encapsulating already we can syntax sugar here deficitParams.evaluate() or something

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -204,30 +198,45 @@ library LiquidationLogic {
...
+    uint256 collateralSharesLiquidated;
+    uint256 collateralSharesToLiquidator;
     (
-      vars.collateralSharesToLiquidate,
-      vars.collateralSharesToLiquidator,
-      vars.isCollateralPositionEmpty
+      collateralSharesLiquidated,
+      collateralSharesToLiquidator,
+      evaluateDeficitParams.isCollateralPositionEmpty
     ) = _liquidateCollateral(
```
DhairyaSethi: not a huge fan of this standard ngl, populate later i feel is a bad practice because it is error prone 

also we said to not keep structs for input params len <= 4 when this was being written, i think we can continue to hold that

yan-man: agree, this was part of the initial motivation for the refactor where we had some structs that were partially populated at diff points

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -70,26 +70,28 @@ interface ISpokeBase {
...
-  /// @param premiumDelta A struct representing the changes to premium debt after liquidation.
-  /// @param collateralToLiquidate The total amount of collateral asset to be liquidated, inclusive of liquidation fee.
-  /// @param collateralSharesToLiquidate The total amount of collateral shares to liquidate.
-  /// @param collateralSharesToLiquidator The amount of collateral shares that the liquidator received.
   /// @param liquidator The address of the liquidator.
   /// @param receiveShares True if the liquidator received collateral in supplied shares rather than underlying assets.
+  /// @param collateralToLiquidate The amount of collateral to liquidate.
+  /// @param collateralToLiquidator The amount of collateral to transfer to the liquidator.
+  /// @param debtToLiquidate The amount of debt to liquidate.
+  /// @param collateralSharesLiquidated The amount of collateral shares liquidated.
```
yan-man: i think a bit inconsistent here - some are future tense (`to liquidate`, `to transfer to the liquidator`) and some are past tense (`shares liquidated`, `...received`)

yan-man: but agnostic on the ordering otherwise. When i added it initially just thought that it made sense to group collaterals together and debts together, rather than assets and shares together

src/spoke/libraries/LiquidationLogic.sol
avniculae: N/A anymore [5318e2b](https://github.com/aave/aave-v4/pull/1000/commits/5318e2b0044dba3d63e23523e5838bf90de39b28)

src/spoke/libraries/LiquidationLogic.sol
avniculae: [5318e2b](https://github.com/aave/aave-v4/pull/1000/commits/5318e2b0044dba3d63e23523e5838bf90de39b28)

src/spoke/interfaces/ISpokeBase.sol
avniculae: N/A anymore [5318e2b](https://github.com/aave/aave-v4/pull/1000/commits/5318e2b0044dba3d63e23523e5838bf90de39b28)

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Base.t.sol
```diff
@@ -417,133 +388,119 @@ contract SpokeLiquidationCallBaseTest is LiquidationLogicBaseTest {
...
-    }
-    {
+
+    if (!params.receiveShares && liquidationMetadata.collateralToLiquidator > 0) {
       vm.expectCall(
-        address(vars.debtHub),
+        address(_hub(params.spoke, params.collateralReserveId)),
         abi.encodeCall(
-          IHubBase.restore,
+          IHubBase.remove,
```
avniculae, line 427: remove expect call was missing

PR #1020, opened 2025-11-24
fix : add flag to block `receiveShares` in liquidation - 1 : extra flag slot

Kogaroshi:
RWA support 1 : 
Add flag to block receiving collateral shares during liquidation if the asset should not allow it.

Option 1 : add extra flag in `ISpoke.reserve`, needing a new slot, and refactor `_validateLiquidationCall()` to reduce cas costs impacts


Kogaroshi:
Closed in favor of #1021

PR #1021, opened 2025-11-24
fix: add ReserveFlags type and new receiveSharesEnabled flag

### LINKED ISSUE #1026: Prevent receiving collateral shares during liquidation for certain assets
For certains asset that might be listed (for example some RWAs), we might encounter restriction to how liquidation must be operated, and in this case, prevent the liquidator to receive shares of RWA collaterals when liquidating.

To allow us to block that behavior on certain asset, we want to add a new flag to the `Reserve` config in the Spokes, and verify that flag during the validation of the liquidation call parameters.

Kogaroshi:
- Adds flag to block receiving collateral shares during liquidation. This is helpful in scenarios where liquidators should not be able to permissionlessly hold collateral shares, ex. permissioned RWA assets.
- Introduces a new `ReserveFlags` type & lib to pack all booleans of reserve tightly.

Fixes #1026

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -150,12 +150,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       decimals: decimals,
       dynamicConfigKey: dynamicConfigKey,
       collateralRisk: config.collateralRisk,
-      flags: ReserveFlagsMap.create(
-        config.paused,
-        config.frozen,
-        config.borrowable,
-        config.receiveSharesEnabled
-      )
+      flags: ReserveFlagsMap.create({
+        initPaused: config.paused,
+        initFrozen: config.frozen,
+        initBorrowable: config.borrowable,
+        initReceiveSharesEnabled: config.receiveSharesEnabled
+      })
     });
     _dynamicConfig[reserveId][dynamicConfigKey] = dynamicConfig;
     _reserveExists[hub][assetId] = true;
@@ -175,12 +175,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     Reserve storage reserve = _getReserve(reserveId);
     _validateReserveConfig(config);
     reserve.collateralRisk = config.collateralRisk;
-    reserve.flags = ReserveFlagsMap.create(
-      config.paused,
-      config.frozen,
-      config.borrowable,
-      config.receiveSharesEnabled
-    );
+    reserve.flags = ReserveFlagsMap.create({
+      initPaused: config.paused,
+      initFrozen: config.frozen,
+      initBorrowable: config.borrowable,
+      initReceiveSharesEnabled: config.receiveSharesEnabled
+    });
     emit UpdateReserveConfig(reserveId, config);
   }
 
--- a/src/spoke/interfaces/ISpoke.sol
+++ b/src/spoke/interfaces/ISpoke.sol
@@ -3,7 +3,6 @@
 pragma solidity ^0.8.0;
 
 import {IAccessManaged} from 'src/dependencies/openzeppelin/IAccessManaged.sol';
-import {ReserveFlags} from 'src/spoke/libraries/ReserveFlagsMap.sol';
 import {INoncesKeyed} from 'src/interfaces/INoncesKeyed.sol';
 import {IMulticall} from 'src/interfaces/IMulticall.sol';
 import {IHubBase} from 'src/hub/interfaces/IHubBase.sol';
--- a/tests/mocks/ReserveFlagsMapWrapper.sol
+++ b/tests/mocks/ReserveFlagsMapWrapper.sol
@@ -13,7 +13,13 @@ contract ReserveFlagsMapWrapper {
     bool initBorrowable,
     bool initReceiveSharesEnabled
   ) external pure returns (ReserveFlags) {
-    return ReserveFlagsMap.create(initPaused, initFrozen, initBorrowable, initReceiveSharesEnabled);
+    return
+      ReserveFlagsMap.create({
+        initPaused: initPaused,
+        initFrozen: initFrozen,
+        initBorrowable: initBorrowable,
+        initReceiveSharesEnabled: initReceiveSharesEnabled
+      });
   }
 
   function setPaused(ReserveFlags flags, bool status) external pure returns (ReserveFlags) {
--- a/tests/unit/ReserveFlags.t.sol
+++ b/tests/unit/ReserveFlags.t.sol
@@ -31,7 +31,12 @@ contract ReserveFlagsTests is Test {
     bool borrowable,
     bool receiveSharesEnabled
   ) public view {
-    ReserveFlags flags = w.create(paused, frozen, borrowable, receiveSharesEnabled);
+    ReserveFlags flags = w.create({
+      initPaused: paused,
+      initFrozen: frozen,
+      initBorrowable: borrowable,
+      initReceiveSharesEnabled: receiveSharesEnabled
+    });
 
     assertEq(w.paused(flags), paused);
     assertEq(w.frozen(flags), frozen);


src/spoke/interfaces/ISpoke.sol
```diff
@@ -2,6 +2,7 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity ^0.8.0;
 
+import {ReserveFlags} from 'src/spoke/libraries/ReserveFlags.sol';
 import {IAccessManaged} from 'src/dependencies/openzeppelin/IAccessManaged.sol';
```
```suggestion
import {IAccessManaged} from 'src/dependencies/openzeppelin/IAccessManaged.sol';
import {ReserveFlags} from 'src/spoke/libraries/ReserveFlags.sol';
```

src/spoke/interfaces/ISpoke.sol
```diff
@@ -18,9 +19,11 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
...
   /// @dev decimals The number of decimals of the underlying asset.
   /// @dev dynamicConfigKey The key of the last reserve dynamic config.
-  /// @dev paused True if all actions are prevented for the reserve.
-  /// @dev frozen True if new activity is prevented for the reserve.
-  /// @dev borrowable True if the reserve is borrowable.
+  /// @dev flags The packed boolean flags of the reserve.
+  /// @dev flags.paused True if all actions are prevented for the reserve.
+  /// @dev flags.frozen True if new activity is prevented for the reserve.
+  /// @dev flags.borrowable True if the reserve is borrowable.
+  /// @dev flags.canReceiveShares True if the liquidator can receive collateral shares during liquidation.
```
miguelmtzinf: not sure if we should put docs here or move all of them to `ReserveFlags`.
Oh, they can be moved to `ReserveConfig` maybe

src/spoke/interfaces/ISpoke.sol
```diff
@@ -29,9 +32,7 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     uint16 assetId;
     uint8 decimals;
     uint24 dynamicConfigKey;
-    bool paused;
-    bool frozen;
-    bool borrowable;
+    ReserveFlags flags;
```
miguelmtzinf: do not like that we don't know how much space takes. Alternative would be use `uint8` directly here instead.

src/spoke/interfaces/ISpoke.sol
```diff
@@ -300,7 +302,7 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @notice Thrown when a debt to cover input is zero.
   error InvalidDebtToCover();
 
-  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen.
+  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is not allowing it or is frozen.
```
```suggestion
  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen or not configured to allow share receival.
```

src/spoke/interfaces/ISpokeConfigurator.sol
```diff
@@ -91,6 +91,12 @@ interface ISpokeConfigurator {
   /// @param borrowable The new borrowable flag.
   function updateBorrowable(address spoke, uint256 reserveId, bool borrowable) external;
 
+  /// @notice Updates the canReceiveShares flag of a reserve.
```
miguelmtzinf: alternative name can be `receiveSharesEnabled`, so here we could say "Updates whether receiving shares on liquidation is enabled."
We can also say now smt like "Updates whether the liquidator can receive shares on liquidation." or "Updates whether receiving shares on liquidation is allowed."

src/spoke/libraries/ReserveFlags.sol
```diff
@@ -0,0 +1,109 @@
...
+type ReserveFlags is uint8;
+
+using ReserveFlagsLib for ReserveFlags global;
+
+/// @title ReserveFlags Library
+/// @author Aave Labs
+/// @notice Packs all boolean flags of a Reserve as a single uint8.
+library ReserveFlagsLib {
+  /// @dev Mask for the `paused` flag.
+  uint8 internal constant PAUSED_MASK = 0x001;
```
miguelmtzinf:
```suggestion
  uint8 internal constant PAUSED_MASK = 0x01;
```
is it clearer to do this for all? 
only 2 digits in hexadecimal because we take 8 bits

src/spoke/libraries/ReserveFlags.sol
```diff
@@ -0,0 +1,109 @@
...
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.20;
+
+type ReserveFlags is uint8;
+
+using ReserveFlagsLib for ReserveFlags global;
+
+/// @title ReserveFlags Library
+/// @author Aave Labs
+/// @notice Packs all boolean flags of a Reserve as a single uint8.
```
miguelmtzinf:
```suggestion
/// @notice Implements the bitmap logic to handle the Reserve flags configuration.
```

src/spoke/Spoke.sol
```diff
@@ -14,6 +14,7 @@ import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {KeyValueList} from 'src/spoke/libraries/KeyValueList.sol';
 import {LiquidationLogic} from 'src/spoke/libraries/LiquidationLogic.sol';
 import {PositionStatusMap} from 'src/spoke/libraries/PositionStatusMap.sol';
+import {ReserveFlags, ReserveFlagsLib} from 'src/spoke/libraries/ReserveFlags.sol';
```
miguelmtzinf: do we need `ReserveFlags` import?

tests/unit/libraries/LiquidationLogic/LiquidationLogic.ValidateLiquidationCall.t.sol
```diff
@@ -50,55 +51,63 @@ contract LiquidationLogicValidateLiquidationCallTest is LiquidationLogicBaseTest
...
 
-    // receiveShares = true; liquidatorUsingAsCollateral = true; frozen = true; => revert
+    // receiveShares = true; liquidatorUsingAsCollateral = true; frozen = true; canReceiveShares = true; => revert
     params.receiveShares = true;
     liquidationLogicWrapper.setLiquidatorCollateralStatus(collateralReserveId, true);
     params.collateralReserveFrozen = true;
     vm.expectRevert(ISpoke.CannotReceiveShares.selector);
     liquidationLogicWrapper.validateLiquidationCall(params);
+
+    // receiveShares = true; liquidatorUsingAsCollateral = true; frozen = false; canReceiveShares = false; => revert
```
miguelmtzinf: missing:
    // receiveShares = true; liquidatorUsingAsCollateral = true; frozen = true; canReceiveShares = false; => revert

src/spoke/interfaces/ISpoke.sol
Kogaroshi: Could we comment the size ? I find it a bit messy and unclear to use the Lib on a raw `uint8`

src/spoke/interfaces/ISpoke.sol
Kogaroshi: Agree, gonna move it to `ReserveConfig`

src/spoke/Spoke.sol
Kogaroshi: not needed, removed

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -8,6 +8,7 @@ import {MathUtils} from 'src/libraries/math/MathUtils.sol';
 import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
 import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {PositionStatusMap} from 'src/spoke/libraries/PositionStatusMap.sol';
+import {ReserveFlags, ReserveFlagsLib} from 'src/spoke/libraries/ReserveFlags.sol';
```
```suggestion
import {ReserveFlagsLib} from 'src/spoke/libraries/ReserveFlags.sol';
```

src/spoke/libraries/ReserveFlags.sol
```diff
@@ -0,0 +1,110 @@
...
+pragma solidity ^0.8.20;
+
+type ReserveFlags is uint8;
+
+using ReserveFlagsLib for ReserveFlags global;
+
+/// @title ReserveFlags Library
+/// @author Aave Labs
+/// @notice Implements the bitmap logic to handle the Reserve flags configuration.
+library ReserveFlagsLib {
```
miguelmtzinf: we can rename to `ReserveFlagsMap`, similar to `PositionStatusMap`
let's rename file too

src/spoke/interfaces/ISpoke.sol
```diff
@@ -300,7 +303,7 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @notice Thrown when a debt to cover input is zero.
   error InvalidDebtToCover();
 
-  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen.
+  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen or not configured to allow share receival.
```
```suggestion
  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen or where receiving shares is not enabled.
```
oh, think this is clearer

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -0,0 +1,110 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.20;
+
+type ReserveFlags is uint8;
+
+using ReserveFlagsMap for ReserveFlags global;
```
miguelmtzinf: is there another way to do this, more aligned with how we use other of libraries? thinking that `using global` is a one-off in the code

Kogaroshi: It prevents from importing both the type and the lib in every file, and repeating the `using` everywhere.
We currently don't have convention for those defined types, so if we want we can adapt.

Kogaroshi: It prevents from importing both the type and the lib in every file, and repeating the `using` everywhere.
We currently don't have convention for those defined types, so if we want we can adapt.

src/spoke/interfaces/ISpoke.sol
yan-man: wdym by comment the size?

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -0,0 +1,110 @@
...
+  /// @param _paused The `pause` flag status.
+  /// @param _frozen The `frozen` flag status.
+  /// @param _borrowable The `borrowable` flag status.
+  /// @param _receiveSharesEnabled The `receiveSharesEnabled` flag status.
+  /// @return The initialized ReserveFlags.
+  function initFlags(
+    bool _paused,
+    bool _frozen,
+    bool _borrowable,
+    bool _receiveSharesEnabled
```
yan-man: i dont think we've used this `_var` convention for arg inputs in other methods

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -0,0 +1,110 @@
...
+    flags = _setFlag(flags, PAUSED_MASK, _paused);
+    flags = _setFlag(flags, FROZEN_MASK, _frozen);
+    flags = _setFlag(flags, BORROWABLE_MASK, _borrowable);
+    flags = _setFlag(flags, RECEIVE_SHARES_ENABLED_MASK, _receiveSharesEnabled);
+    return ReserveFlags.wrap(flags);
+  }
+
+  /// @notice Sets the new status for the 'paused' flag.
+  /// @param flags The current ReserveFlags.
+  /// @param status The new status for the 'paused' flag.
```
yan-man: i believe these should be code blocks `` rather than quotes '', same below

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -0,0 +1,110 @@
...
+  /// @param flags The current ReserveFlags.
+  /// @return True if the flag is set.
+  function borrowable(ReserveFlags flags) internal pure returns (bool) {
+    return (ReserveFlags.unwrap(flags) & BORROWABLE_MASK) != 0;
+  }
+
+  /// @notice Returns the 'receiveSharesEnabled' flag status.
+  /// @param flags The current ReserveFlags.
+  /// @return True if the flag is set.
+  function receiveSharesEnabled(ReserveFlags flags) internal pure returns (bool) {
```
yan-man, line 102: nit: these could be named `isBorrowable`, etc to align with v3, ex. https://github.com/aave-dao/aave-v3-origin/blob/1ce897ba99d5a9ab659861b591576cd4278e9e27/src/contracts/protocol/libraries/configuration/UserConfiguration.sol#L94

src/spoke/Spoke.sol
```diff
@@ -168,9 +172,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    reserve.paused = config.paused;
-    reserve.frozen = config.frozen;
-    reserve.borrowable = config.borrowable;
+    reserve.flags = ReserveFlagsMap.setPaused(reserve.flags, config.paused);
+    reserve.flags = ReserveFlagsMap.setFrozen(reserve.flags, config.frozen);
+    reserve.flags = ReserveFlagsMap.setBorrowable(reserve.flags, config.borrowable);
+    reserve.flags = ReserveFlagsMap.setReceiveSharesEnabled(
+      reserve.flags,
+      config.receiveSharesEnabled
+    );
```
yan-man: do you think worth it to create a helper method like setFlags (similar to initFlags) so that these can be set all at once rather than calling it 4 times, or do you prefer verbosity here

src/spoke/Spoke.sol
```diff
@@ -931,32 +940,32 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   }
 
   function _validateSupply(Reserve storage reserve) internal view {
```
yan-man: on these validation methods can just pass in ReserveFlags reserve.flags directly as well

src/spoke/interfaces/ISpoke.sol
```diff
@@ -300,7 +303,7 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @notice Thrown when a debt to cover input is zero.
   error InvalidDebtToCover();
 
-  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen.
+  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen or configured to not allow share receival.
```
yan-man:
```suggestion
  /// @notice Thrown when the liquidator tries to receive shares for a collateral reserve that is frozen or is not enabled to receive shares.
```

tests/unit/ReserveFlags.t.sol
```diff
@@ -0,0 +1,170 @@
...
+      receiveSharesEnabled
+    );
+
+    assertEq(ReserveFlagsMap.paused(flags), paused);
+    assertEq(ReserveFlagsMap.frozen(flags), frozen);
+    assertEq(ReserveFlagsMap.borrowable(flags), borrowable);
+    assertEq(ReserveFlagsMap.receiveSharesEnabled(flags), receiveSharesEnabled);
+  }
+
+  function test_set_flags() public pure {
```
yan-man: similar to `test_initFlags_fuzz` i think this one can just be a fuzz

tests/unit/ReserveFlags.t.sol
```diff
@@ -0,0 +1,170 @@
...
+pragma solidity ^0.8.0;
+
+import {Test} from 'forge-std/Test.sol';
+import {ReserveFlags, ReserveFlagsMap} from 'src/spoke/libraries/ReserveFlagsMap.sol';
+
+contract ReserveFlagsTests is Test {
+  uint8 internal constant PAUSED_MASK = 0x001;
+  uint8 internal constant FROZEN_MASK = 0x002;
+  uint8 internal constant BORROWABLE_MASK = 0x004;
+  uint8 internal constant RECEIVE_SHARES_ENABLED_MASK = 0x008;
```
yan-man: should we update this similarly to 2 digit `0x01`, etc?

tests/unit/ReserveFlags.t.sol
```diff
@@ -0,0 +1,170 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {Test} from 'forge-std/Test.sol';
+import {ReserveFlags, ReserveFlagsMap} from 'src/spoke/libraries/ReserveFlagsMap.sol';
+
+contract ReserveFlagsTests is Test {
+  uint8 internal constant PAUSED_MASK = 0x001;
```
yan-man: worth it to write some trivial tests to check that these MASK constants match the ones from the lib?

tests/unit/ReserveFlags.t.sol
```diff
@@ -0,0 +1,170 @@
...
+
+    expectedRawFlags = expectedRawFlags & ~RECEIVE_SHARES_ENABLED_MASK;
+
+    flags = ReserveFlagsMap.setReceiveSharesEnabled(flags, false);
+
+    assertEq(ReserveFlagsMap.receiveSharesEnabled(flags), false);
+    assertEq(ReserveFlags.unwrap(flags), expectedRawFlags);
+  }
+
+  function _sanitizeFlags(uint8 rawFlags) internal pure returns (ReserveFlags) {
```
yan-man, line 180: can you write a small comment for this helper to explain it 

how do you mean by rawFlags, it's the unwrapped ones?

src/spoke/libraries/ReserveFlagsMap.sol
Kogaroshi: we had `isBorrowing` & `isUsingAsCollateral` which are now merged as `getUserReserveStatus`.
and here for the ReserveFlag, we were not using the "is__" naming when the booleans were simply in the struct, so didn't see a need to change the naming logic here.

src/spoke/Spoke.sol
Kogaroshi: I prefer the verbosity, but if needed we can add another method to change all the flags at once seeing how we currently use the lib. No strong convictions here.

tests/unit/ReserveFlags.t.sol
Kogaroshi: the `initFlags` test is already a fuzz, as well as each setter test, I was seeing this one more as a test to ensure changing one of the flag wouldn't impact the other ones, didn't see the need to fuzz here since we simply have `True/False` values.

src/spoke/interfaces/ISpoke.sol
Kogaroshi: https://github.com/aave/aave-v4/blob/64e7605457a1732177daf281ee5b229691d80662/src/spoke/interfaces/ISpoke.sol#L22

tests/unit/ReserveFlags.t.sol
Kogaroshi: No, it's a random uint8 here, except only 4 bits out of 8 are needed here, so we remove the other bits irrelevant to our tests here.
Will add a comment here.

tests/unit/ReserveFlags.t.sol
Kogaroshi: It's kind of done implicitly during the other tests via the `expectedRawFlags` checks, don't know if more verification is needed ?

src/spoke/Spoke.sol
Kogaroshi: fd75418bc8fd051f08e22bc1cace2f2285e4e312

src/spoke/libraries/ReserveFlagsMap.sol
yan-man: fairly agnostic nbd either way

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -10,32 +10,32 @@ using ReserveFlagsMap for ReserveFlags global;
...
 
   /// @notice Initializes the ReserveFlags with the given values.
-  /// @param _paused The `pause` flag status.
-  /// @param _frozen The `frozen` flag status.
-  /// @param _borrowable The `borrowable` flag status.
-  /// @param _receiveSharesEnabled The `receiveSharesEnabled` flag status.
+  /// @param initPaused The initial 'paused' flag status.
+  /// @param initFrozen The initial 'frozen' flag status.
+  /// @param initBorrowable The initial 'borrowable' flag status.
+  /// @param initReceiveSharesEnabled The initial 'receiveSharesEnabled' flag status.
```
yan-man: also `` comment block quotes here?

tests/unit/ReserveFlags.t.sol
yan-man: minor but think it'd be safer to have, ie `test_constants` tests we have for other libs

Kogaroshi: okay, will add a wrapper for the lib and extra test

yan-man (approved):

src/spoke/interfaces/ISpoke.sol
```diff
@@ -29,17 +28,21 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     uint16 assetId;
     uint8 decimals;
     uint24 dynamicConfigKey;
-    bool paused;
-    bool frozen;
-    bool borrowable;
+    ReserveFlags flags;
```
miguelmtzinf: should we move to the end in case we wanna expand it later?

src/spoke/libraries/ReserveFlagsMap.sol
avniculae: I'd prefer to repeat `using` and import the type whenever is needed just bc this is what we do for the other libraries as well, but nbd

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -0,0 +1,110 @@
...
+  ) internal pure returns (ReserveFlags) {
+    return
+      ReserveFlags.wrap(_setFlag(ReserveFlags.unwrap(flags), RECEIVE_SHARES_ENABLED_MASK, status));
+  }
+
+  /// @notice Returns the `paused` flag status.
+  /// @param flags The current ReserveFlags.
+  /// @return True if the flag is set.
+  function paused(ReserveFlags flags) internal pure returns (bool) {
+    return (ReserveFlags.unwrap(flags) & PAUSED_MASK) != 0;
```
avniculae, line 82: we could create a helper function
`function getStatus(flags, mask) returns bool`

src/spoke/interfaces/ISpoke.sol
miguelmtzinf: if moving, not sure if worth re-ordering `ReserveConfig` too

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -0,0 +1,110 @@
...
+  /// @notice Returns the `receiveSharesEnabled` flag status.
+  /// @param flags The current ReserveFlags.
+  /// @return True if the flag is set.
+  function receiveSharesEnabled(ReserveFlags flags) internal pure returns (bool) {
+    return (ReserveFlags.unwrap(flags) & RECEIVE_SHARES_ENABLED_MASK) != 0;
+  }
+
+  /// @notice Sets the new status for the given flag.
+  function _setFlag(uint8 flags, uint8 mask, bool status) internal pure returns (uint8) {
+    return status ? flags | mask : flags & ~mask;
```
avniculae, line 108: we could also do
```
return getStatus(flags, mask) != status ? flags ^ mask : flags
```
probably the same from gas perspective?

src/spoke/Spoke.sol
avniculae: I don't follow, why don't we just do `reserve.flags = initFlags` here as well?

src/spoke/interfaces/ISpoke.sol
Kogaroshi: we can move it down yeah, if we don't see anything else that might need bigger size in the future

src/spoke/Spoke.sol
```diff
@@ -223,7 +231,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   ) external onlyPositionManager(onBehalfOf) returns (uint256, uint256) {
     Reserve storage reserve = _getReserve(reserveId);
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
-    _validateSupply(reserve);
+    _validateSupply(reserve.flags);
```
avniculae, line 235: noice

src/spoke/libraries/ReserveFlagsMap.sol
Kogaroshi: I'll try it

tests/unit/libraries/LiquidationLogic/LiquidationLogic.ValidateLiquidationCall.t.sol
```diff
@@ -50,55 +51,71 @@ contract LiquidationLogicValidateLiquidationCallTest is LiquidationLogicBaseTest
...
 
-    // receiveShares = true; liquidatorUsingAsCollateral = true; frozen = true; => revert
+    // receiveShares = true; liquidatorUsingAsCollateral = true; frozen = true; receiveSharesEnabled = true; => revert
     params.receiveShares = true;
     liquidationLogicWrapper.setLiquidatorCollateralStatus(collateralReserveId, true);
     params.collateralReserveFrozen = true;
     vm.expectRevert(ISpoke.CannotReceiveShares.selector);
     liquidationLogicWrapper.validateLiquidationCall(params);
+
+    // receiveShares = true; liquidatorUsingAsCollateral = true; frozen = false; receiveSharesEnabled = false; => revert
```
avniculae, line 140: we had all possible combinations here (2^3). Should we have all 2^4 combinations now?

src/spoke/Spoke.sol
Kogaroshi: We could, just found it a bit uglier to "re-init" the flags.
Otherwise we can rename to `buildFlags()` and use that at once if preferred

avniculae: I think we should use the same function, we could rename to `getFlags`, wdyt?

Kogaroshi: "get" keyword is not great, as we mainly wrap from a new uint, we don't fetch anything

avniculae: I'm fine with calling it even `init` tbh (just init, not initFlags)

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -157,15 +158,16 @@ library LiquidationLogic {
         debtToCover: params.debtToCover,
         collateralReserveHub: address(collateralReserve.hub),
         debtReserveHub: address(debtReserve.hub),
-        collateralReservePaused: collateralReserve.paused,
-        collateralReserveFrozen: collateralReserve.frozen,
-        debtReservePaused: debtReserve.paused,
+        collateralReservePaused: collateralReserve.flags.paused(),
```
miguelmtzinf: should we just pass all flags? does it make any difference in terms of gas costs/bytecode size?

src/spoke/Spoke.sol
Kogaroshi: Renamed `initFlags` to `create` (since we create a new pack of flags every time), and used in this method now.

src/spoke/libraries/LiquidationLogic.sol
Kogaroshi: Decreases liquidation call gas costs slightly. Changed was made for simplicity too (reduces the number of parameters in `ValidateLiquidationCallParams`).

src/spoke/libraries/ReserveFlagsMap.sol
Kogaroshi: having a helper function here for all those getters increase bytecode size, so this change was discarded in favor of Spoke code size.

src/spoke/libraries/ReserveFlagsMap.sol
Kogaroshi: Here bytecode size is negligable, will apply the change.

Kogaroshi:
New Spoke codesize margin : 443B

src/spoke/libraries/ReserveFlagsMap.sol
avniculae: True, it increases bytecode by 200B, but I think it's worth it. Here's how I'm imagining it: https://github.com/aave/aave-v4/pull/1034

Kogaroshi: Yes, close to the version I tested. Let's keep it on the side, and if we have some codesize margin we can try to merge it in, wdyt ?

avniculae: sg. can you at least make the helper fns (`_setStatus`, `_getStatus`) private?

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -0,0 +1,114 @@
...
+  function receiveSharesEnabled(ReserveFlags flags) internal pure returns (bool) {
+    return (ReserveFlags.unwrap(flags) & RECEIVE_SHARES_ENABLED_MASK) != 0;
+  }
+
+  /// @notice Sets the new status for the given flag.
+  function _setStatus(uint8 flags, uint8 mask, bool status) private pure returns (uint8) {
+    return _getStatus(flags, mask) != status ? flags ^ mask : flags;
+  }
+
+  function _getStatus(uint8 flags, uint8 mask) private pure returns (bool) {
```
miguelmtzinf: only used once so i d rm

tests/mocks/ReserveFlagsMapWrapper.sol
```diff
@@ -0,0 +1,69 @@
...
+    return ReserveFlagsMap.FROZEN_MASK;
+  }
+
+  function BORROWABLE_MASK() external pure returns (uint8) {
+    return ReserveFlagsMap.BORROWABLE_MASK;
+  }
+
+  function RECEIVE_SHARES_ENABLED_MASK() external pure returns (uint8) {
+    return ReserveFlagsMap.RECEIVE_SHARES_ENABLED_MASK;
+  }
```
miguelmtzinf: move to the end

snapshots/Hub.Operations.json
```diff
@@ -11,8 +11,8 @@
   "remove: partial": "83740",
   "reportDeficit": "121107",
   "restore: full": "86184",
-  "restore: full - with transfer": "197201",
+  "restore: full - with transfer": "207201",
   "restore: partial": "94550",
-  "restore: partial - with transfer": "144241",
+  "restore: partial - with transfer": "146741",
```
miguelmtzinf: PR changes does not impact gas costs. This is because of foundry version mismatch.

snapshots/Hub.Operations.json
avniculae, line 1: mention issue number in description, since issue exists

avniculae: also fix PR title pls

src/spoke/interfaces/ISpoke.sol
```diff
@@ -18,28 +19,30 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
...
+  /// @dev frozen True if new activity is prevented for the reserve.
+  /// @dev borrowable True if the reserve is borrowable.
+  /// @dev receiveSharesEnabled True if the liquidator can receive collateral shares during liquidation.
+  /// @dev collateralRisk The risk associated with a collateral asset, expressed in BPS.
   struct ReserveConfig {
     bool paused;
     bool frozen;
     bool borrowable;
+    bool receiveSharesEnabled;
     uint24 collateralRisk;
```
miguelmtzinf: should we put this one first just in case we add more flags later, they don't end up in the middle of the struct?

src/spoke/Spoke.sol
avniculae: can we use named params for all create calls pls?

src/spoke/libraries/ReserveFlagsMap.sol
```diff
@@ -0,0 +1,110 @@
...
+  /// @notice Returns the `receiveSharesEnabled` flag status.
+  /// @param flags The current ReserveFlags.
+  /// @return True if the flag is set.
+  function receiveSharesEnabled(ReserveFlags flags) internal pure returns (bool) {
+    return (ReserveFlags.unwrap(flags) & RECEIVE_SHARES_ENABLED_MASK) != 0;
+  }
+
+  /// @notice Sets the new status for the given flag.
+  function _setStatus(uint8 flags, uint8 mask, bool status) private pure returns (uint8) {
+    return status ? flags | mask : flags & ~mask;
```
avniculae, line 108: why you reverted from flags or flags xor mask?

Kogaroshi: easier to read & less code size until we implement the full refactor

src/spoke/Spoke.sol
Kogaroshi: change made in 8d9a2dfb3df7513ebc05facb53ab5200e5acb170

src/spoke/Spoke.sol
```diff
@@ -16,6 +14,7 @@ import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
 import {KeyValueList} from 'src/spoke/libraries/KeyValueList.sol';
 import {LiquidationLogic} from 'src/spoke/libraries/LiquidationLogic.sol';
 import {PositionStatusMap} from 'src/spoke/libraries/PositionStatusMap.sol';
+import {ReserveFlags, ReserveFlagsMap} from 'src/spoke/libraries/ReserveFlagsMap.sol';
```
miguelmtzinf, line 17: `ReserveFlags` is declared in `ISpoke` but imported from `ReserveFlagsMap` now

src/spoke/Spoke.sol
```diff
@@ -226,7 +232,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   ) external onlyPositionManager(onBehalfOf) returns (uint256, uint256) {
     Reserve storage reserve = _getReserve(reserveId);
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
-    _validateSupply(reserve);
+    _validateSupply(reserve.flags);
```
DhairyaSethi, line 235: these can be moved to flag lib for reserve.flags.validateSupply()

src/spoke/Spoke.sol
```diff
@@ -540,10 +546,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
       ReserveConfig({
-        paused: reserve.paused,
-        frozen: reserve.frozen,
-        borrowable: reserve.borrowable,
-        collateralRisk: reserve.collateralRisk
+        collateralRisk: reserve.collateralRisk,
+        paused: reserve.flags.paused(),
+        frozen: reserve.flags.frozen(),
+        borrowable: reserve.flags.borrowable(),
+        receiveSharesEnabled: reserve.flags.receiveSharesEnabled()
```
DhairyaSethi, line 553: i think you could reduce code size if this read from a cached flags var unless optimiser already does it

miguelmtzinf: tried, it's actually worse...

PR #1024, opened 2025-11-24
opt: remove notify on repay v2

yan-man:
- abstract logic from repay into `_getPremiumDelta` method


src/spoke/Spoke.sol
```diff
@@ -300,54 +302,82 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+    uint256 restoredShares = reserve.hub.previewRestoreByAssets(reserve.assetId, drawnDebtRestored);
+    uint256 totalDebtRestored;
+    {
+      uint256 newDrawnShares = userPosition.drawnShares.uncheckedSub(restoredShares).toUint120();
+      premiumDelta = _getPremiumDelta({
+        userPosition: userPosition,
+        newDrawnShares: newDrawnShares,
+        drawnIndex: drawnIndex,
+        riskPremium: positionStatus.riskPremium,
+        restoredPremiumRay: premiumDelta.restoredPremiumRay
```
yan-man: maybe you won't like this part as it's kind of circular @miguelmtzinf

miguelmtzinf: yeah, ugly and error-prone. I d prefer to return 0 and override right after...

yan-man: see updated version: https://github.com/aave/aave-v4/pull/1024/commits/fe952bafd7a0296499d65c655e03a5776b8748f5

returns only 2 params of the premiumDelta struct and opts to rm the redundant calc of accrued premium

src/spoke/Spoke.sol
```diff
@@ -300,54 +302,67 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+        riskPremium: positionStatus.riskPremium
+      });
+      totalDebtRestored = drawnDebtRestored + premiumDelta.restoredPremiumRay.fromRayUp();
 
-    userPosition.applyPremiumDelta(premiumDelta);
-    userPosition.drawnShares -= restoredShares.toUint120();
-    if (userPosition.drawnShares == 0) {
-      _positionStatus[onBehalfOf].setBorrowing(reserveId, false);
+      reserve.underlying.safeTransferFrom(msg.sender, address(reserve.hub), totalDebtRestored);
+      reserve.hub.restore(reserve.assetId, drawnDebtRestored, premiumDelta);
```
Kogaroshi, line 340: Can we use the hub returned shares amount here to verify the preview amount used before was indeed correct ?

src/spoke/Spoke.sol
```diff
@@ -973,9 +993,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   function _getUserDebt(
     IHubBase hub,
     uint256 assetId,
-    UserPosition storage userPosition
+    UserPosition storage userPosition,
+    uint256 drawnIndex
```
Kogaroshi, line 997: might be worth to have 2 internal method so we don't need to fetch and pass `drawnIndex` when we already pass the `hub` + `assetId`, or remove `hub` & `assetId` params here as it seems we do not need them anymore.

src/spoke/Spoke.sol
```diff
@@ -300,54 +302,67 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-      onBehalfOf,
-      restoredShares,
-      drawnDebtRestored + premiumDebtRestored,
-      premiumDelta
-    );
+    return (restoredShares, totalDebtRestored);
+  }
 
-    return (restoredShares, drawnDebtRestored + premiumDebtRestored);
+  function _getPremiumDelta(
```
Kogaroshi, line 354: To move with other internal methods.

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -776,24 +776,45 @@ contract SpokeBase is Base {
...
+    ISpoke.UserPosition memory pos = spoke.getUserPosition(reserveId, user);
+    uint256 riskPremiumCalculated = pos.premiumShares.percentDivDown(pos.drawnShares); // back calculated from premShares/drawnShares
+    uint256 riskPremiumOnTheFly = spoke.getUserAccountData(user).riskPremium;
+    uint256 riskPremiumStored = _getUserRpStored(spoke, reserveId, user);
+    assertEq(riskPremiumStored, expectedRP, 'user risk premium mismatch');
+    assertNotEq(riskPremiumOnTheFly, riskPremiumStored, 'user risk premium expected mismatch');
+  }
+
+  // @dev reads `positionStatus.riskPremium` by temporarily upgrading to mock spoke
   function _getUserRpStored(
```
Kogaroshi, line 801: same as https://github.com/aave/aave-v4/pull/1028#discussion_r2564193467

yan-man:
see https://github.com/aave/aave-v4/pull/1030

PR #1028, opened 2025-11-25
opt: remove notify on repay v1

yan-man:
- opt to rm calculated accrued premium twice
- resolve tests


tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -776,24 +776,45 @@ contract SpokeBase is Base {
...
+  /// instead RP should remain same as prior value
+  function _assertUserRpUnchangedAfterRepay(
+    uint256 reserveId,
+    ISpoke spoke,
+    address user,
+    uint256 expectedRP
+  ) internal {
+    ISpoke.UserPosition memory pos = spoke.getUserPosition(reserveId, user);
+    uint256 riskPremiumCalculated = pos.premiumShares.percentDivDown(pos.drawnShares); // back calculated from premShares/drawnShares
+    uint256 riskPremiumOnTheFly = spoke.getUserAccountData(user).riskPremium;
```
Kogaroshi, line 794: maybe rename to `riskPremiumReceived` or something similar, the "OnTheFly" term seems a bit messy

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -776,24 +776,45 @@ contract SpokeBase is Base {
...
+    uint256 riskPremiumStored = _getUserRpStored(spoke, reserveId, user);
+    assertEq(riskPremiumStored, expectedRP, 'user risk premium mismatch');
+    assertNotEq(riskPremiumOnTheFly, riskPremiumStored, 'user risk premium expected mismatch');
+  }
+
+  // @dev reads `positionStatus.riskPremium` by temporarily upgrading to mock spoke
   function _getUserRpStored(
```
Kogaroshi, line 801: ~~why is this change needed ?~~ nvm saw the change in the other PR.
Would it be worth it to have a getter for the user's `riskPremium` new, as it might be fetched externally in other contexts ?

yan-man: yes! mentioned this to miguel too, think it makes sense to expose a getter for the in-storage RP

yan-man:
see https://github.com/aave/aave-v4/pull/1030

PR #1030, opened 2025-11-27
Remove `_notifyRiskPremiumUpdate` from repay action

yan-man:
closes https://github.com/aave/aave-v4/issues/1031


src/spoke/Spoke.sol
```diff
@@ -291,6 +293,15 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
   }
 
+  /// @notice Local variables for repay action.
+  struct RepayVars {
+    uint256 drawnDebtRestored;
+    uint256 premiumDebtRayDelta;
+    uint256 drawnIndex;
+    uint256 premiumDebtRay;
+    uint256 premiumDebtRayRestored;
+  }
```
yan-man: perhaps you will not like this @avniculae, but was not able to resolve stack after a few hrs of trying without doing this. 

Also - not sure if we should move this up to the top of the file as is typical conventions @miguelmtzinf

tests/unit/Spoke/SpokeBase.t.sol
```diff
@@ -725,26 +725,31 @@ contract SpokeBase is Base {
     assertEq(abi.encode(a), abi.encode(b)); // sanity check
   }
 
-  function _assertUserRpUnchanged(uint256 reserveId, ISpoke spoke, address user) internal view {
-    ISpoke.UserPosition memory pos = spoke.getUserPosition(reserveId, user);
-    uint256 riskPremiumStored = pos.premiumShares.percentDivDown(pos.drawnShares);
```
yan-man, line 730: dont think we should ever back calculate like this. It can be wrong when shares are low, ex. RP = 15_00, but drawnShares = 2. Then premiumShares = 1 due to rounding, and riskPremiumStored would be 50_00

src/spoke/Spoke.sol
```diff
@@ -301,37 +313,48 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+    uint256 totalDebtRestored = vars.drawnDebtRestored +
+      premiumDelta.restoredPremiumRay.fromRayUp();
     reserve.underlying.safeTransferFrom(msg.sender, address(reserve.hub), totalDebtRestored);
-    uint256 restoredShares = reserve.hub.restore(reserve.assetId, drawnDebtRestored, premiumDelta);
+    uint256 restoredShares = reserve.hub.restore(
+      reserve.assetId,
+      vars.drawnDebtRestored,
+      premiumDelta
+    );
+    require(restoredShares == vars.previewRestoredShares);
```
yan-man: added this due to suggestion from @Kogaroshi to validate the previewed shares prior to the action. Thoughts?

src/spoke/Spoke.sol
```diff
@@ -1015,4 +1048,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
       fnOut := fnIn
     }
   }
+
+  /// @notice Checks if the caller is an approved positionManager for `onBehalfOf`.
+  function _onlyPositionManager(address onBehalfOf) internal view {
+    require(_isPositionManager({user: onBehalfOf, manager: msg.sender}), Unauthorized());
+  }
```
yan-man: did this for spoke size, open to other suggestions

src/spoke/Spoke.sol
```diff
@@ -933,17 +965,6 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-  /// @return The user's drawn debt.
-  /// @return The user's premium debt.
-  /// @return The user's premium debt, expressed in asset units and scaled by RAY.
-  function _getUserDebt(
-    UserPosition storage userPosition,
-    IHubBase hub,
-    uint256 assetId
-  ) internal view returns (uint256, uint256, uint256) {
-    return _getUserDebt(userPosition, hub.getAssetDrawnIndex(assetId));
-  }
```
yan-man: rm this actually increased spoke size... but i dont see the need for this one

Kogaroshi: Compiler Optimizer might have an easier way to inline it when declared via this internal helper

src/spoke/libraries/UserPositionPremium.sol
```diff
@@ -39,7 +39,7 @@ library UserPositionPremium {
   /// @param riskPremium The new risk premium, expressed in BPS.
   /// @param restoredPremiumRay The amount of premium to be restored, expressed in asset units and scaled by RAY.
   /// @return The computed premium delta.
-  function getPremiumDelta(
```
Kogaroshi, line 44: why the name change here ?

src/spoke/Spoke.sol
Kogaroshi: reverted in a119bc6bb4357bac17f89f058e66d2dcce593690, to gain back some margin on Spoke size.

src/spoke/Spoke.sol
Kogaroshi: reverted in a119bc6bb4357bac17f89f058e66d2dcce593690, Spoke size under buffer again (only 10B margin).

src/spoke/Spoke.sol
Kogaroshi: Removed local vars in a119bc6bb4357bac17f89f058e66d2dcce593690.

src/spoke/Spoke.sol
miguelmtzinf: move up

src/spoke/Spoke.sol
```diff
@@ -299,42 +299,67 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+    {
+      uint256 drawnIndex = reserve.hub.getAssetDrawnIndex(reserve.assetId);
+      uint256 premiumDebtRayDelta;
+      (
+        drawnDebtRestored,
+        premiumDelta.restoredPremiumRay,
+        premiumDebtRayDelta
+      ) = _calculateRestoreAmount(userPosition, drawnIndex, drawnDebtRestored);
+
+      previewRestoredShares = reserve.hub.previewRestoreByAssets(
```
DhairyaSethi: you can just use the cached drawn index here instead of ext call again

src/spoke/Spoke.sol
```diff
@@ -299,42 +299,67 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    IHubBase.PremiumDelta memory premiumDelta = userPosition.getPremiumDelta({
-      drawnIndex: drawnIndex,
-      riskPremium: 0,
-      restoredPremiumRay: premiumDebtRayRestored
-    });
+    IHubBase.PremiumDelta memory premiumDelta;
+    uint256 drawnDebtRestored = amount;
+    uint256 newDrawnShares;
+    uint256 previewRestoredShares;
+    {
```
DhairyaSethi: would make this logic composable in an internal method & reuse in liq as well

src/spoke/Spoke.sol
Kogaroshi: Needed now, since we do not fetch using `previewRestoreByAssets` anymore, but instead calculate using the fetched index, and we now verify against the `restoredShares` returned by `Hub.restore()`

src/spoke/Spoke.sol
Kogaroshi: liquidate behaves a bit differently, only the part that is similar to `_liquidateDebt` could be reused

src/spoke/libraries/UserPositionPremium.sol
```diff
@@ -47,13 +47,40 @@ library UserPositionPremium {
...
+
+  /// @notice Calculates the premium delta for a user position given a new risk premium and new drawn shares.
+  /// @param oldPremiumShares The current value of premium shares.
+  /// @param oldPremiumOffsetRay The current value of premium offset, expressed in asset units and scaled by RAY.
+  /// @param newDrawnShares The new drawn shares, including the restored shares.
+  /// @param drawnIndex The current drawn index.
+  /// @param riskPremium The new risk premium, expressed in BPS.
+  /// @param restoredPremiumRay The amount of premium to be restored, expressed in asset units and scaled by RAY.
+  /// @return The computed premium delta.
+  function getPremiumDelta(
```
avniculae: we are doing all of this refactoring bc we don't want to update userPosition storage before `restore` bc we want to be consistent with the rest of the code? bc otherwise, we could just update userPosition.drawnShares and we could simply use same function from here

Kogaroshi: Yeah, I'd rather keep consistency of transfer => hub interaction => update shares in UserPosition.
But we can consider breaking it to simplify the logic here.

avniculae: yeah, we are doing `transfer -> hub -> spoke storage update` all the time just because hub returns us the info we need to update the storage. However, repay now is an exception, since we need to preview hub op to understand what to pass to the hub. Hence, I think it's fine to break this rule with enough dev comments. I think the code will be much cleaner

Kogaroshi: well, we do estimate the shares hub would return, but after we still verify against what the hub returned, imo it's only there that the value is correct and we can safely update storage.
In our current version of Hub & Spoke, i agree that we should always have the correct shares estimation, just trying to think of future Hubs that might present an issue there.

DhairyaSethi: context is only given away (ext call) on token transfer that happens before hub.restore, it's anyway better practice to update state before this but we explicitly do not support callback tokens. i agree with alex, its fine to make this exception given it simplifies the code

Kogaroshi: After discussions, broke consistency and tried to simplify the repay (an liquidate) flow, removing the unnecessary double `getPremiumDelta`.
Changed allow to simplify the repay function, and remove the tricks used to avoid StackTooDeep error (except 1, see the `require` wrapping `hub.restore()`).
This simplification also comes with reduction of bytecode size, and a negligeable gas cost increase (~200 gas) for repay & liquidate calls.

See 0cb2d9c646fe7f2b18ad8856d81dc1e5f6adc8a8

src/spoke/Spoke.sol
```diff
@@ -299,42 +299,50 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   ) external onlyPositionManager(onBehalfOf) returns (uint256, uint256) {
     Reserve storage reserve = _getReserve(reserveId);
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
+    PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
     _validateRepay(reserve);
 
+    IHubBase.PremiumDelta memory premiumDelta;
```
DhairyaSethi: not need anymore, can be declared when populating

src/spoke/Spoke.sol
```diff
@@ -299,42 +299,50 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+      riskPremium: positionStatus.riskPremium,
+      restoredPremiumRay: premiumDelta.restoredPremiumRay
     });
 
     uint256 totalDebtRestored = drawnDebtRestored + premiumDelta.restoredPremiumRay.fromRayUp();
     reserve.underlying.safeTransferFrom(msg.sender, address(reserve.hub), totalDebtRestored);
-    uint256 restoredShares = reserve.hub.restore(reserve.assetId, drawnDebtRestored, premiumDelta);
+    require(
+      previewRestoredShares == reserve.hub.restore(reserve.assetId, drawnDebtRestored, premiumDelta)
+    );
```
DhairyaSethi: assert makes more sense bc user should never be able to trigger this? 
also if we're doing this (checking config behaviour essentially) probably makes sense to add on report deficit that debt after is 0 as an assert? 

in general, smart contracts shouldn't need run time asserts but i understand the rationale here

src/spoke/Spoke.sol
```diff
@@ -299,42 +299,50 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
     userPosition.applyPremiumDelta(premiumDelta);
```
avniculae, line 328: let's apply premium delta before restore, since we update drawnShares before as well

src/spoke/Spoke.sol
```diff
@@ -299,42 +299,50 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
     Reserve storage reserve = _getReserve(reserveId);
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
+    PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
     _validateRepay(reserve);
 
+    IHubBase.PremiumDelta memory premiumDelta;
+    uint256 drawnDebtRestored = amount;
     uint256 drawnIndex = reserve.hub.getAssetDrawnIndex(reserve.assetId);
-    (uint256 drawnDebtRestored, , uint256 premiumDebtRayRestored) = _getUserDebt(
+    (drawnDebtRestored, premiumDelta.restoredPremiumRay) = _calculateRestoreAmount(
```
avniculae: maybe we can revert this change now?

Kogaroshi: It was kept to save 1 new var in this function, let me try if removing does not recreate a StackTooDeep issue.

src/spoke/interfaces/ISpoke.sol
```diff
@@ -475,6 +475,14 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     address user
   ) external view returns (UserPosition memory);
 
+  /// @notice Returns the most up-to-date user account data information.
+  /// @dev Utilizes user's current dynamic configuration of user position.
+  /// @dev Risk premium is calculated from up-to-date user position data, and will likely differ from the stored user risk premium.
+  function getUserAccountData(address user) external view returns (UserAccountData memory);
+
+  /// @notice Returns the stored user risk premium.
+  function getUserRiskPremium(address user) external view returns (uint256);
```
avniculae, line 484: based on how the other external functions behave in spoke & hub, this function name suggests that this is the up to date user risk premium (from this block). e.g., see `Hub.getAssetDrawnIndex`. Is there a better name for this maybe?

Kogaroshi: `getUserCurrentRiskPremium` ?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -333,33 +333,34 @@ library LiquidationLogic {
...
-      drawnDebtToLiquidate,
-      premiumDelta
+    require(
+      previewDrawnSharesLiquidated ==
+        debtReserve.hub.restore(debtReserve.assetId, drawnDebtToLiquidate, premiumDelta)
     );
-    debtPosition.applyPremiumDelta(premiumDelta);
-    debtPosition.drawnShares -= drawnSharesLiquidated.toUint120();
 
+    debtPosition.applyPremiumDelta(premiumDelta);
```
avniculae, line 356: let's apply premium delta immediately after we get the premium delta. this was we update user position (drawn + premium) in the same place

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -333,33 +333,34 @@ library LiquidationLogic {
     LiquidateDebtParams memory params
   ) internal returns (uint256, IHubBase.PremiumDelta memory, bool) {
     uint256 premiumDebtToLiquidateRay = params.debtToLiquidate.toRay().min(params.premiumDebtRay);
+    uint256 premiumDebtToLiquidate = premiumDebtToLiquidateRay.fromRayUp();
+    uint256 drawnDebtToLiquidate = params.debtToLiquidate - premiumDebtToLiquidate;
+
+    uint256 previewDrawnSharesLiquidated = drawnDebtToLiquidate.rayDivDown(params.drawnIndex);
```
avniculae: I'd call this `drawnSharesLiquidated` and not assert returned value from the hub anymore. We can leave a comment mentioning that here we preview and that the value will be the same. wdyt? you think is scary?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -333,33 +333,34 @@ library LiquidationLogic {
...
       address(debtReserve.hub),
       drawnDebtToLiquidate + premiumDebtToLiquidate
     );
-    uint256 drawnSharesLiquidated = debtReserve.hub.restore(
-      debtReserve.assetId,
-      drawnDebtToLiquidate,
-      premiumDelta
+    require(
+      previewDrawnSharesLiquidated ==
+        debtReserve.hub.restore(debtReserve.assetId, drawnDebtToLiquidate, premiumDelta)
```
avniculae: even if we keep require/assert here, I'd like us to call restore, store returned value in a var and then perform require/assert. I don't like us making the restore call in the require statemenent

src/spoke/libraries/UserPositionDebt.sol
```diff
@@ -13,7 +13,7 @@ import {Premium} from 'src/hub/libraries/Premium.sol';
 /// @author Aave Labs
 /// @notice Implements premium calculations for user positions.
 library UserPositionPremium {
-  using PercentageMath for uint120;
+  using PercentageMath for *;
```
avniculae, line 17: why?

src/spoke/Spoke.sol
```diff
@@ -299,42 +299,50 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   ) external onlyPositionManager(onBehalfOf) returns (uint256, uint256) {
     Reserve storage reserve = _getReserve(reserveId);
     UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
+    PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
     _validateRepay(reserve);
 
+    IHubBase.PremiumDelta memory premiumDelta;
+    uint256 drawnDebtRestored = amount;
```
avniculae: why?

src/spoke/Spoke.sol
```diff
@@ -299,42 +299,50 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
     );
 
-    (drawnDebtRestored, premiumDebtRayRestored) = _calculateRestoreAmount(
-      drawnDebtRestored,
-      premiumDebtRayRestored,
-      amount
-    );
+    // replicate hub.previewRestoreByAssets() instead of calling here
+    // verified later against actual restored shares returned from hub.restore()
+    uint256 previewRestoredShares = drawnDebtRestored.rayDivDown(drawnIndex);
```
avniculae: similar to liquidation logic, I'd like to challenge whether we should just call this `restoredShares` and not assert returned value from hub (will leave dev comments instead)

src/spoke/libraries/UserPositionDebt.sol
Kogaroshi: https://github.com/aave/aave-v4/blob/6ac680b0b376777d023353b79a45671dfe4679d8/src/spoke/libraries/UserPositionPremium.sol#L57

src/spoke/Spoke.sol
Kogaroshi: passing `amount` later in the method was creating StackTooDeep error.
I'll try to remove this later on when done with all logic changes.

src/spoke/Spoke.sol
Kogaroshi: resolved with 23b7151b725fc31ab1767fdbfd41e1bd649c2f64

src/spoke/libraries/LiquidationLogic.sol
Kogaroshi: resolved with 23b7151b725fc31ab1767fdbfd41e1bd649c2f64

src/spoke/libraries/LiquidationLogic.sol
Kogaroshi: resolved with 23b7151b725fc31ab1767fdbfd41e1bd649c2f64

src/spoke/libraries/LiquidationLogic.sol
Kogaroshi: this change was reverted, since `drawnShares` update was reverted back to previous place.

src/spoke/Spoke.sol
Kogaroshi: this change was reverted, since `drawnShares` update was reverted back to previous place

src/spoke/Spoke.sol
Kogaroshi: resolved with https://github.com/aave/aave-v4/commit/23b7151b725fc31ab1767fdbfd41e1bd649c2f64

src/spoke/libraries/UserPositionPremium.sol
Kogaroshi: This was changed again in resolved with https://github.com/aave/aave-v4/commit/23b7151b725fc31ab1767fdbfd41e1bd649c2f64

Now we trust the previewed shares amount in repay, and removed the require/assert.
In liquidate, there is no need for a require anymore since we use `hub.restore()` returned shares amount again.

src/spoke/Spoke.sol
Kogaroshi: This check is not done anymore, we use `previewRestoreByAssets` again, and trust that `hub.restore()` will return the same value.

src/spoke/Spoke.sol
Kogaroshi: fixed in a51137a46df51786a6ef8dd3d4f83efb52780b2c

src/spoke/Spoke.sol
Kogaroshi: reverted in a51137a46df51786a6ef8dd3d4f83efb52780b2c

src/spoke/Spoke.sol
Kogaroshi: fixed in a51137a46df51786a6ef8dd3d4f83efb52780b2c

src/spoke/Spoke.sol
Kogaroshi: outdated, now repay and liquidate do not share a lot of common logic.

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -333,7 +333,10 @@ library LiquidationLogic {
     LiquidateDebtParams memory params
   ) internal returns (uint256, IHubBase.PremiumDelta memory, bool) {
     uint256 premiumDebtToLiquidateRay = params.debtToLiquidate.toRay().min(params.premiumDebtRay);
+
+    // reset premium data for later adjustment in notify
     IHubBase.PremiumDelta memory premiumDelta = debtPosition.getPremiumDelta({
+      drawnSharesTaken: debtPosition.drawnShares,
```
avniculae, line 339: while this has no effect (bc rp is 0), this is incorrect. we don't change drawnShares here, `drawnSharesTaken` should be 0

src/spoke/libraries/UserPositionDebt.sol
avniculae: do we use it somewhere else as well? I think we can do `for uint256` no?

src/spoke/Spoke.sol
```diff
@@ -941,6 +957,21 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     require(config.liquidationFee <= PercentageMath.PERCENTAGE_FACTOR, InvalidLiquidationFee());
   }
 
+  /// @dev Calculates the drawn debt and premium debt to restore for the given user position and amount.
+  /// @param userPosition The user position.
+  /// @param drawnIndex The drawn index of the reserve.
+  /// @param amount The amount to restore.
+  /// @return The amount of drawn debt to restore, expressed in asset units.
+  /// @return The amount of premium debt to restore, expressed in asset units and scaled by RAY.
+  function _calculateRestoreAmount(
```
avniculae, line 966: should we move this in UserPositionDebt? I think it's a good candidate

src/spoke/libraries/UserPositionDebt.sol
Kogaroshi: you're right, let me check and make the change

src/spoke/Spoke.sol
```diff
@@ -302,36 +302,46 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-      drawnDebtRestored,
-      premiumDebtRayRestored,
+    (uint256 drawnDebtRestored, uint256 premiumDebtRayRestored) = _calculateRestoreAmount(
+      userPosition,
+      drawnIndex,
       amount
     );
 
+    // we preview the drawn shares to be restored, should always be the same value returned by hub.restore()
+    uint256 previewRestoredShares = reserve.hub.previewRestoreByAssets(
```
avniculae, line 312: I know it might be a stretch, but what do you think of previewing it by calculating here manually (divUp), and call it `restoredShares`. Hub is immutable, and we can hardcode/duplicate logic here (we already do it  in other places)

src/spoke/interfaces/ISpoke.sol
avniculae: current can suggest up to date no? wdyt about `getUserLastRiskPremium`? or maybe we can keep it as is,  I don't have a strong opinion

src/spoke/Spoke.sol
Kogaroshi: current Hub is immutable, but future ones might have another logic applied during the preview, I'd prefer to trust Hub return values rather than calculating locally.

PR #1019, opened 2025-11-24
opt: Premium accounting with signed offset

avniculae:
- It removes one storage slot from `Hub.Asset`, `Hub.SpokeData` and `Spoke.UserPosition` by merging `realizedPremium` into `premiumOffset`. This requires a signed `premiumOffset` value.
- This is an optimization on gas consumption to reduce gas costs introduced from new premium accounting in RAY precision (https://github.com/aave/aave-v4/pull/991)

PR #1033, opened 2025-11-27
fix: Removes multiple rounding on totalAddedAssets

### LINKED ISSUE #1032: AssetLogic.totalAddedAssets rounds up three times (drawn, premium and deficit)
The calculation of `totalAddedAssets` rounds up 3 times: drawn, premium and deficit. This can lead to a decrease of supply share price when reporting deficit, since it moves assets from drawn and premium to deficit.

https://github.com/aave/aave-v4/blob/06ee85037e851e4ddea6d00b70d26aaa62c6da53/src/hub/libraries/AssetLogic.sol#L81-L90

miguelmtzinf:
- remove multiple layers of rounding on totalAddedAssets, which could lead to a decrease of supply share price when reporting deficit, since it moves assets from drawn and premium to deficit.

Closes #1032

### DIFF
--- a/src/hub/libraries/AssetLogic.sol
+++ b/src/hub/libraries/AssetLogic.sol
@@ -87,9 +87,7 @@ library AssetLogic {
       premiumOffsetRay: asset.premiumOffsetRay,
       realizedPremiumRay: asset.realizedPremiumRay
     });
-    uint256 aggregatedOwedRay = (uint256(asset.drawnShares) * drawnIndex) +
-      premiumRay +
-      asset.deficitRay;
+    uint256 aggregatedOwedRay = (asset.drawnShares * drawnIndex) + premiumRay + asset.deficitRay;
 
     return
       asset.liquidity +
@@ -199,18 +197,18 @@ library AssetLogic {
     uint256 realizedPremiumRay = asset.realizedPremiumRay;
     uint256 deficitRay = asset.deficitRay;
 
-    uint256 aggregatedOwedRayBefore = (uint256(drawnShares) * previousIndex) +
+    uint256 aggregatedOwedRayAfter = (drawnShares * drawnIndex) +
       Premium.calculatePremiumRay({
         premiumShares: premiumShares,
-        drawnIndex: previousIndex,
+        drawnIndex: drawnIndex,
         premiumOffsetRay: premiumOffsetRay,
         realizedPremiumRay: realizedPremiumRay
       }) +
       deficitRay;
-    uint256 aggregatedOwedRayAfter = (uint256(drawnShares) * drawnIndex) +
+    uint256 aggregatedOwedRayBefore = (drawnShares * previousIndex) +
       Premium.calculatePremiumRay({
         premiumShares: premiumShares,
-        drawnIndex: drawnIndex,
+        drawnIndex: previousIndex,
         premiumOffsetRay: premiumOffsetRay,
         realizedPremiumRay: realizedPremiumRay
       }) +
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -2805,14 +2805,13 @@ abstract contract Base is Test {
       MathUtils.calculateLinearInterest(asset.drawnRate, uint40(asset.lastUpdateTimestamp))
     );
 
-    uint256 liquidityGrowthBefore = (uint256(asset.drawnShares) + asset.premiumShares) *
-      previousIndex -
+    uint256 liquidityGrowthAfter = (uint256(asset.drawnShares) + asset.premiumShares) *
+      drawnIndex -
       asset.premiumOffsetRay +
       asset.realizedPremiumRay +
       asset.deficitRay;
-
-    uint256 liquidityGrowthAfter = (uint256(asset.drawnShares) + asset.premiumShares) *
-      drawnIndex -
+    uint256 liquidityGrowthBefore = (uint256(asset.drawnShares) + asset.premiumShares) *
+      previousIndex -
       asset.premiumOffsetRay +
       asset.realizedPremiumRay +
       asset.deficitRay;


src/hub/libraries/AssetLogic.sol
```diff
@@ -80,11 +80,14 @@ library AssetLogic {
   /// @notice Returns the total added assets for the specified asset.
   function totalAddedAssets(IHub.Asset storage asset) internal view returns (uint256) {
     uint256 drawnIndex = asset.getDrawnIndex();
+    uint256 liquidityGrowth = (uint256(asset.drawnShares) + asset.premiumShares) *
```
avniculae: This is not growth in liquidity, it's drawn + premium debt. Also, I think `(drawnShares + premiumShares) * index` is a bit confusing. I personally think a better approach is
```
uint256 premiumDebtRay = Premium.calculatePremiumRay({
      premiumShares: asset.premiumShares,
      drawnIndex: drawnIndex,
      premiumOffsetRay: asset.premiumOffsetRay,
      realizedPremiumRay: asset.realizedPremiumRay
    });

uint256 debtRay = asset.drawnShares * drawnIndex + premiumDebtRay + deficitRay;

return asset.liquidity + asset.swept + debtRay.fromRayUp() + asset.realizedFees - asset.getUnrealizedFees(drawnIndex);
```

wdyt? this way we don't inline the premium logic, which is always abstracted away by the Premium library in the current codebase

src/hub/libraries/AssetLogic.sol
```diff
@@ -183,26 +186,23 @@ library AssetLogic {
...
-    uint256 premiumRayBefore = Premium.calculatePremiumRay({
-      premiumShares: premiumShares,
-      drawnIndex: previousIndex,
-      premiumOffsetRay: premiumOffsetRay,
-      realizedPremiumRay: realizedPremiumRay
-    });
-    uint256 liquidityGrowthPremium = premiumRayAfter.fromRayUp() - premiumRayBefore.fromRayUp();
 
-    return (liquidityGrowthDrawn + liquidityGrowthPremium).percentMulDown(liquidityFee);
+    uint256 liquidityGrowthBefore = (uint256(drawnShares) + premiumShares) *
```
avniculae: I wouldn't name this `liquidityGrowth` --> this is just totalAddedAssets at a point in time. liquidityGrowth should be used when doing a difference of totalAddedAssets at 2 points in time.

Also, we should include the deficitRay as well in this calculation, as it does not trivially cancels out.

src/hub/libraries/AssetLogic.sol
```diff
@@ -80,11 +80,14 @@ library AssetLogic {
...
+    uint256 liquidityGrowth = (uint256(asset.drawnShares) + asset.premiumShares) *
+      drawnIndex -
+      asset.premiumOffsetRay +
+      asset.realizedPremiumRay;
     return
       asset.liquidity +
       asset.swept +
-      asset.deficitRay.fromRayUp() +
-      asset.totalOwed(drawnIndex) -
+      (liquidityGrowth + asset.deficitRay).fromRayUp() -
```
avniculae: should we round down here as discussed?

avniculae: if we do it separately, let's fix supply_share_price_deficit.py

tests/misc/supply_share_price_deficit.py
```diff
@@ -0,0 +1,58 @@
...
+def rayMulUp(a, b):
+    return (a * b + RAY - 1) / RAY
+
+def rayMulDown(a, b):
+    return (a * b) / RAY
+
+def totalAddedAssets(drawnShares, premiumDebtRay, deficitRay, drawnIndex):
+  # return rayMulUp(drawnShares, drawnIndex) + fromRayUp(premiumDebtRay) + fromRayUp(deficitRay)          # this is wrong
+  # return rayMulDown(drawnShares, drawnIndex) + fromRayDown(premiumDebtRay) + fromRayDown(deficitRay)    # this is wrong
+  return fromRayDown(drawnShares * drawnIndex + premiumDebtRay + deficitRay)
```
avniculae: fix rounding here

src/hub/libraries/AssetLogic.sol
```diff
@@ -80,11 +80,21 @@ library AssetLogic {
...
   function totalAddedAssets(IHub.Asset storage asset) internal view returns (uint256) {
     uint256 drawnIndex = asset.getDrawnIndex();
+
+    uint256 premiumRay = Premium.calculatePremiumRay({
+      premiumShares: asset.premiumShares,
+      drawnIndex: drawnIndex,
+      premiumOffsetRay: asset.premiumOffsetRay,
+      realizedPremiumRay: asset.realizedPremiumRay
+    });
+    uint256 aggregatedOwedRay = (uint256(asset.drawnShares) * drawnIndex) +
```
avniculae: do we need explicit uint256 casting here? drawnIndex is uint256 already

src/hub/libraries/AssetLogic.sol
```diff
@@ -183,26 +193,32 @@ library AssetLogic {
...
-      premiumShares: premiumShares,
-      drawnIndex: previousIndex,
-      premiumOffsetRay: premiumOffsetRay,
-      realizedPremiumRay: realizedPremiumRay
-    });
-    uint256 liquidityGrowthPremium = premiumRayAfter.fromRayUp() - premiumRayBefore.fromRayUp();
+    uint256 realizedPremiumRay = asset.realizedPremiumRay;
+    uint256 deficitRay = asset.deficitRay;
+
+    uint256 aggregatedOwedRayBefore = (uint256(drawnShares) * previousIndex) +
```
avniculae: should we do After, and then Before, similar to the current implementations?

src/hub/libraries/AssetLogic.sol
avniculae: is it worth creating a function to get aggregatedOwedRay? we can use it here and twice in getUnrealizedFees, wdyt?

src/hub/libraries/AssetLogic.sol
Kogaroshi: not really important here I think

tests/Base.t.sol
```diff
@@ -2793,18 +2800,26 @@ abstract contract Base is Test {
...
     );
-    uint256 liquidityGrowth = asset.drawnShares.rayMulUp(drawnIndex) -
-      asset.drawnShares.rayMulUp(lastDrawnIndex) +
-      (asset.premiumShares * drawnIndex - asset.premiumOffsetRay + asset.realizedPremiumRay)
-        .fromRayUp() -
-      (asset.premiumShares * lastDrawnIndex - asset.premiumOffsetRay + asset.realizedPremiumRay)
-        .fromRayUp();
 
-    return liquidityGrowth.percentMulDown(asset.liquidityFee);
+    uint256 liquidityGrowthAfter = (uint256(asset.drawnShares) + asset.premiumShares) *
```
avniculae, line 2808: this should be named aggregatedOwedRayAfter. I can fix in #1019

PR #1034, opened 2025-11-27
refactor: ReserveFlags library

avniculae:


avniculae:
We decided to not go forward with this as it adds more gas

PR #1036, opened 2025-11-27
feat: add grace period for liquidations

### LINKED ISSUE #1041: V4 Spoke cannot prevent liquidations of collateral under certain circumstances
Aave V3 includes two mechanisms that help prevent liquidations: the L2 sequencer halting protection and the unpause grace period.
- L2 sequencer halting. The protocol can provide a grace period for users to adjust their positions when the L2 sequencer halts. During this period, users can top up collateral or repay debt, and liquidations are blocked so their positions remain protected. This should not be used to worsen positions, which is why borrowing is disabled during the grace period.
- Unpause with grace period: The protocol can also provide a grace period after unpausing a reserve. Users have time to fix their positions while liquidations are temporarily disabled. This action is typically paired with setting `LTV = 0`, preventing new borrowing against the affected collateral (and forcing users to withdraw that asset first, if applicable).

V4 can have a more flexible mechanism by introducing a liquidatable flag, so V3 use cases can be replicated and extended by building automation on top.
- For L2 sequencer halting, instead of relying on an L2 uptime feed, an automated agent can detect the halting event and push an update from L1 to L2 to disable liquidations for that asset for a defined period.
- For unpausing, the action can be paired with disabling liquidations and introducing a new dynamic config with `CF = 0`, ensuring that users are protected from liquidation for a period of time while still preventing them from making their positions riskier (via borrow or withdraw).

miguelmtzinf:
Closes #1041


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -386,12 +390,17 @@ library LiquidationLogic {
...
-  ) internal pure {
+  ) internal view {
     require(params.user != params.liquidator, ISpoke.SelfLiquidation());
     require(params.debtToCover > 0, ISpoke.InvalidDebtToCover());
     require(!params.collateralReservePaused && !params.debtReservePaused, ISpoke.ReservePaused());
     require(params.collateralReserveBalance > 0, ISpoke.ReserveNotSupplied());
     require(params.debtReserveBalance > 0, ISpoke.ReserveNotBorrowed());
+    require(
+      params.collateralReserveLiquidationGracePeriodUntil <= block.timestamp &&
+        params.debtReserveLiquidationGracePeriodUntil <= block.timestamp,
```
avniculae, line 401: I see that V3 does this as well, but why do we need to check this for debt reserve as well? debt reserve is getting repaid (which is fine), the problem is the coll reserve no?

src/spoke/Spoke.sol
```diff
@@ -5,7 +5,9 @@ pragma solidity 0.8.28;
 import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
 import {IERC20Permit} from 'src/dependencies/openzeppelin/IERC20Permit.sol';
 import {SignatureChecker} from 'src/dependencies/openzeppelin/SignatureChecker.sol';
-import {AccessManagedUpgradeable} from 'src/dependencies/openzeppelin-upgradeable/AccessManagedUpgradeable.sol';
+import {
+  AccessManagedUpgradeable
+} from 'src/dependencies/openzeppelin-upgradeable/AccessManagedUpgradeable.sol';
```
avniculae, line 10: 😭😭 this again

tests/unit/Spoke/Spoke.MultipleHub.t.sol
```diff
@@ -29,6 +29,7 @@ contract SpokeMultipleHubTest is SpokeBase {
     vm.startPrank(ADMIN);
     // Relist hub 2's dai on spoke1
     ISpoke.ReserveConfig memory daiHub2Config = ISpoke.ReserveConfig({
+      liquidationGracePeriodUntil: 0,
```
avniculae, line 32: this ReserveConfig appears in a lot of places. should we have a function `getDefaultReserveConfig(collateralRisk)` in Base?

tests/unit/libraries/LiquidationLogic/LiquidationLogic.ValidateLiquidationCall.t.sol
avniculae, line 1: we need to add tests here to make sure validate reverts when grace period is active

Kogaroshi: let's maybe introduce an internal fct like `triggerValidateLiquidationCall(params, bool shouldRevert, bytes error)` to be reused in all those tests.

miguelmtzinf:
Closing in favor of #1042

PR #1042, opened 2025-11-29
feat: Add liquidatable flag to Spoke

### LINKED ISSUE #1041: V4 Spoke cannot prevent liquidations of collateral under certain circumstances

miguelmtzinf:
Introduces a new `liquidatable` flag in a reserve, which determines whether that reserve can be seized during liquidation when used as collateral. This flag can be adjusted by priviledged roles depending on the scenario and can be used to replicate use cases such as the V3 oracle sentinel and the V3 unpause grace period.

There are two assets involved in a liquidation: the collateral reserve that is seized and the debt reserve that is repaid. Liquidation is only blocked when the collateral reserve is not liquidatable. The system does not consider the `liquidatable` flag of the debt reserve, since any repayment is considered beneficial for the protocol.

In situations where an asset has been unpaused and users are temporarily protected from liquidation, the goal is to prevent illegitimate collateral seizures, not repayments. For example, if ETH is unpaused with a grace period, it makes sense to block liquidations of ETH/DAI positions, but not DAI/ETH positions (repaying ETH is acceptable).

Closes #1041


tests/unit/libraries/LiquidationLogic/LiquidationLogic.ValidateLiquidationCall.t.sol
```diff
@@ -204,6 +204,30 @@ contract LiquidationLogicValidateLiquidationCallTest is LiquidationLogicBaseTest
...
+    vm.expectRevert(ISpoke.CollateralCannotBeLiquidated.selector);
+    liquidationLogicWrapper.validateLiquidationCall(params);
+
+    // collateral.liquidatable = false; debt.liquidatable = true; => revert
+    params.collateralReserveFlags = params.collateralReserveFlags.setLiquidatable(false);
+    params.debtReserveFlags = params.debtReserveFlags.setLiquidatable(true);
+    vm.expectRevert(ISpoke.CollateralCannotBeLiquidated.selector);
+    liquidationLogicWrapper.validateLiquidationCall(params);
+
+    // collateral.liquidatable = true; debt.liquidatable = true; => revert
```
Kogaroshi: comment should not say "revert", same for next one

Kogaroshi (approved):

PR #1050, opened 2025-12-03
fix: allow `reportDeficit` even if spoke is paused

yan-man:
Currently, if a spoke is paused for a peripheral asset involved in liquidations (ie not directly repaid or seized in the liquidation, but merely another debt asset on the liquidated user position), it does not block the liquidation from going through in normal liquidation flows, because `hub.refreshPremium` can still be called by `paused` spokes.

However, liquidations that involve deficit reporting will still be blocked by paused spokes, because in `hub.reportDeficit` the `paused` flag is still validated. Thus, the behavior of paused peripheral assets is inconsistent in liquidations.

Instead, remove validation of paused flag in `reportDeficit` so that there is consistent behavior (ie paused peripheral assets still allow liquidation)

closes https://github.com/aave/aave-v4/issues/1096

### DIFF
--- a/tests/unit/Hub/Hub.ReportDeficit.t.sol
+++ b/tests/unit/Hub/Hub.ReportDeficit.t.sol
@@ -33,47 +33,6 @@ contract HubReportDeficitTest is HubBase {
     _addLiquidity(usdxAssetId, MAX_SUPPLY_AMOUNT);
   }
 
-  function test_reportDeficit_active_paused_scenarios() public {
-    // draw usdx liquidity to be restored
-    _drawLiquidity({
-      assetId: usdxAssetId,
-      amount: 1,
-      withPremium: true,
-      skipTime: true,
-      spoke: address(spoke1)
-    });
-
-    // set spoke to active and paused
-    _updateSpokePaused(hub1, usdxAssetId, address(spoke1), true);
-    _updateSpokeActive(hub1, usdxAssetId, address(spoke1), true);
-
-    vm.prank(address(spoke1));
-    hub1.reportDeficit(usdxAssetId, 1, ZERO_PREMIUM_DELTA);
-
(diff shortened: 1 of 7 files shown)


miguelmtzinf:
https://github.com/aave/aave-v4/blob/6959e3219b5506bf2acae18551cbb2a68a5b8fba/src/hub/interfaces/IHub.sol#L76

This relates to the meaning of the `paused` state on the Hub. So far, it has been used to control any action that has an immediate impact on the liquidity of an asset.

This change slightly bends that definition, although it still makes sense if we think in terms of real liquidity flows, since no assets are actually transferred. We must revisit this and either make sure the definition is applied consistently across the code, or tweak it if needed.

tests/unit/Hub/Hub.ReportDeficit.t.sol
```diff
@@ -49,13 +49,37 @@ contract HubReportDeficitTest is HubBase {
     hub1.reportDeficit(usdxAssetId, 0, ZERO_PREMIUM_DELTA);
   }
 
+  /// @dev paused spoke can still report deficit
```
miguelmtzinf: better to move it down, alongside the rest of positive cases. Also, do we have tests for all other combinations? (active and paused, active and !paused, !active and paused, !active and !paused)

yan-man: moved down and added the combo tests. I think also worthwhile adding these for all the actions on spoke/hub in a dedicated test file, wdyt. I will add that to this PR as well

yan-man: https://github.com/aave/aave-v4/pull/1050/commits/7aaf208dcbbf584976f506701f5268a323abe3b8

PR #1052, opened 2025-12-08
test: Fix interest accrual scenario test

CheyenneAtapour:
Fixes the accrue interest scenario test, which was previously skipped


tests/unit/Spoke/Spoke.AccrueInterest.Scenario.t.sol
```diff
@@ -448,6 +453,7 @@ contract SpokeAccrueInterestScenarioTest is SpokeBase {
           bob
         );
       }
+
       // Workaround for precision loss with RP calc: https://github.com/aave/aave-v4/issues/421
```
DhairyaSethi: this workaround should be removed now, no?

CheyenneAtapour: Good catch, updated

snapshots/Hub.Operations.json
yan-man, line 1: hm weird why do gas snapshots change

tests/unit/Spoke/Spoke.AccrueInterest.Scenario.t.sol
```diff
@@ -56,15 +58,18 @@ contract SpokeAccrueInterestScenarioTest is SpokeBase {
...
     uint40 skipTime
   ) public {
-    vm.skip(true, 'pending rft');
     amounts = _bound(amounts);
     skipTime = bound(skipTime, 0, MAX_SKIP_TIME / 2).toUint40();
 
     // Ensure bob does not draw more than half his normalized supply value
     amounts = _ensureSufficientCollateral(spoke2, amounts);
-    TestAmounts memory originalAmounts = _copyAmounts(amounts); // deep copy original amounts
+    TestAmounts memory originalValues = _copyAmounts(amounts); // deep copy original amounts
```
yan-man: maybe can change the `TestAmounts` and `_copyAmounts` if it's `value` going fwd

CheyenneAtapour: I changed this name back to `originalAmounts` because this struct is indeed mostly "amounts"

tests/unit/Spoke/Spoke.AccrueInterest.Scenario.t.sol
```diff
@@ -56,699 +58,324 @@ contract SpokeAccrueInterestScenarioTest is SpokeBase {
...
-        MAX_SUPPLY_AMOUNT - amounts.wbtcSupplyAmount
-      );
-    }
 
-    // Bob borrows dai from spoke 2
-    if (amounts.daiBorrowAmount > 0) {
-      Utils.borrow(spoke2, _daiReserveId(spoke2), bob, amounts.daiBorrowAmount, bob);
+    // 1 -> DAI, 2 -> WETH, 3 -> USDx, 4 -> WBTC
+    TestAmount[] memory testAmounts = new TestAmount[](4);
+    for (uint256 i = 0; i < 4; ++i) {
```
yan-man: imo dont think there is a need for this loop, seems cumbersome. Could just set this whole array directly

tests/unit/Spoke/Spoke.AccrueInterest.Scenario.t.sol
```diff
@@ -56,699 +58,324 @@ contract SpokeAccrueInterestScenarioTest is SpokeBase {
     TestAmounts memory amounts,
```
yan-man: can you add more comments above to describe what's going on in this test? hard to follow and prefer we refactor this whole test later. nice job fixing this though it's a beast

CheyenneAtapour: Added

tests/unit/Spoke/Spoke.AccrueInterest.Scenario.t.sol
```diff
@@ -19,27 +19,29 @@ contract SpokeAccrueInterestScenarioTest is SpokeBase {
...
-  struct Rates {
-    uint96 daiBaseBorrowRate;
-    uint96 wethBaseBorrowRate;
-    uint96 usdxBaseBorrowRate;
-    uint96 wbtcBaseBorrowRate;
-  }
-
-  struct Indices {
-    uint256 daiIndex;
+    uint40 startTime;
```
Kogaroshi: is this one used ?

CheyenneAtapour: Nice catch! Removed

PR #1047, opened 2025-12-03
feat: deployment engine

yan-man:
- v4 deployment engine
- takes user input file, validates inputs, deploys smart contracts, sets up and grant roles, logs data into output json
- adapted from initial PR: https://github.com/aave/aave-v4/pull/986

PR #1053, opened 2025-12-08
Deployment Engine V2

avniculae:
Same as #1047 , but against main

### DIFF
--- a/scripts/deploy/inputs/AaveV4DeployInput.json
+++ b/scripts/deploy/inputs/AaveV4DeployInput.json
@@ -5,4 +5,3 @@
   "hubLabels": ["Hub 1", "Hub 2", "Hub 3"],
   "spokeLabels": ["Spoke 1", "Spoke 2", "Spoke 3"]
 }
-


scripts/deploy/inputs/AaveV4DeployInput.json
```diff
@@ -0,0 +1,7 @@
+{
+  "admin": "0x0000000000000000000000000000000000000001",
```
avniculae, line 2: I think more granularity would be good here, wdyt? I'd like us to have the possibility of defining hubs admins (one per each hub), spokes admins (one per each spoke), configurator admins (one per each configurator, one configurator per each hub/spoke), etc. wdyt?

scripts/deploy/AaveV4DeployBatch.s.sol
```diff
@@ -0,0 +1,46 @@
...
+import {OrchestrationReports} from 'src/deployments/libraries/OrchestrationReports.sol';
+import {InputUtils} from 'src/deployments/utils/InputUtils.sol';
+import {DeployUtils} from 'src/deployments/utils/DeployUtils.sol';
+import {MetadataLogger} from 'src/deployments/utils/MetadataLogger.sol';
+import {
+  AaveV4DeployOrchestration
+} from 'src/deployments/orchestration/AaveV4DeployOrchestration.sol';
+
+contract AaveV4DeployBatchScript is Script, DeployUtils, InputUtils {
+  string internal constant INPUT_PATH = 'scripts/deploy/inputs/AaveV4DeployInput.json';
```
avniculae, line 16: I think we should be able to have different inputs for different deployments as the same time, wdyt?

scripts/deploy/AaveV4DeployBatch.s.sol
```diff
@@ -0,0 +1,46 @@
...
+  string internal constant OUTPUT_DIR = 'output/reports/deployments/';
+  string internal constant OUTPUT_FILE = 'AaveV4DeployBatch.json';
+
+  constructor() {}
+
+  function run() external {
+    vm.createDir(OUTPUT_DIR, true);
+    MetadataLogger logger = new MetadataLogger(OUTPUT_DIR);
+    FullDeployInputs memory inputs = loadFullDeployInputs(INPUT_PATH);
```
avniculae, line 25: why do we need "Full"?

scripts/deploy/AaveV4DeployBatch.s.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    logger.log('...Starting Aave V4 Batch Deployment...');
+    address deployer = msg.sender;
+    vm.startBroadcast(deployer);
+    OrchestrationReports.FullDeploymentReport memory report = AaveV4DeployOrchestration
+      .deployAaveV4(
```
avniculae, line 31: I'd use named args here

scripts/deploy/inputs/AaveV4DeployInput.json
```diff
@@ -0,0 +1,7 @@
...
+  "nativeWrapperAddress": "0x0000000000000000000000000000000000000002",
+  "setRoles": true,
```
avniculae, line 4: what use case is there for setRoles = false?

scripts/deploy/AaveV4DeployBatch.s.sol
yan-man: i think bc there was an intention to allow for partial deployment inputs, ie just deploying a single hub or spoke and not everything together

scripts/deploy/inputs/AaveV4DeployInput.json
yan-man: i think perhaps deploying an instance via aave labs and not setting roles till an AIP to hand off to DAO, cc @Kogaroshi

scripts/deploy/AaveV4DeployBatch.s.sol
yan-man: can you clarify this, give an ex on the scenario?

yan-man:
refer to https://github.com/aave/aave-v4/pull/1047

scripts/deploy/inputs/AaveV4DeployInput.json
yan-man: i think for simplicity we would just remediate it on a per-deployment basis, ie if a new configurator is needed for another spoke or hub, then do a separate deployment, wdyt

PR #1055, opened 2025-12-10
feat: Introduce minimum liquidation bonus limit in liquidations

avniculae:
- Allows liquidators to pass a minimum liquidation bonus limit, such that liquidation reverts if liquidation bonus applied to liquidated user is less than the limit. This ensure liquidators are not frontrun by other liquidations that reduces liquidation bonus significantly.


avniculae:
Liquidators can do this themselves (and they probably do it already), hence closing this PR.

PR #1058, opened 2025-12-11
Fix/set self as user position manager

yan-man:
- don't allow user to call the setSelfAsUserPositionManagerWithSig function with a positonManager set to a diﬀerent address than the SignatureGateways is calling


yan-man:
refer to https://github.com/aave/aave-v4/pull/1064

PR #1054, opened 2025-12-08
feat : new position managers

Kogaroshi:
New Position Managers : 
- **GiverPositionManager** : allows any user to supply and/or repay on behalf of any other user that activated this PositionManager, with no other permissions or approval required. Using `MAX_UINT256` to repay the full debt is not allowed. 
- **TakerPositionManager** : allows an user to give withdrawal allowance and borrow allowance on their position (all allowance are given for a specific `Spoke` and `ReserveId` pair for each spender), allowing the spender to withdraw from the user position or borrow on the position based on the given allowances. This PositionManager is also inheriting ERC712, allowing to EIP712-typed intent to grant the allowance to spenders. When `MAX_UINT256` allowance is granted, it is treated as infinite allowance and not decreased when used by spender.
- **ConfigPositionManager** : allows a delegator to grand permissions to delegatee to call `setUsingAsCollateral`, `updateUserRiskPremiumO` & `updateUserDynamicConfig` on their behalf

PR #1059, opened 2025-12-12
feat: Temporary approvals in AllowancePositionManager

avniculae:
- Implements temporary approvals in AllowancePositionManager by using transient storage (following draft ERC7674)

### DIFF
--- a/src/position-manager/AllowancePositionManager.sol
+++ b/src/position-manager/AllowancePositionManager.sol
@@ -303,9 +303,9 @@ contract AllowancePositionManager is
   /// @dev Temporary allowance takes precedence over stored allowance, and does not cumulate.
   function _spendWithdrawAllowance(
     address spoke,
+    uint256 reserveId,
     address owner,
     address spender,
-    uint256 reserveId,
     uint256 amount
   ) internal {
     uint256 temporaryAllowance = _temporaryWithdrawAllowancesSlot({
@@ -339,9 +339,9 @@ contract AllowancePositionManager is
   /// @dev Temporary allowance takes precedence over stored allowance, and does not cumulate.
   function _spendCreditDelegation(
     address spoke,
+    uint256 reserveId,
     address owner,
     address spender,
-    uint256 reserveId,
     uint256 amount
   ) internal {
     uint256 temporaryAllowance = _temporaryDelegateCreditsSlot({


src/position-manager/AllowancePositionManager.sol
```diff
@@ -48,14 +62,25 @@ contract AllowancePositionManager is
...
 
-    _withdrawAllowances[user][params.spender][params.reserveId] = params.amount;
-    emit WithdrawApproval(user, params.spender, params.reserveId, params.amount);
+  /// @inheritdoc IAllowancePositionManager
+  function temporaryApproveWithdraw(address spender, uint256 reserveId, uint256 amount) external {
+    _temporaryWithdrawAllowancesSlot(msg.sender, spender, reserveId).tstore(amount);
+  }
+
+  /// @inheritdoc IAllowancePositionManager
+  function temporaryApproveWithdrawWithSig(
```
Kogaroshi: wouldn't I just be able to execute your intent outside the block it was meant to, consume the signature and making the actual expected execution to revert ?
I was thinking to only have `temporaryApproveWithdraw`, no sig version

DhairyaSethi: yes transient cannot have withSig variants

src/position-manager/AllowancePositionManager.sol
```diff
@@ -163,12 +199,122 @@ contract AllowancePositionManager is
...
+    uint256 reserveId,
+    uint256 amount
+  ) internal {
+    uint256 temporaryAllowance = _temporaryWithdrawAllowancesSlot(onBehalfOf, spender, reserveId)
+      .tload();
+    if (temporaryAllowance > 0) {
+      require(
+        temporaryAllowance >= amount,
+        InsufficientWithdrawAllowance(temporaryAllowance, amount)
+      );
```
DhairyaSethi: lets document we're doing either/or & not both like the OZ draft

src/position-manager/AllowancePositionManager.sol
```diff
@@ -24,15 +26,27 @@ contract AllowancePositionManager is
...
+  using TransientSlot for *;
 
   /// @notice Mapping of withdraw allowances.
   mapping(address owner => mapping(address spender => mapping(uint256 reserveId => uint256 amount)))
     private _withdrawAllowances;
 
+  /// @notice Slot for the temporary withdraw allowances.
+  /// @dev keccak256('temporary.withdrawAllowances')
+  bytes32 private constant _TEMPORARY_WITHDRAW_ALLOWANCES_SLOT =
+    0x1c6a61279a13a86a789311ddf30aee38e2f4a9f6c4aad1ff4a2e75a4018e68c3;
```
DhairyaSethi: -1 & ~0ff for collision resistance :gigachad:

src/position-manager/AllowancePositionManager.sol
avniculae: [a443ac9](https://github.com/aave/aave-v4/pull/1059/commits/a443ac9c1d4340ddd4dcb0df274ebaa8b5702bf4)

src/position-manager/AllowancePositionManager.sol
avniculae: [a443ac9](https://github.com/aave/aave-v4/pull/1059/commits/a443ac9c1d4340ddd4dcb0df274ebaa8b5702bf4)

src/position-manager/AllowancePositionManager.sol
avniculae: As discussed, I'm not convinced what values these bring. Curious to understand other's thoughts as well

snapshots/PositionManager.Operations.json
```diff
@@ -1,13 +1,13 @@
...
-  "AllowancePositionManager: approveWithdraw": "46836",
-  "AllowancePositionManager: approveWithdrawWithSig": "63047",
-  "AllowancePositionManager: borrowOnBehalfOf": "308793",
-  "AllowancePositionManager: creditDelegation": "46814",
-  "AllowancePositionManager: creditDelegationWithSig": "63134",
-  "AllowancePositionManager: renounceCreditDelegation": "24967",
-  "AllowancePositionManager: renounceWithdrawAllowance": "24989",
-  "AllowancePositionManager: withdrawOnBehalfOf: full": "123488",
-  "AllowancePositionManager: withdrawOnBehalfOf: partial": "134160",
+  "AllowancePositionManager: approveWithdraw": "46858",
```
avniculae: We can easily add gas snapshots for temporary approvals, but how to properly test gas usage of borrowOnBehalfOf and withdrawOnBehalfOf, while temporary approvals are in place. Challenge is that we are in isolate mode, there are not cheatcodes for transient storage manipulation, and we don't want the temporary approval to count towards the gas consumed by consumer functions. Only solution I see is: create a bundle function in the wrapper, measure bundle fn gas, measure approve fn gas, and then subtract the two. Any suggestions?

src/position-manager/interfaces/IAllowancePositionManager.sol
avniculae: using `creditDelegation` for approving and for reading allowance is v confusing to me. why did we decide to do this @Kogaroshi @DhairyaSethi ? Can we maybe rename this fn to `delegateCredit`?

snapshots/PositionManager.Operations.json
DhairyaSethi: wrapper can call vm.snapshotGasLastCall, no?

src/position-manager/AllowancePositionManager.sol
DhairyaSethi: they bring collision resistance from potentially using this slot (which we clearly don't here) & are generally considered best practice

src/position-manager/AllowancePositionManager.sol
```diff
@@ -58,6 +72,11 @@ contract AllowancePositionManager is
     _updateWithdrawAllowance(params.owner, params.spender, params.reserveId, params.amount, true);
   }
 
+  /// @inheritdoc IAllowancePositionManager
+  function temporaryApproveWithdraw(address spender, uint256 reserveId, uint256 amount) external {
+    _temporaryWithdrawAllowancesSlot(msg.sender, spender, reserveId).tstore(amount);
```
DhairyaSethi: pls use named args here since we have two consecutive address params

src/position-manager/AllowancePositionManager.sol
```diff
@@ -58,6 +72,11 @@ contract AllowancePositionManager is
...
+  }
+
   /// @inheritdoc IAllowancePositionManager
   function creditDelegation(address spender, uint256 reserveId, uint256 amount) external {
     _updateCreditDelegation(msg.sender, spender, reserveId, amount, true);
```
DhairyaSethi: underlying PR also needs named args, we can also handle it directly there once too cc @Kogaroshi

snapshots/PositionManager.Operations.json
avniculae: Hmm not sure it works. if vm.snapshotGasLastCall works correctly for this purpose, it must not consider the temporaryApprove call. If it does not consider it, borrow/withdraw will fail (since transient storage is cleared).

src/position-manager/interfaces/IAllowancePositionManager.sol
```diff
@@ -44,6 +44,13 @@ interface IAllowancePositionManager is IPositionManagerBase {
     bytes calldata signature
   ) external;
 
+  /// @notice Temporarily approves a spender to withdraw assets from the specified reserve on the spoke.
+  /// @dev The allowance is discarded after the transaction.
```
DhairyaSethi, line 71: temp allowance takes precedence over stored allowance should be on ext natspec for both methods also, wdyt?

src/position-manager/AllowancePositionManager.sol
avniculae: If we do -1 & ~0xff everywhere, these measures do not bring any value: we will still have a collision if same string is used.

src/position-manager/AllowancePositionManager.sol
```diff
@@ -219,4 +227,90 @@ contract AllowancePositionManager is
...
+    address spender,
+    uint256 reserveId,
+    uint256 amount
+  ) internal {
+    uint256 temporaryAllowance = _temporaryWithdrawAllowancesSlot(onBehalfOf, spender, reserveId)
+      .tload();
+    if (temporaryAllowance > 0) {
+      require(
+        temporaryAllowance >= amount,
+        InsufficientWithdrawAllowance(temporaryAllowance, amount)
```
DhairyaSethi: do we need a diff err name when using transient allowance? probably not?

tests/unit/position-managers/AllowancePositionManager.t.sol
```diff
@@ -165,9 +166,33 @@ contract AllowancePositionManagerTest is SpokeBase {
...
+    assertEq(positionManager.temporaryWithdrawAllowance(alice, spender, reserveId), amount);
+  }
+
+  /// forge-config: default.isolate = true
+  function test_temporaryApproveWithdraw_TransientStorage() public {
+    // make sure transient storage is used for temporary withdraw allowances
+    vm.prank(alice);
+    positionManager.temporaryApproveWithdraw(bob, _daiReserveId(spoke), 100e18);
+    assertEq(positionManager.temporaryWithdrawAllowance(alice, bob, _daiReserveId(spoke)), 0);
+  }
```
DhairyaSethi, line 241: another way to test this would be using `vm.accesses()` - assertEq(writes.length, 0);

src/position-manager/AllowancePositionManager.sol
DhairyaSethi: why would we do it everywhere? we just it exceptionally here

DhairyaSethi: the idea of this was not a best practice to use strings + mappings, its just to bring additional randomness on slot derivation

snapshots/PositionManager.Operations.json
DhairyaSethi: for the wrapper `tempApprove` & `consumeAllowance` are 2 ext calls, if snapshotGasLastCall considers what's inside its context this would work but could be wrong here

DhairyaSethi: if not then we could do run this test w `isolate=false` and manually `vm.cold()` targets but this is more error prone; thinking more but probably subtract from cumulative usage is a good enough estimate

src/position-manager/AllowancePositionManager.sol
Kogaroshi: a diff one would be better for clarity, since we decide to use only 1 type of Allowance per call

snapshots/PositionManager.Operations.json
avniculae: yeah ended up doing the last suggestion as well [26743fa](https://github.com/aave/aave-v4/pull/1059/commits/26743fa948970c2b73e7612e83d501d850ffcb93). Not sure if cooling is necessary

snapshots/PositionManager.Operations.json
```diff
@@ -1,13 +1,17 @@
...
-  "AllowancePositionManager: withdrawOnBehalfOf: full": "123417",
-  "AllowancePositionManager: withdrawOnBehalfOf: partial": "134071",
+  "AllowancePositionManager: renounceCreditDelegation": "24900",
+  "AllowancePositionManager: renounceWithdrawAllowance": "24922",
+  "AllowancePositionManager: temporaryApproveWithdraw": "22702",
+  "AllowancePositionManager: temporaryCreditDelegation": "22681",
+  "AllowancePositionManager: withdrawOnBehalfOf: full": "123889",
+  "AllowancePositionManager: withdrawOnBehalfOf: full (with temporary allowance)": "56228",
+  "AllowancePositionManager: withdrawOnBehalfOf: partial": "134661",
+  "AllowancePositionManager: withdrawOnBehalfOf: partial (with temporary allowance)": "67028",
```
avniculae: why we don't have partial subcase for borrow? cc @Kogaroshi

snapshots/PositionManager.Operations.json
DhairyaSethi: cooling is very necessary otherwise loading the target for eth_call (of consumeAllowance) is warm

snapshots/PositionManager.Operations.json
Kogaroshi: followed the same pattern as NativeGateway gas snapshots here, do you think it's worth adding ?

src/position-manager/AllowancePositionManager.sol
Kogaroshi: changed made in https://github.com/aave/aave-v4/pull/1054, you can update based on it @avniculae

snapshots/PositionManager.Operations.json
DhairyaSethi: what is partial borrow?

borrow only matters if its first or second
first: borrow balance storage 0 -> non 0
second: non 0 -> non 0

similarly for withdraw we care about partial and full
partial: supply balance non 0 -> non 0
full: non 0 -> 0

src/position-manager/AllowancePositionManager.sol
avniculae: > why would we do it everywhere?

if we follow this pattern (-1 & ~0xff) everywhere, then collisions are still guaranteed when the same underlying string is used. 

> its just to bring additional randomness on slot derivation

additional randomness relative to what?

snapshots/PositionManager.Operations.json
avniculae: partial wrt to allowance, I guess it's not relevant

snapshots/PositionManager.Operations.json
avniculae: if temporary approval is active, then it must have happened in the same tx, so not cooling might make sense here

src/position-manager/AllowancePositionManager.sol
avniculae: [53340e3](https://github.com/aave/aave-v4/pull/1059/commits/53340e32a2e65fb52255c6a1a6f85c6fcdaf3e11)

src/position-manager/interfaces/IAllowancePositionManager.sol
avniculae: [53340e3](https://github.com/aave/aave-v4/pull/1059/commits/53340e32a2e65fb52255c6a1a6f85c6fcdaf3e11)

src/position-manager/AllowancePositionManager.sol
avniculae: [53340e3](https://github.com/aave/aave-v4/pull/1059/commits/53340e32a2e65fb52255c6a1a6f85c6fcdaf3e11)

src/position-manager/AllowancePositionManager.sol
DhairyaSethi: to potentially using a string for slot derivation, we dont know but its just good practice. 
'if we follow this pattern (-1 & ~0xff) everywhere' ==> see we wouldn't use it everywhere, just here 
[oz also does ](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/draft-ERC20TemporaryApproval.sol#L24)

avniculae: oz uses (-1 & ~0xff) for all storage slots
- https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/88a6346545c2928a622b3561bfa21c3df89066b6/contracts/access/manager/AccessManagedUpgradeable.sol#L29
- https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/88a6346545c2928a622b3561bfa21c3df89066b6/contracts/access/OwnableUpgradeable.sol#L27
- etc

avniculae: [01225ac](https://github.com/aave/aave-v4/pull/1059/commits/01225ac05612e3fc04152cf9937d5db0055fc5f5)

tests/unit/position-managers/AllowancePositionManager.t.sol
```diff
@@ -372,11 +453,92 @@ contract AllowancePositionManagerTest is SpokeBase {
...
+        address(spoke1),
+        _daiReserveId(spoke1),
+        alice,
+        bob
+      ),
+      (approvalType == 2) ? temporaryAllowanceBefore - (supplyAmount * 2) : temporaryAllowanceBefore
+    );
+  }
+
+  function test_withdrawOnBehalfOf_fuzz_allBalance_noAllowanceDecrease(
```
Kogaroshi: nice addition

src/position-manager/AllowancePositionManager.sol
```diff
@@ -243,34 +300,108 @@ contract AllowancePositionManager is IAllowancePositionManager, PositionManagerB
...
+          reserveId: reserveId,
+          owner: owner,
+          spender: spender
+        }).tstore(temporaryAllowance.uncheckedSub(amount));
+      }
+    } else {
+      uint256 allowance = _withdrawAllowances[spoke][reserveId][owner][spender];
+      require(allowance >= amount, InsufficientWithdrawAllowance(allowance, amount));
+      if (allowance != type(uint256).max) {
+        _withdrawAllowances[spoke][reserveId][owner][spender] = allowance.uncheckedSub(amount);
```
DhairyaSethi, line 334: emit event (ideally use _use internal method)

PR #1062, opened 2025-12-15
feat: skim on spoke (enable pushing liquidity)

DhairyaSethi:
Introduces `supplySkimmed` & `repaySkimmed` on Spoke; actions identical to their counterparts (`supply`, `repay`) with the exception to liquidity transfer. These methods do not consume allowance and pull required liquidity from the user wallet, and rather expect it be to be transferred outside the context of these actions.

This enables better UX by batching transfers, and cheaper gateways which do not need to temporarily hold tokens by pulling tokens from user, approving spoke & having spoke pull from itself; can pull tokens from user wallet to hub directly and use the skimmed variants.

This also allows anyone to permisionless-ly claim donated tokens to the hub. 

Todo (blocked due to code-size & std constraints atm)
- inline `caller` arg to avoid using `msg.sender` on internal methods
- introduce flag to gatekeep this feature on spokes


src/spoke/Spoke.sol
```diff
@@ -95,7 +95,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
   /// @notice Modifier that checks if the caller is an approved positionManager for `onBehalfOf`.
   modifier onlyPositionManager(address onBehalfOf) {
-    require(_isPositionManager({user: onBehalfOf, manager: msg.sender}), Unauthorized());
+    _onlyPositionManager({onBehalfOf: onBehalfOf, caller: msg.sender});
```
Kogaroshi: code size hack ?

DhairyaSethi: yeah jump instead of inlining but think compiler was already doing it

src/spoke/interfaces/ISpokeBase.sol
```diff
@@ -107,6 +107,12 @@ interface ISpokeBase {
     address onBehalfOf
   ) external returns (uint256, uint256);
 
+  function supplySkimmed(
```
miguelmtzinf, line 121: so you would say that any spoke should have this functionality?

snapshots/Spoke.Operations.ZeroRiskPremium.json
```diff
@@ -1,33 +1,33 @@
...
+  "permitReserve + supply (multicall)": "146868",
+  "permitReserve + supply + enable collateral (multicall)": "160635",
+  "repay: full": "126229",
+  "repay: partial": "131118",
+  "setUserPositionManagerWithSig: disable": "44868",
+  "setUserPositionManagerWithSig: enable": "68897",
+  "supply + enable collateral (multicall)": "140686",
+  "supply: 0 borrows, collateral disabled": "123750",
+  "supply: 0 borrows, collateral enabled": "106672",
+  "supply: second action, same reserve": "106650",
```
Kogaroshi: pls add skim version of those snapshots

src/spoke/Spoke.sol
```diff
@@ -232,17 +232,16 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-    Reserve storage reserve = _getReserve(reserveId);
-    UserPosition storage userPosition = _userPositions[onBehalfOf][reserveId];
-    _validateSupply(reserve.flags);
-
-    IERC20(reserve.underlying).safeTransferFrom(msg.sender, address(reserve.hub), amount);
-    uint256 suppliedShares = reserve.hub.add(reserve.assetId, amount);
-    userPosition.suppliedShares += suppliedShares.toUint120();
-
-    emit Supply(reserveId, msg.sender, onBehalfOf, suppliedShares, amount);
+    return _supply(reserveId, amount, onBehalfOf, false);
```
Kogaroshi, line 231: for both supply and repay, since the transfer caller -> hub does not impact calculations if made earlier than in current logic, might be interesting to do the transfer before calling the internal function, so we don't need to pass a true/false arg & perform an extra if check

Kogaroshi: and would allow to revert earlier if no allowance given.

And with the `nonReentrant` guards we want to add to those functions, this behavior should be safe enough, wdyt ?

src/spoke/interfaces/ISpokeBase.sol
DhairyaSethi: yes given the UX benefits, we can put this behind a gov feature flag to disable for spokes which might want flexibility (maybe RWA regulations don't like batching for ex)

src/spoke/Spoke.sol
DhairyaSethi: nice one! yeah it just increases the diff, mainly on restore since we need to cap transferred amount to max debt and wanted to avoid this complexity

PR #1064, opened 2025-12-15
chore: set self as posm consistency

DhairyaSethi:
`SignatureGateway`: refactor `setSelfPositionManagerWithSig` to accept only required inputs and avoid overloaded `positionManager` to `self`.
Also adds missing tests for this method.


src/position-manager/SignatureGateway.sol
```diff
@@ -170,18 +170,21 @@ contract SignatureGateway is ISignatureGateway, GatewayBase, NoncesKeyed, Multic
   /// @inheritdoc ISignatureGateway
   function setSelfAsUserPositionManagerWithSig(
     address spoke,
-    EIP712Types.SetUserPositionManager calldata params,
```
yan-man, line 173: think this struct now is never used in src, can rm here or in valentin's posm pr

src/position-manager/interfaces/ISignatureGateway.sol
```diff
@@ -95,12 +95,18 @@ interface ISignatureGateway is IMulticall, INoncesKeyed, IGatewayBase {
   /// with a typed signature from `user`.
   /// @dev The signature is consumed on the the specified registered `spoke`.
   /// @dev The given data is passed to the `spoke` for the signature to be verified.
-  /// @param spoke The address of the spoke.
-  /// @param params The structured setSelfAsUserPositionManager parameters.
+  /// @param spoke The address of the registered spoke.
```
yan-man, line 98: why `registered` added, you mean this sig gateway is registered on the spoke, not that spoke necessarily needs to be registered on hub yes?

yan-man (approved):

src/position-manager/SignatureGateway.sol
Kogaroshi: we need to refactor the types & typehash to move them to related Interfaces, and split up the EIP712 Hashes lib for each use cases, so ig better to make a PR just for that refactor, so that this PR and other impacted by signature type rft can rebase afterwards and prevent too much conflicts

src/position-manager/interfaces/ISignatureGateway.sol
Kogaroshi: no, that the given `spoke` needs to be registered as valid target by the Gateway/PosM. See `onlyRegisteredSpoke` modifier.

PR #1066, opened 2025-12-15
feat: multi posm intent

DhairyaSethi:
Refactor `SetUserPositionManagerWithSig` intent to accept an array of position manager updates.
Removes UX check disallowing enabling an inactive position manager.

### DIFF
--- a/tests/Constants.sol
+++ b/tests/Constants.sol
@@ -15,8 +15,5 @@ library Constants {
   uint256 public constant DUST_LIQUIDATION_THRESHOLD = 1000e26;
   uint24 public constant MAX_ALLOWED_COLLATERAL_RISK = 1000_00; // 1000.00%
   uint256 public constant MAX_ALLOWED_DYNAMIC_CONFIG_KEY = type(uint24).max;
-  bytes32 public constant SET_USER_POSITION_MANAGERS_TYPEHASH =
-    // keccak256('SetUserPositionManagers(address user,PositionManagerUpdate[] updates,uint256 nonce,uint256 deadline)PositionManagerUpdate(address positionManager,bool approve)')
-    0xa9a500485f4e7c738838a1c065fe46501b5a92142c290f6a51aa56f61810c5b0;
   uint256 public constant MAX_ALLOWED_ASSET_ID = type(uint16).max;
 }
--- a/tests/mocks/EIP712Types.sol
+++ b/tests/mocks/EIP712Types.sol
@@ -5,6 +5,7 @@ pragma solidity ^0.8.20;
 /// @title EIP712Types library
 /// @author Aave Labs
 /// @notice Defines type structs used in EIP712-typed signatures.
+/// @dev Consolidated types to generate JsonBindings.sol using `forge bind-json` for vm.eip712* cheat-codes.
 library EIP712Types {
   /// @dev Spoke Intents
   struct SetUserPositionManagers {
--- a/tests/unit/Spoke/Spoke.SetUserPositionManagerWithSig.t.sol
+++ b/tests/unit/Spoke/Spoke.SetUserPositionManagerWithSig.t.sol
@@ -71,21 +71,17 @@ contract SpokesetUserPositionManagersWithSigTest is SpokeBase {
 
   function test_setUserPositionManager_typeHash() public view {
     assertEq(
-      Constants.SET_USER_POSITION_MANAGERS_TYPEHASH,
-      EIP712Hash.SET_USER_POSITION_MANAGERS_TYPEHASH
-    );
-    assertEq(
-      Constants.SET_USER_POSITION_MANAGERS_TYPEHASH,
+      EIP712Hash.SET_USER_POSITION_MANAGERS_TYPEHASH,
       vm.eip712HashType('SetUserPositionManagers')
     );
     assertEq(
-      Constants.SET_USER_POSITION_MANAGERS_TYPEHASH,
+      EIP712Hash.SET_USER_POSITION_MANAGERS_TYPEHASH,
       keccak256(
         'SetUserPositionManagers(address user,PositionManagerUpdate[] updates,uint256 nonce,uint256 deadline)PositionManagerUpdate(address positionManager,bool approve)'
       )
     );
     assertEq(
-      Constants.SET_USER_POSITION_MANAGERS_TYPEHASH,
+      EIP712Hash.SET_USER_POSITION_MANAGERS_TYPEHASH,
       spoke1.SET_USER_POSITION_MANAGERS_TYPEHASH()
     );
   }
@@ -96,10 +92,6 @@ contract SpokesetUserPositionManagersWithSigTest is SpokeBase {
       keccak256('PositionManagerUpdate(address positionManager,bool approve)')
     );
     assertEq(EIP712Hash.POSITION_MANAGER_UPDATE, vm.eip712HashType('PositionManagerUpdate'));
-    assertEq(
-      EIP712Hash.POSITION_MANAGER_UPDATE,
-      keccak256('PositionManagerUpdate(address positionManager,bool approve)')
-    );
   }
 
   function test_setUserPositionManagersWithSig_revertsWith_InvalidSignature_dueTo_ExpiredDeadline()


src/spoke/Spoke.sol
```diff
@@ -5,9 +5,10 @@ pragma solidity 0.8.28;
 import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
 import {SafeERC20, IERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
 import {IERC20Permit} from 'src/dependencies/openzeppelin/IERC20Permit.sol';
-import {SignatureChecker} from 'src/dependencies/openzeppelin/SignatureChecker.sol';
+import {SignatureChecker} from 'src/libraries/misc/SignatureChecker.sol';
```
Kogaroshi: let's move with the other lib imports

src/spoke/Spoke.sol
```diff
@@ -5,9 +5,10 @@ pragma solidity 0.8.28;
...
 import {AccessManagedUpgradeable} from 'src/dependencies/openzeppelin-upgradeable/AccessManagedUpgradeable.sol';
 import {EIP712} from 'src/dependencies/solady/EIP712.sol';
+import {EIP712Hash, EIP712Types} from 'src/position-manager/libraries/EIP712Hash.sol';
```
Kogaroshi: I still think this should be a common lib for all contracts in the repo @miguelmtzinf

DhairyaSethi: 100% agree, didn't do it here bc it's done in another PR and wanted to only keep relevant changes

src/spoke/Spoke.sol
```diff
@@ -693,7 +684,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   function _setUserPositionManager(address positionManager, address user, bool approve) internal {
     PositionManagerConfig storage config = _positionManager[positionManager];
     // only allow approval when position manager is active for improved UX
-    require(!approve || config.active, InactivePositionManager());
+    require(!approve || config.active, InactivePositionManager()); // todo rm this ux check given sig batching?
```
Kogaroshi: might be good to have, we don't want to approve an inactive one via the batch by mistake and be surprised if activated by Governance later

DhairyaSethi: my concern is that keeping this fails an entire batch bc of one such case 

since it is only a UX check, i'm leaning towards removing it 
unsure how likely a scenario it is for a user to pre approve a posm which is in line to be (re)approved by gov

miguelmtzinf: i d keep it, or no-op for ease of integraiton (e.g. we only approve if globally active)

src/position-manager/SignatureGateway.sol
```diff
@@ -170,19 +170,14 @@ contract SignatureGateway is ISignatureGateway, GatewayBase, NoncesKeyed, Multic
   /// @inheritdoc ISignatureGateway
   function setSelfAsUserPositionManagerWithSig(
     address spoke,
-    EIP712Types.SetUserPositionManager calldata params,
+    EIP712Types.SetUserPositionManager memory params,
```
miguelmtzinf: I believe it's fine to apply https://github.com/aave/aave-v4/pull/1064 instead and revert this change. We can build the struct with 1-length array locally

DhairyaSethi: yep thats the plan, sg

src/spoke/Spoke.sol
DhairyaSethi: either is risky tbh 
keeping it - one accidental update destroys batch 
no-op - it's not really a no-op due to nonce consumption; user would notice a silent fail

src/libraries/misc/SignatureChecker.sol
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {SignatureChecker as OpenZeppelinSignatureChecker} from 'src/dependencies/openzeppelin/SignatureChecker.sol';
+
+/// @title SignatureChecker
+/// @author Aave Labs
+library SignatureChecker {
```
Kogaroshi: to be removed once the Spoke code size relief PR is merged in.

src/position-manager/SignatureGateway.sol
```diff
@@ -176,15 +176,21 @@ contract SignatureGateway is ISignatureGateway, GatewayBase, NoncesKeyed, Multic
...
-        user: user,
-        approve: approve,
-        nonce: nonce,
-        deadline: deadline,
-        signature: signature
-      })
+      ISpoke(spoke).setUserPositionManagerWithSig(
+        EIP712Types.SetUserPositionManager({
+          updates: updates,
+          user: user,
```
miguelmtzinf:
```suggestion
          user: user,
          updates: updates,
```

src/spoke/Spoke.sol
miguelmtzinf: we need to fix this

DhairyaSethi: being fixed in the other PR which abstracts intent consumption

src/spoke/interfaces/ISpoke.sol
```diff
@@ -522,10 +533,6 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IMulticall {
   /// @return The address of the library.
   function getLiquidationLogic() external pure returns (address);
 
-  /// @notice Returns the type hash for the SetUserPositionManager intent.
-  /// @return The bytes-encoded EIP-712 struct hash representing the intent.
-  function SET_USER_POSITION_MANAGER_TYPEHASH() external view returns (bytes32);
```
Kogaroshi, line 527: we should keep the view methods for type hash, like in other contracts.

DhairyaSethi: yea had RMed for code size, we can bring back

DhairyaSethi: would technically also need PositionManagerUpdate bc of the recursive hash but avoiding it

src/spoke/Spoke.sol
```diff
@@ -45,6 +45,10 @@ abstract contract Spoke is
   using UserPositionDebt for ISpoke.UserPosition;
   using EIP712Hash for *;
 
+  /// @inheritdoc ISpoke
+  bytes32 public constant SET_USER_POSITION_MANAGERS_TYPEHASH =
```
Kogaroshi, line 9: I'd rather we have this as a view function to stay consistant with other contracts.

DhairyaSethi: then it's consistent w oracle. i'll update the rest

Kogaroshi (approved):

src/position-manager/libraries/EIP712Hash.sol
```diff
@@ -8,6 +8,8 @@ import {ISignatureGateway} from 'src/position-manager/interfaces/ISignatureGatew
 /// @author Aave Labs
 /// @notice Helper methods to hash EIP712 typed data structs.
 library EIP712Hash {
+  using EIP712Hash for *;
```
miguelmtzinf: seems unneded

DhairyaSethi: it's needed for the intentHashes loop 

it uses syntax sugar to hash positonManager update

DhairyaSethi: ah no on this file it's not required, will rm

src/spoke/Spoke.sol
```diff
@@ -42,11 +43,11 @@ abstract contract Spoke is
   using PositionStatusMap for *;
   using ReserveFlagsMap for ReserveFlags;
   using UserPositionDebt for ISpoke.UserPosition;
+  using EIP712Hash for *;
```
```suggestion
using EIP712Hash for *;
using KeyValueList for KeyValueList.List;
using LiquidationLogic for *;
using PositionStatusMap for *;
using ReserveFlagsMap for ReserveFlags;
using UserPositionDebt for ISpoke.UserPosition;
```
following imports order

src/spoke/interfaces/ISpoke.sol
```diff
@@ -14,6 +14,26 @@ type ReserveFlags is uint8;
...
+    address positionManager;
+    bool approve;
+  }
+
+  /// @notice Intent data to set user position manager with EIP712-typed signature.
+  /// @param user The address of the user on whose behalf position manager can act.
+  /// @param updates The array of position manager updates.
+  /// @param nonce The nonce for the signature.
+  /// @param deadline The deadline for the signature.
+  struct SetUserPositionManagers {
```
miguelmtzinf, line 22: maybe this one first?

src/spoke/interfaces/ISpoke.sol
```diff
@@ -14,6 +14,26 @@ type ReserveFlags is uint8;
...
 interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IMulticall {
+  /// @notice Sub-Intent data to apply position manager update for user.
+  /// @param positionManager The address of the position manager.
+  /// @param approve True to approve the position manager, false to revoke approval.
+  struct PositionManagerUpdate {
+    address positionManager;
+    bool approve;
+  }
+
+  /// @notice Intent data to set user position manager with EIP712-typed signature.
```
```suggestion
  /// @notice Intent data to set user position managers with EIP712-typed signature.
```

src/spoke/interfaces/ISpoke.sol
```diff
@@ -393,25 +410,19 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IMulticall {
...
-  /// @notice Enables a user to grant or revoke approval for a position manager using an EIP712-typed intent.
+  /// @notice Enables a user to grant or revoke approval for an array of position managers using an EIP712-typed intent.
   /// @dev Uses keyed-nonces where for each key's namespace nonce is consumed sequentially.
-  /// @param positionManager The address of the position manager.
-  /// @param user The address of the user on whose behalf position manager can act.
-  /// @param approve True to approve the position manager, false to revoke approval.
-  /// @param nonce The key-prefixed nonce for the signature.
-  /// @param deadline The deadline for the signature.
+  /// @dev Allows duplicated updates and the last one is persisted. Allows approving inactive position managers.
+  /// @param params The structured setUserPositionManager parameters.
```
```suggestion
  /// @param params The structured setUserPositionManagers parameters.
```

tests/mocks/EIP712Types.sol
```diff
@@ -6,14 +6,18 @@ pragma solidity ^0.8.20;
 /// @author Aave Labs
 /// @notice Defines type structs used in EIP712-typed signatures.
 library EIP712Types {
```
miguelmtzinf, line 9: let's add separators, for each context / directory

tests/Constants.sol
```diff
@@ -15,8 +15,8 @@ library Constants {
   uint256 public constant DUST_LIQUIDATION_THRESHOLD = 1000e26;
   uint24 public constant MAX_ALLOWED_COLLATERAL_RISK = 1000_00; // 1000.00%
   uint256 public constant MAX_ALLOWED_DYNAMIC_CONFIG_KEY = type(uint24).max;
-  bytes32 public constant SET_USER_POSITION_MANAGER_TYPEHASH =
-    // keccak256('SetUserPositionManager(address positionManager,address user,bool approve,uint256 nonce,uint256 deadline)')
-    0x758d23a3c07218b7ea0b4f7f63903c4e9d5cbde72d3bcfe3e9896639025a0214;
+  bytes32 public constant SET_USER_POSITION_MANAGERS_TYPEHASH =
+    // keccak256('SetUserPositionManagers(address user,PositionManagerUpdate[] updates,uint256 nonce,uint256 deadline)PositionManagerUpdate(address positionManager,bool approve)')
+    0xa9a500485f4e7c738838a1c065fe46501b5a92142c290f6a51aa56f61810c5b0;
```
miguelmtzinf: why do we need this still?

DhairyaSethi: general place to find at one place, can rm this specifically but kept to keep spoke structs complete

miguelmtzinf: it's okay to use them for tests, but I think for SignatureGateway we are relying on the eip712 lib itself

tests/mocks/EIP712Types.sol
miguelmtzinf, line 23: this is just for tests? let's add a comment

tests/Constants.sol
DhairyaSethi: will rm typehashes from here then and only keep value constants

tests/mocks/EIP712Types.sol
DhairyaSethi: no tests use the types from interfaces 
this for ease of forge bind-json command, i'll add a comment here

PR #1071, opened 2025-12-17
fix: safecast on `signedSub`

### LINKED ISSUE #1085: Add a safecast check to the int cast in `signedSub`
In certain place, uint values can be big enough to be casted to a negative value, ex. in new calculations of premium offset ray. Therefore we should add a safecast check to the int cast in `signedSub`.

yan-man:
- add safecast on `signedSub` for safety

closes #1085

### DIFF
--- a/src/libraries/math/MathUtils.sol
+++ b/src/libraries/math/MathUtils.sol
@@ -54,7 +54,6 @@ library MathUtils {
   }
 
   /// @notice Returns the difference of two unsigned integers as a signed integer.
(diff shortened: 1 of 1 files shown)


src/libraries/math/MathUtils.sol
```diff
@@ -52,7 +56,7 @@ library MathUtils {
   /// @notice Returns the difference of two unsigned integers as a signed integer.
   /// @dev Does not ensure the `a` and `b` values are within the range of a signed integer.
```
DhairyaSethi:
```suggestion
```

PR #1067, opened 2025-12-16
chore: restricted eliminate deficit

DhairyaSethi:
Add restricted modifier to `Hub.eliminateDeficit` to gatekeep this flow only for authorised active spokes.

PR #1075, opened 2025-12-19
fix: Fix paused state check on Hub functions

yan-man:
Following the approach in #1083 , this PR fixes the `paused` state check on the Hub, allowing calls to functions that are intended to be part of larger Spoke level operations: `eliminateDeficit`, `payFeeShares`, `reportDeficit` and `refreshPremium`.


src/hub/Hub.sol
```diff
@@ -358,9 +358,8 @@ contract Hub is IHub, AccessManaged {
     SpokeData storage coveredSpoke = _spokes[assetId][spoke];
 
     asset.accrue();
-    _validateEliminateDeficit(callerSpoke, amount);
-
     uint256 deficitRay = coveredSpoke.deficitRay;
+    _validateEliminateDeficit(callerSpoke, amount, deficitRay);
     uint256 deficitAmountRay = (amount < deficitRay.fromRayUp()) ? amount.toRay() : deficitRay;
 
     uint120 shares = asset.toAddedSharesUp(deficitAmountRay.fromRayUp()).toUint120();
```
miguelmtzinf, line 364: we are checking `deficitRay > 0`, what if `shares` is still 0?

src/hub/Hub.sol
```diff
@@ -894,9 +893,14 @@ contract Hub is IHub, AccessManaged {
     require(premiumAmountRay <= premiumRay, SurplusPremiumRayDeficitReported(premiumRay));
   }
 
-  function _validateEliminateDeficit(SpokeData storage spoke, uint256 amount) internal view {
+  function _validateEliminateDeficit(
+    SpokeData storage spoke,
+    uint256 amount,
+    uint256 deficitRay
+  ) internal view {
     require(spoke.active, SpokeNotActive());
```
miguelmtzinf, line 895: as discussed, it's incosistent that caller spoke can execute this while being paused, which results in a decrease of added shares (similar to remove, where `paused` flag checks are in place).

src/hub/Hub.sol
DhairyaSethi: can't be bc it rounds up here, no?

src/hub/Hub.sol
```diff
@@ -894,9 +893,14 @@ contract Hub is IHub, AccessManaged {
...
     require(amount > 0, InvalidAmount());
+    require(deficitRay > 0, InsufficientDeficit());
```
DhairyaSethi: i dont think we need both, we could just check `deficitRay` > 0

DhairyaSethi: did this accidently #1067 lmk will rm from there or we can drop this pr

src/hub/Hub.sol
```diff
@@ -894,9 +893,14 @@ contract Hub is IHub, AccessManaged {
     require(premiumAmountRay <= premiumRay, SurplusPremiumRayDeficitReported(premiumRay));
   }
 
-  function _validateEliminateDeficit(SpokeData storage spoke, uint256 amount) internal view {
+  function _validateEliminateDeficit(
+    SpokeData storage spoke,
```
```suggestion
    SpokeData storage coveredSpoke,
```

src/hub/Hub.sol
yan-man: decided today we will leave it in your PR as a validation on existence, and here just validate on paused caller spoke

src/hub/Hub.sol
yan-man: this should be caller spoke, no?

DhairyaSethi: yes it's done in the other pr it's fine

src/hub/Hub.sol
yan-man: after this validation, now reportDeficit is the only one that is inconsistent, but necessary for liquidations to not get blocked by deficit on paused peripheral assets

src/hub/interfaces/IHub.sol
```diff
@@ -73,7 +73,7 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev drawCap The maximum amount that can be drawn by a spoke, expressed in whole assets (not scaled by decimals). A value of `MAX_ALLOWED_SPOKE_CAP` indicates no cap.
   /// @dev riskPremiumThreshold The maximum ratio of premium to drawn shares a spoke can have, expressed in BPS. A value of `MAX_RISK_PREMIUM_THRESHOLD` indicates no threshold.
   /// @dev active True if the spoke is prevented from performing any actions.
-  /// @dev paused True if the spoke is prevented from performing actions that instantly update the liquidity.
+  /// @dev paused True if the spoke is prevented from performing any actions except those involving peripheral assets in liquidation.
```
yan-man: note: now this must be changed, bc actions such as: payFeeShares, transferShares, eliminateDeficit, all do not instantly update liquidity but are still blocked when paused. 

Only reportDeficit and refreshPremium stand out as not abiding by this paused flag, due to the fact of not wanting peripheral assets not directly involved in the repay/withdraw action of a liquidation to block a liquidation from succeeding.

DhairyaSethi: i don't think the concept of liquidation belongs here

the thinking is also kinda narrow, for refresh premium the idea is paused assets should still allow interest rate updates since they continue accruing interest. since gov can update ir params for base rate, refresh premium does not change liquidity but updates 'premium rate'

yan-man: yeah perhaps this is too narrow involving liquidations, but dont have something more encompassing that also covers the report deficit case, want to discuss what description should go here. true for refreshPremium, but also the instantly updating liquidity isn't correct either and we must update this natspec in some way.

DhairyaSethi: we wanted to change the var name as well to avoid confusion w the pause on reserves at spoke, i think there was an issue open w some brainstormed suggestions from @miguelmtzinf?

yan-man: do you think it's too narrow to specifically mention reportDeficit and refreshPremium as the methods allowed? if hub is immutable then there won't be future deviations from this

DhairyaSethi: narrow was on the reasoning, just mentioning the omitted methods w/o the underlying thinking/logic feels like a bandaid soln

miguelmtzinf:
Can we apply this suggestion https://github.com/aave/aave-v4/issues/1083#issuecomment-3756217167?

src/hub/Hub.sol
```diff
@@ -892,14 +891,13 @@ contract Hub is IHub, AccessManaged {
     require(premiumAmountRay <= premiumRay, SurplusPremiumRayDeficitReported(premiumRay));
   }
 
-  function _validateEliminateDeficit(SpokeData storage spoke, uint256 amount) internal view {
-    require(spoke.active, SpokeNotActive());
+  function _validateEliminateDeficit(SpokeData storage callerSpoke, uint256 amount) internal view {
```
```suggestion
  function _validateEliminateDeficit(SpokeData storage caller, uint256 amount) internal view {
```
or `sender`

src/hub/interfaces/IHub.sol
```diff
@@ -73,7 +73,7 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev drawCap The maximum amount that can be drawn by a spoke, expressed in whole assets (not scaled by decimals). A value of `MAX_ALLOWED_SPOKE_CAP` indicates no cap.
   /// @dev riskPremiumThreshold The maximum ratio of premium to drawn shares a spoke can have, expressed in BPS. A value of `MAX_RISK_PREMIUM_THRESHOLD` indicates no threshold.
   /// @dev active True if the spoke is prevented from performing any actions.
-  /// @dev paused True if the spoke is prevented from performing actions that instantly update the liquidity.
+  /// @dev paused True if the spoke is prevented from performing any user facing actions. Does not restrict actions involving peripheral assets.
```
```suggestion
  /// @dev paused True if the spoke is prevented from performing any user-facing actions.
```
We could say that "It doesn’t affect functions that are intended to be called as auxiliary steps within larger actions."
but maybe better to keep it shorter and add to docs, given that there is no notion of peripheral assets from Hub pov, and to avoid confusion

Kogaroshi: cde062878001178a116f1b1acd37682e3c77ff08

src/hub/Hub.sol
Kogaroshi: I replaced with `spoke` here and in `_validatePayFeeShares` for consistency with the previous `_validate` methods.
See cde062878001178a116f1b1acd37682e3c77ff08

src/hub/Hub.sol
miguelmtzinf: it was agreed to allow paused spokes to call eliminateDeficit, given that is subject to be called as part of a larger Spoke operation

src/hub/Hub.sol
```diff
@@ -357,9 +357,8 @@ contract Hub is IHub, AccessManaged {
     SpokeData storage coveredSpoke = _spokes[assetId][spoke];
 
     asset.accrue();
-    _validateEliminateDeficit(callerSpoke, amount);
-
     uint256 deficitRay = coveredSpoke.deficitRay;
+    _validateEliminateDeficit(callerSpoke, amount);
```
DhairyaSethi, line 361: why was this moved down?
it can just take `deficitRay` as input to fix the amount issue & copy over the specific tests from #1067 or we can do it separately in that PR only to keep this constrained (in which case we can also undo this bc it looks meaningles)

src/hub/Hub.sol
```diff
@@ -897,9 +896,8 @@ contract Hub is IHub, AccessManaged {
     require(amount > 0, InvalidAmount());
   }
 
-  function _validatePayFeeShares(SpokeData storage senderSpoke, uint256 feeShares) internal view {
-    require(senderSpoke.active, SpokeNotActive());
-    require(!senderSpoke.paused, SpokePaused());
+  function _validatePayFeeShares(SpokeData storage spoke, uint256 feeShares) internal view {
+    require(spoke.active, SpokeNotActive());
```
DhairyaSethi, line 900: i would retain that this is `senderSpoke` active check

PR #1010, opened 2025-11-17
fix: add Spoke ReportDeficit event and remove complementary getters in Spoke for bytecode savings 

Kogaroshi:
- remove `getDynamicReserveConfig(uint256 reserveId)` getter to fetch the dynamic risk config of the last key. This can be done instead by fetching the last configuration key from the asset and calling `getDynamicReserveConfig(uint256 reserveId, uint256 configKey)`.
- merge `isUsingAsCollateral` && `isBorrowing` into `getUserReserveStatus`.
- add new event `ReportDeficit` on Spoke to signal deficit reporting on Spoke. Provides additional information for offchain indexing. 
- helps to reduce spoke bytecode size to provide additional buffer

PR #1078, opened 2025-12-23
chore: reorder rp update on deficit

DhairyaSethi:
revert changes done on _reportDeficit reorder done in https://github.com/aave/aave-v4/pull/1010

motivation:
```
Someone who reads the code, given the changes, would assume that the _notifyRiskPremiumUpdate would actually perform the main reason for which the function is executed: call hub.refreshPremium and notify the hubs, instead this action is never performed.

The only reason to execute it when we are in the deficit scenario is to only update the positionStatus.hasPositiveRiskPremium state variable and emit the event.

Note: there's one "edge" case for which _notifyRiskPremiumUpdate will early return and will not emit the UpdateUserRiskPremium event: when the liquidated borrower had provided only collaterals with zero collateralRisk. In that case the riskPremium of the user, tracked by positionStatus.riskPremium, would have been already equal to zero.
```

### DIFF
--- a/src/spoke/libraries/LiquidationLogic.sol
+++ b/src/spoke/libraries/LiquidationLogic.sol
@@ -252,7 +252,7 @@ library LiquidationLogic {
 
   /// @notice Reports deficits for all debt reserves of the user.
   /// @dev Deficit validation should already have occurred during liquidation.
-  /// @dev It clears the user position, setting drawn debt and premium debt to zero.
+  /// @dev It clears the user position, setting drawn debt, premium debt and user risk premium to zero.
   /// @param reserves The mapping of reserves per reserve identifier.
   /// @param userPositions The mapping of user positions per reserve per user.
   /// @param positionStatus The mapping of position status per user.


yan-man:
smh code size

Kogaroshi:
14B over the size limit, not that big of a change tho. We can see if bundled with other changes the code size impact is reduced.

src/spoke/Spoke.sol
```diff
@@ -863,6 +862,7 @@ abstract contract Spoke is
   /// @dev It clears the user position, setting drawn debt, premium debt, and risk premium to zero.
   function _reportDeficit(address user) internal {
```
miguelmtzinf: independently from applying this PR:
- It may make sense to rename this fn to `_notifyReportDeficit` or smt similar, given it's analogous to `_notifyRiskPremiumUpdate` imo.
- The natspec docs above is now corrent, but not in head dev atm (this fn does not set user rp to zero)

src/spoke/Spoke.sol
```diff
@@ -395,13 +395,12 @@ abstract contract Spoke is
...
 
-    uint256 newRiskPremium = 0;
     if (isUserInDeficit) {
       _reportDeficit(user);
     } else {
-      newRiskPremium = _calculateUserAccountData(user).riskPremium;
+      uint256 newRiskPremium = _calculateUserAccountData(user).riskPremium;
+      _notifyRiskPremiumUpdate(user, newRiskPremium);
     }
-    _notifyRiskPremiumUpdate(user, newRiskPremium);
```
miguelmtzinf, line 389: here the question, when reading this fn you can assume either
i. `_reportDeficit` only clears user debt and notify so we must `_notifyRiskPremiumUpdate` at the end no matter what (dev)
ii. `_reportDeficit` also clears risk premium so we don't need to notify later. (PR)

imo:
i. option follow single responsability principle, and makes liq fn more readable
ii. option is more self-contained and favors reusability. It's only used once and worsen liq fn readabiltiy

DhairyaSethi: single responsibility is readable but more error prone imo, since the true action is contained in two methods 

i think renaming to notifyReportDefict or reportAndNotifyDeficit is a great suggestion to improve the lost readability

yan-man (approved):
I submit

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -287,6 +288,8 @@ library LiquidationLogic {
 
       emit ISpoke.ReportDeficit(reserveId, user, deficitShares, premiumDelta);
     }
+
+    emit ISpoke.UpdateUserRiskPremium(user, 0);
```
miguelmtzinf, line 303: this is being emitted even if the risk premium was already 0. We d need to do smt like to avoid emitting false events...
```
    if (newRiskPremium == 0 && positionStatus.riskPremium == 0) {
      return;
    }
```
I d keep the code as is OR create an internal helper for updating user risk premium that i) returns early if not needed, ii) update storage and iii) emit event.

DhairyaSethi: its just a redundant event, idt it's 'false'. to me its fine as is 

internal helper will make the logic of _notifyRiskPremium more expensive

DhairyaSethi: see https://github.com/aave/aave-v4/compare/chore/deficit-reorder...poc/setuserrp?expand=1 wdyt?

miguelmtzinf: why are you fine with the redudant event in this case? it d make the logic of `_reportDeficit` more expensive no? the `_notifyRiskPremium` wouldn't be much affected.
the suggestion does not work for the same reason that calling `_notifyRiskPremiumUpdate` but not actually calling `hub.refresh` is misleading (one of the motivations for this PR).

DhairyaSethi: it adds additional comparator which is more expensive in the general case, the redundant event is only redundant in one case when prev rp is 0; im fine here because next rp bc of **report deficit is always 0**
(unlike notify risk premium)

 the logic is simpler w/o suggestion but i'm happy to apply

src/spoke/libraries/LiquidationLogic.sol
```suggestion
  /// @dev It clears the user position, setting drawn debt, premium debt and user risk premium to zero.
```

src/spoke/libraries/LiquidationLogic.sol
miguelmtzinf: We agreed to accept the redundancy, as it makes deficit scenarios more explicit when the system fully clears a user position, without impacting gas efficiency in the average case.

PR #1088, opened 2026-01-06
fix: unit price feed safecast

yan-man:
- fix remaining places where there is potentially unsafe casting

### DIFF
--- a/tests/unit/UnitPriceFeed.t.sol
+++ b/tests/unit/UnitPriceFeed.t.sol
@@ -17,6 +17,14 @@ contract UnitPriceFeedTest is Base {
     unitPriceFeed = new UnitPriceFeed(DECIMALS, _description);
   }
 
+  function test_constructor_revertsWith_Uint8Overflow() public {
+    uint8 invalidDecimals = 77;
+    vm.expectRevert(
+      abi.encodeWithSelector(SafeCast.SafeCastOverflowedUintToInt.selector, 10 ** invalidDecimals)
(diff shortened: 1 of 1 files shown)


src/misc/UnitPriceFeed.sol
```diff
@@ -23,7 +26,7 @@ contract UnitPriceFeed is AggregatorV3Interface {
   constructor(uint8 decimals_, string memory description_) {
     decimals = decimals_;
     description = description_;
-    _units = int256(10 ** decimals_);
+    _units = (10 ** decimals_).toInt256();
```
miguelmtzinf, line 29: `type(int256).max > 2**type(uint8).max` - so I believe int256 casting is fine.

DhairyaSethi: it's 10**; technically the cast is required bc the RHS can be 10 ** 255
```
> (2n ** 255n - 1n) > 10n ** (2n ** 8n - 1n)
false
```

DhairyaSethi, line 29: kinda unnecessary bc this is a one time fn and if any overflow the contract can be discarded. would be more in favour if this was fixing a run time unsafe cast

miguelmtzinf: 10^77 fits in uint256 but not in int256.
10^76 fits in int256 and int256.
we need the safecast for the first case, so we don't sinlently overflow

PR #1093, opened 2026-01-07
chore: reorder gov updates 

DhairyaSethi:
Reorder hub gov updates for sanity.

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -132,6 +132,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     require(hub != address(0), InvalidAddress());
     require(assetId <= MAX_ALLOWED_ASSET_ID, InvalidAssetId());
     require(!_reserveExists[hub][assetId], ReserveExists());
+    _reserveExists[hub][assetId] = true;
 
     _validateReserveConfig(config);
     _validateDynamicReserveConfig(dynamicConfig);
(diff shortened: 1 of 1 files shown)


src/hub/Hub.sol
```diff
@@ -111,7 +111,6 @@ contract Hub is IHub, AccessManaged {
       feeReceiver: feeReceiver,
       liquidityFee: 0
     });
-    _underlyingAssets.add(underlying);
     _addFeeReceiver(assetId, feeReceiver);
```
miguelmtzinf, line 114: i d move this up together the `add`

DhairyaSethi: aesthetics only?
it will anyway be after `_assetCount++` bc it needs assetId, hence i think it's fine here after setting `asset` struct itself
CEI is fixed with the line 81 addition and check

src/spoke/Spoke.sol
```diff
@@ -160,6 +158,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     });
     _dynamicConfig[reserveId][dynamicConfigKey] = dynamicConfig;
     _reserveExists[hub][assetId] = true;
+    _updateReservePriceSource(reserveId, priceSource);
```
```suggestion

    _updateReservePriceSource(reserveId, priceSource);
```

DhairyaSethi: instead reverted this and moved reserveExists bool up like hub.addAsset flow, lmk wdyt

src/hub/Hub.sol
yan-man: would be better to move it after `_assetCount++` but before `setInterestRateData`, no?

miguelmtzinf: ah, true. Ok, no strong opinion

DhairyaSethi: @yan-man why? reentrency is solved

PR #1094, opened 2026-01-07
opt: struct hash

DhairyaSethi:
optimise eip712 digest hashing by rewriting in assembly 

we allocate from free mem pointer, clean addresses,booleans (since it is direct user input) 
we can avoid clearing memory after fmp because as per [solidity memory safety documentation](https://docs.soliditylang.org/en/stable/assembly.html#memory-safety):
```
memory beyond the location pointed to by the free memory pointer can safely be used as temporary scratch space
```


src/spoke/libraries/EIP712Hash.sol
```diff
@@ -18,24 +16,56 @@ library EIP712Hash {
...
     0x187dbd227227274b90655fb4011fc21dd749e8966fc040bd91e0b92609202565;
 
-  function hash(ISpoke.SetUserPositionManagers calldata params) internal pure returns (bytes32) {
-    bytes32[] memory updatesHashes = new bytes32[](params.updates.length);
-    for (uint256 i = 0; i < updatesHashes.length; ++i) {
-      updatesHashes[i] = params.updates[i].hash();
+  function hash(
+    ISpoke.SetUserPositionManagers calldata params
+  ) internal pure returns (bytes32 digest) {
+    assembly ('memory-safe') {
```
DhairyaSethi:
```suggestion
    assembly {
```

src/spoke/libraries/EIP712Hash.sol
```diff
@@ -18,24 +16,56 @@ library EIP712Hash {
...
+        i := add(i, 1)
+      } {
+        let update := add(add(list, 0x20), mul(i, 0x40))
+
+        mstore(0, POSITION_MANAGER_UPDATE)
+        mstore(0x20, shr(96, shl(96, calldataload(update)))) // params.updates[i].positionManager
+        mstore(0x40, iszero(iszero(calldataload(add(update, 0x20))))) // params.updates[i].approve
+
+        mstore(add(m, mul(i, 0x20)), keccak256(0, 0x60))
+      }
```
miguelmtzinf, line 42: we could extract a helper here

miguelmtzinf:
nice contribution @DhairyaSethi ! While this brings positive effects like reduced bytecode size (and small gas savings), we agreed not to apply it in order to keep the diff minimal between security review rounds.

PR #1095, opened 2026-01-07
fix: unit price feed restriction on decimals

yan-man:


src/misc/UnitPriceFeed.sol
```diff
@@ -9,6 +9,8 @@ import {AggregatorV3Interface} from 'src/dependencies/chainlink/AggregatorV3Inte
 /// @notice Price feed that returns the unit price (1), with decimals precision.
 /// @dev This price feed can be set for reserves that use the base currency as collateral.
 contract UnitPriceFeed is AggregatorV3Interface {
+  uint8 private constant MAX_ALLOWED_DECIMALS = 8;
```
DhairyaSethi, line 12: why max 8? can be anything

yan-man: then why restrict this value, what are you proposing for the max 

or you mean just having this set to value that leads to the largest possible uint that doesnt lead to overflow?

yan-man:
without a clear value for max allowable decimals, opt for simpler approach via safecast: https://github.com/aave/aave-v4/pull/1088

PR #1098, opened 2026-01-08
feat: global position manager [WIP]

yan-man:
This PR introduces global position managers without requiring user approval (only DAO approval). This can be useful for emergencies or migrations, where users don’t need to opt in and the DAO can handle the process directly.

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -229,7 +229,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   ) external restricted {
     PositionManager storage positionManagerConfig = _positionManager[positionManager];
     positionManagerConfig.active = config.active;
-    positionManagerConfig.global = config.global;
+    positionManagerConfig.isGlobal = config.isGlobal;
     emit UpdatePositionManager(positionManager, config);
   }
 
@@ -678,7 +678,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
 
   /// @inheritdoc ISpoke
   function isPositionManagerGlobal(address positionManager) external view returns (bool) {
-    return _positionManager[positionManager].global;
+    return _positionManager[positionManager].isGlobal;
   }
 
   /// @inheritdoc ISpoke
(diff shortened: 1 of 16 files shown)


src/spoke/interfaces/ISpoke.sol
```diff
@@ -85,12 +85,22 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     uint24 dynamicConfigKey;
   }
 
-  /// @notice Position manager configuration data.
+  /// @notice Position manager data.
   /// @dev approval The mapping of position manager user approvals.
   /// @dev active True if the position manager is active.
-  struct PositionManagerConfig {
+  /// @dev global True if the position manager is global.
+  struct PositionManager {
```
Kogaroshi, line 92: I'd keep this as `PositionManagerConfig` and call the new struct `PositionManagerData`, or just pass 2 booleans, the parameters are not that complex imo.

src/spoke/Spoke.sol
```diff
@@ -221,9 +223,14 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
-  function updatePositionManager(address positionManager, bool active) external restricted {
-    _positionManager[positionManager].active = active;
-    emit UpdatePositionManager(positionManager, active);
+  function updatePositionManager(
+    address positionManager,
+    PositionManagerConfig calldata config
+  ) external restricted {
+    PositionManager storage positionManagerConfig = _positionManager[positionManager];
+    positionManagerConfig.active = config.active;
+    positionManagerConfig.isGlobal = config.isGlobal;
```
Kogaroshi, line 232: ~~Should we consider that one that has `isGlobal` as true should has `active` as true by default too ?~~ Nvm, modifier check already has that assumption, would be redundant.

src/spoke/interfaces/ISpoke.sol
```diff
@@ -85,12 +85,22 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
...
     mapping(address user => bool) approval;
     bool active;
+    bool isGlobal;
```
miguelmtzinf, line 95:
```suggestion
    bool global;
```
just because we use `active` not `isActive`.

src/spoke/interfaces/ISpoke.sol
miguelmtzinf: yeah, in favor of 2 flags so we save an struct

miguelmtzinf:
Agreed not to move forward with this, since it’s not clear how it would be used and users can easily sign up for new position managers via intents with array support (#1066).

PR #1099, opened 2026-01-08
feat: Debt and collateral limit

CheyenneAtapour:
Impose two new limits:
1. Limit the number of collaterals a user can have
2. Limit the number of debt reserves a user can have

Replaces the old `LiquidationConfig` struct with `SpokeConfig` struct which also includes these new reserve limits.

Closes https://github.com/aave/aave-v4/issues/1100

Relies on #1132 getting merged in first


src/spoke/interfaces/ISpoke.sol
```diff
@@ -376,6 +382,8 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
 
   /// @notice Allows suppliers to enable/disable a specific supplied reserve as collateral.
   /// @dev It reverts if the reserve associated with the given reserve identifier is not listed.
+  /// @dev It reverts if the user exceeds the maximum allowed collateral reserves when enabling.
```
miguelmtzinf, line 401: should we add a similar comment in `borrow`?

src/spoke/Spoke.sol
```diff
@@ -68,6 +68,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   /// @dev The number of decimals used by the oracle.
   uint8 internal constant ORACLE_DECIMALS = 8;
 
+  /// @dev The maximum allowed number of collateral reserves per user.
+  uint256 internal constant MAX_ALLOWED_COLLATERAL_RESERVES = 128;
+
+  /// @dev The maximum allowed number of borrowed reserves per user.
+  uint256 internal constant MAX_ALLOWED_BORROW_RESERVES = 128;
```
miguelmtzinf: we need these as the max value each limit can have, but we said we should make the limits themselves configurable

src/spoke/Spoke.sol
```diff
@@ -916,21 +922,24 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     require(!flags.paused(), ReservePaused());
   }
 
-  function _validateBorrow(ReserveFlags flags) internal pure {
+  function _validateBorrow(ReserveFlags flags, PositionStatus storage positionStatus) internal view {
     require(!flags.paused(), ReservePaused());
     require(!flags.frozen(), ReserveFrozen());
     require(flags.borrowable(), ReserveNotBorrowable());
+    require(positionStatus.borrowedCount(_reserveCount) < MAX_ALLOWED_BORROW_RESERVES, MaximumBorrowedReservesExceeded());
```
miguelmtzinf: it can be this is not a new borrowed reserve

src/spoke/Spoke.sol
```diff
@@ -916,21 +922,24 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
     require(!flags.paused(), ReservePaused());
   }
 
-  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral) internal pure {
+  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral, PositionStatus storage positionStatus) internal view {
     require(!flags.paused(), ReservePaused());
     // can disable as collateral if the reserve is frozen
     require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    // max collateral reserves check, account for the new reserve being enabled by using strict inequality
+    require(!usingAsCollateral || positionStatus.collateralCount(_reserveCount) < MAX_ALLOWED_COLLATERAL_RESERVES, MaximumCollateralReservesExceeded());
```
miguelmtzinf: does this correctly check we are going from not activated to activated? only checked when new collateral is added

CheyenneAtapour: Both statements in a require like this get executed, so this is not short-circuited, if that's what you mean. 

But indeed the logic is correct. 
if you are setting as collateral, the left side is false, so the right side must succeed

src/spoke/Spoke.sol
```diff
@@ -43,6 +43,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     // keccak256('SetUserPositionManager(address positionManager,address user,bool approve,uint256 nonce,uint256 deadline)')
     0x758d23a3c07218b7ea0b4f7f63903c4e9d5cbde72d3bcfe3e9896639025a0214;
 
+  /// @inheritdoc ISpoke
+  uint256 public constant MAX_ALLOWED_COLLATERAL_RESERVES = 128;
+
+  /// @inheritdoc ISpoke
+  uint256 public constant MAX_ALLOWED_BORROWED_RESERVES = 128;
```
CheyenneAtapour: My initial thought here was that we should evaluate the allowed number of reserves for an implementation of a spoke similar to how it's done here: https://github.com/aave/aave-v4/pull/930 and then we can hardcode this value. Making it configurable leads to the issue of misconfiguration, and there should be no reason to change this given an implementation. 

We can discuss if we want it to change upon upgrading, which is fine by me

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: Done

src/spoke/Spoke.sol
CheyenneAtapour: what is the motivation for setting the limit lower than the max value? Just to avoid situations where liquidating positions potentially cost a higher amount of gas (due to the user activating all possible collaterals)?

src/spoke/Spoke.sol
CheyenneAtapour: Thanks, good catch. Updated

src/spoke/Spoke.sol
DhairyaSethi: >  should be no reason to change this given an implementation

idts, better to have configurable bc we might need
- different limits based on chain 
- start out at a smaller limit as a bd strategy and expand later

not in favour of a partial compound approach where params are upgraded with impl

src/spoke/interfaces/ISpoke.sol
```diff
@@ -534,4 +542,10 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
 
   /// @notice Returns the address of the AaveOracle contract.
   function ORACLE() external view returns (address);
+
+/// @notice The maximum allowed number of collateral reserves per user.
+  function MAX_ALLOWED_COLLATERAL_RESERVES() external view returns (uint256);
+  
+  /// @notice The maximum allowed number of borrowed reserves per user.
+  function MAX_ALLOWED_BORROWED_RESERVES() external view returns (uint256);
```
DhairyaSethi: i would rename these to active collateral and borrow limit

current name kinda suggests this is the max reserves supported on the spoke (see max allowed collateral risk for ex)

src/spoke/Spoke.sol
```diff
@@ -927,10 +934,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     require(!flags.paused(), ReservePaused());
   }
 
-  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral) internal pure {
+  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral, PositionStatus storage positionStatus) internal view {
     require(!flags.paused(), ReservePaused());
     // can disable as collateral if the reserve is frozen
     require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    // max collateral reserves check, account for the new reserve being enabled by using strict inequality
+    require(!usingAsCollateral || positionStatus.collateralCount(_reserveCount) < MAX_ALLOWED_COLLATERAL_RESERVES, MaximumCollateralReservesExceeded());
```
DhairyaSethi: do something like 
```
if (usingAsCollateral) {
  require(!frozen)
  // comment about how this must be a new collateral bc of the no-op check
  require(activeCollateralCount within limit) 
}
```

src/spoke/libraries/PositionStatusMap.sol
```diff
@@ -98,6 +98,23 @@ library PositionStatusMap {
     }
   }
 
+  /// @notice Counts the number of reserves borrowed.
+  /// @dev Disregards potential dirty bits set after `reserveCount`.
+  /// @param reserveCount The current `reserveCount`, to avoid reading uninitialized buckets.
+  function borrowedCount(
```
DhairyaSethi:
```suggestion
  function borrowCount(
```

tests/Base.t.sol
```diff
@@ -2875,4 +2875,81 @@ abstract contract Base is Test {
...
+        IAssetInterestRateStrategy.InterestRateData({
+          optimalUsageRatio: 90_00, // 90.00%
+          baseVariableBorrowRate: 5_00, // 5.00%
+          variableRateSlope1: 5_00, // 5.00%
+          variableRateSlope2: 5_00 // 5.00%
+        })
+      );
+
+      // Add asset to hub1
+      vm.startPrank(ADMIN);
```
DhairyaSethi, line 2936: use configurator? also default values helper

src/spoke/interfaces/ISpoke.sol
```diff
@@ -312,6 +312,12 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @notice Thrown when the maximum number of dynamic config keys is reached.
   error MaximumDynamicConfigKeyReached();
 
+  /// @notice Thrown when user attempts to exceed the maximum allowed collateral reserves.
+  error MaximumCollateralReservesExceeded();
```
DhairyaSethi: ActiveCollLimit (optionally +Reached)

src/spoke/interfaces/ISpoke.sol
```diff
@@ -376,6 +382,8 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
 
   /// @notice Allows suppliers to enable/disable a specific supplied reserve as collateral.
   /// @dev It reverts if the reserve associated with the given reserve identifier is not listed.
+  /// @dev It reverts if the user exceeds the maximum allowed collateral reserves when enabling.
+  /// @dev Reserves with zero supplied and zero collateral factor still count towards the collateral limit.
```
DhairyaSethi: 🙏

src/spoke/Spoke.sol
yan-man: hm perhaps we could have 2 values here then. 
1) an immutable max allowed reserves limit, set in constructor. This one shouldnt ever be exceeded as it could brick actions
2) governance controlled one which is <= max allowed, > 0 to allow flexibility

src/spoke/Spoke.sol
```diff
@@ -927,10 +934,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     require(!flags.paused(), ReservePaused());
   }
 
-  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral) internal pure {
+  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral, PositionStatus storage positionStatus) internal view {
     require(!flags.paused(), ReservePaused());
     // can disable as collateral if the reserve is frozen
     require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    // max collateral reserves check, account for the new reserve being enabled by using strict inequality
```
yan-man: you have this comment here but not in the borrow case, could have in both places or neither

src/spoke/interfaces/ISpoke.sol
yan-man: here i think is the only place where `collateral limit` is mentioned, so maybe reference ActiveCollateralLimit like Dhairya mentioned or here can be something like `count towards the allowed collateral reserves`

tests/unit/libraries/PositionStatusMap.t.sol
```diff
@@ -208,6 +208,73 @@ contract PositionStatusMapTest is Base {
...
+    assertEq(p.borrowedCount(200), 1);
+    p.setBorrowing(133, true);
+    assertEq(p.borrowedCount(200), 2);
+
+    p.setBorrowing(383, true);
+    assertEq(p.borrowedCount(300), 3);
+    p.setBorrowing(283, true);
+    assertEq(p.borrowedCount(300), 4);
+
+    p.setBorrowing(511, true);
```
yan-man, line 259: nit: here and for coll count test, can maybe vm randomUint for an intermediate value

tests/unit/libraries/PositionStatusMap.t.sol
```diff
@@ -208,6 +208,73 @@ contract PositionStatusMapTest is Base {
...
+
+  function test_borrowedCount(uint256 reserveCount) public {
+    reserveCount = bound(reserveCount, 0, 1 << 10); // gas limit
+    vm.setArbitraryStorage(address(p));
+
+    uint256 borrowedCount;
+    for (uint256 reserveId; reserveId < reserveCount; ++reserveId) {
+      if (p.isBorrowing(reserveId)) ++borrowedCount;
+      // reserveId is 0-base indexed, assert running borrowedCount is maintained correctly
+      assertEq(p.borrowedCount({reserveCount: reserveId + 1}), borrowedCount);
```
yan-man: nit: dont think this one needs named args

src/spoke/Spoke.sol
yan-man: good pt, i think this would be easier to reason about the logic

src/spoke/Spoke.sol
```diff
@@ -927,10 +934,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     require(!flags.paused(), ReservePaused());
   }
 
-  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral) internal pure {
+  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral, PositionStatus storage positionStatus) internal view {
```
yan-man: nit: minor but I'd prefer having storage args come first in the function

src/spoke/Spoke.sol
miguelmtzinf: don't think these limits are interesting from bd perspective, but yeah, better to have it configurable

src/spoke/Spoke.sol
miguelmtzinf: i was thinking in the case where it's already collateral, but it just returns later.

src/spoke/Spoke.sol
CheyenneAtapour: ok, done now

src/spoke/Spoke.sol
CheyenneAtapour: Added the configurability

src/spoke/Spoke.sol

src/spoke/Spoke.sol
CheyenneAtapour: I think this is what you mean: https://github.com/aave/aave-v4/pull/1099#discussion_r2672368774
It's done

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: borrow limit kinda sounds like it's a draw cap to me.

I'm not convinced that way about the names. Collateral reserves to me imply that it is a user thing. Max reserves would be more so something that means it's what's supportable by the spoke

DhairyaSethi: the naming is very similar to max collateral risk which is a spoke level function. i think currently it's very confusing and intuitively to me this looks like the max supported reserves on the spoke; i would suggest renaming

src/spoke/Spoke.sol
```diff
@@ -43,6 +43,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     // keccak256('SetUserPositionManager(address positionManager,address user,bool approve,uint256 nonce,uint256 deadline)')
     0x758d23a3c07218b7ea0b4f7f63903c4e9d5cbde72d3bcfe3e9896639025a0214;
 
+  /// @inheritdoc ISpoke
+  uint256 public immutable MAX_ALLOWED_COLLATERAL_RESERVES;
+
+  /// @inheritdoc ISpoke
+  uint256 public immutable MAX_ALLOWED_BORROWED_RESERVES;
```
DhairyaSethi: not needed imo - and confusing w the dynamic limit below

src/spoke/Spoke.sol
```diff
@@ -71,6 +77,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   /// @dev Number of reserves listed in the Spoke.
   uint256 internal _reserveCount;
 
+  /// @dev The governance-configurable maximum number of collateral reserves per user.
+  uint256 internal _collateralReservesLimit;
+
+  /// @dev The governance-configurable maximum number of borrowed reserves per user.
+  uint256 internal _borrowedReservesLimit;
```
DhairyaSethi: these do not need to be full unit's

DhairyaSethi: we can pack these three w reserve count

src/spoke/Spoke.sol
```diff
@@ -101,9 +113,20 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+    address oracle_,
+    uint256 maxAllowedCollateralReserves_,
+    uint256 maxAllowedBorrowedReserves_
+  ) {
+    require(maxAllowedCollateralReserves_ > 0 && maxAllowedBorrowedReserves_ > 0, InvalidReservesLimit());
     require(IAaveOracle(oracle_).DECIMALS() == ORACLE_DECIMALS, InvalidOracleDecimals());
+    MAX_ALLOWED_COLLATERAL_RESERVES = maxAllowedCollateralReserves_;
+    MAX_ALLOWED_BORROWED_RESERVES = maxAllowedBorrowedReserves_;
     ORACLE = oracle_;
+    _setReservesLimits({
```
DhairyaSethi: setUserLimits?

src/spoke/Spoke.sol
```diff
@@ -290,6 +324,7 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     uint256 drawnShares = hub.draw(reserve.assetId, amount, msg.sender);
     userPosition.drawnShares += drawnShares.toUint120();
     if (!positionStatus.isBorrowing(reserveId)) {
+      require(positionStatus.borrowedCount(_reserveCount) < _borrowedReservesLimit, MaximumBorrowedReservesExceeded());
```
DhairyaSethi: maximumborrowablereservesexceeded?

src/spoke/Spoke.sol
```diff
@@ -405,12 +440,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     bool usingAsCollateral,
     address onBehalfOf
   ) external onlyPositionManager(onBehalfOf) {
-    _validateSetUsingAsCollateral(_getReserve(reserveId).flags, usingAsCollateral);
     PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
-
     if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) {
       return;
     }
+    _validateSetUsingAsCollateral(_getReserve(reserveId).flags, usingAsCollateral, positionStatus);
```
```suggestion
    _validateSetUsingAsCollateral(positionStatus, _getReserve(reserveId).flags, usingAsCollatera);
```

src/spoke/Spoke.sol
```diff
@@ -684,6 +718,16 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
+    /// @inheritdoc ISpoke
+  function getCollateralReservesLimit() external view returns (uint256) {
+    return _collateralReservesLimit;
+  }
+
+  /// @inheritdoc ISpoke
+  function getBorrowedReservesLimit() external view returns (uint256) {
+    return _borrowedReservesLimit;
+  }
```
DhairyaSethi: use one function to return both 

pls be lil frugal w codesize lol

src/spoke/Spoke.sol
```diff
@@ -927,10 +971,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
   }
 
-  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral) internal pure {
+  function _validateSetUsingAsCollateral(ReserveFlags flags, bool usingAsCollateral, PositionStatus storage positionStatus) internal view {
     require(!flags.paused(), ReservePaused());
-    // can disable as collateral if the reserve is frozen
-    require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    if (usingAsCollateral) {
+      require(!flags.frozen(), ReserveFrozen()); // disabling as collateral is allowed when reserve is frozen
+      // This must be a new collateral due to short-circuiting otherwise. Strict inequality because collateral count incremented later.
```
DhairyaSethi: first sentence incomplete 
i would say no-op instead of short circuit 


also, can make this check <= also it's totally fine to revert on the next attempt - hence would remove the second comment and fix the inequality

src/spoke/Spoke.sol
```diff
@@ -957,6 +1004,22 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
+  function _setReservesLimits(
+    uint256 collateralLimit,
+    uint256 borrowedLimit
+  ) internal {
+    require(
+      collateralLimit > 0 &&
+        collateralLimit <= MAX_ALLOWED_COLLATERAL_RESERVES &&
+        borrowedLimit > 0 &&
+        borrowedLimit <= MAX_ALLOWED_BORROWED_RESERVES,
```
DhairyaSethi: why can't the limits be zero?

src/spoke/interfaces/ISpoke.sol
```diff
@@ -312,10 +320,24 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @notice Thrown when the maximum number of dynamic config keys is reached.
   error MaximumDynamicConfigKeyReached();
 
+  /// @notice Thrown when user attempts to exceed the maximum allowed collateral reserves.
+  error MaximumCollateralReservesExceeded();
+
+  /// @notice Thrown when user attempts to exceed the maximum allowed borrowed reserves.
+  error MaximumBorrowedReservesExceeded();
```
DhairyaSethi: we can use the same error for both, saves a bit of code size, and isn't too confusing bc they are never potentially to be seen in the same txn unless batched

DhairyaSethi:
you can skip the code size check in the ci (i think continue-on-error: true, see on v3-origin), and tag this PR as blocked until code size is resolved such that ci and tests are run

src/spoke/Spoke.sol
yan-man: setUserReservesLimits?

src/spoke/Spoke.sol
yan-man: > first sentence incomplete

i think this is fine as it's inline comment

src/spoke/Spoke.sol
```diff
@@ -121,6 +144,17 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+  /// @inheritdoc ISpoke
+  function updateUserReservesLimits(
+    uint256 collateralReservesLimit_,
+    uint256 borrowedReservesLimit_
+  ) external restricted {
+    _setReservesLimits({
+      collateralLimit: collateralReservesLimit_,
+      borrowedLimit: borrowedReservesLimit_
+    });
+  }
```
yan-man, line 133: i think need to add to spoke configurator as well then

src/spoke/Spoke.sol
```diff
@@ -121,6 +144,17 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     emit UpdateLiquidationConfig(config);
   }
 
+  /// @inheritdoc ISpoke
+  function updateUserReservesLimits(
+    uint256 collateralReservesLimit_,
+    uint256 borrowedReservesLimit_
+  ) external restricted {
+    _setReservesLimits({
+      collateralLimit: collateralReservesLimit_,
```
yan-man: a bit of inconsistency here in naming, they should all be "reservesLimit" or just "limit". I'd prefer the former even though it's longer, so `collateralReservesLimit` and `borrowedReservesLimit` everywhere

src/spoke/Spoke.sol
yan-man: makes sense for debt limit to be 0 (for a supply-only spoke), but coll reserve cant be 0 unless we have a segregated collateral feature right (ie coll not on this spoke), otherwise what coll to borrow against?

DhairyaSethi: temporary measure

src/spoke/libraries/PositionStatusMap.sol
CheyenneAtapour: I also think this naming is fine as is, but let's perhaps fix logic now and then we can revisit this, and get a majority vote from the team on names

tests/Base.t.sol
CheyenneAtapour: I think it's easier as is without the hub configurator. I'm also noticing now that we don't have a generic `updateAssetConfig` on the hub configurator

tests/unit/libraries/PositionStatusMap.t.sol
CheyenneAtapour: I did it like this to stay consistent with the similar collateral count tests

tests/unit/libraries/PositionStatusMap.t.sol
CheyenneAtapour: Really? I think it's not immediately obvious that the function takes a `reserveCount` and that the reserve count is `reserveId + 1`

src/spoke/Spoke.sol

src/spoke/Spoke.sol
CheyenneAtapour: Removed

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: I'm going to leave variable names untouched for now, until we decide fully as a team, but I updated the comment for now. I thought the `limit` was clear here from context to refer to the comment from above, but I changed it to align more with the wording and avoid `limit` for now

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: I can get behind this perhaps. Let's circle back to namings

src/spoke/Spoke.sol
CheyenneAtapour: Seems it was decided to do it this way: https://github.com/aave/aave-v4/pull/1099#discussion_r2674142675

Or do you mean just to not expose these ones?

DhairyaSethi: not configurability - i meant upper bounds

src/spoke/Spoke.sol
CheyenneAtapour: I think fine as is, can revisit names

src/spoke/Spoke.sol
CheyenneAtapour: Done, @yan-man also pointed this out above

src/spoke/Spoke.sol
CheyenneAtapour: Yeah, absolutely, sounds good! Done

src/spoke/Spoke.sol
CheyenneAtapour: That is a complete sentence haha. Also, I don't like <= because then I have to add a +1. I just removed the comment

src/spoke/Spoke.sol
CheyenneAtapour: That was from a discussion above. Updated to remove the 0 check

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: That makes sense, but what name would you suggest for this? `MaximumUserReservesExceeded`? Seems like an unhelpful name

tests/unit/libraries/PositionStatusMap.t.sol
Kogaroshi: Same impl than the same test for collateral count, so imo okay to keep as is. And if modified, please modify both tests.

src/spoke/Spoke.sol
CheyenneAtapour: Added

src/spoke/Spoke.sol
CheyenneAtapour: Done, changed to `setUserReservesLimits`

src/spoke/Spoke.sol
DhairyaSethi: i dont think absolute limits are needed here tbh, not seeing how they help bc theoretical max is anyway reserve count w infinite gas per block

src/spoke/SpokeConfigurator.sol
```diff
@@ -295,6 +295,15 @@ contract SpokeConfigurator is Ownable2Step, ISpokeConfigurator {
...
   }
 
+  /// @inheritdoc ISpokeConfigurator
+  function updateUserReservesLimits(
+    address spoke,
+    uint8 collateralReservesLimit,
+    uint8 borrowedReservesLimit
+  ) external onlyOwner {
+    ISpoke(spoke).updateUserReservesLimits(collateralReservesLimit, borrowedReservesLimit);
+  }
```
DhairyaSethi, line 312: configurator should not have generic updates: we should have two methods here to update each param individually which leaves the other untouched

src/spoke/Spoke.sol
```diff
@@ -71,6 +79,12 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   /// @dev Number of reserves listed in the Spoke.
   uint256 internal _reserveCount;
 
+  /// @dev The governance-configurable maximum number of collateral reserves per user.
+  uint8 internal _collateralReservesLimit;
+
+  /// @dev The governance-configurable maximum number of borrowed reserves per user.
+  uint8 internal _borrowedReservesLimit;
```
DhairyaSethi: perhaps it makes sense to club this with other spoke level configs like `LiquidationConfig` in a broader struct

src/spoke/Spoke.sol
miguelmtzinf: only reason I think is to make sure it fits within the dataype, but we can just safecast or use the final data type directly as the function input

src/spoke/Spoke.sol
miguelmtzinf: i thought the same, but hard given that they are not very related... open to ideas

src/spoke/Spoke.sol
```diff
@@ -927,10 +974,20 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+  function _validateSetUsingAsCollateral(
+    PositionStatus storage positionStatus,
+    ReserveFlags flags,
+    bool usingAsCollateral
+  ) internal view {
     require(!flags.paused(), ReservePaused());
-    // can disable as collateral if the reserve is frozen
-    require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    if (usingAsCollateral) {
+      require(!flags.frozen(), ReserveFrozen()); // disabling as collateral is allowed when reserve is frozen
```
```suggestion
      // disabling as collateral is allowed when reserve is frozen
      require(!flags.frozen(), ReserveFrozen());
```

src/spoke/SpokeConfigurator.sol
CheyenneAtapour: Done

src/spoke/Spoke.sol

src/spoke/Spoke.sol

src/spoke/Spoke.sol
CheyenneAtapour: The upper bounds were decided there in that comment as well

src/spoke/Spoke.sol
CheyenneAtapour: I think these absolute limits are helpful to ensure governance doesn't accidentally misconfigure a spoke and allow for a position to be created which cannot be liquidated due to OOG

DhairyaSethi: yeah like others safe cast is enough

> position to be created which cannot be liquidated due to OOG

gas numbers keep changing w upgrades; doesn't make sense to hardcode a limit; governance should be allowed to do this

src/spoke/Spoke.sol
DhairyaSethi: struct will be renamed to `SpokeConfig`, liq config inlined and these

src/spoke/interfaces/ISpoke.sol
DhairyaSethi: yeah thats good; it's obvious to figure out if this is borrowed or collateral depending on the action you performed

src/spoke/Spoke.sol
DhairyaSethi: i would remove the max bound checks also

src/spoke/libraries/PositionStatusMap.sol
DhairyaSethi: it's more consistent w `collateralCount`

tests/Base.t.sol
DhairyaSethi: we dont need generic on configurator but rather only targetted

tests/unit/libraries/PositionStatusMap.t.sol
DhairyaSethi: yeah i would avoid named args here

src/spoke/Spoke.sol
yan-man: > start out at a smaller limit as a bd strategy and expand later

do you mean then @DhairyaSethi to forgo this smaller BD limit, or just always rely on gov to set the proper OOG limit?

src/spoke/libraries/PositionStatusMap.sol
yan-man: kind of agnostic here, but consistency wise collateral is a noun whereas borrow is a verb. I think perhaps supplyCount is analogous to borrowCount. I'd prefer collateralCount over supplyCount though

src/spoke/Spoke.sol
DhairyaSethi: gov sets correct values, no immutable upper bounds req 


these upper bounds are anyway not constant bc gas pricing can change w EIPs, gas used can change w protocol upgrades etc; so i would avoid completely

tests/unit/libraries/PositionStatusMap.t.sol
yan-man: yeah think we can do it for both sets of tests for a bit more completeness but nbd

tests/unit/libraries/PositionStatusMap.t.sol
yan-man: good pt @CheyenneAtapour agreed it's not readily clear from what the func should be like, but was more just thinking it's clear based on the func sig. (whereas for other places we had the named args there might be lack of clarity around the order of 2 diff input values)

src/spoke/Spoke.sol
CheyenneAtapour: Yes, aligned from sync today. Will remove upper bounds, keeping only the 2 configurable variables

src/spoke/Spoke.sol
```diff
@@ -927,10 +970,21 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+    ReserveFlags flags,
+    bool usingAsCollateral
+  ) internal view {
     require(!flags.paused(), ReservePaused());
-    // can disable as collateral if the reserve is frozen
-    require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    if (usingAsCollateral) {
+      // disabling as collateral is allowed when reserve is frozen
+      require(!flags.frozen(), ReserveFrozen());
+      // this must be a new collateral due to short-circuiting otherwise.
```
yan-man:
```suggestion
      // this must be a new collateral, otherwise will short-circuit
```

?

.github/workflows/tests-pr.yml
```diff
@@ -23,6 +23,7 @@ jobs:
 
       - name: Run Forge size
         uses: bgd-labs/github-workflows/.github/actions/foundry-size@main
+        continue-on-error: true
```
yan-man: todo: rm before merging after via ir

src/spoke/interfaces/ISpoke.sol
```diff
@@ -127,6 +127,11 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
   /// @param config The new liquidation config.
   event UpdateLiquidationConfig(LiquidationConfig config);
 
+  /// @notice Emitted when user reserve limits are updated.
+  /// @param maxUserCollaterals The new max number of collateral reserves per user.
+  /// @param maxUserBorrows The new max number of borrowed reserves per user.
+  event UpdateUserReservesLimits(uint64 maxUserCollaterals, uint64 maxUserBorrows);
```
yan-man:
```suggestion
  event UpdateUserReserveLimits(uint64 maxUserCollaterals, uint64 maxUserBorrows);
```

wdyt?

src/spoke/interfaces/ISpoke.sol
```diff
@@ -444,7 +465,7 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
 
   /// @notice Returns the number of listed reserves on the spoke.
   /// @dev Count includes reserves that are not currently active.
-  function getReserveCount() external view returns (uint256);
+  function getReserveCount() external view returns (uint64);
```
yan-man: jw how you arrived at 64

src/spoke/interfaces/ISpoke.sol
```diff
@@ -534,4 +555,17 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
 
   /// @notice Returns the address of the AaveOracle contract.
   function ORACLE() external view returns (address);
+
+  /// @notice The absolute limit for the maximum allowed number of collateral reserves per user.
+  /// @dev Governance can set the limit on max allowed user collateral reserves, but it cannot exceed this constant.
```
yan-man:
```suggestion
  /// @dev Governance can set the limit on max allowed user collateral reserves, but it cannot exceed this value.
```

nit: not a constant

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: We wanted to pack this and the new limits together in a slot. Could have done 80, but decided that is too much

src/spoke/libraries/PositionStatusMap.sol
CheyenneAtapour: I agree `collateralCount` is fine, it's just a question of `borrowedCount` vs `borrowCount` here. To me this makes sense because it's the "number of borrowed reserves"

src/spoke/Spoke.sol
CheyenneAtapour: it should have happened, so perhaps "otherwise would have short-circuited"
I changed it to this, lemme know

tests/Base.t.sol
CheyenneAtapour: ok, nice, I removed the generic. But anyway, I think this is simpler code, and more similar to how we do this above (without configurator)

src/spoke/Spoke.sol

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: Removed these, and updated the rest of the names, lemme know wdyt

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: Renamed, lemme know your thoughts

tests/unit/libraries/PositionStatusMap.t.sol
CheyenneAtapour: I removed the named args from both tests as @Kogaroshi suggested

src/spoke/Spoke.sol
CheyenneAtapour: Removed upper bounds

src/spoke/Spoke.sol
CheyenneAtapour: Changed again to setUserReserveLimits btw

src/spoke/Spoke.sol
CheyenneAtapour: Name changed, lemme know your thoughts

src/spoke/Spoke.sol
CheyenneAtapour: Names been changed, resolving this one as it's commented in another thread

src/spoke/Spoke.sol

src/spoke/Spoke.sol
CheyenneAtapour: I changed to maxUserCollaterals and maxUserBorrows, lemme know your thoughts

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: Yes, changed everywhere as well as other functions / variable names to be consistent with this

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: This limit has been removed

src/spoke/Spoke.sol
```diff
@@ -69,7 +71,13 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
 
   /// @dev Number of reserves listed in the Spoke.
-  uint256 internal _reserveCount;
+  uint64 internal _reserveCount;
+
+  /// @dev The governance-configurable maximum number of collateral reserves per user.
+  uint64 internal _maxUserCollaterals;
+
+  /// @dev The governance-configurable maximum number of borrowed reserves per user.
+  uint64 internal _maxUserBorrows;
```
DhairyaSethi: lets not touch `_reserveCount` and move the two `_max`s to `LiquidationConfig` under a renamed struct, something like `SpokeConfig`

src/spoke/Spoke.sol
```diff
@@ -107,7 +115,11 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   }
 
   /// @dev To be overridden by the inheriting Spoke instance contract.
-  function initialize(address authority) external virtual;
+  function initialize(
+    address authority,
+    uint64 maxUserCollaterals,
+    uint64 maxUserBorrows
```
DhairyaSethi: lets have this init w spokeConfig

src/spoke/Spoke.sol
```diff
@@ -927,10 +961,21 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
...
+    ReserveFlags flags,
+    bool usingAsCollateral
+  ) internal view {
     require(!flags.paused(), ReservePaused());
-    // can disable as collateral if the reserve is frozen
-    require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    if (usingAsCollateral) {
+      // disabling as collateral is allowed when reserve is frozen
+      require(!flags.frozen(), ReserveFrozen());
+      // this must be a new collateral, otherwise would have short-circuited.
```
```suggestion
      // this must be a new collateral, otherwise would have short-circuited
```

tests/Base.t.sol
DhairyaSethi: above needs a refactor, the whole point of configurator is help w this

CheyenneAtapour: I'm not sure I agree with both of these statements being true. If the whole point of configurator was to help make these setups cleaner, then we would have generic functions on configurator, no?

tests/unit/libraries/PositionStatusMap.t.sol
CheyenneAtapour: I think it would need to be done carefully to ensure we don't re-use the random values, so let's just save for another Pr which refactors tests

src/spoke/Spoke.sol

src/spoke/libraries/PositionStatusMap.sol
Kogaroshi: rest of the lib uses `borrowing` keyword, imo good compromise is `borrowingCount`, as it still is "number of borrowed reserves" at the current moment.

src/spoke/Spoke.sol
Kogaroshi: `LiquidationConfig` only has uint48 left available, so 2x `uint24` for the `_max` values ?

src/spoke/interfaces/ISpoke.sol
DhairyaSethi: cc @CheyenneAtapour

src/spoke/Spoke.sol
DhairyaSethi: yeah - you're also going to be adding `gracePeriod` there now so its fine we overflow 
in any case yep uint24 seems enough

DhairyaSethi: we can also see how to pack liq config more efficiently, i think target hf doesnt need to be 128
you need to 40 for grace period

DhairyaSethi: spoke config
    uint72 targetHealthFactor; (max is 100 WAD)
    uint64 healthFactorForMaxBonus;
    uint16 liquidationBonusFactor;
    uint40 gracePeriod; 

64 bits left

src/spoke/instances/SpokeInstance.sol
```diff
@@ -21,9 +21,16 @@ contract SpokeInstance is Spoke {
   /// @notice Initializer.
   /// @dev The authority contract must implement the `AccessManaged` interface for access control.
   /// @param authority The address of the authority contract which manages permissions.
-  function initialize(address authority) external override reinitializer(SPOKE_REVISION) {
+  /// @param maxUserCollaterals The maximum allowed number of collateral reserves per user.
+  /// @param maxUserBorrows The maximum allowed number of borrowed reserves per user.
+  function initialize(
+    address authority,
+    uint64 maxUserCollaterals,
+    uint64 maxUserBorrows
```
DhairyaSethi: very ugly we need init like this; i would take SpokeConfig struct as input OR set these two as max sentinel value and skip the check when max

src/spoke/Spoke.sol
```diff
@@ -290,6 +310,10 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     uint256 drawnShares = hub.draw(reserve.assetId, amount, msg.sender);
     userPosition.drawnShares += drawnShares.toUint120();
     if (!positionStatus.isBorrowing(reserveId)) {
+      require(
+        positionStatus.borrowedCount(_reserveCount) < _maxUserBorrows,
```
DhairyaSethi: maxUserBorrows != max || maxUserBorrows < borrowCount

same for coll count check

src/spoke/interfaces/ISpoke.sol

src/spoke/Spoke.sol

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: Reverted this to uint256

src/spoke/Spoke.sol

src/spoke/Spoke.sol
CheyenneAtapour: maxUserBorrows == max I think you mean, but yes will do

src/spoke/interfaces/ISpoke.sol
```diff
@@ -528,4 +537,9 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IMulticall {
 
   /// @notice Returns the address of the AaveOracle contract.
   function ORACLE() external view returns (address);
+
+  /// @notice Returns the governance-controlled maximum allowed number of collateral and borrowed reserves per user.
+  /// @return The maximum allowed number of collateral reserves per user.
+  /// @return The maximum allowed number of borrowed reserves per user.
+  function getUserReserveLimits() external view returns (uint24, uint24);
```
DhairyaSethi: we dont need this, this info is part of `getSpokeConfig`

CheyenneAtapour: Good call, done

src/spoke/libraries/PositionStatusMap.sol
CheyenneAtapour: I went with borrowCount now

src/spoke/Spoke.sol

src/spoke/instances/SpokeInstance.sol
CheyenneAtapour: These are set when the rest of SpokeConfig is set now

src/spoke/Spoke.sol

src/spoke/interfaces/ISpoke.sol
```diff
@@ -59,14 +59,18 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IMulticall {
...
-  struct LiquidationConfig {
-    uint128 targetHealthFactor;
+  /// @dev maxUserCollaterals The maximum number of reserves a user can enable as collateral.
+  /// @dev maxUserBorrows The maximum number of reserves a user can borrow.
+  struct SpokeConfig {
+    uint64 targetHealthFactor;
     uint64 healthFactorForMaxBonus;
     uint16 liquidationBonusFactor;
+    uint24 maxUserCollaterals;
+    uint24 maxUserBorrows;
```
Kogaroshi: note : size for another `uint64` left. If the `gracePeriod` logic is not merged in, we can consider reverting back `targetHealthFactor` to an `uint128`

miguelmtzinf: uint16 would be enough?

src/spoke/interfaces/ISpoke.sol
```diff
@@ -309,9 +313,12 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IMulticall {
...
   error MaximumDynamicConfigKeyReached();
 
-  /// @notice Updates the liquidation config.
-  /// @param config The liquidation config.
-  function updateLiquidationConfig(LiquidationConfig calldata config) external;
+  /// @notice Thrown when user attempts to exceed either the maximum allowed collateral or borrowed reserves.
+  error MaximumUserReservesExceeded();
+
+  /// @notice Updates the spoke config.
+  /// @param config The spoke config.
```
```suggestion
  /// @notice Updates the spoke configuration.
  /// @param config The new spoke config.
```

src/spoke/Spoke.sol
```diff
@@ -76,6 +69,9 @@ abstract contract Spoke is
   /// @dev Number of reserves listed in the Spoke.
   uint256 internal _reserveCount;
 
+  /// @dev The spoke configuration including liquidation params and user reserve limits.
```
```suggestion
  /// @dev The spoke configuration, including liquidation params and user reserve limits.
```

src/spoke/Spoke.sol
```diff
@@ -295,6 +288,12 @@ abstract contract Spoke is
     uint256 drawnShares = hub.draw(reserve.assetId, amount, msg.sender);
     userPosition.drawnShares += drawnShares.toUint120();
     if (!positionStatus.isBorrowing(reserveId)) {
+      uint24 maxUserBorrows = _spokeConfig.maxUserBorrows;
+      require(
+        maxUserBorrows == type(uint24).max ||
```
miguelmtzinf: let's create a constant for the uint24 max
`MAX_ALLOWED_USER_RESERVES_LIIMIT` ?

src/spoke/Spoke.sol
```diff
@@ -295,6 +288,12 @@ abstract contract Spoke is
     uint256 drawnShares = hub.draw(reserve.assetId, amount, msg.sender);
     userPosition.drawnShares += drawnShares.toUint120();
     if (!positionStatus.isBorrowing(reserveId)) {
+      uint24 maxUserBorrows = _spokeConfig.maxUserBorrows;
```
miguelmtzinf: This logic is placed in the validate fn for the collateral. I think better to put them both in the same place: either validate or status flag setter.

src/spoke/Spoke.sol
```diff
@@ -928,10 +926,23 @@ abstract contract Spoke is
...
+    ReserveFlags flags,
+    bool usingAsCollateral
+  ) internal view {
     require(!flags.paused(), ReservePaused());
-    // can disable as collateral if the reserve is frozen
-    require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    if (usingAsCollateral) {
+      // disabling as collateral is allowed when reserve is frozen
+      require(!flags.frozen(), ReserveFrozen());
+      // this must be a new collateral, otherwise would have short-circuited
```
miguelmtzinf: short-circuit happens afterwards indeed

src/spoke/instances/SpokeInstance.sol
```diff
@@ -25,9 +25,11 @@ contract SpokeInstance is Spoke {
     emit UpdateOracle(ORACLE);
     require(authority != address(0), InvalidAddress());
     __AccessManaged_init(authority);
-    if (_liquidationConfig.targetHealthFactor == 0) {
-      _liquidationConfig.targetHealthFactor = HEALTH_FACTOR_LIQUIDATION_THRESHOLD;
-      emit UpdateLiquidationConfig(_liquidationConfig);
+    if (_spokeConfig.targetHealthFactor == 0) {
+      _spokeConfig.targetHealthFactor = HEALTH_FACTOR_LIQUIDATION_THRESHOLD;
+      _spokeConfig.maxUserCollaterals = type(uint24).max;
+      _spokeConfig.maxUserBorrows = type(uint24).max;
```
miguelmtzinf: change to constant

snapshots/Spoke.Operations.json
```diff
@@ -1,33 +1,33 @@
...
-  "usingAsCollateral: 1 borrow, disable": "158199",
-  "usingAsCollateral: 1 borrow, enable": "42444",
-  "usingAsCollateral: 2 borrows, disable": "227867",
-  "usingAsCollateral: 2 borrows, enable": "42456",
-  "withdraw: 0 borrows, full": "127660",
-  "withdraw: 0 borrows, partial": "132346",
-  "withdraw: 1 borrow, partial": "209550",
-  "withdraw: 2 borrows, partial": "255263",
-  "withdraw: non collateral": "105615"
+  "borrow: first": "260012",
```
miguelmtzinf: it adds 2k because `SpokeConfig` is only read for liquidations, right

src/spoke/Spoke.sol
CheyenneAtapour: The short-circuit is the first thing that happens in `setUsingAsCollateral`, so it happens before, no?

src/spoke/Spoke.sol
CheyenneAtapour: It seems a little tricky with how the flow of these 2 functions currently is, but I will think on the best way to achieve this

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: updated

src/spoke/interfaces/ISpoke.sol

src/spoke/Spoke.sol

src/spoke/Spoke.sol

src/spoke/instances/SpokeInstance.sol
CheyenneAtapour: Done

snapshots/Spoke.Operations.json
CheyenneAtapour: That, and then we also invoke `PositionStatus.borrowCount()` which we hadn't before

src/spoke/Spoke.sol
CheyenneAtapour: I did this, but can we omit for some space? Space seems to be problematic here. We have another constant with the same value, and also we already do use a `type(uint256).max` in code

src/spoke/Spoke.sol
miguelmtzinf: ah, you moved it. ty

src/spoke/Spoke.sol
```diff
@@ -413,12 +422,11 @@ abstract contract Spoke is
     bool usingAsCollateral,
     address onBehalfOf
   ) external nonReentrant onlyPositionManager(onBehalfOf) {
-    _validateSetUsingAsCollateral(_getReserve(reserveId).flags, usingAsCollateral);
     PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
```
miguelmtzinf, line 425:
```suggestion
    Reserve storage reserve = _getReserve(reserveId);
    PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
```
so we check reserve existence

src/spoke/Spoke.sol
```diff
@@ -413,12 +422,11 @@ abstract contract Spoke is
...
-
     if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) {
       return;
     }
+    _validateSetUsingAsCollateral(positionStatus, _getReserve(reserveId).flags, usingAsCollateral);
```
```suggestion
    _validateSetUsingAsCollateral(positionStatus, reserve.flags, usingAsCollateral);
```

src/spoke/instances/SpokeInstance.sol
```diff
@@ -25,9 +25,11 @@ contract SpokeInstance is Spoke {
     emit UpdateOracle(ORACLE);
     require(authority != address(0), InvalidAddress());
     __AccessManaged_init(authority);
-    if (_liquidationConfig.targetHealthFactor == 0) {
-      _liquidationConfig.targetHealthFactor = HEALTH_FACTOR_LIQUIDATION_THRESHOLD;
-      emit UpdateLiquidationConfig(_liquidationConfig);
+    if (_spokeConfig.targetHealthFactor == 0) {
+      _spokeConfig.targetHealthFactor = HEALTH_FACTOR_LIQUIDATION_THRESHOLD;
+      _spokeConfig.maxUserCollaterals = MAX_ALLOWED_USER_RESERVES_LIMIT;
+      _spokeConfig.maxUserBorrows = MAX_ALLOWED_USER_RESERVES_LIMIT;
```
miguelmtzinf, line 31: i d prefer `maxUserCollateralReserves` and `maxUserBorrowedReserves` but i guess it's too long?

src/spoke/Spoke.sol
```diff
@@ -63,6 +63,9 @@ abstract contract Spoke is
   /// @dev The maximum allowed value for a dynamic configuration key (inclusive).
   uint256 internal constant MAX_ALLOWED_DYNAMIC_CONFIG_KEY = type(uint24).max;
 
+  /// @dev The maximum allowed value for a user reserves limit (inclusive).
```
```suggestion
  /// @dev The maximum allowed value for the maximum number of reserves a user can have (collateral or borrowed) (inclusive).
```

src/spoke/Spoke.sol
miguelmtzinf: where?

src/spoke/interfaces/ISpoke.sol
miguelmtzinf: uint8 may be enough too

CheyenneAtapour: I thought it's possible we might want to exceed that

src/spoke/Spoke.sol
CheyenneAtapour: Space is fine now, so perhaps nvm, but here: https://github.com/aave/aave-v4/blob/7b7db47d0e872d9af91a02acbcab5e13fe3d980e/src/spoke/Spoke.sol#L787

src/spoke/Spoke.sol
CheyenneAtapour: We already have a _getReserve() call above, so we already check reserve existence, right?

src/spoke/Spoke.sol
CheyenneAtapour: Why not just leave it as is? See above

src/spoke/instances/SpokeInstance.sol
CheyenneAtapour: I think too long, and these names are already clear imo

src/spoke/Spoke.sol

src/spoke/Spoke.sol
CheyenneAtapour: I put both outside the validate function

src/spoke/libraries/ValidationLogic.sol
```diff
@@ -0,0 +1,69 @@
...
+
+/// @title ValidationLogic library
+/// @author Aave Labs
+/// @notice Implements validation logic for Spoke operations.
+library ValidationLogic {
+  using PercentageMath for uint256;
+  using ReserveFlagsMap for ReserveFlags;
+
+  /// @notice Validates the supply operation.
+  function validateSupply(ReserveFlags flags) internal pure {
```
DhairyaSethi: how does this help if all internal methods only?

CheyenneAtapour: Reverted

miguelmtzinf:
Closing this in favor of #1155. While configurable limits could be useful to accommodate chain-level changes that affect the theoretical maximum number of reserves a user position can have, we’re proceeding with that approach instead.

PR #1101, opened 2026-01-08
feat: extsloads

DhairyaSethi:
Implement ExtSloads contract to allow inheriting source contract to expose it's state to external contracts.


snapshots/Hub.Operations.json
DhairyaSethi, line 1: gas changes only bc of updates in fn selector binary search routine

src/hub/Hub.sol
```diff
@@ -2,6 +2,7 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity 0.8.28;
 
+import {ExtSload} from 'src/utils/ExtSload.sol';
```
yan-man: nit: should we mv down after external deps to L10

src/utils/ExtSload.sol
```diff
@@ -0,0 +1,37 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {IExtSload} from 'src/interfaces/IExtSload.sol';
+
+/// @title ExtSload
+/// @author Aave Labs
+/// @notice This allows source contract to make it's state available to external contracts.
```
yan-man:
```suggestion
/// @notice This allows the source contract to make its state available to external contracts.
```

tests/unit/misc/ExtSload.t.sol
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  function test_extSload(bytes32) public {
+    vm.setArbitraryStorage(address(w));
+    bytes32 slot = bytes32(vm.randomUint());
+    assertEq(w.extSload(slot), vm.load(address(w), slot));
+  }
+
+  function test_extSloads(uint256 count) public {
+    count = bound(count, 0, 1024); // for performance
```
yan-man, line 22: for performance, as in bounding only to 1024 and not higher?

yan-man:
can create an associated issue

tests/unit/misc/ExtSload.t.sol
DhairyaSethi: this is the loop size

yan-man: yes but what does the comment `for performance` mean in this context

DhairyaSethi: fuzz test time performance

CheyenneAtapour (approved):

PR #1103, opened 2026-01-09
rft : replace liquidatable flag by grace period logic

Kogaroshi:
The `liquidatable` flag checked during during liquidation was only applied on the Collateral Reserve, and was opening some misconfiguration that could block needed liquidations.
The new Grace Period logic replaces that flag by a set period during which the liquidations are not allowed on user position (either blocking the collateral side of the debt side depending which Reserve is in grace period).
The Grace Period is triggered after a Reserve goes from `paused` to `unpaused`, or `frozen` to `unfrozen`.


Kogaroshi:
To close if #1106 is chosen instead

src/spoke/interfaces/ISpoke.sol
```diff
@@ -31,21 +32,21 @@ interface ISpoke is ISpokeBase, IMulticall, INoncesKeyed, IAccessManaged {
     uint24 dynamicConfigKey;
     uint24 collateralRisk;
     ReserveFlags flags;
+    //
+    uint40 gracePeriodEnd;
```
DhairyaSethi: we can pack w `underlying` in the first slot

Kogaroshi: Yup, that's the plan, was checking the gas impact of each version.

src/spoke/Spoke.sol
```diff
@@ -87,6 +87,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
   mapping(uint256 reserveId => mapping(uint24 dynamicConfigKey => DynamicReserveConfig))
     internal _dynamicConfig;
 
+  /// @dev Grace period duration for liquidations after a Reserve is unpaused/unfrozen.
+  uint40 internal _gracePeriod;
+
   /// @dev Liquidation configuration for the Spoke.
   LiquidationConfig internal _liquidationConfig;
```
DhairyaSethi: these two can be merged in one struct since both are global spoke level configurations

src/spoke/Spoke.sol
```diff
@@ -154,9 +157,9 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
         initPaused: config.paused,
         initFrozen: config.frozen,
         initBorrowable: config.borrowable,
-        initLiquidatable: config.liquidatable,
         initReceiveSharesEnabled: config.receiveSharesEnabled
-      })
+      }),
+      gracePeriodEnd: 0
```
DhairyaSethi: block.timestamp instead?

src/spoke/Spoke.sol
Kogaroshi: the global `_gracePeriod` will be removed, after more discussions decision is to go with parameter passed through `ReserveConfig` .

src/spoke/Spoke.sol
```diff
@@ -690,6 +698,15 @@ abstract contract Spoke is ISpoke, Multicall, NoncesKeyed, AccessManagedUpgradea
     emit UpdateReservePriceSource(reserveId, priceSource);
   }
 
+  function _updateReserveGracePeriod(
+    Reserve storage reserve,
+    ReserveConfig calldata config
+  ) internal {
+    if ((reserve.flags.paused() && !config.paused) || (reserve.flags.frozen() && !config.frozen)) {
+      reserve.gracePeriodEnd = uint40(block.timestamp) + _gracePeriod;
```
```suggestion
      reserve.gracePeriodEnd = block.timestamp.toUint40() + _gracePeriod;
```

src/spoke/interfaces/ISpoke.sol
DhairyaSethi: its annoying that it gets "mixed" w the static part of the reserve struct but its fine

src/spoke/Spoke.sol
Kogaroshi: imo we should only set when there is a grace period active, otherwise put 0 as default.

DhairyaSethi: it's more consistent w future state of this bc it will never be zero again after the first time it's set

src/spoke/Spoke.sol
DhairyaSethi: so on each update of paused/frozen gov needs to be pass grace period? 

it's less friction to set a default value applied always so gov doesn't need to add this factor, and if it needs custom values it can always sandwich the reserve config update w grace period update

Kogaroshi: was my thinking too, I prefer the global setting. Argument for the parameter was for granularity.
cc @miguelmtzinf

DhairyaSethi: granularity is not lost w global setting, it's just extra steps on the niche case vs on the normal case when not needing granularity

tests/unit/Spoke/Spoke.Getters.t.sol
```diff
@@ -316,4 +316,25 @@ contract SpokeGettersTest is SpokeBase {
...
+    elapsedTime = uint40(bound(elapsedTime, 0, 150 days));
+
+    uint256 reserveId = _daiReserveId(spoke);
+    ISpoke.ReserveConfig memory config = spoke.getReserveConfig(reserveId);
+    config.gracePeriod = gracePeriod;
+
+    vm.prank(SPOKE_ADMIN);
+    spoke.updateReserveConfig(reserveId, config);
+
+    vm.warp(block.timestamp + elapsedTime);
```
DhairyaSethi: avoid using block.timestamp in tests, rather `vm.getBlockTimestamp()` its fine here but sometimes if used multiple times w warps in between then solc optimiser can optimise it out to a constant 

```suggestion
    skip(elapsedTime);
```

tests/unit/Spoke/Spoke.Getters.t.sol
```diff
@@ -316,4 +316,25 @@ contract SpokeGettersTest is SpokeBase {
     assetPremiumDebtRay = hub1.getAssetPremiumRay(wethAssetId);
     assertEq(assetPremiumDebtRay, spokePremiumDebtRay + spoke1PremiumDebtRay);
   }
+
+  function test_getReserveConfig_gracePeriod(uint40 gracePeriod, uint40 elapsedTime) public {
+    gracePeriod = uint40(bound(gracePeriod, 0, 100 days));
```
DhairyaSethi: safe cast for sanity in tests also pls for consistency

Kogaroshi: 9ad2fc99b845a4a3ac2a2c28fc807f1662afde9d

tests/unit/Spoke/Spoke.Getters.t.sol
Kogaroshi: 9ad2fc99b845a4a3ac2a2c28fc807f1662afde9d

Kogaroshi:
Spoke codesize reached here, would need to reduce via-IR optimizer compiler runs to add this change.

miguelmtzinf:
Although the ability to block liquidations under certain conditions can be useful, there are several considerations around this mechanism:

1. Allowing liquidations to be disabled for a reserve is prone to misconfiguration (for example, blocking liquidations but allowing borrows of an asset would be considered a misconfig). In this sense, a grace period feels more appropriate than a binary flag, where a fixed time window is applied when the lever is activated.
2. It’s unclear whether the check should apply to the collateral side, the debt side, or both during liquidations. For flexibility, having a liquidatable flag on each side would allow the Spoke owner to control this behavior more granularly. 
3. The overall usefulness of this security lever is uncertain, as it heavily depends on the type of emergency the system is facing. Scenarios such as a stablecoin depeg, infinite minting, bridge failure, or an L2 sequencer outage raise open questions about how this mechanism should be used effectively to protect protocol solvency and users.

Given the points above, we decided to close this PR and remove the mechanism entirely (#1106), reducing code complexity, saving gas and bytecode size, and simplifying risk configuration.

PR #1105, opened 2026-01-09
Evaluate minting fee shares after each action

CheyenneAtapour:
Evaluates the approach of minting fee shares after each action

For purposes of examining the extra gas cost incurred with minting fee shares after each action


src/hub/Hub.sol
```diff
@@ -770,9 +775,24 @@ contract Hub is IHub, AccessManaged {
...
+    asset.updateDrawnRate(assetId);
+    return feeShares;
+  }
+
+  function _mintFeeShares(
+    Asset storage asset,
+    uint256 assetId,
+    uint256 fees
+  ) internal returns (uint256) {
+    uint120 shares = asset.toAddedSharesDown(fees, fees).toUint120();
```
Kogaroshi, line 795: worth it to add another getter in `AssetLogic` to prevent passing `(fees, fees)` there (to pass it only once, cleaner) ?

miguelmtzinf:
Closing this PR, as while the approach is valid, it adds roughly ~2k gas across the board. Leaning toward not minting fees on each action, but acting like they are by distributing the generated interest fairly between suppliers and the treasury, based on the proportion of added shares and treasury assets in the system since the last action. (#1154)

PR #1106, opened 2026-01-09
fix: removes liquidatable flag on Spoke

Kogaroshi:
Alternative to #1103 

Instead of grace periods or other alternatives, simply remove the `liquidatable` flag and check during liquidations.


Kogaroshi:
To close if #1103 is chosen instead

PR #1110, opened 2026-01-12
fix: Remove Spoke reserve getters

miguelmtzinf:
This draft implementation gives a sense of the savings on bytecode size we could get by removing auxiliary reserve getters from the Spoke. They are only calling the `Hub` contract, so this is actually more efficient. In case this creates any ux friction, we can consider creating an adapter contract to expose all Spoke information in only one place.

_Bytecode size from 24,523 to 23,870_

### DIFF
--- a/src/spoke/interfaces/ISpokeBase.sol
+++ b/src/spoke/interfaces/ISpokeBase.sol
@@ -169,32 +169,6 @@ interface ISpokeBase {
     bool receiveShares
   ) external;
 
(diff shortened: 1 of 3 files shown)


src/spoke/interfaces/ISpokeBase.sol
yan-man, line 1: should we also write the data provider contract and include here as part of this PR?

PR #1113, opened 2026-01-13
analysis: Gas cost of individual event emission in dynamic config update

CheyenneAtapour:
Analyzes the gas cost if we emit a single event per reserve when user's dynamic config is being updated


src/spoke/Spoke.sol
miguelmtzinf: probably better to change the ternary op to if cond

CheyenneAtapour: sure, I wanted to keep things consistent across all gas analysis branches, but if we are decided to proceed here, then I will update

src/spoke/Spoke.sol
```diff
@@ -714,52 +713,55 @@ abstract contract Spoke is
     KeyValueList.List memory collateralInfo = KeyValueList.init(
       positionStatus.collateralCount(reserveId)
     );
-    bool borrowing;
-    bool collateral;
+
     while (true) {
-      (reserveId, borrowing, collateral) = positionStatus.next(reserveId);
+      ProcessUserAccountDataVars memory vars;
```
Kogaroshi, line 718: are we forced to use that vars struct, or can we play with via ir runs see if it removes the StackTooDeep issue ?

Kogaroshi: or we can see with `assetPrice` & `assetUnit`, as they seem to only be used together on 2 occurrences, they could be merged into 1 var here ?

CheyenneAtapour:
Decided to proceed with current approach to emitting events for updating dynamic reserve config

PR #1114, opened 2026-01-13
chore: fix lint

DhairyaSethi:
Solidity file linting were not being picked up because with the current version of prettier we had to manually specify plugin in `.prettierrc`. This PR fixes this and pins `prettier` & `prettier-plugin-solidity` to the latest stable version.


.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
+  "plugins": ["prettier-plugin-solidity"],
```
DhairyaSethi, line 2: this was missing all this while 😭 
ty detective @avniculae

Kogaroshi (approved):

PR #1115, opened 2026-01-13
feat: access manager on configurators

DhairyaSethi:
replaces ownable with granular access manager for the configurators

### DIFF
--- a/tests/Base.t.sol
+++ b/tests/Base.t.sol
@@ -395,12 +395,8 @@ abstract contract Base is Test {
     selectors[17] = IHubConfigurator.haltSpoke.selector;
     selectors[18] = IHubConfigurator.freezeSpoke.selector;
     selectors[19] = IHubConfigurator.updateInterestRateData.selector;
-    // Handle overloaded addAsset functions - note: Solidity doesn't support disambiguating overloaded
-    // functions via interface selectors. We use the function signature hash as the only option.
-    selectors[20] = bytes4(keccak256('addAsset(address,address,address,uint256,address,bytes)'));
-    selectors[21] = bytes4(
-      keccak256('addAsset(address,address,uint8,address,uint256,address,bytes)')
-    );
+    selectors[20] = IHubConfigurator.addAsset.selector;
+    selectors[21] = IHubConfigurator.addAssetWithDecimals.selector;
     IAccessManager(manager).setTargetFunctionRole(
       hubConfigurator,
       selectors,


src/libraries/types/Roles.sol
```diff
@@ -10,4 +10,6 @@ library Roles {
   uint64 public constant HUB_ADMIN_ROLE = 1;
   uint64 public constant SPOKE_ADMIN_ROLE = 2;
   uint64 public constant USER_POSITION_UPDATER_ROLE = 3;
+  uint64 public constant HUB_CONFIGURATOR_ROLE = 4;
+  uint64 public constant SPOKE_CONFIGURATOR_ROLE = 5;
```
yan-man, line 14: do you think worth it to get super granular on these configurators, ie a role per method? like "freeze role" 
"inactive role", etc. just thinking if we had like a freeze agent that we only want to be able to set freeze status, or a separate asset listing admin that can add reserves, etc.

Kogaroshi: a role per method might be a bit overkill, imo the grouping per "type" of method makes sense.

DhairyaSethi: we decided not to do it now in the tests but theres a dedicated test which tests granularity for some reason

tests/unit/Hub/Hub.Restore.t.sol
```diff
@@ -11,17 +11,15 @@ contract HubRestoreTest is HubBase {
...
     super.setUp();
 
     // Set up a hub configurator to test freezing and pausing assets
-    hubConfigurator = new HubConfigurator(HUB_CONFIGURATOR_ADMIN);
     IAccessManager accessManager = IAccessManager(hub1.authority());
-    // Grant hubConfigurator hub admin role with 0 delay
-    vm.prank(ADMIN);
-    accessManager.grantRole(Roles.HUB_ADMIN_ROLE, address(hubConfigurator), 0);
+    hubConfigurator = new HubConfigurator(address(accessManager));
+    SpokeConfigurator spokeConfigurator = new SpokeConfigurator(address(accessManager));
```
Kogaroshi: why `spokeConfigurator` ?
Maybe we should split `setUpConfiguratorRoles` into 2 methods in Base and have that `setUpConfiguratorRoles` method wrap both.

tests/Base.t.sol
```diff
@@ -355,6 +360,94 @@ abstract contract Base is Test {
...
+    selectors[15] = IHubConfigurator.updateSpokeCaps.selector;
+    selectors[16] = IHubConfigurator.deactivateSpoke.selector;
+    selectors[17] = IHubConfigurator.haltSpoke.selector;
+    selectors[18] = IHubConfigurator.freezeSpoke.selector;
+    selectors[19] = IHubConfigurator.updateInterestRateData.selector;
+    // Handle overloaded addAsset functions - note: Solidity doesn't support disambiguating overloaded
+    // functions via interface selectors. We use the function signature hash as the only option.
+    selectors[20] = bytes4(keccak256('addAsset(address,address,address,uint256,address,bytes)'));
+    selectors[21] = bytes4(
+      keccak256('addAsset(address,address,uint8,address,uint256,address,bytes)')
```
miguelmtzinf: let's fix this please by renaming to `addAssetWithDecimals`

.github/workflows/tests-pr.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - name: Run Foundry setup
         uses: bgd-labs/github-workflows/.github/actions/foundry-setup@main
         with:
-          FOUNDRY_VERSION: stable
+          FOUNDRY_VERSION: nightly
```
yan-man, line 22: is this still needed here

tests/Base.t.sol
```diff
@@ -355,6 +360,90 @@ abstract contract Base is Test {
...
+    setUpSpokeConfiguratorRoles(SPOKE_CONFIGURATOR, address(manager));
+
+    vm.stopPrank();
+  }
+
+  function setUpHubConfiguratorRoles(address hubConfigurator, address manager) internal {
+    vm.startPrank(ADMIN);
+
+    // Grant HUB_ADMIN_ROLE so the configurator can call hub functions
+    IAccessManager(manager).grantRole(Roles.HUB_ADMIN_ROLE, hubConfigurator, 0);
```
yan-man, line 374: are we expecting in prod that we'll use the same manager contract for both hub and hubConfigurator

tests/Base.t.sol
```diff
@@ -355,6 +360,90 @@ abstract contract Base is Test {
...
+    selectors[16] = IHubConfigurator.deactivateSpoke.selector;
+    selectors[17] = IHubConfigurator.haltSpoke.selector;
+    selectors[18] = IHubConfigurator.freezeSpoke.selector;
+    selectors[19] = IHubConfigurator.updateInterestRateData.selector;
+    selectors[20] = IHubConfigurator.addAsset.selector;
+    selectors[21] = IHubConfigurator.addAssetWithDecimals.selector;
+    IAccessManager(manager).setTargetFunctionRole(
+      hubConfigurator,
+      selectors,
+      Roles.HUB_CONFIGURATOR_ROLE
```
yan-man, line 403: let's say we had an agent that can only call `freezeSpoke`. If we define `HUB_CONFIGURATOR_ROLE` with all these selectors, the selectors cannot be shared across multiple roles so it would be impossible to set up as is, correct?

.github/workflows/tests-pr.yml
DhairyaSethi: we decided to do this in the last sync in this pr to unblock ci

tests/Base.t.sol
DhairyaSethi: no, this config is not for prod at all

tests/Base.t.sol
DhairyaSethi: yea, see granular access test. this config is not for prod

src/libraries/types/Roles.sol
yan-man: I think we should either move this to a tests/libraries/types folder, or make this a prod-ready Roles setup. As it stands if we only want to use this set up for testing, it's a bit misleading to have it here

PR #1116, opened 2026-01-13
feat: update oz deps to 5.5

DhairyaSethi:
changelog: https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v5.5.0

relevent:
- `calldata` variant for isValidSignature
- erc6909 out of draft
- safeTransfer lib uses optimised logic

### DIFF
--- a/src/utils/IntentConsumer.sol
+++ b/src/utils/IntentConsumer.sol
@@ -33,7 +33,10 @@ abstract contract IntentConsumer is IIntentConsumer, NoncesKeyed, EIP712 {
   ) internal {
     require(block.timestamp <= deadline, InvalidSignature());
     bytes32 digest = _hashTypedData(intentHash);
-    require(SignatureChecker.isValidSignatureNow(signer, digest, signature), InvalidSignature());
+    require(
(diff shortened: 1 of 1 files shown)


src/dependencies/openzeppelin/Multicall.sol
```diff
@@ -23,7 +23,7 @@ abstract contract Multicall is Context {
    * @dev Receives and executes a batch of function calls on this contract.
    * @custom:oz-upgrades-unsafe-allow-reachable delegatecall
    */
-  function multicall(bytes[] calldata data) external virtual returns (bytes[] memory results) {
+  function multicall(bytes[] calldata data) public virtual returns (bytes[] memory results) {
```
miguelmtzinf, line 26: i wonder why this (ours is `external`)

DhairyaSethi: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5794 they added here, i think it's bc of solhint `interface-only-external` rule

PR #1117, opened 2026-01-14
analysis: Gas cost of array event emission in dynamic config update

CheyenneAtapour:
Analyze the gas cost of emitting an array of reserveIds when dynamic config is updated


CheyenneAtapour:
We decided against this approach

PR #1118, opened 2026-01-14
analysis: Gas cost of dynamic config update

CheyenneAtapour:
Analyzes the current gas cost of updating dynamic config when user has increasingly more collaterals. Meant to serve as a baseline to compare with other approaches


CheyenneAtapour:
Decided to proceed with current approach for event emission upon dynamic reserve config updates

PR #1122, opened 2026-01-15
fix: consistent transfer logic on reclaim

DhairyaSethi:
refactors hub `reclaim` to expect transfer rather than pulling asset pattern to be consistent with `add` and `restore`.

### DIFF
--- a/tests/unit/Hub/Hub.Reclaim.t.sol
+++ b/tests/unit/Hub/Hub.Reclaim.t.sol
@@ -5,6 +5,8 @@ pragma solidity ^0.8.0;
 import 'tests/unit/Hub/HubBase.t.sol';
 
 contract HubReclaimTest is HubBase {
+  using SafeERC20 for *;
+
   function test_reclaim_revertsWith_AssetNotListed() public {
     uint256 assetId = _randomInvalidAssetId(hub1);
     vm.expectRevert(IHub.AssetNotListed.selector);
@@ -82,7 +84,7 @@ contract HubReclaimTest is HubBase {
 
     deal(address(tokenList.dai), reinvestmentController, sweepAmount + 1);
     vm.prank(reinvestmentController);
-    tokenList.dai.transfer(address(hub1), sweepAmount + 1);
+    tokenList.dai.safeTransfer(address(hub1), sweepAmount + 1);
 
     vm.prank(reinvestmentController);
     vm.expectRevert(stdError.arithmeticError);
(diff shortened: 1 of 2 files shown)


src/hub/Hub.sol
```diff
@@ -447,12 +448,13 @@ contract Hub is IHub, AccessManaged {
     asset.accrue();
     _validateReclaim(asset, msg.sender, amount);
 
-    asset.liquidity += amount.toUint120();
+    uint256 liquidity = asset.liquidity + amount;
+    uint256 balance = IERC20(asset.underlying).balanceOf(address(this));
+    require(balance >= liquidity, InsufficientTransferred(liquidity.uncheckedSub(balance)));
+    asset.liquidity = liquidity.toUint120();
     asset.swept -= amount.toUint120();
     asset.updateDrawnRate(assetId);
```
miguelmtzinf, line 456:
```suggestion
asset.swept -= amount.toUint120();

asset.updateDrawnRate(assetId);
```
for readability

src/hub/interfaces/IHub.sol
```diff
@@ -331,6 +331,7 @@ interface IHub is IHubBase, IAccessManaged {
 
   /// @notice Reclaims an amount of liquidity of the corresponding asset from the configured reinvestment controller.
   /// @dev The controller can only reclaim up to swept amount. All accrued interest is distributed offchain.
+  /// @dev Underlying assets must be transferred to the Hub before invocation.
   /// @param assetId The identifier of the asset.
```
miguelmtzinf, line 336:
```suggestion
  /// @dev Extra underlying liquidity retained in the Hub can be skimmed by the investment controller through this action.
  /// @param assetId The identifier of the asset.
```

anything about skim?
- leftovers can be skimmed later
- controller can skim existing leftovers too

DhairyaSethi: yeah, very slight nuance (similar to restore) that you can skim only until previously swept 

i only took the docs on add & restore, we can this there as well bc it's true there too

tests/unit/Hub/Hub.Reclaim.t.sol
```diff
@@ -5,6 +5,8 @@ pragma solidity ^0.8.0;
 import 'tests/unit/Hub/HubBase.t.sol';
 
 contract HubReclaimTest is HubBase {
+  using SafeERC20 for *;
```
agryaznov: .SafeERC20 functions only make sense on IERC20 types

PR #1124, opened 2026-01-19
feat: sig gateway permit2 flows

DhairyaSethi:
introduce permit2 witness transfer on sig gateway supply & repay flows to improve UX & reduce user signatures required 

- reuses existing supply/repay struct to which duplicates nonce & expiry. this second nonce & deadline of gateway are validated - should we validate sig more? technically, if permit2 impl is correct, we dont need to validate and even have these fields 
- for repay, the permitWitnessTransferFrom intent requests repaid amount to be transferred which might be less than signed amount - this is ok bc permit2 supports this
- sig cannot be frontrun bc permit2 checks the signed spender is `msg.sender`

### DIFF
--- a/src/position-manager/SignatureGateway.sol
+++ b/src/position-manager/SignatureGateway.sol
@@ -277,6 +277,8 @@ contract SignatureGateway is ISignatureGateway, GatewayBase, IntentConsumer, Mul
   ) external onlyRegisteredSpoke(params.spoke) returns (uint256, uint256) {
     require(block.timestamp <= params.deadline, InvalidSignature());
     _useCheckedNonce(params.onBehalfOf, params.nonce);
+    address underlying = _getReserveUnderlying(params.spoke, params.reserveId);
+    require(permit.permitted.token == underlying, InvalidSignature());
 
     ISignatureTransfer(PERMIT2).permitWitnessTransferFrom(
       permit,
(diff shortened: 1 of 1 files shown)


tests/unit/misc/SignatureGateway/SignatureGateway.Base.t.sol
```diff
@@ -5,13 +5,15 @@ pragma solidity ^0.8.0;
 import 'tests/unit/Spoke/SpokeBase.t.sol';
 
 contract SignatureGatewayBaseTest is SpokeBase {
+  address internal constant CANONICAL_PERMIT2 = 0x000000000022D473030F116dDEE9F6B43aC78BA3;
```
Kogaroshi, line 8: Should be in `Constants`

tests/unit/misc/SignatureGateway/SignatureGateway.Permit2.Base.t.sol
```diff
@@ -0,0 +1,258 @@
...
+          params.reserveId,
+          params.amount,
+          params.onBehalfOf,
+          params.nonce,
+          params.deadline
+        )
+      );
+  }
+
+  function _approvePermit2(ISpoke spoke, uint256 reserveId, address who) internal {
```
Kogaroshi, line 204: Should we move this to Base and approve Constants.PERMIT2 ?

Kogaroshi: Same for non Gateway related Permit2 helpers

Kogaroshi:
> * reuses existing supply/repay struct to which duplicates nonce & expiry. this second nonce & deadline of gateway are validated - should we validate sig more? technically, if permit2 impl is correct, we dont need to validate and even have these fields

Having 2 nonces does not seem to be such an issue, since it's supposed to come from 2 different verifying contracts. Having twice the deadline seems a bit overkill, we need to ensure they are always the same.
Not validating the Spoke action in the Gateway is a high trust assumption on Permit2, I'd rather we verify in the Gateway too, but can't see a clean way to do so without redundant parts.

PR #1087, opened 2026-01-06
fix: vm.random* behaviour in tests

DhairyaSethi:

PR #1125, opened 2026-01-19
test: Fix `vm.random*` behaviour

avniculae:
resolve failing tests as a result of updates to vm.random behavior 

original PR: https://github.com/aave/aave-v4/pull/1087


tests/Base.t.sol
```diff
@@ -2233,6 +2233,10 @@ abstract contract Base is Test {
     }
   }
 
+  function _spokeDrawnIndex(ISpoke spoke, uint256 reserveId) internal view returns (uint256) {
```
yan-man: nit: this may fit more as _reserveDrawnIndex

avniculae: named it like so bc there is another fn called `_spokeAssetId`, but I agree `_reserveDrawnIndex` (and `_reserveAssetId` is better). do we have other PRs where we could do this change, or shall we do it now?

yan-man: i dont think there's another PR we could fit this into atm, but we can also leave this out of here anyway to get these tests resolved, nbd

foundry.toml
```diff
@@ -47,6 +47,14 @@ gas_snapshot_check = true
 test = 'tests/gas'
 isolate = true
 
+[profile.coverage]
+optimizer = true
+optimizer_runs = 444444444444
```
DhairyaSethi, line 52: @Kogaroshi what was the reason for this number here again? kinda scared of an forge integration mishap that it picks up wrong artifacts for the spoke/hub deployment bc in the default profile every other contract is compiled w these settings. perhaps we can do a unique setting here, i know forge coverage always recompiles looks like which is a plus

foundry.toml
```diff
@@ -34,7 +35,6 @@ lint_on_build = false
 
 [profile.default.fuzz]
 runs = 1000
-seed = "0x640"
```
DhairyaSethi, line 37: should we bring this back and instead switch to nightly given our recent trauma until the fix gets in stable xD

DhairyaSethi: actually think its coming in stable soon enough https://github.com/foundry-rs/foundry/pull/13124

foundry.toml
Kogaroshi: No real reason for this value, we can replace and change with any other.

foundry.toml
avniculae: no, I think this is fine. I think our challenges are mostly due to forge's limitations

tests/unit/Hub/Hub.Config.t.sol
```diff
@@ -690,21 +690,6 @@ contract HubConfigTest is HubBase {
     assertEq(hub1.getSpokeAddedShares(assetId, newFeeReceiver), newFees);
   }
 
-  /// Updates the fee receiver to an existing spoke of the hub1, so ends up with existing supplied shares plus accrued fees
-  function test_updateAssetConfig_fuzz_UseExistingSpokeAsFeeReceiver_revertsWith_SpokeAlreadyListed(
```
CheyenneAtapour, line 692: why did we delete this test?

avniculae: https://github.com/aave/aave-v4/pull/1087#discussion_r2690160643

tests/Base.t.sol
yan-man: this method was actually not used at all so i removed it and renamed the other one reserveAssetId

PR #1126, opened 2026-01-20
fix: eliminate deficit zero check

DhairyaSethi:
existence check on covered spoke, check non zero amount on actual deficit cleared

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -893,10 +893,10 @@ contract Hub is IHub, AccessManaged {
 
   function _validateEliminateDeficit(
     SpokeData storage callerSpoke,
-    uint256 deficitCoveredRay
+    uint256 deficitAmountRay
   ) internal view {
     require(callerSpoke.active, SpokeNotActive());
-    require(deficitCoveredRay > 0, InvalidAmount());
+    require(deficitAmountRay > 0, InvalidAmount());
   }
 
   function _validatePayFeeShares(SpokeData storage spoke, uint256 feeShares) internal view {
--- a/tests/unit/Hub/Hub.EliminateDeficit.t.sol
+++ b/tests/unit/Hub/Hub.EliminateDeficit.t.sol
@@ -32,7 +32,6 @@ contract HubEliminateDeficitTest is HubBase {
 
   function test_eliminateDeficit_revertsWith_InvalidAmount_ZeroAmountWithDeficit() public {
     _createDeficit(_assetId, _coveredSpoke, _deficitAmountRay);
-    assertEq(hub1.getSpokeDeficitRay(_assetId, _coveredSpoke), _deficitAmountRay);
     vm.expectRevert(IHub.InvalidAmount.selector);
     vm.prank(_callerSpoke);
     hub1.eliminateDeficit(_assetId, 0, _coveredSpoke);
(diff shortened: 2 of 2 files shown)


yan-man:
ha can you rephrase the title to something maybe like "zero check on `eliminateDeficit`", just when I read it initially thought you were removing the zero check on deficits

src/hub/Hub.sol
```diff
@@ -891,9 +891,12 @@ contract Hub is IHub, AccessManaged {
     require(premiumAmountRay <= premiumRay, SurplusPremiumRayDeficitReported(premiumRay));
   }
 
-  function _validateEliminateDeficit(SpokeData storage spoke, uint256 amount) internal view {
-    require(spoke.active, SpokeNotActive());
-    require(amount > 0, InvalidAmount());
+  function _validateEliminateDeficit(
+    SpokeData storage callerSpoke,
+    uint256 deficitCoveredRay
```
yan-man: maybe we can align var names to have them both be `deficitCoveredRay` or `deficitAmountRay` for consistency

tests/unit/Hub/Hub.EliminateDeficit.t.sol
```diff
@@ -38,6 +38,24 @@ contract HubEliminateDeficitTest is HubBase {
     hub1.eliminateDeficit(_assetId, 0, _coveredSpoke);
   }
 
+  function test_eliminateDeficit_revertsWith_SpokeNotActive_on_UnregisteredAsset() public {
```
yan-man, line 40: for completeness, perhaps another revert test on `SpokeNotActive` error on a registered asset that has been set inactive

tests/unit/Hub/Hub.EliminateDeficit.t.sol
```diff
@@ -38,6 +38,24 @@ contract HubEliminateDeficitTest is HubBase {
...
+    _createDeficit(_assetId, _coveredSpoke, _deficitAmountRay);
+    assertEq(hub1.getSpokeDeficitRay(_assetId, _coveredSpoke), _deficitAmountRay);
```
yan-man, line 42: i think this assert can be moved to the helper `_createDeficit` so that it's always checked. And perhaps update the helper to accept hub input rather than defaulting to hub1

tests/unit/Hub/Hub.EliminateDeficit.t.sol
DhairyaSethi: exists below

tests/unit/Hub/Hub.EliminateDeficit.t.sol
DhairyaSethi: helper isn't used outside this file and its defined here itself. specific input requires creating `_hub` input to follow this file's schemantics and solidity doesnt like bc it we have a _hub(spoke, reserveId) helper smh

yan-man: ok that part's fine, but good to still just add that assert as a sanity check in the helper itself

DhairyaSethi: added assert in helper yep

CheyenneAtapour (approved):

PR #1127, opened 2026-01-20
feat: naive strongly typed sig batcher

DhairyaSethi:
dynamically builds typestring at runtime based on input

todo opts:
- execute individual action alongside building the typestring to save on loops 
- OR the final loop for executing actions can be a multicall
- tstore map/mem fixed array for actionId to action lookup

### DIFF
--- a/src/position-manager/interfaces/ISignatureGateway.sol
+++ b/src/position-manager/interfaces/ISignatureGateway.sol
@@ -22,181 +22,161 @@ interface ISignatureGateway is IGatewayBase, IIntentConsumer, IMulticall {
     UpdateUserDynamicConfig // 6
   }
 
-  /// @notice Batch action for supplying assets (without nonce/deadline/onBehalfOf - lifted to batch level).
+  /// @notice Parameters for supplying assets.
   /// @param spoke The address of the registered spoke.
   /// @param reserveId The identifier of the reserve.
   /// @param amount The amount of assets to supply.
-  struct SupplyAction {
+  struct SupplyParams {
     address spoke;
     uint256 reserveId;
     uint256 amount;
   }
 
-  /// @notice Batch action for withdrawing assets.
+  /// @notice Parameters for withdrawing assets.
   /// @param spoke The address of the registered spoke.
   /// @param reserveId The identifier of the reserve.
   /// @param amount The amount of assets to withdraw.
-  struct WithdrawAction {
+  struct WithdrawParams {
     address spoke;
     uint256 reserveId;
     uint256 amount;
   }
 
-  /// @notice Batch action for borrowing assets.
+  /// @notice Parameters for borrowing assets.
   /// @param spoke The address of the registered spoke.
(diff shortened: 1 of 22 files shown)


src/position-manager/interfaces/ISignatureGateway.sol
```diff
@@ -11,6 +11,79 @@ import {ISignatureTransfer} from 'lib/permit2/src/interfaces/ISignatureTransfer.
 /// @author Aave Labs
 /// @notice Minimal interface for protocol actions involving signed intents.
 interface ISignatureGateway is IGatewayBase, IIntentConsumer, IMulticall {
+  /// @notice Action type enumeration for batch operations.
+  enum ActionType {
```
Kogaroshi, line 15: How about using constants + encoding all actions as `bytes` ?
Similar to https://github.com/Uniswap/v4-periphery/blob/main/src/libraries/Actions.sol & https://github.com/Uniswap/v4-periphery/blob/main/src/base/BaseActionsRouter.sol

src/position-manager/interfaces/ISignatureGateway.sol
```diff
@@ -11,6 +11,79 @@ import {ISignatureTransfer} from 'lib/permit2/src/interfaces/ISignatureTransfer.
...
+
+  /// @notice Batch action for repaying assets.
+  /// @param spoke The address of the registered spoke.
+  /// @param reserveId The identifier of the reserve.
+  /// @param amount The amount of assets to repay.
+  struct RepayAction {
+    address spoke;
+    uint256 reserveId;
+    uint256 amount;
+  }
```
Kogaroshi, line 63: can we use only 1 struct instead of those 4 since it's the same parameters in the end ? Only the action type is the difference there.

Kogaroshi: nvm, now used for specific sig methods as well, might be better to keep for clarity.

src/position-manager/interfaces/ISignatureGateway.sol
DhairyaSethi: yes definitely next on the list to optimise decoding; im still thinking through the types given the idea that user should see exactly what they're signing, the supplemented data can be as compressed as needed

im also thinking the final loop for executing actions can be a multicall lol

Kogaroshi: not 100% sure for the multicall xD but yes on the rest !

DhairyaSethi: i think we should 
- have non withSig method for this to enable cross spoke communication?
- do something where this gateway can skim tokens as well bc rn to withdraw from one spoke & deposit to another user has to receive tokens to themselves and have allowance to pull tokens to supply to another

DhairyaSethi: im now thinking the calldata opt really doesnt help us because we anyway need to lift the vars in memory to build the struct hash

PR #1131, opened 2026-01-21
fix: rename spoke `paused` flag to `halted` in hub

### LINKED ISSUE #1083: Same term name for Hub `paused` asset and Spoke `paused` reserve can be confusing.
Using the same `paused` term in both the Hub and the Spoke can be confusing.

- The `paused` state in the Hub means that no actions with an immediate impact on the Hub’s liquidity from a given Spoke are allowed.
- The `paused` state in the Spoke means that no actions are allowed at all.

Since these are different components with different semantics, it would make sense to use distinct terminology for each to avoid confusion.

Some alternatives to consider (with clearer semantics):
- sealed: implies that liquidity changes on the Hub are blocked
- locked: may be softer than `active`, and suggests a temporary state
- halted: implies a temporary stop that can be resumed

Open to suggestions!

yan-man:
- rename `SpokeData` `paused` flag to `halted` in IHub struct definition to better describe the intended temporary nature of a halt operation

closes #1083

### DIFF
--- a/src/hub/interfaces/IHub.sol
+++ b/src/hub/interfaces/IHub.sol
@@ -72,7 +72,7 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev addCap The maximum amount that can be added by a spoke, expressed in whole assets (not scaled by decimals). A value of `MAX_ALLOWED_SPOKE_CAP` indicates no cap.
   /// @dev drawCap The maximum amount that can be drawn by a spoke, expressed in whole assets (not scaled by decimals). A value of `MAX_ALLOWED_SPOKE_CAP` indicates no cap.
   /// @dev riskPremiumThreshold The maximum ratio of premium to drawn shares a spoke can have, expressed in BPS. A value of `MAX_RISK_PREMIUM_THRESHOLD` indicates no threshold.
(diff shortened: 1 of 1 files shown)


src/hub/interfaces/IHub.sol
```diff
@@ -73,7 +73,7 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev drawCap The maximum amount that can be drawn by a spoke, expressed in whole assets (not scaled by decimals). A value of `MAX_ALLOWED_SPOKE_CAP` indicates no cap.
   /// @dev riskPremiumThreshold The maximum ratio of premium to drawn shares a spoke can have, expressed in BPS. A value of `MAX_RISK_PREMIUM_THRESHOLD` indicates no threshold.
   /// @dev active True if the spoke is prevented from performing any actions.
```
CheyenneAtapour: Isn't this incorrect?

yan-man: yes you are right, it should be opposite. Good catch!

CheyenneAtapour: maybe keep it as "true" but say "allowed to perform any actions."

CheyenneAtapour (approved):

PR #1133, opened 2026-01-21
feat: assetId from underlying getter

Kogaroshi:
Add a new getter on the Hub to fetch the assetId with a given listed underlying asset address. Simplifies need for offchain indexing.


src/hub/Hub.sol
```diff
@@ -463,6 +467,17 @@ contract Hub is IHub, AccessManaged {
     return _underlyingAssets.contains(underlying);
   }
 
+  /// @inheritdoc IHub
+  function getUnderlyingAssets() external view returns (address[] memory) {
+    return _underlyingAssets.values();
+  }
+
+  /// @inheritdoc IHub
+  function getUnderlyingAssetId(address underlying) external view returns (uint256) {
```
yan-man: imo can just be `getAssetId`

src/hub/Hub.sol
```diff
@@ -55,6 +55,9 @@ contract Hub is IHub, AccessManaged {
   /// @dev Set of underlying addresses listed as assets in the Hub.
   EnumerableSet.AddressSet internal _underlyingAssets;
 
+  /// @dev Map of underlying addresses to asset identifiers.
+  mapping(address underlying => uint256) internal _underlyingToAssetId;
```
yan-man:
```suggestion
  mapping(address underlying => uint256 assetId) internal _underlyingToAssetId;
```

jw why not this? or vice versa both without

Kogaroshi: c7601d0897a502bb6e53b497be4ddd36cbf1705c

src/hub/Hub.sol
miguelmtzinf: in favor

Kogaroshi: c73ef32875868387fcee194a7c642d9c786908ee

src/hub/Hub.sol
```diff
@@ -111,6 +111,7 @@ contract Hub is IHub, AccessManaged {
       feeReceiver: feeReceiver,
       liquidityFee: 0
     });
+    _underlyingToAssetId[underlying] = assetId;
```
DhairyaSethi: add it first, right after the require pls
there's no reentrency concern w gov methods but we want to follow it for sanity after certora's suggestions, its why we did #1093

src/hub/Hub.sol
```diff
@@ -459,8 +461,14 @@ contract Hub is IHub, AccessManaged {
   }
 
   /// @inheritdoc IHub
-  function isUnderlyingListed(address underlying) external view returns (bool) {
-    return _underlyingAssets.contains(underlying);
+  function isUnderlyingListed(address underlying) public view returns (bool) {
+    return _assets[_underlyingToAssetId[underlying]].underlying == underlying;
+  }
```
DhairyaSethi, line 466: love the idea!

would like more tests on this, i think one specifically covering unregistered assetId (false) case
you can write a clean symbolic one using setArbitraryStorage

DhairyaSethi (approved):

src/hub/Hub.sol
CheyenneAtapour: def agree we need a test for the false case. 

This gives me a headache, but yeah seems fine. Tricky cause my alternative idea is just to check the _underlyingToAssetId dictionary, but 0 can ofc be a valid ID

src/hub/interfaces/IHub.sol
```diff
@@ -342,6 +342,11 @@ interface IHub is IHubBase, IAccessManaged {
   /// @return True if the underlying asset is listed.
   function isUnderlyingListed(address underlying) external view returns (bool);
 
+  /// @notice Returns the assetId for the specified underlying asset.
```
yan-man: i think this is our only getter where it reverts on an invalid input, the others i believe just default to base values. Maybe can include this detail in natspec

Kogaroshi: b4ec027a60395a0a0b82755cfcf41cf9f0e1bbf5

src/hub/interfaces/IHub.sol
```diff
@@ -342,6 +342,12 @@ interface IHub is IHubBase, IAccessManaged {
...
+  /// @dev Reverts with `AssetNotListed` if the underlying is not listed.
+  /// @param underlying The address of the underlying asset.
+  /// @return The assetId of the underlying asset.
```
```suggestion
  /// @return The `assetId` of the underlying asset.
```

src/hub/interfaces/IHub.sol
```diff
@@ -342,6 +342,12 @@ interface IHub is IHubBase, IAccessManaged {
   /// @return True if the underlying asset is listed.
   function isUnderlyingListed(address underlying) external view returns (bool);
 
+  /// @notice Returns the assetId for the specified underlying asset.
```
```suggestion
  /// @notice Returns the asset identifier for the specified underlying asset address.
```

Kogaroshi: d16ddd91078e91f86b85810ded75f6a85e548019

src/hub/interfaces/IHub.sol
Kogaroshi: d16ddd91078e91f86b85810ded75f6a85e548019

PR #1135, opened 2026-01-22
fix: Fix intent struct onBehalfOf param name on SignatureGateway and Spoke

miguelmtzinf:
- In `SignatureGateway`, some functions use `onBehalfOf` as parameter name (`setUsingAsCollateralWithSig()`), while others use `user` (`setSelfAsUserPositionManagerWithSig()`) to refer to the same address.

https://github.com/aave/aave-v4/blob/7b7db47d0e872d9af91a02acbcab5e13fe3d980e/src/position-manager/interfaces/ISignatureGateway.sol#L84-L100

- In `Spoke`, the `SetUserPositionManagers` data struct uses `user` instead of  `onBehalfOf`.
https://github.com/aave/aave-v4/blob/7b7db47d0e872d9af91a02acbcab5e13fe3d980e/src/spoke/interfaces/ISpoke.sol#L23-L27

### DIFF
--- a/src/position-manager/SignatureGateway.sol
+++ b/src/position-manager/SignatureGateway.sol
@@ -57,9 +57,9 @@ contract SignatureGateway is ISignatureGateway, GatewayBase, IntentConsumer, Mul
   ) external onlyRegisteredSpoke(params.spoke) returns (uint256, uint256) {
     address spoke = params.spoke;
     uint256 reserveId = params.reserveId;
-    address user = params.onBehalfOf;
+    address onBehalfOf = params.onBehalfOf;
     _verifyAndConsumeIntent({
-      signer: user,
+      signer: onBehalfOf,
       intentHash: params.hash(),
       nonce: params.nonce,
       deadline: params.deadline,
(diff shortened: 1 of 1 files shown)


tests/mocks/EIP712Types.sol
```diff
@@ -83,7 +83,7 @@ library EIP712Types {
 
   struct UpdateUserDynamicConfig {
     address spoke;
-    address user;
+    address onBhela;
```
miguelmtzinf:
```suggestion
    address onBehalfOf;
```

tests/mocks/EIP712Types.sol
miguelmtzinf:
```suggestion
    address onBehalfOf;
```

src/position-manager/interfaces/ISignatureGateway.sol
yan-man, line 1: should GatewayBase `renouncePositionManagerRole` also refer to onBehalfOf instead of user?

yan-man, line 1: also _supplyNative in NativeTokenGateway?

do you care about internal vars within a function maintaining consistency too, ie `supplyWithSig` referencing user instead of onBehalfOf

miguelmtzinf: i don't think so, i d only do it for sig methods.

PR #1029, opened 2025-11-26
feat: init tokenisation spoke

DhairyaSethi:
Implements ERC4626 vault to tokenise hub positions. Share price is maintained by the hub, fee-less accounting. Only acts as a proxy to tokenise supply-only positions.

PR #1136, opened 2026-01-22
fix: natspec clean up; remove unneeded imports

Kogaroshi:
- Update natspec in gateway contracts
- Remove unneeded import from SpokeInstance

To be merged before #1029


DhairyaSethi:
i think its fine 1029 does this cleanup

Kogaroshi:
Goal is to have only TokenizedSpoke changes in 1029, and get all cleanup/small changes beforehand, so it's easier for reviews

src/spoke/instances/SpokeInstance.sol
```diff
@@ -2,7 +2,6 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity 0.8.28;
 
-import {LiquidationLogic} from 'src/spoke/libraries/LiquidationLogic.sol';
```
DhairyaSethi, line 5: this is the only new cleanup all of the others are done bc of the new tests introduced in that PR so i would keep scope of test updates there so its more trackable

src/spoke/instances/SpokeInstance.sol
DhairyaSethi, line 1: same issue on INativeTokenGateway

yan-man (approved):

yan-man (approved):

DhairyaSethi (approved):

PR #1141, opened 2026-01-26
feat : new getter for reserveId

Kogaroshi:
Add a new getter on the Spoke to fetch a `reserveId` for a given `Hub` & `assetId` couple.

### DIFF
--- a/src/spoke/SpokeStorage.sol
+++ b/src/spoke/SpokeStorage.sol
@@ -18,7 +18,7 @@ abstract contract SpokeStorage {
   /// @dev Map of reserve identifiers to their Reserve data.
   mapping(uint256 reserveId => ISpoke.Reserve) internal _reserves;
 
-  /// @dev Map of hub addresses and asset identifiers to the reserveId.
+  /// @dev Map of hub addresses and asset identifiers to the reserve identifier.
   mapping(address hub => mapping(uint256 assetId => uint256 reserveId))
     internal _hubAssetIdToReserveId;
 


tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -262,6 +263,20 @@ contract SpokeConfigTest is SpokeBase {
...
+  function test_getReserveId_fuzz(uint256 reserveId) public view {
+    reserveId = bound(reserveId, 0, spoke1.getReserveCount() - 1);
+    uint256 assetId = spoke1.getReserve(reserveId).assetId;
+
+    uint256 returnedId = spoke1.getReserveId(address(hub1), assetId);
+    assertEq(returnedId, getReserveIdByAssetId(spoke1, hub1, assetId));
+  }
+
+  function test_getReserveId_fuzz_revertsWith_ReserveNotListed(uint256 assetId) public {
+    assetId = bound(assetId, hub1.getAssetCount() + 1, 100);
```
yan-man:
```suggestion
    assetId = bound(assetId, hub1.getAssetCount(), 100);
```
nit: think we can start from assetCount

src/spoke/interfaces/ISpoke.sol
```diff
@@ -455,6 +455,13 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IExtSload, IMul
   /// @dev Count includes reserves that are not currently active.
   function getReserveCount() external view returns (uint256);
 
+  /// @notice Returns the reserve identifier for a given assetId in a Hub.
+  /// @dev It reverts if no reserve is associated with the given assetId.
```
yan-man, line 466: nice

tests/unit/Spoke/Spoke.Config.t.sol
Kogaroshi: 83aec94fe2c4a6c5c87a36205d96fcb146f2cf27

src/spoke/Spoke.sol
```diff
@@ -150,6 +149,9 @@ abstract contract Spoke is
     (address underlying, uint8 decimals) = IHubBase(hub).getAssetUnderlyingAndDecimals(assetId);
     require(underlying != address(0), AssetNotListed());
 
+    require(_reserves[_assetIdToReserveId[hub][assetId]].underlying != underlying, ReserveExists());
```
yan-man: do you think worth it to move this and L150 to L142 to fail earlier, prior to the `_validateReserveConfig` and `_validateDynamicReserveConfig`

Kogaroshi: We need to fetch `underlying` from the Hub to do this require, this is why I set it there, with the previous require ensuring `underlying != address(0)`

src/spoke/Spoke.sol
```diff
@@ -543,6 +545,16 @@ abstract contract Spoke is
     return reserve.hub.getSpokeTotalOwed(reserve.assetId, address(this));
   }
 
+  /// @inheritdoc ISpoke
+  function getReserveId(address hub, uint256 assetId) external view returns (uint256) {
+    uint256 reserveId = _assetIdToReserveId[hub][assetId];
+    require(
+      _reserves[reserveId].assetId == assetId && address(_reserves[reserveId].hub) == hub,
+      ReserveNotListed()
+    );
```
miguelmtzinf: moving this to an internal method `_isHubAssetListed`/`_isAssetListed` saves anything in terms of codesize?

Kogaroshi: saves 24B of codesize. Will push the change

Kogaroshi: b2e77e3f3c0d13d9ac83ed21f1d21868e809c66d

src/spoke/Spoke.sol
```diff
@@ -148,8 +149,6 @@ abstract contract Spoke is
   ) external restricted returns (uint256) {
     require(hub != address(0), InvalidAddress());
     require(assetId <= MAX_ALLOWED_ASSET_ID, InvalidAssetId());
-    require(!_reserveExists[hub][assetId], ReserveExists());
```
DhairyaSethi, line 131: can we add require(!_isAssetListed(hub, assetId, reserveId)) here instead?

Kogaroshi: 5b99072f29f174587b7ed28571bd49614d4a4937

src/spoke/Spoke.sol
```diff
@@ -905,6 +912,14 @@ abstract contract Spoke is
     }
   }
 
+  function _isAssetListed(
```
DhairyaSethi: maybe more accurate is _isAssetListedAt(Reserve)? bc the current name seems like it should just expect hub, assetId or underlying

src/spoke/SpokeStorage.sol
```diff
@@ -18,8 +18,9 @@ abstract contract SpokeStorage {
   /// @dev Map of reserve identifiers to their Reserve data.
   mapping(uint256 reserveId => ISpoke.Reserve) internal _reserves;
 
-  /// @dev Map of hub addresses and asset identifiers to whether the reserve exists.
-  mapping(address hub => mapping(uint256 assetId => bool)) internal _reserveExists;
+  /// @dev Map of hub addresses and asset identifiers to the reserveId.
+  mapping(address hub => mapping(uint256 assetId => uint256 reserveId))
+    internal _assetIdToReserveId;
```
DhairyaSethi:
```suggestion
    internal _assetToReserveId;
```

_asset/_underlying = (_hub, _assetId)
?

DhairyaSethi: actually current is ok as well, nbd

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -278,6 +279,20 @@ contract SpokeConfigTest is SpokeBase {
...
+  function test_getReserveId_fuzz(uint256 reserveId) public view {
+    reserveId = bound(reserveId, 0, spoke1.getReserveCount() - 1);
+    uint256 assetId = spoke1.getReserve(reserveId).assetId;
+
+    uint256 returnedId = spoke1.getReserveId(address(hub1), assetId);
+    assertEq(returnedId, getReserveIdByAssetId(spoke1, hub1, assetId));
+  }
+
+  function test_getReserveId_fuzz_revertsWith_ReserveNotListed(uint256 assetId) public {
+    assetId = bound(assetId, hub1.getAssetCount(), 100);
```
DhairyaSethi:
```suggestion
    assetId = bound(assetId, hub1.getAssetCount(), UINT256_MAX);
```
why 100?

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -278,6 +279,20 @@ contract SpokeConfigTest is SpokeBase {
     );
   }
 
+  function test_getReserveId_fuzz(uint256 reserveId) public view {
+    reserveId = bound(reserveId, 0, spoke1.getReserveCount() - 1);
```
DhairyaSethi, line 305: should we add a bunch of reserves before these tests to be more robust?

and also add reserves with multihub (same _assetId different hub) etc

src/spoke/Spoke.sol
```diff
@@ -905,6 +912,14 @@ abstract contract Spoke is
...
+    address hub,
+    uint256 assetId,
+    uint256 reserveId
+  ) internal view returns (bool) {
+    return _reserves[reserveId].assetId == assetId && address(_reserves[reserveId].hub) == hub;
```
DhairyaSethi, line 906: cache reserve struct, does it reduce code size? opt must do something here but just in case, in the naive impl we could avoid slot derivation calc twice 
also these two are in the same slot, maybe some trick we can use in assembly to do it ourselves (i know you will say not worth it)

DhairyaSethi: the trick would SIMD basically https://github.com/aave/aave-v4/pull/1180

src/spoke/SpokeStorage.sol
Kogaroshi: I'd rather keep the current name, since we only track the Id and no other asset data.

src/spoke/Spoke.sol
Kogaroshi: renamed to `_isAssetIdListed`, since it's the info we check on a given Hub.
but open to rename if we find something more clear.

tests/unit/Spoke/Spoke.Config.t.sol
Kogaroshi: dunno, arbitrary

tests/unit/Spoke/Spoke.Config.t.sol
Kogaroshi: 791984ab9849c8bd70c9696d98e78d72ea2f1e97

tests/unit/Spoke/Spoke.Config.t.sol
Kogaroshi: 791984ab9849c8bd70c9696d98e78d72ea2f1e97

src/spoke/Spoke.sol
Kogaroshi: 791984ab9849c8bd70c9696d98e78d72ea2f1e97

src/spoke/interfaces/ISpoke.sol
miguelmtzinf: hope this behaviour is okay for integrators... otherwise we could consider returning max_uint or smt

src/spoke/SpokeStorage.sol
miguelmtzinf: we could also do `_hubAssetIdToReserveId`, probably not worth

src/spoke/Spoke.sol
miguelmtzinf, line 145:
```suggestion
(address underlying, uint8 decimals) = IHubBase(hub).getAssetUnderlyingAndDecimals(assetId);
require(underlying != address(0), AssetNotListed());

uint256 reserveId = _reserveCount++;
 _assetIdToReserveId[hub][assetId] = reserveId;

_updateReservePriceSource(reserveId, priceSource);

uint24 dynamicConfigKey; // 0 as first key to use
_reserves[reserveId] = Reserve({
```

this order makes more sense to me, hope i'm not missing anything

DhairyaSethi: no we cannot do this, the reentrance lock (assigning the _assetIdToReserveItMapping) needs to happen before the hub ext call from tomer's paranoia

src/spoke/SpokeStorage.sol
Kogaroshi: `_hubAssetIdToReserveId` could work too, so all keys are mentioned. No strong prefs tho

src/spoke/Spoke.sol
Kogaroshi: Yeah, I prefer to keep the current order as we do not need to use the `underlying` in the check, so all those validation and updates are done before the ext call.
`dynamicConfigKey` can be moved down tho yes.

src/spoke/SpokeStorage.sol
Kogaroshi: 1f704d63b16d11961f06c8d472489016a169badd

src/spoke/SpokeStorage.sol
```diff
@@ -18,8 +18,9 @@ abstract contract SpokeStorage {
   /// @dev Map of reserve identifiers to their Reserve data.
   mapping(uint256 reserveId => ISpoke.Reserve) internal _reserves;
 
-  /// @dev Map of hub addresses and asset identifiers to whether the reserve exists.
-  mapping(address hub => mapping(uint256 assetId => bool)) internal _reserveExists;
+  /// @dev Map of hub addresses and asset identifiers to the reserveId.
```
miguelmtzinf:
```suggestion
  /// @dev Map of hub addresses and asset identifiers to the reserve identifier.
```

Kogaroshi: 98de9e4779d870ca2113011817277369b3d99a57

src/spoke/Spoke.sol
Kogaroshi: https://github.com/aave/aave-v4/pull/1180#issuecomment-3861162450

PR #1134, opened 2026-01-21
opt: Health Factor & Liquidations Precision

avniculae:
- Increase precision of debt when calculating user's health factor.
- Perform liquidation calculations in share terms for increased precision. Handle case when recalculated debt to liquidate can exceed total debt (in case collateral is not sufficient or there is collateral dust).

PR #1142, opened 2026-01-26
fix: consistent rounding on risk premium computation

DhairyaSethi:
- changes rounding on risk premium calculation for user by rounding up the weighted average; consistent with how risk premium is applied (rounding up when computing `premiumShares`)

- includes proof for max risk premium threshold value to be set as max_collateral_risk + percentage_factor 
  - this is needed because of precision error accumulation:
   we round up when computing user premiumShares: 
https://github.com/aave/aave-v4/blob/b15540a7fc5311c63cb9fe54a918e59fceedc41c/src/spoke/libraries/UserPositionDebt.sol#L59-L61
  and when comparing the aggregated values of the spoke: 
  https://github.com/aave/aave-v4/blob/b15540a7fc5311c63cb9fe54a918e59fceedc41c/src/hub/Hub.sol#L775-L779

  where `spoke{Premium,Drawn}Shares = sum(user{Premium,Drawn}Shares)`; the individual error will accumulate due to rounding ==> this makes max(riskPremiumBps) == riskPremiumThreshold an incorrect choice; hence this PR includes a proof of the overshoot required by governance while setting risk premium threshold  
    
 we will be providing a writeup on this recommendation, a more accurate but less consecrative minima also can be `divUp(max_collateral_risk, percentageFactor) * percentageFactor`, ie the nearest whole multiple of percentage_factor

### DIFF
--- a/tests/misc/risk_premium_threshold_overshoot.py
+++ b/tests/misc/risk_premium_threshold_overshoot.py
@@ -12,7 +12,7 @@ def percentMulUp(value, percentage):
 # we want to minimize RISK_PREMIUM_THRESHOLD such that:
 # Σ ⌈(drawnShares_i × riskPremium_i) / PERCENTAGE_FACTOR⌉ ≤ ⌈((Σ drawnShares_i) × RISK_PREMIUM_THRESHOLD) / PERCENTAGE_FACTOR⌉
 
-# minimize LHS using the property ⌈x⌉ ≤ x + 1, and maximize RHS using the property ⌈x⌉ ≥ x
+# maximize LHS using the property ⌈x⌉ ≤ x + 1, and minimize RHS using the property ⌈x⌉ ≥ x
 # Σ ((drawnShares_i × riskPremium_i) / PERCENTAGE_FACTOR + 1) ≤ ((Σ drawnShares_i) × RISK_PREMIUM_THRESHOLD) / PERCENTAGE_FACTOR
 
 # for N as number of users, the above simplifies to:


tests/misc/risk_premium_weighted_average_bound.py
```diff
@@ -0,0 +1,24 @@
...
+# N-agnostic: represent sum(k_i * v_i) as numerator, sum(k_i) as denominator
+numerator = Int('numerator')
+denominator = Int('denominator')
+
+s.add(denominator >= 1)
+s.add(numerator >= 0)
+# v_i <= v_max
+# implies; k_i * v_i <= k_i * v_max
+# implies; sum(k_i * v_i) <= v_max * sum(k_i)
+s.add(numerator <= MAX_KEY * denominator)
```
avniculae: this should be percentMulUp, no?

also, not sure how useful this script is, as all it will tell us is: when premium is MAX_KEY, there is no rounding error. if premium is not MAX_KEY, then after rounding effective rounding cannot be more than MAX_KEY. This we know from percentMulUp properties, no? (or we can just prove this percentMulUp property more generally)

tests/misc/risk_premium_weighted_average_bound.py
```diff
@@ -0,0 +1,24 @@
...
+from z3 import *
+
+MAX_KEY = IntVal(1000_00) # MAX_ALLOWED_COLLATERAL_RISK
+
+def divUp(a, b):
+    return (a + b - 1) / b
+
+s = Solver()
+
+# N-agnostic: represent sum(k_i * v_i) as numerator, sum(k_i) as denominator
```
avniculae: same as this https://github.com/aave/aave-v4/pull/1142/changes#r2727591802

this assume we store premiumSharesBps? (which we don't)

it should be sum(percentMulUp(k_i, v_i)), no?

tests/misc/risk_premium_threshold_overshoot.py
```diff
@@ -0,0 +1,45 @@
...
+
+PERCENTAGE_FACTOR = IntVal(100_00)
+MAX_COLLATERAL_RISK = IntVal(1000_00)
+
+def percentMulUp(value, percentage):
+    return (value * percentage + PERCENTAGE_FACTOR - 1) / PERCENTAGE_FACTOR
+
+# ∀ drawnShares_i ≥ 1, ∀ riskPremium_i ≤ MAX_COLLATERAL_RISK,
+# we want to minimize RISK_PREMIUM_THRESHOLD such that:
+# Σ ⌈(drawnShares_i × riskPremium_i) / PERCENTAGE_FACTOR⌉ ≤ ⌈((Σ drawnShares_i) × RISK_PREMIUM_THRESHOLD) / PERCENTAGE_FACTOR⌉
```
avniculae, line 13: is this correct? it should be `(drawnShares_i × riskPremium_i + PERCENTAGE_FACTOR - 1) / PERCENTAGE_FACTOR`, no?

src/spoke/Spoke.sol
```diff
@@ -812,7 +813,10 @@ abstract contract Spoke is
     }
 
     if (debtValueLeftToCover < accountData.totalDebtValue) {
-      accountData.riskPremium /= accountData.totalDebtValue.uncheckedSub(debtValueLeftToCover);
+      accountData.riskPremium = Math.ceilDiv(
```
avniculae, line 827: todo: change to divUp from MathUtils (introduced in #1134 )

tests/misc/risk_premium_threshold_overshoot.py
DhairyaSethi: `⌈` means `ceil`

src/spoke/Spoke.sol
DhairyaSethi: ehh kinda pointless to rewrite oz functions (for this and `min`)

tests/misc/risk_premium_weighted_average_bound.py
DhairyaSethi: this was just to generally prove the wavg property; and yea the script is just a sanity check on the `divUp`. i'll adapt it to premium stuff more specifically

> when premium is MAX_KEY, there is no rounding error

yes exactly; which is why its correct but im putting it up for redundancy in the next commit

tests/misc/risk_premium_weighted_average_bound.py
DhairyaSethi: it was supposed to be general, can adapt it to be more specific

src/spoke/Spoke.sol
avniculae: then let's drop mulDivUp/Down from MathUtils as well, wdyt?

tests/misc/risk_premium_threshold_overshoot.py
```diff
@@ -0,0 +1,45 @@
...
+
+# minimize LHS using the property ⌈x⌉ ≤ x + 1, and maximize RHS using the property ⌈x⌉ ≥ x
+# Σ ((drawnShares_i × riskPremium_i) / PERCENTAGE_FACTOR + 1) ≤ ((Σ drawnShares_i) × RISK_PREMIUM_THRESHOLD) / PERCENTAGE_FACTOR
+
+# for N as number of users, the above simplifies to:
+# (1/PERCENTAGE_FACTOR) × Σ (drawnShares_i × riskPremium_i) + N ≤ (RISK_PREMIUM_THRESHOLD / PERCENTAGE_FACTOR) × Σ drawnShares_i
+
+# worst case occurs when riskPremium_i = MAX_COLLATERAL_RISK for all i and drawnShares_i = 1 for all i (to maximize N)
```
avniculae, line 21: wdym `and drawnShares_i = 1 for all i (to maximize N) ` ?

src/spoke/Spoke.sol
DhairyaSethi: not the same. those are 256 bit variants, oz only exposes 512 bit versions for mulDiv ops, we added the 256 bit ones in MathUtils for gas savings

tests/misc/risk_premium_threshold_overshoot.py
DhairyaSethi: we want to maximise N/D from the equation above, in the worst case this is when each user has the minimal drawn share (1)

tests/misc/risk_premium_threshold_overshoot.py
```diff
@@ -0,0 +1,45 @@
...
+# which implies: Σ drawnShares_i = N, and Σ (drawnShares_i × riskPremium_i) = N × MAX_COLLATERAL_RISK
+# substituting these values gives:
+# (1/PERCENTAGE_FACTOR) × (N × MAX_COLLATERAL_RISK) + N ≤ (RISK_PREMIUM_THRESHOLD / PERCENTAGE_FACTOR) × N
+
+# MAX_COLLATERAL_RISK + PERCENTAGE_FACTOR ≤ RISK_PREMIUM_THRESHOLD
+
+RISK_PREMIUM_THRESHOLD = MAX_COLLATERAL_RISK + PERCENTAGE_FACTOR
```
avniculae, line 28: I think this is correct for MAX_COLLATERAL_RISK. but say we know spoke does not use risk premium more than 15%, then I think it would be good to know what can be the min risk premium threshold we should use. I still think formula should be `roundUp(MAX_RISK / PERCENTAGE_FACTOR) * PERCENTAGE_FACTOR`, wdyt?

tests/misc/risk_premium_threshold_overshoot.py
yan-man: what's D?

DhairyaSethi: sum of drawn shares

tests/misc/risk_premium_threshold_overshoot.py
yan-man: but if we know a spoke does not use a RP > 15%, then MAX_COLLATERAL_RISK just equals 15% in that instance right?

tests/misc/risk_premium_threshold_overshoot.py
```diff
@@ -0,0 +1,45 @@
...
+
+# for N as number of users, the above simplifies to:
+# (1/PERCENTAGE_FACTOR) × Σ (drawnShares_i × riskPremium_i) + N ≤ (RISK_PREMIUM_THRESHOLD / PERCENTAGE_FACTOR) × Σ drawnShares_i
+
+# worst case occurs when riskPremium_i = MAX_COLLATERAL_RISK for all i and drawnShares_i = 1 for all i (to maximize N)
+# which implies: Σ drawnShares_i = N, and Σ (drawnShares_i × riskPremium_i) = N × MAX_COLLATERAL_RISK
+# substituting these values gives:
+# (1/PERCENTAGE_FACTOR) × (N × MAX_COLLATERAL_RISK) + N ≤ (RISK_PREMIUM_THRESHOLD / PERCENTAGE_FACTOR) × N
+
+# MAX_COLLATERAL_RISK + PERCENTAGE_FACTOR ≤ RISK_PREMIUM_THRESHOLD
```
yan-man, line 26: algebraically this proof makes sense to me

tests/misc/risk_premium_threshold_overshoot.py
DhairyaSethi: the proof does not utilise the fact that max risk is a factor of percentage factor for any of the simplifications and it works even if we make max_risk symbolic

tests/misc/risk_premium_threshold_overshoot.py
yan-man: maybe would be helpful in hubConfigurator where you can `updateSpokeRiskPremiumThreshold` based on the `MAX_COLLATERAL_RISK` value input, and then internally it will calc the `RISK_PREMIUM_THRESHOLD` value to use (ie applying this formula)

src/spoke/Spoke.sol
```diff
@@ -7,6 +7,7 @@ import {SafeERC20, IERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
 import {IERC20Permit} from 'src/dependencies/openzeppelin/IERC20Permit.sol';
 import {ReentrancyGuardTransient} from 'src/dependencies/openzeppelin/ReentrancyGuardTransient.sol';
 import {AccessManagedUpgradeable} from 'src/dependencies/openzeppelin-upgradeable/AccessManagedUpgradeable.sol';
+import {Math} from 'src/dependencies/openzeppelin/Math.sol';
```
```suggestion
import {Math} from 'src/dependencies/openzeppelin/Math.sol';
import {AccessManagedUpgradeable} from 'src/dependencies/openzeppelin-upgradeable/AccessManagedUpgradeable.sol';
```
sorry

tests/misc/risk_premium_threshold_overshoot.py
```diff
@@ -0,0 +1,45 @@
...
+MAX_COLLATERAL_RISK = IntVal(1000_00)
+
+def percentMulUp(value, percentage):
+    return (value * percentage + PERCENTAGE_FACTOR - 1) / PERCENTAGE_FACTOR
+
+# ∀ drawnShares_i ≥ 1, ∀ riskPremium_i ≤ MAX_COLLATERAL_RISK,
+# we want to minimize RISK_PREMIUM_THRESHOLD such that:
+# Σ ⌈(drawnShares_i × riskPremium_i) / PERCENTAGE_FACTOR⌉ ≤ ⌈((Σ drawnShares_i) × RISK_PREMIUM_THRESHOLD) / PERCENTAGE_FACTOR⌉
+
+# minimize LHS using the property ⌈x⌉ ≤ x + 1, and maximize RHS using the property ⌈x⌉ ≥ x
```
DhairyaSethi:
```suggestion
# maximize LHS using the property ⌈x⌉ ≤ x + 1, and minimize RHS using the property ⌈x⌉ ≥ x
```

PR #1143, opened 2026-01-26
fix: revert back to 512 bit shares math

DhairyaSethi:
reverts https://github.com/aave/aave-v4/pull/729 given it is much safer to avoid overflows given hub is immutable; almost all impls use 512 bit arithmetic for share price calcs (oz vault, euler etc); it's more useful for max uint conversions to not revert


src/hub/libraries/SharesMath.sol
yan-man, line 1: why do you want to do this?

DhairyaSethi: much safer to avoid overflows given hub is immutable 
almost all impl use 512 bit arithmetic for share price calcs (oz vault, euler etc) 
it's more useful for max uint conversions to not revert 

i suggested removing this before & moving to 256 bit arithmetic (see linked pr & its issue in the description) after analysing overflow bounds to save gas 

i think the gas bandwidth given by via ir means we can afford to revert this minor gas opt

foundry.toml
```diff
@@ -14,13 +14,13 @@ gas_limit = 1099511627776
 dynamic_test_linking = true
 
 additional_compiler_profiles = [
-  { name = "hub", optimizer = true, via_ir = true, optimizer_runs = 22_250 },
```
Kogaroshi: reducing runs to 21k + maths lib change doesn't seem to impact gas costs too much + large codesize buffer for Hub.

src/hub/libraries/SharesMath.sol
Kogaroshi: I agree, feels safer, and gas wise the impact is minimal

foundry.toml
DhairyaSethi: yea i was surprised at the minimal gas impact lol; this code size bandwidth also allows us to add `ExtSload` again for consistency (even tho it does not provide value bc all storage already has getters)

src/hub/libraries/SharesMath.sol
yan-man: feels like a lifetime ago

foundry.toml
```diff
@@ -14,13 +14,13 @@ gas_limit = 1099511627776
 dynamic_test_linking = true
 
 additional_compiler_profiles = [
-  { name = "hub", optimizer = true, via_ir = true, optimizer_runs = 21_000 },
+  { name = "hub", optimizer = true, via_ir = true, optimizer_runs = 22_300 },
```
Kogaroshi, line 17: nice !

PR #1144, opened 2026-01-26
rft: resize nonce & key

DhairyaSethi:
update nonce-key sizes from 192, 64 bits to 160, 96 bits

Field | Before | After
-- | -- | --
Key | 192 bits, ≈ 6.26e57 | 160 bits, ≈ 1.46e48
Nonce | 64 bits, ≈ 1.84e18 | 96 bits, ≈ 7.922e28

this is done in an effort to improve resilience on fixed namespace usage patterns (for ex permit is restricted to one namespace (key) in #1029 in order to have a simpler implementation)


miguelmtzinf:
Although we could increase the nonce size without issues, we prefer to keep the code because it is impractical to ever overflow (given how large 2^64 nonce space is) and to stay completely aligned with standards like ERC-4337.

PR #1149, opened 2026-01-27
debug: Try virtual fee shares

CheyenneAtapour:


src/hub/interfaces/IHub.sol
```diff
@@ -357,11 +357,11 @@ interface IHub is IHubBase, IAccessManaged {
...
 
-  /// @notice Returns the accrued fees for the asset, expressed in asset units.
-  /// @dev Accrued fees are excluded from total added assets.
+  /// @notice Returns the pending fee shares for the asset.
+  /// @dev Pending fee shares are included in total shares for share price calculation.
   /// @param assetId The identifier of the asset.
-  /// @return The amount of accrued fees.
-  function getAssetAccruedFees(uint256 assetId) external view returns (uint256);
+  /// @return The amount of pending fee shares.
+  function getAssetPendingFeeShares(uint256 assetId) external view returns (uint256);
```
Kogaroshi, line 370: might be interesting to keep a getter for the fees in assets directly (tho the conversion can be made externally from those shares ofc)

yan-man: were you thinking to keep both, ie getter for fees in assets and in shares

snapshots/Hub.Operations.json
yan-man, line 1: wow omg thought we had a lot of hub space, cant believe hub is the one exceeding size

src/hub/Hub.sol
```diff
@@ -782,10 +780,13 @@ contract Hub is IHub, AccessManaged {
     SpokeData storage feeReceiverSpoke = _spokes[assetId][feeReceiver];
     require(feeReceiverSpoke.active, SpokeNotActive());
 
+    // Convert shares to assets for the event
+    uint256 assets = asset.toAddedAssetsDown(shares);
```
yan-man, line 792: think we can rm the comment, and keep this var as `fees` as before

src/hub/Hub.sol
```diff
@@ -592,9 +592,8 @@ contract Hub is IHub, AccessManaged {
   }
 
   /// @inheritdoc IHub
-  function getAssetAccruedFees(uint256 assetId) external view returns (uint256) {
-    Asset storage asset = _assets[assetId];
-    return asset.realizedFees + asset.getUnrealizedFees(asset.getDrawnIndex());
+  function getAssetPendingFeeShares(uint256 assetId) external view returns (uint256) {
+    return _assets[assetId].pendingFeeShares;
```
yan-man, line 604: should this also add unrealized fees? `_assets[assetId].toAddedAssetsDown(_assets[assetId].getUnrealizedFees(asset.getDrawnIndex()))`

CheyenneAtapour:
Closing this as we are proceeding with an approach to call mintFeeShares periodically and leave this logic outside of the protocol

PR #1154, opened 2026-01-27
feat: separate fee accounting pro rata assets

CheyenneAtapour:
Implements an approach to unrealized fee shares, which maintains separate accrual accounting for suppliers vs treasury (borrowed fees)


src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +212,29 @@ library AssetLogic {
...
+    // fees = protocol's cut of the delta
+    uint256 fees = delta.percentMulDown(liquidityFee);
+
+    // interest = supplier's cut of the delta
+    uint256 interest = delta - fees;
+
+    // Calculate how much fees are borrowed
+    uint256 realizedFees = asset.realizedFees;
+    uint256 liquidity = asset.liquidity;
+    uint256 feesBorrowed = realizedFees > liquidity ? realizedFees - liquidity : 0;
```
miguelmtzinf: this formula is working with the assumption that there is nothing else apart from `realizedFees`, `liquidity` and `drawn`; which is wrong.
Example:
```
liquidity	200
swept		100
drawn		300
realizedFees	350
deposits   (totalAddedAssets) 200+100+300 - 350 = 600 - 350 = 250
depositsBorrowed = min(deposits, drawn) = min(250, 300) = 250
feesBorrowed = drawn - depositsBorrowed = 300 - 250 = 50 
```

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +212,29 @@ library AssetLogic {
...
+
+    // interest earned by the fees
+    uint256 drawnAmount = drawnShares.rayMulUp(previousIndex);
+    uint256 interestForFees = interest.mulDivDown(feesBorrowed, drawnAmount);
```
miguelmtzinf: denominator should be total debt (`drawn` + `premium`). Not clear to me whether we should count with `deficit` too (or `aggregatedOwedRayAfter`)

avniculae: why do we want to calculate `feesBorrowed` and then calculate `interestForFees` based on that? Interest is split across all suppliers at all time, irrespective of whether their "liquidity" was borrowed or not. Hence, I don't think we should calculate `feesBorrowed`. What do you think of the following:

```
uint256 unmintedFeeShares = asset.toAddedSharesDown(realizedFees);
uint256 interestForFees = interest.mulDivDown(unmintedFeeShares, asset.suppliedShares + unmintedFeeShares)
````

Comment (miguelmtzinf) [reply src/hub/libraries/AssetLogic.sol]: aren't they equivalent? but this is a good idea: instead of minting fee shares in every accrual, we accrue to unrealizedFees in form of assets rather than increasing the shares part of the fee receiver  
simpler and easy to understand imo

Comment (miguelmtzinf) [reply src/hub/libraries/AssetLogic.sol]: as pointed out by @avniculae, not equivalent as this would be fairer: treating `realizedFees` as they were converted into shares, effectively being part of the suppliers, leading to distributing interests evenly across all suppliers added shares.
before, we were prioritizing users' deposits to be lent out first.

File: src/hub/libraries/AssetLogic.sol [hunk 3]
```diff
@@ -210,10 +212,25 @@ library AssetLogic {
...
+
+    // fees = protocol's cut of the delta
+    uint256 fees = delta.percentMulDown(liquidityFee);
+
+    // interest = supplier's cut of the delta
+    uint256 interest = delta - fees;
+
+    // Distribute interestForFees pro-rata to realizedFees
+    uint256 realizedFees = asset.realizedFees;
+    uint256 totalAssetsBefore = asset.liquidity + asset.swept + aggregatedOwedRayBefore.fromRayUp();
```
yan-man: i think we should rename this to totalAddedAssetsBefore 

```suggestion
    uint256 totalAddedAssetsBefore = asset.liquidity + asset.swept + aggregatedOwedRayBefore.fromRayUp() - realizedFees;
```

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +212,25 @@ library AssetLogic {
...
+    uint256 interestForFees = interest.mulDivDown(realizedFees, totalAssetsBefore);
```
yan-man:
```suggestion
    uint256 interestForFees = interest.mulDivDown(realizedFees, totalAddedAssetsBefore + realizedFees);
```

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,573 @@
...
+    });
+
+    uint256 accruedFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
+
+    // No fees should accrue
+    assertEq(accruedFees, 0, 'fees should be 0 when liquidityFee is 0');
+  }
+
+  /// @dev Test fee accrual with swept funds
+  function test_unrealizedFees_withSweptFunds() public {
```
yan-man: do you have a test where realizedFees > 0

src/hub/libraries/AssetLogic.sol
CheyenneAtapour: I thought we didn't want to do this because it was cleaner as is

src/hub/libraries/AssetLogic.sol
CheyenneAtapour: I thought it was simpler to leave as is, since this realizedFees part will get canceled out

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: Added: test_unrealizedFees_withPriorRealizedFees

src/hub/libraries/AssetLogic.sol
avniculae: realizedFees should not be added here since they are already part of `totalAssetsBefore` -- note that `totalAssetsBefore` is a bit confusing here, since it does not contain the subtraction of realizedFees. 

However, the denominator should contain `+ VIRTUAL_ASSETS`

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +212,25 @@ library AssetLogic {
       drawnIndex: previousIndex
     });
 
-    return
-      (aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp()).percentMulDown(
-        liquidityFee
-      );
+    // delta = total growth in aggregated owed
+    uint256 delta = aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp();
+    if (delta == 0) {
```
miguelmtzinf: this is done before already, by comparing indexes.

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,1203 @@
...
+      addSpoke: address(spoke1),
+      addAmount: SUPPLY_AMOUNT,
+      drawUser: bob,
+      drawSpoke: address(spoke1),
+      drawAmount: BORROW_AMOUNT,
+      skipTime: 365 days
+    });
+
+    // Get accrued fees
+    uint256 accruedFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
```
miguelmtzinf: let's assert this is `getAssetAccruedFees`.
also, let's calculate and assert how much interest is generated.
let's assert how much user position grew, and how much accrued fees grew.
let's do this programtically, and would like to have some scenarios with exact numbers too.

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,1203 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import 'tests/unit/Hub/HubBase.t.sol';
+
+contract HubUnrealizedFeesTest is HubBase {
```
miguelmtzinf:
```suggestion
contract HubAccruedFeesTest is HubBase {
```

src/hub/libraries/AssetLogic.sol
avniculae: Note that we can achieve an equivalence with the changes from [4d8a205](https://github.com/aave/aave-v4/pull/1154/commits/4d8a205857fdc2e34ccbc1f19ecb9d785b7cc7b9), as long as we do this change https://github.com/aave/aave-v4/pull/1154#discussion_r2741094966

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,1203 @@
...
+    uint256 previousFees = 0;
+
+    // Accrue over multiple periods
+    for (uint256 i = 0; i < 4; i++) {
+      skip(90 days);
+
+      uint256 currentFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
+
+      // Fees should increase each period
+      assertGt(currentFees, previousFees, 'fees should increase over time');
```
miguelmtzinf: this is very vague, why not eq?

miguelmtzinf: what about a test where we have 1 user with the same amount of added assets than treasury, interests should grow at the same pace. correct?

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,1203 @@
...
+
+    // Verify fees were accrued even with deficit
+    assertGt(accruedFees, 0, 'fees should accrue with deficit');
+
+    // Verify deficit is tracked
+    assertGt(hub1.getAsset(daiAssetId).deficitRay, 0, 'deficit should be tracked');
+  }
+
+  /// @dev Test fee accrual with premium debt present
+  function test_unrealizedFees_withPremiumDebt() public {
```
miguelmtzinf: we can have a combined test, fuzzing deficit, swept, drawn and premium
and a combined test with exact numbers

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,1203 @@
...
+    // Therefore fees should be exactly 0
+    assertEq(accruedFees, 0, 'fees should be exactly 0 due to rounding');
+
+    // But interest still exists, so share price should increase (suppliers get everything)
+    assertGt(delta, 0, 'delta should be > 0');
+    assertGt(finalSharePrice, initialSharePrice, 'share price should increase');
+  }
+
+  /// @dev Test that share price never decreases
+  function test_unrealizedFees_sharePriceNeverDecreases() public {
```
miguelmtzinf: don't we have thsi test somewhere else already?

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,1203 @@
...
+    // Get accrued fees
+    uint256 accruedFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
+
+    // Verify fees were accrued in complex scenario
+    assertGt(accruedFees, 0, 'fees should accrue in combined scenario');
+
+    // Verify all state is tracked correctly
+    IHub.Asset memory asset = hub1.getAsset(daiAssetId);
+    assertEq(asset.swept, sweepAmount, 'swept should be tracked');
+    assertGt(asset.deficitRay, 0, 'deficit should be tracked');
```
miguelmtzinf: why gt?

tests/unit/Hub/Hub.UnrealizedFees.t.sol
miguelmtzinf, line 1: please point me to the test where
- we check that treasury fees receives the same amount of interest that any other user (per share)
- users receive same amount of interest no matter how big is realizedFees
- supply share price does not increase sharply if realized fees are not minted

src/hub/libraries/AssetLogic.sol
yan-man: @CheyenneAtapour depends how explicit we want to be, also your suggestion of adding a comment would be helpful too. But think it's weird to just have it without realizedFees bc we have it in the totalAddedAssets() method normally. and yeah including virtual assets as well like you did in shares approach

src/hub/libraries/AssetLogic.sol
yan-man: or add comment

but I do think we should rename to `totalAddedAssetsBefore` for consistency

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,1203 @@
...
+    // getAssetAccruedFees returns realizedFees + unrealizedFees
+    // So it should equal realizedFeesAfterFirst + expectedNewUnrealizedFees
+    uint256 totalAccruedFees = hub1.getAssetAccruedFees(daiAssetId);
+    uint256 expectedTotalAccruedFees = realizedFeesAfterFirst + expectedNewUnrealizedFees;
+
+    // Verify the formula
+    assertApproxEqAbs(
+      totalAccruedFees,
+      expectedTotalAccruedFees,
+      2,
```
yan-man: hm is the 2 wei expected? this seems a bit high for one accrual

tests/unit/Hub/Hub.UnrealizedFees.t.sol
yan-man: bc deficit is reported in L492

src/hub/libraries/AssetLogic.sol
yan-man: hm but it is possible that, ex. drawnIndex = previousIndex + 1, and the rounded fromRayUp values match so delta == 0, no?

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +212,25 @@ library AssetLogic {
       drawnIndex: previousIndex
     });
 
-    return
-      (aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp()).percentMulDown(
-        liquidityFee
-      );
+    // delta = total growth in aggregated owed
+    uint256 delta = aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp();
```
yan-man: delta is generic as well, maybe `totalGrowth`?

tests/unit/Hub/Hub.UnrealizedFees.t.sol
```diff
@@ -0,0 +1,1203 @@
...
+
+    // Calculate interest for suppliers (the rest of interest after fees take their share)
+    uint256 interestForSuppliers = interest - expectedInterestForFees;
+
+    // Verify accounting: interestForFees + interestForSuppliers = interest (exact)
+    assertEq(
+      expectedInterestForFees + interestForSuppliers,
+      interest,
+      'interest split should be exact'
+    );
```
yan-man: dont think this assertion is helpful as it exactly falls out from L872

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1247 @@
...
+    hub1.reportDeficit(daiAssetId, 30e18, ZERO_PREMIUM_DELTA);
+
+    skip(180 days);
+
+    uint256 realizedFees = hub1.getAsset(daiAssetId).realizedFees;
+    uint256 expectedAccruedFees = realizedFees + _calcUnrealizedFees(hub1, daiAssetId);
+    uint256 accruedFees = hub1.getAssetAccruedFees(daiAssetId);
+
+    assertApproxEqAbs(accruedFees, expectedAccruedFees, 2);
+    assertGt(accruedFees, 0);
```
yan-man, line 202: dont really think we need this assertion

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1247 @@
...
+
+    IHub.Asset memory asset = hub1.getAsset(daiAssetId);
+    assertEq(asset.swept, 100e18);
+    assertGt(asset.deficitRay, 0);
```
yan-man: can this be exact assertion on deficitRay == 30e18 RAY

src/hub/libraries/AssetLogic.sol
CheyenneAtapour: I changed to totalAddedAssetsBefore

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: Adjusted

src/hub/libraries/AssetLogic.sol
CheyenneAtapour: Changed

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: Removed

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1247 @@
...
+      skipTime: 365 days
+    });
+
+    uint256 accruedFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
+    uint256 drawnDebt = getAssetDrawnDebt(daiAssetId);
+    uint256 totalInterest = drawnDebt - 10;
+
+    assertGt(drawnDebt, 10);
+    assertLe(accruedFees, totalInterest);
+    assertApproxEqAbs(accruedFees + (totalInterest - accruedFees), totalInterest, 2);
```
yan-man: dont think this is helpful, it's a direct equality as accruedFees cancel and it's just totalInterest == totalInterest

yan-man: https://github.com/aave/aave-v4/pull/1161

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1152 @@
...
+
+    Utils.draw({hub: hub1, assetId: daiAssetId, to: bob, caller: address(spoke1), amount: 100});
+    skip(1 hours);
+
+    (uint256 drawnDebt, ) = hub1.getAssetOwed(daiAssetId);
+    uint256 delta = drawnDebt - 100;
+    uint256 accruedFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
+    uint256 finalSharePrice = hub1.previewAddByShares(daiAssetId, 1e18);
+
+    assertEq((delta * 1_00) / 100_00, 0);
```
yan-man: did you not want to use percentMulDown specifically here? at least should prob be PercentageMath.PERCENTAGE_FACTOR for clarity

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: I think it's fine, just a sanity check to show accrued fees were non-zero

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: 1. test_unrealizedFees_equalAssetsEqualGrowth
2. testFuzz_unrealizedFees_interestIndependentOfFeeSize
3. testFuzz_unrealizedFees_gradualSharePriceGrowth

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: We don't necessarily have an explicit test like this, but we try to verify this is the case by utilizing a helper (_checkSupplyRateIncreasing) in some withdraw tests

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: Done with: 
* testFuzz_unrealizedFees_combinedScenario
* test_unrealizedFees_combinedScenario_exactNumbers

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: This one was just saying that the fees were growing, but it's perhaps trivial, so I removed it. I added the latter test: test_unrealizedFees_equalAssetsEqualGrowth

src/hub/libraries/AssetLogic.sol
CheyenneAtapour: I renamed to totalAddedAssetsBefore

src/hub/libraries/AssetLogic.sol
CheyenneAtapour: This has since been updated

tests/unit/Hub/Hub.UnrealizedFees.t.sol
CheyenneAtapour: Did it, I just added a couple tests which call the fuzzed one, hope that works

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +211,29 @@ library AssetLogic {
...
+    uint256 totalGrowth = aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp();
+    if (totalGrowth == 0) {
+      return 0;
+    }
+
+    // fees = protocol's cut of the total growth
+    uint256 fees = totalGrowth.percentMulDown(liquidityFee);
+
+    // interest = supplier's cut of the total growth
+    uint256 interest = totalGrowth - fees;
```
```suggestion
    uint256 growthDelta = aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp();
    uint256 unrealizedFees = growthDelta.percentMulDown(liquidityFee);
```
short-circuit happens already by comparing previous and current indexes.

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +211,29 @@ library AssetLogic {
...
+    uint256 totalAddedAssetsBefore = asset.liquidity +
+      asset.swept +
+      aggregatedOwedRayBefore.fromRayUp();
+    uint256 interestForFees = interest.mulDivDown(
+      realizedFees,
+      totalAddedAssetsBefore + SharesMath.VIRTUAL_ASSETS
+    );
+
+    // Total unrealized fees = protocol fee cut + interest earned by fee portion
+    return fees + interestForFees;
```
```suggestion
    // distribute the remaining interest in proportion to the `realizedFees` relative to total amount of assets that belong to
    // suppliers, effectively as if the `realizedFees` were minted as added shares
    uint256 suppliersTotalAddedAssets = asset.liquidity +
      asset.swept +
      aggregatedOwedRayBefore.fromRayUp();
    uint256 realizedFeesInterest = (growDelta - fees).mulDivDown(
      asset.realizedFees,
      suppliersTotalAddedAssets + SharesMath.VIRTUAL_ASSETS
    );

    return unrealizedFees + realizedFeesInterest;
```

tests/Base.t.sol
```diff
@@ -2942,22 +2958,47 @@ abstract contract Base is Test {
...
     uint256 aggregatedOwedRayBefore = (((uint256(asset.drawnShares) + asset.premiumShares) *
       previousIndex).toInt256() - asset.premiumOffsetRay).toUint256() + asset.deficitRay;
 
-    return
-      (aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp()).percentMulDown(
-        asset.liquidityFee
-      );
+    // delta = total growth in aggregated owed
+    uint256 totalGrowth = aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp();
+    if (totalGrowth == 0) {
```
miguelmtzinf: reminder to update this logic based on src

src/hub/libraries/AssetLogic.sol
yan-man: https://github.com/aave/aave-v4/pull/1154#discussion_r2742224717

tests/Base.t.sol

src/hub/libraries/AssetLogic.sol

src/hub/libraries/AssetLogic.sol

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1190 @@
...
+    (uint256 drawnDebt, ) = hub1.getAssetOwed(daiAssetId);
+    uint256 totalInterest = drawnDebt - borrowAmount;
+    uint256 finalTotalAssets = hub1.getAddedAssets(daiAssetId);
+
+    assertEq(accruedFees, 0);
+    assertEq(accruedFees, _calcUnrealizedFees(hub1, daiAssetId));
+    assertEq(finalTotalAssets - initialTotalAssets, totalInterest);
+  }
+
+  function testFuzz_unrealizedFees_maxFeeAllToTreasury(
```
yan-man: can you add some natspec description to these tests that lack it

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +211,21 @@ library AssetLogic {
       drawnIndex: previousIndex
     });
 
-    return
-      (aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp()).percentMulDown(
-        liquidityFee
-      );
+    // Take treasury cut, remainder is interest to distribute
```
miguelmtzinf: i d rm (last part is already said later)

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +211,21 @@ library AssetLogic {
...
+    uint256 growthDelta = aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp();
+    uint256 unrealizedFees = growthDelta.percentMulDown(liquidityFee);
+
+    // distribute the remaining interest in proportion to the `realizedFees` relative to total amount of assets that belong to
+    // suppliers, effectively as if the `realizedFees` were minted as added shares
```
```suggestion
    // distribute the remaining interest in proportion to the `realizedFees` relative to total amount of assets
    // that belong to suppliers, effectively as if the `realizedFees` were minted as added shares
```
just so it's under 120 char limit

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Done

src/hub/libraries/AssetLogic.sol

src/hub/libraries/AssetLogic.sol

tests/unit/Hub/HubBase.t.sol
```diff
@@ -3,11 +3,13 @@
 pragma solidity ^0.8.0;
 
 import 'tests/Base.t.sol';
+import {WadRayMath} from 'src/libraries/math/WadRayMath.sol';
```
yan-man: not sure if this is my fault, but we dont need this import

tests/unit/Hub/HubBase.t.sol
```diff
@@ -227,4 +229,18 @@ contract HubBase is Base {
   function _randomInvalidAssetId(IHub hub) internal returns (uint256) {
     return vm.randomUint(hub.getAssetCount(), UINT256_MAX);
   }
+
+  function _calcTotalAddedAssets(IHub.Asset memory asset) internal pure returns (uint256) {
```
yan-man: we're calling this `suppliersTotalAddedAssets` in AssetLogic now, so this should be `_calcSuppliersTotalAddedAssets`

yan-man: can you also add natspec above clarifying how this is different than regular totalAddedAssets

src/hub/libraries/AssetLogic.sol
miguelmtzinf: I do not think so, because we round up in both sides.
Quick proof:
```
aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp()
=
drawnShares * drawnIndexAfter + premiumShares * drawnIndexAfter - offset + deficit
-
drawnShares  * drawnIndexBefore + premiumShares * drawnIndexBefore - offset + deficit

Given that drawnIndexAfter = drawnIndexBefore + 1:

aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp()
=
drawnShares * (drawnIndex + 1) + premiumShares * (drawnIndex + 1) - offset + deficit
-
drawnShares * drawnIndex + premiumSHares * drawnIndex - offset + deficit
=
drawnShares * drawnIndex + drawnShares + premiumShares * drawnIndex + premiumShares - offfset + deficit
-
drawnShares * drawnIndex + premiumShares * drawnIndex - offset + deficit
=
drawnShares + premiumShares
```

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+
+    assertEq(delta.percentMulDown(liquidityFee), 0);
+    assertEq(accruedFees, 0);
+    assertEq(hub1.getAssetAccruedFees(daiAssetId), 0);
+    assertGt(delta, 0);
+    _checkSupplyRateIncreasing(initialSharePrice, finalSharePrice, 'share price');
+  }
+
+  /// @dev Verifies that accrued fees themselves earn interest over time
+  function test_unrealizedFees_feesEarnInterest() public {
```
miguelmtzinf: it's not accrued fees what earns interest here, but minted fee shares
is the intention of this test to showcase the fact that accrued fees are earning interests even without becoming added shares?

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+
+    uint256 drawnIndex = hub1.getAssetDrawnIndex(daiAssetId);
+    uint256 liquidityFee = hub1.getAssetConfig(daiAssetId).liquidityFee;
+
+    uint256 expectedTotalInterest = drawnIndex.rayMulUp(initialDrawnDebt) -
+      initialDrawnIndex.rayMulUp(initialDrawnDebt);
+
+    assertEq(expectedTotalInterest, totalInterest);
+    assertEq(drawnDebt, initialDrawnDebt + expectedTotalInterest);
+    assertEq(accruedFees, totalInterest.percentMulDown(liquidityFee));
```
miguelmtzinf: if we are forcing the case wher 1 wei is the total interest, we can use exact numbers for accruedFees and drawnDebt (0 delta),
so i d remove generic calculations

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    uint256 accruedFees = hub1.getAssetAccruedFees(daiAssetId);
+
+    assertEq(accruedFees, expectedAccruedFees);
+    assertGt(accruedFees, 0);
+
+    IHub.Asset memory asset = hub1.getAsset(daiAssetId);
+    assertEq(asset.swept, swept);
+    assertEq(asset.deficitRay, expectedDeficitRay);
+    assertEq(asset.drawnShares, expectedDrawnShares);
+  }
```
miguelmtzinf, line 208: what's the point of this test? just creating this complex scenario for running the same logic in both sides (src and test have the same fn to calculate accrued fees)

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+      vm.prank(address(spoke1));
+      hub1.reportDeficit(daiAssetId, deficitAmount, ZERO_PREMIUM_DELTA);
+    }
+
+    skip(timeSkip);
+
+    uint256 realizedFees = hub1.getAsset(daiAssetId).realizedFees;
+    uint256 expectedAccruedFees = realizedFees + _calcUnrealizedFees(hub1, daiAssetId);
+    uint256 accruedFees = hub1.getAssetAccruedFees(daiAssetId);
+    assertEq(accruedFees, expectedAccruedFees);
```
miguelmtzinf, line 263: same as above

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+  }
+
+  /// @dev Verifies fees >= protocol cut and share price increases
```
miguelmtzinf: i think the inequality is wrong. also, what's fees and what's protocol cut?

miguelmtzinf: if this is an invariant, does it make sense to do n iterations?

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    uint256 protocolCut = totalDelta.percentMulDown(20_00);
+    assertGe(accruedFees, protocolCut);
+    _checkSupplyRateIncreasing(
+      initialSharePrice,
+      hub1.previewAddByShares(daiAssetId, 1e18),
+      'share price'
+    );
+  }
+
+  /// @dev Tests 90% fee with 100% utilization, verifies 90/10 split between fees and suppliers
```
miguelmtzinf: i thought this was doing 90/10 split in terms of assets of suppliers/treasury. Do we have this test also? We can have both as fuzz instead.

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+      sharePriceBefore,
+      hub1.previewAddByShares(daiAssetId, 1e18),
+      'share price'
+    );
+
+    uint256 supplierAssetsBefore = hub1.previewRemoveByShares(
+      daiAssetId,
+      hub1.getSpokeAddedShares(daiAssetId, address(spoke1))
+    );
+    Utils.mintFeeShares(hub1, daiAssetId, ADMIN);
```
miguelmtzinf: why do we need to mintFeeShares now?

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    uint256 supplierAssetsAfter = hub1.previewRemoveByShares(
+      daiAssetId,
+      hub1.getSpokeAddedShares(daiAssetId, address(spoke1))
+    );
+    assertEq(supplierAssetsAfter, supplierAssetsBefore);
+  }
+
+  /// @dev Tests 5 years of fee accumulation, verifies fees grow and mint doesn't dilute suppliers
```
miguelmtzinf: can you point me how we are checking that there is no dilution? If it's by minting fee shares at the end, that's obvious no? and belongs to mintFeeShares test more than here.

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+
+    uint256 supplierSharesBefore = hub1.getSpokeAddedShares(daiAssetId, address(spoke1));
+    uint256 sharePriceBefore = hub1.previewAddByShares(daiAssetId, 1e18);
+    Utils.mintFeeShares(hub1, daiAssetId, ADMIN);
+
+    assertEq(hub1.getSpokeAddedShares(daiAssetId, address(spoke1)), supplierSharesBefore);
+    assertEq(hub1.previewAddByShares(daiAssetId, 1e18), sharePriceBefore);
+  }
+
+  /// @dev Verifies exact fee formula: first accrual fees = protocolCut when no prior fees
```
miguelmtzinf, line 351: let's do fuzz pls. also, wondering if better to remove `protocolCut` term across the code, and here just be explicit: `unrealizedFees only comes from liquidityFee` or smt

miguelmtzinf: also, what does `preciseCalculation` mean

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+        assertGt(expectedInterestForFees, 0);
+      }
+    }
+    uint256 interestForSuppliers = interest - expectedInterestForFees;
+    assertEq(totalAccruedFees - realizedFeesAfterFirst, expectedNewUnrealizedFees);
+    assertEq(protocolCut + expectedInterestForFees + interestForSuppliers, delta);
+  }
+
+  /// @dev When user assets equal treasury's fee shares, both should earn interest at same rate
+  function test_unrealizedFees_equalAssetsEqualGrowth() public {
```
miguelmtzinf: i d rename to signify that interest is distributed based on assets proportion

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    assertGt(carolGrowth, 0);
+    assertGt(treasuryGrowth, 0);
+    assertEq(carolGrowth, treasuryGrowth);
+    assertEq(
+      hub1.getAssetAccruedFees(daiAssetId) - hub1.getAsset(daiAssetId).realizedFees,
+      _calcUnrealizedFees(hub1, daiAssetId)
+    );
+  }
+
+  /// @dev Verifies minting fee shares mid-period doesn't affect supplier's position value
```
miguelmtzinf: is this saying that `no matter whether realized fees are minted or not, they earn at the same pace and also users are earning at the same pace`? I do think this is not completely equivalent because of the potential donation on `mintedFeeShares`, but good scenario to think about

is this what you are testing here? if so, don't really see how this is verified in the test code (after the second 180 days wait)

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    uint256 supplierAssetsAfterMint = hub1.previewRemoveByShares(daiAssetId, supplierShares);
+    assertApproxEqAbs(supplierAssetsAfterMint, supplierAssetsBeforeMint, 1);
+    assertEq(hub1.getAssetAccruedFees(daiAssetId) - hub1.getAsset(daiAssetId).realizedFees, 0);
+    assertEq(_calcUnrealizedFees(hub1, daiAssetId), 0);
+
+    skip(180 days);
+    assertGt(hub1.previewRemoveByShares(daiAssetId, supplierShares), supplierAssetsAfterMint);
+  }
+
+  /// @dev Verifies new supplier gets fair share price after large fee accumulation
```
miguelmtzinf: why is this relevant? is this testing than accruedFees earning interest does not have any impact on new suppliers? how would this be verified? what is this test doing

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    (uint256 drawnDebt, ) = hub1.getAssetOwed(daiAssetId);
+    assertGt(drawnDebt, BORROW_AMOUNT);
+
+    uint256 bobFinalValue = hub1.previewRemoveByShares(daiAssetId, bobShares);
+    uint256 aliceFinalValue = hub1.previewRemoveByShares(daiAssetId, aliceShares);
+    assertGt(bobFinalValue, bobValueAfter);
+    assertGt(aliceFinalValue, aliceRedeemable);
+  }
+
+  /// @dev Tests 90% fee with 100% utilization over 3 years, verifies mint doesn't dilute
```
miguelmtzinf: same test as above, can you point me to how are you checking no dilution happens?

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+
+    uint256 finalAccruedFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
+    assertEq(
+      finalAccruedFees - hub1.getAsset(daiAssetId).realizedFees,
+      _calcUnrealizedFees(hub1, daiAssetId)
+    );
+    assertGt(finalAccruedFees, 0);
+  }
+
+  /// @dev Fuzz: Share price growth bounded by debt growth, minting doesn't cause sharp jump
```
miguelmtzinf, line 562: let's add accruals also, so we check share price grows well while realizedFees are earning interest too

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+
+    uint256 accruedFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
+    (uint256 drawnDebt, ) = hub1.getAssetOwed(daiAssetId);
+    uint256 totalInterest = drawnDebt - borrowAmount;
+    assertEq(accruedFees, _calcUnrealizedFees(hub1, daiAssetId));
+
+    assertLe(accruedFees, totalInterest);
+    if (liquidityFee == 0) assertEq(accruedFees, 0);
+    if (liquidityFee == PercentageMath.PERCENTAGE_FACTOR && totalInterest > 0)
+      assertEq(accruedFees, totalInterest);
```
miguelmtzinf, line 675: add brackets

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+      assetId: daiAssetId,
+      to: bob,
+      caller: address(spoke1),
+      amount: borrowAmount
+    });
+
+    for (uint256 i = 0; i < 5; i++) {
+      skip(30 days);
+      uint256 currentSharePrice = hub1.previewAddByShares(daiAssetId, 1e18);
+      _checkSupplyRateIncreasing(lastSharePrice, currentSharePrice, 'share price');
```
miguelmtzinf: a bit useless without an accrual in each step imo

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+      lastSharePrice = currentSharePrice;
+      assertEq(
+        _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId),
+        _calcUnrealizedFees(hub1, daiAssetId)
```
miguelmtzinf: why/how this asset is helpful?

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+      );
+    }
+  }
+
+  /// @dev Fuzz test verifying minting fee shares preserves supplier position value
+  function testFuzz_unrealizedFees_mintPreservesSupplierValue(
```
miguelmtzinf: a bit useless test

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+
+    Utils.mintFeeShares(hub1, daiAssetId, ADMIN);
+
+    uint256 supplierValueAfter = hub1.previewRemoveByShares(daiAssetId, supplierShares);
+    assertApproxEqAbs(supplierValueAfter, supplierValueBefore, 4);
+    assertEq(_calcUnrealizedFees(hub1, daiAssetId), 0);
+  }
+
+  /// @dev Fuzz test verifying total interest equals fees plus supplier yield
+  function testFuzz_unrealizedFees_conservationOfValue(
```
miguelmtzinf: also redudant imho, i d rm

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    assertEq(accruedFees, _calcUnrealizedFees(hub1, daiAssetId));
+
+    uint256 supplierYield = hub1.getAddedAssets(daiAssetId) - supplyAmount;
+    (uint256 drawnDebt, ) = hub1.getAssetOwed(daiAssetId);
+    uint256 totalInterest = drawnDebt - borrowAmount;
+
+    assertEq(accruedFees + supplierYield, totalInterest);
+  }
+
+  /// @dev Fuzz test verifying zero fee rate directs all interest to suppliers
```
miguelmtzinf, line 679: the interesting point to check here is we are still accruing fees because of `realizedFeesInterest`, please add

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    (uint256 drawnDebt, ) = hub1.getAssetOwed(daiAssetId);
+    uint256 totalInterest = drawnDebt - borrowAmount;
+    uint256 finalTotalAssets = hub1.getAddedAssets(daiAssetId);
+
+    assertEq(accruedFees, 0);
+    assertEq(accruedFees, _calcUnrealizedFees(hub1, daiAssetId));
+    assertEq(finalTotalAssets - initialTotalAssets, totalInterest);
+  }
+
+  /// @dev Fuzz test verifying 100% fee rate directs all interest to treasury
```
miguelmtzinf, line 717: does the system work well in this case too? please check this (unrealizedFeesInterest must be 0)

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    uint256 totalInterest = drawnDebt - borrowAmount;
+    uint256 finalSharePrice = hub1.previewAddByShares(daiAssetId, 1e18);
+
+    assertEq(accruedFees, totalInterest);
+    assertEq(accruedFees, _calcUnrealizedFees(hub1, daiAssetId));
+    assertEq(finalSharePrice, initialSharePrice);
+  }
+
+  /// @dev Fuzz test verifying fee split matches configured liquidity fee percentage
+  function testFuzz_unrealizedFees_feeSplitProportional(
```
miguelmtzinf: is this the fuzz version of https://github.com/aave/aave-v4/pull/1154#discussion_r2750212592 ? `test_unrealizedFees_preciseCalculation`

tests/unit/Hub/Hub.AccruedFees.t.sol
```diff
@@ -0,0 +1,1199 @@
...
+    (uint256 drawnDebt, ) = hub1.getAssetOwed(daiAssetId);
+    uint256 totalDelta = drawnDebt - borrowAmount;
+    uint256 accruedFees = _getExpectedFeeReceiverAddedAssets(hub1, daiAssetId);
+
+    uint256 expectedProtocolCut = totalDelta.percentMulDown(liquidityFee);
+    assertEq(accruedFees, expectedProtocolCut);
+    assertEq(accruedFees, _calcUnrealizedFees(hub1, daiAssetId));
+  }
+
+  /// @dev Fuzz: Users earn same interest per share regardless of realizedFees size
```
miguelmtzinf, line 762: makes sense to test this even without minting fee shares

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Yeah, in this test there should have been an additional action to show the unrealized fees becoming realized fees, and then showing interest is accrued on those. But this is already covered by `test_unrealizedFees_fuzz_withPriorRealizedFees` so can remove this one

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Yeah, we wanted a version of this test that included all of swept, deficit, etc and ensure everything still works. But maybe we should create a helper to calc accrued fees which does not copy src and use that

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: protocol cut is just the amount from the current accrual that is given to the treasury. Fees includes this along with the amount earned as interest on unrealized fees, so the inequality is correct. 

Indeed I can do more iterations to show this

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: It's 90% treasury, 10% suppliers, and sure can do

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: We don't, removed

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: It's obvious because we are doing math to calculate the distribution as if they are already minted, but it's just to check our mechanism works properly, and that after fee shares are minted, the accounting isn't messed up. I can remove though to cut down this test suite and we can perhaps move it there if desired

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Bad name, but yeah just saying "protocolCut" matches all the accrued fees here, since there are no realizedFees which earn interest

tests/unit/Hub/HubBase.t.sol
CheyenneAtapour: Done

tests/unit/Hub/HubBase.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: This is supposed to be that suppliers' yield is not diluted out after minting fee shares, but this firstly doesn't test it well, and also we have above that minting fee shares does not dilute suppliers, so I can remove

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Done

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: This is the more correct version of this test: testFuzz_unrealizedFees_interestIndependentOfFeeSize
Will remove this one

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Ah, you mean first having non-zero liquidity fee, so we get some realizedFees, then change to 0 and show treasury is still earning the interest on realizedFees?

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: I think this test doesn't even belong here, so will remove

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Done, added occasional accruals

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: Is it sufficient to check like so? That suppliers get nothing?

tests/unit/Hub/Hub.AccruedFees.t.sol
CheyenneAtapour: The nice thing here is that the 2 users supply at the same block, so makes this comparison easier. How do you think we can do it without minting?

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +211,20 @@ library AssetLogic {
...
     });
 
-    return
-      (aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp()).percentMulDown(
-        liquidityFee
-      );
+    uint256 growthDelta = aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp();
+    uint256 unrealizedFees = growthDelta.percentMulDown(liquidityFee);
+
+    // distribute the remaining interest in proportion to the `realizedFees` relative to total amount of assets
```
avniculae, line 217: this seems a bit cumbersome, is it just me? 
what about simply `calculate interest of unminted fee shares`?

CheyenneAtapour:
Proceeding with an approach to call mintFeeShares() when it's desired to realize treasury fees

PR #1155, opened 2026-01-28
feat: Immutable limit on user collaterals and borrows

### LINKED ISSUE #1100: Unbounded number of collateral and debt reserves
Currently there is no bound on the number of collateral and debt reserves a user position can have

CheyenneAtapour:
Adds a limit on how many reserves a user can utilize as borrows or collaterals

This introduces a maximum number of reserves a user can have as collateral or as borrowed, counted separately (for example, a value of 10 allows up to 10 collateral assets and 10 borrowed reserves for each user position).

The goal is to set a value below the theoretical limit (for the sorting mechanism when computing user RP and for loops in the HF calculation), while still being high enough not to restrict normal user activity. This avoids the need to frequently update the limit (which would require upgrades), while protecting the system from users locking their positions by adding an excessive number of collateral or borrowed reserves.

Closes #1100


src/spoke/Spoke.sol
```diff
@@ -411,15 +424,24 @@ abstract contract Spoke is
...
     if (usingAsCollateral) {
+      // disabling as collateral is allowed when reserve is frozen
+      require(!flags.frozen(), ReserveFrozen());
+      // this must be a new collateral, otherwise would have short-circuited
+      // We set as collateral above in order to potentially refresh, so check <= here
+      require(
+        MAX_USER_RESERVES_LIMIT == MAX_ALLOWED_USER_RESERVES_LIMIT ||
+          positionStatus.collateralCount(_reserveCount) <= MAX_USER_RESERVES_LIMIT,
+        MaximumUserReservesExceeded()
+      );
```
DhairyaSethi: why can't this be in the validate function itself?

src/spoke/interfaces/ISpoke.sol
```diff
@@ -535,4 +540,7 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IExtSload, IMul
 
   /// @notice Returns the address of the AaveOracle contract.
   function ORACLE() external view returns (address);
+
+  /// @notice Returns the maximum allowed number of reserves per user (for both collaterals and borrows).
```
```suggestion
  /// @notice Returns the maximum allowed number of collateral and borrow reserves per user (each counted separately).
```

src/spoke/Spoke.sol
```diff
@@ -109,9 +115,11 @@ abstract contract Spoke is
 
   /// @dev Constructor.
   /// @param oracle_ The address of the AaveOracle contract.
-  constructor(address oracle_) {
+  /// @param maxUserReservesLimit_ The maximum number of reserves a user can have (both collaterals and borrows).
```
```suggestion
  /// @param maxUserReservesLimit_ The maximum number of collateral and borrow reserves a user can have.
```

src/spoke/Spoke.sol
miguelmtzinf: i suggested moving it out of validation, so it's similar to the borrow function. Maybe this is confusing and we should consider either let them be asymetric or do the logic on validation in both functions?

miguelmtzinf: the frozen flag check does not belong to here imo, i d revert that anyway

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
```diff
@@ -177,4 +177,46 @@ contract SpokeBorrowValidationTest is SpokeBase {
     vm.expectRevert(abi.encodeWithSelector(IHub.DrawCapExceeded.selector, drawCap));
     Utils.borrow(spoke1, daiReserveId, bob, 1, bob);
   }
+
+  function test_borrow_revertsWith_MaximumUserReservesExceeded() public {
+    // Verify spoke3 has a limit of 10
+    assertEq(spoke3.MAX_USER_RESERVES_LIMIT(), 10);
```
miguelmtzinf: think it's better to fetch the limit in a variable and use that below, dynamicly.
we can add an assert > 0 so we make sure we test this path.

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
```diff
@@ -177,4 +177,46 @@ contract SpokeBorrowValidationTest is SpokeBase {
...
+
+    // Bob tries to borrow from an 11th reserve - should revert due to limit of 10
+    vm.expectRevert(ISpoke.MaximumUserReservesExceeded.selector);
+    vm.prank(bob);
+    spoke3.borrow(10, 1e18, bob);
+  }
+
+  function test_borrow_unlimited_whenLimitIsMax() public {
+    // Verify that when MAX_USER_RESERVES_LIMIT is type(uint16).max, many reserves can be borrowed
+    assertEq(spoke1.MAX_USER_RESERVES_LIMIT(), type(uint16).max);
```
miguelmtzinf: add to test constants file, do not hardcode

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
miguelmtzinf, line 1: mising a test where we borrow the limit, repay 1, we have space again for 1
mising an analogous test for setUsingAsCollateral

tests/Base.t.sol
```diff
@@ -2975,4 +2985,80 @@ abstract contract Base is Test {
...
+          optimalUsageRatio: 90_00, // 90.00%
+          baseVariableBorrowRate: 5_00, // 5.00%
+          variableRateSlope1: 5_00, // 5.00%
+          variableRateSlope2: 5_00 // 5.00%
+        })
+      );
+
+      // Add asset to hub1
+      vm.startPrank(ADMIN);
+      uint256 newTokenAssetId = hub1.addAsset(
```
miguelmtzinf: does make sense to pass the hub instance to use, no?
the function name is a bit generic, not sure if we should say smt to signal the reserve is collateral/borrowable and completely activated.

src/spoke/instances/SpokeInstance.sol
```diff
@@ -14,7 +14,8 @@ contract SpokeInstance is Spoke {
   /// @dev Constructor.
   /// @dev During upgrade, must ensure that the new oracle is supporting existing assets on the spoke and the replaced oracle.
   /// @param oracle_ The address of the oracle.
-  constructor(address oracle_) Spoke(oracle_) {
+  /// @param maxUserReservesLimit_ The maximum number of reserves a user can have (both collaterals and borrows).
```
yan-man:
```suggestion
  /// @param maxUserReservesLimit_ The maximum number of collateral and debt reserves allowed in a user position (limit applies independently to each).
```

src/spoke/Spoke.sol
```diff
@@ -109,9 +115,11 @@ abstract contract Spoke is
...
+  constructor(address oracle_, uint16 maxUserReservesLimit_) {
```
yan-man, line 119: i think we should add some validation on maxUserReservesLimit_ > 0

src/spoke/instances/SpokeInstance.sol
yan-man, line 25: i know you didnt change it here, but feels weird that this event is emitted even before validation occurs. was there a reason for this?

src/spoke/Spoke.sol
yan-man: in favor of making it asymmetric, and moving frozen back into _validateSetUsingAsCollateral check. In that approach, we would need to check `usingAsCollateral` in `_validateSetUsingAsCollateral` anyway for frozen, so makes sense to also bundle check reserves limit there, to fail earlier. 

for borrow, doesnt make as much sense to add validation into the validateBorrow method, as then we'd need to access positionStatus.isBorrowing(reserveId) twice.

src/spoke/instances/SpokeInstance.sol
CheyenneAtapour: This is not accurate, they are counted separately. But will change to be consistent with the spoke as @miguelmtzinf mentioned above

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
yan-man: agree with first part

> we can add an assert > 0

wdym here, what thing to assert > 0 ?

tests/unit/Spoke/Spoke.Config.t.sol
```diff
@@ -14,6 +14,7 @@ contract SpokeConfigTest is SpokeBase {
     vm.mockCall(oracle, abi.encodeCall(IPriceOracle.DECIMALS, ()), abi.encode(8));
     ISpoke instance = ISpoke(address(DeployUtils.deploySpokeImplementation(oracle)));
     assertEq(instance.ORACLE(), oracle);
+    assertEq(instance.MAX_USER_RESERVES_LIMIT(), type(uint16).max);
```
yan-man: also compare vs Constants file value

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
```diff
@@ -12,7 +12,7 @@ contract SpokeDynamicConfigTest is SpokeBase {
   function setUp() public override {
     super.setUp();
     spoke = MockSpoke(address(spoke1));
-    address mockSpokeImpl = address(new MockSpoke(address(spoke.ORACLE())));
+    address mockSpokeImpl = address(new MockSpoke(address(spoke.ORACLE()), type(uint16).max));
```
yan-man: same

tests/Base.t.sol
CheyenneAtapour: I changed it to pass the hub, but I think it's fine to have it generic imo

src/spoke/instances/SpokeInstance.sol
yan-man: ty, updated

src/spoke/instances/SpokeInstance.sol
CheyenneAtapour: I'm not sure about this, cc @miguelmtzinf @DhairyaSethi we can change here if desired

tests/unit/Spoke/Spoke.DynamicConfig.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Config.t.sol
CheyenneAtapour: Done

src/spoke/Spoke.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
CheyenneAtapour: Done

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
CheyenneAtapour: He means just that the limit is nonzero I think, I updated this

src/spoke/Spoke.sol

src/spoke/interfaces/ISpoke.sol

src/spoke/Spoke.sol
CheyenneAtapour: I updated back to a version I had before. I still don't like how I'm checking if a user is setting as collateral or not twice in the flow, but I think it makes sense to keep the validate function as a view and not alter state. Lemme know what you guys think

src/spoke/instances/SpokeInstance.sol
CheyenneAtapour: Can you check above and let me know if you like @miguelmtzinf 's version better? We should pick one and be consistent

yan-man: yeah miguel's is fine and yep sg just stay consistent

src/spoke/instances/SpokeInstance.sol
miguelmtzinf: main reason is emitting info about the oracle init here instead of doing it in the implementation constructor, so all events come from the proxied/real contract.
wondering if we should add an event for `maxReservesLimit` too... 😿

src/spoke/interfaces/ISpoke.sol
```diff
@@ -322,8 +322,14 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IExtSload, IMul
   /// @notice Thrown when the maximum number of dynamic config keys is reached.
   error MaximumDynamicConfigKeyReached();
 
+  /// @notice Thrown when user attempts to exceed either the maximum allowed collateral or borrowed reserves.
+  error MaximumUserReservesExceeded();
+
+  /// @notice Thrown upon attempt to set a max user reserves limit of 0.
```
miguelmtzinf: can you move right after `InvalidOracleDecimals` please? 
```suggestion
  /// @notice Thrown when the maximum ruser reserves limit is zero in the constructor.
```

src/spoke/instances/SpokeInstance.sol
yan-man: maybe also an event for reserves limit... 

but besides that, should we have all events emitted at the end of initialize?

src/spoke/interfaces/ISpoke.sol
yan-man: Both should be moved to the top then no? given they come first

src/spoke/Spoke.sol
```diff
@@ -920,10 +933,22 @@ abstract contract Spoke is
...
+    ReserveFlags flags,
+    bool usingAsCollateral
+  ) internal view {
     require(!flags.paused(), ReservePaused());
-    // can disable as collateral if the reserve is frozen
-    require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    if (usingAsCollateral) {
+      // disabling as collateral is allowed when reserve is frozen
+      require(!flags.frozen(), ReserveFrozen());
+      // this must be a new collateral, otherwise would have short-circuited
```
yan-man, line 946: dont think this comment really needed but nbd

CheyenneAtapour: I think it's a nice sanity check, will leave

src/spoke/instances/SpokeInstance.sol
CheyenneAtapour: Added event

src/spoke/interfaces/ISpoke.sol
CheyenneAtapour: I put it there, I feel both errors are fine where they're at

src/spoke/instances/SpokeInstance.sol
yan-man: why are we emitting them before the validation?

miguelmtzinf: because they are already initialized at deployment time.

src/spoke/instances/SpokeInstance.sol
```diff
@@ -22,6 +23,7 @@ contract SpokeInstance is Spoke {
   /// @param authority The address of the authority contract which manages permissions.
   function initialize(address authority) external override reinitializer(SPOKE_REVISION) {
     emit UpdateOracle(ORACLE);
+    emit UpdateMaxUserReservesLimit(MAX_USER_RESERVES_LIMIT);
```
miguelmtzinf, line 26:
```suggestion
    emit UpdateMaxUserReservesLimit(MAX_USER_RESERVES_LIMIT);

```
just a bit of separation here, so it's clear they are separate stuff

DhairyaSethi: these can be part of the same event no? init spoke immutables or something

src/spoke/Spoke.sol
```diff
@@ -411,12 +425,11 @@ abstract contract Spoke is
     bool usingAsCollateral,
     address onBehalfOf
   ) external nonReentrant onlyPositionManager(onBehalfOf) {
-    _validateSetUsingAsCollateral(_getReserve(reserveId).flags, usingAsCollateral);
     PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
-
     if (positionStatus.isUsingAsCollateral(reserveId) == usingAsCollateral) {
       return;
     }
+    _validateSetUsingAsCollateral(positionStatus, _getReserve(reserveId).flags, usingAsCollateral);
```
miguelmtzinf: we return when seting false and reserve is not listed. fix pushed.

tests/DeployUtils.sol
```diff
@@ -8,39 +8,70 @@ import {IHub} from 'src/hub/interfaces/IHub.sol';
...
     address oracle,
     bytes32 salt
+  ) internal returns (ISpokeInstance spoke) {
+    return deploySpokeImplementation(oracle, Constants.MAX_ALLOWED_USER_RESERVES_LIMIT, salt);
+  }
+
+  function deploySpokeImplementation(
+    address oracle,
+    uint16 maxUserReservesLimit,
+    bytes32 salt
```
Kogaroshi, line 30: might be worth to add function `deploySpokeImplementation(address oracle, uint16 maxUserReservesLimit)` also, to match with other methods with default salt

src/spoke/Spoke.sol
```diff
@@ -920,10 +934,22 @@ abstract contract Spoke is
...
-    require(!usingAsCollateral || !flags.frozen(), ReserveFrozen());
+    if (usingAsCollateral) {
+      // disabling as collateral is allowed when reserve is frozen
+      require(!flags.frozen(), ReserveFrozen());
+      // this must be a new collateral, otherwise would have short-circuited
+      require(
+        MAX_USER_RESERVES_LIMIT == MAX_ALLOWED_USER_RESERVES_LIMIT ||
+          positionStatus.collateralCount(_reserveCount) < MAX_USER_RESERVES_LIMIT,
+        MaximumUserReservesExceeded()
+      );
```
DhairyaSethi, line 951: bc of this maybe now we dont incentivise keeping coll flag open and disable it on full withdraw? nbd i think either way

tests/unit/Spoke/Spoke.Borrow.Validation.t.sol
```diff
@@ -177,4 +177,85 @@ contract SpokeBorrowValidationTest is SpokeBase {
     vm.expectRevert(abi.encodeWithSelector(IHub.DrawCapExceeded.selector, drawCap));
     Utils.borrow(spoke1, daiReserveId, bob, 1, bob);
   }
+
+  function test_borrow_revertsWith_MaximumUserReservesExceeded() public {
+    // Fetch the user reserves limit
+    uint16 maxUserReservesLimit = spoke3.MAX_USER_RESERVES_LIMIT();
+    assertGt(maxUserReservesLimit, 0, 'Reserve limit is nonzero');
```
DhairyaSethi, line 184: i would also validate this is not the short circuit case, not immediately obvious that w config we have spoke3 here which has exceptional limit set

miguelmtzinf (approved):

src/spoke/instances/SpokeInstance.sol
miguelmtzinf: happy to do it in a separate pr

src/spoke/Spoke.sol
CheyenneAtapour: I'm still in favor of just keeping that logic outside of smart contract, and can have that in ui via multicall

PR #1157, opened 2026-01-28
fix: asset IR strat validation

yan-man:


src/hub/AssetInterestRateStrategy.sol
```diff
@@ -49,6 +52,11 @@ contract AssetInterestRateStrategy is IAssetInterestRateStrategy {
       InvalidOptimalUsageRatio()
     );
     require(rateData.variableRateSlope1 <= rateData.variableRateSlope2, Slope2MustBeGteSlope1());
+    require(
+      rateData.variableRateSlope2 * rateData.optimalUsageRatio >=
+        rateData.variableRateSlope1 * (MAX_USAGE_RATIO - rateData.optimalUsageRatio),
```
yan-man, line 56: variableRateSlope1 / variableRateSlope2 should be renamed to be precise, maybe something like: variableRateGrowth or variableRateIncrease? open to suggestions

yan-man:
leave flexibility in IR strat contract and instead we can add validation in future config engine

PR #1158, opened 2026-01-29
feat: Separate fee accounting pro rata shares

CheyenneAtapour:


src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +210,33 @@ library AssetLogic {
...
+    uint256 interest = delta - fees;
+
+    // Distribute interestForFees pro-rata to unminted fee shares
+    uint256 realizedFees = asset.realizedFees;
+    uint256 addedShares = asset.addedShares;
+    uint256 totalAddedAssetsBefore = asset.liquidity +
+      asset.swept +
+      aggregatedOwedRayBefore.fromRayUp() -
+      realizedFees;
+    uint256 unmintedFeeShares = realizedFees.mulDivDown(addedShares, totalAddedAssetsBefore);
```
yan-man: dont think we can do mulDivDown, need to do toSharesDown to include virtual shares/assets

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +210,37 @@ library AssetLogic {
...
+      aggregatedOwedRayBefore.fromRayUp() -
+      realizedFees;
+    uint256 unmintedFeeShares = SharesMath.toSharesDown(
+      realizedFees,
+      totalAddedAssetsBefore,
+      addedShares
+    );
+    uint256 interestForFees = interest.mulDivDown(
+      unmintedFeeShares,
+      addedShares + SharesMath.VIRTUAL_SHARES + unmintedFeeShares
```
CheyenneAtapour, line 239: Remove virtual shares because this is just a proportional calc ?

avniculae: I think it's correct to use virtual shares here, as they are also earning interest

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +210,37 @@ library AssetLogic {
...
+    // interest = supplier's cut of the delta
+    uint256 interest = delta - fees;
+
+    // Distribute interestForFees pro-rata to unminted fee shares
+    uint256 realizedFees = asset.realizedFees;
+    uint256 addedShares = asset.addedShares;
+    uint256 totalAddedAssetsBefore = asset.liquidity +
+      asset.swept +
+      aggregatedOwedRayBefore.fromRayUp() -
+      realizedFees;
```
avniculae, line 231: initially I though this should be `(realizedFees + fees)`, however you are right, this is correct. We need to simulate as if we were to mint the fee shares immediately after the last accrual

src/hub/libraries/AssetLogic.sol
```diff
@@ -210,10 +210,37 @@ library AssetLogic {
...
+    uint256 unmintedFeeShares = SharesMath.toSharesDown(
+      realizedFees,
+      totalAddedAssetsBefore,
+      addedShares
```
miguelmtzinf, line 235: do we need to add virtual shares here too?

avniculae: already there, since they are part of `toSharesDown`

miguelmtzinf:
Closing this in favor of #1154 , as it's equivalent and provides a more direct way to calculate the proportion between user deposits and treasury fees that are borrowed and generating interest.

PR #1168, opened 2026-01-31
gas analysis: Single reserve event emission for dynamic config update

CheyenneAtapour:


src/spoke/interfaces/ISpoke.sol
```diff
@@ -182,10 +182,6 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IMulticall {
     bool usingAsCollateral
   );
 
-  /// @notice Emitted when a user's dynamic config is refreshed for all reserves to their latest config key.
-  /// @param user The address of the user.
-  event RefreshAllUserDynamicConfig(address indexed user);
-
   /// @notice Emitted when a user's dynamic config is refreshed for a single reserve to its latest config key.
   /// @param user The address of the user.
   /// @param reserveId The identifier of the reserve.
```
yan-man, line 187: i think we should rename this, no need for single. like `RefreshUserDynamicConfig`

src/spoke/Spoke.sol
```diff
@@ -759,6 +760,9 @@ abstract contract Spoke is ISpoke, AccessManagedUpgradeable, IntentConsumer, Mul
             accountData.activeCollateralCount = accountData.activeCollateralCount.uncheckedAdd(1);
           }
         }
+        if (refreshConfig) {
```
yan-man, line 763: wonder if it is worth moving up to L740 and combining the logic with the collFactor
anyway stermi had mentioned having the ternary op there was hard to read

tests/gas/Spoke.UpdateUserDynamicConfig.gas.t.sol
```diff
@@ -0,0 +1,281 @@
...
+    uint256 debtReserveId = _daiReserveId(spoke);
+
+    // Wait one block so alice is liquidatable
+    skip(1);
+    assertLt(spoke.getUserAccountData(alice).healthFactor, HEALTH_FACTOR_LIQUIDATION_THRESHOLD);
+
+    // Liquidate alice
+    vm.prank(bob);
+    spoke.liquidationCall(collateralReserveId, debtReserveId, alice, 1e18, false);
+    vm.snapshotGasLastCall(NAMESPACE, 'L. liquidateUser: 2 collaterals');
```
yan-man, line 125: think these all should be `liquidationCall` rather than liquidateUser

CheyenneAtapour:
Decided not to proceed with this, will keep current convention for emitting events upon dynamic reserve config updates

PR #1169, opened 2026-01-31
poc: emit bitmap

DhairyaSethi:


src/spoke/libraries/PositionStatusMap.sol
```diff
@@ -250,4 +250,71 @@ library PositionStatusMap {
...
+  function compressCollateral(uint256 word) internal pure returns (uint128 packed) {
+    assembly ('memory-safe') {
+      // (word & COLLATERAL_MASK) >> 1 moves bits from positions 1,3,5,... to 0,2,4,...
+      let x := shr(1, and(word, COLLATERAL_MASK))
+
+      // Progressively merge bits from sparse even positions into dense consecutive positions
+
+      // Merge (0,2)→(0,1), (4,6)→(4,5), (8,10)→(8,9), ...
+      x := and(or(x, shr(1, x)), 0x3333333333333333333333333333333333333333333333333333333333333333)
+      // Merge (0-3)→(0-3), (8-11)→(8-11), ...
```
miguelmtzinf: are these comments wrong?

snapshots/Spoke.Operations.json
```diff
@@ -17,18 +17,18 @@
   "supply: 0 borrows, collateral disabled": "122874",
   "supply: 0 borrows, collateral enabled": "105845",
   "supply: second action, same reserve": "105774",
-  "updateUserDynamicConfig: 1 collateral": "74603",
-  "updateUserDynamicConfig: 2 collaterals": "89520",
+  "updateUserDynamicConfig: 1 collateral": "76744",
+  "updateUserDynamicConfig: 2 collaterals": "91661",
```
miguelmtzinf, line 21: shouldn't be the same increase as for 2 collaterals? ~2.1k
gas increase isn't flat now?

src/spoke/interfaces/ISpoke.sol
miguelmtzinf: how it d be for this event?

snapshots/Spoke.Operations.json
Kogaroshi: the gas increase between both old snasphots and new ones is flat here (+2141 on both)

src/spoke/interfaces/ISpoke.sol
Kogaroshi: If we replace this "single" event, the action would be more expensive dur to the bitmap compression needed, so didn't change.

src/spoke/libraries/PositionStatusMap.sol
Kogaroshi: they just describe masks, let me replace with more explanation

Kogaroshi: e2f2b7b4391009047837dd3cf6423afa8c7a93b6

DhairyaSethi:
decided the integrator dx is not worth the additional complexity of this soln

PR #1170, opened 2026-02-02
feat: mint after every action

DhairyaSethi:
todo

- fix update asset config dup mint fee shares
- event likely needs unmintedFees in asset terms as well


src/hub/libraries/AssetLogic.sol
```diff
@@ -132,18 +132,46 @@ library AssetLogic {
...
+    mapping(uint256 assetId => mapping(address spoke => IHub.SpokeData)) storage spokes,
+    IHub.Asset storage asset,
+    uint256 assetId
+  ) internal {
     uint256 drawnIndex = asset.drawnIndex;
+    uint256 swept = asset.swept;
+    uint256 deficitRay = asset.deficitRay;
+    uint120 drawnShares = asset.drawnShares;
+    uint256 liquidity = asset.liquidity;
+    uint256 fees = asset.realizedFees;
```
DhairyaSethi, line 144: `asset.realizedFees` can be packed smaller given it can never be greater than 1 share worth of assets; also it can be removed from storage and unrealisedFees can be maintained in mem/tstore/stack like in https://github.com/aave/aave-v4/pull/927/changes where we are ok loosing rounded fee shares (and be susceptible to treasury grief)

yan-man: asset.realizedFees needs to be set somewhere prior. In accrue if we use initial share price (pre-action), or here if we use final share price (post-action)

src/hub/libraries/AssetLogic.sol
```diff
@@ -132,18 +132,46 @@ library AssetLogic {
...
+    if (fees > 0) {
+      uint256 aggregatedOwedRay = _calculateAggregatedOwedRay({
+        drawnShares: drawnShares,
+        premiumShares: asset.premiumShares,
+        premiumOffsetRay: asset.premiumOffsetRay,
+        deficitRay: deficitRay,
+        drawnIndex: drawnIndex
+      });
+      uint256 totalAdded = liquidity + swept + aggregatedOwedRay.fromRayUp() - fees;
+      feeShares = fees.toSharesDown(totalAdded, asset.addedShares).toUint120(); // donated share price
```
yan-man, line 155: griefing of protocol fee can be done here

DhairyaSethi: yes ofcourse, that's always there. the point is it's not possible to grief 100% bc of > 0 but rather 50% (or 33% i think) of possible fees like it is on dev rn

yan-man: options: 
1) maintain current approach of using post-action share price, but mint as perfect of shares as possible, and leave behind remainder in realizedFees, NOT zero-ing it out. 

ex:
```
uint256 exactFeeAmount = feeShares.toAssetsDown(totalAdded, asset.addedShares + feeShares); 
...
L161: asset.realizedFees = fees - exactFeeAmount;
```
at least this reduces the donation from being up to <1 share worth of assets to at most 1 wei assets, so adds safety. Now griefing is not profitable 

2) simpler - mint with initial share price, pre-donation

DhairyaSethi: either not needed, i had thought about the scenario you're mentioning and it's acceptable imo

yan-man: previously we made this restricted to avoid it, so not sure if it's acceptable

src/hub/Hub.sol
```diff
@@ -168,7 +168,7 @@ contract Hub is IHub, AccessManaged {
       require(irData.length == 0, InvalidInterestRateStrategy());
     }
 
-    asset.updateDrawnRate(assetId);
+    _spokes.updateDrawnRateAndMintFeeShares(asset, assetId);
```
miguelmtzinf: this is ugly, i prefer to use `asset` and pass `_spokes` tbf

src/hub/libraries/AssetLogic.sol
miguelmtzinf: given that we are keeping `realizedFees` in the storage anyway, i d consider the 1st option of minting "perfect" shares and save the excess

PR #1171, opened 2026-02-03
fix: Remove redundant spoke getters

CheyenneAtapour:
Removes spoke getters which are redundant because they are wrappers to existing hub getters:
* getReserveSuppliedAssets
* getReserveSuppliedShares
* getReserveDebt
* getReserveTotalDebt

Similarly removes getters on the TreasurySpoke because they are wrappers to existing hub getters:
* getSuppliedAmount
* getSuppliedShares

Also moves reportDeficit logic to the LiquidationLogic library


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -522,4 +522,45 @@ library LiquidationLogic {
     }
     return !isDebtPositionEmpty || borrowCount > 1;
   }
+
+  /// @notice Reports deficits for all debt reserves of the user, including the reserve being repaid during liquidation.
+  /// @dev Deficit validation should already have occurred during liquidation.
+  /// @dev It clears the user position, setting drawn debt, premium debt, and risk premium to zero.
+  function _reportDeficit(
```
yan-man: let's rm the underscore for this

CheyenneAtapour: It's been done upstream

CheyenneAtapour:
Decided to keep the getters, originally this PR was in case we needed code size space on the spoke contract

PR #1172, opened 2026-02-03
Feat: Mint fee shares with pre-donation share price 

yan-man:


src/hub/Hub.sol
```diff
@@ -152,7 +151,6 @@ contract Hub is IHub, AccessManaged {
 
     address oldFeeReceiver = asset.feeReceiver;
     if (oldFeeReceiver != config.feeReceiver) {
-      _mintFeeShares(asset, assetId);
```
yan-man, line 155: note: bc I opted to combine into `updateDrawnRateAndMintFeeShares` rather than split into 2 separate fns, this means that old fee receiver can lose fees accrued since last action. Shouldnt be much lost though now given fees are accrued on every action. Did this to avoid having to always remember to call mintFeeShares after updateDrawnRate and associated mental overhead of that 

another option is to leave a restricted mintFeeShares action in place so gov can call that before changing feeReceiver

src/hub/libraries/AssetLogic.sol
```diff
@@ -153,7 +182,10 @@ library AssetLogic {
     }
 
     uint256 drawnIndex = asset.getDrawnIndex();
-    asset.realizedFees += asset.getUnrealizedFees(drawnIndex).toUint120();
+    uint256 feeAmount = asset.getUnrealizedFees(drawnIndex);
+    FEE_AMOUNT_SLOT.tstore(feeAmount);
+    uint256 feeShares = asset.toAddedSharesDown(feeAmount);
```
yan-man, line 187: note: if we call `toAddedSharesDown` after FEE_AMOUNT_SLOT is set, then we can avoid calling `getUnrealizedFees` twice, instead we can read from transient: https://github.com/aave/aave-v4/blob/8caf6724910f0bf9d7fbe69d97c1cbaf7347066f/src/hub/libraries/AssetLogic.sol#L108

src/hub/libraries/AssetLogic.sol
```diff
@@ -143,7 +161,18 @@ library AssetLogic {
     });
     asset.drawnRate = newDrawnRate.toUint96();
 
-    emit IHub.UpdateAsset(assetId, drawnIndex, newDrawnRate, asset.realizedFees);
+    uint120 feeShares = FEE_SHARES_SLOT.tload().toUint120();
+    if (feeShares > 0) {
```
yan-man, line 165: note: issue is that this still opens griefing vector for fees < 1 share

DhairyaSethi: griefing allowed for > 1 also. here 100% of the fees can be griefed

yan-man: yeah, just meant it doesnt protect < 1 share either. so yep vector still present

yan-man:
see https://github.com/aave/aave-v4/pull/1178

PR #1173, opened 2026-02-03
feat: mint fee shares with pre-donation share price, perfect shares

yan-man:
- calc fee shares at the beginning of an action, before user donation affects share price, then cache into transient storage
- at the end of the action, mint the equivalent fee shares to feeReceiver
- utilize as "perfect" of shares as possible, whatever remainder exists we keep in realizedFeesRay
- use RAY precision to have a more precise remainder 
- this reduces the griefing vector, as rounded off donation is not thrown away but rather cached for next accrual action


src/hub/libraries/AssetLogic.sol
```diff
@@ -153,7 +194,12 @@ library AssetLogic {
     }
 
     uint256 drawnIndex = asset.getDrawnIndex();
-    asset.realizedFees += asset.getUnrealizedFees(drawnIndex).toUint120();
+    uint256 feeAmountRay = asset.realizedFeesRay + asset.getUnrealizedFees(drawnIndex).toRay();
+    uint256 feeShares = asset.toAddedSharesDown(feeAmountRay.fromRayDown());
+    uint256 perfectFeeAmountRay = feeShares == 0 ? 0 : asset.toAddedAssetsUp(feeShares.toRay());
```
yan-man, line 199: note: feeShares.toRay() first before converting to assets. Ex. if share price is 1.33333...., and feeShares is 1, then this will yield 1.33333 RAY. If we did toRay() after the conversion instead, it would only give us 2 RAY

also, round up so that the remaining realizedFeesRay is minimized https://github.com/aave/aave-v4/blob/c7fbdfce6e9d45119a7b60be699050c401a7d755/src/hub/libraries/AssetLogic.sol#L183

src/hub/libraries/AssetLogic.sol
```diff
@@ -143,7 +170,21 @@ library AssetLogic {
...
+    uint256 perfectFeeAmountRay = PERFECT_FEE_AMOUNT_RAY_SLOT.tload();
+    uint120 feeShares = FEE_SHARES_SLOT.tload().toUint120();
+    if (feeShares > 0) {
+      address feeReceiver = asset.feeReceiver;
+      asset.addedShares += feeShares;
+      spokes[assetId][feeReceiver].addedShares += feeShares;
+      FEE_SHARES_SLOT.tstore(0);
+      emit IHub.MintFeeShares(assetId, feeReceiver, feeShares, perfectFeeAmountRay.fromRayDown());
+    }
+    asset.realizedFeesRay = (feeAmountRay - perfectFeeAmountRay).toUint200();
```
yan-man, line 183: note: whatever remainder exists beyond the perfect shares, cache for the next accrual 

maybe needs to be capped bc perfectFeeAmountRay was rounded up, `feeAmountRay <= perfectFeeAmountRay ? 0: feeAmountRay - perfectFeeAmountRay`

yan-man:
see https://github.com/aave/aave-v4/pull/1178

PR #1174, opened 2026-02-03
fix: Move reportDeficit to external lib in favor of Spoke bytecode size

miguelmtzinf:
Moves `reportDeficit` logic to `LiquidationLogic` external library to save bytecode size in the Spoke.

- Saves 439 B bytecode.
- Gas costs of a liquidation with deficit increases ~1716 and ~1695 units depending on whether risk premium is enabled.

### DIFF
--- a/snapshots/Spoke.Operations.ZeroRiskPremium.json
+++ b/snapshots/Spoke.Operations.ZeroRiskPremium.json
@@ -3,6 +3,7 @@
   "borrow: second action, same reserve": "169471",
   "liquidationCall (receiveShares): full": "295586",
   "liquidationCall (receiveShares): partial": "295304",
+  "liquidationCall (reportDeficit): full": "350919",
   "liquidationCall: full": "305078",
   "liquidationCall: partial": "304796",
   "permitReserve + repay (multicall)": "164506",
--- a/snapshots/Spoke.Operations.json
+++ b/snapshots/Spoke.Operations.json
@@ -3,6 +3,7 @@
   "borrow: second action, same reserve": "201454",
   "liquidationCall (receiveShares): full": "327674",
   "liquidationCall (receiveShares): partial": "327392",
+  "liquidationCall (reportDeficit): full": "345548",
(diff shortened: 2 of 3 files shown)


src/spoke/Spoke.sol
```diff
@@ -844,38 +844,11 @@ abstract contract Spoke is
     }
     emit UpdateUserRiskPremium(user, newRiskPremium);
   }
-
   /// @notice Reports deficits for all debt reserves of the user, including the reserve being repaid during liquidation.
   /// @dev Deficit validation should already have occurred during liquidation.
-  /// @dev It clears the user position, setting drawn debt, premium debt, and risk premium to zero.
+  /// @dev It clears the user position, setting drawn debt and premium debt to zero.
   function _reportDeficit(address user) internal {
```
avniculae, line 849: why don't we remove this and call liq logic directly?

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -243,6 +243,52 @@ library LiquidationLogic {
       });
   }
 
+  /// @notice Reports deficits for all debt reserves of the user.
+  /// @dev Deficit validation should already have occurred during liquidation.
+  /// @dev It clears the user position, setting drawn debt, premium debt, and risk premium to zero.
+  /// @param reserves The mapping of reserves per reserve identifier.
+  /// @param userPositions The mapping of user positions per reserve per user.
```
avniculae, line 250:
```suggestion
  /// @param userPositions The mapping of user positions per user per reserve.
```

src/spoke/Spoke.sol
```diff
@@ -390,6 +390,7 @@ abstract contract Spoke is
 
     uint256 newRiskPremium = 0;
     if (isUserInDeficit) {
```
DhairyaSethi, line 392: at this point lets just move the entire flow in liquidation logic lib tbh, _processUserAccountData can be moved to another lib and be internally shared by spoke & liq

PR #1178, opened 2026-02-03
rft: mint fee shares at the start

DhairyaSethi:


src/hub/Hub.sol
```diff
@@ -602,7 +602,7 @@ contract Hub is IHub, AccessManaged {
   /// @inheritdoc IHub
   function getAssetAccruedFees(uint256 assetId) external view returns (uint256) {
     Asset storage asset = _assets[assetId];
-    return asset.realizedFees + asset.getUnrealizedFees(asset.getDrawnIndex());
+    return asset.realizedFees; // todo fix
```
yan-man: should this be `+ asset.unrealizedFeeShares()`?

src/hub/Hub.sol
yan-man, line 627: todo: needs updating too, for if spoke is feeReceiver

yan-man: todo: needs updating too, for if spoke is feeReceiver

src/hub/Hub.sol
DhairyaSethi: no this getter will be removed along with realised fees var

this pr is very early stage rn

yan-man: hm ok we should finalize the impl first then before working on tests @CheyenneAtapour

src/hub/Hub.sol
yan-man, line 154: todo: rm

src/hub/Hub.sol
```diff
@@ -209,7 +209,7 @@ contract Hub is IHub, AccessManaged {
   function mintFeeShares(uint256 assetId) external restricted returns (uint256) {
```
yan-man, line 208: todo: determine if we want to keep this around, but doesnt seem necessary

src/hub/Hub.sol
DhairyaSethi: the core impl should be done for the 99% of the important tests that we want to migrate, also pls continue thinking about the consequences 

but yea feel free to branch off and finish impl meanwhile in your work day

src/hub/Hub.sol
CheyenneAtapour: We mentioned that if there is no action for a while, we might want to manually trigger this, even with this kind of implementation

src/hub/interfaces/IHub.sol
```diff
@@ -118,6 +118,8 @@ interface IHub is IHubBase, IAccessManaged {
     uint256 accruedFees
   );
 
+  event AccrueFees(uint256 indexed assetId, address feeReceiver, uint256 feeShares);
```
CheyenneAtapour, line 113: TODO: we need to add natspec here

src/hub/libraries/AssetLogic.sol
```diff
@@ -76,10 +76,15 @@ library AssetLogic {
     return asset.drawn(drawnIndex) + asset.premium(drawnIndex);
   }
 
-  /// @notice Returns the total added assets for the specified asset.
```
CheyenneAtapour, line 79: Why was this comment removed?

src/hub/libraries/AssetLogic.sol
```diff
@@ -76,10 +76,15 @@ library AssetLogic {
...
   function totalAddedAssets(IHub.Asset storage asset) internal view returns (uint256) {
-    uint256 drawnIndex = asset.getDrawnIndex();
+    return asset.totalAddedAssets(asset.getDrawnIndex());
+  }
 
+  /// @notice Returns the total added assets for the specified asset.
```
CheyenneAtapour, line 83: this could at the end say "at the specified drawnIndex"

src/hub/libraries/AssetLogic.sol
```diff
@@ -88,44 +93,75 @@ library AssetLogic {
       drawnIndex: drawnIndex
     });
 
-    return
-      asset.liquidity +
-      asset.swept +
-      aggregatedOwedRay.fromRayUp() -
-      asset.realizedFees -
-      asset.getUnrealizedFees(drawnIndex);
+    return asset.liquidity + asset.swept + aggregatedOwedRay.fromRayUp();
```
CheyenneAtapour, line 96: Why not subtract out unrealizedFees?

CheyenneAtapour: Oh, so we are including the unrealized both in the numerator and denominator now? (both assets and shares)

DhairyaSethi: no we will remove unrealised fees from this impl and go back to what we had before (only on denominator)

src/hub/libraries/AssetLogic.sol
```diff
@@ -194,25 +254,28 @@ library AssetLogic {
...
-    });
+    }).fromRayUp();
+
+    uint256 feeAmount = (aggregatedOwedAfter - aggregatedOwedBefore).percentMulDown(liquidityFee);
 
     return
-      (aggregatedOwedRayAfter.fromRayUp() - aggregatedOwedRayBefore.fromRayUp()).percentMulDown(
-        liquidityFee
+      feeAmount.toSharesDown(
+        asset.liquidity + asset.swept + aggregatedOwedAfter - feeAmount,
```
CheyenneAtapour, line 277: subtracted out feeAmount because this should be the share rate before the current action (before these fees are accrued), so this looks good

src/hub/libraries/AssetLogic.sol
```diff
@@ -173,14 +228,19 @@ library AssetLogic {
...
-  /// @param drawnIndex The current drawn index.
-  function getUnrealizedFees(
+  /// @notice Calculates the amount of unrealized fee shares since last accrual.
+  function unrealizedFeeShares(IHub.Asset storage asset) internal view returns (uint256) {
+    return asset.getFeeShares(asset.getDrawnIndex(), asset.drawnIndex);
+  }
+
+  /// @notice Calculates the amount of fee shares derived from the index growth due to interest accrual.
+  /// @dev The true liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee.
+  function getFeeShares(
```
CheyenneAtapour, line 238: Should we specify that we are rounding down (which I think is correct btw) in natspec dev comment?

src/hub/libraries/AssetLogic.sol
```diff
@@ -147,20 +183,39 @@ library AssetLogic {
...
+    uint120 feeShares = asset.getFeeShares(drawnIndex, previousIndex).toUint120();
+    if (feeShares > 0) {
+      address feeReceiver = asset.feeReceiver;
+      asset.addedShares += feeShares;
+      spokes[assetId][feeReceiver].addedShares += feeShares;
+      emit IHub.AccrueFees(assetId, feeReceiver, feeShares);
+    }
   }
 
-  /// @notice Calculates the drawn index of a specified asset based on the existing drawn rate and index.
```
CheyenneAtapour, line 161: Why removing natspec?

PR #1180, opened 2026-02-03
opt: simd comparison

DhairyaSethi:


Kogaroshi:
This change does not seem to decrease gas costs, and actually increase Spoke codesize by 20B.
Imo not worth it.

PR #1181, opened 2026-02-03
feat: Analyze griefing via graphs

CheyenneAtapour:


tests/misc/fee_griefing_18dec.png
yan-man, line 1: i think maybe we can start the x values from 1 onwards, rather than 0 (which approach y inf)

tests/misc/fee_griefing_6dec.png
yan-man, line 1: i dont think the rhs y-ax is correct, $100 should be at like 100e6 in tokens, no?

tests/misc/fee_griefing_analysis.py
```diff
@@ -0,0 +1,495 @@
...
+def create_rate_comparison_plot(filename: str = "fee_griefing_rate_comparison.png") -> None:
+    """
+    Create a plot comparing boundaries at different interest rates.
+    """
+    rates = [
+        (int(0.01 * RAY), "1% APR"),
+        (int(0.05 * RAY), "5% APR"),
+        (int(0.10 * RAY), "10% APR"),
+        (int(0.20 * RAY), "20% APR"),
+        (int(0.50 * RAY), "50% APR"),
```
yan-man, line 356: yeah maybe we dont need a 50% one but nbd

tests/misc/fee_griefing_18dec.png
yan-man, line 1: maybe eth/dai we can split into 2 separate graphs so that we can show the $ on the rhs, which would be diff scales for those 2

tests/misc/fee_griefing_6dec.png
CheyenneAtapour: in tokens means the number is already * 1e6, like 1 "token" = 1e6

CheyenneAtapour:
Closing as we are no longer minting fee shares upon accrual

PR #1185, opened 2026-02-04
opt: position manager update hash

DhairyaSethi:
trimmed version of #1094

### motivation
very useful for user position update since this method will be called repeatedly inside a loop when user updates multiple position managers using `setUserPositionManagersWithSig` improving gas usage for user onboarding

### description
rewrite of hash in assembly for <= 3 params because of major efficiency gains.
this PR implements digest hashing on nested EIP712 struct `PositionManagerUpdate` in yul, optimising over the solidity implementation to skip memory expansion by hashing in scratch space and temporarily using fmp


src/spoke/libraries/EIP712Hash.sol
```diff
@@ -35,7 +35,17 @@ library EIP712Hash {
...
-    return keccak256(abi.encode(POSITION_MANAGER_UPDATE, params.positionManager, params.approve));
+  function hash(
+    ISpoke.PositionManagerUpdate calldata params
+  ) internal pure returns (bytes32 digest) {
+    // equivalent to: keccak256(abi.encode(POSITION_MANAGER_UPDATE, params.positionManager, params.approve))
+    assembly ('memory-safe') {
+      let fmp := mload(0x40)
+      mstore(0, POSITION_MANAGER_UPDATE)
+      mstore(0x20, shr(96, shl(96, calldataload(params)))) // params.positionManager
+      mstore(0x40, iszero(iszero(calldataload(add(params, 0x20))))) // params.update
```
CheyenneAtapour:
```suggestion
      mstore(0x40, iszero(iszero(calldataload(add(params, 0x20))))) // params.approve
```

CheyenneAtapour (dismissed):

src/spoke/libraries/EIP712Hash.sol
```diff
@@ -35,7 +35,17 @@ library EIP712Hash {
       );
   }
 
-  function hash(ISpoke.PositionManagerUpdate calldata params) internal pure returns (bytes32) {
-    return keccak256(abi.encode(POSITION_MANAGER_UPDATE, params.positionManager, params.approve));
+  function hash(
+    ISpoke.PositionManagerUpdate calldata params
+  ) internal pure returns (bytes32 digest) {
+    // equivalent to: keccak256(abi.encode(POSITION_MANAGER_UPDATE, params.positionManager, params.approve))
```
yan-man, line 41: i think would be good to have a test to verify this

DhairyaSethi: already exists in eip712hash.t.sol which builds it the naive way (exactly as mentioned in the comment) and builds it using the forge cheatcode helper

Kogaroshi (dismissed):

yan-man (dismissed):

src/spoke/libraries/EIP712Hash.sol
```diff
@@ -35,7 +35,17 @@ library EIP712Hash {
...
+    assembly ('memory-safe') {
```
DhairyaSethi:
```suggestion
    assembly {
```

src/spoke/libraries/EIP712Hash.sol
```diff
@@ -35,7 +35,17 @@ library EIP712Hash {
...
-    return keccak256(abi.encode(POSITION_MANAGER_UPDATE, params.positionManager, params.approve));
+  function hash(
+    ISpoke.PositionManagerUpdate calldata params
+  ) internal pure returns (bytes32 digest) {
+    // equivalent to: keccak256(abi.encode(POSITION_MANAGER_UPDATE, params.positionManager, params.approve))
+    assembly {
+      let fmp := mload(0x40)
+      mstore(0, POSITION_MANAGER_UPDATE)
+      mstore(0x20, shr(96, shl(96, calldataload(params)))) // params.positionManager
+      mstore(0x40, iszero(iszero(calldataload(add(params, 0x20))))) // params.approve
```
avniculae, line 46: After back & forth with Dhairya, it turns out this (and address sanity check above) deviates from the canonical solidity ABI encoding/decoding. Solidity does not accept any dirty bits in sub-word values (e.g., reverts if bool is 0x100...00). It would be fine to remove any arg cleaning here, since both fields of the calldata are accessed later in Solidity (outside of yul), and at that point sanity checks are performed. Note that in this case, these checks are not performed when entering the external function because params are `calldata`, not `memory`.

miguelmtzinf (dismissed):

PR #1189, opened 2026-02-04
fix: increase size of dynamicConfigKey to uint32

yan-man:
- change dynamicConfigKey from uint24 to uint32
- allows for a larger number of possible dynamic config keys (from 16M with uint24 to 4B with uint32)
- move dynamicConfigKey field to end of the struct to allow for future upgrades that expand it further if needed

The main motivation for these changes is having more space for more dynamic configuration keys, which provides greater flexibility for more frequent updates of risk params (e.g. automated agents).

### DIFF
--- a/src/spoke/Spoke.sol
+++ b/src/spoke/Spoke.sol
@@ -147,14 +147,14 @@ abstract contract Spoke is
       hub: IHubBase(hub),
       assetId: assetId.toUint16(),
       decimals: decimals,
-      dynamicConfigKey: dynamicConfigKey,
       collateralRisk: config.collateralRisk,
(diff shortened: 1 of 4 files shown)


src/spoke/interfaces/ISpoke.sol
```diff
@@ -49,7 +49,7 @@ interface ISpoke is ISpokeBase, IAccessManaged, IIntentConsumer, IExtSload, IMul
     IHubBase hub;
     uint16 assetId;
     uint8 decimals;
-    uint24 dynamicConfigKey;
+    uint40 dynamicConfigKey;
```
yan-man: uint32 and move to bottom

PR #1190, opened 2026-02-04
fix: gas tests with interest accrual (vs dev)

yan-man:
- update gas snapshot tests so that they are more representative of typical flows, with existing assets supplied and borrowed, earning interest. Also skip time between actions so that the actions trigger meaningful interest accrual calculations. 
- should be merged prior to `mintFeeShares` fix to get a more representative view of gas changes associated with accrual actions

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -38,10 +38,10 @@ runs = 1000
 seed = "0x640"
 
 [profile.pr.fuzz]
-runs = 5000
+runs = 5
 
 [profile.ci.fuzz]
-runs = 10000
+runs = 5
 
 [profile.gas]
 gas_snapshot_check = true


tests/gas/Gateways.Operations.gas.t.sol
```diff
@@ -112,96 +124,130 @@ contract SignatureGateway_Gas_Tests is SignatureGatewayBaseTest {
...
-    Utils.supply(spoke1, p.reserveId, alice, p.amount, alice);
 
     gateway.supplyWithSig(p, signature);
     vm.snapshotGasLastCall(NAMESPACE, 'supplyWithSig');
   }
 
   function test_withdrawWithSig() public {
+    uint256 reserveId = _wethReserveId(spoke1);
+    uint256 amount = 100e18;
+    Utils.supply(spoke1, reserveId, alice, 200e18, alice);
```
CheyenneAtapour: If you're already editing to put the withdraw amount in a variable, what about amount * 2 for the supply?

tests/gas/Hub.Operations.gas.t.sol
```diff
@@ -20,6 +39,8 @@ contract HubOperations_Gas_Tests is Base {
     hub1.add(usdxAssetId, 1000e6);
     vm.snapshotGasLastCall('Hub.Operations', 'add');
```
CheyenneAtapour, line 40: Do we not want this add to have an accrue with it?

tests/gas/Hub.Operations.gas.t.sol
CheyenneAtapour, line 58: Do we not want this remove to have an accrue with it?

CheyenneAtapour, line 325: Skip before this one too?

tests/gas/Spoke.Operations.gas.t.sol
```diff
@@ -16,19 +16,32 @@ contract SpokeOperations_Gas_Tests is SpokeBase {
...
+    spoke.setUsingAsCollateral(reserveId.usdx, true, bob);
+    spoke.borrow(reserveId.dai, 100e18, bob);
+    spoke.borrow(reserveId.weth, 1e18, bob);
+    spoke.borrow(reserveId.wbtc, 1e5, bob);
+    vm.stopPrank();
+    skip(100);
   }
 
   function test_supply() public {
+    skip(100);
```
CheyenneAtapour: This skip prob redundant because setup already does it

tests/gas/Spoke.Operations.gas.t.sol
CheyenneAtapour, line 56: What about skipping before this action?

CheyenneAtapour, line 68: Need a skip in between

CheyenneAtapour, line 80: Need a skip in between

CheyenneAtapour, line 88: skip in between

CheyenneAtapour: This skip is unneeded I think

CheyenneAtapour, line 134: We need a skip cause the supply would have triggered accrue

tests/gas/Spoke.Operations.gas.t.sol
```diff
@@ -147,6 +170,8 @@ contract SpokeOperations_Gas_Tests is SpokeBase {
   function test_liquidation_full() public {
     _liquidationSetup(85_00);
 
+    skip(100);
+
     vm.startPrank(bob);
```
CheyenneAtapour: not you, but we could just prank bob, no need for startPrank

tests/gas/Spoke.Operations.gas.t.sol
```diff
@@ -157,6 +182,8 @@ contract SpokeOperations_Gas_Tests is SpokeBase {
   function test_liquidation_receiveShares_partial() public {
     _liquidationSetup(85_00);
 
+    skip(100);
+
     vm.startPrank(bob);
```
CheyenneAtapour: same, could just prank

tests/gas/Hub.Operations.gas.t.sol
yan-man: this one should be fine right bc of the skip from setup?

tests/gas/Spoke.Operations.gas.t.sol
yan-man: i think was from me before anyway

snapshots/Hub.Operations.json
DhairyaSethi, line 1: i kinda don't agree w the premise of the pr, it was discussed before as well and the existing tests found a balance to the average user experience 

this makes it for the unlikely worst case, ie only the first user in a block will experience this, every other user won't. for debugging & analysis i 100% agree this is the way to go but not sure overall about the goal. wdyt

CheyenneAtapour: wouldn't the average user experience involve an accrue within each action? I think it's less likely that a user would interact at the same block as another, but perhaps not?

yan-man: depends on what we deem as the "canonical" user experience. Was discussing with Cheyenne, I feel like the user action with accrual included is that. It's kind of a ceiling vs floor gas cost. 

but when we're looking at mintFeeShares, we can't evaluate it properly unless it includes an accrual action. At the moment it's simply not consistent in any way. I don't think it's a balance for avg UX, it's just kind of haphazard imo atm

tests/gas/Hub.Operations.gas.t.sol
yan-man: ty, thought the setup covered this but missed the add

PR #1192, opened 2026-02-04
fix: gas tests with interest accrual (1193 vs dev)

yan-man:
branched off #1193, vs dev


yan-man:
decided to leave minting out of accrual

PR #1193, opened 2026-02-04
rft: Mint fee shares when realized fees exceed one share

CheyenneAtapour:


src/hub/libraries/AssetLogic.sol
yan-man: ok 1 opt i can see here - L280 already defines feeAmount as asset.realizedFees, so here you can just do feeAmount +=

src/hub/libraries/AssetLogic.sol
```diff
@@ -112,18 +112,46 @@ library AssetLogic {
     return asset.addedShares + feeShares;
   }
 
+  /// @notice Returns both totalAddedAssets and totalAddedShares with a single getFee() call.
```
yan-man:
```suggestion
  /// @notice Returns both `totalAddedAssets` and `totalAddedShares` with a single `getFee` call.
```

src/hub/libraries/AssetLogic.sol
```diff
@@ -112,18 +112,46 @@ library AssetLogic {
...
+  function getTotalAssetsAndShares(
+    IHub.Asset storage asset,
+    uint256 drawnIndex,
+    uint256 previousIndex
+  ) internal view returns (uint256, uint256) {
+    (uint256 feeAmount, uint256 feeShares) = asset.getFee(drawnIndex, previousIndex);
+
+    uint256 projectedRealizedFees = feeShares > 0 ? 0 : feeAmount;
```
yan-man: think we can just call this realizedFees

src/hub/Hub.sol
```diff
@@ -325,8 +325,7 @@ contract Hub is IHub, AccessManaged {
     spoke.drawnShares -= drawnShares;
     _applyPremiumDelta(asset, spoke, premiumDelta);
 
-    uint256 deficitAmountRay = uint256(drawnShares) *
```
yan-man: have we always been doing this? perhaps we should do safeCast here? can we drawnShares.toUint256(), but ig uint120 never can overflow on 256, so maybe nbd

src/hub/libraries/AssetLogic.sol
CheyenneAtapour: Actually to be consistent with the rest of the library, I think I should do it like

```suggestion
  /// @notice Returns both total added assets and shares for the specified asset. 
  /// @dev Uses only a single getFee() call.
```

src/hub/libraries/AssetLogic.sol

src/hub/libraries/AssetLogic.sol
```diff
@@ -71,14 +71,37 @@ library AssetLogic {
...
   }
 
-  /// @notice Returns the total amount owed for the specified asset, including drawn and premium.
+  /// @notice Returns the total amount owed for the specified asset at specified drawnIndex.
   function totalOwed(IHub.Asset storage asset, uint256 drawnIndex) internal view returns (uint256) {
     return asset.drawn(drawnIndex) + asset.premium(drawnIndex);
   }
 
   /// @notice Returns the total added assets for the specified asset.
   function totalAddedAssets(IHub.Asset storage asset) internal view returns (uint256) {
```
DhairyaSethi: i think we can rm these two now in favor of gettotalassetsandsharea, only used by ext getter

src/hub/interfaces/IHub.sol
```diff
@@ -26,17 +27,18 @@ interface IHub is IHubBase, IAccessManaged {
   /// @dev irStrategy The address of the interest rate strategy.
   /// @dev reinvestmentController The address of the reinvestment controller.
   /// @dev feeReceiver The address of the fee receiver spoke.
-  /// @dev deficitRay The amount of outstanding bad debt across all spokes, expressed in asset units and scaled by RAY.
   struct Asset {
     uint120 liquidity;
-    uint120 realizedFees;
+    uint120 swept;
```
DhairyaSethi: revert this reorder but retain deficit ray moving up

src/hub/libraries/AssetLogic.sol
```diff
@@ -71,14 +71,37 @@ library AssetLogic {
...
+  /// @notice Returns the total added shares for the specified asset.
+  function totalAddedShares(IHub.Asset storage asset) internal view returns (uint256) {
+    uint256 previousIndex = asset.drawnIndex;
+    uint256 drawnIndex = asset.getDrawnIndex(previousIndex);
+    (, uint256 totalShares) = asset.getTotalAssetsAndShares(drawnIndex, previousIndex);
+    return totalShares;
+  }
+
+  /// @notice Returns both total added assets and shares for the specified asset.
+  /// @dev Uses only a single getFee() call.
```
DhairyaSethi: rm

src/hub/libraries/AssetLogic.sol

src/hub/interfaces/IHub.sol
CheyenneAtapour: Done

src/hub/libraries/AssetLogic.sol

src/hub/libraries/AssetLogic.sol
```diff
@@ -71,14 +71,20 @@ library AssetLogic {
...
-    uint256 drawnIndex = asset.getDrawnIndex();
+  /// @notice Returns both total added assets and shares for the specified asset.
+  function getTotalAssetsAndShares(
+    IHub.Asset storage asset,
+    uint256 drawnIndex,
+    uint256 previousIndex
+  ) internal view returns (uint256, uint256) {
+    (uint256 feeAmount, uint256 feeShares) = asset.getFee(drawnIndex, previousIndex);
+
+    uint256 realizedFees = feeShares > 0 ? 0 : feeAmount;
```
DhairyaSethi, line 87: this is kinda confusing at first glance so pls add a comment saying how when feeShares are non zero it means all of `feeAmount` is minted else not OR revisit how `getFee` helper returns information, possibly passing it through an intermediary wrapper which does this logic

src/hub/libraries/AssetLogic.sol
```diff
@@ -71,14 +71,20 @@ library AssetLogic {
...
 
     uint256 aggregatedOwedRay = _calculateAggregatedOwedRay({
```
DhairyaSethi, line 89: this is being recalculated for the third from bc of `getFee`, lets see if we can cache it

src/hub/libraries/AssetLogic.sol
DhairyaSethi: the logic is correct tho, and exactly how i did it in accrue but in isolation is weird

src/hub/libraries/AssetLogic.sol
```diff
@@ -173,47 +228,59 @@ library AssetLogic {
...
   }
 
-  /// @notice Calculates the amount of fees derived from the index growth due to interest accrual.
-  /// @param drawnIndex The current drawn index.
-  function getUnrealizedFees(
+  /// @notice Calculates the amount of unrealized fee shares since last accrual.
+  function unrealizedFeeShares(IHub.Asset storage asset) internal view returns (uint256) {
+    (, uint256 feeShares) = asset.getFee(asset.getDrawnIndex(), asset.drawnIndex);
+    return feeShares;
+  }
```
DhairyaSethi, line 235: this should not be needed anymore right

src/hub/libraries/AssetLogic.sol
```diff
@@ -173,47 +228,59 @@ library AssetLogic {
...
+
+  /// @notice Calculates the amount of fee shares derived from the index growth due to interest accrual.
+  /// @dev The true liquidity growth is always greater than accrued fees, even with 100.00% liquidity fee.
```
DhairyaSethi, line 238: this comment is not true anymore i think? anyway something for later

CheyenneAtapour:
Closing this as we are leaving mint fee shares to be handled outside of src

PR #1194, opened 2026-02-04
feat: calculate `riskPremiumThreshold` from `collateralRisk` in hub configurator 

yan-man:
In `HubConfigurator`, allow for setting the Risk Premium Threshold from the maximum collateral risk associated with a given spoke. The RP Threshold must be set to a value sufficiently higher than the max collateral risk in order to prevent rounding situations that block `refreshPremium` calls from that spoke. Prevents misconfiguration by helping to calculate the RP Threshold for the user. 

- see https://github.com/aave/aave-v4/pull/1142/changes#r2729207563
- formula reference: https://github.com/aave/aave-v4/pull/1142/changes#r2727737374
- extension of PR: https://github.com/aave/aave-v4/pull/1142

### DIFF
--- a/src/hub/HubConfigurator.sol
+++ b/src/hub/HubConfigurator.sol
@@ -327,7 +327,7 @@ contract HubConfigurator is AccessManaged, IHubConfigurator {
     hub.updateAssetConfig(assetId, config, new bytes(0));
   }
 
-  /// @dev Calculates the risk premium threshold from the maximum collateral risk, both expressed in BPS.
+  /// @dev Calculates the risk premium threshold associated with the maximum collateral risk, both expressed in BPS.
   function _calculateRiskPremiumThreshold(
     uint256 maxCollateralRisk
   ) internal pure returns (uint256) {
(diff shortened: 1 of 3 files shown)


src/hub/HubConfigurator.sol
```diff
@@ -235,10 +237,11 @@ contract HubConfigurator is AccessManaged, IHubConfigurator {
     address hub,
     uint256 assetId,
     address spoke,
-    uint256 riskPremiumThreshold
+    uint256 maxCollateralRisk
```
DhairyaSethi, line 240: there is no concept of 'collateral' and collateral risk on the hub. should we rename this to max risk premium or max premium ratio or smth

yan-man: mm i think max risk premium is kind of too similar to risk premium threshold. I see what you mean, but it's also setting this value for the spoke, so it's from the spoke's perspective of the MAX_ALLOWED_COLLATERAL_RISK which will be set, which is the value intended to be added here. Open to everyone's thoughts on this.

DhairyaSethi: i mean max coll risk is this spoke specific property. most other spokes won't have this (tokenisation, treasury, etc) 
we could come up w something more generic 

from the current spoke perspective it's just the maximum possible individual risk premium it can report or something along these lines

yan-man: ig that's a good point, what would we set as this RP threshold for tokenization or treasury, max sentinel?

DhairyaSethi: yeah for that this needs an overload which accepts threshold input only such that we can set it to zero for those spokes

yan-man:
leave for later, or don't do it

PR #1196, opened 2026-02-05
fix: checked nonce increment

DhairyaSethi:
use checked increment for nonces, an alternate mitigation for the scenario described in #1144


miguelmtzinf:
2^64 nonces is a very large space. Even if a user consumes 10^6 nonces per second, it would take on the order of hundreds of thousands of years to overflow. Given how impractical this is, we’re fine keeping the code as is (not checking overflow via unchecked operation).

PR #1199, opened 2026-02-08
fix: Add getter to compute last asset drawn rate on Hub

DhairyaSethi:
- change `getAssetDrawnRate` to return latest computed drawn rate

### DIFF
--- a/src/hub/Hub.sol
+++ b/src/hub/Hub.sol
@@ -537,12 +537,6 @@ contract Hub is IHub, AccessManaged {
     return _assets[assetId].addedShares;
   }
 
-  /// @inheritdoc IHub
-  function getAssetDrawnRate(uint256 assetId) external view returns (uint256) {
-    Asset storage asset = _assets[assetId];
-    return asset.getDrawnRate(assetId, asset.getDrawnIndex());
-  }
-
   /// @inheritdoc IHubBase
   function getAssetOwed(uint256 assetId) external view returns (uint256, uint256) {
     Asset storage asset = _assets[assetId];
@@ -614,6 +608,12 @@ contract Hub is IHub, AccessManaged {
   /// @inheritdoc IHub
   function getAssetSwept(uint256 assetId) external view returns (uint256) {
     return _assets[assetId].swept;
+  }
+
+  /// @inheritdoc IHub
+  function getAssetDrawnRate(uint256 assetId) external view returns (uint256) {
+    Asset storage asset = _assets[assetId];
+    return asset.getDrawnRate(assetId, asset.getDrawnIndex());
   }
 
   /// @inheritdoc IHub


src/hub/interfaces/IHubBase.sol
```diff
@@ -220,7 +220,12 @@ interface IHubBase {
   /// @notice Calculates the current drawn index for the specified asset.
   /// @param assetId The identifier of the asset.
   /// @return The current drawn index of the asset.
-  function getAssetDrawnIndex(uint256 assetId) external view returns (uint256);
+  function computeAssetDrawnIndex(uint256 assetId) external view returns (uint256);
```
miguelmtzinf: i d revert back and use `get` convention, because we have other getters that are extrapolating anyway (e.g. `getAddedAssets`). So as long as natspec docs are clear i'm okay.

src/hub/interfaces/IHub.sol
```diff
@@ -374,9 +374,9 @@ interface IHub is IHubBase, IAccessManaged {
   /// @return The amount of liquidity swept.
   function getAssetSwept(uint256 assetId) external view returns (uint256);
 
-  /// @notice Returns the current drawn rate for the specified asset.
+  /// @notice Returns the last stored drawn rate for the specified asset.
   /// @param assetId The identifier of the asset.
-  /// @return The current drawn rate of the asset.
+  /// @return The last stored drawn rate of the asset.
   function getAssetDrawnRate(uint256 assetId) external view returns (uint256);
```
miguelmtzinf, line 380: this can be removed since it's returned in `getAsset`?

src/hub/interfaces/IHubBase.sol
```diff
@@ -220,7 +220,12 @@ interface IHubBase {
...
+
+  /// @notice Calculates the current drawn rate for the specified asset.
+  /// @param assetId The identifier of the asset.
+  /// @return The current drawn rate of the asset.
+  function computeAssetDrawnRate(uint256 assetId) external view returns (uint256);
```
miguelmtzinf: is drawn rate generic enough for having it here? prob better to move to `IHub`

src/hub/libraries/AssetLogic.sol
```diff
@@ -173,6 +165,24 @@ library AssetLogic {
       );
   }
 
+  /// @notice Calculates the latest drawn rate of a specified asset using the specified drawnIndex.
```
```suggestion
  /// @notice Calculates the drawn rate of a specified asset using the specified drawn index.
```

src/hub/interfaces/IHub.sol
DhairyaSethi: we also have getAssetLiq & swept etc so we keep those yeah

src/hub/interfaces/IHubBase.sol
DhairyaSethi: drawn index getter is in base, wdyt? move both out? i think either both out or both in

miguelmtzinf (approved):

src/hub/interfaces/IHub.sol
```diff
@@ -374,8 +374,7 @@ interface IHub is IHubBase, IAccessManaged {
   /// @return The amount of liquidity swept.
   function getAssetSwept(uint256 assetId) external view returns (uint256);
 
-  /// @notice Returns the last updated drawn rate for the specified asset.
-  /// @dev Drawn rate can be outdated due to passage of time.
+  /// @notice Calculates the current drawn index for the specified asset.
```
DhairyaSethi, line 377:
```suggestion
  /// @notice Calculates the current drawn rate for the specified asset.
```

PR #1201, opened 2026-02-09
fix: max deposit rounding on TokenizationSpoke

DhairyaSethi:
CS issue: `
maxDeposit can over-estimate the amount. It uses totalAssets() which rounds down the assets amount, and subtracts it, so that the result rounds up. The returned amount can then cause _validateAdd to revert
`


tests/unit/TokenizationSpoke/TokenizationSpoke.MaxGetters.t.sol
```diff
@@ -156,13 +156,13 @@ contract TokenizationSpokeMaxGettersAddCapVariablePartialTest is
 
   function test_maxDeposit_returnsRemaining() public view {
     uint256 capWithDecimals = uint256(addCap) * MathUtils.uncheckedExp(10, vault.decimals());
-    uint256 expected = capWithDecimals - vault.totalAssets();
+    uint256 expected = capWithDecimals - vault.previewMint(vault.totalSupply());
     assertEq(vault.maxDeposit(alice), expected);
   }
 
   function test_maxMint_returnsSharesOfRemaining() public view {
     uint256 capWithDecimals = uint256(addCap) * MathUtils.uncheckedExp(10, vault.decimals());
```
yan-man: capWithDecimals used a lot in this file, could create a helper here

src/spoke/TokenizationSpoke.sol
```diff
@@ -275,7 +275,7 @@ abstract contract TokenizationSpoke is ITokenizationSpoke, ERC20Upgradeable, Int
       return type(uint256).max;
     }
     uint256 allowed = config.addCap * ASSET_UNITS;
-    uint256 balance = totalAssets();
+    uint256 balance = previewMint(totalSupply());
```
yan-man, line 278: understand doing this to change the rounding, but it's a bit weird to do a previewMint on an asset amt instead of a shares amt. Maybe can leave a comment to explain this

DhairyaSethi: it's not an asset amount, it's a shares amount only 
totalSupply is the total shares issued by this spoke (and it should theoretically always be equal to hub.spokeAddedShares(address(this)))

if you mean the other way, previewMint(x) means how many assets will i need to mint x shares

yan-man: misread, thought that totalSupply was in asset

Kogaroshi: to avoid small misunderstandings like this one, maybe worth to add a quick comment here.

DhairyaSethi: the flow hasn't changed from before, totalAssets does `previewRedeem(totalShares())` so i think its fine as is

avniculae (approved):

miguelmtzinf (approved):

PR #1202, opened 2026-02-09
chore: precise validateAdd

DhairyaSethi:


tests/unit/Spoke/Spoke.Supply.t.sol
```diff
@@ -80,7 +80,7 @@ contract SpokeSupplyTest is SpokeBase {
   function test_supply_revertsWith_InvalidSupplyAmount() public {
```
yan-man, line 80: update test name

src/hub/Hub.sol
```diff
@@ -832,16 +831,16 @@ contract Hub is IHub, AccessManaged {
...
-    require(amount > 0, InvalidAmount());
+    require(shares > 0, InvalidShares());
     require(spoke.active, SpokeNotActive());
     require(!spoke.halted, SpokeHalted());
     uint256 addCap = spoke.addCap;
     require(
       addCap == MAX_ALLOWED_SPOKE_CAP ||
         addCap * MathUtils.uncheckedExp(10, asset.decimals) >=
-          asset.toAddedAssetsUp(spoke.addedShares) + amount,
+          asset.toAddedAssetsUp(spoke.addedShares + shares),
```
yan-man, line 843: i think need a test scenario for this specifically, where it passes on amount but now on shares hits AddCapExceeded

src/hub/Hub.sol
miguelmtzinf, line 869: pls add a test to show we don't need to do the same for `_validateDraw`

src/hub/Hub.sol
```diff
@@ -832,16 +831,16 @@ contract Hub is IHub, AccessManaged {
   function _validateAdd(
     Asset storage asset,
     SpokeData storage spoke,
-    uint256 amount
+    uint256 shares
```
miguelmtzinf, line 834:
```suggestion
    uint256 sharesAmount
```
wdyt

src/hub/Hub.sol
```diff
@@ -832,16 +831,16 @@ contract Hub is IHub, AccessManaged {
...
   ) internal view {
-    require(amount > 0, InvalidAmount());
+    require(shares > 0, InvalidShares());
```
miguelmtzinf, line 836: we re using `InvalidAmount` for similar checks (>0), so maybe better to keep `InvalidAmount`?

DhairyaSethi: its different here, will be incorrect bc at share price 2, amount 1 is also invalid bc it results in 0 shares added

DhairyaSethi: we do not need invalid amount here bc we compare shares, since share price > 1 (bc of virtual assets/shares)if shares are non 0 then assets have to be non 0

src/hub/Hub.sol
DhairyaSethi: not a fan 😅

src/hub/Hub.sol
DhairyaSethi: [6835ca5](https://github.com/aave/aave-v4/pull/1202/commits/6835ca516451a8d12707957331ab59d3b5f6ae08) lmk wdyt

src/hub/Hub.sol
DhairyaSethi: [6835ca5](https://github.com/aave/aave-v4/pull/1202/commits/6835ca516451a8d12707957331ab59d3b5f6ae08)

tests/unit/Hub/Hub.Draw.t.sol
```diff
@@ -458,4 +458,47 @@ contract HubDrawTest is HubBase {
...
+      premiumDelta: ZERO_PREMIUM_DELTA
+    });
+    vm.stopPrank();
+
+    Utils.add({hub: hub1, assetId: daiAssetId, caller: address(spoke2), amount: 10e18, user: bob});
+
+    uint40 drawCap = 5;
+    updateDrawCap(hub1, daiAssetId, address(spoke1), drawCap);
+
+    uint256 drawAmount = drawCap * 1e18 - 1;
```
miguelmtzinf, line 495: may be better to use `_decimals.dai`?

tests/unit/Hub/Hub.Draw.t.sol
```diff
@@ -458,4 +458,47 @@ contract HubDrawTest is HubBase {
...
+
+    vm.prank(address(spoke1));
+    hub1.draw({assetId: daiAssetId, amount: drawAmount, to: alice});
+
```
miguelmtzinf, line 499: we could add here that getTotalOwed into shares == drawAmount into shares

src/hub/Hub.sol
miguelmtzinf: was thinking from the user pov. If they add 0 assets or too little so it's less than a share, they will get `InvalidAmount` which is fair and clear. Getting `InvalidShares` may be a bit confusing, but i guess it's debatable, so no strong opinion.

tests/unit/Hub/Hub.Add.t.sol
```diff
@@ -803,4 +803,39 @@ contract HubAddTest is HubBase {
...
+      drawSpoke: address(spoke1),
+      drawAmount: 100e18,
+      skipTime: 365 days * 10
+    });
+
+    uint256 existingShares = hub1.getSpokeAddedShares(daiAssetId, address(spoke2));
+
+    // minimumAssetsPerAddedShare + 1: the +1 is truncated by floor rounding, minting exactly 1 share
+    uint256 addAmount = minimumAssetsPerAddedShare(hub1, daiAssetId) + 1;
+    assertEq(hub1.previewAddByAssets(daiAssetId, addAmount), 1);
```
avniculae, line 825: should we assert that hub.previewAddByShares(1) + 1 == addAmount as well?

tests/unit/Hub/Hub.Add.t.sol
```diff
@@ -803,4 +803,39 @@ contract HubAddTest is HubBase {
...
+
+    uint256 totalAddedAssets = hub1.getAddedAssets(daiAssetId);
+    uint256 totalAddedShares = hub1.getAddedShares(daiAssetId);
+    uint256 postStateAssets = (existingShares + 1).toAssetsUp(totalAddedAssets, totalAddedShares);
```
avniculae, line 829: why don't we use `previewAddByShares` here?

tests/unit/Hub/Hub.Add.t.sol
```diff
@@ -803,4 +803,39 @@ contract HubAddTest is HubBase {
...
+    uint40 newAddCap = ((postStateAssets / 1e18) + 1).toUint40();
```
avniculae, line 830: this tests succeeds on main as well, due to this rounding error

I'd suggest to find a testcase that fails on main but succeeds on this branch. we can also set newAddCap 1 wei smaller initially, expect revert, and then increase newAddCap by 1 and ensure it succeeds, wdyt?

tests/unit/Hub/Hub.Draw.t.sol
avniculae, line 495: why don't we draw exactly draw cap? and then latest try to draw 1?

PR #1203, opened 2026-02-09
fix : prevent _revokeRole invalid tracking

Kogaroshi:
Prevent usage of `_revokeRole()` through `renounceRole()` or `revokeRole()` to incorrectly track an `account` as role member.

### DIFF
--- a/tests/unit/AccessManagerEnumerable.t.sol
+++ b/tests/unit/AccessManagerEnumerable.t.sol
@@ -486,6 +486,24 @@ contract AccessManagerEnumerableTest is Test {
     assertEq(roleMembers[1], user3);
   }
 
+  function test_revokeRole_shouldNotTrack() public {
+    uint64 roleId = 1;
+    address user1 = makeAddr('user1');
+
+    (bool isMember, ) = accessManagerEnumerable.hasRole(roleId, user1);
+    assertFalse(isMember);
+    assertEq(accessManagerEnumerable.getRoleMemberCount(roleId), 0);
+
+    vm.prank(ADMIN);
+    accessManagerEnumerable.revokeRole(roleId, user1);
+
+    assertEq(accessManagerEnumerable.getRoleMemberCount(roleId), 0);
+    assertEq(accessManagerEnumerable.getRoleMembers(roleId, 0, 1).length, 0);
+
+    (isMember, ) = accessManagerEnumerable.hasRole(roleId, user1);
+    assertFalse(isMember);
+  }
+
   function test_renounceRole_shouldNotTrack() public {
     uint64 roleId = 1;
     address user1 = makeAddr('user1');


yan-man (approved):

src/access/AccessManagerEnumerable.sol
```diff
@@ -219,9 +219,12 @@ contract AccessManagerEnumerable is AccessManager, IAccessManagerEnumerable {
 
   /// @dev Overrides AccessManager `_revokeRole` function to tracked removed role members.
```
CheyenneAtapour:
```suggestion
  /// @dev Overrides AccessManager `_revokeRole` function to track removed role members.
```
this is what is meant, right?

tests/unit/AccessManagerEnumerable.t.sol
CheyenneAtapour, line 1: What about a test where the user has the role, and then renounces it?

Kogaroshi: f3fab7d775af4e1904098cb10ea6713e55231d81

src/access/AccessManagerEnumerable.sol
Kogaroshi: f3fab7d775af4e1904098cb10ea6713e55231d81

src/access/AccessManagerEnumerable.sol
```diff
@@ -217,11 +217,14 @@ contract AccessManagerEnumerable is AccessManager, IAccessManagerEnumerable {
     return granted;
   }
 
-  /// @dev Overrides AccessManager `_revokeRole` function to tracked removed role members.
+  /// @dev Overrides AccessManager `_revokeRole` function to track removed role members.
   function _revokeRole(uint64 roleId, address account) internal override returns (bool) {
+    (bool hadRole, ) = hasRole(roleId, account);
```
DhairyaSethi: i think this is still incorrect bc original revokeRole shortcircuit is based on `since != 0` whereas this uses `since != 0 && since <= now`

DhairyaSethi: i know we said we might not use this functionality but better to still be accurate here?
how about only remove tracking when _revokeRole succeeded 

```solidity
function _revokeRole(uint64 roleId, address account) internal override returns (bool) {
  bool revoked = super._revokeRole(roleId, account);
  if (revoked) {
    _trackRoleMember(roleId, account, false);
  }
  return revoked;
}
```

DhairyaSethi: this is similar to how we're it on `grantRole` (there also i think we can target a `true` as the last org or add wrappers around `_trackRoleMember` which are `_track` & `_untrack` or smthing

Kogaroshi: Yes, this should be simple, making the change

Kogaroshi: 205226499b8378f73cc29eca7ff8ec10c79f8f3d

miguelmtzinf (approved):

PR #1210, opened 2026-02-10
0.5.6 <- 0.5.9

DhairyaSethi:


src/access/AccessManagerEnumerable.sol
```diff
@@ -9,27 +9,91 @@ import {IAccessManagerEnumerable} from 'src/access/interfaces/IAccessManagerEnum
...
 /// @notice Extension of AccessManager that tracks role members and their function selectors using EnumerableSet.
+/// @dev Roles, target contracts, and function selectors assigned to `ADMIN_ROLE` are excluded from tracking.
 contract AccessManagerEnumerable is AccessManager, IAccessManagerEnumerable {
-  using EnumerableSet for EnumerableSet.AddressSet;
-  using EnumerableSet for EnumerableSet.Bytes32Set;
+  using EnumerableSet for *;
+
+  /// @dev Set of all role identifiers.
+  /// @dev `PUBLIC_ROLE` and `ADMIN_ROLE` are not part of this set.
+  EnumerableSet.UintSet private _rolesSet;
```
StErMi: [💡 SUGGESTION] Maybe it's implicit by the fact that you had to create a start/end getter for `_rolesSet` and `_adminRolesSet` but it's worth adding a dev comment that underlying the fact that these sets are **always growing** sets where items added are **never** removed, even when they should.

src/access/AccessManagerEnumerable.sol
```diff
@@ -38,77 +102,195 @@ contract AccessManagerEnumerable is AccessManager, IAccessManagerEnumerable {
...
   ) internal override {
-    uint64 oldRoleId = getTargetFunctionRole(target, selector);
     super._setTargetFunctionRole(target, selector, roleId);
+
+    _trackRoleTargetSelector(roleId, target, selector);
+  }
+
+  /// @dev Tracks all role identifiers when a new role is created.
+  function _trackRole(uint64 roleId) internal {
+    if (roleId == ADMIN_ROLE) {
```
StErMi: [💡 SUGGESTION] I think that it's impossible, with the current logic and how OZ `AccessManager` works to enter `_trackRole` when `roleId == PUBLIC_ROLE`. `_setRoleAdmin` and `_setRoleGuardian` reverts when `roleId == ADMIN_ROLE || roleId == PUBLIC_ROLE` and `_grantRole` reverts when `roleId == PUBLIC_ROLE`.

I know that you managed "manually" the upgrade to OZ libs, but maybe, to be "future proof" and avoid any possible issue I would also add the early revert when `roleId == PUBLIC_ROLE` and call it a day.
In the natspec you are quite explicit: "/// @dev `PUBLIC_ROLE` and `ADMIN_ROLE` are not part of this set."

src/access/AccessManagerEnumerable.sol
StErMi: I think that it should also be documented that the fact that `_rolesSet` has a role tracked it **does not** mean that the role has members. `_trackRole` is also called when `_setRoleAdmin` and `_setRoleGuardian` are executed, those functions can be called even if the role has not been granted to anyone.

src/access/AccessManagerEnumerable.sol
```diff
@@ -38,77 +102,195 @@ contract AccessManagerEnumerable is AccessManager, IAccessManagerEnumerable {
...
+  /// @dev Overrides AccessManager `_setTargetFunctionRole` function to track function selectors attributed to roles.
   function _setTargetFunctionRole(
     address target,
     bytes4 selector,
     uint64 roleId
   ) internal override {
-    uint64 oldRoleId = getTargetFunctionRole(target, selector);
     super._setTargetFunctionRole(target, selector, roleId);
+
+    _trackRoleTargetSelector(roleId, target, selector);
```
StErMi: [❓ QUESTION] Should this function also call the `_trackRole` function?

src/hub/Hub.sol
StErMi: [💡 SUGGESTION] I'm not sure if I have already suggested it, but it would be worth it to validate if `underlying` is also a **valid** and **existing** contract. In the `addAsset` logic is never invoked and `HubConfigurator` could call `addAsset` via `addAssetWithDecimals` which also doesn't invoke any function on `underlying`.

PR #1213, opened 2026-02-12
opt : skip RP update in withdraw if no debt

Kogaroshi:
Gas savings : 
Check that `onBehalfOf` is currently borrowing before executing
 ```     
uint256 newRiskPremium = _refreshAndValidateUserAccountData(onBehalfOf).riskPremium;
```
in `withdraw()`
Saves between 18k - 23k gas when user is not borrowing, minimal increase when he is.

### CODE REVIEW

File: src/spoke/Spoke.sol [hunk 1]
```diff
@@ -261,7 +261,10 @@ abstract contract Spoke is
 
     userPosition.suppliedShares -= withdrawnShares.toUint120();
 
-    if (_positionStatus[onBehalfOf].isUsingAsCollateral(reserveId)) {
+    PositionStatus storage positionStatus = _positionStatus[onBehalfOf];
+    if (
+      positionStatus.isUsingAsCollateral(reserveId) && positionStatus.borrowCount(_reserveCount) > 0
```
DhairyaSethi: maybe add a isBorrowingAny() which will be cheaper bc it can return false early when it finds it in the first word, and also doesn't need to maintain a counter

src/spoke/Spoke.sol
```diff
@@ -261,7 +261,10 @@ abstract contract Spoke is
...
+    ) {
       uint256 newRiskPremium = _refreshAndValidateUserAccountData(onBehalfOf).riskPremium;
```
DhairyaSethi, line 268: when i initially thought about it, the opt went in `_processUserAccountData` as a short circuit

DhairyaSethi: smth along these lines https://github.com/aave-dao/aave-v3-origin/blob/v3.6.0/src/contracts/protocol/libraries/logic/GenericLogic.sol#L72-L74

src/spoke/Spoke.sol
Kogaroshi: c4b68471b896a8306164d55ebdff53138ef740d5

src/spoke/Spoke.sol
miguelmtzinf: debatable but we could assume a user has borrows if the asset is enabled as collateral. They could skip hf check by disabling the asset as collateral first (that runs HF check anyway) or not enabling it from the start.

This opt would benefit depositors (only; with no borrow intention), and not sure it's needed given that `supply` does not activate the asset as collateral by default.

Kogaroshi:
Decided not to apply and favor codesize vs gas optimization

PR #1216, opened 2026-02-17
feat: Add `FeeSharesMinter` contract

CheyenneAtapour:
Adds a fee minter contract which will enforce specific conditions for calling mintFeeShares() on the hub, and allow for mintFeeShares() to be called consistently


src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
...
+import {Ownable} from 'src/dependencies/openzeppelin/Ownable.sol';
+
+/// @title FeeSharesMinterBase
+/// @author Aave Labs
+/// @notice Contract to mint fee shares on the Hub when specific conditions are met.
+contract FeeSharesMinterBase is Ownable {
+  struct MintConfig {
+    uint256 minTimeInterval;
+    uint256 minUnrealizedFeePercent; // 1e4 = 100% (basis points)
+  }
```
Kogaroshi: could be packed tighter since `minUnrealizedFeePercent` is in BPS

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
...
+  /// @dev Constructor.
+  /// @param owner The owner of the contract.
+  constructor(address owner) Ownable(owner) {}
+
+  /// @notice Sets the automation configuration for a specific asset.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
+  /// @param config The new configuration.
+  function setConfig(address hub, uint256 assetId, MintConfig memory config) external onlyOwner {
+    _configs[hub][assetId] = config;
```
Kogaroshi: Do we want to add input checks there, tho it's owner gated, a missconfig can be troublesome

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    IHub hubContract = IHub(hub);
+    uint256 accruedFees = hubContract.getAssetAccruedFees(assetId);
+
+    uint256 totalAddedAssets = hubContract.getAddedAssets(assetId);
+    if (totalAddedAssets == 0) return false;
+
+    // Check if accruedFees / totalAddedAssets >= minUnrealizedFeePercent (in bps)
+    if ((accruedFees * 10000) / totalAddedAssets < config.minUnrealizedFeePercent) {
```
Kogaroshi: Add a constant instead of `10000` there

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
...
+      return false;
+    }
+
+    // Ensure at least 1 fee share is minted
+    uint256 expectedShares = hubContract.previewAddByAssets(assetId, accruedFees);
+    if (expectedShares < 1) {
```
Kogaroshi: `if (expectedShares == 0)` ?

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
...
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+import {Ownable} from 'src/dependencies/openzeppelin/Ownable.sol';
+
+/// @title FeeSharesMinterBase
+/// @author Aave Labs
+/// @notice Contract to mint fee shares on the Hub when specific conditions are met.
+contract FeeSharesMinterBase is Ownable {
```
Kogaroshi: Worth to inherit Rescuable as well, just in case asset are sent here by mistake

tests/unit/FeeSharesMinterBase.t.sol
```diff
@@ -0,0 +1,188 @@
...
+      minTimeInterval: 1 days,
+      minUnrealizedFeePercent: 100 // 1%
+    });
+
+    vm.prank(bob);
+    vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, bob));
+    minter.setConfig(address(hub1), daiAssetId, config);
+  }
+
+  function test_execute_success() public {
```
Kogaroshi: can we also add a fuzz test there ?

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
...
+    emit ConfigUpdated(hub, assetId, config);
+  }
+
+  /// @notice Executes the fee minting if conditions are met.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
+  function execute(address hub, uint256 assetId) external {
+    if (!_checkExecute(hub, assetId)) {
+      revert ConditionsNotMet();
+    }
```
DhairyaSethi: require style

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
...
+  /// @return True if conditions are met, false otherwise.
+  function checkExecute(address hub, uint256 assetId) external view returns (bool) {
+    return _checkExecute(hub, assetId);
+  }
+
+  /// @dev Internal function to check execution conditions.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
+  /// @return True if conditions are met, false otherwise.
+  function _checkExecute(address hub, uint256 assetId) internal view returns (bool) {
```
DhairyaSethi:
```suggestion
  function _checkExecute(address hub, uint256 assetId) internal virtual view returns (bool) {
```

src/hub/FeeSharesMinterBase.sol
DhairyaSethi: call this a steward maybe

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+import {Ownable} from 'src/dependencies/openzeppelin/Ownable.sol';
```
DhairyaSethi: always use Ownable2Step pls

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: idk about Steward, but still open to suggestions about names

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: Sure, tho I think 0 is fine for both configs. I can surely check the `minUnrealizedFeePercent` is under 100%, but then unsure how to bound the time interval. I was thinking maybe 1 year is fine? Likely we won't go over that

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: Done

tests/unit/FeeSharesMinterBase.t.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
Kogaroshi: `minTimeInterval` is still `uint256` tho, so still taking 2 slots there.

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,109 @@
...
+    // Check mint interval
+    if (block.timestamp - lastMintTime[hub][assetId] < config.minTimeInterval) {
+      return false;
+    }
+
+    IHub hubContract = IHub(hub);
+    uint256 accruedFees = hubContract.getAssetAccruedFees(assetId);
+
+    uint256 totalAddedAssets = hubContract.getAddedAssets(assetId);
+    if (totalAddedAssets == 0) return false;
```
DhairyaSethi: why? we can avoid the division below by making it mul on rhs. for active market to become empty this would still never be zero bc some assets will belong to virtual shares so is purely on market creation/non borrowable empty? i would rm from here

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    // Check if accruedFees / totalAddedAssets >= minUnrealizedFeePercent (in bps)
+    if ((accruedFees * MAX_BPS) / totalAddedAssets < config.minUnrealizedFeePercent) {
```
DhairyaSethi: max_bps not needed, it's percentage factor from percentage math
pls using percentDiv{explicitRounding} instead

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,109 @@
...
+      return false;
+    }
+
+    // Ensure at least 1 fee share is minted
+    uint256 expectedShares = hubContract.previewAddByAssets(assetId, accruedFees);
+    if (expectedShares == 0) {
+      return false;
+    }
+
+    return true;
```
DhairyaSethi:
```suggestion
    // Ensure at least 1 fee share is minted
    uint256 expectedShares = hubContract.previewAddByAssets(assetId, accruedFees);
    return expectedShares > 0;
```

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,109 @@
...
+
+  /// @dev Internal function to check execution conditions.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
+  /// @return True if conditions are met, false otherwise.
+  function _checkExecute(address hub, uint256 assetId) internal view virtual returns (bool) {
+    MintConfig memory config = _configs[hub][assetId];
+
+    // Check mint interval
+    if (block.timestamp - lastMintTime[hub][assetId] < config.minTimeInterval) {
```
DhairyaSethi: i dont think we should place a time based min interval, but rather amount based instead?

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,109 @@
...
+  uint256 public constant MAX_BPS = 1e4;
+  uint256 public constant MAX_TIME_INTERVAL = 365 days;
+
+  mapping(address => mapping(uint256 => MintConfig)) internal _configs;
+  mapping(address => mapping(uint256 => uint256)) public lastMintTime;
+
+  event ConfigUpdated(address indexed hub, uint256 indexed assetId, MintConfig config);
+
+  error ConditionsNotMet();
+  error InvalidConfig();
```
DhairyaSethi: move to interface, natspec them, max_bps not needed

src/hub/FeeSharesMinterBase.sol
DhairyaSethi: @CheyenneAtapour

src/hub/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {Ownable} from 'src/dependencies/openzeppelin/Ownable.sol';
+import {Ownable2Step} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
```
DhairyaSethi:
```suggestion
import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
```

src/utils/FeeSharesMinter.sol
DhairyaSethi, line 1: probably should be in `utils/`

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: I think the amount does make more sense. cc @miguelmtzinf on this

src/utils/FeeSharesMinter.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Removed

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
```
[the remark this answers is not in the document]
Kogaroshi: Valid one, please change to match other contracts @CheyenneAtapour

src/utils/IFeeSharesMinterBase.sol
```diff
@@ -0,0 +1,60 @@
...
+
+/// @title IFeeSharesMinterBase
+/// @author Aave Labs
+/// @notice Interface for the FeeSharesMinterBase contract
+interface IFeeSharesMinterBase {
+  /// @notice Configuration for automated fee share minting on a specific asset.
+  /// @param minTimeInterval Minimum number of seconds that must elapse between mint executions.
+  /// @param minUnrealizedFeePercent Minimum ratio of accrued fees to total assets, in bps.
+  struct MintConfig {
+    uint32 minTimeInterval;
```
Kogaroshi: `uint48` ? purely so 4 structs packed together fit exactly in an 32 bytes.
(alternatively, can increase the `unit16` as well too)

src/utils/FeeSharesMinterBase.sol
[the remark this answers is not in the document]
CheyenneAtapour: Done

src/utils/IFeeSharesMinterBase.sol
CheyenneAtapour: Done

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+      return false;
+    }
+
+    IHub hubContract = IHub(hub);
+    uint256 accruedFees = hubContract.getAssetAccruedFees(assetId);
+    uint256 totalAddedAssets = hubContract.getAddedAssets(assetId);
+
+    // Check if accruedFees / totalAddedAssets >= minUnrealizedFeePercent (in bps)
+    if (
+      PercentageMath.percentDivDown(accruedFees, totalAddedAssets) < config.minUnrealizedFeePercent
```
CheyenneAtapour: minAccruedFeesPercent instead of config.minUnrealizedFeePercent

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+// Copyright (c) 2025 Aave Labs
+pragma solidity 0.8.28;
+
+import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
+import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
+import {Rescuable} from 'src/utils/Rescuable.sol';
+import {IFeeSharesMinterBase} from 'src/utils/IFeeSharesMinterBase.sol';
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+
+/// @title FeeSharesMinterBase
```
yan-man: wondering why this is base? is it meant to be extended

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+/// @author Aave Labs
+/// @notice Contract to mint fee shares on the Hub when specific conditions are met.
+contract FeeSharesMinterBase is IFeeSharesMinterBase, Ownable2Step, Rescuable {
+  /// @inheritdoc IFeeSharesMinterBase
+  uint256 public constant MAX_TIME_INTERVAL = 365 days;
+
+  /// @inheritdoc IFeeSharesMinterBase
+  mapping(address => mapping(uint256 => uint256)) public lastMintTime;
```
yan-man: can we add the vars into the mapping, like 

```suggestion
  mapping(address hub => mapping(uint256 assetId => uint256 timestamp)) public lastMintTime;
```

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+import {Rescuable} from 'src/utils/Rescuable.sol';
+import {IFeeSharesMinterBase} from 'src/utils/IFeeSharesMinterBase.sol';
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+
+/// @title FeeSharesMinterBase
+/// @author Aave Labs
+/// @notice Contract to mint fee shares on the Hub when specific conditions are met.
+contract FeeSharesMinterBase is IFeeSharesMinterBase, Ownable2Step, Rescuable {
+  /// @inheritdoc IFeeSharesMinterBase
+  uint256 public constant MAX_TIME_INTERVAL = 365 days;
```
yan-man: do we want to make this immutable and set in constructor?

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+      config.minUnrealizedFeePercent <= PercentageMath.PERCENTAGE_FACTOR &&
+        config.minTimeInterval <= MAX_TIME_INTERVAL,
+      InvalidConfig()
+    );
+    _configs[hub][assetId] = config;
+    emit ConfigUpdated(hub, assetId, config);
+  }
+
+  /// @inheritdoc IFeeSharesMinterBase
+  function execute(address hub, uint256 assetId) external {
```
yan-man: perhaps prefer a diff fn name that's more descriptive, like mintFeeShares

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+    require(_checkExecute(hub, assetId), ConditionsNotMet());
```
yan-man: i think L40-43 can be in a helper as it's re-used by performUpkeep

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+    IHub(hub).mintFeeShares(assetId);
+  }
+
+  /// @inheritdoc IFeeSharesMinterBase
+  function checkUpkeep(
+    bytes calldata checkData
+  ) external view override returns (bool upkeepNeeded, bytes memory performData) {
+    (address hub, uint256 assetId) = abi.decode(checkData, (address, uint256));
+    upkeepNeeded = _checkExecute(hub, assetId);
+    performData = checkData;
```
yan-man: in other places we'd had explicit return statement, even if we had named return vars

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  /// @inheritdoc IFeeSharesMinterBase
+  function checkExecute(address hub, uint256 assetId) external view returns (bool) {
+    return _checkExecute(hub, assetId);
+  }
+
+  /// @dev Internal function to check execution conditions.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
```
yan-man: dont think we need these 2 param natspec for this given its internal fn but nbd

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+  /// @inheritdoc IFeeSharesMinterBase
+  function checkExecute(address hub, uint256 assetId) external view returns (bool) {
+    return _checkExecute(hub, assetId);
+  }
+
+  /// @dev Internal function to check execution conditions.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
+  /// @return True if conditions are met, false otherwise.
+  function _checkExecute(address hub, uint256 assetId) internal view virtual returns (bool) {
```
yan-man: would rather prefer this as hubAddress, and below as `IHub hub = IHub(hubAddress)` for consistency

src/utils/IFeeSharesMinterBase.sol
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
```
yan-man: now with updated license

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,107 @@
...
+    // Check mint interval
+    if (block.timestamp - lastMintTime[hub][assetId] < config.minTimeInterval) {
+      return false;
+    }
+
+    IHub hubContract = IHub(hub);
+    uint256 accruedFees = hubContract.getAssetAccruedFees(assetId);
+    uint256 totalAddedAssets = hubContract.getAddedAssets(assetId);
+
+    // Check if accruedFees / totalAddedAssets >= minUnrealizedFeePercent (in bps)
```
yan-man: other places we've used BPS over bps in natspec

src/utils/IFeeSharesMinterBase.sol
```diff
@@ -0,0 +1,75 @@
...
+
+import {AutomationCompatibleInterface} from 'src/dependencies/chainlink/AutomationCompatibleInterface.sol';
+
+/// @title IFeeSharesMinterBase
+/// @author Aave Labs
+/// @notice Interface for the FeeSharesMinterBase contract
+interface IFeeSharesMinterBase is AutomationCompatibleInterface {
+  /// @notice Configuration for automated fee share minting on a specific asset.
+  /// @param minTimeInterval Minimum number of seconds that must elapse between mint executions.
+  /// @param minUnrealizedFeePercent Minimum ratio of accrued fees to total assets, in bps.
```
yan-man: BPS over bps

src/utils/IFeeSharesMinter.sol
```diff
@@ -0,0 +1,75 @@
...
+  /// @param assetId The identifier of the asset.
+  /// @param config The new configuration to apply.
+  function setConfig(address hub, uint256 assetId, MintConfig memory config) external;
+
+  /// @notice Executes fee share minting if all conditions are met.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
+  function execute(address hub, uint256 assetId) external;
+
+  /// @notice Chainlink Automation on-chain execution entry point.
```
yan-man: nbd but just need to align on onchain vs on-chain consistency

src/utils/IFeeSharesMinterBase.sol
```diff
@@ -0,0 +1,75 @@
...
+  function execute(address hub, uint256 assetId) external;
+
+  /// @notice Chainlink Automation on-chain execution entry point.
+  /// @dev performData must be abi.encoded as (address hub, uint256 assetId).
+  /// @inheritdoc AutomationCompatibleInterface
+  function performUpkeep(bytes calldata performData) external;
+
+  /// @notice Chainlink Automation off-chain simulation check.
+  /// @dev checkData must be abi.encoded as (address hub, uint256 assetId).
+  /// @dev Returns upkeepNeeded=true and the same bytes as performData when conditions are met.
```
yan-man:
```suggestion
  /// @dev Returns whether upkeep is needed and the performData in bytes when conditions are met.
```

reads a bit weird. Maybe reword, or rm this line completely and only rely on AutomationCompatibleInterface docs, or explicitly define the @return vars in the natspec

src/utils/IFeeSharesMinterBase.sol
```diff
@@ -0,0 +1,75 @@
...
+  /// @notice Returns the current automation configuration for a specific asset.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
+  /// @return The stored `MintConfig` struct.
+  function getConfig(address hub, uint256 assetId) external view returns (MintConfig memory);
+
+  /// @notice Returns the last timestamp at which fee shares were minted for a given asset.
+  /// @param hub The address of the hub.
+  /// @param assetId The identifier of the asset.
+  /// @return The block timestamp of the last successful `execute` call.
```
yan-man:
```suggestion
  /// @return The timestamp of the last successful `execute` call.
```

tests/contracts/utils/FeeSharesMinterBase.t.sol
```diff
@@ -0,0 +1,525 @@
...
+import 'tests/unit/Hub/HubBase.t.sol';
+import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
+import {FeeSharesMinterBase} from 'src/utils/FeeSharesMinterBase.sol';
+import {IFeeSharesMinterBase} from 'src/utils/IFeeSharesMinterBase.sol';
+import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
+
+contract FeeSharesMinterBaseTest is HubBase {
+  using SafeCast for uint256;
+
+  FeeSharesMinterBase internal minter;
```
yan-man: public or _minter

tests/contracts/utils/FeeSharesMinterBase.t.sol
```diff
@@ -0,0 +1,525 @@
...
+
+  function setUp() public override {
+    super.setUp();
+    minter = new FeeSharesMinterBase(ADMIN);
+
+    // Grant minter the HUB_ADMIN_ROLE so it can call mintFeeShares
+    vm.prank(ADMIN);
+    accessManager.grantRole(Roles.HUB_ADMIN_ROLE, address(minter), 0);
```
yan-man: role will also need to update to HUB_DOMAIN_ADMIN_ROLE w latest changes

tests/contracts/utils/FeeSharesMinter.t.sol
```diff
@@ -0,0 +1,525 @@
...
+      minTimeInterval: 0,
+      minUnrealizedFeePercent: minUnrealizedFeePercent
+    });
+
+    vm.prank(ADMIN);
+    vm.expectRevert(IFeeSharesMinterBase.InvalidConfig.selector);
+    minter.setConfig(address(hub1), daiAssetId, config);
+  }
+
+  function test_rescueToken() public {
```
yan-man: nit: here could define 1000e18 in a cached var

tests/contracts/utils/FeeSharesMinterBase.t.sol
```diff
@@ -0,0 +1,525 @@
...
+    vm.expectRevert(IFeeSharesMinterBase.ConditionsNotMet.selector);
+    minter.performUpkeep(performData);
+  }
+
+  function test_performUpkeep_revertsWith_ConditionsNotMet_percentThresholdNotMet_withMinShares()
+    public
+  {
+    IFeeSharesMinterBase.MintConfig memory config = IFeeSharesMinterBase.MintConfig({
+      minTimeInterval: 0,
+      minUnrealizedFeePercent: 5000
```
yan-man:
```suggestion
      minUnrealizedFeePercent: 50_00
```

tests/contracts/utils/FeeSharesMinterBase.t.sol
```diff
@@ -0,0 +1,525 @@
...
+      drawAmount: 100e18,
+      skipTime: 365 days
+    });
+
+    uint256 fees = hub1.getAssetAccruedFees(daiAssetId);
+    uint256 totalAssets = hub1.getAddedAssets(daiAssetId);
+
+    assertGt(fees, 0, 'Fees must be nonzero');
+    assertGt(hub1.previewAddByAssets(daiAssetId, fees), 0, 'At least 1 share would be minted');
+    assertLt(fees, totalAssets / 2, 'Fees must be < 50% of total');
```
yan-man:
```suggestion
    assertLt(fees, totalAssets.percentMulDown(config.minUnrealizedFeePercent), 'Fees must be < 50% of total');
```

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: Yeah, this was meant to be a simple version, but it's also possible to just make a brand new one too with other logic. Up for discussion

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: our pattern for mappings seems to be that we do not label the last variable, right?

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: names were meant to match this contract: https://github.com/aave-dao/gho-origin/blob/main/src/contracts/facilitators/gsm/swapFreezer/OracleSwapFreezerBase.sol

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: hmm, but I'm just using "hub" variable as address type consistently in this contract

src/utils/FeeSharesMinter.sol
CheyenneAtapour: I think nice cause it's virtual to have natspec

src/utils/IFeeSharesMinterBase.sol
CheyenneAtapour: why not block timestamp to be more precise?

tests/contracts/utils/FeeSharesMinterBase.t.sol
CheyenneAtapour: These changes haven't happened yet, right? This role is not yet in the roles library

yan-man: sorry meant latest *planned updates, yes it is still an open pr atm https://github.com/aave/aave-v4/pull/1287

src/utils/IFeeSharesMinterBase.sol
yan-man: just for consistency with like IHub where we have `lastUpdateTimestamp The timestamp of the last accrual`

src/utils/FeeSharesMinterBase.sol
yan-man: mmm good pt, not sure whether worthwhile to update everything

src/utils/FeeSharesMinterBase.sol
yan-man: i think we've done it in most places, just when struct is the last var it gets skipped. But (for me) would prefer everywhere if possible

tests/contracts/utils/FeeSharesMinterBase.t.sol
```diff
@@ -0,0 +1,544 @@
...
+    assertFalse(_minter.checkExecute(address(hub1), daiAssetId));
+
+    vm.expectRevert(IFeeSharesMinterBase.ConditionsNotMet.selector);
+    _minter.execute(address(hub1), daiAssetId);
+  }
+
+  function test_execute_revertsWith_ConditionsNotMet_PercentThresholdNotMet() public {
+    IFeeSharesMinterBase.MintConfig memory config = IFeeSharesMinterBase.MintConfig({
+      minTimeInterval: 0,
+      minAccruedFeesPercent: 5000 // 50% threshold
```
yan-man:
```suggestion
      minAccruedFeesPercent: 50_00 // 50% threshold
```

tests/contracts/utils/FeeSharesMinterBase.t.sol
```diff
@@ -0,0 +1,544 @@
...
+      drawAmount: 100e18,
+      skipTime: 365 days
+    });
+
+    uint256 fees = hub1.getAssetAccruedFees(daiAssetId);
+    uint256 totalAssets = hub1.getAddedAssets(daiAssetId);
+
+    assertGt(fees, 0, 'Fees must be nonzero');
+    assertGt(hub1.previewAddByAssets(daiAssetId, fees), 0, 'At least 1 share would be minted');
+    assertLt(fees, totalAssets / 2, 'Fees must be < 50% of total');
```
yan-man: also should be the percentMulDown?

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: Ah ok you are right - Done

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: cc @DhairyaSethi wdyt?

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: Done

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: Yeah, good catch. Done

src/utils/IFeeSharesMinterBase.sol
CheyenneAtapour: Done

src/utils/IFeeSharesMinterBase.sol
CheyenneAtapour: I applied it as your suggestion, but perhaps we don't even need it. Open to whatever you think is best

src/utils/IFeeSharesMinterBase.sol
CheyenneAtapour: Ok cool - Done

src/utils/IFeeSharesMinterBase.sol
CheyenneAtapour: Done

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: Done

tests/contracts/utils/FeeSharesMinterBase.t.sol
CheyenneAtapour: Done, was actually the HUB_FEE_MINTER_ROLE btw

tests/contracts/utils/FeeSharesMinter.t.sol
CheyenneAtapour: Done

tests/contracts/utils/FeeSharesMinterBase.t.sol
CheyenneAtapour: I think this is fine, I don't wanna check the exact same math as src against the exact same math here

tests/contracts/utils/FeeSharesMinterBase.t.sol
CheyenneAtapour: Done

src/utils/FeeSharesMinterBase.sol
CheyenneAtapour: I removed the time component now, but kept the ratio

src/hub/FeeSharesMinterBase.sol
CheyenneAtapour: The word "keeper" is used for these. I'm open to that

src/utils/IFeeSharesMinterBase.sol
yan-man: dont feel too strongly

src/utils/FeeSharesMinterBase.sol
```diff
@@ -0,0 +1,87 @@
...
+import {Rescuable} from 'src/utils/Rescuable.sol';
+import {IFeeSharesMinterBase} from 'src/utils/IFeeSharesMinterBase.sol';
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+
+/// @title FeeSharesMinterBase
+/// @author Aave Labs
+/// @notice Contract to mint fee shares on the Hub when specific conditions are met.
+contract FeeSharesMinterBase is IFeeSharesMinterBase, Ownable2Step, Rescuable {
+  mapping(address hub => mapping(uint256 assetId => uint16 minAccruedFeesPercent))
+    internal _configs;
```
Kogaroshi: should be renamed now ?

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,87 @@
...
+  /// @return True if conditions are met, false otherwise.
+  function _checkUpkeep(address hub, uint256 assetId) internal view virtual returns (bool) {
+    uint16 minAccruedFeesPercent = _configs[hub][assetId];
+
+    IHub hubContract = IHub(hub);
+    uint256 accruedFees = hubContract.getAssetAccruedFees(assetId);
+    uint256 totalAddedAssets = hubContract.getAddedAssets(assetId);
+
+    if (PercentageMath.percentDivDown(accruedFees, totalAddedAssets) < minAccruedFeesPercent) {
```
yan-man: let's say we do this for an asset which has no added assets yet, will this overflow or revert if totalAddedAssets == 0, ie div by 0

src/utils/FeeSharesMinterBase.sol
yan-man: to what?

Kogaroshi: `minAccruedFeesPercent` ? it is not a struct anymore, just an uint16, calling it "configs" seems strange when there is only 1 parameter to configure

src/utils/FeeSharesMinter.sol
Kogaroshi: yeah, I think better to return false early if `totalAddedAssets == 0`

src/utils/IFeeSharesMinter.sol
Kogaroshi: we seem to use on-chain in other parts of the repo, so I'd keep as is.

src/utils/FeeSharesMinter.sol
Kogaroshi: 7bb20ece4c38bb5a75e4be26058990caabd393cb

src/utils/FeeSharesMinterBase.sol
Kogaroshi: 7bb20ece4c38bb5a75e4be26058990caabd393cb

src/utils/IFeeSharesMinter.sol
yan-man: i dont really mind either way but in JP's tech docs it's using onchain: https://github.com/aave/aave-v4/pull/1297/changes

@CanonicalJP thoughts?

src/utils/IFeeSharesMinter.sol
```diff
@@ -0,0 +1,46 @@
...
+pragma solidity 0.8.28;
+
+import {AutomationCompatibleInterface} from 'src/dependencies/chainlink/AutomationCompatibleInterface.sol';
+
+/// @title IFeeSharesMinter
+/// @author Aave Labs
+/// @notice Interface for the FeeSharesMinter contract
+interface IFeeSharesMinter is AutomationCompatibleInterface {
+  /// @notice Emitted when the configuration for an asset is updated.
+  /// @param hub The address of the hub.
```
yan-man:
```suggestion
  /// @param hub The address of the Hub.
```

src/utils/IFeeSharesMinter.sol
```diff
@@ -0,0 +1,46 @@
...
+  event ConfigUpdated(address indexed hub, uint256 indexed assetId, uint16 minAccruedFeesPercent);
+
+  /// @notice Thrown upon minting when the required conditions are not met.
+  error ConditionsNotMet();
+
+  /// @notice Thrown when `setConfig` is called with an invalid value.
+  error InvalidConfig();
+
+  /// @notice Sets the minimum accrued fees percent for a specific asset.
+  /// @param hub The address of the hub.
```
yan-man:
```suggestion
  /// @param hub The address of the Hub.
```

src/utils/IFeeSharesMinter.sol
```diff
@@ -0,0 +1,46 @@
...
+  /// @notice Chainlink Automation off-chain simulation check.
+  /// @dev checkData must be abi.encoded as (address hub, uint256 assetId).
+  /// @dev Returns whether upkeep is needed and the performData in bytes when conditions are met.
+  /// @inheritdoc AutomationCompatibleInterface
+  function checkUpkeep(
+    bytes calldata checkData
+  ) external view returns (bool upkeepNeeded, bytes memory performData);
+
+  /// @notice Returns the minimum accrued fees percent for a specific asset.
+  /// @param hub The address of the hub.
```
yan-man:
```suggestion
  /// @param hub The address of the Hub.
```

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,89 @@
...
+
+    if (totalAddedAssets == 0) {
+      return false;
+    }
+    if (PercentageMath.percentDivDown(accruedFees, totalAddedAssets) < minAccruedFeesPercent) {
+      return false;
+    }
+
+    // Ensure at least 1 fee share would be minted
+    uint256 expectedShares = hubContract.previewAddByAssets(assetId, accruedFees);
```
yan-man: ok so if there is an unconfigured hub/asset pair here it would always return true whenever fee shares is > 0 on the hub, bc the default minAccruedFeesPercent is 0. If we accidentally register it on keepers before configuring it here, we would end up constantly minting when shares > 0, which could be wasteful on LINK. 

I think we could short circuit and return false if minAccruedFeesPercent == 0

tests/deployments/batches/AaveV4FeeSharesMinterBatch.t.sol
yan-man, line 1: tbd update readme

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,95 @@
...
+import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
+import {Rescuable} from 'src/utils/Rescuable.sol';
+import {IFeeSharesMinter} from 'src/utils/IFeeSharesMinter.sol';
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+
+/// @title FeeSharesMinter
+/// @author Aave Labs
+/// @notice Contract to mint fee shares on the Hub when specific conditions are met.
+contract FeeSharesMinter is IFeeSharesMinter, Ownable2Step, Rescuable {
+  mapping(address hub => mapping(uint256 assetId => uint16)) internal _minAccruedFeesPercent;
```
avniculae, line 18: I think we should name last value as well

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,95 @@
...
+    _performUpkeep(hub, assetId);
+  }
+
+  /// @inheritdoc IFeeSharesMinter
+  function checkUpkeep(
+    bytes calldata checkData
+  ) external view override returns (bool, bytes memory) {
+    (address hub, uint256 assetId) = abi.decode(checkData, (address, uint256));
+    bool upkeepNeeded = _checkUpkeep(hub, assetId);
+    bytes memory performData = checkData;
```
avniculae: we can store `checkdata` in memory in the args as well, to avoid needing this, wdyt?

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,95 @@
...
+  /// @param hub The address of the Hub.
+  /// @param assetId The identifier of the asset.
+  /// @return True if conditions are met, false otherwise.
+  function _checkUpkeep(address hub, uint256 assetId) internal view virtual returns (bool) {
+    uint16 minAccruedFeesPercent = _minAccruedFeesPercent[hub][assetId];
+    if (minAccruedFeesPercent == 0) {
+      return false;
+    }
+
+    IHub hubContract = IHub(hub);
```
avniculae: we use targetHub for this in configurators

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,95 @@
...
+    uint256 accruedFees = hubContract.getAssetAccruedFees(assetId);
+    uint256 totalAddedAssets = hubContract.getAddedAssets(assetId);
+
+    if (totalAddedAssets == 0) {
+      return false;
+    }
+    if (PercentageMath.percentDivDown(accruedFees, totalAddedAssets) < minAccruedFeesPercent) {
```
avniculae: we can do using .. for to be consistent

avniculae: is bps enough precision here?

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  /// @dev Internal function to check execution conditions.
+  /// @param hub The address of the Hub.
+  /// @param assetId The identifier of the asset.
+  /// @return True if conditions are met, false otherwise.
+  function _checkUpkeep(address hub, uint256 assetId) internal view virtual returns (bool) {
+    uint16 minAccruedFeesPercent = _minAccruedFeesPercent[hub][assetId];
+    if (minAccruedFeesPercent == 0) {
+      return false;
```
avniculae, line 95: is this dangerous, in case we forget to configure one asset?

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,95 @@
...
+  constructor(address owner) Ownable(owner) {}
+
+  /// @inheritdoc IFeeSharesMinter
+  function setConfig(
+    address hub,
+    uint256 assetId,
+    uint16 minAccruedFeesPercent
+  ) external onlyOwner {
+    require(
+      minAccruedFeesPercent > 0 && minAccruedFeesPercent <= PercentageMath.PERCENTAGE_FACTOR,
```
avniculae: if we want to disable minting for one asset, we have no way bc we require percentage to be >0

src/utils/IFeeSharesMinter.sol
```diff
@@ -0,0 +1,46 @@
...
+  /// @param hub The address of the Hub.
+  /// @param assetId The identifier of the asset.
+  /// @param minAccruedFeesPercent The new minimum ratio of accrued fees to total added assets, in BPS.
+  event ConfigUpdated(address indexed hub, uint256 indexed assetId, uint16 minAccruedFeesPercent);
+
+  /// @notice Thrown upon minting when the required conditions are not met.
+  error ConditionsNotMet();
+
+  /// @notice Thrown when `setConfig` is called with an invalid value.
+  error InvalidConfig();
```
avniculae: I think we should include the percentage value that was passed here, wdyt?

src/utils/IFeeSharesMinter.sol
```diff
@@ -0,0 +1,46 @@
...
+
+  /// @notice Sets the minimum accrued fees percent for a specific asset.
+  /// @param hub The address of the Hub.
+  /// @param assetId The identifier of the asset.
+  /// @param minAccruedFeesPercent Minimum ratio of accrued fees to total added assets, in BPS. Must be greater than zero.
```
avniculae: let's also say it must be at most PERCENTAGE_FACTOR

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,95 @@
...
+      InvalidConfig()
+    );
+    _minAccruedFeesPercent[hub][assetId] = minAccruedFeesPercent;
```
avniculae, line 34: should we also check assetId was registered on the hub? we don't check this anywhere in this contract

tests/contracts/utils/FeeSharesMinter.t.sol
```diff
@@ -0,0 +1,287 @@
...
+      addAmount: addAmount,
+      drawUser: bob,
+      drawSpoke: address(spoke1),
+      drawAmount: drawAmount,
+      skipTime: skipTime
+    });
+
+    bytes memory checkData = abi.encode(address(hub1), daiAssetId);
+    (bool upkeepNeeded, bytes memory performData) = minter.checkUpkeep(checkData);
+    assertEq(performData, checkData, 'performData should equal checkData');
```
avniculae: this can be checked in a different test also, they always must be equal

tests/contracts/utils/FeeSharesMinter.t.sol
```diff
@@ -0,0 +1,287 @@
...
+
+    if (upkeepNeeded) {
```
avniculae: I don't like these fuzz tests, we don't have visbility of what's being tested concretely. I think we should split these scenarios into 2 tests, and make sure each of them check on of the 2 braches

tests/contracts/utils/FeeSharesMinter.t.sol
```diff
@@ -0,0 +1,287 @@
...
+
+    bytes memory checkData = abi.encode(address(hub1), daiAssetId);
+    (bool upkeepNeeded, ) = minter.checkUpkeep(checkData);
+    assertFalse(upkeepNeeded, 'checkUpkeep should return false for unconfigured pair');
+  }
+
+  function test_setConfig_revertsWith_InvalidConfig_zero() public {
+    vm.prank(ADMIN);
+    vm.expectRevert(IFeeSharesMinter.InvalidConfig.selector);
+    minter.setConfig(address(hub1), daiAssetId, 0);
```
avniculae: let's move up, next to the other invalid config test

tests/contracts/utils/FeeSharesMinter.t.sol
avniculae: we also don't have any test to expectCall the hub call to mint fees

src/deployments/orchestration/AaveV4DeployOrchestration.sol
```diff
@@ -378,7 +397,7 @@ library AaveV4DeployOrchestration {
     return report;
   }
 
-  /// @dev Setup roles for the hub and spoke configurators.
+  /// @dev Setup roles for the Hub and spoke configurators.
```
avniculae: spoke is lowercase, maybe we can revert?

src/deployments/utils/libraries/InputUtils.sol
```diff
@@ -8,10 +8,12 @@ library InputUtils {
   /// @dev accessManagerAdmin The default admin of the access manager. Only used when grantRoles is true.
   /// @dev proxyAdminOwner The owner of the Hub and Spoke ProxyAdmin contracts. Required at deploy time (constructor arg).
   ///      When grantRoles is `false`, defaults to the deployer; ownership can be transferred post-deployment.
-  /// @dev hubAdmin The admin of the hub. Only used when grantRoles is true.
+  /// @dev hubAdmin The admin of the Hub. Only used when grantRoles is true.
```
avniculae: what was the convention here?

src/utils/FeeSharesMinter.sol
Kogaroshi: the last value is `_minAccruedFeesPercent`, makes sense not to name it here

src/utils/FeeSharesMinter.sol
Kogaroshi: Yes, this would prevent setting the config for unlisted assets

src/deployments/utils/libraries/InputUtils.sol
Kogaroshi: We seem to be using "Hub" across Natspec overall

src/utils/IFeeSharesMinter.sol
```diff
@@ -0,0 +1,46 @@
...
+  function setConfig(address hub, uint256 assetId, uint16 minAccruedFeesPercent) external;
+
+  /// @notice Chainlink Automation on-chain execution entry point.
+  /// @dev performData must be abi.encoded as (address hub, uint256 assetId).
+  /// @inheritdoc AutomationCompatibleInterface
+  function performUpkeep(bytes calldata performData) external;
```
avniculae: IFeeSharesMinter inherits AutomationCompatibleInterface and redefines them to add dev comments. I'm not sure I like the current approach and I'm thinking the alternatives could be:
- remove inheritance, keep definition here
- keep inheritance, remove these redefinitions and add dev comment in FeeSharesMinter OR at the top of IFeeSharesMinter.

Thoughts? cc @Kogaroshi

Kogaroshi: Yeah, I'd keep inheritence, and have the complete updated Natspec in FeeSharesMinter directly. This feels more consistent with how we did with inheriting and overriding overall no ?

avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/deployments/utils/libraries/InputUtils.sol
avniculae: I reverted to lowercase bc there are still places where we use lowercase. There does not seem to be a consistency atm: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/deployments/orchestration/AaveV4DeployOrchestration.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

tests/contracts/utils/FeeSharesMinter.t.sol
avniculae: N/A anymore [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

tests/contracts/utils/FeeSharesMinter.t.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

tests/contracts/utils/FeeSharesMinter.t.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/utils/FeeSharesMinter.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/utils/IFeeSharesMinter.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/utils/IFeeSharesMinter.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/utils/FeeSharesMinter.sol
avniculae: I am now allowing 0, wdyt? [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/utils/FeeSharesMinter.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/utils/FeeSharesMinter.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/utils/FeeSharesMinter.sol
avniculae: [98c4cc4](https://github.com/aave/aave-v4/pull/1216/commits/98c4cc4ed47fd7b7fe5275cfaa5c0cc2f409af6a)

src/utils/FeeSharesMinter.sol
```diff
@@ -0,0 +1,94 @@
...
+import {Ownable2Step, Ownable} from 'src/dependencies/openzeppelin/Ownable2Step.sol';
+import {PercentageMath} from 'src/libraries/math/PercentageMath.sol';
+import {Rescuable} from 'src/utils/Rescuable.sol';
+import {IFeeSharesMinter, AutomationCompatibleInterface} from 'src/utils/IFeeSharesMinter.sol';
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+
+/// @title FeeSharesMinter
+/// @author Aave Labs
+/// @notice Contract to mint fee shares on the Hub when specific conditions are met.
+contract FeeSharesMinter is IFeeSharesMinter, Ownable2Step, Rescuable {
```
avniculae, line 15: who will own this contract? if EXECUTOR, then what I don't like is that we have no option to disable fee minting without an AIP

wdyt?

DhairyaSethi: lets add OwnableWithGuardian ac and allow disable fee through guardian, or longer way to introduce AccessManager here

src/utils/FeeSharesMinter.sol
```diff
@@ -24,27 +26,26 @@ contract FeeSharesMinter is IFeeSharesMinter, Ownable2Step, Rescuable {
...
     _performUpkeep(hub, assetId);
   }
 
-  /// @inheritdoc IFeeSharesMinter
-  function checkUpkeep(
-    bytes calldata checkData
-  ) external view override returns (bool, bytes memory) {
+  /// @dev `checkData` must be abi.encoded as (address hub, uint256 assetId).
+  /// @inheritdoc AutomationCompatibleInterface
+  function checkUpkeep(bytes memory checkData) external view override returns (bool, bytes memory) {
```
DhairyaSethi: calldata above, mem here, lets be consistent. doesnt really matter bc youre lowering to memory anyway when abi.decoding (you can decode calldata directly for the tbh) and returning (question: why does checkup need to return passed call data?)

src/utils/FeeSharesMinter.sol
DhairyaSethi: why? it does not harm bc we're doing batched versions or seomthing

src/utils/FeeSharesMinter.sol
```diff
@@ -24,27 +26,26 @@ contract FeeSharesMinter is IFeeSharesMinter, Ownable2Step, Rescuable {
...
     _minAccruedFeesPercent[hub][assetId] = minAccruedFeesPercent;
     emit ConfigUpdated(hub, assetId, minAccruedFeesPercent);
   }
 
-  /// @inheritdoc IFeeSharesMinter
+  /// @dev `performData` must be abi.encoded as (address hub, uint256 assetId).
+  /// @inheritdoc AutomationCompatibleInterface
   function performUpkeep(bytes calldata performData) external override {
     (address hub, uint256 assetId) = abi.decode(performData, (address, uint256));
     _performUpkeep(hub, assetId);
```
DhairyaSethi: should we add batched versions? probably not worth the complexity

src/utils/IFeeSharesMinter.sol
```diff
@@ -17,27 +17,14 @@ interface IFeeSharesMinter is AutomationCompatibleInterface {
...
-  /// @notice Chainlink Automation on-chain execution entry point.
-  /// @dev performData must be abi.encoded as (address hub, uint256 assetId).
-  /// @inheritdoc AutomationCompatibleInterface
-  function performUpkeep(bytes calldata performData) external;
-
-  /// @notice Chainlink Automation off-chain simulation check.
-  /// @dev checkData must be abi.encoded as (address hub, uint256 assetId).
-  /// @dev Returns whether upkeep is needed and the performData in bytes when conditions are met.
-  /// @inheritdoc AutomationCompatibleInterface
-  function checkUpkeep(
```
DhairyaSethi: if you killed these from the interface i think its fine to inherit AutomatoinCompatibleInterface on IFeeSharesMinter as well

PR #1217, opened 2026-02-17
rft : add create1 helpers in tests

Kogaroshi:
Use Create1 instead of Create2 in tests when no salt is used for deploy.


tests/Create1Utils.sol
```diff
@@ -0,0 +1,23 @@
...
+import {Vm} from 'forge-std/Vm.sol';
+
+library Create1Utils {
+  error Create1DeploymentFailed();
+
+  Vm internal constant vm = Vm(address(uint160(uint256(keccak256('hevm cheat code')))));
+
+  function create1Deploy(bytes memory bytecode) internal returns (address addr) {
+    /// @solidity memory-safe-assembly
+    assembly {
```
DhairyaSethi:
```suggestion
    assembly ('memory-safe') {
```

tests/DeployUtils.sol
DhairyaSethi, line 1: tysm
still not a fan of the silent continue if deployed already exists i would revert there

Kogaroshi: true, will change that too.

tests/Create1Utils.sol
Kogaroshi: 842ecc475ccc91f56118cc7d35b0842093922fb9

tests/DeployUtils.sol
Kogaroshi: 842ecc475ccc91f56118cc7d35b0842093922fb9

Kogaroshi:
Discarded after change with https://github.com/aave/aave-v4/pull/1218 & https://github.com/aave/aave-v4/pull/1287

PR #1108, opened 2026-01-09
fix: consistent naming on constants

yan-man:
maintain consistent naming conventions on constants and immutables with capitalized var names

closes #1048

PR #1221, opened 2026-02-18
feat: Use latestAnswer instead of latestRoundData for reserve price on AaveOracle

CheyenneAtapour:
This PR replaces `latestRoundData()` with `latestAnswer()` call to fetch price of a reserve on `AaveOracle` contract.

Chainlink team confirmed that both functions will continue to be supported, with no plans to deprecate either function. They are also removing the deprecation notice for `latestAnswer` from their documentation.

Their rationale for exposing `latestRoundData()` is that it enables staleness checks, which `latestAnswer()` does not. Since we are not currently performing staleness checks, we can switch to `latestAnswer()` for simplicity, gas efficiency, and alignment with the Aave V3 `AaveOracle` contract.

Going forward, we can consider adding staleness checks or other safeguards to ensure the Chainlink price feed data is fresh, accurate, and up to date.

Additionally, the `UnitPriceFeed` is removed given that [OneUSDFixedAdapter](https://github.com/bgd-labs/aave-price-feeds/blob/main/src/contracts/misc-adapters/OneUSDFixedAdapter.sol) can be used.

### DIFF
--- a/tests/unit/Spoke/Spoke.MultipleHub.Base.t.sol
+++ b/tests/unit/Spoke/Spoke.MultipleHub.Base.t.sol
@@ -44,10 +44,7 @@ contract SpokeMultipleHubBase is SpokeBase {
 
     // New hub and spoke
     newHub = DeployUtils.deployHub(address(accessManager), hex'02');
-    (newSpoke, newOracle) = _deploySpokeWithOracle(
-      ADMIN,
-      address(accessManager)
-    );
+    (newSpoke, newOracle) = _deploySpokeWithOracle(ADMIN, address(accessManager));
     newIrStrategy = new AssetInterestRateStrategy(address(newHub));
 
     assetA = new TestnetERC20('Asset A', 'A', 18);


src/misc/UnitPriceFeed.sol
```diff
@@ -67,7 +69,7 @@ contract UnitPriceFeed is AggregatorV3Interface {
       uint80 answeredInRound
     )
   {
-    roundId = uint80(block.timestamp);
+    roundId = block.timestamp.toUint80();
     answer = UNITS;
```
DhairyaSethi: answer = latestAnswer();

src/misc/UnitPriceFeed.sol
```diff
@@ -2,14 +2,16 @@
...
 import {AggregatorV3Interface} from 'src/dependencies/chainlink/AggregatorV3Interface.sol';
+import {AggregatorInterface} from 'src/dependencies/chainlink/AggregatorInterface.sol';
 import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
 
 /// @title UnitPriceFeed contract
 /// @author Aave Labs
 /// @notice Price feed that returns the unit price (1), with decimals precision.
 /// @dev This price feed can be set for reserves that use the base currency as collateral.
-contract UnitPriceFeed is AggregatorV3Interface {
+contract UnitPriceFeed is AggregatorV2V3Interface {
```
miguelmtzinf: I think it's fine just using V2

src/spoke/AaveOracle.sol
miguelmtzinf, line 75: we better say that this funciton reverts if price is zero. Same for `getReservesPrices`.
Also, `setReserveSource` docs needs update.

src/misc/UnitPriceFeed.sol
yan-man: i thought the point of this unit price feed was hard coding it to UNITS

DhairyaSethi: yes that's what latestAnswer returns

DhairyaSethi: suggestion is to avoid logic duplication, can be done either way

yan-man: hm then we need to make `latestAnswer` public or create a `_latestAnswer` internal helper to return UNITS, also there's already some code duplication in `getAnswer` too

DhairyaSethi: yes make it public

leekzor:
Security Issue: This PR changes from `latestRoundData()` to `latestAnswer()` for fetching price data. This removes the ability to verify price freshness, which is critical for a lending protocol that uses these prices for collateral valuation, borrowing power calculations, and liquidation decisions.

`latestRoundData()` returns timestamp metadata (`updatedAt`, `startedAt`, `roundId`) that enables staleness checks. `latestAnswer()` only returns the price value with no timestamp information. If a Chainlink price feed stops updating (due to node issues, market halt, or other disruptions), `latestAnswer()` will continue returning the last known price indefinitely.

src/misc/UnitPriceFeed.sol
CheyenneAtapour: I feel it's kinda a waste to do a function call just to return UNITS, we can save that and just return UNITS (there's not really logic in this case), but sure I can do this

src/spoke/AaveOracle.sol
CheyenneAtapour: This does not revert if the source is address(0), right?

src/misc/UnitPriceFeed.sol
CheyenneAtapour: Done

src/misc/UnitPriceFeed.sol
CheyenneAtapour: Updated to use `AggregatorInterface`

src/spoke/AaveOracle.sol
CheyenneAtapour: Done with the rest

src/misc/UnitPriceFeed.sol
```diff
@@ -2,80 +2,52 @@
...
-import {AggregatorV3Interface} from 'src/dependencies/chainlink/AggregatorV3Interface.sol';
+import {AggregatorInterface} from 'src/dependencies/chainlink/AggregatorInterface.sol';
 import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
 
 /// @title UnitPriceFeed contract
 /// @author Aave Labs
 /// @notice Price feed that returns the unit price (1), with decimals precision.
 /// @dev This price feed can be set for reserves that use the base currency as collateral.
-contract UnitPriceFeed is AggregatorV3Interface {
+contract UnitPriceFeed is AggregatorInterface {
```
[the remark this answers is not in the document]
DhairyaSethi: @CheyenneAtapour this looks to be correct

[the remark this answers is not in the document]
CheyenneAtapour: @miguelmtzinf let's go back to V2V3, RE: https://github.com/aave/aave-v4/pull/1221#discussion_r2823660190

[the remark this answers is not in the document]
CheyenneAtapour: Reverted back to v2v3

src/spoke/interfaces/IAaveOracle.sol
```diff
@@ -49,12 +49,13 @@ interface IAaveOracle is IPriceOracle {
...
   /// @notice Sets the price feed source of a reserve.
   /// @dev Must be called by the spoke.
-  /// @dev The source must implement the AggregatorV3Interface.
+  /// @dev The source must implement the AggregatorV2V3Interface.
   /// @param reserveId The identifier of the reserve.
   /// @param source The price feed source of the reserve.
   function setReserveSource(uint256 reserveId, address source) external;
 
   /// @notice Returns the prices of multiple reserves.
+  /// @dev Reverts if the price of one of the reserves is 0.
```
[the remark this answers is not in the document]
miguelmtzinf: @CheyenneAtapour let's say `greater than 0` so we are more precise.

src/misc/UnitPriceFeed.sol
DhairyaSethi: it's no additional cost, since it'll be an internal jump and optimiser will likely inline (we're not using external, which would create an additional staticall)

src/spoke/interfaces/IAaveOracle.sol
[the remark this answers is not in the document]
CheyenneAtapour: It seems weird to say `not greater than 0`, but I changed it to that

src/spoke/interfaces/IPriceFeed.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+/// @title IPriceFeed
+/// @author Aave Labs
+/// @notice Defines the minimal functions needed to work with the AaveOracle contract.
+interface IPriceFeed {
+  /// @notice Returns the number of decimals used to represent the price.
```
miguelmtzinf, line 9:
```suggestion
  /// @notice Returns the description of the feed.
  /// @return string description
    function description() external view returns (string memory);

  /// @notice Returns the number of decimals used to represent the price.
```
let's add

CheyenneAtapour: Done

src/misc/UnitPriceFeed.sol
```diff
@@ -2,80 +2,35 @@
...
 
-  /// @inheritdoc AggregatorV3Interface
+  /// @inheritdoc IPriceFeed
+  uint8 public immutable decimals;
+
+  /// @inheritdoc IPriceFeed
   string public description;
 
-  uint8 private immutable DECIMALS;
   int256 private immutable UNITS;
```
DhairyaSethi: solidity style guide order of variables (reorder needed  based on visibility and immutable)

src/spoke/interfaces/IPriceFeed.sol
```diff
@@ -0,0 +1,20 @@
...
+  /// @return string description
+  function description() external view returns (string memory);
+
+  /// @notice Returns the number of decimals used to represent the price.
+  /// @return The number of decimals.
+  function decimals() external view returns (uint8);
+
+  /// @notice Returns the latest price answer.
+  /// @return The latest price.
+  function latestAnswer() external view returns (int256);
```
DhairyaSethi:
```suggestion
  /// @notice Returns the description of the feed.
  function description() external view returns (string memory);

  /// @notice Returns the number of decimals used to represent the price.
  function decimals() external view returns (uint8);

  /// @notice Returns the latest price answer.
  function latestAnswer() external view returns (int256);
```

DhairyaSethi: also reorder based on above

tests/Base.t.sol
```diff
@@ -2905,7 +2903,7 @@ abstract contract Base is Test {
 
   function _deployMockPriceFeed(ISpoke spoke, uint256 price) internal returns (address) {
     AaveOracle oracle = AaveOracle(spoke.ORACLE());
-    return address(new MockPriceFeed(oracle.DECIMALS(), oracle.DESCRIPTION(), price));
```
DhairyaSethi, line 2908: let's retain desc on the mock

tests/Base.t.sol
```diff
@@ -2886,9 +2886,7 @@ abstract contract Base is Test {
   function _mockReservePrice(ISpoke spoke, uint256 reserveId, uint256 price) internal {
     require(price > 0, 'mockReservePrice: price must be positive');
     AaveOracle oracle = AaveOracle(spoke.ORACLE());
-    address mockPriceFeed = address(
-      new MockPriceFeed(oracle.DECIMALS(), oracle.DESCRIPTION(), price)
-    );
+    address mockPriceFeed = address(new MockPriceFeed(oracle.DECIMALS(), int256(price)));
```
DhairyaSethi: oracle should just expose lowercase decimals (like price feed) and let's just rely on that on our code

tests/mocks/MockPriceFeed.sol
```diff
@@ -2,49 +2,24 @@
...
-  function version() external pure override returns (uint256) {
-    return 1;
-  }
-
-  function getRoundData(
-    uint80
-  ) external pure override returns (uint80, int256, uint256, uint256, uint80) {
-    revert OperationNotSupported();
+    _price = price_;
+    description = 'Mock Price Feed';
```
DhairyaSethi: let this be custom pls

tests/mocks/MockPriceFeed.sol
```diff
@@ -2,49 +2,24 @@
...
 
-contract MockPriceFeed is AggregatorV3Interface {
-  uint8 public immutable override decimals;
-  string public override description;
+contract MockPriceFeed is IPriceFeed {
+  uint8 public immutable decimals;
+
+  string public description;
 
   int256 private immutable _price;
```
DhairyaSethi: price setter missing on the mock

src/misc/UnitPriceFeed.sol
CheyenneAtapour: What's the rule here? Should I flip description and decimals?

tests/mocks/MockPriceFeed.sol
CheyenneAtapour: You don't want it to be immutable?

CheyenneAtapour: Added

tests/mocks/MockPriceFeed.sol
CheyenneAtapour: Done

tests/Base.t.sol

tests/Base.t.sol
CheyenneAtapour: Done

src/spoke/interfaces/IPriceFeed.sol
```diff
@@ -6,12 +6,15 @@ pragma solidity ^0.8.0;
...
 /// @notice Defines the minimal functions needed to work with the AaveOracle contract.
 interface IPriceFeed {
-  /// @notice Returns the description of the feed.
-  function description() external view returns (string memory);
-
   /// @notice Returns the number of decimals used to represent the price.
   function decimals() external view returns (uint8);
 
+  /// @notice Returns the number of units used to represent the price.
+  function units() external view returns (int256);
```
DhairyaSethi: this doesn't belong here, it should not be in the interface for price feed. what we discussed what units can be optionally made public only for the unit price feed

src/misc/UnitPriceFeed.sol
```diff
@@ -2,80 +2,35 @@
...
-contract UnitPriceFeed is AggregatorV3Interface {
+contract UnitPriceFeed is IPriceFeed {
   using SafeCast for uint256;
 
-  /// @inheritdoc AggregatorV3Interface
-  string public description;
+  /// @inheritdoc IPriceFeed
+  uint8 public immutable decimals;
+
+  int256 public immutable units;
```
yan-man: worth adding some natspec here given it's not in any interface?

src/misc/UnitPriceFeed.sol
```diff
@@ -2,80 +2,35 @@
 // Copyright (c) 2025 Aave Labs
 pragma solidity 0.8.28;
 
-import {AggregatorV3Interface} from 'src/dependencies/chainlink/AggregatorV3Interface.sol';
+import {IPriceFeed} from 'src/spoke/interfaces/IPriceFeed.sol';
 import {SafeCast} from 'src/dependencies/openzeppelin/SafeCast.sol';
```
yan-man: nit: we'd been importing the external deps at the top of the file in other contracts

src/spoke/AaveOracle.sol
```diff
@@ -12,44 +12,44 @@ import {IAaveOracle, IPriceOracle} from 'src/spoke/interfaces/IAaveOracle.sol';
...
+    spoke = spoke_;
+    emit SetSpoke(spoke_);
   }
 
   /// @inheritdoc IAaveOracle
   function setReserveSource(uint256 reserveId, address source) external {
-    require(msg.sender == SPOKE, OnlySpoke());
-    AggregatorV3Interface targetSource = AggregatorV3Interface(source);
-    require(targetSource.decimals() == DECIMALS, InvalidSourceDecimals(reserveId));
+    require(msg.sender == spoke, OnlySpoke());
```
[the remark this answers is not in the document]
yan-man: worth adding some validation on address(0) part at least?

src/spoke/interfaces/IPriceFeed.sol
```diff
@@ -0,0 +1,17 @@
...
+/// @notice Defines the minimal functions needed to work with the AaveOracle contract.
+interface IPriceFeed {
+  /// @notice Returns the number of decimals used to represent the price.
+  function decimals() external view returns (uint8);
+
+  /// @notice Returns the description of the feed.
+  function description() external view returns (string memory);
+
+  /// @notice Returns the latest price answer.
+  function latestAnswer() external view returns (int256);
```
yan-man, line 16: im in favor of adding @return natspec to all these, for consistency given it's an external method, and we've maintained that convention in other places

src/misc/UnitPriceFeed.sol
yan-man: immutables at the top, even the private one?

src/spoke/interfaces/IPriceFeed.sol
CheyenneAtapour: haha yeah I had it that way. @DhairyaSethi thoughts?

src/spoke/AaveOracle.sol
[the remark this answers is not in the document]
CheyenneAtapour: That happens later on with the `_getSourcePrice()` call later in this function

src/misc/UnitPriceFeed.sol
CheyenneAtapour: Discussed during sync, now it's all public, immutables at the top.

Discussion for later, but I'm in favor of separating by visibility. Then within each visibility, constant and then immutable at the top

src/spoke/interfaces/IPriceFeed.sol
CheyenneAtapour: Done

src/spoke/interfaces/IPriceFeed.sol
CheyenneAtapour: Done

src/misc/UnitPriceFeed.sol
CheyenneAtapour: Done

src/misc/UnitPriceFeed.sol
CheyenneAtapour: Can't believe I missed that, thank you for catching! Done

src/spoke/interfaces/IPriceFeed.sol
DhairyaSethi: i think we settled this. we don't need to mention if it's super obvious or notice natspec itself starts with Return 


it's consistent w/o bc we have that as a convension, for ex see typehashes and domain separator natspec in isignaturegateway

src/misc/UnitPriceFeed.sol
DhairyaSethi: can someone give me a quick tldr or crash course on import order pls

src/misc/UnitPriceFeed.sol
```diff
@@ -2,80 +2,36 @@
...
+contract UnitPriceFeed is IPriceFeed {
   using SafeCast for uint256;
 
-  /// @inheritdoc AggregatorV3Interface
-  string public description;
+  /// @inheritdoc IPriceFeed
+  uint8 public immutable decimals;
+
+  /// @notice The number of units used to represent the price.
+  int256 public immutable units;
```
DhairyaSethi: this is not really "the number of units" (i think that would be 8?) but rather the unit price (ie 1)

DhairyaSethi: should probably be called price? idk

src/spoke/AaveOracle.sol
```diff
@@ -12,44 +12,44 @@ import {IAaveOracle, IPriceOracle} from 'src/spoke/interfaces/IAaveOracle.sol';
...
   /// @inheritdoc IPriceOracle
-  address public SPOKE;
+  address public spoke;
 
   /// @dev The address of the deployer.
-  address private immutable DEPLOYER;
+  address private immutable _deployer;
 
-  mapping(uint256 reserveId => AggregatorV3Interface) internal _sources;
+  mapping(uint256 reserveId => IPriceFeed) internal _sources;
```
miguelmtzinf, line 24:
```suggestion
  /// @dev Map of reserve identifiers to their price feed.
  mapping(uint256 reserveId => IPriceFeed) internal _sources;
```

src/spoke/interfaces/IAaveOracle.sol
```diff
@@ -68,5 +69,5 @@ interface IAaveOracle is IPriceOracle {
 
   /// @notice Returns the description of the oracle.
   /// @return The description of the oracle.
-  function DESCRIPTION() external view returns (string memory);
+  function description() external view returns (string memory);
 }
```
miguelmtzinf, line 69: missing spoke() getter function

src/spoke/interfaces/IPriceFeed.sol
yan-man: i thought we agreed on external fns, always having full natspec. And internal we are flexible. imo better to be explicit and verbose on external ones.

what do you guys think @avniculae @miguelmtzinf @Kogaroshi

src/misc/UnitPriceFeed.sol
yan-man: i've been doing:
- foundry/internal first
- external (OZ)
- internal src deps

dont know if that's correct ha

src/spoke/interfaces/IAaveOracle.sol
CheyenneAtapour: This comes already from the `IPriceOracle` which `IAaveOracle` inherits

src/misc/UnitPriceFeed.sol
CheyenneAtapour: Order by most external stuff first, coming towards src. Then within each import section from same source, longest order path comes first

src/spoke/AaveOracle.sol
```diff
@@ -12,44 +12,45 @@ import {IAaveOracle, IPriceOracle} from 'src/spoke/interfaces/IAaveOracle.sol';
...
-  string public DESCRIPTION;
+  string public description;
 
   /// @inheritdoc IPriceOracle
-  address public SPOKE;
+  address public spoke;
 
   /// @dev The address of the deployer.
-  address private immutable DEPLOYER;
+  address private immutable _deployer;
```
yan-man: if it's immutable, should still be capitalized right? 
_DEPLOYER

src/spoke/AaveOracle.sol
```diff
@@ -12,44 +12,45 @@ import {IAaveOracle, IPriceOracle} from 'src/spoke/interfaces/IAaveOracle.sol';
 /// @dev Oracles are spoke-specific, due to the usage of reserve id as index of the `_sources` mapping.
 contract AaveOracle is IAaveOracle {
   /// @inheritdoc IPriceOracle
-  uint8 public immutable DECIMALS;
+  uint8 public immutable decimals;
```
yan-man: should remain DECIMALS if immutable

src/spoke/AaveOracle.sol
miguelmtzinf: agree

src/spoke/AaveOracle.sol
miguelmtzinf: agree

src/spoke/AaveOracle.sol
DhairyaSethi: w/o underscore, that's what we've done everywhere else despite my protest

src/spoke/AaveOracle.sol
DhairyaSethi: no, then we need another decimals() as well

DhairyaSethi: the standard we have is lowercase for this. i'm not against upper case bc of immutable but that results in adding a redundant getter

CheyenneAtapour: Im fine as well with immutable to be capitalized, the redundant getter is nbd imo. Better to just keep the convention anyway

src/spoke/AaveOracle.sol
CheyenneAtapour: we can do underscore

DhairyaSethi: omg we had a whole discussion about this. we had before, i always wanted to keep it. but then it was slowly nuked away and i had to not use underscore in tokenisation. see that and the gateways

src/spoke/AaveOracle.sol
DhairyaSethi: yea im ok w redundant getter also but pls have it. the convention is to expose lowercase decimals in the interface only

src/spoke/AaveOracle.sol
yan-man: ah yes we nuked underscore here: https://github.com/aave/aave-v4/pull/1108/changes

but I prefer it being underscore too, i think @miguelmtzinf didn't prefer it? We should just at least make sure it's consistent everywhere

src/spoke/AaveOracle.sol
CheyenneAtapour: Added

src/spoke/AaveOracle.sol
CheyenneAtapour: I think I did not like _ originally with full cap variable names, but it makes sense with the other convention for internal / private variables, so I'm coming around

PR #1224, opened 2026-02-21
docs: Add TokenizationSpoke tech doc

CanonicalJP:
Adds `docs/features/TokenizationSpoke.md`, a technical reference for the TokenizationSpoke contract.

### DIFF
--- a/docs/features/TokenizationSpoke.md
+++ b/docs/features/TokenizationSpoke.md
@@ -21,7 +21,7 @@ Unlike `Spoke.supply`, which restricts `onBehalfOf` to approved Position Manager
 **Deposit flow**
 
 1. The caller approves the underlying ERC-20 asset to the TokenizationSpoke.
-2. `deposit` transfers underlying directly from the caller to the Hub via `safeTransferFrom`. The TokenizationSpoke calls `Hub.add` to account for the deposited amount against its position.
+2. `deposit` transfers the input amount of underlying assets directly from the caller to the Hub via `safeTransferFrom`. The TokenizationSpoke calls `Hub.add` to account for the deposited amount against its position.
 3. Shares are minted to `receiver`; ERC-4626 events are emitted.
 
 **Mint flow**


docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
+# TokenizationSpoke
+
+## Summary
+
+The TokenizationSpoke is a minimal ERC-4626–compliant contract that registers as a Spoke on an Aave V4 Hub, wrapping supply-only Hub positions into transferable ERC-20 shares. Its primary purpose is DeFi composability: the standard vault interface allows external protocols to interact with Aave V4 liquidity without custom adapters. The TokenizationSpoke does not implement borrowing or the full `ISpoke` and `ISpokeBase` interface. It is a pure supply tokenization primitive.
+
+## Relationship to the Hub/Spoke Architecture
+
+Standard Spokes in Aave V4 manage both supply and borrow flows, enforce collateralization constraints, apply risk premium accounting, and support the full `ISpoke` and `ISpokeBase` interface including PositionManager delegation. The TokenizationSpoke operates at a narrower scope. It connects exclusively to the supply side of the Hub, calling `Hub.add` on deposit and `Hub.remove` on withdrawal, and exposes no debt surface.
```
yan-man: not really applying risk premium accounting, as hub also enforces premium debt. I think more like "manage asset level risk configuration" or smt like that 

let's just supply instead of deposit (even if slightly repetitive)

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+A TokenizationSpoke instance wraps exactly one underlying ERC-20 asset. Where a standard Spoke manages multiple Reserves, each TokenizationSpoke deployment is scoped to a single token. The Hub registers the TokenizationSpoke as a distinct entity from any co-existing standard Spoke for the same underlying asset. This means Hub exposure to a given asset can be partitioned across standard Spokes and one or more TokenizationSpoke instances, each governed by independently configured add caps.
```
yan-man: why `Reserves` instead of `reserves`

are "Standard Spokes" how we're officially referring to the vanilla spokes, should it be capitalized then?

yan-man: do we actually want potentially one or more tokenization spokes across a single asset? what's the use case there

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+Because the TokenizationSpoke sits on the Hub's supply side without drawing debt, it does not participate in the risk premium system. Positions held through the TokenizationSpoke cannot be used as collateral and do not contribute to a user's health factor.
+
+## ERC-4626 Interface
+
+The TokenizationSpoke implements the ERC-4626 standard. Entry points include `deposit`, `mint`, `withdraw`, and `redeem`, alongside the standard view surface: `totalAssets`, `convertToShares`, `convertToAssets`, `maxDeposit`, `maxMint`, `maxWithdraw`, and `maxRedeem`.
```
yan-man:
```suggestion
The TokenizationSpoke implements the ERC-4626 standard. User entry points include `deposit`, `mint`, `withdraw`, and `redeem` alongside the standard view functions: `totalAssets`, `convertToShares`, `convertToAssets`, `maxDeposit`, `maxMint`, `maxWithdraw`, and `maxRedeem`.
```

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+Unlike `Spoke.supply`, which restricts `onBehalfOf` to approved PositionManagers, the ERC-4626 interface permits callers to specify any `receiver` on deposit and any `owner` on withdrawal, following standard allowance semantics. `withSig` and EIP-2612 permit operations are natively supported within the TokenizationSpoke, covering the majority of meta-transaction use cases without requiring external PositionManager approval flows.
```
yan-man: I think "PositionManagers" should either be `PositionManagers` or Position Managers - ie contract reference with code blocks, or just named

yan-man: maybe just use supply terminology on all these actions over deposit?

actually this is a bit tricky bc hub uses supply and 4626 is deposit. So ig just consistency is fine

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+**Deposit flow**
+
+1. The caller approves the underlying ERC-20 to the TokenizationSpoke.
+2. `deposit` transfers underlying directly from the caller to the Hub via `safeTransferFrom`.
+3. The TokenizationSpoke calls `Hub.add` to account for the deposited amount against its position.
```
yan-man: maybe this should be 2b) or a part of 2, as it's invoked within the deposit flow

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+1. `withdraw` burns the corresponding shares from `owner`.
+2. The TokenizationSpoke calls `Hub.remove`.
+3. Underlying is transferred to `receiver`.
+
+All four entry points round in favor of the vault per ERC-4626 conventions: `deposit` rounds down shares minted, `mint` rounds up assets required, `withdraw` rounds up shares burned, and `redeem` rounds down assets returned. This asymmetry protects the vault from rounding-based value extraction.
+
+## Share Price and Accounting
+
+The TokenizationSpoke carries no fee logic at the vault layer. There are no performance fees, management fees, or protocol spreads applied by the contract itself. The share price (the ratio of `totalAssets` to total share supply) drifts solely as a function of hub-level interest accrual on the underlying asset. As the Hub accrues yield for the TokenizationSpoke's registered position, the balance attributable to that position increases, and `totalAssets` reflects this, causing outstanding shares to appreciate in underlying terms over time.
```
yan-man: instead of `drifts` maybe "grows" 

> As the Hub accrues yield for the TokenizationSpoke's registered position, the balance attributable to that position increases, and `totalAssets` reflects this, causing outstanding shares to appreciate in underlying terms over time.

As the Hub accrues yield for the TokenizationSpoke's position, totalAssets increases accordingly, and outstanding shares appreciate in underlying terms over time.

smt like this?

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+`totalAssets` converts the vault's total share supply to underlying via the Hub's exchange rate (`previewRemoveByShares`), denominated in the underlying asset's smallest unit. Caps (`addCap`, type `uint40`) are stored in whole asset units and scaled by `10^decimals` during enforcement.
```
yan-man: > during enforcement

during validation?

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Cap Management and Deployment
+
+The TokenizationSpoke does not deploy through the standard Spoke factory. Because it registers directly on the Hub and requires governor-authorized `addCap` configuration, each instance must be deployed and registered manually. This is a deliberate constraint: the Hub's cap accounting for a TokenizationSpoke instance is independent from the add cap of any standard Spoke for the same underlying asset. Risk managers must allocate add cap budget across both entities separately when configuring exposure limits for a given token.
```
yan-man: > does not deploy

does not get deployed

yan-man: > Because it registers directly on the Hub

i think should be passive as the object, ie "Because it gets registered directly on the Hub"

yan-man: > and registered manually

and registered separately?

yan-man: > This is a deliberate constraint: the Hub's cap accounting for a TokenizationSpoke instance is independent from the add cap of any standard Spoke for the same underlying asset. Risk managers must allocate add cap budget across both entities separately when configuring exposure limits for a given token.

what does this mean? it reads like there's 2 addCaps? what are the 2 separate entities

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+The TVL ceiling for a TokenizationSpoke instance is controlled by the `addCap` field in `SpokeConfig` (type `uint40`). In practice it is commonly managed via governance-authorized calls to `HubConfigurator.updateSpokeSupplyCap`, but the source of truth is the Hub’s per-asset and per-spoke configuration (`Hub.updateSpokeConfig`). The cap is enforced by the Hub on every supply-add path (`deposit` and `mint`) when `Hub.add` is invoked: `addCap × 10^decimals ≥ toAddedAssetsUp(spoke.addedShares) + amount` must hold or the transaction reverts. The left-hand side is the decimal-scaled cap; the right-hand side is the spoke's existing position (converted from Hub shares rounding up) plus the incoming assets.
```
yan-man: I have a PR out to rename `updateSpokeSupplyCap` -> updateSpokeAddCap for correctness

yan-man: the spoke's -> the Spoke's?

how should this be? are we doing "the Hub" and "a hub" and "the Spoke", "a spoke"?

yan-man: > `addCap × 10^decimals ≥ toAddedAssetsUp(spoke.addedShares) + amount`

this should be descriptive instead of formula I think. It's a bit hard to read

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Upgradeability
+
+The TokenizationSpoke deploys behind an upgradeable proxy. The current implementation uses `TransparentUpgradeableProxy` per instance (same as standard Spokes).
```
yan-man: > deploys

is deployed

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Safety Controls
+
+The TokenizationSpoke supports the same emergency control states as standard Spokes: `halted` and `active` are both checked for deposits and withdrawals. Enforcement is layered: the TokenizationSpoke's `maxDeposit`/`maxWithdraw` view functions return zero when either flag is in its blocking state, and the Hub's `_validateAdd`/`_validateRemove` independently enforce the same checks, reverting on-chain if violated. Both flags are governance-controlled and enforced per Spoke at Hub validation time.
```
yan-man: first sentence is framed like it is the spoke that supports the flags, but it is a hub-imposed control on the spokes, so it has no choice. so it's more like governed by the same emrgency control states

yan-man: > on-chain

onchain? how do we want to do this generally

docs/features/TokenizationSpoke.md
CanonicalJP: > why `Reserves` instead of `reserves`

It was agreed internally with the team leading the dev of the aave docs and it's now part of our Style Guide.

> Capitalize protocol concepts when they function as nouns (singular or plural) referring to Aave V4 architectural entities:
> - Hub, Hubs
> - Spoke, Spokes
> - Reserve, Reserves
> - Governor
> - PositionManager
> - Reinvestment Module
> - TokenizationSpoke
> - Specific PositionManager names: SupplyRepayPositionManager, AllowancePositionManager, PositionConfigPositionManager
> 
> Use lowercase when the term functions as a variable (e.g. function input paramenter) or as an adjective modifying another noun.

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+1. The caller approves the underlying ERC-20 to the TokenizationSpoke.
+2. `deposit` transfers underlying directly from the caller to the Hub via `safeTransferFrom`.
+3. The TokenizationSpoke calls `Hub.add` to account for the deposited amount against its position.
+4. Shares are minted to `receiver`; ERC-4626 events are emitted.
+
+**Withdrawal flow**
+
+1. `withdraw` burns the corresponding shares from `owner`.
+2. The TokenizationSpoke calls `Hub.remove`.
+3. Underlying is transferred to `receiver`.
```
Kogaroshi: In the Deposit Flow we mention that transfers is executed by the TokenizationSpoke from caller to Hub, so here we might want to say that transfer is done by the Hub directly to the receiver ? 
So we present the difference of flow in terms of transfers.

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+All four entry points round in favor of the vault per ERC-4626 conventions: `deposit` rounds down shares minted, `mint` rounds up assets required, `withdraw` rounds up shares burned, and `redeem` rounds down assets returned. This asymmetry protects the vault from rounding-based value extraction.
```
Kogaroshi, line 45: Also, do we want to add mint/burn flows as well, seems weird to only present the deposit/withdraw, but right after talk about the conversions in the 4 different flows ?

docs/features/TokenizationSpoke.md
CheyenneAtapour: I echo @yan-man 's point, I don't think we'd have the use for 2 tokenization spokes for the same hub, same asset cc @DhairyaSethi

docs/features/TokenizationSpoke.md
CheyenneAtapour: Yeah, I agree, I saw 2 flows, and then the "4" threw me off

docs/features/TokenizationSpoke.md
CheyenneAtapour: yeah @yan-man 's version sounds better. The way it's currently written sounds a bit convoluted. We can def make this paragraph more concise and clearer

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Key Differences from Standard Spokes
+
+**No debt surface**: Standard Spokes expose both `supply` and `borrow` paths. The TokenizationSpoke exposes only the supply side via ERC-4626. There is no `draw`, no `repay`, and no risk premium calculation.
+
+**No PositionManager integration**: Standard Spokes restrict `onBehalfOf` operations to approved PositionManagers. The TokenizationSpoke instead uses ERC-4626 `receiver`/`owner` semantics and natively supports `withSig` flows, reducing friction for integrators who do not need PositionManager delegation.
+
+**Single asset per deployment**: A standard Spoke manages multiple Reserves across multiple assets. Each TokenizationSpoke instance corresponds to exactly one underlying ERC-20, making it a per-asset tokenization contract rather than a market-level entry point.
+
+**Independent cap accounting**: The Hub treats the TokenizationSpoke as a distinct entity from any standard Spoke for the same asset. Add cap must be allocated to the TokenizationSpoke independently; it does not inherit or share cap budget from a co-existing standard Spoke.
```
CheyenneAtapour: This is actually the same as any other spoke: each spoke does have it's own add cap. 

But this is a useful point, so perhaps we can just mention this elsewhere, not in this section as a key difference from other spokes (because it's a similarity to other spokes)

docs/features/TokenizationSpoke.md
CanonicalJP: >how should this be? are we doing "the Hub" and "a hub" and "the Spoke", "a spoke"?

Capitalized in both occurrences, it was a typo

docs/features/TokenizationSpoke.md
CanonicalJP: onchain as single words with no hyphen

docs/features/TokenizationSpoke.md
CanonicalJP: good point, I removed it cause after current updates, this behavior is reinforced throughout the text

docs/features/TokenizationSpoke.md
CanonicalJP: > are "Standard Spokes" how we're officially referring to the vanilla spokes...?

Let's discuss this one and update when settled

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,85 @@
...
+
+## ERC-4626 Interface
+
+The TokenizationSpoke implements the ERC-4626 standard. User entry points include `deposit`, `mint`, `withdraw`, and `redeem` alongside the standard view functions: `totalAssets`, `convertToShares`, `convertToAssets`, `maxDeposit`, `maxMint`, `maxWithdraw`, and `maxRedeem`.
+
+Unlike `Spoke.supply`, which restricts `onBehalfOf` to approved Position Managers, the ERC-4626 interface permits callers to specify any `receiver` on deposit and any `owner` on withdrawal, following standard allowance semantics. `withSig` and EIP-2612 permit operations are natively supported within the TokenizationSpoke, covering the majority of meta-transaction use cases without requiring external PositionManager approval flows.
+
+**Deposit flow**
+
+1. The caller approves the underlying ERC-20 to the TokenizationSpoke.
```
CheyenneAtapour: Small nit: For the others flows, we specify whether it's an asset or shares input, but here we don't. Should add for consistency

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Cap Management and Deployment
+
+The TokenizationSpoke does not get deployed through the standard Spoke factory. Each instance is deployed and registered on the Hub with governor-authorized `SpokeConfig` (including `addCap`), the same pattern as adding any Spoke via `HubConfigurator.addSpoke`.
+
+The TVL ceiling for a TokenizationSpoke instance is controlled by the `addCap` field in `SpokeConfig` (type `uint40`). In practice it is commonly managed via governance-authorized calls to `HubConfigurator.updateSpokeAddCap`, but the source of truth is the Hub’s per-asset and per-spoke configuration (`Hub.updateSpokeConfig`). The cap is enforced by the Hub on every supply-add path (`deposit` and `mint`) when `Hub.add` is invoked. The Hub scales the configured cap to the asset’s decimals, values the Spoke’s existing position in underlying by converting its Hub shares with rounding up, adds the incoming amount, and reverts if that total would exceed the cap.
```
CheyenneAtapour: The "with rounding up" is awkward phrasing, maybe better like this
```suggestion
The TVL ceiling for a TokenizationSpoke instance is controlled by the `addCap` field in `SpokeConfig` (type `uint40`). In practice it is commonly managed via governance-authorized calls to `HubConfigurator.updateSpokeAddCap`, but the source of truth is the Hub’s per-asset and per-spoke configuration (`Hub.updateSpokeConfig`). The cap is enforced by the Hub on every supply-add path (`deposit` and `mint`) when `Hub.add` is invoked. The Hub scales the configured cap to the asset’s decimals, values the Spoke’s existing position in underlying by converting its Hub shares (rounding up), adds the incoming amount, and reverts if that total would exceed the cap.
```

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Out of Scope
+
+The following are explicitly excluded from the TokenizationSpoke:
+
+- **Borrowing**: No draw, repay, or collateralization logic. Positions through the TokenizationSpoke are supply-only.
+- **PositionManagers**: External Position Managers cannot be plugged in. `withSig` and permit cover the key meta-transaction use cases natively.
+- **Fees**: No performance or management fees at the vault layer.
+- **Multi-asset**: Each deployment handles exactly one underlying ERC-20.
+- **Factory deployment**: Must be deployed and registered manually with `addCap` governance setup.
```
CheyenneAtapour: I can see how it can be confusing here. Let's just explicitly state they are not deployed with a factory like standard spokes are, as we do above, so it's crystal clear

docs/features/TokenizationSpoke.md
```diff
@@ -0,0 +1,85 @@
...
+## ERC-4626 Interface
+
+The TokenizationSpoke implements the ERC-4626 standard. User entry points include `deposit`, `mint`, `withdraw`, and `redeem` alongside the standard view functions: `totalAssets`, `convertToShares`, `convertToAssets`, `maxDeposit`, `maxMint`, `maxWithdraw`, and `maxRedeem`.
+
+Unlike `Spoke.supply`, which restricts `onBehalfOf` to approved Position Managers, the ERC-4626 interface permits callers to specify any `receiver` on deposit and any `owner` on withdrawal, following standard allowance semantics. `withSig` and EIP-2612 permit operations are natively supported within the TokenizationSpoke, covering the majority of meta-transaction use cases without requiring external PositionManager approval flows.
+
+**Deposit flow**
+
+1. The caller approves the underlying ERC-20 asset to the TokenizationSpoke.
+2. `deposit` transfers underlying directly from the caller to the Hub via `safeTransferFrom`. The TokenizationSpoke calls `Hub.add` to account for the deposited amount against its position.
```
CheyenneAtapour: I'd like to see the amount reference come a bit earlier, so I think like: 
```suggestion
2. `deposit` transfers the specified amount of underlying directly from the caller to the Hub via `safeTransferFrom`. The TokenizationSpoke calls `Hub.add` to account for the deposited amount against its position.
```

docs/features/TokenizationSpoke.md
CheyenneAtapour: This can be discussed on the aggregated PR (docs/tech-docs branch) so we can apply what we choose consistently across all docs

PR #1230, opened 2026-02-24
feat: invariant tests

DhairyaSethi:


invariants/protocol-suite/specs/InvariantsSpec.t.sol
```diff
@@ -0,0 +1,87 @@
...
+    ///   - Properties that should always hold true in the system.
+    ///   - Implemented in the /invariants folder.
+
+    /////////////////////////////////////////////////////////////////////////////////////////////*/
+
+  ///////////////////////////////////////////////////////////////////////////////////////////////
+  //                                         HUB                                               //
+  ///////////////////////////////////////////////////////////////////////////////////////////////
+
+  string constant INV_HUB_A2 = 'INV_HUB_A2: If hub assets = 0 => shares 0';
```
Kogaroshi: There is no more `INV_HUB_A1` no ? we can rename to `INV_HUB_A`

PR #1234, opened 2026-02-27
feat : config Payload for Governance V3

Kogaroshi:
Add a new Config Engine & Payload to be compatible with Governance V3 via aave-helpers.

To be updated to have a more granular list or Roles & config functions based on #1047

### DIFF
--- a/tests/config-engine/AaveV4Payload.t.sol
+++ b/tests/config-engine/AaveV4Payload.t.sol
@@ -1036,7 +1036,7 @@ contract AaveV4PayloadTest is BaseConfigEngineTest {
 
     vm.startPrank(ADMIN);
     accessManager.grantRole(Roles.SPOKE_CONFIGURATOR_ROLE, address(freshPayload), 0);
-    accessManager.grantRole(Roles.DEFAULT_ADMIN_ROLE, address(freshPayload), 0);
+    accessManager.grantRole(Roles.ACCESS_MANAGER_ADMIN_ROLE, address(freshPayload), 0);
     vm.stopPrank();
 
     IAaveV4ConfigEngine.AssetHalt[] memory halts = new IAaveV4ConfigEngine.AssetHalt[](1);
@@ -1063,7 +1063,7 @@ contract AaveV4PayloadTest is BaseConfigEngineTest {
 
     vm.startPrank(ADMIN);
     accessManager.grantRole(Roles.HUB_CONFIGURATOR_ROLE, address(freshPayload), 0);
-    accessManager.grantRole(Roles.DEFAULT_ADMIN_ROLE, address(freshPayload), 0);
+    accessManager.grantRole(Roles.ACCESS_MANAGER_ADMIN_ROLE, address(freshPayload), 0);
     vm.stopPrank();
 
     IAaveV4ConfigEngine.ReserveConfigUpdate[]
@@ -1106,7 +1106,7 @@ contract AaveV4PayloadTest is BaseConfigEngineTest {
       abi.encodeWithSelector(
         IAccessManager.AccessManagerUnauthorizedAccount.selector,
         address(freshPayload),
-        Roles.DEFAULT_ADMIN_ROLE
+        Roles.ACCESS_MANAGER_ADMIN_ROLE
       )
     );
     freshPayload.execute();
@@ -1120,7 +1120,7 @@ contract AaveV4PayloadTest is BaseConfigEngineTest {
     vm.startPrank(ADMIN);
     accessManager.grantRole(Roles.HUB_CONFIGURATOR_ROLE, address(freshPayload), 0);
     accessManager.grantRole(Roles.SPOKE_CONFIGURATOR_ROLE, address(freshPayload), 0);
-    accessManager.grantRole(Roles.DEFAULT_ADMIN_ROLE, address(freshPayload), 0);
+    accessManager.grantRole(Roles.ACCESS_MANAGER_ADMIN_ROLE, address(freshPayload), 0);
     vm.stopPrank();
 
     PositionManagerBaseWrapper deadPm = new PositionManagerBaseWrapper(address(0xdead));


src/config-engine/libraries/EngineFlags.sol
DhairyaSethi, line 1: should go inside `src/config-engine/libraries/`?

src/config-engine/libraries/AccessManagerEngine.sol
```diff
@@ -0,0 +1,256 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {IAaveV4ConfigEngine} from 'src/config-engine/interfaces/IAaveV4ConfigEngine.sol';
+import {IAccessManager} from 'src/dependencies/openzeppelin/IAccessManager.sol';
+import {Roles} from 'src/libraries/types/Roles.sol';
```
DhairyaSethi: these roles are not final atm unfortunately, we need to decide a strategy asap bc one selector can only belong to one role

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
DhairyaSethi, line 1: can we also add a readme like https://github.com/aave-dao/aave-v3-origin/blob/main/src/contracts/extensions/v3-config-engine/README.md would be awesome!

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,300 @@
...
+          listings[i].feeReceiver,
+          listings[i].liquidityFee,
+          listings[i].irStrategy,
+          listings[i].irData
+        );
+      } else {
+        listings[i].hubConfigurator.addAssetWithDecimals(
+          listings[i].hub,
+          listings[i].underlying,
+          uint8(listings[i].decimals),
```
DhairyaSethi:
```suggestion
          listings[i].decimals.toUint8(),
```

although this should already be an uint8

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,300 @@
...
+/// @title HubEngine
+/// @author Aave Labs
+/// @notice Library containing hub configurator logic for AaveV4ConfigEngine.
+library HubEngine {
+  /// @notice Lists new assets on hubs via the HubConfigurator.
+  /// @param listings The asset listings to execute.
+  function executeHubAssetListings(IAaveV4ConfigEngine.AssetListing[] calldata listings) external {
+    uint256 length = listings.length;
+    for (uint256 i; i < length; ++i) {
+      if (listings[i].decimals == 0) {
```
DhairyaSethi: maybe we do this through an engine flag? dynamic/static maybe

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,847 @@
...
+  /// @param spoke The address of the Spoke.
+  /// @param assetIds The list of asset identifiers to register the spoke for.
+  /// @param configs The list of Spoke configurations to register.
+  struct SpokeToAssetsAddition {
+    IHubConfigurator hubConfigurator;
+    address hub;
+    address spoke;
+    uint256[] assetIds;
+    IHub.SpokeConfig[] configs;
+  }
```
DhairyaSethi: maybe just have the second multiple one to reduce or just the single one?

src/config-engine/libraries/AccessManagerEngine.sol
yan-man: i think something as granular as possible helps that ie minimize the number of selectors per role. That allows us more flexibility later, if we have stewards that only need a small number of selectors. So we can build complex roles out of a combination of granular roles. 

This will be better than grouping too many selectors together into a single role, and later having to redefine them and needing to update all existing admins.

DhairyaSethi: > So we can build complex roles out of a combination of granular roles.

how? we cannot group roles together, right? or are you just saying the actor needing this will just be assigned all roles or smth

src/config-engine/AaveV4Payload.sol
```diff
@@ -0,0 +1,1134 @@
...
+  }
+
+  /// @notice Hook called before executing any actions. Override to add pre-execution logic.
+  function _preExecute() internal virtual {}
+  /// @notice Hook called after executing all actions. Override to add post-execution logic.
+  function _postExecute() internal virtual {}
+
+  /// @notice Returns the hub asset listings to execute. Override to provide listings.
+  /// @return An array of AssetListing structs (empty by default).
+  function hubAssetListings() internal virtual returns (IAaveV4ConfigEngine.AssetListing[] memory) {
```
yan-man:
```suggestion
  function _hubAssetListings() internal virtual returns (IAaveV4ConfigEngine.AssetListing[] memory) {
```
?

DhairyaSethi: i think it's fine to make the exception here given how it's used


actually tbh this should be a public function instead 

```suggestion
  function hubAssetListings() public virtual returns (IAaveV4ConfigEngine.AssetListing[] memory) {
```

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: There is both, because my thinking was the multiple one would be more efficient for initial market config (right after deploy), while the single one is better for specific new listings.

DhairyaSethi: i think just having the single one reduced moat, the multiple one is not really that much more efficient anyway since it just abstracts the 2 ext delegatecalls where it this be called by a contract only

Kogaroshi: I agree the abstraction is not much. But I'm thinking on the auditing of payloads, it might be easier to read 1 set hub & spoke with the list of configs & assetIds, rather than the same call copy/pasted X times where one of the spoke address or hub address might be incorrect.
But ofc it doesn't reduce the potential mistakes in assetIds or configs there.

src/config-engine/AaveV4Payload.sol
Kogaroshi: Gonna change to public for now, but due to contract codesize, we might need to revert that change.
Because those Payloads will need to be deployed onchain (see https://github.com/aave-dao/aave-proposals-v3), so we might want to limit the external getters there.

Kogaroshi: 17422af916f53f98b223aeea5720948ff3762e0a

src/config-engine/libraries/HubEngine.sol
Kogaroshi: 17422af916f53f98b223aeea5720948ff3762e0a

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 17422af916f53f98b223aeea5720948ff3762e0a

src/config-engine/libraries/EngineFlags.sol
Kogaroshi: 17422af916f53f98b223aeea5720948ff3762e0a

src/config-engine/AaveV4Payload.sol
DhairyaSethi: bruh yeah that IS an issue :( 

yeah perhaps let's keep internal but without the prefix, like you had it before :hiding:

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
DhairyaSethi: can we add the diagram as well 😸

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
DhairyaSethi: yeah probably thats why multiple config at once only makes sense, thats what v3 does mostly

single update ones we can remove

src/config-engine/AaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,346 @@
...
+    HubEngine.executeHubFeeConfigUpdates(updates);
+  }
+
+  /// @inheritdoc IAaveV4ConfigEngine
+  function executeHubInterestRateUpdates(InterestRateUpdate[] calldata updates) external {
+    HubEngine.executeHubInterestRateUpdates(updates);
+  }
+
+  /// @inheritdoc IAaveV4ConfigEngine
+  function executeHubReinvestmentControllerUpdates(
```
DhairyaSethi: should be included in update asset configs, do not need this granularity, user will KEEP_CURRENT on rest

same idea for all below

Kogaroshi: 1db364e2c54691c08fdd7e8787062104984ffbe2

src/config-engine/AaveV4Payload.sol
Kogaroshi: 1db364e2c54691c08fdd7e8787062104984ffbe2 => size is okay in the current version, should allow for most usage in proposals.
If there is a limit, we can always set external contract with getters to return the lists for Payload usage.

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 1db364e2c54691c08fdd7e8787062104984ffbe2

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 1db364e2c54691c08fdd7e8787062104984ffbe2

src/config-engine/libraries/AccessManagerEngine.sol
Kogaroshi: currently decided to remove the Roles from that logic, as it was increase a lot the Engine & Payload codesize. Instead we have just the generic methods, and we can have an helper contract to generate the `IAaveV4ConfigEngine.RoleMembership` & such with preset roleIds based on our defined Role list.

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+import {IHubConfigurator} from 'src/hub/interfaces/IHubConfigurator.sol';
+import {ISpokeConfigurator} from 'src/spoke/interfaces/ISpokeConfigurator.sol';
```
CheyenneAtapour:
```suggestion
import {IHubConfigurator} from 'src/hub/interfaces/IHubConfigurator.sol';
import {ISpokeConfigurator} from 'src/spoke/interfaces/ISpokeConfigurator.sol';
import {IHub} from 'src/hub/interfaces/IHub.sol';
import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
```

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+
+/// @title IAaveV4ConfigEngine
+/// @author Aave Labs
+/// @notice Interface for the Aave V4 Config Engine, defining all structs and engine method signatures.
+/// The engine is stateless and invoked via delegatecall from payload contracts.
+/// All numeric fields in config structs use uint256 so that type(uint256).max can serve as
```
CheyenneAtapour, line 14: can we use this sentinel like that though? Like for example, what if we want to go from a non-inf `addCap` to inf `addCap`? Then we'd pass `type(uint256).max` but that wouldn't actually change the `addCap` as desired. I think similar behavior for some other configs that use this value as a sentinel, no?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @param configs The list of Spoke configurations to register.
+  struct SpokeToAssetsAddition {
+    IHubConfigurator hubConfigurator;
+    address hub;
+    address spoke;
+    uint256[] assetIds;
+    IHub.SpokeConfig[] configs;
+  }
+
+  /// @notice Parameters for updating spoke config (caps, risk premium threshold, status) on a hub.
```
CheyenneAtapour:
```suggestion
  /// @notice Parameters for updating spoke config (caps, risk premium threshold, status) on a Hub.
```

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+import {ISpokeConfigurator} from 'src/spoke/interfaces/ISpokeConfigurator.sol';
+
+/// @title IAaveV4ConfigEngine
+/// @author Aave Labs
+/// @notice Interface for the Aave V4 Config Engine, defining all structs and engine method signatures.
+/// The engine is stateless and invoked via delegatecall from payload contracts.
+/// All numeric fields in config structs use uint256 so that type(uint256).max can serve as
+/// the universal KEEP_CURRENT sentinel. Boolean fields use uint256 (0=false, 1=true, KEEP_CURRENT=skip).
+interface IAaveV4ConfigEngine {
+  /// @notice Parameters for listing a new asset on a hub.
```
CheyenneAtapour:
```suggestion
  /// @notice Parameters for listing a new asset on a Hub.
```
We want to be consistent about a capitalized `Hub`, right? There are a few here that need changing if so

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+    IHubConfigurator hubConfigurator;
+    address hub;
+    uint256 assetId;
+  }
+
+  /// @notice Parameters for resetting asset caps on a hub.
+  /// @param hubConfigurator The HubConfigurator to use for this action.
+  /// @param hub The address of the Hub.
+  /// @param assetId The identifier of the asset.
+  struct AssetCapsReset {
```
CheyenneAtapour, line 135: Looking at this now, without context it's kind of unclear what "reset" means. For example, it could be resetting to some older value. Would it be better to call this struct ZeroAssetCaps, and the function on `HubConfigurator` to be `zeroAssetCaps` or similar?

Or if not, maybe this natspec can be more clear to say that it's setting the asset caps to 0

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @param spokeConfigurator The SpokeConfigurator to use for this action.
+  /// @param spoke The address of the Spoke.
+  struct SpokePause {
+    ISpokeConfigurator spokeConfigurator;
+    address spoke;
+  }
+
+  /// @notice Parameters for freezing all reserves on a spoke.
+  /// @param spokeConfigurator The SpokeConfigurator to use for this action.
+  /// @param spoke The address of the Spoke.
```
CheyenneAtapour: On other structs, we denote the fields with `@dev` instead of `@param`

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @param user The address of the user to renounce the role for.
+  struct PositionManagerRoleRenouncement {
+    address positionManager;
+    address spoke;
+    address user;
+  }
+
+  /// @notice Parameters for granting or revoking a role via AccessManager.
+  /// When granted=true → grantRole(roleId, account, executionDelay).
+  /// When granted=false → revokeRole(roleId, account). executionDelay is ignored.
```
CheyenneAtapour: should these 2 lines be prefixed with `@dev`?

src/config-engine/AaveV4Payload.sol
```diff
@@ -0,0 +1,469 @@
...
+
+  /// @param configEngine The IAaveV4ConfigEngine implementation to delegatecall into.
+  constructor(IAaveV4ConfigEngine configEngine) {
+    require(address(configEngine) != address(0), InvalidConfigEngine());
+    CONFIG_ENGINE = configEngine;
+  }
+
+  /// @notice Main execution entry point called by governance. Runs all configured actions.
+  /// @dev Expected to be called by a governance executor. No on-chain access control is applied;
+  ///   the caller is responsible for authorization. Idempotency is not guaranteed.
```
CheyenneAtapour, line 27: why is no access control applied on-chain exactly?

src/config-engine/libraries/AccessManagerEngine.sol
```diff
@@ -0,0 +1,83 @@
...
+    }
+  }
+
+  /// @notice Updates role configuration (admin, guardian, grant delay, label) via AccessManager.
+  /// @param updates The role updates to execute.
+  function executeRoleUpdates(IAaveV4ConfigEngine.RoleUpdate[] calldata updates) external {
+    uint256 length = updates.length;
+    for (uint256 i; i < length; ++i) {
+      IAccessManager authority = IAccessManager(updates[i].authority);
+      if (updates[i].admin != type(uint64).max) {
```
CheyenneAtapour: Should we use constants for these, like defined in EngineFlags?

src/config-engine/AaveV4Payload.sol
```diff
@@ -0,0 +1,469 @@
...
+  }
+
+  /// @notice Hook called before executing any actions. Override to add pre-execution logic.
+  function _preExecute() internal virtual {}
+  /// @notice Hook called after executing all actions. Override to add post-execution logic.
+  function _postExecute() internal virtual {}
+
+  /// @notice Returns the hub asset listings to execute. Override to provide listings.
+  /// @return An array of AssetListing structs (empty by default).
+  function hubAssetListings() public virtual returns (IAaveV4ConfigEngine.AssetListing[] memory) {
```
CheyenneAtapour: Firstly these functions should be above the internals if we are following solidity style conventions.
Also, do we not want them as `view` because they might have effects?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+
+  /// @notice Updates dynamic reserve configs on spokes.
+  /// @param updates The dynamic reserve config updates to execute.
+  function executeSpokeDynamicReserveConfigUpdates(
+    DynamicReserveConfigUpdate[] calldata updates
+  ) external;
+
+  /// @notice Pauses all reserves on spokes.
+  /// @param pauses The spoke pauses to execute.
+  function executeSpokeAllReservesPauses(SpokePause[] calldata pauses) external;
```
CheyenneAtapour:
```suggestion
  function executeSpokePauseAllReserves(SpokePause[] calldata pauses) external;
```
Function name seems awkward

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+
+  /// @notice Freezes all reserves on spokes.
+  /// @param freezes The spoke freezes to execute.
+  function executeSpokeAllReservesFreezes(SpokeFreeze[] calldata freezes) external;
```
CheyenneAtapour:
```suggestion
  function executeSpokeFreezeAllReserves(SpokeFreeze[] calldata freezes) external;
```
?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  function executeSpokeDynamicReserveConfigUpdates(
+    DynamicReserveConfigUpdate[] calldata updates
+  ) external;
+
+  /// @notice Pauses all reserves on spokes.
+  /// @param pauses The spoke pauses to execute.
+  function executeSpokeAllReservesPauses(SpokePause[] calldata pauses) external;
+
+  /// @notice Freezes all reserves on spokes.
+  /// @param freezes The spoke freezes to execute.
```
CheyenneAtapour: Also kinda awkwardly worded

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @param updates The role updates to execute.
+  function executeRoleUpdates(RoleUpdate[] calldata updates) external;
+
+  /// @notice Updates target function roles via AccessManager.
+  /// @param updates The target function role updates to execute.
+  function executeTargetFunctionRoleUpdates(TargetFunctionRoleUpdate[] calldata updates) external;
+
+  /// @notice Updates target admin delays via AccessManager.
+  /// @param updates The target admin delay updates to execute.
+  function executeTargetAdminDelayUpdates(TargetAdminDelayUpdate[] calldata updates) external;
```
CheyenneAtapour, line 431: I thought we didn't want to use this functionality. But we are just adding for completeness in case we do?

src/config-engine/libraries/AccessManagerEngine.sol
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {IAaveV4ConfigEngine} from 'src/config-engine/interfaces/IAaveV4ConfigEngine.sol';
+import {IAccessManager} from 'src/dependencies/openzeppelin/IAccessManager.sol';
```
CheyenneAtapour:
```suggestion
import {IAccessManager} from 'src/dependencies/openzeppelin/IAccessManager.sol';
import {IAaveV4ConfigEngine} from 'src/config-engine/interfaces/IAaveV4ConfigEngine.sol';
```

src/config-engine/libraries/SpokeEngine.sol
```diff
@@ -0,0 +1,223 @@
...
+        updates[i].reserveId,
+        updates[i].dynamicConfigKey.toUint32(),
+        current
+      );
+    }
+  }
+
+  /// @notice Pauses all reserves on spokes.
+  /// @param pauses The spoke pauses to execute.
+  function executeSpokeAllReservesPauses(
```
CheyenneAtapour:
```suggestion
  function executeSpokePauseAllReserves(
```
It's awkward wording, but I guess we are trying to keep a convention here perhaps, because this is to pause all reserves on mutliple spokes

src/config-engine/AaveV4Payload.sol
```diff
@@ -0,0 +1,469 @@
...
+
+  /// @notice Delegatecalls the config engine with the given calldata.
+  /// @param data The ABI-encoded function call to forward to CONFIG_ENGINE.
+  function _delegateCallEngine(bytes memory data) internal {
+    address(CONFIG_ENGINE).functionDelegateCall(data);
+  }
+
+  /// @notice Hook called before executing any actions. Override to add pre-execution logic.
+  function _preExecute() internal virtual {}
+  /// @notice Hook called after executing all actions. Override to add post-execution logic.
```
CheyenneAtapour:
```suggestion

  /// @notice Hook called after executing all actions. Override to add post-execution logic.
```
whitespace to separate?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: we could modify the sentinel value to `type(uint256).max - 1`, so we can pass infinite value in those case, wdyt ?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: Yes, as we might want to use it later on, kept here for completeness.

src/config-engine/AaveV4Payload.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/libraries/SpokeEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/libraries/AccessManagerEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/AaveV4Payload.sol
Kogaroshi: reordering : 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/libraries/AccessManagerEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 4a095603e966803b07d32d3b4dedb6903bdd3265

tests/config-engine/BaseConfigEngine.t.sol
```diff
@@ -0,0 +1,418 @@
...
+    internal
+    view
+    returns (IAaveV4ConfigEngine.DynamicReserveConfigAddition memory)
+  {
+    return
+      IAaveV4ConfigEngine.DynamicReserveConfigAddition({
+        spokeConfigurator: ISpokeConfigurator(address(mockSpokeConfigurator)),
+        spoke: SPOKE,
+        reserveId: RESERVE_ID,
+        dynamicConfig: ISpoke.DynamicReserveConfig({
```
CheyenneAtapour: Should we perhaps also have a general default DynamicReserveConfig so we can use that here and not copy paste over the values?

Also not helpful yet in this file, but if it's worth it, same thing for a default ReserveConfig

tests/config-engine/HubEngine.t.sol
```diff
@@ -0,0 +1,1002 @@
...
+    update.irData = '';
+    update.reinvestmentController = EngineFlags.KEEP_CURRENT_ADDRESS;
+
+    vm.expectRevert(MockHubConfigurator.UpdateFeeConfigReverted.selector);
+    engine.executeHubAssetConfigUpdates(_toAssetConfigUpdateArray(update));
+  }
+
+  // ---------------------------------------------------------------------------
+  // AssetConfigUpdate tests — interest rate scenarios
+  // ---------------------------------------------------------------------------
```
CheyenneAtapour: Do we want this kind of thing? There was discussion in another PR about removing these

tests/config-engine/HubEngine.t.sol
```diff
@@ -0,0 +1,1002 @@
...
+  }
+
+  function testFuzz_executeHubSpokeConfigUpdates_status(uint256 active, uint256 halted) public {
+    // Bound to valid flag values (ENABLED or DISABLED only, not KEEP_CURRENT)
+    active = bound(active, 0, 1);
+    halted = bound(halted, 0, 1);
+
+    IAaveV4ConfigEngine.SpokeConfigUpdate memory update = _defaultSpokeConfigUpdate();
+    update.active = active;
+    update.halted = halted;
```
CheyenneAtapour: ig if the active or halted flag is the same, it stills calls into the function, but perhaps that's fine

tests/config-engine/HubEngine.t.sol
```diff
@@ -0,0 +1,1002 @@
...
+    mockHubConfigurator.setShouldRevert(IHubConfigurator.resetSpokeCaps.selector, true);
+
+    IAaveV4ConfigEngine.SpokeCapsReset memory reset = IAaveV4ConfigEngine.SpokeCapsReset({
+      hubConfigurator: IHubConfigurator(address(mockHubConfigurator)),
+      hub: HUB,
+      spoke: SPOKE
+    });
+
+    vm.expectRevert(MockHubConfigurator.ResetSpokeCapsReverted.selector);
+    engine.executeHubSpokeCapsResets(_toSpokeCapsResetArray(reset));
```
CheyenneAtapour: I see we are just setting it to revert in the MockHubConfigurator. Is this useful? Or should we just be testing real cases where it would revert?

tests/config-engine/SpokeEngine.t.sol
```diff
@@ -0,0 +1,696 @@
...
+      true
+    );
+
+    IAaveV4ConfigEngine.LiquidationConfigUpdate memory update = _defaultLiquidationConfigUpdate();
+
+    vm.expectRevert(MockSpokeConfigurator.UpdateLiquidationConfigReverted.selector);
+    engine.executeSpokeLiquidationConfigUpdates(_toLiquidationConfigUpdateArray(update));
+  }
+
+  function _setupDynamicReserveConfig(
```
CheyenneAtapour: Perhaps this helper should go at the end

tests/config-engine/SpokeEngine.t.sol
```diff
@@ -0,0 +1,696 @@
...
+        spokeConfigurator: ISpokeConfigurator(address(mockSpokeConfigurator)),
+        spoke: address(mockSpokeReader),
+        reserveId: RESERVE_ID,
+        dynamicConfigKey: DYNAMIC_CONFIG_KEY,
+        collateralFactor: EngineFlags.KEEP_CURRENT,
+        maxLiquidationBonus: EngineFlags.KEEP_CURRENT,
+        liquidationFee: EngineFlags.KEEP_CURRENT
+      });
+
+    // When all fields are KEEP_CURRENT the external write call should be skipped entirely.
```
CheyenneAtapour: I think we can remove this comment, we don't have it elsewhere

tests/mocks/config-engine/MockHubConfigurator.sol
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  function updateSpokeRiskPremiumThreshold(
+    address hub,
+    uint256 assetId,
+    address spoke,
+    uint256 riskPremiumThreshold
+  ) external {
+    if (shouldRevert[msg.sig]) revert UpdateSpokeRiskPremiumThresholdReverted();
+    emit UpdateSpokeRiskPremiumThresholdCalled(hub, assetId, spoke, riskPremiumThreshold);
```
CheyenneAtapour: Are we doing it like this because we are decoupling from the behavior of the actual hubConfigurator?

tests/config-engine/HubEngine.t.sol
Kogaroshi: yeah, I keep removing, missed this one, let me fix

tests/config-engine/BaseConfigEngine.t.sol
Kogaroshi: That could be a good addition
wdyt @miguelmtzinf @DhairyaSethi ?

tests/config-engine/HubEngine.t.sol
Kogaroshi: I tried to keep those tests closer to unit tests, since we will have some more e2e tests with actual payloads in repos that import that engine.
But we can see to replace the mocks with actual Configurators, Hubs & Spokes if we want here as well.

tests/config-engine/SpokeEngine.t.sol
Kogaroshi: actually gonna move it in `BaseConfigEngineTest` with all internal helpers

tests/mocks/config-engine/MockHubConfigurator.sol
Kogaroshi: those are just meant to confirm the Engine passes the expected configs parameters to the Configurator

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @dev irStrategy The address of the interest rate strategy contract.
+  /// @dev irData The interest rate data to apply to the given asset, encoded in bytes.
+  struct AssetListing {
+    IHubConfigurator hubConfigurator;
+    address hub;
+    address underlying;
+    uint256 decimals;
+    address feeReceiver;
+    uint256 liquidityFee;
+    address irStrategy;
```
DhairyaSethi, line 42: shouldnt this be deployed or taken care of automatically?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+    bytes irData;
```
DhairyaSethi: why is this bytes? lets have the struct here pls for type safety even tho we check later, no need for this to be bytes

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @notice Parameters for listing a new asset on a Hub.
+  /// @dev hubConfigurator The HubConfigurator to use for this action.
+  /// @dev hub The address of the Hub.
+  /// @dev underlying The address of the underlying asset.
+  /// @dev decimals Explicit decimals (0 means use addAsset which auto-detects decimals).
+  /// @dev feeReceiver The address of the fee receiver Spoke.
+  /// @dev liquidityFee The liquidity fee of the asset, in BPS.
+  /// @dev irStrategy The address of the interest rate strategy contract.
+  /// @dev irData The interest rate data to apply to the given asset, encoded in bytes.
+  struct AssetListing {
```
DhairyaSethi, line 36: this needs to include tokenized name and symbol

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @dev liquidityFee The new liquidity fee (KEEP_CURRENT to skip).
+  /// @dev feeReceiver The new fee receiver (KEEP_CURRENT_ADDRESS to skip).
+  /// @dev irStrategy The new interest rate strategy (KEEP_CURRENT_ADDRESS to skip strategy update).
+  /// @dev irData The interest rate data. If irStrategy != KEEP_CURRENT_ADDRESS, calls updateInterestRateStrategy.
+  ///   Otherwise if irData.length > 0, calls updateInterestRateData.
+  /// @dev reinvestmentController The new reinvestment controller (KEEP_CURRENT_ADDRESS to skip).
+  struct AssetConfigUpdate {
+    IHubConfigurator hubConfigurator;
+    address hub;
+    uint256 assetId;
```
DhairyaSethi: let's reference by underlying pls, this should be able fetch asset id

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+    uint256 liquidityFee;
+    address feeReceiver;
+    address irStrategy;
+    bytes irData;
```
DhairyaSethi: exact type pls

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @dev hubConfigurator The HubConfigurator to use for this action.
+  /// @dev hub The address of the Hub.
+  /// @dev spoke The address of the Spoke.
+  /// @dev assetIds The list of asset identifiers to register the Spoke for.
+  /// @dev configs The list of Spoke configurations to register.
+  struct SpokeToAssetsAddition {
+    IHubConfigurator hubConfigurator;
+    address hub;
+    address spoke;
+    uint256[] assetIds;
```
DhairyaSethi: ref by underlying

also for spoke reserve listings, we can start to reference by hub and underlying instead of reserve id i think that'd be better and cleaner

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+
+  /// @notice Parameters for updating asset config (fee, interest rate, reinvestment) on a Hub.
+  /// @dev hubConfigurator The HubConfigurator to use for this action.
+  /// @dev hub The address of the Hub.
+  /// @dev assetId The identifier of the asset.
+  /// @dev liquidityFee The new liquidity fee (KEEP_CURRENT to skip).
+  /// @dev feeReceiver The new fee receiver (KEEP_CURRENT_ADDRESS to skip).
+  /// @dev irStrategy The new interest rate strategy (KEEP_CURRENT_ADDRESS to skip strategy update).
+  /// @dev irData The interest rate data. If irStrategy != KEEP_CURRENT_ADDRESS, calls updateInterestRateStrategy.
+  ///   Otherwise if irData.length > 0, calls updateInterestRateData.
```
DhairyaSethi: each individual flag should have keep current variant on this struct

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+    uint256 riskPremiumThreshold;
+    uint256 active;
+    uint256 halted;
+  }
+
+  /// @notice Parameters for halting an asset on a Hub.
+  /// @dev hubConfigurator The HubConfigurator to use for this action.
+  /// @dev hub The address of the Hub.
+  /// @dev assetId The identifier of the asset.
+  struct AssetHalt {
```
DhairyaSethi, line 115: we dont need these right? should use keep current and asset update

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
DhairyaSethi: ie we need to maintain this per hub, either offchain or onchain

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+    address spoke;
+  }
+
+  /// @notice Parameters for freezing all reserves on a Spoke.
+  /// @dev spokeConfigurator The SpokeConfigurator to use for this action.
+  /// @dev spoke The address of the Spoke.
+  struct SpokeFreeze {
+    ISpokeConfigurator spokeConfigurator;
+    address spoke;
+  }
```
DhairyaSethi: we said we wont have these right? only whole struct updates w keep current

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,465 @@
...
+  /// @dev hub The address of the Hub.
+  /// @dev spoke The address of the Spoke.
+  /// @dev assetIds The list of asset identifiers to register the Spoke for.
+  /// @dev configs The list of Spoke configurations to register.
+  struct SpokeToAssetsAddition {
+    IHubConfigurator hubConfigurator;
+    address hub;
+    address spoke;
+    uint256[] assetIds;
+    IHub.SpokeConfig[] configs;
```
DhairyaSethi: these should be in struct not two arrays

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
yan-man: in v3 they use more random/exotic vals for the sentinel like `type(uint256).max - 42`, which is more unlikely to be used. We could do similar (ie more random), although max-1 is prob also not super likely 

https://github.com/aave-dao/aave-v3-origin/blob/1e3d70c4151a94166ebc59e2eaa4aff6e6ba6978/src/contracts/extensions/v3-config-engine/EngineFlags.sol#L7

src/config-engine/AaveV4Payload.sol
```diff
@@ -0,0 +1,437 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {Address} from 'src/dependencies/openzeppelin/Address.sol';
+
```
yan-man: nit: whitespace

src/config-engine/AaveV4Payload.sol
```diff
@@ -0,0 +1,437 @@
...
+  using Address for address;
+
+  /// @notice The config engine used to execute payload actions via delegatecall.
+  IAaveV4ConfigEngine public immutable CONFIG_ENGINE;
+
+  /// @dev Thrown when the config engine address is zero.
+  error InvalidConfigEngine();
+
+  /// @param configEngine The IAaveV4ConfigEngine implementation to delegatecall into.
+  constructor(IAaveV4ConfigEngine configEngine) {
```
yan-man:
```suggestion
  constructor(IAaveV4ConfigEngine configEngine_) {
```
nit for consistency, but dont know if we want to have that here too

src/config-engine/AaveV4Payload.sol
```diff
@@ -0,0 +1,437 @@
...
+    CONFIG_ENGINE = configEngine;
+  }
+
+  /// @notice Main execution entry point called by governance. Runs all configured actions.
+  /// @dev Expected to be called by a governance executor. No on-chain access control is applied;
+  ///   the caller is responsible for authorization. Idempotency is not guaranteed.
+  function execute() external {
+    _preExecute();
+    _executeHubActions();
+    _executeSpokeActions();
```
yan-man, line 31: do you think there's ever situations where we might want diff order of ops here. Should that be a separate payload, or configurable? or handle later as need arises

Kogaroshi: yeah, not sure, we might need specific payloads for that, or a Spoke only payload and then a Hub only payload executed in the wanted order.

yan-man: maybe we can handle that as needed later then

src/config-engine/AaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,131 @@
+// SPDX-License-Identifier: UNLICENSED
+// Copyright (c) 2025 Aave Labs
+pragma solidity ^0.8.0;
+
+import {HubEngine} from 'src/config-engine/libraries/HubEngine.sol';
+import {SpokeEngine} from 'src/config-engine/libraries/SpokeEngine.sol';
+import {AccessManagerEngine} from 'src/config-engine/libraries/AccessManagerEngine.sol';
+import {PositionManagerEngine} from 'src/config-engine/libraries/PositionManagerEngine.sol';
```
yan-man, line 7: do you think we need one for treasury spoke, ie like the supply actions, will those happen thru here, or just simpler fashion (ie not thru payload but just a tx from a MS maybe)

src/config-engine/AaveV4Payload.sol
yan-man: just that it gets applied at the time of execution, rather than separately validating access control here?

src/config-engine/libraries/EngineFlags.sol
```diff
@@ -0,0 +1,43 @@
...
+
+/// @title EngineFlags
+/// @author Aave Labs
+/// @notice Sentinel values for partial updates in config engine structs.
+library EngineFlags {
+  /// @dev Thrown when toBool receives a value other than 0 or 1.
+  error InvalidBoolValue(uint256 value);
+
+  /// @dev Sentinel value to keep the current uint value.
+  uint256 internal constant KEEP_CURRENT = type(uint256).max - 1;
```
yan-man: perhaps these can be a bit more random offset values?

tests/config-engine/EngineFlags.t.sol
```diff
@@ -0,0 +1,81 @@
...
+    return EngineFlags.toBool(flag);
+  }
+
+  function fromBool(bool value) external pure returns (uint256) {
+    return EngineFlags.fromBool(value);
+  }
+}
+
+contract EngineFlagsTest is Test {
+  EngineFlagsHarness harness;
```
yan-man:
```suggestion
  EngineFlagsHarness internal _harness;
```

tests/config-engine/EngineFlags.t.sol
```diff
@@ -0,0 +1,81 @@
...
+    assertFalse(harness.toBool(0));
+  }
+
+  function test_toBool_one_returnsTrue() public view {
+    assertTrue(harness.toBool(1));
+  }
+
+  function test_toBool_revertsOnInvalidValue() public {
+    vm.expectRevert(abi.encodeWithSelector(EngineFlags.InvalidBoolValue.selector, 2));
+    harness.toBool(2);
```
yan-man: nit: can do randomUint > 1 for more coverage

tests/mocks/config-engine/MockSpokeReader.sol
```diff
@@ -0,0 +1,42 @@
...
+
+  function getReserveId(address hub, uint256 assetId) external view returns (uint256) {
+    return _reserveIds[keccak256(abi.encode(hub, assetId))];
+  }
+
+  function getDynamicReserveConfig(
+    uint256 reserveId,
+    uint32 dynamicConfigKey
+  ) external view returns (ISpoke.DynamicReserveConfig memory) {
+    if (shouldRevertOnRead) revert(REVERT_MSG);
```
yan-man:
```suggestion
    if (shouldRevertOnRead) {
    revert(REVERT_MSG);
    }
```

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,435 @@
...
+/// @notice Interface for the Aave V4 Config Engine, defining all structs and engine method signatures.
+/// The engine is stateless and invoked via delegatecall from payload contracts.
+/// All numeric fields in config structs use uint256 so that type(uint256).max can serve as
+/// the universal KEEP_CURRENT sentinel. Boolean fields use uint256 (0=false, 1=true, KEEP_CURRENT=skip).
+interface IAaveV4ConfigEngine {
+  /// @notice Parameters for tokenization of an asset on a Hub when listing the asset.
+  /// @dev addCap The add cap for the TokenizationSpoke (0 means no tokenization).
+  /// @dev name The name for the TokenizationSpoke.
+  /// @dev symbol The symbol for the TokenizationSpoke.
+  struct TokenizationConfig {
```
yan-man:
```suggestion
  struct TokenizationSpokeConfig {
```
?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -0,0 +1,435 @@
...
+    uint256 addCap;
+    string name;
+    string symbol;
+  }
+
+  /// @notice Parameters for listing a new asset on a Hub.
+  /// @dev hubConfigurator The HubConfigurator to use for this action.
+  /// @dev hub The address of the Hub.
+  /// @dev underlying The address of the underlying asset.
+  /// @dev decimals Explicit decimals (0 means use addAsset which auto-detects decimals).
```
yan-man:
```suggestion
  /// @dev decimals Explicit decimals (0 means use `addAsset` which reads decimals from token).
```

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
yan-man: maybe we should have data provider contracts like in v3

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,337 @@
...
+          listings[i].underlying,
+          listings[i].decimals.toUint8(),
+          listings[i].feeReceiver,
+          listings[i].liquidityFee,
+          listings[i].irStrategy,
+          irData
+        );
+      }
+
+      if (listings[i].tokenization.addCap > 0) {
```
yan-man: hm do you think that we would ever deploy a tokenization spoke with addCap of 0 legitimately (ie just to launch with), maybe we can base this on name or symbol or somethign else like another engine flag

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,337 @@
...
+        _deployAndRegisterTokenizationSpoke(listings[i]);
```
yan-man: also tbh it is a little weird to have the hub engine be deploying a spoke here. ig it depends on framing, do we see tokenization spoke as an inherent part of the hub? perhaps this can be in the spoke engine and we call it from there

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,337 @@
...
+  ///   Reinvestment: address set → updateReinvestmentController.
+  /// @param updates The asset config updates to execute.
+  function executeHubAssetConfigUpdates(
+    IAaveV4ConfigEngine.AssetConfigUpdate[] calldata updates
+  ) external {
+    uint256 length = updates.length;
+    for (uint256 i; i < length; ++i) {
+      uint256 assetId = IHubBase(updates[i].hub).getAssetId(updates[i].underlying);
+
+      bool updateFee = updates[i].liquidityFee != EngineFlags.KEEP_CURRENT;
```
yan-man, line 53: i think we can extract this till L95 into a separate helper for readability, same for ir strat

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  /// @notice Updates asset config (fee, interest rate, reinvestment) for assets on Hubs.
+  /// @dev Dispatches to the appropriate HubConfigurator methods based on sentinel values:
+  ///   Fee: both set → updateFeeConfig; only fee → updateLiquidityFee; only receiver → updateFeeReceiver.
+  ///   IR: strategy set → updateInterestRateStrategy; strategy kept + non-sentinel irData fields →
+  ///   read-modify-write via updateInterestRateData.
+  ///   Reinvestment: address set → updateReinvestmentController.
+  /// @param updates The asset config updates to execute.
+  function executeHubAssetConfigUpdates(
```
yan-man, line 46: any reason you wanted to name it this way instead of `executeHubUpdateAssetConfig` and `executeHubAddSpokeToAssets` below? but i also dont think we need to have the "Hub" part in these fns given it's the HubEngine

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,337 @@
...
+import {IHubBase} from 'src/hub/interfaces/IHubBase.sol';
+import {IHub} from 'src/hub/interfaces/IHub.sol';
+import {IAssetInterestRateStrategy} from 'src/hub/interfaces/IAssetInterestRateStrategy.sol';
+
+import {IAaveV4ConfigEngine} from 'src/config-engine/interfaces/IAaveV4ConfigEngine.sol';
+
+/// @title HubEngine
+/// @author Aave Labs
+/// @notice Library containing hub configurator logic for AaveV4ConfigEngine.
+library HubEngine {
```
yan-man, line 15: i see v3 has quite a bit of granularity on their engines, like a CapsEngine, a ListingEngine, etc. Do you think worth it to split into more granularity, or is that too much to keep track of?

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,337 @@
...
+
+    anyUpdated = updateOptimal || updateBase || updateBefore || updateAfter;
+    if (!anyUpdated) return '';
+
+    address irStrategy = IHub(hub).getAssetConfig(assetId).irStrategy;
+    IAssetInterestRateStrategy.InterestRateData memory current = IAssetInterestRateStrategy(
+      irStrategy
+    ).getInterestRateData(assetId);
+
+    if (updateOptimal) current.optimalUsageRatio = irData.optimalUsageRatio;
```
yan-man: nit: braces on the if branches

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
DhairyaSethi: we definitely should

src/config-engine/libraries/HubEngine.sol
```diff
@@ -0,0 +1,337 @@
...
+    IAaveV4ConfigEngine.SpokeConfigUpdate[] calldata updates
+  ) external {
+    uint256 length = updates.length;
+    for (uint256 i; i < length; ++i) {
+      uint256 assetId = IHubBase(updates[i].hub).getAssetId(updates[i].underlying);
+
+      bool updateAdd = updates[i].addCap != EngineFlags.KEEP_CURRENT;
+      bool updateDraw = updates[i].drawCap != EngineFlags.KEEP_CURRENT;
+
+      if (updateAdd && updateDraw) {
```
yan-man: same, maybe this can be in a helper

tests/mocks/config-engine/AaveV4PayloadWrapper.sol
```diff
@@ -0,0 +1,383 @@
...
+
+  function _postExecute() internal override {
+    postExecuteCalled = true;
+    postExecuteOrder = ++_callCounter;
+  }
+
+  // Hub setters
+  function setHubAssetListings(IAaveV4ConfigEngine.AssetListing[] memory items) external {
+    delete _hubAssetListings;
+    for (uint256 i = 0; i < items.length; i++) _hubAssetListings.push(items[i]);
```
yan-man: nit: braces

tests/mocks/config-engine/MockAccessManager.sol
```diff
@@ -0,0 +1,156 @@
...
+  error LabelRoleReverted();
+  error SetGrantDelayReverted();
+  error SetTargetAdminDelayReverted();
+
+  function setShouldRevert(bytes4 selector, bool revert_) external {
+    shouldRevert[selector] = revert_;
+  }
+
+  function grantRole(uint64 roleId, address account, uint32 executionDelay) external {
+    if (shouldRevert[msg.sig]) revert GrantRoleReverted();
```
yan-man: nit: braces

src/config-engine/AaveV4Payload.sol
Kogaroshi: Yes, try to say that we do not verify the needed roles are given before executing the actions of the Payload, and Executor will need all the required Roles for the complete execution beforehand.
And that we rely on the Configurators & contracts access control directly.

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: We will have the address-book (or a Data Provider on chain works too) to fill those addresses when building the payload.
But yes I agree that a Data Provider would be more future proof.

src/config-engine/libraries/HubEngine.sol
Kogaroshi: In that case, we'd need a new bool in the struct to know if we want to deploy it, wdyt @DhairyaSethi ?

src/config-engine/libraries/HubEngine.sol
Kogaroshi: Imo the Spoke Engine is for Spoke.sol & SpokeConfigurator.sol, and the TokenizationSpoke is more of a "periphery" of an asset on the Hub, hence why it was here (and also because we only need deploy + call a config on the Hub here, so made more sense there as well).

src/config-engine/libraries/HubEngine.sol
Kogaroshi: The `Hub` part in the name is because that method is in `IAaveV4ConfigEngine`.
As for the naming, it was to show that we have a list of updates (and additions), ant not only one, but open to other names here.

src/config-engine/libraries/HubEngine.sol
Kogaroshi: Since each library will be deployed onchain, I don't think there is a need for splitting those Engine libraries even more, we already have enough granularity via the different methods imo.

tests/mocks/config-engine/MockAccessManager.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

tests/mocks/config-engine/AaveV4PayloadWrapper.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/libraries/HubEngine.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/libraries/HubEngine.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/libraries/HubEngine.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

tests/mocks/config-engine/MockSpokeReader.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

tests/config-engine/EngineFlags.t.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

tests/config-engine/EngineFlags.t.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/libraries/EngineFlags.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/AaveV4Payload.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/AaveV4Payload.sol
Kogaroshi: 01f8064b52ce83283032aa860c956a9152c476e9

src/config-engine/libraries/AccessManagerEngine.sol
yan-man: > the actor needing this will just be assigned all roles or smth

yes

src/config-engine/libraries/HubEngine.sol
DhairyaSethi: yes we want to deploy w 0 addCap for coverage

src/config-engine/README.md
```diff
@@ -0,0 +1,146 @@
...
+### The `KEEP_CURRENT` sentinel pattern
+
+The `EngineFlags` library defines sentinel values for each width that needs a "skip" marker:
+
+| Constant               | Type      | Value                        |
+| ---------------------- | --------- | ---------------------------- |
+| `KEEP_CURRENT`         | `uint256` | `type(uint256).max`          |
+| `KEEP_CURRENT_ADDRESS` | `address` | `address(type(uint160).max)` |
+| `KEEP_CURRENT_UINT64`  | `uint64`  | `type(uint64).max`           |
+| `KEEP_CURRENT_UINT32`  | `uint32`  | `type(uint32).max`           |
```
CheyenneAtapour: These have been updated, so this is out of date and no longer matches the code. 
For example, `KEEP_CURRENT` is now `type(uint256).max - 256`

Kogaroshi: 43c7352aec0db56c41b0eb40a6c4043c376bda77

src/config-engine/libraries/HubEngine.sol
Kogaroshi: 43c7352aec0db56c41b0eb40a6c4043c376bda77

tests/config-engine/BaseConfigEngine.t.sol
yan-man: i like this idea (particularly for tests) and in general to have default configs

Kogaroshi: addressed in ce77d0ed51fa0f623dbf02351605c31e9e0841ff
we can see after merging the test rft to have a more global default config.

PR #1235, opened 2026-03-02
chore: cleanup prototype

DhairyaSethi:


yan-man:
lean towards not including the prototype and rm instead as it's additional overhead if we ever add spoke changes to keep in sync, but dont feel super strongly

PR #1236, opened 2026-03-02
wip: tokenized fee receiver

DhairyaSethi:


DhairyaSethi:
we have planned to implement this outside the core protocol by doing a (2 share) lossy transfer to tokenise instead

PR #1237, opened 2026-03-03
docs: Add IsolationMode tech doc

CanonicalJP:
Adds `docs/features/IsolationMode.md`, a technical reference for the IsolationMode implementation.


docs/features/IsolationMode.md
```diff
@@ -0,0 +1,51 @@
+# Isolation Mode
+
+## Summary
+
+Isolation Mode enables new assets to be added to the protocol with contained risk by deploying them in a dedicated Spoke and Liquidity Hub pair separate from the canonical market. Users who supply an isolated collateral asset in this Spoke can only borrow assets configured as borrowable for that market, up to per-asset draw caps set by the Governor. Because the isolated market lives in its own Spoke, users continue to hold unrestricted positions in the Canonical Spoke and any other Spoke simultaneously.
```
CheyenneAtapour:
```suggestion
Isolation Mode enables new collateral assets to be added to the protocol with contained risk by deploying them in a dedicated Spoke and Liquidity Hub pair separate from the canonical market. Users who supply an isolated collateral asset in this Spoke can only borrow assets configured as borrowable for that market, up to per-asset draw caps set by the Governor. Because the isolated market lives in its own Spoke, users continue to hold unrestricted positions in the Canonical Spoke and any other Spoke simultaneously.
```

I want us to emphasize that the point of isolation mode is to allow new or experimental `collateral` assets, specifically. This is in contrast to siloed borrowing which deals with "isolating" borrowable assets

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,51 @@
...
+
+To bootstrap USDC liquidity, USDC is listed as an additional reserve (USDC') on the isolated Spoke with `Reserve.hub` pointing to the Canonical LH. The Governor sets a draw cap on USDC' in the Canonical LH, limiting total exposure of canonical USDC suppliers to the isolated market. Borrowers in the isolated Spoke draw USDC' from the Canonical LH under this cap.
+
+Once the isolated LH has accumulated sufficient native USDC liquidity, the Governor sets the draw cap for USDC' to zero. Existing USDC' borrow positions can still be repaid, but no new borrows are possible. Users then borrow USDC directly from the isolated LH. At this point the credit line is wound down and the isolated market is self-sustaining.
+
+## Out of Scope
+
+The following are explicitly excluded from Isolation Mode as a configuration pattern:
+
+- **Account-level enforcement**: Isolation Mode is a Spoke-level pattern, not an account-level lock. The contracts do not prevent a user from holding positions in both the isolated Spoke and the Canonical Spoke simultaneously.
```
CheyenneAtapour:
```suggestion
- **Account-level enforcement**: Isolation Mode is a Spoke-level pattern, not an account-level lock. The contracts do not prevent a user from holding positions in the isolated Spoke, the Canonical Spoke, or any other spoke simultaneously.
```

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,51 @@
...
+
+Each reserve on a Spoke references a specific Liquidity Hub through `Reserve.hub`. This means a single Spoke can draw liquidity from multiple Hubs, each subject to its own draw cap per asset. Draw caps are configured via `HubConfigurator.updateSpokeDrawCap` by governance-authorized roles (AccessManaged `restricted` access).
+
+## Collateral and Borrow Configuration
+
+The collateral set in an isolated market is defined by which reserves on the Spoke carry a non-zero `collateralFactor` in their `DynamicReserveConfig`. A reserve with `collateralFactor > 0` contributes to a user's health factor when the user has activated it as collateral via `setUsingAsCollateral`. Any number of reserves can be configured this way within the isolated Spoke.
+
+The borrowable set is controlled by the `borrowable` flag in `ReserveConfig`. Reserves with `borrowable = false` reject borrow requests; the Spoke reverts with `ReserveNotBorrowable`. The Governor designates borrowable assets by setting `borrowable = true` on selected reserves. There is no protocol-level constraint limiting borrowable assets to stablecoins.
```
CheyenneAtapour: I think you can omit the last sentence. It would be relevant maybe if you were making a comparison to something else, but that is not present here. You could say there is no protocol-level constraint for limiting borrowable assets; any type of reserve can be configured as borrowable in an isolated market

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,51 @@
...
+
+Once the isolated LH has accumulated sufficient native USDC liquidity, the Governor sets the draw cap for USDC' to zero. Existing USDC' borrow positions can still be repaid, but no new borrows are possible. Users then borrow USDC directly from the isolated LH. At this point the credit line is wound down and the isolated market is self-sustaining.
+
+## Out of Scope
+
+The following are explicitly excluded from Isolation Mode as a configuration pattern:
+
+- **Account-level enforcement**: Isolation Mode is a spoke-level pattern, not an account-level lock. The contracts do not prevent a user from holding positions in the isolated Spoke, the Canonical Spoke, or any other Spoke simultaneously.
+- **Cross-Spoke health factor aggregation**: Collateral and debt in the isolated Spoke are accounted independently. Health factors are not aggregated across Spokes.
+- **Stablecoin-only borrowable constraint**: There is no protocol-level rule restricting borrowable assets to stablecoins; that is a governance configuration choice.
```
CheyenneAtapour: This is mentioned already here: https://github.com/aave/aave-v4/pull/1237/changes#r2991940195 and also again below. I think both in my previous comment and here should be removed. We can keep the one at the bottom which compares to V3. I think that's the point that is most relevant

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
+# Isolation Mode
+
+## Summary
+
+Isolation Mode enables new collateral assets to be added to the protocol with contained risk by deploying them in a dedicated Spoke and Liquidity Hub pair separate from the canonical market. Users who supply an isolated collateral asset in this Spoke can only borrow assets configured as borrowable for that market, up to per-asset draw caps set by the Governor. Because the isolated market lives in its own Spoke, users continue to hold unrestricted positions in the Canonical Spoke and any other Spoke simultaneously.
```
yan-man:
```suggestion
Isolation Mode allows the protocol to contain risk associated with certain collateral assets by adding them in a dedicated Spoke and Liquidity Hub pair separated from the canonical market. Users who supply an isolated collateral asset in this Spoke can only borrow assets up to per-asset draw caps set by the Governor. Because the isolated market lives in its own Spoke, users continue to hold independent positions in the Canonical Spoke and any other Spoke simultaneously.
```

- let's not use "deploying" (invoking smart contract terms) but rather "adding" or "registering"
- what defines a market in your view? is it a hub/spoke cluster

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+Unlike Aave V3, which enforced a single aggregated borrow ceiling per isolated market and locked the borrower to one isolated collateral asset per account, Aave V4 assigns draw caps independently per asset and permits any number of assets to be configured as collateral within the isolated Spoke. The risk boundary is structural: it is scoped to the Spoke, not to the user's account.
+
+## Relationship to the Hub/Spoke Architecture
+
+Isolation Mode is not a contract-level primitive but a configuration pattern applied to a Spoke and its associated Liquidity Hub. Any Spoke becomes an isolated market when its reserves are configured with a restricted borrowable set and its Hub enforces per-asset draw caps on that Spoke.
```
yan-man: there is not a single associated LH for a given Spoke, as it is tied to the reserve on the Spoke. So a spoke could have 1 (coll) asset which is tied to hubA but 10 other assets tied to hubB. Would this be considered isolation mode still? the isolation part refers mostly to the collateral risk right, and so the debt assets are "safe" in that theyre exposed from hubB to a risky asset on the spoke, but still only up to the drawCap. Here in v4 we have some cross hub exposure due to this (but still only up to caps)

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Collateral and Borrow Configuration
+
+The collateral set in an isolated market is defined by which reserves on the Spoke carry a non-zero `collateralFactor` in their `DynamicReserveConfig`. A reserve with `collateralFactor > 0` contributes to a user's health factor when the user has activated it as collateral via `setUsingAsCollateral`. Any number of reserves can be configured this way within the isolated Spoke.
+
+The borrowable set is controlled by the `borrowable` flag in `ReserveConfig`. Reserves with `borrowable = false` reject borrow requests; the Spoke reverts with `ReserveNotBorrowable`. The Governor designates borrowable assets by setting `borrowable = true` on selected reserves.
+
+A draw cap on the Hub supplements the `borrowable` flag. The `borrowable` flag gates access at the Spoke level; the draw cap enforces an absolute ceiling on how much the Spoke may draw from a given Hub asset. Both constraints must be satisfied for a borrow to succeed.
```
yan-man, line 23: draw cap also includes current outstanding reported deficit

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Bootstrapping Liquidity
+
+A new isolated market initially has no native liquidity. Without liquidity there are no borrowers, and without borrowers, suppliers earn no competitive rates. Aave V4 addresses this by allowing a reserve in the isolated Spoke to reference the Canonical Liquidity Hub rather than the new isolated LH.
```
yan-man: Define LH above somewhere like Liquidity Hub (LH), or just "rather than the new isolated hub." ?

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+To bootstrap USDC liquidity, USDC is listed as an additional reserve (USDC') on the isolated Spoke with `Reserve.hub` pointing to the Canonical LH. The Governor sets a draw cap on USDC' in the Canonical LH, limiting total exposure of canonical USDC suppliers to the isolated market. Borrowers in the isolated Spoke draw USDC' from the Canonical LH under this cap.
```
yan-man: -> "For example, to bootstrap USDC liquidity..."

-> "...USDC from the Canonical LH is listed as an additional reserve (USDC') on the isolated Spoke..."

-> "from the Canonical LH up to this cap"

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+Once the isolated LH has accumulated sufficient native USDC liquidity, the Governor sets the draw cap for USDC' to zero. Existing USDC' borrow positions can still be repaid, but no new borrows are possible. Users then borrow USDC directly from the isolated LH. At this point the credit line is wound down and the isolated market is self-sustaining.
```
yan-man: this one confuses me. I don't think USDC' can translate directly to USDC liquidity bc they're considered separate assets. Let's say the new Iso Hub has: USDC, CollA collateral. iso spoke has USDC', USDC, and CollA listed
- users use CollA to borrow USDC' up to cap
- USDC' drawCap on Canonical Hub then is set to 0
- it's the `existing` "stuck" CollA collateral that then attracts more organic USDC suppliers to Iso Spoke right? The USDC' still needs to get restored back to the Canonical Hub, and isnt maintained. 

So the bootstrapping is more like USDC' -> attract collateral -> attract USDC?

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Out of Scope
+
+The following are explicitly excluded from Isolation Mode as a configuration pattern:
+
+- **Account-level enforcement**: Isolation Mode is a spoke-level pattern, not an account-level lock. The contracts do not prevent a user from holding positions in the isolated Spoke, the Canonical Spoke, or any other Spoke simultaneously.
```
yan-man, line 37:
```suggestion
- **Account-level enforcement**: Isolation Mode is a spoke-level pattern, not an account-level lock. The contracts do not prevent a user from holding positions in the Isolated Spoke, the Canonical Spoke, or any other Spoke simultaneously.
```

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+- **Cross-Spoke health factor aggregation**: Collateral and debt in the isolated Spoke are accounted independently. Health factors are not aggregated across Spokes.
```
yan-man:
```suggestion
- **Cross-Spoke health factor aggregation**: Collateral and debt in the Isolated Spoke are accounted for independently. Health factors are not aggregated across Spokes.
```

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Multiple isolated collaterals**: In Aave V3, only one isolated asset can serve as collateral per account. In Aave V4, any number of reserves in the isolated Spoke can be configured with a non-zero `collateralFactor`, making them eligible as collateral simultaneously.
+
+**Per-asset draw caps**: Aave V3 enforces a single aggregated borrow ceiling shared across all borrowable assets in the isolated market. A 10,000 USD cap shared among USDC, USDT, and GHO allows a single asset to consume the entire budget, blocking all others. Aave V4 assigns draw caps independently per asset: a 3,000 USDC draw cap, 3,000 USDT draw cap, and 4,000 GHO draw cap each enforce a distinct limit so no single asset crowds out the rest.
+
+**Unrestricted borrowable assets**: In Aave V3, isolated mode restricts borrowing to a governance-approved list of stablecoins. In Aave V4, any asset can be marked `borrowable = true` in the isolated Spoke; the Governor defines the composition without a stablecoin-only constraint.
```
yan-man:
```suggestion
**Unrestricted borrowable assets**: In Aave V3, Isolation Mode restricts borrowing to a governance-approved list of stablecoins. In Aave V4, any asset can be marked `borrowable = true` in the isolated Spoke; the Governor defines the composition without a stablecoin-only constraint.
```

but in v3 it isn't enforced that it must be a stablecoin, no? just that it is an approved asset list

docs/features/IsolationMode.md
CheyenneAtapour: The ' is just to show that it's a USDC from a different hub, so don't worry too much about that.
You are just borrowing USDC from one hub and supplying into another. That's the bootstrapped liquidity

docs/features/IsolationMode.md
CheyenneAtapour: I kinda like the way it is written, because it gets to the exact point about the new collateral assets. Yours feels a bit more passively written. But if you don't like those words (and I can see your point there), we can re-write it to mention the collateral assets as the focal point of the sentence.

I'd say a market is a spoke

docs/features/IsolationMode.md
CheyenneAtapour: Sure, but unsure if we need to go into that detail here

docs/features/IsolationMode.md
CheyenneAtapour: They are governance-approved stablecoins that can be borrowed against the isolated collateral. But the isolated collateral can be whatever asset, doesn't have to be stablecoin

docs/features/IsolationMode.md
CheyenneAtapour: That is technically true, you could have a setup that falls under "Isolation Mode", but it's just 1 collateral asset, and then the borrowables come from 2 or more different hubs. But I think to mention that is too complex that necessary to mention here, and we likely would just do isolation mode as 1 spoke 1 hub, unless there is some extraneous reason to have it multi-hub

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+- **Account-level enforcement**: Isolation Mode is a spoke-level pattern, not an account-level lock. The contracts do not prevent a user from holding positions in the isolated Spoke, the Canonical Spoke, or any other Spoke simultaneously.
+- **Cross-Spoke health factor aggregation**: Collateral and debt in the isolated Spoke are accounted independently. Health factors are not aggregated across Spokes.
+
+## Key Differences from Aave V3
+
+**Spoke-scoped isolation**: In Aave V3, Isolation Mode applies at the account level. A user who supplies an isolated collateral asset cannot use any other asset as collateral in that account; if non-isolated collateral is already active, the isolated asset cannot be added until it is removed. In Aave V4, Isolation Mode is spoke-scoped: the user is isolated within the isolated Spoke but can hold unrestricted positions in any other Spoke at the same time.
+
+**Multiple isolated collaterals**: In Aave V3, only one isolated asset can serve as collateral per account. In Aave V4, any number of reserves in the isolated Spoke can be configured with a non-zero `collateralFactor`, making them eligible as collateral simultaneously.
```
CheyenneAtapour: We could also mention that on V4, a user, with one account, can participate in multiple isolated markets at once. Either can put here or in the point above. It's hinted at on the point above, but we can be explicit about it

docs/features/IsolationMode.md
yan-man: dont feel too strongly about the stylistic changes, i think just the "deploying" part matters to me that i'd prefer changed

CheyenneAtapour: cool, yeah im aligned with that

docs/features/IsolationMode.md
CanonicalJP: I made just small tweaks since it's not enforced and we're not certain how this will evolve under the DAO's management

docs/features/IsolationMode.md
CanonicalJP: not strong feelings here, inclined to keep it as it's

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+- **Account-level enforcement**: Isolation Mode is a spoke-level pattern, not an account-level lock. The contracts do not prevent a user from holding positions in the isolated Spoke, the Canonical Spoke, or any other Spoke simultaneously.
+- **Cross-Spoke health factor aggregation**: Collateral and debt in the isolated Spoke are accounted independently. Health factors are not aggregated across Spokes.
+
+## Key Differences from Aave V3
+
+**Spoke-scoped isolation**: In Aave V3, Isolation Mode applies at the account level. A user who supplies an isolated collateral asset cannot use any other asset as collateral in that account; if non-isolated collateral is already active, the isolated asset cannot be added until it is removed. In Aave V4, Isolation Mode is spoke-scoped: the user is isolated within the isolated Spoke but can hold unrestricted positions in any other Spoke at the same time. This also means a single account can participate in multiple isolated markets simultaneously. For example, supplying collateral in one isolated Spoke while also borrowing against collateral in a separate isolated Spoke, with neither position affecting the other.
```
CheyenneAtapour: Easier to read imo just to have a clean break and 2 separate sentences
```suggestion
**Spoke-scoped isolation**: In Aave V3, Isolation Mode applies at the account level. A user who supplies an isolated collateral asset cannot use any other asset as collateral in that account. If non-isolated collateral is already active, the isolated asset cannot be added until it is removed. In Aave V4, Isolation Mode is spoke-scoped: the user is isolated within the isolated Spoke but can hold unrestricted positions in any other Spoke at the same time. This also means a single account can participate in multiple isolated markets simultaneously. For example, supplying collateral in one isolated Spoke while also borrowing against collateral in a separate isolated Spoke, with neither position affecting the other.
```

docs/features/IsolationMode.md
CheyenneAtapour: This change did not get applied

CheyenneAtapour: can keep "isolated" lower-case, but need the "for"

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
+# Isolation Mode
+
+## Summary
+
+Isolation Mode enables new collateral assets to be added to the protocol with contained risk by registering them in a dedicated Spoke and Hub pair separate from the canonical market. Users who supply an isolated collateral asset in this Spoke can only borrow assets configured as borrowable for that market, up to per-asset draw caps set by the Governor. Because the isolated market lives in its own Spoke, users continue to hold unrestricted positions in the Canonical Spoke and any other Spoke simultaneously.
```
CheyenneAtapour:
```suggestion
Isolation Mode enables new collateral assets to be added to the protocol with contained risk by registering them in a dedicated Spoke and Hub pair separate from the canonical market. Users who supply an isolated collateral asset in this Spoke can only borrow assets configured as borrowable for that market, up to per-asset draw caps set by the Governor. Because the isolated market lives in its own Spoke, users can continue to hold unrestricted positions in the Canonical Spoke and any other Spoke simultaneously.
```

docs/features/IsolationMode.md
```diff
@@ -0,0 +1,50 @@
...
+
+**Multiple isolated collaterals**: In Aave V3, only one isolated asset can serve as collateral per account. In Aave V4, any number of Reserves in the isolated Spoke can be configured with a non-zero `collateralFactor`, making them eligible as collateral simultaneously.
+
+**Per-asset draw caps**: Aave V3 enforces a single aggregated borrow ceiling shared across all borrowable assets in the isolated market. A 10,000 USD cap shared among USDC, USDT, and GHO allows a single asset to consume the entire budget, blocking all others. Aave V4 assigns draw caps independently per asset: a 3,000 USDC draw cap, 3,000 USDT draw cap, and 4,000 GHO draw cap each enforce a distinct limit so no single asset crowds out the rest.
+
+**Unrestricted borrowable assets**: In Aave V3, Isolation Mode restricts borrowing to a governance-approved list of assets. In Aave V4, any asset can be marked `borrowable = true` in the isolated Spoke; the Governor defines the composition without a stablecoin-only constraint.
```
CheyenneAtapour:
```suggestion
**Unrestricted borrowable assets**: In Aave V3, Isolation Mode restricts borrowing to a governance-approved list of stablecoins. In Aave V4, any asset can be marked `borrowable = true` in the isolated Spoke; the Governor defines the composition without a stablecoin-only constraint.
```

PR #1241, opened 2026-03-03
feat: Revert if collateral dust in liquidations

avniculae:


avniculae:
`debtToTarget` restriction, together with new restriction added in the PR, might prevent liquidations from happening

PR #1243, opened 2026-03-04
opt: cache on stack for ir strategy

DhairyaSethi:
Replace InterestRateData struct with a PackedInterestRateData UDVT (uint256) for internal storage and computation, eliminating unnecessary memory allocations by keeping rate data on the stack


DhairyaSethi:
ahead of its time 😞

PR #1249, opened 2026-03-07
fix: remove ISpokeBase

miguelmtzinf:
The initial motivation for `ISpokeBase` was to define the minimal functionality a Spoke should implement, so integrators (and position managers) could rely on it and work across all existing Spokes.

However, this approach does not hold in practice because a spoke is any borrowing logic that connects to the Hub to access liquidity, and the implementations can vary widely. For instance, `TokenizationSpoke` and `TreasurySpoke` do not follow `ISpokeBase`, and future Spokes may also not implement the functionality defined there

As a result, we consolidated everything into a single interface to simplify maintainability and integrations.

Blocked by #950

### DIFF
--- a/tests/unit/Spoke/TreasurySpoke.t.sol
+++ b/tests/unit/Spoke/TreasurySpoke.t.sol
@@ -75,16 +75,18 @@ contract TreasurySpokeTest is SpokeBase {
     assertEq(Ownable2Step(address(treasurySpoke)).pendingOwner(), address(0));
   }
 
-  function test_supply_revertsWith_Unauthorized(address caller) public {
-    vm.assume(caller != TREASURY_ADMIN);
+  function test_supply_revertsWith_OwnableUnauthorizedAccount() public {
+    address caller = vm.randomAddress();
+    while (caller == TREASURY_ADMIN || caller == ADMIN) caller = vm.randomAddress();
 
     vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, caller));
     vm.prank(caller);
     treasurySpoke.supply(address(hub1), address(tokenList.dai), 1);
   }
 
-  function test_withdraw_revertsWith_Unauthorized(address caller) public {
-    vm.assume(caller != TREASURY_ADMIN);
+  function test_withdraw_revertsWith_OwnableUnauthorizedAccount() public {
+    address caller = vm.randomAddress();
+    while (caller == TREASURY_ADMIN || caller == ADMIN) caller = vm.randomAddress();
 
     vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, caller));
     vm.prank(caller);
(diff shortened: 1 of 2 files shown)


src/spoke/TreasurySpoke.sol
```diff
@@ -73,49 +74,37 @@ contract TreasurySpoke is ITreasurySpoke, Ownable2Step {
     return HUB.getSpokeAddedShares(reserveId, address(this));
   }
 
-  /// @inheritdoc ISpokeBase
   function borrow(uint256, uint256, address) external pure returns (uint256, uint256) {
     revert UnsupportedAction();
   }
```
DhairyaSethi: we should even remove all of these no

miguelmtzinf: yes, to be made in the TreasurySpoke PR, so better to apply that one first

src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -620,7 +620,8 @@ library LiquidationLogic {
         collateralRemaining.toValue({
           decimals: params.collateralAssetDecimals,
           price: params.collateralAssetPrice
-        }) < DUST_LIQUIDATION_THRESHOLD;
+        }) <
```
avniculae: do you have the latest npm dependencies installed? If I run lint:fix on this branch all these changes are reverted

miguelmtzinf: fixed now, ty

tests/Base.t.sol
```diff
@@ -57,7 +57,7 @@ import {
 } from 'src/hub/AssetInterestRateStrategy.sol';
 
 // spoke
-import {ISpoke, ISpokeBase} from 'src/spoke/interfaces/ISpoke.sol';
+import {ISpoke, ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
```
yan-man: rm

tests/Utils.sol
```diff
@@ -5,7 +5,7 @@ pragma solidity ^0.8.0;
 import {Vm} from 'forge-std/Vm.sol';
 import {SafeERC20, IERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
 import {IHub, IHubBase} from 'src/hub/interfaces/IHub.sol';
-import {ISpokeBase, ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+import {ISpoke, ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
```
yan-man: rm

tests/unit/Spoke/TreasurySpoke.t.sol
```diff
@@ -75,16 +75,18 @@ contract TreasurySpokeTest is SpokeBase {
     assertEq(Ownable2Step(address(treasurySpoke)).pendingOwner(), address(0));
   }
 
-  function test_supply_revertsWith_Unauthorized(address caller) public {
-    vm.assume(caller != TREASURY_ADMIN);
+  function test_supply_revertsWith_OwnableUnauthorizedAccount() public {
+    address caller = vm.randomAddress();
+    while (caller == TREASURY_ADMIN || caller == ADMIN) caller = vm.randomAddress();
```
DhairyaSethi, line 80: for access control we said we wanted multiple iterations so an empty fuzz param or the same as before but conditions in vm assume are also ok 

but nbd

DhairyaSethi (approved):

tests/unit/position-manager/PositionManagerBase.t.sol
```diff
@@ -214,9 +214,11 @@ contract PositionManagerBaseTest is SpokeBase {
   }
 
   function test_registerSpoke_revertsWith_OwnableUnauthorizedAccount() public {
-    address user = vm.randomAddress();
-    vm.expectRevert(abi.encodeWithSelector(Ownable.OwnableUnauthorizedAccount.selector, user));
-    vm.prank(user);
+    address caller = vm.randomAddress();
+    while (caller == ADMIN) caller = vm.randomAddress();
```
avniculae, line 218: should we use an assume here instead?

DhairyaSethi: not when its a unit test, it will reject the only run, only useful for fuzz test

avniculae: yes, saw afterwards it's unit test. I think we shouldn't use vm.randomAddress() in unit tests when we also need to restrict its space, wdyt?

PR #1250, opened 2026-03-07
feat: Move deficit functionalities to IHub

CheyenneAtapour:
Moves the following from `IHubBase.sol` into `IHub.sol`:
* getAssetDeficitRay() function
* getSpokeDeficitRay() function

### DIFF
--- a/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenarios.t.sol
+++ b/tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenarios.t.sol
@@ -756,7 +756,7 @@ contract SpokeLiquidationCallScenariosTest is SpokeLiquidationCallBaseTest {
     _openSupplyPosition(spoke, collateralReserveId, MAX_SUPPLY_AMOUNT);
 
     vm.expectCall(
-      address(hub), 
+      address(hub),
       abi.encodeWithSelector(IHubBase.reportDeficit.selector, usdxAssetId)
     );
 


src/spoke/libraries/LiquidationLogic.sol
```diff
@@ -274,7 +275,7 @@ library LiquidationLogic {
     ) {
       ISpoke.UserPosition storage userPosition = userPositions[user][reserveId];
       ISpoke.Reserve storage reserve = reserves[reserveId];
-      IHubBase hub = reserve.hub;
+      IHub hub = IHub(address(reserve.hub));
```
CheyenneAtapour: Did it this way because there is a `hub.reportDeficit()` call later in this test, and I felt it cleaner to cast here rather than during the call

DhairyaSethi: then we cannot do this PR. normal spoke cannot depend on IHub, only IHubBase otherwise there's no point

src/hub/interfaces/IHubBase.sol
avniculae, line 261: I think we should also move `event ReportDeficit` and `function reportDeficit` right?

DhairyaSethi: we cannot bc the spoke relies on `reportDeficit` being in base

CheyenneAtapour:
We've decided not to proceed with this, because we want to keep spoke (liquidation logic here, specifically) dependent only on IHubBase, and not IHub. Moving the deficit functions to IHub breaks this. Moving just the getters is possible, but then it is quite strange to have a related function but getters on 2 different interfaces. So discarding this.

PR #1251, opened 2026-03-09
fix: TakerPositionManager reduces the allowance by the effective amount impacting the user position

Kogaroshi:
Refactor allowance deduction logic in TakerPositionManager to reduce the allowance by the true asset liquidity change (rounded up), to prevent scenario where share prices allow the spender to extract more than the given allowance.

### DIFF
--- a/src/position-manager/TakerPositionManager.sol
+++ b/src/position-manager/TakerPositionManager.sol
@@ -11,7 +11,7 @@ import {PositionManagerIntentBase} from 'src/position-manager/PositionManagerInt
 
 /// @title TakerPositionManager
 /// @author Aave Labs
-/// @notice Position manager to handle withdraw permit and borrow permit actions on behalf of users.
+/// @notice Position manager to handle withdraw and borrow actions on behalf of users.
 contract TakerPositionManager is ITakerPositionManager, PositionManagerIntentBase {
   using SafeERC20 for IERC20;
   using MathUtils for uint256;


src/position-manager/TakerPositionManager.sol
```diff
@@ -269,60 +320,54 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+  }
+
+  /// @dev Deducts the corrected amount from the given allowance.
+  /// `correctedAmount` may exceed `currentAllowance` by a rounding delta;
+  /// consumption is capped at `currentAllowance` to prevent underflow.
+  function _deductAllowance(
+    uint256 currentAllowance,
+    uint256 correctedAmount
+  ) internal pure returns (uint256 newAllowance) {
+    if (currentAllowance == type(uint256).max) return type(uint256).max;
```
miguelmtzinf: add brackets please

src/position-manager/TakerPositionManager.sol
```diff
@@ -269,60 +320,54 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+    uint256 consumption = currentAllowance >= correctedAmount ? correctedAmount : currentAllowance;
+    return currentAllowance - consumption;
```
miguelmtzinf:
```suggestion
    return currentAllowance > correctedAmount ? currentAllowance - correctedAmount : 0;
```
is not this the classic zeroFloorSub? we can use MathUtils.zeroFloorSub

src/position-manager/TakerPositionManager.sol
```diff
@@ -269,60 +320,54 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
-      });
-    }
+  }
+
+  /// @dev Deducts the corrected amount from the given allowance.
+  /// `correctedAmount` may exceed `currentAllowance` by a rounding delta;
+  /// consumption is capped at `currentAllowance` to prevent underflow.
+  function _deductAllowance(
+    uint256 currentAllowance,
+    uint256 correctedAmount
```
miguelmtzinf:
```suggestion
    uint256 amount
```

src/position-manager/TakerPositionManager.sol
```diff
@@ -269,60 +320,54 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
-        spender: spender,
-        newCreditDelegation: currentAllowance.uncheckedSub(amount)
-      });
-    }
+  }
+
+  /// @dev Deducts the corrected amount from the given allowance.
+  /// `correctedAmount` may exceed `currentAllowance` by a rounding delta;
+  /// consumption is capped at `currentAllowance` to prevent underflow.
+  function _deductAllowance(
```
miguelmtzinf: this is called inside `!=max` and we are checking `== max` here

Kogaroshi: forgot to remove this when adding the `!= max` check, will remove

src/position-manager/TakerPositionManager.sol
```diff
@@ -194,21 +219,47 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+    // Simply decreasing the allowance by the input `amount` is not ideal for shares-based
+    // debt. Due to rounding, the actual increase in the user's debt can differ slightly from
+    // the input `amount`. To handle this, the allowance consumption is based on the before/after
+    // delta of `previewRestoreByShares`, and capped at `currentAllowance` to prevent underflow
+    // from rounding.
+    if (currentAllowance != type(uint256).max) {
+      uint256 drawnSharesAfter = ISpoke(spoke).getUserPosition(reserveId, onBehalfOf).drawnShares;
+      uint256 newAllowance = _deductAllowance({
+        currentAllowance: currentAllowance,
+        correctedAmount: reserve.hub.previewRestoreByShares(reserve.assetId, drawnSharesAfter) -
```
miguelmtzinf: couldn't we just do this?
```suggestion
        correctedAmount: reserve.hub.previewRestoreByShares(reserve.assetId, borrowedShares)
```

Kogaroshi: No, after testing out using the Sim fuzz test, it appears that the returned shares can differ by 1 wei, and doing the before/after is more precise to get the true liquidity cost.
In the borrow case, using the returned shares can be precise enough in all tested case, but it it not for the withdraw case, where the "shares before/after => converted => difference" method is required to have the correct true liquidity cost to apply to the allowance.
So for consistency, this method was applied to both withdraw & borrow logics.

src/position-manager/TakerPositionManager.sol
Kogaroshi: 3b0f53fd2cb7376411757976fda630b836bc1f54

src/position-manager/TakerPositionManager.sol
Kogaroshi: 3b0f53fd2cb7376411757976fda630b836bc1f54

src/position-manager/TakerPositionManager.sol
Kogaroshi: 3b0f53fd2cb7376411757976fda630b836bc1f54

src/position-manager/TakerPositionManager.sol
Kogaroshi: 3b0f53fd2cb7376411757976fda630b836bc1f54

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +167,47 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+    });
+
+    // Simply decreasing the allowance by the input `amount` is not ideal for shares-based
+    // positions. Due to rounding in supply-side SharesMath, the actual decrease in the user's
+    // position value can differ slightly from the input `amount`. To handle this, the allowance
+    // consumption is based on the before/after delta of `previewAddByShares`, and capped at
+    // `currentAllowance` to prevent underflow from rounding.
+    if (currentAllowance != type(uint256).max) {
+      uint256 sharesAfter = ISpokeBase(spoke).getUserSuppliedShares(reserveId, onBehalfOf);
+      uint256 newAllowance = _deductAllowance({
```
DhairyaSethi: how much more does it cost if we move this up and do an additional previewWithdrawByAssets to simulate the preview action instead?
breaking cei in the context of this, even for purely known interactions warrants reentrency guard (sanity) but its not enough bc of read only reentrency (again only if the spoke was untrusted, which it is not here) so perhaps better to do this with an avoidable ext call & cold computation if gas increase is minimal

src/position-manager/TakerPositionManager.sol
```diff
@@ -194,21 +219,47 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+    // debt. Due to rounding, the actual increase in the user's debt can differ slightly from
+    // the input `amount`. To handle this, the allowance consumption is based on the before/after
+    // delta of `previewRestoreByShares`, and capped at `currentAllowance` to prevent underflow
+    // from rounding.
+    if (currentAllowance != type(uint256).max) {
+      uint256 drawnSharesAfter = ISpoke(spoke).getUserPosition(reserveId, onBehalfOf).drawnShares;
+      uint256 newAllowance = _deductAllowance({
+        currentAllowance: currentAllowance,
+        amount: reserve.hub.previewRestoreByShares(reserve.assetId, drawnSharesAfter) -
+          reserve.hub.previewRestoreByShares(reserve.assetId, drawnSharesBefore)
```
DhairyaSethi: maybe add a comment that this is enough and `premiumDebt` should remain unchanged, borrow can realise premium debt and potentially change its "rate"

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +167,47 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
-      amount,
-      onBehalfOf
-    );
-    asset.safeTransfer(msg.sender, withdrawnAmount);
+    uint256 sharesBefore;
+    if (currentAllowance != type(uint256).max) {
+      sharesBefore = ISpokeBase(spoke).getUserSuppliedShares(reserveId, onBehalfOf);
+    }
+
+    (uint256 withdrawnShares, uint256 withdrawnAmount) = ISpokeBase(spoke).withdraw({
```
avniculae: would it hurt if we would be a bit less exact, where we would convert `withdrawnShares` to assets (by rounding up), and we use that result to update the allowance? withdrawnShares into assets should be >= what we compute now as  `reserve.hub.previewAddByShares(reserve.assetId, sharesBefore) -reserve.hub.previewAddByShares(reserve.assetId, sharesAfter)`. to me this is acceptable and would help greatly with readability. We can pass `withdrawnShares` to `_spendWithdrawAllowance`, and skip check inside if allowance is max (as before). wdyt?

same for borrow ofc

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +167,47 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
     uint256 amount,
     address onBehalfOf
   ) external onlyRegisteredSpoke(spoke) returns (uint256, uint256) {
-    IERC20 asset = IERC20(_getReserveUnderlying(spoke, reserveId));
-    _spendWithdrawAllowance({
+    ISpoke.Reserve memory reserve = ISpoke(spoke).getReserve(reserveId);
+    uint256 currentAllowance = _checkWithdrawAllowance({
```
DhairyaSethi: it cannot be that this "check" will revert on happy path cases, right?

src/position-manager/TakerPositionManager.sol
```diff
@@ -269,60 +320,52 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+  }
+
+  /// @dev Deducts the corrected amount from the given allowance.
+  /// `amount` may exceed `currentAllowance` by a rounding delta;
+  /// consumption is capped at `currentAllowance` to prevent underflow.
+  function _deductAllowance(
+    uint256 currentAllowance,
+    uint256 amount
+  ) internal pure returns (uint256 newAllowance) {
+    return currentAllowance > amount ? currentAllowance - amount : 0;
```
DhairyaSethi:
```suggestion
    return currentAllowance.zeroFloorSub(amount);
```

src/position-manager/TakerPositionManager.sol
```diff
@@ -269,60 +320,52 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
-        reserveId: reserveId,
-        owner: owner,
-        spender: spender,
-        newCreditDelegation: currentAllowance.uncheckedSub(amount)
-      });
-    }
+  }
+
+  /// @dev Deducts the corrected amount from the given allowance.
+  /// `amount` may exceed `currentAllowance` by a rounding delta;
```
DhairyaSethi: i think we should call the `amount` param as `correctAmount` in this function args for clarity

src/position-manager/TakerPositionManager.sol
DhairyaSethi: also this function only abstracting this zeroFloorSub so probably do not need this function and inline it with the comment

src/position-manager/TakerPositionManager.sol
Kogaroshi: changed based on comment https://github.com/aave/aave-v4/pull/1251#discussion_r2905610478

tests/misc/allowance-sim/WithdrawAllowanceSim.t.sol
```diff
@@ -0,0 +1,87 @@
...
+    // Method A: direct conversion of withdrawnShares to assets (round up)
+    uint256 methodA = hub1.previewAddByShares(daiAssetId, withdrawnShares);
+
+    // Method B: before/after delta
+    uint256 assetsBefore = hub1.previewAddByShares(daiAssetId, sharesBefore);
+    uint256 assetsAfter = hub1.previewAddByShares(daiAssetId, sharesAfter);
+    uint256 methodB = assetsBefore - assetsAfter;
+
+    // Log all values
+    console.log('--- WithdrawAllowanceSim ---');
```
DhairyaSethi: we usually do not commit sims in test, so far we have written and checked these cases with z3 to symbolically prove them for all values instead of relying on fuzzing

i would strongly recommend the same here, let's keep this as multiple fuzz tests to cover and assert edge cases (this is missing from this PR so far)

tests/misc/allowance-sim/WithdrawAllowanceSim.t.sol
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    // Skip time to accrue interest so addedIndex != RAY
+    skip(timeSkip);
+
+    // Alice approves bob to withdraw on her behalf
+    vm.prank(alice);
+    positionManager.approveWithdraw(address(spoke1), reserveId, bob, withdrawAmount * 10);
+
+    // Snapshot before
```
DhairyaSethi: pls rm comments which make this seem like ai gen

tests/misc/allowance-sim/WithdrawAllowanceSim.t.sol
DhairyaSethi: also ideally no console logs in tests, this is ofc sim and prev point - we do not commit the sim work

src/position-manager/TakerPositionManager.sol
DhairyaSethi: can you add specific test for that pls and prove w z3

src/position-manager/TakerPositionManager.sol
DhairyaSethi: pls prove this w z3

src/position-manager/TakerPositionManager.sol
DhairyaSethi: i think this should be `correctedAmount` only to be more clear @miguelmtzinf

src/position-manager/TakerPositionManager.sol
DhairyaSethi: ok lets discuss there, anyway i think this function can be removed and can be inlined with the comment

src/position-manager/TakerPositionManager.sol
DhairyaSethi: we could then over consume allowance by a few wei, no? that's the problem this PR intends to solve also

for cases when there is exact allowance

src/position-manager/TakerPositionManager.sol
DhairyaSethi: we already cap tho, no?

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +167,47 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+    // positions. Due to rounding in supply-side SharesMath, the actual decrease in the user's
+    // position value can differ slightly from the input `amount`. To handle this, the allowance
+    // consumption is based on the before/after delta of `previewAddByShares`, and capped at
+    // `currentAllowance` to prevent underflow from rounding.
+    if (currentAllowance != type(uint256).max) {
+      uint256 sharesAfter = ISpokeBase(spoke).getUserSuppliedShares(reserveId, onBehalfOf);
+      uint256 newAllowance = _deductAllowance({
+        currentAllowance: currentAllowance,
+        amount: reserve.hub.previewAddByShares(reserve.assetId, sharesBefore) -
+          reserve.hub.previewAddByShares(reserve.assetId, sharesAfter)
```
DhairyaSethi: i think we cannot do this boss
the operation itself changes the share price (unlike v3) bc of donation and hence the true balance before would be doing the first preview here before the operation

src/position-manager/TakerPositionManager.sol
avniculae: How can we do exact allowance in asset terms anyways?

For example
- drawn index: 1.249
- user X drawn shares: 10
- user X wants to approve user Y to borrow 5 assets on their behalf

**option 1**
- user Y borrows 5 assets at drawn index 1.249
- user X drawn assets before: roundUp(10*1.249) = 13
- user X drawn shares: 15 
- user X drawn assets after: roundUp(15*1.249) = 19
- increase in user X drawn assets: 6

**option 2**
- user Y borrows 5 assets at drawn index 1.25
- user X drawn assets before: roundUp(10*1.25) = 13
- user X drawn shares: 14
- user X drawn assets after: roundUp(14*1.25) = 18
- increase in user X drawn assets: 5

**option 3**
- user Y borrows 5 assets at drawn index 1.3
- user X drawn assets before: roundUp(10*1.3) = 13
- user X drawn shares: 14
- user x drawn assets after: roundUp(14*1.3) = 19

src/position-manager/TakerPositionManager.sol
Kogaroshi: 34c0877409535ac5e487fa086dddcb62fb47cf4d

src/position-manager/TakerPositionManager.sol
Kogaroshi: 34c0877409535ac5e487fa086dddcb62fb47cf4d

src/position-manager/TakerPositionManager.sol
Kogaroshi: current version using asset amounts before/after :  34c0877409535ac5e487fa086dddcb62fb47cf4d

src/position-manager/TakerPositionManager.sol
Kogaroshi: 34c0877409535ac5e487fa086dddcb62fb47cf4d

src/position-manager/TakerPositionManager.sol
Kogaroshi: new logic using asset amounts before/after : 34c0877409535ac5e487fa086dddcb62fb47cf4d
will work on z3 scripts next.

src/position-manager/TakerPositionManager.sol
Kogaroshi: 34c0877409535ac5e487fa086dddcb62fb47cf4d (removed `correctedAmount` keyword)

src/position-manager/TakerPositionManager.sol
```diff
@@ -245,18 +245,18 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
-        currentAllowance: currentAllowance,
-        amount: reserve.hub.previewRestoreByShares(reserve.assetId, drawnSharesAfter) -
-          reserve.hub.previewRestoreByShares(reserve.assetId, drawnSharesBefore)
-      });
+      // We only need drawnAssets to calculate the corrected amount, so it's sufficient to check the position
+      // after the borrow action since drawn assets only increase, and the premium debt should remain unchanged.
+      uint256 drawnAssetsAfter = _getUserDrawnDebt(spoke, reserveId, onBehalfOf);
+      /// Deducts the corrected amount from the given allowance.
+      /// the correctedAmount (`drawnAssetsAfter` - `drawnAssetsBefore`) may exceed `currentAllowance` by a rounding delta;
+      /// consumption is capped at `currentAllowance` to prevent underflow.
```
DhairyaSethi: i mean at this point we can just use totalDebt and we dont the first have above comment 

uint256 debtAssets{Before,After} = ISpoke(spoke).getUserTotalDebt(user, reserveId)

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +169,45 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+    (uint256 withdrawnShares, uint256 withdrawnAmount) = ISpokeBase(spoke).withdraw({
+      reserveId: reserveId,
+      amount: amount,
+      onBehalfOf: onBehalfOf
+    });
+
+    // Simply decreasing the allowance by the input `amount` is not ideal for shares-based
+    // positions. Due to rounding in supply-side SharesMath, the actual decrease in the user's
+    // position value can differ slightly from the input `amount`. To handle this, the allowance
+    // consumption is based on the before/after delta of `previewAddByShares`, and capped at
```
DhairyaSethi:
```suggestion
    // consumption is based on the before/after delta of `previewRemoveByShares`, and capped at
```

or just say getUserSuppliedAmount (this uses the remove preview roundings rather than add)

src/position-manager/TakerPositionManager.sol
Kogaroshi: Can't we have a case where `_refreshAndValidateUserAccountData` updates the premiumShares, so that the premiumDebt calculated by getUserTotalDebt() might be slightly bigger than the before, which would false our calculation for allowance deduction ?

DhairyaSethi: nop we cannot, overall premium debt in refreshPremium in the same block always remains unchanged (see assertion in hub.refreshPremium)
premiumShares change is exactly offset in premiumOffsetRay

src/position-manager/TakerPositionManager.sol
Kogaroshi: d0ffeaab4e728b845b59ea143c281db5d87e711c

src/position-manager/TakerPositionManager.sol
Kogaroshi: d0ffeaab4e728b845b59ea143c281db5d87e711c

src/position-manager/TakerPositionManager.sol
```diff
@@ -85,7 +86,7 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
       reserveId: reserveId,
       owner: msg.sender,
       spender: spender,
-      newCreditDelegation: amount
+      newAllowance: amount
```
yan-man, line 88: wondering why renamed to borrowAllowance instead of credit delegation, just a separate desire?

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +169,45 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+      onBehalfOf: onBehalfOf
+    });
+
+    // Simply decreasing the allowance by the input `amount` is not ideal for shares-based
+    // positions. Due to rounding in supply-side SharesMath, the actual decrease in the user's
+    // position value can differ slightly from the input `amount`. To handle this, the allowance
+    // consumption is based on the before/after delta of `getUserSuppliedAssets`, and capped at
+    // `currentAllowance` to prevent underflow from rounding.
+    if (currentAllowance != type(uint256).max) {
+      uint256 suppliedAssetsAfter = ISpokeBase(spoke).getUserSuppliedAssets(reserveId, onBehalfOf);
```
yan-man: if != uint256 max we use ISpokeBase(spoke) a few times, could consider caching it

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +169,45 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+    }
+
+    (uint256 withdrawnShares, uint256 withdrawnAmount) = ISpokeBase(spoke).withdraw({
+      reserveId: reserveId,
+      amount: amount,
+      onBehalfOf: onBehalfOf
+    });
+
+    // Simply decreasing the allowance by the input `amount` is not ideal for shares-based
+    // positions. Due to rounding in supply-side SharesMath, the actual decrease in the user's
```
yan-man: mentioning SharesMath here points to virtual shares/assets, but the rounding is more general right, ie just from imprecision due to shares/assets rounding in hub?

src/position-manager/TakerPositionManager.sol
```diff
@@ -194,21 +219,45 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+      uint256 drawnAssetsAfter = ISpoke(spoke).getUserTotalDebt(reserveId, onBehalfOf);
+      /// Deducts the corrected amount from the given allowance.
+      /// the correctedAmount (`drawnAssetsAfter` - `drawnAssetsBefore`) may exceed `currentAllowance` by a rounding delta;
+      /// consumption is capped at `currentAllowance` to prevent underflow.
+      _updateBorrowAllowance({
+        spoke: spoke,
+        reserveId: reserveId,
+        owner: onBehalfOf,
+        spender: msg.sender,
+        newAllowance: currentAllowance.zeroFloorSub(drawnAssetsAfter - drawnAssetsBefore)
```
yan-man: given we're use "drawn" here I think the "borrowed" refs should be "drawn" too, or vice versa, for consistency

tests/misc/z3/commons.py
```diff
@@ -42,6 +42,46 @@ def rayMulDown(a, b):
...
 
+def rayDivUp(a, b):
+    return (a * RAY + b - 1) / b
+
+
+def rayDivDown(a, b):
+    return (a * RAY) / b
+
+
+def toDrawnSharesUp(assets, drawnIndex):
```
avniculae: these can be moved a bit down, with the other functions

tests/misc/z3/withdraw_allowance_methods.py
```diff
@@ -0,0 +1,53 @@
...
+s.add(0 <= totalAddedAssets, totalAddedAssets <= 10**30)
+s.add(0 <= totalAddedShares, totalAddedShares <= 10**30)
+s.add(0 <= userShares, userShares <= totalAddedShares)
+s.add(1 <= withdrawAmount, withdrawAmount <= 10**30)
+s.add(withdrawAmount <= totalAddedAssets)
+
+withdrawnShares = toAddedSharesUp(withdrawAmount, totalAddedAssets, totalAddedShares)
+s.add(withdrawnShares <= userShares)
+
+userSuppliedBefore = toAddedAssetsUp(userShares, totalAddedAssets, totalAddedShares)
```
avniculae: this should round down, right?

src/position-manager/TakerPositionManager.sol
Kogaroshi: it was already renamed in the main PR for those Position Managers, this one was forgotten.
As of why, to have unity in the naming in this posm, since credit delegation is in fact allowance, it also reduces potential confusion.

src/position-manager/TakerPositionManager.sol
Kogaroshi: in borrow() in Spoke, we have `drawnShares`, simplify following naming, but don't mind renaming here.
Would `borrowedAssets` make sense, or `totalDebtAssets` to match getter ?

tests/misc/z3/withdraw_allowance_methods.py
Kogaroshi: true, gonna change

src/position-manager/TakerPositionManager.sol
Kogaroshi: we could, it was left this way to match style from other Position Managers, but how should we name it ? `targetSpoke` ?

DhairyaSethi: no pls target* thing sucks, its fine to cast again. caching does not save anything

src/position-manager/TakerPositionManager.sol
DhairyaSethi: no this talks about total debt also not just base debt

tests/misc/z3/withdraw_allowance_methods.py
Kogaroshi: 8a85b28507e5f3dc57252f62f4a0be870101536e

tests/misc/z3/commons.py
Kogaroshi: 8a85b28507e5f3dc57252f62f4a0be870101536e

src/position-manager/TakerPositionManager.sol
Kogaroshi: 8a85b28507e5f3dc57252f62f4a0be870101536e

src/position-manager/TakerPositionManager.sol
Kogaroshi: e9f2bff43507842f17342588c467513d84291d50

src/position-manager/interfaces/ITakerPositionManager.sol
```diff
@@ -180,12 +180,12 @@ interface ITakerPositionManager is IPositionManagerBase {
...
   ) external view returns (uint256);
 
-  /// @notice Returns the credit delegation allowance for a spender on behalf of an owner.
+  /// @notice Returns the borrow allowance for a spender on behalf of an owner.
   /// @param spoke The address of the Spoke.
   /// @param reserveId The identifier of the reserve.
   /// @param owner The address of the owner.
   /// @param spender The address of the spender.
-  /// @return The amount of credit delegation allowance.
+  /// @return The amount of borrow allowance.
```
avniculae, line 188: let's also rename test_renounceCreditDelegation from PositionManager_Gas_Tests

avniculae: and test_creditDelegation, test_delegateCreditWithSig

avniculae: and test_hash_creditDelegation_fuzz from PositionManagerEIP712HashTest

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +168,45 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+
+    // Simply decreasing the allowance by the input `amount` is not ideal for shares-based
+    // positions. Due to rounding in Hub conversions, the actual decrease in the user's
+    // position value can differ slightly from the input `amount`. To handle this, the allowance
+    // consumption is based on the before/after delta of `getUserSuppliedAssets`, and capped at
+    // `currentAllowance` to prevent underflow from rounding.
+    if (currentAllowance != type(uint256).max) {
+      uint256 suppliedAssetsAfter = ISpoke(spoke).getUserSuppliedAssets(reserveId, onBehalfOf);
+      /// Deducts the corrected amount from the given allowance.
+      /// the correctedAmount (`suppliedAssetsBefore` - `suppliedAssetsAfter`) may exceed `currentAllowance` by a rounding delta;
```
avniculae: I would add that this is true because we checked at the beginning of the function that allowance is at least amount

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +168,45 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+      /// consumption is capped at `currentAllowance` to prevent underflow.
+      _updateWithdrawAllowance({
+        spoke: spoke,
+        reserveId: reserveId,
+        owner: onBehalfOf,
+        spender: msg.sender,
+        newAllowance: currentAllowance.zeroFloorSub(suppliedAssetsBefore - suppliedAssetsAfter)
+      });
```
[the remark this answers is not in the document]
avniculae: this is right, we should/usually use // inside functions I think

src/position-manager/TakerPositionManager.sol
```diff
@@ -194,21 +218,45 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+
+    // Simply decreasing the allowance by the input `amount` is not ideal for shares-based
+    // debt. Due to rounding, the actual increase in the user's debt can differ slightly from
+    // the input `amount`. To handle this, the allowance consumption is based on the before/after
+    // delta of `getUserTotalDebt`, and capped at `currentAllowance` to prevent underflow
+    // from rounding.
+    if (currentAllowance != type(uint256).max) {
+      uint256 borrowedAssetsAfter = ISpoke(spoke).getUserTotalDebt(reserveId, onBehalfOf);
+      /// Deducts the corrected amount from the given allowance.
+      /// the correctedAmount (`borrowedAssetsAfter` - `borrowedAssetsBefore`) may exceed `currentAllowance` by a rounding delta;
```
avniculae: same comment as for withdraw

tests/misc/z3/borrow_allowance_methods.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# Both methods should always be >= borrowAmount (safety).
+# We verify whether A == B always holds and bound any divergence.
+from commons import *
+
+s = Solver()
+s.set("timeout", 300000)  # 5min per check
+
+drawnIndex = Int("drawnIndex")
+userShares = Int("userShares")
```
avniculae: I know we are in borrow case here, but I would still call this `userDrawnShares`, wdyt?

tests/misc/z3/borrow_allowance_methods.py
```diff
@@ -0,0 +1,49 @@
...
+#   borrowedShares = rayDivUp(amount, drawnIndex)
+#   drawnIndex does NOT change during a single borrow.
+#
+# Method A: rayMulUp(borrowedShares, drawnIndex)
+#           — direct round-trip of borrowed shares back to assets
+# Method B: rayMulUp(userShares + borrowedShares, drawnIndex) - rayMulUp(userShares, drawnIndex)
+#           — before/after delta of user drawn debt
+#
+# Both methods should always be >= borrowAmount (safety).
+# We verify whether A == B always holds and bound any divergence.
```
avniculae, line 15: we usually state the conclusion here, such that we don't need to run this

tests/misc/z3/borrow_allowance_methods.py
```diff
@@ -0,0 +1,49 @@
...
+# Safety (both methods >= actual borrow amount)
+proveValid(s, "A >= borrowAmount", methodA >= borrowAmount)
+proveValid(s, "B >= borrowAmount", methodB >= borrowAmount)
+
+# Equality check
+proveValid(s, "A == B", methodA == methodB)
+
+# Divergence bounds (in case A != B)
+proveValid(s, "A >= B", methodA >= methodB)
+proveValid(s, "A - B <= 1", methodA - methodB <= 1)
```
avniculae, line 49: I see these 2 are true, which means A is never an underapproximation. Should we not use shares in the impl such that it's easier?

tests/misc/z3/withdraw_allowance_methods.py
```diff
@@ -0,0 +1,53 @@
...
+s = Solver()
+s.set("timeout", 300000)  # 5min per check
+
+totalAddedAssets = Int("totalAddedAssets")
+totalAddedShares = Int("totalAddedShares")
+userShares = Int("userShares")
+withdrawAmount = Int("withdrawAmount")
+
+s.add(0 <= totalAddedAssets, totalAddedAssets <= 10**30)
+s.add(0 <= totalAddedShares, totalAddedShares <= 10**30)
```
avniculae: this means supply share price can by anything (even smaller than 1). I would do something similar to liquidation_logic.py

```solidity
addedShares = Int("addedShares")
s.add(0 <= addedShares, addedShares <= MAX_SUPPLY_AMOUNT)
totalAddedAssets = Int("totalAddedAssets")
s.add(
    (addedShares + VIRTUAL_SHARES) <= (totalAddedAssets + VIRTUAL_ASSETS),
    (totalAddedAssets + VIRTUAL_ASSETS)
    <= MAX_SUPPLY_PRICE * (addedShares + VIRTUAL_SHARES),
)
```

tests/misc/z3/withdraw_allowance_methods.py
```diff
@@ -0,0 +1,53 @@
+# Verifies the outcome of each calculation method for the asset liquidity cost of a
+# withdraw action, to find which ones give a good representation of the reduction in
+# the user's supplied assets — suitable for decreasing the taker withdraw allowance.
+#
+# Method A: toAddedAssetsUp(withdrawnShares)               — ceil conversion of burned shares
+# Method B: toAddedAssetsUp(before) - toAddedAssetsUp(after) — delta of ceil asset views
```
avniculae:
```suggestion
# Method B: toAddedAssetsDown(before) - toAddedAssetsDown(after) — delta of floor asset views
```

tests/misc/z3/withdraw_allowance_methods.py
```diff
@@ -0,0 +1,53 @@
...
+s.add(0 <= userShares, userShares <= totalAddedShares)
+s.add(1 <= withdrawAmount, withdrawAmount <= 10**30)
+s.add(withdrawAmount <= totalAddedAssets)
+
+withdrawnShares = toAddedSharesUp(withdrawAmount, totalAddedAssets, totalAddedShares)
+s.add(withdrawnShares <= userShares)
+
+userSuppliedBefore = previewRemoveByShares(userShares, totalAddedAssets, totalAddedShares)
+
+roundTripShares = toAddedSharesDown(withdrawAmount, totalAddedAssets, totalAddedShares)
```
avniculae: why is this down? remove rounds up removed shares

avniculae: I think we should use withdrawnShares here right?

avniculae: I think method C is just method A, but it uses a different supply price for shares -> assets

Kogaroshi: they are similar, methodC is supposed to do `previewAddByAssets` => `previewAddByShares` before the action is executed to estimate the true liquidity change.
while method A does `previewAddByShares` on the output or the action, so in this case the share price will have the donation as well.

Kogaroshi: but method C is invalid, since it can produce an asset amount that is actually lower than `withdrawAmount`, so maybe my logic there was wrong

src/position-manager/TakerPositionManager.sol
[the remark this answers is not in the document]
DhairyaSethi: agree @Kogaroshi

src/position-manager/TakerPositionManager.sol
```diff
@@ -194,21 +218,45 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+      /// consumption is capped at `currentAllowance` to prevent underflow.
+      _updateBorrowAllowance({
+        spoke: spoke,
+        reserveId: reserveId,
+        owner: onBehalfOf,
+        spender: msg.sender,
+        newAllowance: currentAllowance.zeroFloorSub(borrowedAssetsAfter - borrowedAssetsBefore)
```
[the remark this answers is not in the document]
DhairyaSethi: @Kogaroshi can you close irrelevent comments also pls to reduce noise or just disable copilot lol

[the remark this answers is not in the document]
Kogaroshi: Ah, I keep hiding them, I thought it was for everybody ... And it is disabled on my account, no clue why I keep getting those

tests/misc/z3/borrow_allowance_methods.py
Kogaroshi: The issue with `A` is that it sometimes overshoot too much, and instead of having the actual user position change, we have slightly more, which would mean we would decrease the allowance by too much if we do not reach `currentAllowance`.

tests/misc/z3/withdraw_allowance_methods.py
Kogaroshi: b4304d49e53e4148f71a4735303453bda9177397

tests/misc/z3/withdraw_allowance_methods.py
Kogaroshi: b4304d49e53e4148f71a4735303453bda9177397

tests/misc/z3/borrow_allowance_methods.py
Kogaroshi: b4304d49e53e4148f71a4735303453bda9177397

src/position-manager/TakerPositionManager.sol
Kogaroshi: b4304d49e53e4148f71a4735303453bda9177397

src/position-manager/TakerPositionManager.sol
Kogaroshi: b4304d49e53e4148f71a4735303453bda9177397

src/position-manager/interfaces/ITakerPositionManager.sol
Kogaroshi: b4304d49e53e4148f71a4735303453bda9177397

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +168,44 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+      amount: amount,
+      onBehalfOf: onBehalfOf
+    });
+
+    if (currentAllowance != type(uint256).max) {
+      // Simply decreasing the allowance by the input `amount` is not ideal for shares-based
+      // positions. Due to rounding in Hub conversions, the actual decrease in the user's
+      // position value can differ slightly from the input `amount`.
+      // To handle this, the allowance consumption is based on a corrected amount, and capped at
+      // `currentAllowance` to prevent underflow from rounding, , given that `amount` was already checked against the allowance.
```
DhairyaSethi:
```suggestion
      // `currentAllowance` to prevent underflow from rounding, given that `amount` was already checked against the allowance.
```

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +168,44 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerBase {
...
+    uint256 suppliedAssetsBefore;
+    if (currentAllowance != type(uint256).max) {
+      suppliedAssetsBefore = ISpoke(spoke).getUserSuppliedAssets(reserveId, onBehalfOf);
+    }
+
+    (uint256 withdrawnShares, uint256 withdrawnAmount) = ISpoke(spoke).withdraw({
+      reserveId: reserveId,
+      amount: amount,
+      onBehalfOf: onBehalfOf
+    });
```
DhairyaSethi: named args avoidable here i think

Kogaroshi: 113fcb7596865fc296e3557fb563afc29f7582f5

src/position-manager/TakerPositionManager.sol
Kogaroshi: 113fcb7596865fc296e3557fb563afc29f7582f5

src/position-manager/TakerPositionManager.sol
```diff
@@ -3,8 +3,8 @@
 pragma solidity 0.8.28;
 
 import {SafeERC20, IERC20} from 'src/dependencies/openzeppelin/SafeERC20.sol';
-import {MathUtils} from 'src/libraries/math/MathUtils.sol';
 import {EIP712Hash} from 'src/position-manager/libraries/EIP712Hash.sol';
+import {MathUtils} from 'src/libraries/math/MathUtils.sol';
```
CheyenneAtapour, line 6: I think this was fine where it was

Kogaroshi: this is to match `using` order, but we can revert it back and change `using` order as well, no big pref.

CheyenneAtapour: I think we should keep the import order consistent with our style guide and also just change the using order

Kogaroshi: d051e5248509ef387351318a4f7b71ccf721a87d

tests/unit/position-manager/TakerPositionManager/TakerPositionManager.t.sol
```diff
@@ -487,22 +499,91 @@ contract TakerPositionManagerTest is TakerPositionManagerBaseTest {
...
+
+    Utils.supplyCollateral(spoke1, _daiReserveId(spoke1), alice, supplyAmount, alice);
+    Utils.supplyCollateral(spoke1, _daiReserveId(spoke1), bob, supplyAmount, bob);
+
+    Utils.borrow(spoke1, _daiReserveId(spoke1), bob, initialBorrowAmount, bob);
+
+    skip(322 days);
+
+    uint256 availableLiquidity = tokenList.dai.balanceOf(address(hub1));
+    uint256 maxBorrow = spoke1.getUserSuppliedAssets(_daiReserveId(spoke1), alice) / 10;
```
CheyenneAtapour, line 548: Could make this more than 10% of supplied, like even 50%, but not a big deal

src/position-manager/TakerPositionManager.sol
```diff
@@ -168,21 +168,44 @@ contract TakerPositionManager is ITakerPositionManager, PositionManagerIntentBas
     uint256 amount,
     address onBehalfOf
   ) external onlyRegisteredSpoke(spoke) returns (uint256, uint256) {
-    IERC20 asset = IERC20(_getReserveUnderlying(spoke, reserveId));
-    _spendWithdrawAllowance({
+    IERC20 underlying = IERC20(_getReserveUnderlying(spoke, reserveId));
+    uint256 currentAllowance = _checkWithdrawAllowance({
```
miguelmtzinf: not clear what "check" is. I think it's clearer doing `_getWithdrawAllowance` and have a require directly here

tests/misc/z3/withdraw_allowance_methods.py
```diff
@@ -0,0 +1,58 @@
...
+s.add(
+    (totalAddedShares + VIRTUAL_SHARES) <= (totalAddedAssets + VIRTUAL_ASSETS),
+    (totalAddedAssets + VIRTUAL_ASSETS)
+    <= MAX_SUPPLY_PRICE * (totalAddedShares + VIRTUAL_SHARES),
+)
+userSuppliedShares = Int("userSuppliedShares")
+withdrawAmount = Int("withdrawAmount")
+
+s.add(0 <= totalAddedAssets, totalAddedAssets <= 10**30)
+s.add(0 <= totalAddedShares, totalAddedShares <= 10**30)
```
miguelmtzinf: is this dup? we have it already in L19
Also, should we use `MAX_SUPPLY_AMOUNT` instead? same in the other z3 file

Kogaroshi: 83d490e10f0ced58918fa73b0e94b982e8378f70

src/position-manager/TakerPositionManager.sol
Kogaroshi: 83d490e10f0ced58918fa73b0e94b982e8378f70

src/position-manager/TakerPositionManager.sol
miguelmtzinf:
```suggestion
/// @notice Position manager to handle withdraw and borrow actions on behalf of users.
```

Kogaroshi: 472f0847f1d4ce498a7c85aed6564bc5adae4cf2

PR #1253, opened 2026-03-10
fix: Skipped accrue test

CheyenneAtapour:
Fixes the previously skipped `test_accrueInterest_fuzz_RPBorrowAndSkipTime` test, passing now with 100k fuzz runs


tests/Base.t.sol
```diff
@@ -1932,13 +1932,13 @@ abstract contract Base is Test {
...
       expectedPremiumDebt,
-      3,
+      4,
       string.concat('asset premium debt ', label)
     );
     assertApproxEqAbs(
       hub1.getAssetTotalOwed(assetId),
       expectedDrawnDebt + expectedPremiumDebt,
-      3,
+      4,
```
yan-man, line 1947: tbh even with the abs of 3, seems a bit much. made more sense with the old prem accounting, but feel like now it should be quite exact. Wonder why that's happening

yan-man (dismissed):

tests/Base.t.sol
CheyenneAtapour: It could be 0, but it's just that I'm calculating premium debt / rp in a different way than it happens on the contract, but I think it's fine

yan-man:
i think i was seeing a warning related to this test before, just wanted to make sure it's resolved here

PR #1256, opened 2026-03-10
feat: add supply skimmed on TreasurySpoke

CheyenneAtapour:
Adds a new method: `supplySkimmed(hub, underlying, amount)` to TreasurySpoke, allowing the owner to claim untracked liquidity already present on a Hub without pulling tokens from the owner (/caller).
Refactors supply and supplySkimmed into a shared _supply internal function with a skim flag to avoid code duplication

The Hub's add() function validates that actual token balance covers the claimed amount (balance >= liquidity + amount), so supplySkimmed is safe against over-claiming. Tokens must be transferred to the Hub before calling supplySkimmed, otherwise the transaction reverts with InsufficientTransferred.

This is useful for claiming tokens that were sent directly to the Hub (e.g. donations, fee sweeps, or batched transfers from a governance executor).

### DIFF
--- a/tests/unit/Spoke/TreasurySpoke.t.sol
+++ b/tests/unit/Spoke/TreasurySpoke.t.sol
@@ -93,6 +93,19 @@ contract TreasurySpokeTest is SpokeBase {
     treasurySpoke.supplySkimmed(address(hub1), address(tokenList.dai), 1);
   }
 
+  function test_supplySkimmed_revertsWith_InsufficientTransferred(uint256 amount) public {
+    amount = bound(amount, 1, MAX_SUPPLY_AMOUNT);
+    uint256 transferAmount = vm.randomUint(0, amount - 1);
+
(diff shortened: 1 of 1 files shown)


src/spoke/TreasurySpoke.sol
```diff
@@ -39,16 +43,16 @@ abstract contract TreasurySpoke is ITreasurySpoke, Ownable2StepUpgradeable {
...
-      amount,
-      targetHub.getSpokeAddedAssets(assetId, address(this))
-    );
-    uint256 withdrawnShares = targetHub.remove(assetId, withdrawnAmount, msg.sender);
+    return _withdraw(hub, underlying, amount, true);
+  }
 
-    return (withdrawnShares, withdrawnAmount);
+  /// @inheritdoc ITreasurySpoke
+  function withdrawSkimmed(
```
DhairyaSethi: there is no need for a withdraw skim version, it doesn't make sense

PR #1258, opened 2026-03-11
fix: default on ex rate for debt assets

yan-man:
- ensure that all preview getters in the hub default to a 1:1 exchange rate


miguelmtzinf:
The debt side only reverts if the asset is not listed (configured to 1 RAY upon listing). Therefore, adding custom logic to return 1 on the debt side would not really make sense, and preference is to revert on both sides if the asset does not exist. Behaviour is already documented, so closing this since implementing additional logic would introduce extra gas costs for little benefit.

PR #1260, opened 2026-03-12
feat: monotonic liquidations enhancement

yan-man:


tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenarios.t.sol
```diff
@@ -811,6 +811,76 @@ contract SpokeLiquidationCallScenariosTest is SpokeLiquidationCallBaseTest {
...
+        debtReserveId: _daiReserveId(spoke),
+        user: user,
+        debtToCover: type(uint256).max,
+        liquidator: liquidator,
+        isSolvent: true,
+        receiveShares: false
+      })
+    );
+
+    // All collateral liquidated, all debt liquidated
```
avniculae: would be nice to find an example where after recalculation debt amount decreases, such that we can properly tests this, wdyt?

yan-man: see scenario 9: https://github.com/aave/aave-v4/pull/1260/commits/1d79a73a1c8e526e68b7493abd0f9cd7969b561b

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenarios.t.sol
```diff
@@ -811,6 +811,72 @@ contract SpokeLiquidationCallScenariosTest is SpokeLiquidationCallBaseTest {
...
+    spoke.liquidationCall({
+      collateralReserveId: _usdxReserveId(spoke),
+      debtReserveId: _wethReserveId(spoke),
+      user: user,
+      debtToCover: 184,
+      receiveShares: false
+    });
+
+    uint256 liquidatorDebtAfter = tokenList.weth.balanceOf(liquidator);
+    // without recalculation, the debt to liquidate would be 183 (drawnShares to liquidate 122)
```
avniculae, line 873: this says 183, but comment above says 184

> // liquidate 184 debt assets, 184 / 1.5 -> 122 drawn shares to liquidate (rounded down)

I think it would be better if we can make comments and checks in share terms (drawnShares) , this would be more accurate I think

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenarios.t.sol
```diff
@@ -811,6 +811,72 @@ contract SpokeLiquidationCallScenariosTest is SpokeLiquidationCallBaseTest {
...
+    // increase debt price so that user is unhealthy
+    // now debt worth is 195 * $2 = $390
+    _mockReservePrice(spoke, _wethReserveId(spoke), 2e8 * 1e18 * 1e6);
+
+    // User is liquidatable
+    assertLe(spoke.getUserAccountData(user).healthFactor, 1e18, 'User should be unhealthy');
+    uint256 liquidatorDebtBefore = tokenList.weth.balanceOf(liquidator);
+
+    // liquidate 184 debt assets, 184 / 1.5 -> 122 drawn shares to liquidate (rounded down)
+    // 122 drawn shares to liquidate -> 183 debt assets, $366 collateral value
```
avniculae: this says 183, but below we say 182

> // now with recalculation, the debt to liquidate is 182 (drawnShares to liquidate 121)

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidationAmounts.t.sol
```diff
@@ -97,6 +97,40 @@ contract LiquidationLogicLiquidationAmountsTest is LiquidationLogicBaseTest {
...
+    vm.assume(rawLiquidationAmounts.collateralSharesToLiquidate > 0);
+
+    // Set suppliedShares to exactly the raw collateral
+    params.suppliedShares = rawLiquidationAmounts.collateralSharesToLiquidate;
+
+    // Recompute debt to ensure it is consistent with the actual collateral being taken.
+    LiquidationLogic.LiquidationAmounts
+      memory expectedLiquidationAmounts = _calculateAdjustedLiquidationAmounts(params);
+
+    params.debtToCover = bound(
```
avniculae, line 119: isn't this already bounded? don't we invalidate scenario here? can we check after liquidation that entire collateral was liquidated (by checking drawn shares is 0), just to be extra sure?

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenarios.t.sol
yan-man: > checks in share terms

reason why i didnt do that is, due to all coll being liquidated, it means deficit so the user position no longer has any drawnShares remaining so it is more roundabout to get the "true" shares diff as a result of the liquidation. 

> this says 183, but comment above says 184

comment discrepancy is bc: input is 184 (debtToCover), which is 122 drawnShares to liquidate, but drawnAmountToRestore & debtRayToLiquidate is 183. Added some comments to clarify https://github.com/aave/aave-v4/pull/1260/changes/7d4681b5cb1b8a943a870507b101eba6a4608000

tests/unit/Spoke/Liquidations/Spoke.LiquidationCall.Scenarios.t.sol
yan-man: i meant this part as before the recalculation, and below when it comes to 182 is after a recalculation that we've added here. added more commetns  https://github.com/aave/aave-v4/pull/1260/changes/7d4681b5cb1b8a943a870507b101eba6a4608000

tests/unit/libraries/LiquidationLogic/LiquidationLogic.LiquidationAmounts.t.sol
yan-man: need to bound this again in order to avoid the dust condition on the collateral, ie must be high enough so that it leads to 100% of coll being liquidated. 

> by checking drawn shares is 0

I think you mean added shares on coll side, added this https://github.com/aave/aave-v4/pull/1260/changes/7d4681b5cb1b8a943a870507b101eba6a4608000

yan-man:
decided we will not proceed with this addition

PR #1261, opened 2026-03-12
fix: abstract assetId, reserveId

DhairyaSethi:


src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -31,48 +32,55 @@ interface IAaveV4ConfigEngine {
     address feeReceiver;
     uint256 liquidityFee;
     address irStrategy;
-    bytes irData;
```
Kogaroshi, line 34: let's add here a struct for `Tokenization` to deploy the TokenizationSpoke when adding the asset ?

Kogaroshi: decided to do this in subsequent PR

PR #1267, opened 2026-03-13
fix: validation on delegatee zero addr

yan-man:


src/position-manager/ConfigPositionManager.sol
```diff
@@ -425,6 +425,7 @@ contract ConfigPositionManager is IConfigPositionManager, PositionManagerIntentB
     ConfigPermissions oldPermissions,
     ConfigPermissions newPermissions
   ) internal {
+    require(delegatee != address(0), InvalidAddress());
```
DhairyaSethi, line 428: i think avoidable but should add this natspec as well

PR #1270, opened 2026-03-17
feat: separate deploy periphery for posms

yan-man:


yan-man:
decided today that we will leave the deployment script intact and not separate out the posms. Instead, the control for deploying the posms will depend on the bool inputs

PR #1273, opened 2026-03-17
style: Add ascii art

CheyenneAtapour:


src/hub/Hub.sol
```diff
@@ -15,6 +15,35 @@ import {Premium} from 'src/hub/libraries/Premium.sol';
 import {IBasicInterestRateStrategy} from 'src/hub/interfaces/IBasicInterestRateStrategy.sol';
 import {IHubBase, IHub} from 'src/hub/interfaces/IHub.sol';
 
+//                                     @%%%%%%%%%%%%%%%%%%%%%%%%%%@
```
yan-man, line 18: tbh i think too big, personally was thinking 20 lines ish (in height)

src/spoke/Spoke.sol
```diff
@@ -27,6 +27,35 @@ import {IHubBase} from 'src/hub/interfaces/IHubBase.sol';
 import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
 import {SpokeStorage} from 'src/spoke/SpokeStorage.sol';
 
+//                                     @%%%%%%%%%%%%%%%%%%%%%%%%%%@
```
DhairyaSethi: should be on the SpokeInstance instead imo

DhairyaSethi: can we make it say v4 probably

src/hub/Hub.sol
CheyenneAtapour: Uniswap's is also 40 lines

src/spoke/Spoke.sol
CheyenneAtapour: Moved. But yeah maybe we can regenerate more art

src/hub/Hub.sol
miguelmtzinf: wow, this is way bigger than expected. I was thinking smt like 10 lines max tbh

yan-man:
is it too late or we can still add this

src/hub/Hub.sol
DhairyaSethi: you don't get anything in 10-20 range tbh

DhairyaSethi: to me size is fine but these need to be on the instance contracts (which are the entry points of the impl as mentioned before)

src/spoke/instances/SpokeInstance.sol
```diff
@@ -4,6 +4,35 @@ pragma solidity 0.8.28;
 
 import {Spoke} from 'src/spoke/Spoke.sol';
 
+//                                     @%%%%%%%%%%%%%%%%%%%%%%%%%%@
```
DhairyaSethi, line 7:
```suggestion
//                                     @4444444444444444444444444444444444444@
```

PR #1278, opened 2026-03-20
test : check used LiqLogic bytecode

Kogaroshi:


Kogaroshi:
This was used as a check during deployments, closing since not needed anymore, we are using the default compiler `LiquidationLogic`

PR #1285, opened 2026-03-26
docs: Update overview.md

CanonicalJP:
Makes small changes to `./docs/overview.md`


docs/overview.md
yan-man, line 52: comparing vs the WP, maybe we should add asset registration/listing here too

docs/overview.md
```diff
@@ -44,7 +44,7 @@ A Hub can have an unspecified number of Spokes, each one contributing to the tot
 
 ## Hub
 
-The Hub is immutable and serves as the central coordinator for liquidity management in Aave V4. The design allows for multiple Hubs to exist, with each Hub maintaining oversight of its own set of Spokes. Each Hub sets the add/draw caps for its Spokes and enforces crucial accounting invariants. The design objective was to make the Hub as simple as possible.
+The Hub is upgradeable and serves as the central coordinator for liquidity management in Aave V4. The design allows for multiple Hubs to exist, with each Hub maintaining oversight of its own set of Spokes. Each Hub sets the add/draw caps for its Spokes and enforces crucial accounting invariants. The design objective was to make the Hub as simple as possible.
```
yan-man: "The design objective was to simplify the Hub as much as possible." -> this last sentence I dont think flows logically from the stuff preceding it, as in the things mentioned dont necessarily explain how it simplifies the Hub

docs/overview.md
```diff
@@ -202,16 +202,16 @@ The Reinvestment Module offers an optional tool to support capital efficiency wh
...
-- **Opt‑In for Users**: Participation in reinvestment is not compulsory. Reinvestment is enabled at the Hub level on a per-asset basis; users choose whether to interact with a reinvestment-enabled asset by supplying to a spoke Reserve that is connected to that Hub asset. Users who prefer not to participate can supply to spoke Reserves linked to Hub assets without an active reinvestment controller.
-- **Enhanced Yields**: By deploying otherwise idle liquidity into external strategies, opt‑in liquidity providers can earn incremental returns on top of borrower interest.
+- **Opt‑In for Users**: Participation in reinvestment is not compulsory. Reinvestment is enabled at the Hub level on a per-asset basis; users choose whether to interact with a reinvestment-enabled asset by supplying to a spoke Reserve that is connected to that Hub asset. Since the reinvestment controller is a governance-controlled parameter that can be activated at any time. It is the Governor's responsibility to properly communicate any planned activation and its timeframe in advance, giving existing suppliers the opportunity to withdraw their funds if they wish to opt out.
```
yan-man:
```suggestion
- **Opt‑In for Users**: Participation in reinvestment is not compulsory. Reinvestment is enabled at the Hub level on a per-asset basis; users choose whether to interact with a reinvestment-enabled asset by supplying to a spoke Reserve that is connected to that Hub asset. Since the reinvestment controller is a governance-controlled parameter that can be activated at any time, it is the Governor's responsibility to properly communicate any planned activation and its timeframe in advance. This gives existing suppliers the opportunity to withdraw their funds if they wish to opt out.
```

yan-man: tbh the title here is opt-in but this feels more like opt-out, esp the way this is written and the feature impl bc like mentioned it can be changed any time for a given asset connected to your existing position, and isnt enforced that the governor must give notice

docs/overview.md
```diff
@@ -202,16 +202,16 @@ The Reinvestment Module offers an optional tool to support capital efficiency wh
...
+- **New Yield Opportunities**: By deploying otherwise idle liquidity into external strategies, opt‑in liquidity providers can potentially earn incremental returns on top of borrower interest.
```
yan-man:
```suggestion
- **New Yield Opportunities**: By deploying otherwise idle liquidity into external strategies, liquidity providers opting in can earn additional returns on top of borrower interest.
```

returns are always >= 0 right, bc losses are absorbed by treasury? potentially just sounds like it could be a loss

docs/overview.md
```diff
@@ -202,16 +202,16 @@ The Reinvestment Module offers an optional tool to support capital efficiency wh
...
 - **Optional by Design**: The module can remain disabled without affecting core supply and borrow functionality. When inactive, all liquidity stays in the Hub.
 
 The reinvestment infrastructure enables the protocol to optimize capital efficiency while maintaining user choice and clear risk allocation. Implementation details, strategy selection, and risk management frameworks are governance decisions that can evolve based on protocol needs and market conditions.
 
 # Dynamic Risk Configuration
 
-One of the major risk‑side limitations of V3 lies in its single, global risk configuration per asset. This design creates significant governance overhead and potential user harm through unexpected liquidations, as any parameter change, in particular lowering the liquidation threshold, immediately affects every open position.
+One of the major risk‑side limitations of Aave V3 lies in its single, global risk configuration per asset. This design creates significant governance overhead and potential user harm through unexpected liquidations, as any parameter change, in particular lowering the liquidation threshold, immediately affects every open position.
```
yan-man:
```suggestion
One of the major risk‑side limitations of Aave V3 lies in its single, global risk configuration per asset. This design creates significant governance overhead and potential user friction through unexpected liquidations, as any parameter change, in particular lowering the liquidation threshold, immediately affects every open position.
```

harm just sounds too negative imo

docs/overview.md
```diff
@@ -250,18 +250,18 @@ The architecture of dynamic configuration comes with several practical constrain
...
-- **Dutch‑Auction Style Liquidation Bonus:** V3 applies a static liquidation bonus that does not depend on the borrower’s health factor. Aave V4 introduces a variable liquidation bonus that increases linearly as the health factor decreases. Governance can specify two spoke‑wide parameters that shape the liquidation bonus: `healthFactorForMaxBonus` and `liquidationBonusFactor`.
+- **Target Health Factor vs Close Factor:** In Aave V3, the default close factor is 50% (with a 100% close factor when HF < 0.95 or when liquidation amounts are under a given base currency threshold). Liquidators would typically repay half of a borrower’s debt and seize half of their collateral. Aave V4 removes the default close‑factor: the maximum a liquidator can repay (in situations that do not result in dust collateral or debt remaining) is the amount needed to bring the borrower back to the `TargetHealthFactor` determined by the Governor.
```
yan-man:
```suggestion
- **Target Health Factor vs Close Factor:** In Aave V3, the default close factor is 50% (with a 100% close factor when HF < 0.95 or when liquidation amounts are under a given base currency threshold). Liquidators would typically repay half of a borrower’s debt and seize a corresponding amount of their collateral. Aave V4 removes the default close factor: the maximum a liquidator can repay (in situations that do not result in dust collateral or debt remaining) is the amount needed to bring the borrower back to the `TargetHealthFactor` determined by the Governor.
```

half of user debt != half of user's coll necessarily. Also need to account for LB on top

docs/overview.md
```diff
@@ -250,18 +250,18 @@ The architecture of dynamic configuration comes with several practical constrain
...
+- **Dynamic Dust Handling during Liquidations**: Both Aave V3 and Aave V4 revert when the remaining amount is below a hard‑coded threshold, while dynamically adjusting the maximum debt that can be liquidated and, if the liquidator opts to fully repay, allow full repayment to prevent dust. However, Aave V4 allows more flexibility because of removing the close-factor and facilitating the liquidation steps required to bring the position back to the target HF. Dust may still remain if either the collateral or debt Reserve is fully liquidated.
```
yan-man:
```suggestion
- **Dynamic Dust Handling during Liquidations**: Both Aave V3 and Aave V4 revert when the remaining amount is below a hard‑coded threshold, while dynamically adjusting the maximum debt that can be liquidated and, if the liquidator intends to fully repay, allow full repayment to prevent dust. However, whereas Aave V3's close factor is a hard constraint, Aave V4 allows more flexibility by selectively bypassing the target HF enforcement in order to prevent dust accumulation. Dust can still only remain if either the collateral or debt Reserve is fully liquidated.
```

yan-man: @avniculae can you also give feedback here

avniculae: to me this paragraph is very hard to read and understand. what if we use a similar paragraph as the whitepaper, plus one of your suggestions

Similar to V3, Aave V4 uses a dust threshold of $1,000 (encoded as $1,000 \times 10^{26}$ in internal value units) to prevent residual positions that are uneconomical to liquidate. If the remaining debt or collateral post-liquidation falls below this threshold, the protocol expands the liquidation to fully exhaust the smaller side, or reverts if the liquidator's intended `debtToCover` is insufficient. However, whereas Aave V3's close factor is a hard constraint, Aave V4 allows more flexibility by selectively bypassing the target HF enforcement in order to prevent dust accumulation.

yan-man: sg, only 2 suggestions on top:
- Similar to V3 -> Similar to Aave V3
- (encoded as..) -> remove this part

docs/overview.md
```diff
@@ -250,18 +250,18 @@ The architecture of dynamic configuration comes with several practical constrain
...
+- **Dutch‑Auction Style Liquidation Bonus:** Aave V3 applies a static liquidation bonus that does not depend on the borrower’s health factor. Aave V4 introduces a variable liquidation bonus that increases linearly as the health factor decreases. Governance can specify two spoke‑wide parameters that shape the liquidation bonus: `healthFactorForMaxBonus` and `liquidationBonusFactor`.
```
yan-man:
```suggestion
- **Dutch‑Auction Style Liquidation Bonus:** Aave V3 applies a static liquidation bonus that does not depend on the borrower’s health factor. Aave V4 introduces a variable liquidation bonus that increases linearly as the health factor decreases. Governance can specify two spoke‑wide parameters that determine the liquidation bonus calculation: `healthFactorForMaxBonus` and `liquidationBonusFactor`.
```

docs/overview.md
CheyenneAtapour: This is strange to say when "liquidity is restored on the hub ..." the points about base interest and premium debt. Shouldn't that say "borrowed" or "drawn" instead of "restored"?

CheyenneAtapour: When does `borrow` not increase debt? I don't think we need to specify there

CheyenneAtapour:
```suggestion
Premium Debt is the portion of a user’s debt that represents the additional interest accumulated due to the quality of the user’s collateral assets (i.e., their Risk Premium on top of the base drawn rate).
```

docs/overview.md
CheyenneAtapour: I agree with your suggestion, in response to your second comment:

I think it is written correctly: if it's already enabled on an asset, then it's opt in. If there's advance notice that it will be enabled, then you effectively opt out. Imo it's fine as written. The idea is the user is informed about it and can make a decision to participate or not

docs/overview.md
CheyenneAtapour:
```suggestion
Each Reserve stores the latest `dynamicConfigKey`, which represents the current up-to-date risk configuration. In contrast, every user position retains a snapshot of the active `dynamicConfigKey` corresponding to the configuration in effect at the time of its last risk-increasing event. This snapshot is refreshed across all assets of a user position only when the user performs an action which elevates the risk posed to the system, such as disabling an asset as collateral, withdrawing collateral, or borrowing. When a user designates a new asset as collateral, only the `dynamicConfigKey` snapshot of the asset in play is refreshed.
```

CheyenneAtapour: Please clarify that the user position has to be healthy after the action, with the latest dynamic configuration. This kind of makes it sound like the position before the action is checked to work with the dynamic config, then the action proceeds. But the check happens at the end, after the action, with the new dynamic config

CheyenneAtapour:
```suggestion
   3. `withdraw` when the withdrawn Reserve had been set as collateral
```

CheyenneAtapour: Can add withdraw when the Reserve has not been set as collateral

CheyenneAtapour:
```suggestion
3. The snapshot does **not** update on actions that do not increase the risk exposure of the system:
```

docs/overview.md
CheyenneAtapour: I agree it doesn't flow logically, but I think it's fine. Just meant to be an overview of the hub design, and I think it's important to mention that this was the design objective

docs/overview.md
CheyenneAtapour: Yeah, can perhaps be `registry of authorized spokes and assets`

docs/overview.md
CheyenneAtapour: I think this is fine as written, because 0 means you don't earn

docs/overview.md
CheyenneAtapour: unexpected liquidations are indeed harmful imo and not friction, and we are emphasizing the advantages of V4 here, so I think it's fine as written

docs/overview.md
CheyenneAtapour: I'd be fine with "shape" (more flavor text), but ok

docs/overview.md
CheyenneAtapour, line 142: I think we should make a note at the bottom here saying that if the user's collateral is not sufficient to cover all the debt (such as in the deficit case), the algorithm terminates early and just returns the calculation with all of the user's collateral value, and the remaining debt is ignored

docs/overview.md
CanonicalJP: made a small tweak to improve the flow and keep the concept

docs/overview.md
CanonicalJP: I'd prefer to keep the title since the rule is that users will have clear visibility of reinvestment-enabled reservers. The Governor activating reinvestment in reservers is not expected to be a regular situation

docs/overview.md
CanonicalJP: I'll change the first part but keep "potentially." Otherwise, it looks like additional earnings are guaranteed. In the next bullet point, it clarifies that any losses are absorbed by the Governor, removing doubts about this situation

docs/overview.md
CanonicalJP: Not a strong opinion here, open to both alternatives

docs/overview.md
```diff
@@ -86,7 +87,7 @@ The debt interest of each user is directly impacted by the quality of the assets
 
 ## Collateral Risk
 
-The Collateral Risk $CR_i$ is specified by the quality of the asset $i$, which is a BPS value, ranging from 0 to 1000_00. A value of 0 means highest quality and risk-free, while a value of 1000_00 signifies the lowest quality and maximum risk possible for a collateral.
+The Collateral Risk $CR_i$ is specified by the quality of the asset $i$, which is a BPS value, ranging from 0 to `1000_00`. A value of 0 means highest quality and risk-free, while a value of `1000_00` signifies the lowest quality and maximum risk possible for a collateral.
```
CheyenneAtapour: Looks strange to me if we do not also put the back ticks on the 0:

```suggestion
The Collateral Risk $CR_i$ is specified by the quality of the asset $i$, which is a BPS value, ranging from `0` to `1000_00`. A value of `0` means highest quality and risk-free, while a value of `1000_00` signifies the lowest quality and maximum risk possible for a collateral.
```

docs/overview.md
```diff
@@ -138,6 +139,8 @@ $RP_u = f(CR_0, C_{u,0}, P_0) = CR_0$
 
 $RP_u = f(CR_i, C_{u, i}, P_i) = \frac{CR_0C_{u,0}P_0 + CR_1C_{u,1}P_1}{C_{u,0}P_0+C_{u,1}P_1}$
 
+> **Note:** If the user's total collateral value is insufficient to cover all outstanding debt (e.g., in a deficit scenario), the algorithm terminates early once all collateral has been allocated. The Risk Premium is computed using only the collateral value available, and the remaining uncovered debt is ignored in the calculation.
```
CheyenneAtapour: Nit, but just a suggestion to consider
```suggestion
> **Note:** If the user's total collateral value is insufficient to cover all outstanding debt (e.g., in a deficit scenario), the algorithm terminates early once all collateral has been exhausted. The Risk Premium is computed using only the collateral value available, and the remaining uncovered debt is ignored in the calculation.
```

docs/overview.md
```diff
@@ -2,7 +2,7 @@
...
 
-The protocol implements sophisticated risk management through its Risk Premiums system, where each asset receives a specific risk factor (called Collateral Risk) ranging from 0 to 1000_00 (BPS) based on the asset's implied volatility, market conditions, liquidity, risk, etc. Base drawn rates are determined purely by utilisation, while risk premium rates are adjusted according to collateral risk profiles, from low‑risk pristine‑quality assets (such as ETH) to higher‑risk collateral, scaling borrowing costs proportionally.
+The protocol implements sophisticated risk management through its Risk Premiums system, where each asset receives a specific risk factor (called Collateral Risk) ranging from 0 to `1000_00` (BPS) based on the asset's implied volatility, market conditions, liquidity, risk, etc. Base drawn rates are determined purely by utilisation, while risk premium rates are adjusted according to collateral risk profiles, from low‑risk pristine‑quality assets (such as ETH) to higher‑risk collateral, scaling borrowing costs proportionally.
```
CheyenneAtapour: I think it reads better this way, and also use back ticks for all numbers
```suggestion
The protocol implements sophisticated risk management through its Risk Premium system, where each asset receives a specific risk factor (called Collateral Risk) ranging from `0` to `1000_00` (BPS) based on the asset's implied volatility, market conditions, liquidity, risk, etc. Base drawn rates are determined purely by utilisation, while risk premium rates are adjusted according to collateral risk profiles, from low‑risk pristine‑quality assets (such as ETH) to higher‑risk collateral, scaling borrowing costs proportionally.
```

docs/overview.md
```diff
@@ -241,27 +244,28 @@ The architecture of dynamic configuration comes with several practical constrain
...
+- **Target Health Factor vs Close Factor:** In Aave V3, the default close factor is 50% (with a 100% close factor when HF < 0.95 or when liquidation amounts are under a given base currency threshold). Liquidators would typically repay half of a borrower’s debt and seize a corresponding amount of their collateral. Aave V4 removes the default close factor: the maximum a liquidator can repay (in situations that do not result in dust collateral or debt remaining) is the amount needed to bring the borrower back to the `TargetHealthFactor` determined by the Governor.
+- **Dynamic Dust Handling during Liquidations**: Similar to Aave V3, Aave V4 uses a dust threshold of `$1_000` to prevent residual positions that are uneconomical to liquidate. If the remaining debt or collateral post-liquidation falls below this threshold, the protocol expands the liquidation to fully exhaust the smaller side, or reverts if the liquidator's intended `debtToCover` is insufficient. However, whereas Aave V3's close factor is a hard constraint, Aave V4 allows more flexibility by selectively bypassing the target HF enforcement in order to prevent dust accumulation.
```
yan-man:
```suggestion
- **Dynamic Dust Handling during Liquidations**: Similar to Aave V3, Aave V4 uses a dust threshold of `$1_000` to prevent residual positions that are uneconomical to liquidate. If the remaining debt or collateral post-liquidation were to fall below this threshold, the protocol expands the liquidation to fully exhaust the smaller side, or reverts if the liquidator's intended `debtToCover` is insufficient. However, whereas Aave V3's close factor is a hard constraint, Aave V4 allows more flexibility by selectively bypassing the target HF enforcement in order to prevent dust accumulation.
```

docs/overview.md
```diff
@@ -284,12 +288,12 @@ The following high‑level steps outline the Aave V4 liquidation flow:
...
+5. **Apply Debt Repayment & Transfer Collateral**: Reduce the borrower’s debt amount by the repaid amount. Transfer the corresponding collateral to the liquidator with the liquidation bonus applied, minus the protocol fee (as in Aave V3). If `receiveSharesEnabled` is true for the collateral Reserve and the Reserve is not frozen, the liquidator can opt to receive added shares directly instead of underlying assets by setting the `receiveShares` parameter to true. Added shares accrue yield in the Hub, providing a more capital-efficient liquidation mechanism. The fee portion is sent to the protocol/fee receiver via the Hub as shares, accrues yield there, and the shares are assigned directly via Hub accounting.
 6. **Emit Events and Update State:** A `LiquidationCall` event is emitted containing details of the liquidation. The borrower’s and Reserve’s interest indices are updated. If the borrower still has debt outstanding and no remaining collateral, the system will record a protocol deficit. Reporting deficit is allowed even when the reporting Spoke is halted (as long as it’s active).
```
CheyenneAtapour: There is no borrower (user) interest index, can remove that part. Or more accurate to just say 
```suggestion
6. **Emit Events and Update State:** A `LiquidationCall` event is emitted containing details of the liquidation. The borrower’s and Reserve’s debt amounts are updated. If the borrower still has debt outstanding and no remaining collateral, the system will record a protocol deficit. Reporting deficit is allowed even when the reporting Spoke is halted (as long as it’s active).
```

docs/overview.md
CheyenneAtapour: Configured and constant are a bit strange to use in the same sentence, might mistake for "configurable", but it's not. This should say just: 
```suggestion
- $HF\_LIQ\_THRESHOLD$: a spoke-wide constant. Represents the health factor threshold under which the user becomes liquidatable. Equals 1.
```

docs/overview.md
yan-man: was only thinking to soften language given some prior complaints about us bashing v3 but perhaps that isnt as relevant now, so no strong opinion, your call @CanonicalJP

CanonicalJP: I see your point now, moving forward with it

docs/overview.md
```diff
@@ -58,11 +59,10 @@ The key aspects of the Hub include:
...
 - Setting the liquidity fee per asset, determining the share of interest revenue retained by the protocol.
-- Enforcing accounting invariants:
-  1. Total borrowed shares == sum of Spoke debt shares
-  2. Hub added assets amount >= sum of Spoke added assets amount (converted from shares)
-  3. Hub added shares == sum of Spoke added shares
-  4. Supply share price and drawn index cannot decrease (remains constant or increases)
+- Enforcing accounting and solvency invariants:
+  1. Hub level aggregate fields == sum across all per-Spoke fields (`addedShares`, `drawnShares`, `premiumShares`, `premiumOffsetRay`, `deficitRay`)
+  2. Supply share price and drawn index cannot decrease (remains constant or increases)
+  3. Hub's actual token balance for a given asset is always >= internally tracked available liquidity (`asset.liquidity`)
```
yan-man, line 65: pls review updated hub invariants

docs/overview.md
```diff
@@ -79,14 +79,17 @@ Users interact with the Spokes, which then interact directly with the Hubs. The
 - Employing reentrancy guards for extra protection against reentrancy attacks, even though the Hub, Interest Rate Strategy, and Price Feeds are trusted and Aave V4 does not support tokens with callbacks.
 - Enforcing position constraints through a configurable `MAX_USER_RESERVES_LIMIT` which limits the number of collateral Reserves and the number of borrow Reserves a user can have (each counted separately).
 - Configuring and enforcing per-reserve liquidation parameters at the Spoke level.
+- Enforcing position safety invariants:
+  1. Users without collateral cannot assume debt
+  2. No user action can worsen their position's health factor below the liquidation threshold
```
yan-man, line 84: pls review updated spoke invariants

PR #1308, opened 2026-05-29
fix : remove compiler restriction for tests & use Foundry stable for CI

Kogaroshi:


foundry.toml
```diff
@@ -26,9 +26,9 @@ additional_compiler_profiles = [
 ]
 
 compilation_restrictions = [
-  { paths = "src/hub/instances/HubInstance.sol", optimizer = true, via_ir = true, optimizer_runs = 22_300 },
-  { paths = "src/spoke/instances/SpokeInstance.sol", optimizer = true, via_ir = true, optimizer_runs = 750 },
-  { paths = "tests/**", optimizer = true, via_ir = false, optimizer_runs = 44444444 },
+  { paths = "src/hub/instances/HubInstance.sol", via_ir = true, optimizer_runs = 22_300 },
+  { paths = "src/spoke/instances/SpokeInstance.sol", via_ir = true, optimizer_runs = 750 },
+  { paths = "tests/**", via_ir = false, optimizer_runs = 44444444 },
```
DhairyaSethi: tests compiler profile is different from before now no? the first two should be the same

Kogaroshi: tests is the same, just dropped `optimizer = true,` from all restrictions since Foundry does not support it anymore

DhairyaSethi (approved):

PR #1309, opened 2026-06-05
feat : positionIds with salt for single address

Kogaroshi:
Replaces single position per address to a combination of user address & `salt` to derive `positionIds`, allowing the same user address to have multiple isolated positions.
Current positions are kept by respecting previous mapping keys when the default `0x0` salt is given.

`processUserAccountData()` is now an external view functions, where the refreshing of dynamic configs is done outside this method, and moved to an external library to reduce `Spoke.sol` codesize.
The `tests/fork` test suite is temporary to ensure compatibility and pre-upgrade position preservation, and meant to be replaced once the full V4 test suite for fork an integration tests are imported in this repo.


src/spoke/Spoke.sol
```diff
@@ -929,19 +1194,17 @@ abstract contract Spoke is
...
   }
 
+  function _getPositionIdentifier(address user, bytes32 salt) internal pure returns (bytes32) {
+    // The default salt maps to the legacy address-keyed storage slot, so positions
+    // created before the position-nonces upgrade remain accessible. The chance of a non-default,
+    // keccak-derived identifier colliding with such a value is negligible (~2^-96).
+    return
+      salt == USER_POSITION_DEFAULT_SALT
+        ? bytes32(uint256(uint160(user)))
+        : keccak256(abi.encodePacked(user, salt));
```
```suggestion
        : Hashes.efficientKeccak256(user, salt);
```

src/spoke/Spoke.sol
```diff
@@ -929,19 +1194,17 @@ abstract contract Spoke is
...
-    function(address, bool) internal returns (UserAccountData memory) fnIn
-  )
-    internal
-    pure
-    returns (function(address, bool) internal view returns (UserAccountData memory) fnOut)
-  {
-    assembly ('memory-safe') {
-      fnOut := fnIn
-    }
+    return ('Spoke', '2');
```
DhairyaSethi: dont think this should change

PR #1310, opened 2026-06-11
feat: Add V4 Addresses Provider

avniculae:


tests/contracts/addresses-provider/V4AddressesProvider.t.sol
```diff
@@ -0,0 +1,626 @@
...
+      tag: 'PERIPHERY',
+      newAddress: makeAddr('NEW_CONFIG_ENGINE')
+    });
+
+    vm.expectRevert(abi.encodeWithSelector(IV4AddressesProvider.AddressAlreadySet.selector, id));
+    provider.setAddress({name: 'CONFIG_ENGINE', tag: 'PERIPHERY', newAddress: configEngine});
+    vm.stopPrank();
+  }
+
+  function test_setAddress_idCollision_revertsWith_AddressAlreadySet() public {
```
avniculae: Added this test to highlight that id collision is possible, but it will revert if already set (that's why I decided to not allow updating an existing id, but rather owner needs to remove --by setting addr(0) and add). Owner could use the general `setAddress` fn to set "MAIN_TOKENIZATION" as name, and `SPOKE` as tag, which would result in a different tag tracking for that address, but to me this is acceptable since it's owner-controlled fn. Mistakes should be prevented by the fact that we revert on collision, hence I decided to concatenate name and tag in the same string, rather than using abi.encode() to distinguish between these collisions. thoughts/

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
```diff
@@ -0,0 +1,626 @@
...
+
+    vm.startPrank(OWNER);
+    provider.setAddress({name: 'A', tag: 'B_C', newAddress: makeAddr('A')});
+
+    vm.expectRevert(abi.encodeWithSelector(IV4AddressesProvider.AddressAlreadySet.selector, id));
+    provider.setAddress({name: 'A_B', tag: 'C', newAddress: makeAddr('A_B')});
+    vm.stopPrank();
+  }
+
+  function test_setAddress_sameAddressUnderMultipleIds() public {
```
avniculae: this is currently possible, in case owner wants to track same address under multiple ids, eg babylon spoke can be in babylon tag and in third party tag. thoughts?

xnvx123 (commented):
a57e24a

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
Kogaroshi: Agree with the chosen method to prevent overriding address on same id.
I don't think in normal usage we'll have too much couple `tag` + `name` that would create issues there, so this logic should be sufficient, and then up to Owner to respect a coherent tag and naming convention.

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
Kogaroshi: yes, this was my 1st thought here, how much would it be an issue if the same address leaves under multiple tags/names, but since there is no tracking `address -> id` or `address -> tag & name`, I don't think it should be an issue actually, and just up to owner to correctly clean each id when removing an address from the Provider

src/addresses-provider/AddressesProvider.sol
```diff
@@ -0,0 +1,166 @@
+// SPDX-License-Identifier: LicenseRef-BUSL
```
Kogaroshi, line 1: can we add it to the Deploy procedures, and the Config Engine (tho it might be managed by a different admin than Gov, can be useful to have support if Gov ends up controling the Provider)

avniculae: yes, I'll add a commit to have procedures as well, although I was thinking to not be part of the config engine. can be separate and can be configured in an AIP, wdyt?

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
avniculae: although I think address -> tag & name would be useful imo, wdyt? also I'm thinking whether for an id, we should track the name as well, not just tag and address. wdyt?

avniculae: (in which case maybe one address shouldn't be registered under multiple ids / name&tag)?

src/addresses-provider/AddressesProvider.sol
avniculae: [a82f500](https://github.com/aave/aave-v4/pull/1310/commits/a82f500113092a1f39c6e077ce8b2a7f304b2c16)

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
Kogaroshi: if we add the other side of the tracking, then yes deciding if we allow address registered multiple times would need to be considered. But it might be useful to easy verify an address identity, or if it is a legit V4 approved address.

src/addresses-provider/AddressesProvider.sol
Kogaroshi: Yeah, the Config Engine part can be done later on when needed.
Keeping this open as a reminder.

avniculae: I don't think it needs to be part of config engine, I'm thinking it can be configured in an AIP, wdyt?

Kogaroshi: I'm thinking because we will use the Config Engine to list new Spokes or Hubs, so later those listing in the Provider can be done via the Engine as well as part of the listing/credit line allocation process (if of course the Executor got the ownership of the Provider)

DhairyaSethi: separate AIPs just for this will be very tedious, i am much more inclined to auto update address provider as part of config engine. or we can think of a guardian based solution gated by a steward which validates existence

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
DhairyaSethi: i think its fine to be registered across multiple tags, the reverse map can be address to an array of (name, tag)s

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
DhairyaSethi: whats the benefit here over abi.encode i am not following - if we accept over being stupid then we should allow this and use abi.encode

src/addresses-provider/V4AddressesProvider.sol
```diff
@@ -0,0 +1,166 @@
...
+
+  /// @inheritdoc IV4AddressesProvider
+  function getAddressEntry(bytes32 id) external view returns (AddressEntry memory) {
+    return _addressEntries[id];
+  }
+
+  /// @inheritdoc IV4AddressesProvider
+  function getIds(string memory tag) external view returns (bytes32[] memory) {
+    return _taggedIds[tag].values();
+  }
```
DhairyaSethi: annoyingly we might need to overload with index and length methods instead of this bc we'll get a finding says array size can make inflate over gas limit (unlikely), we've done this way already elsewhere with array getters in the codebase

src/addresses-provider/V4AddressesProvider.sol
```diff
@@ -0,0 +1,166 @@
...
+
+  /// @inheritdoc IV4AddressesProvider
+  function getTags() external view returns (string[] memory) {
+    return _tags.values();
+  }
+
+  /// @inheritdoc IV4AddressesProvider
+  function getCanonicalHub(string memory name) external view returns (address) {
```
DhairyaSethi: calldata all of these pls, even tho to get id you lower to memory for hashing

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
avniculae: I'm thinking if you want to replicate the id yourself it might be easier to do it with the current impl, but I'm good with abi.encode as well. it's also about how you think it conceptually I guess

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
avniculae: you want the revert map if multiple tags are supported as well?

src/addresses-provider/AddressesProvider.sol
avniculae: wdym separate AIPs? The address provider will be the ground truth (like address-book), so I think it should always go through an AIP

src/addresses-provider/V4AddressesProvider.sol
avniculae: yes true, we should do this

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
DhairyaSethi: sure why not, unless youre concerned about bloat

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
DhairyaSethi: can you share a usecase? im not following, would prefer to go w abi encode

src/addresses-provider/AddressesProvider.sol
DhairyaSethi: if its not included in config engine, do you mean we do separate AIPs? my purpose to include this update while listing (which is what config engine is best at) - since you said no - i assume you meant a separate AIP to maintain this

avniculae: hmm I see, you mean when you add a new hub/spoke. I thought on the initialisation at the moment, since we already deployed and we need to do it retroactively. I will add in config engine for future listings

src/addresses-provider/V4AddressesProvider.sol
avniculae: [c012fac](https://github.com/aave/aave-v4/pull/1310/commits/c012fac5d9084b8436e6cd2269fbcd65dcf01bde)

src/addresses-provider/V4AddressesProvider.sol
avniculae: [c012fac](https://github.com/aave/aave-v4/pull/1310/commits/c012fac5d9084b8436e6cd2269fbcd65dcf01bde)

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
avniculae: [c012fac](https://github.com/aave/aave-v4/pull/1310/commits/c012fac5d9084b8436e6cd2269fbcd65dcf01bde)

tests/contracts/addresses-provider/V4AddressesProvider.t.sol
avniculae: [c012fac](https://github.com/aave/aave-v4/pull/1310/commits/c012fac5d9084b8436e6cd2269fbcd65dcf01bde)

Kogaroshi (dismissed):
I still think we should add specific logic in the Config Engine, but impl lgtm

src/addresses-provider/AddressesProvider.sol
avniculae: [80e400b](https://github.com/aave/aave-v4/pull/1310/commits/80e400bed82d3d5ece90ae9c5c701b3190b2c27d) cc @Kogaroshi

DhairyaSethi: we also have the option to do this in the configurator instead

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -42,6 +50,11 @@ interface IAaveV4ConfigEngine {
     address irStrategy;
     IAssetInterestRateStrategy.InterestRateData irData;
     TokenizationSpokeConfig tokenization;
+    IV4AddressesProvider addressesProvider;
+    bool registerHub;
+    string hubName;
+    bool registerTokenizationSpoke;
+    string tokenizationSpokeName;
```
Kogaroshi: could be 2 structs, as their are optional here ?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -174,6 +191,9 @@ interface IAaveV4ConfigEngine {
     address priceSource;
     ISpoke.ReserveConfig config;
     ISpoke.DynamicReserveConfig dynamicConfig;
+    IV4AddressesProvider addressesProvider;
+    bool registerSpoke;
+    string spokeName;
```
Kogaroshi: same here, a struct ?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: idea if if they are not set, struct will be false & empty string by default, and then skipped

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
avniculae: [8e0bf2a](https://github.com/aave/aave-v4/pull/1310/commits/8e0bf2aa8e1d106d1c630e55d2d3c22b3d557fe2)

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
avniculae: [8e0bf2a](https://github.com/aave/aave-v4/pull/1310/commits/8e0bf2aa8e1d106d1c630e55d2d3c22b3d557fe2)

Kogaroshi (dismissed):

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,238 @@
...
+
+  /// @notice Returns the number of entries registered for an address.
+  /// @param addr The registered address.
+  /// @return The number of entries.
+  function getAddressIdCount(address addr) external view returns (uint256);
+
+  /// @notice Returns the identifiers of all entries registered for an address.
+  /// @param addr The registered address.
+  /// @return The list of identifiers.
+  function getAddressIds(address addr) external view returns (bytes32[] memory);
```
DhairyaSethi: we do not need this if we have the one below

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,238 @@
...
+  /// @return The address entry associated with the identifier.
+  function getAddressEntry(bytes32 id) external view returns (AddressEntry memory);
+
+  /// @notice Returns the number of tags with at least one registered entry.
+  /// @return The number of tags.
+  function getTagCount() external view returns (uint256);
+
+  /// @notice Returns all tags with at least one registered entry.
+  /// @return The list of tags.
+  function getTags() external view returns (string[] memory);
```
DhairyaSethi:
```suggestion
  /// @notice Returns the number of tags with at least one registered entry.
  function getTagCount() external view returns (uint256);

  /// @notice Returns all tags with at least one registered entry.
  function getTags() external view returns (string[] memory);
```

src/addresses-provider/AddressesProvider.sol
DhairyaSethi, line 1: idk if this should be in a separate folder, perhaps misc? nbd either way

src/addresses-provider/V4AddressesProvider.sol
```diff
@@ -0,0 +1,253 @@
...
+
+import {Ownable2StepUpgradeable} from 'src/dependencies/openzeppelin-upgradeable/Ownable2StepUpgradeable.sol';
+import {EnumerableSet} from 'src/dependencies/openzeppelin/EnumerableSet.sol';
+import {V4AddressesProviderStorage} from 'src/addresses-provider/V4AddressesProviderStorage.sol';
+import {IV4AddressesProvider} from 'src/addresses-provider/interfaces/IV4AddressesProvider.sol';
+
+/// @title V4AddressesProvider
+/// @author Aave Labs
+/// @notice Main registry of Aave V4 contract addresses.
+abstract contract V4AddressesProvider is
```
DhairyaSethi: thinking more its a registry

src/addresses-provider/AddressesProvider.sol
```diff
@@ -0,0 +1,253 @@
...
+    return _toAddresses(_taggedIds[CANONICAL_SPOKE_TAG].values());
+  }
+
+  /// @inheritdoc IV4AddressesProvider
+  function getTokenizationSpoke(string calldata name) external view returns (address) {
+    return _getAddress({name: name, tag: TOKENIZATION_SPOKE_TAG});
+  }
+
+  /// @inheritdoc IV4AddressesProvider
+  function getTokenizationSpokes() external view returns (address[] memory) {
```
DhairyaSethi: these dont have pointer getters, i would just keep those and not have full arr return one's for all

src/addresses-provider/AddressesProvider.sol
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    emit AddressSet(id, name, tag, oldEntry.addr, newAddress);
+  }
+
+  function _getAddress(string memory name, string memory tag) internal view returns (address) {
+    return _addressEntries[_getId({name: name, tag: tag})].addr;
+  }
+
+  function _getId(string memory name, string memory tag) internal pure returns (bytes32) {
```
DhairyaSethi: could be useful as a public fn

src/config-engine/libraries/HubEngine.sol
```diff
@@ -215,14 +222,40 @@ library HubEngine {
...
   }
 
+  /// @dev Registers the Hub on the AddressesProvider when requested.
+  /// @dev Only allowed when the listed asset is the Hub's first (asset id 0), to avoid registering an
+  /// already-configured Hub; reverts otherwise.
+  function _registerHub(
+    IAaveV4ConfigEngine.AssetListing calldata listing,
+    uint256 assetId
+  ) private {
+    if (!listing.hubRegistration.register) {
```
DhairyaSethi: if input is false i believe we should assert non zero len label

src/config-engine/libraries/HubEngine.sol
```diff
@@ -233,8 +266,6 @@ library HubEngine {
       listing.tokenization.symbol
     );
 
-    uint256 assetId = IHubBase(listing.hub).getAssetId(listing.underlying);
```
DhairyaSethi, line 247: this was dead code?

Kogaroshi: the assetId is retrieved at listing in `_registerHub()` now, and passed, instead of fetching after. Less gas intensive, same result

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: this is meant for convenience

src/addresses-provider/V4AddressesProvider.sol
avniculae: I followed the v3 naming convention here (also with the plural in "addresses" which I don't like). I'm fine changing the name, thoughts @Kogaroshi @miguelmtzinf ?

src/addresses-provider/AddressesProvider.sol
avniculae: we can add range getters, but would keep these for convenience as well, wdyt?

src/addresses-provider/AddressesProvider.sol
avniculae: we have one already

src/config-engine/libraries/HubEngine.sol
avniculae: I agree, we should check these for sanity, in all of the structs used as "optional" params

src/addresses-provider/V4AddressesProvider.sol
Kogaroshi: I think it's better to keep v3 name convention, so we don't have more difference of naming in the tooling.

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: [b55faac](https://github.com/aave/aave-v4/pull/1310/commits/b55faacb732afd056e17869044cc2c42361c033b)

src/addresses-provider/AddressesProvider.sol
avniculae: [b55faac](https://github.com/aave/aave-v4/pull/1310/commits/b55faacb732afd056e17869044cc2c42361c033b)

src/config-engine/libraries/HubEngine.sol
avniculae: [b55faac](https://github.com/aave/aave-v4/pull/1310/commits/b55faacb732afd056e17869044cc2c42361c033b)

src/addresses-provider/interfaces/IV4AddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
...
+  /// @param addr The registered address.
+  /// @param name The name of the entry.
+  /// @param tag The tag grouping the entry.
+  struct AddressEntry {
+    address addr;
+    string name;
+    string tag;
+  }
+
+  /// @notice Emitted when the address associated with a name and tag is updated.
```
miguelmtzinf: when the address of an entry is updated

src/addresses-provider/interfaces/IV4AddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
...
+
+/// @title IV4AddressesProvider
+/// @author Aave Labs
+/// @notice Main registry of the Hub and Spoke addresses of an Aave V4 instance.
+interface IV4AddressesProvider {
+  /// @notice Address entry registered under an identifier.
+  /// @param addr The registered address.
+  /// @param name The name of the entry.
+  /// @param tag The tag grouping the entry.
+  struct AddressEntry {
```
miguelmtzinf: this can be just `Entry`

src/addresses-provider/interfaces/IV4AddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
+// SPDX-License-Identifier: LicenseRef-BUSL
+pragma solidity ^0.8.0;
+
+/// @title IV4AddressesProvider
```
miguelmtzinf: the V4 preffix may not be needed

src/addresses-provider/interfaces/IV4AddressesProvider.sol
miguelmtzinf: event name can be changed to `UpdateEntry` or `SetEntry`
(if keeping, it d be SetAddress ig, for style consistency)

src/addresses-provider/interfaces/IV4AddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
...
+  /// @param newAddress The new address of the entry.
+  event AddressSet(
+    bytes32 indexed id,
+    string name,
+    string tag,
+    address indexed oldAddress,
+    address indexed newAddress
+  );
+
+  /// @notice Thrown when an empty tag is supplied.
```
miguelmtzinf:
```suggestion
  /// @notice Thrown when the specified tag is invalid.
```
same for `InvalidName`

src/addresses-provider/interfaces/IV4AddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
...
+  function CANONICAL_HUB_TAG() external view returns (string memory);
+
+  /// @notice Returns the tag grouping all canonical Spokes.
+  function CANONICAL_SPOKE_TAG() external view returns (string memory);
+
+  /// @notice Returns the tag grouping all tokenization Spokes.
+  function TOKENIZATION_SPOKE_TAG() external view returns (string memory);
+
+  /// @notice Returns the tag grouping all treasury Spokes.
+  function TREASURY_SPOKE_TAG() external view returns (string memory);
```
miguelmtzinf: view functions of constants at the bottom

src/addresses-provider/interfaces/IV4AddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
...
+
+  /// @notice Associates an address with a name, grouped under a tag.
+  /// @dev Associating the zero address removes the entry and its identifier from enumeration, it reverts if no address is registered.
+  /// @dev Reverts if an address is already registered under the identifier, it must be removed first.
+  /// @param name The name of the entry.
+  /// @param tag The tag grouping the entry.
+  /// @param newAddress The address to associate with the name and tag.
+  function setAddress(string calldata name, string calldata tag, address newAddress) external;
```
miguelmtzinf: we can rename to `setEntry` maybe?

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
...
+  /// @notice Returns the number of tags with at least one registered entry.
+  function getTagCount() external view returns (uint256);
+
+  /// @notice Returns all tags with at least one registered entry.
+  function getTags() external view returns (string[] memory);
+
+  /// @notice Returns a slice of the tags with at least one registered entry.
+  /// @param start The start index of the slice.
+  /// @param end The end index of the slice, capped to the number of tags.
+  /// @return The list of tags in the slice.
```
miguelmtzinf, line 68:
```suggestion
  /// @notice Returns the list of tags between the specified indexes.
  /// @dev It ignores provided bounds if they exceed current tag list lengths.
  /// @param start The starting index of the tags list.
  /// @param end The ending index of the tags list.
  /// @return The list of tags.
```

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
...
+
+  /// @notice Returns the address entry associated with an identifier.
+  /// @param id The identifier of the entry.
+  /// @return The address entry associated with the identifier.
+  function getAddressEntry(bytes32 id) external view returns (AddressEntry memory);
+
+  /// @notice Returns the number of tags with at least one registered entry.
+  function getTagCount() external view returns (uint256);
+
+  /// @notice Returns all tags with at least one registered entry.
```
miguelmtzinf:
```suggestion
  /// @notice Returns the list of tags.
```

miguelmtzinf: it actually does not matter if there is at least one registered entry, no?

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,263 @@
...
+  function getAddressEntries(
+    address addr,
+    uint256 start,
+    uint256 end
+  ) external view returns (AddressEntry[] memory);
+
+  /// @notice Returns the canonical Hub associated with a name.
+  /// @param name The name of the Hub.
+  /// @return The address of the Hub, the zero address if none is registered.
+  function getCanonicalHub(string calldata name) external view returns (address);
```
miguelmtzinf: why not by id as well?

src/addresses-provider/V4AddressesProviderStorage.sol
```diff
@@ -0,0 +1,26 @@
...
+  mapping(bytes32 id => IV4AddressesProvider.AddressEntry) internal _addressEntries;
+
+  /// @dev Map of tags to set of entry identifiers.
+  mapping(string tag => EnumerableSet.Bytes32Set ids) internal _taggedIds;
+
+  /// @dev Set of all tags with at least one registered entry.
+  EnumerableSet.StringSet internal _tags;
+
+  /// @dev Map of registered addresses to set of entry identifiers.
+  mapping(address addr => EnumerableSet.Bytes32Set ids) internal _addressIds;
```
miguelmtzinf: so address - id is not 1:1?
do we allow to have multiple entries for the same address? why and what's the use case?

src/addresses-provider/V4AddressesProviderStorage.sol
```diff
@@ -0,0 +1,26 @@
...
+
+  /// @dev Reserved storage space to allow for future layout updates.
+  uint256[49] private __gap;
```
miguelmtzinf: why 49? and not 50

src/addresses-provider/V4AddressesProviderStorage.sol
miguelmtzinf: this would mean that a single address has multiple (name,tag) combinations... don't know if that makes sense

src/addresses-provider/V4AddressesProviderStorage.sol
```diff
@@ -0,0 +1,26 @@
...
+import {EnumerableSet} from 'src/dependencies/openzeppelin/EnumerableSet.sol';
+import {IV4AddressesProvider} from 'src/addresses-provider/interfaces/IV4AddressesProvider.sol';
+
+/// @title V4AddressesProviderStorage
+/// @author Aave Labs
+/// @notice Storage layout for the V4AddressesProvider contract.
+/// @dev This contract defines all storage variables used by the V4AddressesProvider.
+abstract contract V4AddressesProviderStorage {
+  /// @dev Map of entry identifiers to address entries.
+  mapping(bytes32 id => IV4AddressesProvider.AddressEntry) internal _addressEntries;
```
miguelmtzinf: i d follow the same conventions we used as for `AccessManagerEnumerable`.
It's a bit unclear what each var is for (without looking at the definition/docs)

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -24,6 +25,18 @@ interface IAaveV4ConfigEngine {
     string symbol;
   }
 
+  /// @notice Optional registration of a listed Hub or Spoke on the V4AddressesProvider.
+  /// @dev Left unset (the default), `register` is false and the registration is skipped.
+  /// @dev All fields must be set when `register` is true, and left unset when false; reverts otherwise.
```
miguelmtzinf: different behavior than for TokenizationSpoke.. do we want to make them both consistent?
we derive the need for doing the action (registering/deploying) by checking all params are set.

src/config-engine/libraries/HubEngine.sol
```diff
@@ -36,7 +44,8 @@ library HubEngine {
         irData
       );
 
-      _deployAndRegisterTokenizationSpoke(listings[i]);
+      _registerHub(listings[i], assetId);
```
miguelmtzinf: why is this coupled with listing an asset? Is the intention to do this automatically, and allow registering the hub separately even with no assets via direct call?

src/config-engine/libraries/HubEngine.sol
```diff
@@ -215,14 +224,45 @@ library HubEngine {
     }
   }
 
+  /// @dev Registers the Hub on the AddressesProvider when requested.
+  /// @dev Only allowed when the listed asset is the Hub's first (asset id 0), to avoid registering an
+  /// already-configured Hub; reverts otherwise.
+  function _registerHub(
+    IAaveV4ConfigEngine.AssetListing calldata listing,
+    uint256 assetId
```
miguelmtzinf: this is input param is a bit weird, and I d manage this condition outside (better reusability)

src/config-engine/libraries/SpokeEngine.sol
```diff
@@ -225,4 +234,25 @@ library SpokeEngine {
     uint256 assetId = IHubBase(hub).getAssetId(underlying);
     return ISpoke(spoke).getReserveId(hub, assetId);
   }
+
+  /// @dev Registers the Spoke on the AddressesProvider when requested.
+  /// @dev Only allowed when the listed reserve is the Spoke's first (reserve id 0), to avoid
+  /// registering an already-configured Spoke; reverts otherwise.
+  function _registerSpoke(
+    IAaveV4ConfigEngine.ReserveListing calldata listing,
+    uint256 reserveId
```
miguelmtzinf: same, i d move this condition out

miguelmtzinf (changes requested):

src/addresses-provider/V4AddressesProviderStorage.sol
avniculae: yeah, as discussed we'd like to have support for multiple (name,tag) for one address, in case that address is needed for 2 different purposes: e.g., Babylon canonical spoke can be both under canonical and babylon tag

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
avniculae: I think this changed in the latest config engine, we revert if fields are partially set for tokenization spokes. to me this is better since we avoid footguns, wdyr

src/config-engine/libraries/HubEngine.sol
avniculae: yes, idea is to have it both automatic and manual if needed

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: It does matter — a tag is only in the set while it has at least one entry. `_setEntry` removes it from `_tagsSet` as soon as the tag's id set empties, so `getTags()` never returns a tag with zero entries (`test_setEntry_remove` covers it). Kept the qualifier for that reason, but took the rest of your natspec suggestions.

src/addresses-provider/interfaces/IV4AddressesProvider.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/interfaces/IV4AddressesProvider.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/interfaces/IV4AddressesProvider.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/interfaces/IV4AddressesProvider.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/interfaces/IV4AddressesProvider.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/interfaces/IV4AddressesProvider.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/V4AddressesProviderStorage.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/V4AddressesProviderStorage.sol
avniculae: [d0631c3](https://github.com/aave/aave-v4/pull/1310/commits/d0631c39c149b75f601c6f6ebda4b9bf07747641)

src/addresses-provider/interfaces/IV4AddressesProvider.sol
avniculae: [6a1614a](https://github.com/aave/aave-v4/pull/1310/commits/6a1614a1)

src/config-engine/libraries/HubEngine.sol
avniculae: I agree it's weird, but I don't like that it would split the validation into 2 parts. Another option is to fetch the asset/reserve count and assess in this way, wdyt about this?

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,272 @@
+// SPDX-License-Identifier: LicenseRef-BUSL
+pragma solidity ^0.8.0;
+
+/// @title IAddressesProvider
+/// @author Aave Labs
+/// @notice Main registry of the Hub and Spoke addresses of an Aave V4 instance.
```
Kogaroshi: Not consistent with the Natspec for other interfaces

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,272 @@
...
+  /// @notice Thrown when no address is registered under the identifier.
+  error AddressNotSet(bytes32 id);
+
+  /// @notice Associates an address with a name, grouped under a tag.
+  /// @dev Associating the zero address removes the entry and its identifier from enumeration, it reverts if no address is registered.
+  /// @dev Reverts if an address is already registered under the identifier, it must be removed first.
+  /// @param name The name of the entry.
+  /// @param tag The tag grouping the entry.
+  /// @param newAddress The address to associate with the name and tag.
+  function setEntry(string calldata name, string calldata tag, address newAddress) external;
```
Kogaroshi, line 43: Could those params be ordered the same way as in the struct ? for consistency ?

src/config-engine/libraries/HubEngine.sol
Kogaroshi: more in favor of fetching the count instead of passing the parameter here

avniculae: @Kogaroshi I agree, the question is: are you in fabor of fetching the count instead of splitting the validation into 2 parts (Miguel's suggestion)?

Kogaroshi: since we only need it for the require here, I would fetch it locally here only

avniculae: [d7e3340](https://github.com/aave/aave-v4/pull/1310/commits/d7e33401040ebe17d2ecd8ee5f2469651f855aa2)

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: [d7e3340](https://github.com/aave/aave-v4/pull/1310/commits/d7e33401040ebe17d2ecd8ee5f2469651f855aa2)

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: [d7e3340](https://github.com/aave/aave-v4/pull/1310/commits/d7e33401040ebe17d2ecd8ee5f2469651f855aa2)

Kogaroshi (dismissed):

src/addresses-provider/AddressesProvider.sol
DhairyaSethi: its more fool proof if we do it in the configurator @avniculae

src/addresses-provider/AddressesProvider.sol
DhairyaSethi: i would kill these, unnecessary take space

DhairyaSethi: we NEED range getters bc of potential out of gas, these arent needed

src/config-engine/libraries/EngineUtils.sol
```diff
@@ -0,0 +1,22 @@
...
+pragma solidity ^0.8.0;
+
+import {IAaveV4ConfigEngine} from 'src/config-engine/interfaces/IAaveV4ConfigEngine.sol';
+
+/// @title EngineUtils
+/// @author Aave Labs
+/// @notice Library containing shared helpers for the AaveV4ConfigEngine libraries.
+library EngineUtils {
+  /// @dev Returns whether an optional AddressesProvider registration is consistent: all fields must
+  /// be set when registering, and left unset otherwise.
```
DhairyaSethi: it should be left unset only when it's already populated no? i would not keep it optional, i thought we settled on this

src/config-engine/libraries/HubEngine.sol
```diff
@@ -19,19 +20,26 @@ library HubEngine {
   /// KEEP_CURRENT sentinel. All fields must be explicitly set when the strategy changes.
   error InvalidIrDataWithNewStrategy();
 
+  /// @dev Thrown when an addresses provider registration is requested for a listing that does not
```
DhairyaSethi: i would keep it strict by default, optional only when its already set

src/config-engine/libraries/HubEngine.sol
```diff
@@ -219,19 +228,48 @@ library HubEngine {
...
+  /// already-configured Hub; reverts otherwise.
+  function _registerHub(IAaveV4ConfigEngine.AssetListing calldata listing) private {
+    require(
+      EngineUtils.isConsistentRegistration(listing.hubRegistration),
+      InvalidAddressesProviderRegistration()
+    );
+    if (!listing.hubRegistration.register) {
+      return;
+    }
+    require(IHub(listing.hub).getAssetCount() == 1, InvalidAddressesProviderRegistration());
```
DhairyaSethi: this is too strict, we cannot retrofix

DhairyaSethi: i would also check if its present in the address provider here

src/addresses-provider/AddressesProvider.sol
avniculae: settled offline with Dhairya: the engine now reverts if a hub/spoke it targets is not registered in the provider, and registering/unregistering is a dedicated engine action (executeAddressesProviderEntryUpdates, runs before hub/spoke actions in the payload). [5650778](https://github.com/aave/aave-v4/pull/1310/commits/5650778bec3275218d88a79ade021bd99bd55510)

src/config-engine/libraries/EngineUtils.sol
avniculae: [5650778](https://github.com/aave/aave-v4/pull/1310/commits/5650778bec3275218d88a79ade021bd99bd55510) — registration is no longer optional, engine actions require the target to be registered upfront

src/config-engine/libraries/HubEngine.sol
avniculae: [5650778](https://github.com/aave/aave-v4/pull/1310/commits/5650778bec3275218d88a79ade021bd99bd55510)

src/config-engine/libraries/HubEngine.sol
avniculae: [5650778](https://github.com/aave/aave-v4/pull/1310/commits/5650778bec3275218d88a79ade021bd99bd55510) — the count guard is gone, listings now require the hub to already be registered in the provider, so existing hubs can be registered retroactively via the entry updates action

src/addresses-provider/AddressesProvider.sol
```diff
@@ -0,0 +1,288 @@
...
+  function getEntries(
+    address addr,
+    uint256 start,
+    uint256 end
+  ) external view returns (Entry[] memory) {
+    return _toEntries(_addressToIdSet[addr].values(start, end));
+  }
+
+  /// @inheritdoc IAddressesProvider
+  function isRegistered(address addr, string calldata tag) external view returns (bool) {
```
avniculae, line 115: flagging: this walks every id of the address and hashes each tag, so it degrades if one address is registered under many names/tags. The engine can also check the same address several times in one payload (`requireRegisteredSpoke` does up to 3 `isRegistered` calls per item). Probably fine to leave for now since it is owner-driven and the sets stay small, but if AIP gas becomes a concern we may want to optimise (e.g. a tag-keyed lookup).

DhairyaSethi: config engine can just check if its registered on any tag, if so it passes no?

DhairyaSethi: in the optional tag input case

src/addresses-provider/AddressesProvider.sol
DhairyaSethi: @avniculae

src/config-engine/libraries/EngineUtils.sol
```diff
@@ -1,22 +1,55 @@
...
+  /// @dev Reverts unless the Spoke is registered on the AddressesProvider under a spoke tag
+  /// (canonical, tokenization or treasury). For flows supporting any spoke participant.
+  function requireRegisteredSpoke(
+    IAddressesProvider addressesProvider,
+    address spoke
+  ) internal view {
+    require(
+      addressesProvider.isRegistered(spoke, addressesProvider.CANONICAL_SPOKE_TAG()) ||
+        addressesProvider.isRegistered(spoke, addressesProvider.TOKENIZATION_SPOKE_TAG()) ||
+        addressesProvider.isRegistered(spoke, addressesProvider.TREASURY_SPOKE_TAG()),
```
DhairyaSethi, line 39: maybe we should instead maintain a reverse map as well such that we can just do `isRegistered(spoke)`? i reckon safer that we validate type as well so not sure

src/addresses-provider/AddressesProvider.sol
avniculae: I am not aligned here. What's the concern if we have both range-based and full getters?

DhairyaSethi: overpopulation

src/addresses-provider/AddressesProvider.sol
avniculae: I'd rather keep strongly checking the type — an entry under an unrelated tag shouldn't authorize hub/spoke actions. Made the typed check O(1) instead by tracking a per-address tag count in setEntry: [e7c9097](https://github.com/aave/aave-v4/pull/1310/commits/e7c9097de478c933734d72105c51ad325052e709)

src/config-engine/libraries/EngineUtils.sol
avniculae: [e7c9097](https://github.com/aave/aave-v4/pull/1310/commits/e7c9097de478c933734d72105c51ad325052e709) — kept the typed checks and made isRegistered a single mapping read (per-address tag count maintained in setEntry), so we don't have to trade type validation for gas

src/addresses-provider/AddressesProvider.sol
avniculae: went further as discussed: dropped all convenience functions, not just the full-array ones — typed set*/get* and unbounded getters are gone, consumers use setEntry and the count/range getters with the public tag constants. [f369c9a](https://github.com/aave/aave-v4/pull/1310/commits/f369c9a99b9aedd0d0f816dfbf5e759d8d509720)

src/addresses-provider/AddressesProvider.sol
Kogaroshi, line 29: are those still useful ?

Kogaroshi: is the goal to just use them as "preset" tags for ease listing ? Maybe they can live in their own lib and be used only by the ConfigEngine then, kind of like we do for the Roles ?

avniculae: yeah, I was thinking to include them here since we've seen challenges with Roles pattern. But we can do it for consistency, wdyt? tagging @DhairyaSethi as well

src/addresses-provider/AddressesProvider.sol
```diff
@@ -0,0 +1,175 @@
...
+import {AddressesProviderStorage} from 'src/addresses-provider/AddressesProviderStorage.sol';
+import {IAddressesProvider} from 'src/addresses-provider/interfaces/IAddressesProvider.sol';
+
+/// @title AddressesProvider
+/// @author Aave Labs
+/// @notice Main registry of Aave V4 contract addresses.
+abstract contract AddressesProvider is
+  AddressesProviderStorage,
+  Ownable2StepUpgradeable,
+  IAddressesProvider
```
miguelmtzinf:
```suggestion
  IAddressesProvider,
  AddressesProviderStorage,
  Ownable2StepUpgradeable
```

src/addresses-provider/AddressesProvider.sol
```diff
@@ -0,0 +1,175 @@
...
+    address addr,
+    uint256 start,
+    uint256 end
+  ) external view returns (Entry[] memory) {
+    return _toEntries(_addressToIdSet[addr].values(start, end));
+  }
+
+  /// @inheritdoc IAddressesProvider
+  function isRegistered(address addr, string calldata tag) external view returns (bool) {
+    return _addressToTagCount[addr][keccak256(bytes(tag))] > 0;
```
miguelmtzinf: here is the only place where this is used.
we could achieve the same thing by iterating over `_idToEntry[_addressToIdSet[addr]][i]` right
i wonder if makes more sense to store entires per tag instead: `_idToTagToEntry`

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,160 @@
...
+  /// @param name The name of the entry.
+  /// @param tag The tag grouping the entry.
+  /// @param oldAddress The previous address of the entry.
+  /// @param newAddress The new address of the entry.
+  event SetEntry(
+    bytes32 indexed id,
+    string name,
+    string tag,
+    address indexed oldAddress,
+    address indexed newAddress
```
miguelmtzinf: not aligned with other events of V4.
`oldAddress` or `newAddress` is zero when creating or removing.

src/config-engine/libraries/EngineUtils.sol
```diff
@@ -0,0 +1,55 @@
...
```
AlbertoCentonze, line 39: Should we have some extra validation on overlaps between a treasury spoke and other spokes? I'm thinking that if something is registered as both a treasury spoke and a tokenization spoke there's probably something wrong with it?

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,160 @@
...
+  /// @notice Returns the tag grouping all canonical Hubs.
+  function CANONICAL_HUB_TAG() external view returns (string memory);
+
+  /// @notice Returns the tag grouping all canonical Spokes.
+  function CANONICAL_SPOKE_TAG() external view returns (string memory);
+
+  /// @notice Returns the tag grouping all tokenization Spokes.
+  function TOKENIZATION_SPOKE_TAG() external view returns (string memory);
+
+  /// @notice Returns the tag grouping all treasury Spokes.
```
miguelmtzinf:
```suggestion
  /// @notice Returns the tag grouping all Treasury Spokes.
```

src/addresses-provider/interfaces/IAddressesProvider.sol
```diff
@@ -0,0 +1,160 @@
...
+  /// @return True if at least one entry associates the address with the tag.
+  function isRegistered(address addr, string calldata tag) external view returns (bool);
+
+  /// @notice Returns the tag grouping all canonical Hubs.
+  function CANONICAL_HUB_TAG() external view returns (string memory);
+
+  /// @notice Returns the tag grouping all canonical Spokes.
+  function CANONICAL_SPOKE_TAG() external view returns (string memory);
+
+  /// @notice Returns the tag grouping all tokenization Spokes.
```
miguelmtzinf:
```suggestion
  /// @notice Returns the tag grouping all Tokenization Spokes.
```

src/addresses-provider/AddressesProvider.sol
```diff
@@ -0,0 +1,175 @@
...
+    Entry memory oldEntry = _idToEntry[id];
+
+    if (newAddress == address(0)) {
+      require(oldEntry.addr != address(0), AddressNotSet(id));
+      _tagToIdSet[oldEntry.tag].remove(id);
+      if (_tagToIdSet[oldEntry.tag].length() == 0) {
+        _tagsSet.remove(oldEntry.tag);
+      }
+      _addressToIdSet[oldEntry.addr].remove(id);
+      _addressToTagCount[oldEntry.addr][keccak256(bytes(oldEntry.tag))]--;
```
miguelmtzinf:
```suggestion
      _addressToTagToIdCount[oldEntry.addr][keccak256(bytes(oldEntry.tag))]--;
```
it's the count of entries indeed, and the var name seems to refer to number of tags

src/addresses-provider/AddressesProvider.sol
avniculae: The counter is there to keep this constant-time: the config engine calls `isRegistered` up to three times per action (EngineUtils), and iterating the address's entries means loading two strings per entry plus a string compare, unbounded in the number of entries. The extra cost is one slot write in `setEntry`, which is owner-only and rare, so I think the trade is worth it.

On `_idToTagToEntry` — can you expand on the layout you have in mind? The id already encodes the tag, and `isRegistered` takes `(addr, tag)` with no name, so I don't see how to resolve it from that mapping. If you mean per-address per-tag id sets replacing `_addressToIdSet`, that makes this a `length() > 0` check, but then `getAddressIds`/`getEntries` would have to iterate tags to enumerate an address's entries.

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: Done in 5788639c — dropped `oldAddress` from the event.

src/addresses-provider/AddressesProvider.sol
avniculae: Done in 8f438029.

src/addresses-provider/AddressesProvider.sol
avniculae: Done in 8f438029 — renamed to `_addressToTagToIdCount`.

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: Done in 8f438029 — capitalized all four tag comments so the block stays consistent.

src/addresses-provider/interfaces/IAddressesProvider.sol
avniculae: Done in 8f438029.

src/addresses-provider/AddressesProvider.sol
avniculae: @miguelmtzinf @DhairyaSethi one more angle on reducing tracking: on-chain, only `isRegistered` and the point lookups are consumed. The enumeration views — `getTags`, `getIds`/`getAddresses` per tag, `getAddressIds`/`getEntries` per address — have no on-chain callers, and each group carries its own index (`_tagsSet`, `_tagToIdSet`, `_addressToIdSet`). They're there for convenience, so tooling can read the registry without an indexer.

If we want fewer data structures, we can drop any of these views and reconstruct them off-chain from `SetEntry` events — in the extreme down to just `_idToEntry` plus the `isRegistered` counter, keeping all on-chain consumers and their gas exactly as they are. Open to trimming if we agree on-chain enumeration isn't a requirement.

DhairyaSethi: i like enumeration honestly, we literally created an additional contract on top of access manager to enable enumeration (whose idea it was doesn;t matter ;))

miguelmtzinf: i prefer to avoid reliance on offchain indexing. My comment was about whether the current storage layout can be enhanced so we have fewer structures and better capabilities.

src/config-engine/libraries/SpokeEngine.sol
```diff
@@ -14,12 +16,17 @@ library SpokeEngine {
...
+  /// @dev The Spoke must be registered on the AddressesProvider as a canonical Spoke.
   /// @param listings The reserve listings to execute.
+  /// @param addressesProvider The AddressesProvider authorizing the actions.
   function executeSpokeReserveListings(
-    IAaveV4ConfigEngine.ReserveListing[] calldata listings
+    IAaveV4ConfigEngine.ReserveListing[] calldata listings,
+    IAddressesProvider addressesProvider
   ) external {
     uint256 length = listings.length;
     for (uint256 i; i < length; ++i) {
```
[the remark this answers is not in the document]
avniculae: Accepted for `executeSpokeReserveListings` — the Spoke accepts any address implementing `getAssetUnderlyingAndDecimals`, so the engine is the only place to enforce hub registration at listing time. Fixed in ff1e2989.

The three update functions are not the same gap: they resolve the reserve via `ISpoke.getReserveId(hub, assetId)`, which reverts with `ReserveNotListed` unless that (hub, assetId) already backs a listed reserve on the spoke. A wrong or unregistered hub can't reach any state change there, so no check added on those paths.

PR #1315, opened 2026-06-30
feat: beacon proxify tokenization spoke

DhairyaSethi:


src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -15,10 +15,12 @@ import {IAssetInterestRateStrategy} from 'src/hub/interfaces/IAssetInterestRateS
 /// the universal KEEP_CURRENT sentinel. Boolean fields use uint256 (0=false, 1=true, KEEP_CURRENT=skip).
 interface IAaveV4ConfigEngine {
   /// @notice Parameters for tokenization of an asset on a Hub when listing the asset.
+  /// @dev beacon The shared TokenizationSpoke beacon the deployed proxy points to.
   /// @dev addCap The add cap for the TokenizationSpoke (0 means no tokenization).
   /// @dev name The name for the TokenizationSpoke.
   /// @dev symbol The symbol for the TokenizationSpoke.
   struct TokenizationSpokeConfig {
+    address beacon;
```
DhairyaSethi, line 23: should this be here or should as we add it as an immutable in hubEngine and it's non configurable? in the config engine or helpers we can abstract this by pulling from address-book (v3 takes the former approach)

src/spoke/TokenizationSpokeStorage.sol
```diff
@@ -0,0 +1,25 @@
...
+/// @title TokenizationSpokeStorage
+/// @author Aave Labs
+/// @notice Storage layout for the TokenizationSpoke contract.
+/// @dev This contract defines all storage variables used by TokenizationSpoke.
+abstract contract TokenizationSpokeStorage {
+  /// @dev The associated Hub contract.
+  IHub internal _hub;
+
+  /// @dev The identifier of the tokenized asset on the Hub.
+  uint256 internal _assetId; // todo: change to 96 to save an sload, prob ok limitation
```
DhairyaSethi, line 15: i wanna make this 96 such that these are packed in one slot, it makes the spoke limited, and i think it's probably ok limitation since reasonably should not be able to cross this asset id on a hub realistically (as its incremental)
normal spoke restricts to uint16

DhairyaSethi: or lower actually to bring in decimals here as well for the preview getters which read hub, assetId & decimals but not asset

src/spoke/instances/TokenizationSpokeInstance.sol
```diff
@@ -6,25 +6,29 @@ import {TokenizationSpoke} from 'src/spoke/TokenizationSpoke.sol';
 /// @title TokenizationSpokeInstance
 /// @author Aave Labs
 /// @notice Implementation contract for the TokenizationSpoke.
+/// @dev A single instance is shared by all TokenizationSpoke beacon proxies, so the Hub and
+/// tokenized asset details are provided to the initializer rather than the constructor.
```
miguelmtzinf, line 10:
```suggestion
/// @dev A single stateless implementation is shared across all TokenizationSpoke beacon proxies.
```

src/spoke/TokenizationSpokeStorage.sol
miguelmtzinf: TokenizationSpoke must be closer to Hub than to the Spoke in general, given that goal would be provide easy access to Hubs' deposits. Having said that, 96 means 7.9e28 ids.. this is completely okay.
88 bits d fine as well, in case we wanna pack decimals but not sure if there is much benefit over there

src/spoke/TokenizationSpoke.sol
```diff
@@ -32,35 +38,30 @@ abstract contract TokenizationSpoke is ITokenizationSpoke, ERC20Upgradeable, Int
...
+  function initialize(
+    address hub_,
+    address underlying_,
+    string memory shareName,
+    string memory shareSymbol
+  ) external virtual;
 
-  /// @dev Constructor.
+  /// @dev Sets the Hub and tokenized asset details and the vault share token's ERC20 name and
+  /// symbol. Must be called at first initialization.
```
miguelmtzinf, line 51:
```suggestion
  /// @dev Initializes the TokenizationSpoke vault parameters, including vault share token's ERC20 name and symbol.
  /// Must be called at first initialization.
```

src/spoke/TokenizationSpoke.sol
```diff
@@ -32,35 +38,30 @@ abstract contract TokenizationSpoke is ITokenizationSpoke, ERC20Upgradeable, Int
...
-
-  /// @dev To be overridden by the inheriting TokenizationSpokeInstance contract.
-  function initialize(string memory shareName, string memory shareSymbol) external virtual;
-
-  /// @dev Sets the vault share token's ERC20 name and symbol. Must be called at first initialization.
+  /// @param shareName The ERC20 name of the share issued by this vault.
+  /// @param shareSymbol The ERC20 symbol of the share issued by this vault.
   function __TokenizationSpoke_init(
+    address hub_,
+    address underlying_,
```
miguelmtzinf, line 58:
```suggestion
    address hub,
    address underlying,
```
fix docs too pls

src/spoke/TokenizationSpoke.sol
```diff
@@ -32,35 +38,30 @@ abstract contract TokenizationSpoke is ITokenizationSpoke, ERC20Upgradeable, Int
...
-  /// @dev Immutable references to the Hub and tokenized asset details.
-  IHub internal immutable HUB;
-  uint256 internal immutable ASSET_ID;
-  address internal immutable ASSET;
-  uint8 internal immutable DECIMALS;
-  uint256 internal immutable ASSET_UNITS;
+  /// @dev To be overridden by the inheriting TokenizationSpokeInstance contract.
+  function initialize(
+    address hub_,
+    address underlying_,
```
miguelmtzinf, line 45:
```suggestion
    address hub,
    address underlying,
```

src/spoke/TokenizationSpoke.sol
DhairyaSethi: this was intentional because of shadow warning (we have hub() getter) but yeah will see

src/spoke/TokenizationSpokeStorage.sol
DhairyaSethi: `decimals` packed here allows the same warm sload for supply preview getters where we need to normalize addCap with assetUnits

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
Kogaroshi: If the goal is to have 1 beacon per chain, should be an immutable in the Hub Engine, shouldn't allow to pass it to avoid mistakes.
If we want 1 beacon per market (for example between current one, and a whitelabel one or Horizon), then we should leave it configurable, or derived from the HubConfigurator managing a specific instance

src/spoke/TokenizationSpokeStorage.sol
Kogaroshi: Agree, packing all 3 together should be fine here, chance to reach the max possible assetId here is really low.

PR #1317, opened 2026-07-09
perf(libraries): branchless MathUtils.add(uint256,int256)

### LINKED ISSUE #641: opt: branchless MathUtils.add(uint,int)
https://github.com/aave/aave-v4/blob/63caec0c2d2d22f345fe4372c959465dc2f294b4/src/libraries/math/MathUtils.sol#L46-L56

This can be optimised to be written branchless-ly. See [min](https://github.com/aave/aave-v4/blob/63caec0c2d2d22f345fe4372c959465dc2f294b4/src/libraries/math/MathUtils.sol#L42) implementation above or [oz](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/0034c302241c4b1a1685272d4df42ca5d64b8c34/contracts/utils/math/SignedMath.sol#L13-L26) for inspiration.

ozpool:
Closes #641.

Rewrites `MathUtils.add(uint256, int256)` branchlessly, as suggested in the issue.

### Approach
`add` wraps modulo 2**256, so reinterpreting `b` as its two's-complement word gives the correct result for both signs whenever it fits in a `uint256`. The sum is invalid exactly when `(c < a) != (b < 0)`:

- `b >= 0` overflows ⇒ `c < a`
- `b < 0` underflows ⇒ `c >= a`

which folds to a single `xor(lt(c, a), slt(b, 0))`. Reverts are kept as an arithmetic panic (`0x11`), so the observable behaviour is unchanged and the existing tests pass untouched.

### Note on `type(int256).min`
The old code reverted for `b == type(int256).min` regardless of `a`, purely as an artifact of the checked negation `-b`. The branchless version instead returns the mathematically correct `a - 2**255` when it fits (`a >= 2**255`). This is only reachable outside the range the existing tests exercise; `a < 2**255` still reverts. Covered by a dedicated edge test.

### Tests & gas
- Added `test_fuzz_add` covering the full `int256` range (incl. `type(int256).min`) and `test_add_int256_min_edge`.
- `forge test` green; gas snapshots refreshed — reductions across borrow, repay, liquidation and premium paths.


Annkkitaaa (commented):
I went through the assembly by hand against the full two's-complement range and this holds up. Walking through why, for the record:

`c := add(a, b)` reinterprets `b`'s two's-complement bit pattern as a raw word and relies on `ADD` wrapping mod 2**256, which is exactly right: if `b` represents `-m` for some `m > 0`, its unsigned word value is `2**256 - m`, so `add(a, b) mod 2**256 = (a - m) mod 2**256`, giving the correct wrapped result in both the valid and the underflowing case.

The revert condition `xor(lt(c, a), slt(b, 0))` is the correct overflow/underflow check:
- `b >= 0`: overflow occurred exactly when `c < a` (the sum wrapped past 2**256). No overflow means `c >= a`. `slt(b, 0)` is `0` in this branch, so the xor is `0` (no revert) iff `c >= a`, matching.
- `b < 0`: valid iff `a >= m`, which gives `c = a - m <= a`, i.e. `lt(c, a)` is `1` when `a > m` and `0` only at the exact boundary `a == m` (where `c == 0`, still `< a` unless `a == 0`). `slt(b, 0)` is `1` in this branch, so xor is `0` (no revert) iff `lt(c, a)` is `1`, i.e. iff `a > m`... worth double-checking the `a == m` boundary specifically since that's `c == 0`: `lt(0, a)` is `1` for any `a > 0`, so it correctly does not revert there.

I specifically checked `b == type(int256).min`, since that's the case that breaks naive "negate and subtract" implementations (negating `INT256_MIN` overflows Solidity's checked int256 range). Two's complement of `type(int256).min` is `2**255`, so `m = 2**255`. At `a = 2**255`: `c = add(2**255, 2**255) mod 2**256 = 0`, `lt(0, 2**255) = 1`, `slt(b,0) = 1`, xor = 0, correctly no revert, and `0` is indeed the right answer (`2**255 + (-2**255) = 0`). At `a = 2**255 - 1`: `c` wraps to `2**256 - 1`, `lt(UINT256_MAX, a) = 0`, xor = 1, correctly reverts (true result is negative). The fuzz test and the dedicated `test_add_int256_min_edge` both cover this correctly.

One thing worth calling out explicitly in the PR description, since it's easy to miss under a "perf" label: **this is also a correctness fix, not only a gas optimization.** The old implementation (`if (b >= 0) return a + uint256(b); return a - uint256(-b);`) unconditionally reverts whenever `b == type(int256).min`, regardless of `a`, because unary negation of `INT256_MIN` overflows Solidity's checked arithmetic before the subtraction ever runs. So `add(2**255, type(int256).min)` reverts today even though the mathematically correct result (`0`) is well-defined and in range. The new version fixes that as a side effect of the branchless rewrite.

For what it's worth on current exploitability: I checked the two call sites (`Hub.sol:945` and `UserPositionUtils.sol:41`), both apply `add` to premium-share deltas computed as `newPremiumShares.signedSub(oldPremiumShares)`, where both operands are bounded well below `type(int256).min` in practice (shares are packed into `uint120` elsewhere), so today's callers can't actually hit the old bug. Still a real robustness improvement, and worth having regardless of current reachability.

Gas snapshot diffs across `Hub.Operations.json`, `Spoke.Operations.json`, etc. all move in the same direction (small decreases), consistent with a pure efficiency win layered on top of the correctness fix.

PR #1322, opened 2026-07-20
added the hub connected with TreasurySpoke in the constructor, so each treasury spoke will be connected with a hub

### LINKED ISSUE #799: Update Treasury Spoke - allow for registering reserveIds
- currently treasury spoke doesnt really behave like a regular spoke; it is connected to 1 hub only and the functions like supply and withdraw accept a reserveId that MUST match the assetId on the hub, as it calls the hub directly with that reserveId input in place of assetId input
- treasury spoke can be misconfigured on the hub, if the same treasury spoke is used across multiple hubs, therefore need to deploy a new treasury spoke per hub

could lead to fees being bricked if the same treasury spoke is used across hubs.
- recently we made sure on addAsset and updateAssetConfig when there is a new feeReceiver, it must be a new spoke (for that hub) which then gets added to the hub
- in treasury spoke, you must define the hub it is connected with in the constructor and it is immutable

See the following scenario:
- deploy treasurySpoke feeReceiver1 connected to hub1
- on hub1, add assetA with feeReceiver1; feeReceiver1 gets added as a new spoke
- on hub2, add assetA with feeReceiver1 as well; this is allowed bc feeReceiver1 is a new spoke to hub2

ref of discussion: https://avaraxyz.slack.com/archives/C04HCBMEEBZ/p1758595490572469

web3devLuBdF:
Closes #799 

An Immutable variabile named "HUB" has been added at the contract level , because it is immutable it will be part of the final bytecode and its value will not can be changed after deployment and it has been marked as public.
A constructor has been added in "TreasurySpoke" contract and the variabile "HUB" has been initialized in it, so each treasury spoke will have a hub.
Although "TreasurySpoke" contains upgradeable elements (for Ownable2StepUpgradeable), the presence of constructor is not a problem if it does not affect upgradeable pattern logic.
Naturally "TreasurySpokeInstance" and "MockTreasurySpokeInstance" contracts that inherit directly by "TreasurySpoke"  had to indicate "HUB" variable as argument in their constructors.
Initially the deployment of treasury spoke instance failed in treasury spoke tests with the error "FailedCreate2FactoryCall" because it is deployed with CREATE2, so the addition of constructor in "TreasurySpoke" contract has changed final expected bytecode in deterministic deployment.
Because of this has been necessary to encode the "TreasurySpoke" argument in the final bytecode, and all files involved in treasury spoke deployment have added this argument to ensure that deployment succeds.
Since all this files are quite numerous, is necessary to be careful to how the argument integrates with the other variables.
For this reason if is needed change "HUB" variable visibility, modify the position of treasury spoke instance argument in all files involved in deployment or any other change and addition, i am willing to do it.


DhairyaSethi:
feature is not planned

PR #1325, opened 2026-07-23
ether.fi Cash whitelabel instance on OP Mainnet — two-phase launch (dormant config + activation), mirroring the Avalanche v4 pattern

seongyun-ko:
## What this is

Launch configuration and deployment tooling for the dedicated Aave V4 whitelabel instance, fully managed by ether.fi, powering ether.fi Cash on OP Mainnet.

Governance reference: [ARFC — Deploy a dedicated Aave v4 whitelabel instance fully managed by EtherFi on OP Mainnet to power ether.fi Cash](https://governance.aave.com/t/arfc-deploy-a-dedicated-aave-v4-whitelabel-instance-fully-managed-by-etherfi-on-op-mainnet-to-power-ether-fi-cash/25314)

This is a review branch in the spirit of the `feat/deployment` mainnet-launch branch — we do not necessarily expect a merge; we want Aave eng eyes on it.

## Launch model: two-phase (config-dormant → verify → activate)

Same pattern as the Aave V4 Avalanche launch ([activation payload](https://github.com/aave-dao/aave-proposals-v3/blob/7bb6bcb394723856ebeb04e020a566da0978efa4/src/20260708_AaveV4Avalanche_AaveV4AvalancheActivation/AaveV4Avalanche_AaveV4AvalancheActivation_20260708.sol), proposal 504):

1. **Phase 1 — `EtherfiCashLaunchPayload`**: configures everything with every spoke registered `active: false`. 19 asset listings (USDC and WETH borrowable, everything else collateral-only with a flat 0% curve), Cash Spoke registration with final add/draw caps, reserve listings with the final risk parameters from Nonce Capital (risk curator), liquidation engine (target HF 1.24, HF for max bonus 0.90), and two granular operator roles (hub caps = role 201, spoke dynamic reserve config = role 401) granted to the risk-curator Safe and the owner Safe. The market is fully configured but unusable.
2. **On-chain verification** of the dormant state, field by field (`VerifyEtherfiCashLive`, `EXPECT_ACTIVE=false`).
3. **Phase 2 — `EtherfiCashActivationPayload`**: enumerates `getAssetCount()` / `getSpokeCount()` / `getSpokeAddress()` and flips `updateSpokeActive(true)` on every pair — identical shape to the Avalanche activation, no hardcoded asset list.

Both phases are executed by the Owner Safe (Safe v1.4.1 on OP) as delegatecall transactions — same topology as `Deploy.s.sol` on the `feat/deployment` branch, with the Safe in the executor seat. Both payloads hold only immutables and constants (no storage writes), so the delegatecalls cannot touch Safe state.

## Contents

- **`src/etherfi/EtherfiCashLaunchPayload.sol`** — phase 1 (config-engine based, dormant).
- **`src/etherfi/EtherfiCashActivationPayload.sol`** — phase 2 (enumeration-based activation).
- **`src/etherfi/EtherfiCashOpMainnet.sol`** — the one file holding all addresses and caps; every underlying and price feed verified on-chain; instance addresses predicted from deterministic simulation (see PR comment).
- **`scripts/etherfi/`** — instance deployment (one hub `CASH_HUB`, one spoke `CASH_SPOKE`, all admin seats = the ether.fi Owner Safe), config-engine deployment, payload deployment via CREATE2, read-only preflight validation, Owner-Safe delegatecall transaction generation (both phases), post-execution state verification (dormant and active modes), and a one-command pipeline (`launch.sh`) that refuses to broadcast until a fork dress rehearsal of the full two-phase sequence passes.
- **`scripts/etherfi/LAUNCH_SPEC.md`** — the review artifact: full parameter tables and predicted addresses, generated from the payload source so it cannot drift.
- **`tests/etherfi/`** — governance-topology simulation (9 tests, including dormant registration and activation) plus a fork rehearsal test covering deploy → dormant config → verify → activate → verify.

## What we'd like from Aave eng

1. **Deployment blessing / license confirmation** — the v4 contracts are BUSL-licensed; please confirm the whitelabel arrangement covers this instance deployment, and whether you prefer to run the instance deployment yourselves or are fine with us running `DeployEtherfiCashInstance.s.sol` (all admin/owner seats point at the ether.fi Owner Safe).
2. **Canonical commit** — which commit/tag you consider the audited release to deploy the instance from (this branch is cut from `main` @ `2524fe40`).
3. **Payload review** — both payloads will be deployed at immutable addresses (predicted in the PR comment) for a bytecode-vs-source review.
4. **Address book** — post-deployment, we'd appreciate an `aave-address-book` entry for the instance so integrators get the addresses the standard way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


seongyun-ko:
## Predicted deployment addresses (deterministic simulation) — REVISED (hardcoded payloads)

We ran the full deployment as the launch deployer (`0xf8a86ea1Ac39EC529814c377Bd484387D395421e`) on an OP Mainnet fork, three times from clean state — all runs reproduce the same instance addresses. Contracts deploy via CREATE2 through the Safe Singleton Factory (`0x914d7Fec6aaC8cd542e72Bca78B30650d45643d7`). Assumptions: this branch's commit + compiler settings, same deployer, same salts.

Following review feedback, both payloads are now fully hardcoded from the `AaveV4EtherfiCash` address-book libraries (zero constructor args, all functions `pure`) — which also makes both payload addresses pure-CREATE2 deterministic. The nonce-sensitivity is confined to the AaveOracle and the Cash Spoke themselves (the oracle is `new`-deployed and feeds the spoke deployment): they require the deployer to start at **nonce 3** with the exact tx sequence and no interleaved transactions. If the instance were ever deployed differently, the payloads would still land at their addresses but point at stale constants — the preflight validator gates on live code at every pinned address before any broadcast.

| Contract | Address | Requires |
|---|---|---|
| AccessManager | `0x188d7173772499FB6375F23FdFd130CE6107286b` | CREATE2 only |
| AaveV4ConfigEngine | `0x84210b3087E952Be0f3610fD75f0f045995eAF22` | CREATE2 only (deployer-independent; broadcast includes its 4 external engine libraries) |
| CASH_HUB (proxy) | `0x66753c4e3fC84f1eD0e3C267C927284E9d90C572` | CREATE2 only |
| CASH_HUB implementation | `0x697aF34263Bc8E5E6B0e6b0C37A1EC58CeDE2cE9` | CREATE2 only |
| HubConfigurator | `0xA39bEf2fD611fb9c5a69D63277b4Af97a30F0dbC` | CREATE2 only |
| InterestRateStrategy | `0x51d07C362f9c4716F96EbEB63DB985EF9D2aCd7C` | CREATE2 only |
| SpokeConfigurator | `0xFEe9E8cCE1c40D3bd9F025437D3A11cA0DAe9f8b` | CREATE2 only |
| TreasurySpoke | `0x7EB4d25F137868662350603A2863F682287b0768` | CREATE2 only |
| LiquidationLogic | `0x818E84198224535FAeaEc1b583d3Ff6b812A5AF3` | CREATE2 only |
| NativeTokenGateway | `0xB53A2a51f6Ed4B15B7822De4aB678E5Fe554A5Ae` | CREATE2 only |
| Config / Giver / Taker PositionManagers | `0x94911F…D611` / `0x161a9B…3f54` / `0xd871eE…e16c` | CREATE2 only |
| AaveOracle (CASH_SPOKE) | `0xf71F96C9570459af50519532b1503cB19Af5acb1` | deployer nonce 3 + exact sequence |
| CASH_SPOKE (proxy) | `0x208fAF5F20abb9E23A8E004CC813415C54448D8e` | deployer nonce 3 + exact sequence |
| CASH_SPOKE implementation | `0xD5c34Cab0D9db0Ec68881F42BdF9cD4D5e6A24C8` | deployer nonce 3 + exact sequence |
| **Launch payload (phase 1, dormant config)** | `0x74F38848f551f9ec237BfDf667db408e18Da07f4` | CREATE2 only |
| **Activation payload (phase 2)** | `0x5F64dE77e63E9CDfF87d818FA373c8593b0b20f3` | CREATE2 only |

Deployment salt: `0xf74ad75659c1f58e8b093c24f4bf439b7961aaaa19e6abc1bc12d2e55d745319` (derived). Estimated instance deployment: ~61M gas on OP.

## Fork dress rehearsal result (two-phase, hardcoded payloads)

Instance deploy → both payloads deployed at the predicted addresses → Owner Safe (`0x082B85ED50F1cd120C597EF860ece712e54CE844`) delegatecall-executes phase 1 → **dormant state verified field by field** (all 19 reserves with final CF / liquidation bonus / liquidation fee, caps, curves, liquidation engine 1.24 / 0.90, operator roles 201/401, every spoke `active: false`) → phase 2 activation executes → **live state verified**.

The full launch specification is committed at [`scripts/etherfi/LAUNCH_SPEC.md`](https://github.com/etherfi-protocol/aave-v4/blob/etherfi/cash-op-launch/scripts/etherfi/LAUNCH_SPEC.md) — this is the review artifact.

We're ready to broadcast as soon as the asks in the PR description are answered (deployment blessing, canonical commit, review slot).

scripts/etherfi/DeployEtherfiCashConfigEngine.s.sol
```diff
@@ -0,0 +1,28 @@
...
+/// @notice Deploys the stateless AaveV4ConfigEngine on OP Mainnet via the Safe Singleton
+/// Factory (CREATE2, fixed salt) — the address is fully deterministic and independent of the
+/// deployer. The engine holds no state or permissions; payloads delegatecall into it.
+///   forge script scripts/etherfi/DeployEtherfiCashConfigEngine.s.sol --rpc-url optimism \
+///     --account <keystore> --broadcast --verify
+contract DeployEtherfiCashConfigEngineScript is Script {
+  bytes32 internal constant SALT = keccak256('ETHERFI_CASH_AAVE_V4_CONFIG_ENGINE_V1');
+
+  function run() external returns (address engine) {
+    require(block.chainid == 10, 'expected OP Mainnet (chainid 10)');
```
Kogaroshi: some scripts use `WrongChain()` instead of error string, could be consistent and chose only 1 require style

scripts/etherfi/GenerateEtherfiCashLaunchSpec.s.sol
```diff
@@ -0,0 +1,146 @@
...
+import {EtherfiCashLaunchPayload} from 'src/etherfi/EtherfiCashLaunchPayload.sol';
+import {EtherfiCashOpMainnet} from 'src/etherfi/EtherfiCashOpMainnet.sol';
+import {DeployEtherfiCashLaunchPayloadScript} from 'scripts/etherfi/DeployEtherfiCashLaunchPayload.s.sol';
+
+/// @title GenerateEtherfiCashLaunchSpec
+/// @notice Generates the launch specification document from the payload's own specs, so the
+/// document can never drift from what the contract actually does. Writes
+/// output/etherfi/launch-spec.md. Optional env: PAYLOAD (deployed address).
+///   forge script scripts/etherfi/GenerateEtherfiCashLaunchSpec.s.sol --sig 'generate()' --rpc-url optimism
+contract GenerateEtherfiCashLaunchSpecScript is DeployEtherfiCashLaunchPayloadScript {
```
Kogaroshi: instead of inheriting `DeployEtherfiCashLaunchPayloadScript` here and in Verify script, there could be a common base for all `EtherfiCashLaunchScript` that has `_instanceAddresses()` &  `_assetAddresses()` (the 2 only reused methods in both script inheriting the Deploy script it seems)

src/etherfi/EtherfiCashLaunchPayload.sol
```diff
@@ -0,0 +1,574 @@
...
+/// @title EtherfiCashLaunchPayload
+/// @author ether.fi
+/// - Discussion: https://governance.aave.com/t/arfc-deploy-a-dedicated-aave-v4-whitelabel-instance-fully-managed-by-etherfi-on-op-mainnet-to-power-ether-fi-cash/25314
+/// @notice Launch payload for the ether.fi Cash Aave V4 instance on OP Mainnet (whitelabel:
+/// executed by the OWNER Safe via a delegatecall Safe transaction — no Aave Governance V3).
+/// The executing Safe must hold the AccessManager admin role (0) and the configurator domain
+/// admin roles (200/400), which the instance deployment grants.
+///
+/// Executes, in the AaveV4Payload fixed order:
+///   1. AccessManager actions — labels the two operator roles, reassigns the cap /
```
Kogaroshi: order of that step inconsistant with actual implementation

src/etherfi/EtherfiCashOpMainnet.sol
```diff
@@ -0,0 +1,130 @@
+// SPDX-License-Identifier: MIT
```
Kogaroshi: This should live in the address-book, but nbd

src/etherfi/EtherfiCashLaunchPayload.sol
```diff
@@ -0,0 +1,574 @@
...
+  address public immutable SETHFI;
+  address public immutable OP;
+  address public immutable WHYPE;
+  address public immutable BEHYPE;
+  address public immutable LIQUID_ETH;
+  address public immutable LIQUID_BTC;
+  address public immutable LIQUID_USD;
+  address public immutable LIQUID_RESERVE;
+  address public immutable WEEUR;
+  address public immutable LIQUID_RWA;
```
Kogaroshi: why not importing from `EtherfiCashOpMainnet` like for the Caps ? Same for the price feeds ?

scripts/etherfi/DeployEtherfiCashConfigEngine.s.sol
seongyun-ko: Good catch — standardized on custom errors everywhere in a7e6505b: `WrongChain`, `MissingInstanceAddress` and `NoCodeAtAddress` now live once in the new shared base and all string requires are gone.

scripts/etherfi/GenerateEtherfiCashLaunchSpec.s.sol
seongyun-ko: Agreed, that inheritance was lazy. Added `EtherfiCashScriptBase` in a7e6505b holding `_instanceAddresses()` / `_assetAddresses()`, the chain guard and the shared errors — deploy, verify, safe-tx, launch-spec, validate and config-engine scripts all extend it now; no read-only script inherits the broadcasting deploy script anymore.

src/etherfi/EtherfiCashLaunchPayload.sol
seongyun-ko: Fixed in a7e6505b — the header now describes the engine's actual sub-order (role grants → labels → selector reassignment). Comment-only change, so the payload bytecode and the predicted addresses in the PR comment are unchanged.

src/etherfi/EtherfiCashOpMainnet.sol
seongyun-ko: Done in d7cabfd6 — restructured into aave-address-book format: `AaveV4EtherfiCash` / `AaveV4EtherfiCashHubs` / `AaveV4EtherfiCashSpokes` / `AaveV4EtherfiCashAssets` (`*_UNDERLYING`/`*_ORACLE`/`*_DECIMALS`), same shape as `AaveV4Avalanche.sol`, so it can be upstreamed verbatim post-deployment. Launch caps live in a separate `AaveV4EtherfiCashCaps` library since they're launch config rather than address-book material.

src/etherfi/EtherfiCashLaunchPayload.sol
seongyun-ko: You're right — done in d7cabfd6. Both payloads now import every address from the `AaveV4EtherfiCash` libraries like the caps: zero constructor args, all action functions `pure`, matching the production ConfigPayload pattern. Unit tests were reworked to assert the generated actions/parameters directly against the final parameter sheet, and execution coverage moved to the fork dress rehearsal (which re-passed end to end). Nice side effect: both payload addresses are now pure-CREATE2 deterministic — updated in the address table above.

scripts/etherfi/DeployEtherfiCashInstance.s.sol
```diff
@@ -3,7 +3,8 @@ pragma solidity ^0.8.0;
 
 import {AaveV4DeployBatchBaseScript} from 'scripts/deploy/AaveV4DeployBatchBase.s.sol';
 import {InputUtils} from 'src/deployments/utils/libraries/InputUtils.sol';
-import {EtherfiCashOpMainnet} from 'src/etherfi/EtherfiCashOpMainnet.sol';
+import {AaveV4EtherfiCash} from 'src/etherfi/AaveV4EtherfiCash.sol';
+import {AaveV4EtherfiCashAssets} from 'src/etherfi/AaveV4EtherfiCash.sol';
```
Kogaroshi: `import {AaveV4EtherfiCash, AaveV4EtherfiCashAssets} from 'src/etherfi/AaveV4EtherfiCash.sol';`

seongyun-ko: Fixed in e89fa49a — merged into a single import (leftover from a mechanical rename; the other files were already combined).

scripts/etherfi/DeployEtherfiCashInstance.s.sol
```diff
@@ -24,6 +28,42 @@ import {AaveV4EtherfiCash, AaveV4EtherfiCashAssets} from 'src/etherfi/AaveV4Ethe
...
+    logger.log('deployer', deployer);
+    logger.logHeader1('starting Aave V4 batch deployment');
+
+    OrchestrationReports.FullDeploymentReport memory report = AaveV4DeployOrchestration
+      .deployAaveV4(
+        logger,
+        deployer,
+        inputs,
+        BytecodeHelper.getHubBytecode(),
+        vm.getCode('src/etherfi/EtherFiSpokeInstance.sol:EtherFiSpokeInstance')
```
Kogaroshi: pls add a method in `BytecodeHelper` for consistency

scripts/etherfi/LAUNCH_SPEC.md
```diff
@@ -56,15 +56,15 @@ Operator roles carved out by the payload: HUB_CAPS_OPERATOR_ROLE (201) for updat
...
+| Launch payload (phase 1, dormant config) | 0x7de5c5B3595F2Cc1b2f4C446773d3597374AbDa4 |
 | Activation payload (phase 2) | 0x5F64dE77e63E9CDfF87d818FA373c8593b0b20f3 |
 | Owner Safe | 0x082B85ED50F1cd120C597EF860ece712e54CE844 |
 | Operator Safe | 0x23c30c38d73a0D1609ffAAe47aA7d6D1a3e46f03 |
 | AccessManager | 0x188d7173772499FB6375F23FdFd130CE6107286b |
 | Config Engine | 0x84210b3087E952Be0f3610fD75f0f045995eAF22 |
 | Hub | 0x66753c4e3fC84f1eD0e3C267C927284E9d90C572 |
 | Hub Configurator | 0xA39bEf2fD611fb9c5a69D63277b4Af97a30F0dbC |
-| Cash Spoke | 0x208fAF5F20abb9E23A8E004CC813415C54448D8e |
+| Cash Spoke | 0x6eAb1dC9eA3E6557dCE44B52c340a514a5ed5b83 |
```
Kogaroshi: can you confirm which LiquidationLogic library address this is deployed with ?

src/etherfi/EtherFiSpokeInstance.sol
```diff
@@ -0,0 +1,56 @@
...
+    uint256 reserveId,
+    uint256 amount,
+    address onBehalfOf
+  ) public override returns (uint256, uint256) {
+    require(
+      IEtherFiDataProvider(ETHERFI_DATA_PROVIDER).isEtherFiSafe(onBehalfOf),
+      OnlyEtherFiSafe(onBehalfOf)
+    );
+    return super.borrow(reserveId, amount, onBehalfOf);
+  }
```
Kogaroshi: The TakerPositionManager could shortcut that restriction using "credit delegation", pls make sure not to add and list that PositionManager, or a degraded version (with only the withdraw side).

src/etherfi/EtherFiSpokeInstance.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: LicenseRef-BUSL
+pragma solidity 0.8.28;
+
+import {SpokeInstance} from 'src/spoke/instances/SpokeInstance.sol';
+
+/// @dev The ether.fi data provider used to recognize Cash Safes (a proxy; address is stable).
+interface IEtherFiDataProvider {
+  function isEtherFiSafe(address account) external view returns (bool);
+}
```
Kogaroshi: this should be in its own interface file

tests/etherfi/EtherFiSpokeGate.t.sol
```diff
@@ -0,0 +1,45 @@
...
+
+import {Test} from 'forge-std/Test.sol';
+
+import {EtherFiSpokeInstance} from 'src/etherfi/EtherFiSpokeInstance.sol';
+
+contract MockOracle {
+  function decimals() external pure returns (uint8) {
+    return 8;
+  }
+}
```
Kogaroshi: pls move it to its own file in the mocks dir

tests/etherfi/EtherFiSpokeGate.t.sol
```diff
@@ -0,0 +1,45 @@
...
+
+/// @dev Verifies the borrow gate against the LIVE EtherFiDataProvider on an OP Mainnet fork: a
+/// non-safe onBehalfOf is rejected with OnlyEtherFiSafe BEFORE any Spoke logic runs; a recognized
+/// safe passes the gate and proceeds into the parent borrow (which then reverts Unauthorized for
+/// lack of a position manager — proving the gate is pass-through, not a wall). Skips itself unless
+/// running against chainid 10:
+///   forge test --match-path tests/etherfi/EtherFiSpokeGate.t.sol --fork-url <op-rpc> -vv
+contract EtherFiSpokeGateTest is Test {
+  function test_fork_borrowGate() public {
```
Kogaroshi, line 175: Is this the only test for that new restriction ? There should be fuzzing here, and some other scenario tested.
The current `SpokeInstance` test suite should also be run again for `EtherFiSpokeInstance`, with the expected restricted scenarios on borrow., to ensure the rest of the Spoke behavior was not impacted.

tests/etherfi/EtherFiSpokeGate.t.sol
```diff
@@ -0,0 +1,45 @@
...
+    vm.expectRevert(abi.encodeWithSelector(EtherFiSpokeInstance.OnlyEtherFiSafe.selector, notASafe));
+    spoke.borrow(0, 1, notASafe);
+
+    // data provider says yes -> gate passes, parent borrow takes over (Unauthorized: no position manager)
+    vm.mockCall(
+      spoke.ETHERFI_DATA_PROVIDER(),
+      abi.encodeWithSignature('isEtherFiSafe(address)', fakeSafe),
+      abi.encode(true)
+    );
+    vm.expectRevert(); // parent's Unauthorized (onlyPositionManager), NOT OnlyEtherFiSafe
```
Kogaroshi: please add the correct expected revert Error here

src/etherfi/EtherFiSpokeInstance.sol
```diff
@@ -0,0 +1,56 @@
...
+  error OnlyEtherFiSafe(address account);
+
+  constructor(
+    address oracle_,
+    uint16 maxUserReservesLimit_
+  ) SpokeInstance(oracle_, maxUserReservesLimit_) {}
+
+  /// @notice Borrows `amount` of reserve `reserveId` for `onBehalfOf`, which must be an ether.fi
+  ///         Cash Safe.
+  function borrow(
```
Kogaroshi, line 59: either import the full natspec, or inherit and put the new change under a "@dev" comment

scripts/etherfi/VerifyEtherfiCashLive.s.sol
```diff
@@ -0,0 +1,150 @@
...
+/// @title VerifyEtherfiCashLive
+/// @notice Read-only check that the LIVE hub/spoke state matches the launch payload spec —
+/// every listing, cap, collateral factor, bonus, curve and the liquidation config.
+/// Run after the payload executes (fork rehearsal or the real post-AIP state):
+///   forge script scripts/etherfi/VerifyEtherfiCashLive.s.sol --rpc-url <optimism|fork>
+/// Reverts with a mismatch count if anything differs. Never broadcasts.
+contract VerifyEtherfiCashLiveScript is EtherfiCashScriptBase {
+  uint256 internal mismatches;
+
+  function verify() external returns (uint256) {
```
Kogaroshi, line 39: this should now also verify that the correct SpokeInstance implementation was used (and that `ETHERFI_DATA_PROVIDER()` returns the expected Provider)

scripts/etherfi/DeployEtherfiCashInstance.s.sol
```diff
@@ -0,0 +1,112 @@
...
+      );
+    vm.stopBroadcast();
+    logger.writeJsonReportMarket(report);
+    _logDeploySummary(logger);
+    logger.logHeader1('batch deployment completed');
+    logger.logHeader1('saving logs');
+    logger.save({fileName: _outputFileName, withTimestamp: true});
+  }
```
Kogaroshi, line 154: a check at the end of this run() to ensure all addresses in the Deploy Report are matching the expected AaveV4EtherfiCash.sol registry would prevent any misconfiguration not being caught early enough

scripts/etherfi/DeployEtherfiCashInstance.s.sol
```diff
@@ -0,0 +1,112 @@
...
+    address ownerSafe = AaveV4EtherfiCash.OWNER_SAFE;
+
+    string[] memory hubLabels = new string[](1);
+    hubLabels[0] = 'CASH_HUB';
+
+    string[] memory spokeLabels = new string[](1);
+    spokeLabels[0] = 'CASH_SPOKE';
+
+    uint16[] memory spokeMaxReservesLimits = new uint16[](1);
+    spokeMaxReservesLimits[0] = 64; // 19 launch reserves + room for stocks and future listings
```
Kogaroshi: consider using constants for those values

tests/etherfi/EtherFiSpokeGate.t.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: MIT
```
Kogaroshi, line 1: linter seems to fail on this file, and some other are raising issue, please run the prettier

scripts/etherfi/launch.sh
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env bash
+# ether.fi Cash Aave V4 launch pipeline (OP Mainnet, whitelabel — Safe-administered).
+#
+# Usage:
+#   scripts/etherfi/launch.sh rehearse                       # validate + fork dress rehearsal
+#   account=<keystore> scripts/etherfi/launch.sh broadcast   # the real thing (asks for confirmation)
+#
+# Stages:
+#   1. validate   read-only preflight of every address in EtherfiCashOpMainnet.sol
```
Kogaroshi: pls update, this file does not exist anymore

foundry.toml
```diff
@@ -27,6 +27,7 @@ additional_compiler_profiles = [
 compilation_restrictions = [
   { paths = "src/hub/instances/HubInstance.sol", via_ir = true, optimizer_runs = 22_300 },
   { paths = "src/spoke/instances/SpokeInstance.sol", via_ir = true, optimizer_runs = 750 },
+  { paths = "src/etherfi/EtherFiSpokeInstance.sol", via_ir = true, optimizer_runs = 750 },
```
Kogaroshi, line 30: Nit: this new contract seems right on the limit with this compiler settings, feel free to introduce a new profile and reduce the optimizer runs to free a bigger codesize buffer

src/etherfi/EtherFiSpokeInstance.sol
Kogaroshi: alternatively, verifying `msg.sender == onBehalfOf` as well here would prevent such shortcuts

scripts/etherfi/LAUNCH_SPEC.md
shivam-ef: LiquidationLogic at 0x818E84198224535FAeaEc1b583d3Ff6b812A5AF3

src/etherfi/EtherFiSpokeInstance.sol
shivam-ef: Makes sense, thanks for the suggestion

scripts/etherfi/LAUNCH_SPEC.md
Kogaroshi: Please use https://etherscan.io/address/0x88dF535473C5adf1f57789734A05E555F7Deb8DB#code instead

Kogaroshi: via the `FOUNDRY_LIBRARIES=src/spoke/libraries/LiquidationLogic.sol:LiquidationLogic:0x88dF535473C5adf1f57789734A05E555F7Deb8DB` env variable

scripts/etherfi/DeployEtherfiCashInstance.s.sol
seongyun-ko: Done in 3a7c8f9d — run() now asserts every report address against the AaveV4EtherfiCash registry after the report JSON is saved (custom ReportMismatch(name, actual, expected) so the mismatch is inspectable). It immediately proved its worth: it caught that the previously pinned CASH_SPOKE 0x6eAb…5b83 does not reproduce from a clean fork at deployer nonce 3 — the correct predictions are proxy 0xc5645c9956c088847616012855C786036A280a5C / implementation 0xB76c5cA6958eCbdA4050c72279015B1d5a07A5eE, now re-pinned (payload addresses updated accordingly in LAUNCH_SPEC).

scripts/etherfi/VerifyEtherfiCashLive.s.sol
seongyun-ko: Done in 3a7c8f9d — the verify script now reads the Cash Spoke proxy's EIP-1967 implementation slot and asserts it equals the pinned CASH_SPOKE_IMPLEMENTATION (new registry entry), and calls ETHERFI_DATA_PROVIDER() through the proxy asserting the expected provider address — a stock SpokeInstance deployment fails both checks. Covered in the fork rehearsal, which passes end to end.

scripts/etherfi/launch.sh
seongyun-ko: Fixed in 3a7c8f9d — both stale references (the stage description and the NOT-READY hint) now point at AaveV4EtherfiCash.sol.

scripts/etherfi/LAUNCH_SPEC.md
seongyun-ko: we'll link against the canonical LiquidationLogic 0x88dF535473C5adf1f57789734A05E555F7Deb8DB via FOUNDRY_LIBRARIES and drop our own predeploy. One question so we can pin addresses: 0x88dF… doesn't have code on OP Mainnet yet (only Ethereum) — will you deploy it there, or can you share the factory + salt so our deployer can reproduce the address? Since relinking changes the spoke implementation bytecode (and thus the spoke/payload predictions), we'll bundle this with the gate/test changes and re-pin everything in one pass.

seongyun-ko: LAUNCH_SPEC corrected in b6c15ac2: it now states the current predictions were simulated with the self-deployed library and that the deployment will link the canonical 0x88dF535473C5adf1f57789734A05E555F7Deb8DB instead, with the re-pin bundled into the gate/test pass.

src/etherfi/EtherFiSpokeInstance.sol
seongyun-ko: Resolved in f5882c12 (merged as c7de1bdc) — the stock position managers are not deployed at all with this instance (deployPositionManagers: false), and the deploy script's registry check asserts the Taker/Giver/Config position managers and SignatureGateway do NOT exist. With zero position managers, Spoke._isPositionManager only passes for msg.sender == onBehalfOf, so this is the structural form of your alternative fix as well: borrow proceeds can only ever be paid to the safe itself. The invariant (cash-v3 LendGateway as the only PM that may ever be activated) is documented in EtherFiSpokeInstance.

tests/etherfi/EtherFiSpokeGate.t.sol
seongyun-ko: Resolved in f5882c12 (merged as c7de1bdc) — EtherFiSpokeGate.t.sol now has fuzz coverage (gate-reverts for any non-safe onBehalfOf across arbitrary reserveId/amount; gate-passthrough reaching the parent auth for arbitrary callers) plus scenario tests (collateral does not open the gate; an approved position manager cannot bypass it; safe-loses-status keeps repay/withdraw working). EtherFiSpokeSuiteRerun.t.sol re-runs the full stock Spoke suite against EtherFiSpokeInstance via a virtual _spokeBytecode() hook in Base.t.sol with the gate transparent — 640/640 pass, gas snapshots unchanged.

tests/etherfi/EtherFiSpokeGate.t.sol
seongyun-ko: Resolved in f5882c12 — the passthrough assertions now expect the specific errors (OnlyEtherFiSafe with the offending account, and the parent's Unauthorized on the position-manager check).

src/etherfi/EtherFiSpokeInstance.sol
seongyun-ko: Resolved in f5882c12 — IEtherFiDataProvider and IEtherFiSpokeInstance now live in src/etherfi/interfaces/.

tests/etherfi/EtherFiSpokeGate.t.sol
seongyun-ko: Resolved in f5882c12 — MockEtherFiDataProvider and MockOracle moved to tests/helpers/mocks/.

src/etherfi/EtherFiSpokeInstance.sol
seongyun-ko: Resolved in f5882c12 — the override now uses @inheritdoc ISpoke with the restriction documented under @dev.

scripts/etherfi/DeployEtherfiCashInstance.s.sol
seongyun-ko: Resolved in f5882c12 — BytecodeHelper.getEtherFiSpokeBytecode(), used by the deploy script.

scripts/etherfi/DeployEtherfiCashInstance.s.sol
seongyun-ko: Done in 5455c028 — hub/spoke labels and the max-reserves limit hoisted to named constants.

tests/etherfi/EtherFiSpokeGate.t.sol
seongyun-ko: Done — prettier now passes across src/etherfi, scripts/etherfi and tests/etherfi (f5882c12 reformatted the new files; 5455c028 fixed the one remaining verify-script warning).

scripts/etherfi/LAUNCH_SPEC.md
seongyun-ko: Implemented in etherfi-protocol/aave-v4#5 — the spoke now links against the canonical 0x88dF535473C5adf1f57789734A05E555F7Deb8DB, with the FOUNDRY_LIBRARIES pin in committed tooling (make targets + launch.sh) rather than the machine-local .env, and the preflight validator gates deployment on the library having code on OP Mainnet. Working assumption per the OP/Aave side: it will be deployed there at the same address as Ethereum mainnet — please flag if that's wrong. Final predicted addresses re-pinned accordingly (spoke 0xdffcC3536D932eb51Df51a7F5FA407c4270d5308, impl 0xA1f75D801633a1941cae6670352d627884dC3b68, phase-1 payload 0xAEC828886f4259a24c2E284907151e448B35e794).

foundry.toml
seongyun-ko: Closing this out in etherfi-protocol/aave-v4#5 — measured after the canonical relink: runtime 24,556 bytes, 20-byte EIP-170 margin at the shared spoke profile. We kept your suggestion's intent but not the mechanism: a dedicated lower-runs profile would silently recompile the stock SpokeInstance (shared compilation unit), so instead there's now a size-gate test that logs the margin on every run and fails below a 10-byte alarm threshold. If the margin gets consumed by an upstream change, the fallback is restructuring the instance as a compilation-unit sibling (tested, gains ~350 bytes) — deferred since it's a contract change requiring re-review.

scripts/etherfi/LAUNCH_SPEC.md
Kogaroshi: Ah yes indeed, mb here, this is the canonical lib on Eth mainnet, not OP.
Please then do the pre-deploy & verification of the lib on OP, and add it in the specs.

scripts/etherfi/DeployEtherfiCashInstance.s.sol
```diff
@@ -54,14 +64,92 @@ contract DeployEtherfiCashInstanceScript is AaveV4DeployBatchBaseScript {
...
     logger.logHeader1('batch deployment completed');
     logger.logHeader1('saving logs');
     logger.save({fileName: _outputFileName, withTimestamp: true});
+
+    // after the report is persisted: every deployed address must match the pinned registry,
+    // so a misconfiguration is caught here rather than at payload execution
+    _assertReportMatchesRegistry(report);
+  }
+
+  error ReportMismatch(string name, address actual, address expected);
```
Kogaroshi: pls move it at the top of contract, for consistency with the other files in repo

shivam-ef: fixed in [932e768](https://github.com/aave/aave-v4/pull/1325/commits/932e768c41ecf44cfd291c1c99dab29c11bedeb0)

scripts/etherfi/LAUNCH_SPEC.md
shivam-ef: deployed and verified: https://optimistic.etherscan.io/address/0x88dF535473C5adf1f57789734A05E555F7Deb8DB

Also added it to the launch spec in commit [d86feca](https://github.com/aave/aave-v4/pull/1325/commits/d86feca16a20dc256df7bfa8f25553bb9ba3c0d7)

src/etherfi/EtherFiSpokeInstance.sol
```diff
@@ -0,0 +1,70 @@
...
+///      on OP Mainnet) rather than a constructor argument, so this contract keeps the exact
+///      `(oracle, maxUserReservesLimit)` constructor shape the Aave deployment framework appends
+///      to the spoke bytecode — the instance deployment pipeline needs no framework changes. Code,
+///      not storage: zero storage-layout risk across Aave upgrades.
+///
+///      This file is maintained verbatim in both etherfi-protocol/cash-v3 (the audited copy) and
+///      the aave-v4 launch branch (the deployed copy); only the SpokeInstance import path differs
+///      per repo. Keep them in sync.
+contract EtherFiSpokeInstance is SpokeInstance {
+  /// @notice The ether.fi data provider used to recognize Cash Safes (prod proxy, OP Mainnet).
```
miguelmtzinf, line 46: makes sense to inherit `IEtherFiSpokeInstance`, and its natspec docs here

seongyun-ko:
Thanks for the thorough review (21 threads, all resolved — including the canonical LiquidationLogic adoption and the final address re-pin in etherfi-protocol/aave-v4#5). Per the agreement with the Aave eng team, we're moving the launch work into our fork as the main repo going forward: the equivalent PR is etherfi-protocol/aave-v4#7, and this repo's branch will keep tracking it until launch. Closing this review PR; we'll share the deployed payload addresses and the post-deployment state (make etherfi-verify output + LAUNCH_SPEC) for your review as agreed, and we'd still appreciate the aave-address-book entry once the instance is live. 🙏

Kogaroshi:
Verified the deployed LiquidationLogic library deployed on Optimism at `0x88dF535473C5adf1f57789734A05E555F7Deb8DB` is the exact same bytecode as the Canonical one deployed at the same address on Ethereum and Avalanche network

src/etherfi/EtherfiCashLaunchPayload.sol
```diff
@@ -0,0 +1,532 @@
...
+
+  // ------------------------------- liquidation engine (global) --------------------------------
+  uint256 public constant TARGET_HEALTH_FACTOR = 1.24e18; // WAD
+  uint256 public constant HEALTH_FACTOR_FOR_MAX_BONUS = 0.9e18; // WAD
+
+  // ------------------------------- operator roles (Nonce curator) -----------------------------
+  // Granular roles carved out of the configurator domain-admin roles, following the Roles.sol
+  // evolution rules (next free IDs; domain admins are granted the new roles to retain access).
+  uint64 public constant HUB_CAPS_OPERATOR_ROLE = 201; // updateSpokeCaps / AddCap / DrawCap
+  uint64 public constant SPOKE_RISK_OPERATOR_ROLE = 401; // add/updateDynamicReserveConfig
```
miguelmtzinf, line 85: I d suggest moving to `Roles` lib

src/etherfi/EtherfiCashLaunchPayload.sol
```diff
@@ -0,0 +1,532 @@
...
+  // ------------------------------- shared reserve parameters ----------------------------------
+  uint16 internal constant LIQUIDATION_FEE = 10_00; // 10% for every reserve
+  uint24 internal constant COLLATERAL_RISK = 0; // risk premium unused at launch
+
+  // maxLiquidationBonus encoding: 100_00 == 0.00% bonus
+  uint32 internal constant BONUS_1_0 = 101_00;
+  uint32 internal constant BONUS_2_0 = 102_00;
+  uint32 internal constant BONUS_3_5 = 103_50;
+  uint32 internal constant BONUS_4_0 = 104_00;
+  uint32 internal constant BONUS_5_0 = 105_00;
```
miguelmtzinf, line 96: do not think it's useful... i d suggest using raw values instead

PR #1333, opened 2026-08-10
feat: spoke gate + admin position transfers

avniculae:
Adds the two spoke-level primitives needed for permissioned instances (EtherFi Cash whitelabel, Horizon-style RWA): an optional access gate on position actions, and admin position transfers.

## Spoke gate

`updateGate(address)` sets an optional gate contract on the Spoke. When set, `supply`, `withdraw`, `borrow`, `repay` and `setUsingAsCollateral` call `ISpokeGate.isActionAllowed(selector, caller, onBehalfOf, reserveId, amount)` (view) and revert with `GateAccessDenied` unless it allows the action. The default `address(0)` short-circuits: permissionless spokes pay one SLOAD per action and never make the external call.

This is the `feat/add-spoke-hooks` shape with three changes:
- the hook is view-only and receives the full action context including `reserveId`/`amount` (taken from `feat/permissioned-borrow-hook`), so policies like per-user borrow limits are expressible without letting third-party code mutate state inside user actions;
- the gate is a storage address behind a restricted setter instead of a constructor immutable, so policy can be changed without an implementation upgrade;
- `liquidationCall` is not gated (see PR comment).

Why this over the other explored options: the EtherFi-style `virtual`/`override` and the compile-time `_isBorrowAllowed` hook require a diverging SpokeInstance per licensee, while the gate keeps every instance on the same audited artifact with policy in a small external contract (EtherFi becomes a ~15-line gate checking `isEtherFiSafe(onBehalfOf)` on borrow). Mandatory position managers break direct self-calls per gated selector and turn the PM into a funds-forwarding contract (`borrow` pays the caller). A custom AccessManager `canCall` cannot see `onBehalfOf`, which is the address that actually needs gating.

## transferPosition

`transferPosition(reserveId, from, to, shares)` moves supplied shares between positions — the v4 equivalent of Horizon's `RwaAToken.authorizedTransfer` (issuer-forced transfers: lost keys, sanctioned users). Semantics mirror v3: the health factor of `from` is validated after the move (with risk premium refresh), transfers are allowed while frozen but blocked while paused, `shares > balance` signals a full transfer, and the recipient's collateral flag is not auto-enabled. Doing this natively instead of via a global position manager keeps the underlying inside the Hub (no withdraw/re-supply round-trip, no allowlisting of the manager on the RWA token, single event for the issuer's transfer agent) and preserves exact share amounts.

Access control is a new `SPOKE_POSITION_TRANSFER_ADMIN_ROLE` (303). The roles procedure wires the selector, but the role is granted to no one by default — an instance without an explicit grant has no forced transfers. Per-reserve delegation to issuers (v3's `RwaATokenManager` analog) can be layered on as a periphery contract holding the role.

## Spoke size

`SpokeInstance` runtime was at 24,291/24,576 bytes (285 margin) and the two features add ~900. To fit, this PR removes five getters whose data is available elsewhere, freeing ~695 bytes:

- `getReserveSuppliedAssets` / `getReserveSuppliedShares` — pure passthroughs of `hub.getSpokeAddedAssets/Shares(assetId, spoke)`
- `getReserveTotalDebt` — sum of the `getReserveDebt` pair
- `getLiquidationBonus` — computable off-chain from `getLiquidationConfig` + the user's dynamic config via `LiquidationLogic.calculateLiquidationBonus` (public)
- `getUserLastRiskPremium` — readable via ExtSload (tests now do this)

Final size: 24,496 bytes (80 margin). If any of these getters matter for integrators, say so — each is independently revertable, but then something else has to give (the externalized-functions refactor would be the structural fix).

Test suite updated accordingly; new suites in `Spoke.Gate.t.sol` and `Spoke.TransferPosition.t.sol`.


avniculae:
The gate is deliberately not enforced in `liquidationCall`: we don't want a gate to be able to block liquidations. Horizon-style instances don't need it either — with `receiveSharesEnabled = false` the collateral leaves as an underlying transfer, so the RWA token's own allowlist already constrains who can effectively liquidate. Open question whether we want the flexibility anyway: a permissioned spoke for other use-cases might legitimately need to restrict liquidators at the spoke level. If so, I'd add it later as an explicit opt-in (e.g. a separate flag) rather than gating liquidations by default.

avniculae:
Superseded by #1334 — same goals via a mandatory position manager on a separate PermissionedSpokeInstance, leaving the standard SpokeInstance byte-identical (no gate calls in entry points, no transferPosition, no getter removals).

PR #1334, opened 2026-08-11
feat: permissioned spoke via gate

avniculae:
Adds `PermissionedSpokeInstance`, a spoke variant where a gate contract decides authorization on position actions and liquidations. This enables permissioned borrowing (e.g. the EtherFi Cash instance) and Horizon-style RWA spokes without forking the core.

The `onlyAllowed` modifier in `Spoke` calls a virtual `_isAllowed(caller, user, data)`. The default allows anyone to liquidate and requires an approved position manager otherwise, so canonical spokes behave as before, at +181 gas per position action and +231 on liquidations. `PermissionedSpoke` overrides the hook to call `ISpokeGate.isCallAllowed`, with the gate an immutable set at construction: no storage, no setter. A gate can restrict (only eligible users borrow, only permitted liquidators liquidate), extend (an RWA manager moves any user's position via withdraw and re-supply, with HF checks intact), or preserve the default behavior by allowing `liquidationCall` and calling back `ISpoke(msg.sender).isPositionManager`. Note the callback alone is not enough: a gate that wants permissionless liquidations must allow the `liquidationCall` selector itself. Replacing a gate is an implementation upgrade; policy changes can stay mutable inside the gate.

`PermissionedSpokeInstance` inherits `SpokeInstance` and `PermissionedSpoke` and fits the standard compiler profile (750 runs) at 24,557 bytes, 19 bytes under the size limit. Reference policy gates and their gas snapshots are in `PermissionedSpoke.Operations.json`.

Supersedes #1333.


src/spoke/interfaces/ISpokeGate.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: LicenseRef-BUSL
```
Kogaroshi, line 1: shouldn't this be in position-manager/ ? Or named differently to avoir confusion ?

foundry.toml
```diff
@@ -22,11 +22,13 @@ dynamic_test_linking = true
 additional_compiler_profiles = [
   { name = "hub", optimizer = true, via_ir = true, optimizer_runs = 22_300 },
   { name = "spoke", optimizer = true, via_ir = true, optimizer_runs = 750 },
+  { name = "permissioned-spoke", optimizer = true, via_ir = true, optimizer_runs = 200 },
```
Kogaroshi: are the run needed to be so low ?

src/spoke/interfaces/ISpokeGate.sol
AlbertoCentonze: I think it should be named differently but we haven't settled on a name for now.

src/spoke/interfaces/IMandatoryPositionManager.sol
```diff
@@ -0,0 +1,20 @@
...
+  /// @param caller The transaction initiator on the Spoke.
+  /// @param onBehalfOf The owner of the position being modified.
+  /// @param data The full calldata of the Spoke call, allowing per-action decoding.
+  /// @return True if the call is allowed.
+  function isCallAllowed(
+    address caller,
+    address onBehalfOf,
+    bytes calldata data
+  ) external view returns (bool);
+}
```
AlbertoCentonze: I can't get myself to like the redundancy between a mandatory position manager's `isCallAllowed` and access manager's `canCall`. If we have to go for this approach we might as well merge the two concepts and create our own abstraction that is flexible enough for both use cases?

src/spoke/interfaces/ISpokeGate.sol
Kogaroshi: we have Gateways for signatures and native asset interactions, so maybe Gatekeeper ?

src/spoke/instances/PermissionedSpokeInstance.sol
AlbertoCentonze, line 1: Isn't the mandatory position manager an abstraction that would be useful for non permissioned spokes too? Genuine question.

src/spoke/interfaces/ISpokeGate.sol
AlbertoCentonze: Mandatory position managers can also allow to seize assets, in that case they would do the opposite of gatekeeping

src/spoke/instances/PermissionedSpokeInstance.sol
```diff
@@ -0,0 +1,61 @@
...
+  /// @dev When a mandatory position manager is set, it replaces the default authorization and fully
+  /// decides whether the call is allowed, based on the caller, the position owner and the calldata.
+  function _isAuthorizedPositionManagerCall(address user) internal view override returns (bool) {
+    address mandatoryPositionManager = _permissionedSpokeStorage().mandatoryPositionManager;
+    if (mandatoryPositionManager == address(0)) {
+      return super._isAuthorizedPositionManagerCall(user);
+    }
+    return
+      IMandatoryPositionManager(mandatoryPositionManager).isCallAllowed({
+        caller: msg.sender,
```
DhairyaSethi: don't use env data inside internal functions, should always be passed down (msg sender, data)

avniculae: fixed, the modifier now passes msg.sender and msg.data down

src/spoke/instances/PermissionedSpokeInstance.sol
```diff
@@ -0,0 +1,60 @@
...
+    address caller,
+    address user,
+    bytes calldata data
+  ) internal view override returns (bool) {
+    address gate = _permissionedSpokeStorage().gate;
+    if (gate == address(0)) {
+      return super._isAuthorizedPositionManagerCall(caller, user, data);
+    }
+    return ISpokeGate(gate).isCallAllowed({caller: caller, onBehalfOf: user, data: data});
+  }
```
AlbertoCentonze: if mandatory gate is set normal position managers can't be used anymore?

avniculae: it can, since the gate can call back into the spoke to see if posm is approved

avniculae: see `PositionManagerPolicyGate` in PolicyGats.sol in mocks

src/spoke/instances/PermissionedSpokeInstance.sol
```diff
@@ -0,0 +1,38 @@
...
+  /// @dev The gate fully decides whether the call is allowed, based on the caller, the position
+  /// owner and the calldata. It can preserve the default authorization by calling back
+  /// `isPositionManager`.
+  function _isAuthorizedPositionManagerCall(
+    address caller,
+    address user,
+    bytes calldata data
+  ) internal view override returns (bool) {
+    return ISpokeGate(GATE).isCallAllowed({caller: caller, onBehalfOf: user, data: data});
+  }
```
miguelmtzinf, line 29: unless we call super here under certain conditions, this kills the position manager check of the base spoke logic (this gate implementation is only calling back to the Spoke to access the posm config)

src/spoke/instances/PermissionedSpokeInstance.sol
```diff
@@ -0,0 +1,27 @@
...
+import {PermissionedSpoke} from 'src/spoke/PermissionedSpoke.sol';
+import {SpokeInstance} from 'src/spoke/instances/SpokeInstance.sol';
+
+/// @title PermissionedSpokeInstance
+/// @author Aave Labs
+/// @notice Implementation contract for the PermissionedSpoke.
+contract PermissionedSpokeInstance is SpokeInstance, PermissionedSpoke {
+  /// @dev Applies the gate authorization of the PermissionedSpoke.
+  modifier onlyPositionManager(address onBehalfOf) override(Spoke, PermissionedSpoke) {
+    _checkCallAllowed(msg.sender, onBehalfOf, msg.data);
```
avniculae: We cannot do super here bc this is a modifier. thoughts? @DhairyaSethi 

We need to override to resolve inhteritance overlap

DhairyaSethi: override the internal method?

avniculae: can't override `_isPositionManager` directly: it also backs the external `isPositionManager`, which gates call back into for the default authorization, so it would recurse into the gate. and the modifier didn't call anything else overridable. latest commit adds a dedicated virtual `_isAllowed` hook that the modifier calls, overridden in `PermissionedSpoke` instead

src/horizon/HorizonSpokeGate.sol
```diff
@@ -0,0 +1,56 @@
...
+import {ISpokeGate} from 'src/spoke/interfaces/ISpokeGate.sol';
+import {IHorizonSpokeGate} from 'src/horizon/interfaces/IHorizonSpokeGate.sol';
+
+/// @title HorizonSpokeGate
+/// @author Aave Labs
+/// @notice Gate for Horizon RWA spokes. Managers authorized for a reserve can supply, withdraw,
+/// repay and set collateral usage on the position of any user for that reserve (e.g. forced
+/// transfers via withdraw and re-supply, forced repayments), but never borrow. Liquidations are
+/// permissionless and all other calls follow the default position manager authorization.
+contract HorizonSpokeGate is Ownable, IHorizonSpokeGate {
```
Kogaroshi, line 15: Do we want to have this Gate upgradeable ? In case more role/enforcement is required on future Horizon updates ? 
While I see that we could also replace the Gate when upgrading the `PermissionedSpoke`, that would require to port and copy all the previously done configuration for the reserve managers.

AlbertoCentonze: Another option I was thinking about is having the gate storage live in the spoke and delegatecall it since they're tightly coupled

src/spoke/PermissionedSpoke.sol
```diff
@@ -0,0 +1,32 @@
...
+
+  /// @dev The gate fully decides whether the call is allowed, based on the caller, the position
+  /// owner and the calldata. It can preserve the default authorization by calling back
+  /// `isPositionManager` and allowing `liquidationCall`.
+  function _isAllowed(
+    address caller,
+    address user,
+    bytes calldata data
+  ) internal view virtual override returns (bool) {
+    return ISpokeGate(GATE).isCallAllowed({caller: caller, onBehalfOf: user, data: data});
```
Kogaroshi, line 30: Do we want to consider a "per asset" filter here, to skip enforcing gate on some specific assets ? (such as USDC/GHO deposits in Horizon ?)
Or modify the current HorizonGate to skip reserve managers checks on those assetIds, but that might require come callback or extra configuration

AlbertoCentonze: IMO gate should always be called, it's an immutable so the overhead isn't too big and we can have early return conditions (like asset to skip inside the gate logic). Otherwise the access control abstractions (posm + gate + restricted) get too messy

src/spoke/PermissionedSpoke.sol
```diff
@@ -0,0 +1,41 @@
...
+import {ISpokeGate} from 'src/spoke/interfaces/ISpokeGate.sol';
+
+/// @title PermissionedSpoke
+/// @author Aave Labs
+/// @notice Spoke where a gate replaces the default position manager authorization on position
+/// actions.
+abstract contract PermissionedSpoke is Spoke {
+  /// @notice The gate deciding whether position actions are allowed.
+  address public immutable GATE;
+
```
avniculae, line 14: eip712 name and version

src/spoke/Spoke.sol
```diff
@@ -413,7 +413,7 @@ abstract contract Spoke is
 
   /// @inheritdoc ISpoke
   function updateUserRiskPremium(address onBehalfOf) external nonReentrant {
-    if (!_isPositionManager({user: onBehalfOf, manager: msg.sender})) {
+    if (!_isAllowed(msg.sender, onBehalfOf, msg.data)) {
```
avniculae, line 416: if we choose to not change the bytecode, we are not going to be able to change behaviour in these 2 fns

src/horizon/HorizonSpokeGate.sol
```diff
@@ -0,0 +1,56 @@
...
+    bytes calldata data
+  ) external view returns (bool) {
+    bytes4 selector = bytes4(data);
+    if (selector == ISpoke.liquidationCall.selector) return true;
+    if (
+      (selector == ISpoke.supply.selector ||
+        selector == ISpoke.withdraw.selector ||
+        selector == ISpoke.repay.selector ||
+        selector == ISpoke.setUsingAsCollateral.selector) &&
+      isReserveManager({reserveId: uint256(bytes32(data[4:36])), manager: caller})
```
AlbertoCentonze, line 50: Isn't this too much power for the horizon usecase? To execute a forceful transfer we just need withdraw and repay no? Also does this mean that seizing collateral requires the reserve manager to repay? Is this how it works in horizon v3?

AlbertoCentonze:
Added coverage for the three previously missing gated operations and measured each policy on a fresh PermissionedSpoke deployment at b264e068.

| operation | direct Spoke | position-manager policy | global-manager policy | borrow-allowlist policy |
| --- | ---: | ---: | ---: | ---: |
| liquidationCall: full | 365,891 | 374,898 (+9,007 / +2.46%) | 374,986 (+9,095 / +2.49%) | 374,970 (+9,079 / +2.48%) |
| updateUserDynamicConfig: 1 collateral | 76,432 | 81,016 (+4,584 / +6.00%) | 81,107 (+4,675 / +6.12%) | 81,163 (+4,731 / +6.19%) |
| updateUserRiskPremium: 1 borrow | 158,839 | 164,698 (+5,859 / +3.69%) | 164,789 (+5,950 / +3.75%) | 164,845 (+6,006 / +3.78%) |

The scenarios mirror the existing direct-Spoke snapshots: full liquidation, one active collateral, and one active borrow. The delta is the gate callback plus the selected policy logic.

PR #1336, opened 2026-08-13
feat: add permissioned borrow access manager

avniculae:


src/access/PermissionedBorrowAccessManager.sol
avniculae, line 1: can you please add gas snapshots for using this access manager ? I suspect this adds ~5-10k gas costs

src/dependencies/openzeppelin/AccessManager.sol
```diff
@@ -155,6 +156,62 @@ contract AccessManager is Context, Multicall, IAccessManager {
     }
   }
 
+  /// @inheritdoc IAccessManager
+  function canCall(
```
avniculae, line 160: I personally don't like the complexity of mantaining a modified OZ Access Manager

src/spoke/Spoke.sol
```diff
@@ -226,7 +220,7 @@ abstract contract Spoke is
     uint256 reserveId,
     uint256 amount,
     address onBehalfOf
-  ) external nonReentrant onlyPositionManager(onBehalfOf) returns (uint256, uint256) {
+  ) external nonReentrant restricted returns (uint256, uint256) {
```
avniculae, line 223: one advantage of this is that we can reuse the same spoke impl for canonical and permissioned/horizon spokes

src/dependencies/openzeppelin/AccessManager.sol
miguelmtzinf: `canCall` is virtual so we can just add custom logic in `PermissionedBorrowAccessManager` without touching OZ deps no? @AlbertoCentonze 
smt like
```
function canCall() override {
  if target and selector is spoke and borrow(), then call custom logic
  else super.canCall()
}
```

miguelmtzinf: we can keep the `onlyPositionManager` as is, and only add `restricted`

src/spoke/Spoke.sol
miguelmtzinf: this is because we are moving the `onlyPositionManager` to the access manager. If we don't do it, we may need a seaprate implementation anyway for collateral seizing

src/dependencies/openzeppelin/AccessManager.sol
AlbertoCentonze: We need a method overload that uses `bytes` instead `bytes4` but yes technical we can just build on top. It still a custom implementation technically

src/spoke/Spoke.sol
miguelmtzinf, line 100: upside is that by just updating the `authority` via `setAuthority` we can modify general access control
no upgrade required, no separate implemenation for restricting access

PR #1337, opened 2026-08-14
feat: permissioned spoke via immutable gate

avniculae:
Variant of #1334 where the gate is immutable: a constructor argument instead of storage set at initialization. `_isAuthorizedPositionManagerCall` reads an immutable instead of a cold storage slot, and the instance has no storage, no `updateGate`, and no special compiler profile — it fits the standard 750 runs (24,543 bytes, vs 24,393 at 600 runs for #1334).

Gated actions get ~2.1k cheaper per action (supply with the borrow-allowlist policy is 132,289 vs 134,409, against 137,066 for the #1336 equivalent). The trade: replacing the gate contract requires an implementation upgrade instead of a config call — policy changes can still be mutable inside the gate itself, so this only bites when the gate's logic shape changes.

Gates rely on the spoke being their caller (`ISpoke(msg.sender)` callbacks), so the gate deploys first without knowing the spoke and the spoke takes it as a constructor argument.

One implementation can serve canonical and permissioned spokes alike: canonical spokes run a gate replicating the position-manager authorization (supply 132,022, +4.3k over the stock instance), permissioned ones run their policy. Gas snapshots for the three reference policies are regenerated in `PermissionedSpoke.Operations.json`.


avniculae:
Merged into #1334's branch — the gate design continues there.

PR #1340, opened 2026-08-22
test: cover liquidator health factor scenarios

Amlan2006:
Adds liquidation scenario coverage for liquidators with existing unhealthy positions.

- Covers a liquidator receiving collateral shares and improving their health factor
- Covers a competing liquidation where another liquidator liquidates the unhealthy liquidator first

Note: only the targeted scenario suite was run locally.

Testing:
```bash
forge test --match-path tests/contracts/spoke/liquidation/Spoke.LiquidationCall.Scenarios.t.sol


Amlan2006:
@avniculae @Kogaroshi 

The failing workflow appears to fail during Foundry setup before running the test suite:

`/home/runner/.foundry/bin/foundryup: cannot execute binary file`

The targeted suite passes locally:

```bash
forge test --match-path tests/contracts/spoke/liquidation/Spoke.LiquidationCall.Scenarios.t.sol

PR #1345, opened 2026-08-27
feat: Add BabylonSpoke

avniculae:
Adds a `BabylonSpoke` variant for the Babylon integration, with vaultBTC as collateral.

Liquidations are restricted to an immutable liquidation manager and sized by a collateral removal cap instead of the target health factor. The call repays up to `debtToCover` in the user's single debt reserve, premium first and without dust validation, removes collateral priced by the canonical bonus formula, and returns the applied bonus and the collateral removed. When the priced removal exceeds the cap, the repayment is resized to exactly consume it. The cap is given in asset units and converted into shares rounding down, so the payout never exceeds it at any supply share price. The z3 scripts in `tests/misc/z3/` prove the sizing and the rounding. No liquidation fee is charged and the liquidator always receives underlying assets.

Users hold at most one collateral and one debt reserve. Only the managed collateral reserve can be registered as collateral, no reserve can charge a liquidation fee, and the managed collateral reserve can never be borrowable, checked on listing, on update and on upgrade.

The liquidation manager and the managed collateral reserve are constructor immutables, taken by the deployment engine per Babylon spoke. The spoke carries the canonical spoke roles and the config engine is unchanged. Rotating the manager means a new implementation.

`BabylonLiquidationLogic` reuses the canonical validation, deficit evaluation and collateral and debt liquidation. Canonical `Spoke` only gains `virtual` markers and `public` visibility on `addReserve`, `updateReserveConfig` and `setUsingAsCollateral`, leaving `SpokeInstance` bytecode identical to main. `BabylonSpokeInstance` compiles at via-ir/450 with 180 bytes of EIP-170 margin.

Tests mirror the canonical suites: a library suite for the Babylon-specific functions, the integration engine on a fourth spoke deployed through the engine, the fuzz variants and scenarios on the managed collateral, config and upgrade tests, and gas snapshots. The shared deployment checks (implementation, oracle, authority, roles, proxy admin, bytecode and the two immutables) and the JSON report round trip now cover Babylon spokes, and the default orchestration fixture deploys one.

Unrelated to Babylon, one commit tightens the canonical small-liquidation-bonus fuzz bound from `MAX_LIQUIDATION_BONUS` to `MIN_LIQUIDATION_BONUS`, which is what the setup draws from; the old bound was vacuous.


src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,138 @@
...
+
+    BabylonSpokeStorage storage $ = _getBabylonSpokeStorage();
+    $.liquidationManager = liquidationManager;
+    $.managedCollateralReserveId = managedCollateralReserveId.toUint96();
+    emit UpdateBabylonLiquidationConfig(liquidationManager, managedCollateralReserveId);
+  }
+
+  /// @dev Restricts users to at most one registered collateral.
+  // todo: the registered collateral is not validated to be the managed collateral reserve, so a
+  // user whose single collateral is another reserve is not reachable by `liquidationCall` unless
```
avniculae: Known gap, deliberately accepted for now — there's an inline todo at the override to collect opinions. It's only reachable if governance gives another reserve a nonzero collateral factor: with CF 0 the registered collateral grants no borrowing power, so no unliquidatable debt can form. Open to enforcing `managedCollateralReserveId` in the hook instead if reviewers prefer.

src/spoke/Spoke.sol
```diff
@@ -350,7 +350,7 @@ abstract contract Spoke is
     address user,
     uint256 debtToCover,
     bool receiveShares
-  ) external nonReentrant {
+  ) external virtual nonReentrant {
```
avniculae: We're marking `virtual` only on the functions BabylonSpoke overrides (this one and `_validateSetUsingAsCollateral`), which is not consistent from the Spoke perspective. My preference would be to take the opportunity and mark all relevant functions virtual — opinions?

avniculae, line 353: We're marking `virtual` only on the functions BabylonSpoke overrides (this one and `_validateSetUsingAsCollateral`), which is not consistent from the Spoke perspective. My preference would be to take the opportunity and mark all relevant functions virtual — opinions?

src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /// @inheritdoc IBabylonSpoke
+  function getBabylonLiquidationConfig() external view returns (address, uint256) {
+    BabylonSpokeStorage storage babylonSpokeStorage = _getBabylonSpokeStorage();
+    return (babylonSpokeStorage.liquidationManager, babylonSpokeStorage.managedCollateralReserveId);
+  }
+
+  /// @dev Restricts users to at most one registered collateral.
+  function _validateSetUsingAsCollateral(
```
avniculae, line 124: This caps the count but deliberately does not check the reserve against `managedCollateralReserveId`: a user whose single registered collateral is another reserve is unreachable by `liquidationCall`, which is only safe while every other reserve keeps a zero collateral factor.

This hook does not receive the reserve id, so enforcing the id here would need a `reserveId` parameter added to the canonical hook. 

The alternative is overriding the external `setUsingAsCollateral` (made `public virtual` for `super`) — which leaves the canonical spoke bytecode the same

Question for reviewers: keep the count-only check, or enforce the managed reserve id via the external override? The override is proven bytecode-free for canonical instances, so it's purely a question of whether we want the stricter invariant.

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,535 @@
...
+      params.collateralFactor > 0 &&
+        userPositionStatus.isUsingAsCollateral(params.collateralReserveId),
+      ISpoke.ReserveNotEnabledAsCollateral()
+    );
+
+    for (uint256 i = 0; i < params.debtReserveIds.length; ++i) {
+      uint256 debtReserveId = params.debtReserveIds[i];
+      require(params.debtToCoverAmounts[i] > 0, ISpoke.InvalidDebtToCover());
+      require(!reserves.get(debtReserveId).flags.paused(), ISpoke.ReservePaused());
+      // quadratic duplicate scan; the list is bounded by the user's borrowed reserves
```
avniculae, line 531: Duplicates are rejected with a quadratic scan: the list is bounded by the user's borrowed reserves, so N should be small, which is why I preferred this over a temporary bitmap (which needed either inline bit manipulation or new PositionStatusMap helpers). Flagging in case anyone prefers a different mechanism here.

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,535 @@
...
+    );
+
+    // the collateral pricing is constant across the debt reserves repaid
+    // the bonus is derived from the entry health factor and the collateral price is read once
+    LiquidateDebtReservesParams memory liquidateDebtReservesParams = LiquidateDebtReservesParams({
+      collateralHub: params.collateralHub,
+      collateralAssetId: params.collateralAssetId,
+      collateralAssetUnit: MathUtils.uncheckedExp(10, params.collateralAssetDecimals),
+      collateralAssetPrice: IAaveOracle(params.oracle).getReservePrice(params.collateralReserveId),
+      liquidationBonus: LiquidationLogic.calculateLiquidationBonus({
```
avniculae, line 220: We compute lb only once, using the entry health factor, instead of recomputing HF after each debt reserve liquidated and recomputing lb. This should be fine, as lb will be the same or higher in this scenario, which means that we do not incentivise liquidaton splitting.

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,535 @@
...
+    (uint256 collateralSharesLiquidated, uint256 collateralAmountRemoved) = _liquidateDebtReserves({
+      reserves: reserves,
+      userPositions: userPositions,
+      collateralUserPosition: collateralUserPosition,
+      collateralLiquidatorPosition: userPositions[params.liquidator][params.collateralReserveId],
+      userPositionStatus: userPositionStatus,
+      params: liquidateDebtReservesParams
+    });
+
+    emit IBabylonSpoke.BabylonLiquidationCallSummary({
```
avniculae, line 244: what do you think about the current event structure? one summary event + one event per each debt reserve.

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,535 @@
...
+      user: params.user,
+      liquidator: params.liquidator,
+      collateralAmountRemoved: collateralAmountRemoved,
+      collateralSharesLiquidated: collateralSharesLiquidated
+    });
+
+    return
+      collateralUserPosition.suppliedShares == 0 &&
```
avniculae, line 252: we don't check activeCollateralCount bc we restrict collateral count to 1 already in the BabylonSpoke.

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,535 @@
...
+    LiquidationLogic.LiquidateCollateralResult memory liquidateCollateralResult = LiquidationLogic
+      ._liquidateCollateral(
+        collateralUserPosition,
+        collateralLiquidatorPosition,
+        LiquidationLogic.LiquidateCollateralParams({
+          hub: params.collateralHub,
+          assetId: params.collateralAssetId,
+          sharesToLiquidate: liquidationAmounts.collateralSharesToLiquidate,
+          sharesToLiquidator: liquidationAmounts.collateralSharesToLiquidate,
+          liquidator: params.liquidator,
```
avniculae, line 377: this means that we execute multiple collateral transfers to the liquidator. I preferred this option over either:
1. passing address(this) here and then send the transfer at the end
2. deferring removing collateral towards the end and removing in one go bc then we would repay debt before removing collateral, which we know is problematic. 

we have a third approach, which is same as 2, but we also defer debt repayment such that we still remove coll before repaying debt, but that would increase complexity.

worth noting that 2 might not be problematic is the collateral has a share price of 1, but I don't want to make assumptions about the share price in this implementation (although we could ofc, if we want).

thoughts?

src/spoke/BabylonSpoke.sol
ngfam: Should a full reserve for loop using `positionStatus.isUsingAsCollateral()` be good for this?

The detailed change might be a function `_validateBabylonPositionStatus()` where all other reserves (`!= managedCollateralReserveId`) are enforced 0. It might create a gas cost overhead but should be more intuitive.

src/spoke/libraries/BabylonLiquidationLogic.sol
ngfam: From Babylon's point of view, if Aave does not enforce the price assumption, the adapter will have an extra layer of rounding to make sure that it receives enough collateral for a vault. 

This though should not be an issue and the rounding should return an output exactly the same as the input, but seems to be addressed a lot more easily on the Spoke side.

avniculae: We cannot enforce the supply share price assumption in the spoke, as the hub can let other spokes borrow the same asset in theory

tests/contracts/spoke/liquidation/Spoke.BabylonLiquidationCall.t.sol
avniculae, line 1: FYI Tests CI time goes up from 1h40 to 2h40

src/spoke/BabylonSpoke.sol
avniculae: Implemented: `BabylonSpoke` now overrides `setUsingAsCollateral` (made `public virtual` on the canonical Spoke, verified byte-identical) and requires `reserveId == managedCollateralReserveId` when enabling, with a dedicated `UnsupportedCollateralReserve` error. No loop needed since only the managed id can ever be registered.

The count check in `_validateSetUsingAsCollateral` stays on top: it still matters after a config change, when a user may hold the previously managed reserve while registering the new one.

src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,130 @@
...
+      _notifyRiskPremiumUpdate(user, newRiskPremium);
+    }
+  }
+
+  /// @inheritdoc IBabylonSpoke
+  function getBabylonLiquidationConfig() external view returns (address, uint256) {
+    BabylonSpokeStorage storage babylonSpokeStorage = _getBabylonSpokeStorage();
+    return (babylonSpokeStorage.liquidationManager, babylonSpokeStorage.managedCollateralReserveId);
+  }
+
```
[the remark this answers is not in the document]
avniculae: Accepted — fixed in the current revision. `BabylonSpoke` overrides `setUsingAsCollateral` (made `public virtual` on the canonical Spoke, verified byte-identical there) and requires `reserveId == managedCollateralReserveId` when enabling, reverting with `UnsupportedCollateralReserve`. The count check in `_validateSetUsingAsCollateral` stays underneath: it still matters after a config change, when a user may hold the previously managed reserve. Covered by unit tests for both errors and the reconfiguration path.

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: I think we can replace `collateralUserPosition.suppliedShares == 0` with `liquidateDebtReservesParams.suppliedShares == collateralSharesRemoved`

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,540 @@
...
+      uint256 maxRemovableShares = params
+        .collateralHub
+        .previewAddByAssets(
+          params.collateralAssetId,
+          params.maxCollateralToRemove - totalCollateralAmountRemoved
+        )
+        .min(params.suppliedShares - totalCollateralSharesRemoved);
+      // no further repayments once the removal cap is consumed; the first repayment always runs,
+      // so debt can be liquidated even when the corresponding collateral amount to receive is zero
+      if (maxRemovableShares == 0 && i > 0) break;
```
avniculae, line 290: wdyt about this condition?

src/deployments/orchestration/AaveV4DeployOrchestration.sol
```diff
@@ -124,14 +136,21 @@ library AaveV4DeployOrchestration {
           hubConfiguratorAdmin: deployInputs.hubConfiguratorAdmin
         });
       }
-      if (deployInputs.spokeLabels.length > 0) {
+      if (deployInputs.spokeLabels.length > 0 || deployInputs.babylonSpokeLabels.length > 0) {
```
DhairyaSethi, line 139: is this or needed?

src/spoke/interfaces/IBabylonSpoke.sol
```diff
@@ -0,0 +1,109 @@
...
+  /// @param debtReserveId The identifier of the repaid debt reserve.
+  /// @param user The address of the borrower getting liquidated.
+  /// @param liquidator The address of the liquidator.
+  /// @param debtAmountRestored The amount of debt restored, expressed in asset units.
+  /// @param drawnSharesLiquidated The amount of drawn shares liquidated.
+  /// @param premiumDelta A struct representing the changes to premium debt after liquidation.
+  /// @param collateralAmountRemoved The amount of collateral removed, expressed in asset units.
+  /// @param collateralSharesLiquidated The amount of collateral shares liquidated.
+  event BabylonLiquidationCall(
+    uint256 indexed debtReserveId,
```
DhairyaSethi, line 37: still fine i think to emit managedCollReserveId

src/spoke/libraries/BabylonLiquidationLogic.sol
DhairyaSethi: yea and also bc we might be performing a liquidation when position has become healthy due to previous liqs

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,540 @@
...
+    emit IBabylonSpoke.BabylonLiquidationCallSummary({
+      user: params.user,
+      liquidator: params.liquidator,
+      collateralAmountRemoved: collateralAmountRemoved,
+      collateralSharesLiquidated: collateralSharesRemoved
+    });
+
+    return
+      collateralUserPosition.suppliedShares == 0 &&
+      userPositionStatus.nextBorrowing(params.reserveCount) != PositionStatusMap.NOT_FOUND;
```
DhairyaSethi, line 253: just borrowCount > 1 is cheaper i think

src/spoke/libraries/BabylonLiquidationLogic.sol
DhairyaSethi: imo == 0 is also fine

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,540 @@
...
+      // rounded down so the removed collateral cannot exceed the remaining cap
+      uint256 maxRemovableShares = params
+        .collateralHub
+        .previewAddByAssets(
+          params.collateralAssetId,
+          params.maxCollateralToRemove - totalCollateralAmountRemoved
+        )
+        .min(params.suppliedShares - totalCollateralSharesRemoved);
+      // no further repayments once the removal cap is consumed; the first repayment always runs,
+      // so debt can be liquidated even when the corresponding collateral amount to receive is zero
```
DhairyaSethi, line 289: why do we want to allow the first repayment again sorry?

src/spoke/libraries/BabylonLiquidationLogic.sol
DhairyaSethi: agree, still believe it's safe to assume for this collateral (it's only on the hub bc we didnt want to build a custom hub for this collateral) but regardless the current option of multiple transfer seems fine to me (we're non reenetrent so there's no execution context handover and the underlying token does not have extra weird rounding on each transfer for allowance or balance)

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,540 @@
...
+      ._liquidateCollateral(
+        collateralUserPosition,
+        collateralLiquidatorPosition,
+        LiquidationLogic.LiquidateCollateralParams({
+          hub: params.collateralHub,
+          assetId: params.collateralAssetId,
+          sharesToLiquidate: liquidationAmounts.collateralSharesToLiquidate,
+          sharesToLiquidator: liquidationAmounts.collateralSharesToLiquidate,
+          liquidator: params.liquidator,
+          receiveShares: false
```
DhairyaSethi, line 378: can you remind me why pls

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,540 @@
...
+  /// resized to exactly consume it, mirroring the canonical full-collateral sizing.
+  /// @param params The calculate liquidation amounts params.
+  /// @return The liquidation amounts.
+  function _calculateLiquidationAmounts(
+    CalculateLiquidationAmountsParams memory params
+  ) internal view returns (LiquidationAmounts memory) {
+    // premium debt is liquidated first, up to `debtToCover`
+    uint256 premiumDebtRayToLiquidate = params.premiumDebtRay;
+    // strict inequality is mandatory given rounding
+    if (params.debtToCover < premiumDebtRayToLiquidate.fromRayUp()) {
```
DhairyaSethi, line 428: dont we lose out on a dust premium case (premium < 1 ray anyway is fine to discard?)

src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,145 @@
...
+import {IBabylonSpoke} from 'src/spoke/interfaces/IBabylonSpoke.sol';
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+import {Spoke} from 'src/spoke/Spoke.sol';
+
+/// @title BabylonSpoke
+/// @author Aave Labs
+/// @notice Spoke variant for the Babylon integration: liquidations are restricted to a configured
+/// liquidation manager and sized by a collateral cap instead of a target health factor. Users can
+/// register at most one reserve as collateral.
+abstract contract BabylonSpoke is IBabylonSpoke, Spoke {
```
DhairyaSethi, line 19: should we override _validate config methods to ensure liquidationFee is zero and no additional coll

src/spoke/Spoke.sol
DhairyaSethi: im ok with all virtual tbh

DhairyaSethi: `public virtual`

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: > also bc we might be performing a liquidation when position has become healthy due to previous liqs

yes although in this case you could update the lb calculation fn to return min lb is healthy

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: for the same reason we allow no collateral to be liquidated while debt is being repaid. in an edge case scenario, the supply share price can be so high that one share is worth more than the entire debt. if we don't allow first repayment here, we would basically disallow 0 collateral liquidations

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: if we receive shares, anyone can bypass the liquidation manager (ie the adapter) and liquidate collateral up to any granularity

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: I don't think we lose any case here, this is btw the same code as the canonical spoke. please think through it again and make sure it's correct, will do the same

src/deployments/orchestration/AaveV4DeployOrchestration.sol
avniculae: I think it is bc the roles for the standard spoke fns (that are inherited by Babylon spoke) should still be configured, wdyt?

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: if you call borrowCount() on the map it's more expensive right?
if we use borrow count from previous user account data, is a bit tricky since that is not updated and there can be edge cases. We can also track and return info from liquidateDebtReserves fn, but I figured we may be fine to accept this gas cost increase here to lower the code complexity.

wdyt?

src/spoke/Spoke.sol
avniculae: Kept these `external virtual` rather than `public virtual`: an override can still widen to `public` where a variant needs an internal caller, and Spoke's existing rule of `public` only where there's an internal caller (`setUsingAsCollateral`) still holds. Measured both, identical bytecode. Same treatment for the `_validate*` internals, which the fee guard needed anyway. Also flipped the disabled canonical `liquidationCall` override in `BabylonSpoke` to `external`; it was `public` for no reason. b27e2186

src/spoke/interfaces/IBabylonSpoke.sol
avniculae: Added `collateralReserveId` as an indexed topic on `BabylonLiquidationCallSummary` rather than on the per-repayment event: a ninth argument next to the `PremiumDelta` struct pushes the ABI encoder over the stack limit, and the collateral reserve is constant across one call anyway, so once per liquidation carries the same information. b27e2186

src/spoke/BabylonSpoke.sol
avniculae: Done in b27e2186: `BabylonSpoke` overrides `_validateDynamicReserveConfig` and rejects a non-zero `liquidationFee` on every reserve, since only the managed reserve can ever be collateral. Covered on all three config paths. The "no additional coll" half was already enforced by the `setUsingAsCollateral` override, so nothing added there. The test fixtures were configuring fees the Babylon path never charged, which is exactly what this catches.

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: No dust case is lost. `debtToCover` is validated non-zero, so it's at least one asset unit, and a sub-unit premium has `fromRayUp() == 1 <= debtToCover`, which takes the full-liquidation branch. The strict `<` is required: at `debtToCover == fromRayUp(premium)`, `<=` would set `premiumToLiquidate = debtToCover.toRay() = ceil(premium) * RAY >= premium`, overstating the premium cleared. Same as canonical.

src/spoke/interfaces/IBabylonSpoke.sol
avniculae: Superseded in 2ff7d85f: with one debt reserve per user there is a single `BabylonLiquidationCall` event, and `collateralReserveId` is an indexed topic on it. The summary event is gone.

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: No longer applies after 2ff7d85f: users hold at most one debt reserve now (`MAX_USER_RESERVES_LIMIT` fixed at 1), so there is no skip path anymore. A debt reserve the user has not borrowed reverts with `ReserveNotBorrowed` through the canonical `_validateLiquidationCall`, and deficit detection reuses `LiquidationLogic._evaluateDeficit` with the counts from the pre-liquidation account data, same as the canonical spoke.

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,353 @@
...
+          underlying: params.debtUnderlying,
+          reserveId: params.debtReserveId,
+          drawnSharesToLiquidate: liquidationAmounts.drawnSharesToLiquidate,
+          premiumDebtRayToLiquidate: liquidationAmounts.premiumDebtRayToLiquidate,
+          drawnIndex: debtComponents.drawnIndex,
+          liquidator: params.liquidator
+        })
+      );
+
+    emit IBabylonSpoke.BabylonLiquidationCall({
```
avniculae, line 246: Should this emit the canonical `ISpoke.LiquidationCall` instead of a Babylon-specific event, so existing indexers pick up these liquidations without a second decoder? The extra canonical fields are cheap to fill here (`receiveShares` is always false, `collateralSharesToLiquidator` equals `collateralSharesLiquidated`).

Probably not very relevant in practice though: anyone reconstructing a Babylon liquidation has to decode the adapter's events anyway to know which vault was seized and what was paid. Leaning towards leaving it as is, but worth a second opinion.

src/spoke/interfaces/IBabylonSpoke.sol
```diff
@@ -85,27 +67,32 @@ interface IBabylonSpoke is ISpoke {
...
-    uint256[] calldata debtReserveIds,
-    uint256[] calldata debtToCoverAmounts,
+    uint256 debtReserveId,
+    uint256 debtToCover,
     address user,
     uint256 maxCollateralToRemove
-  ) external;
+  )
+    external
+    returns (
```
avniculae, line 91: I have decided to return more information from the liquidation call to help the adapter skip a few duplicate operations. Any thoughts?

src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,152 @@
...
+  /// @dev Constructor.
+  /// @param oracle_ The address of the oracle.
+  constructor(address oracle_) Spoke(oracle_, USER_RESERVES_LIMIT) {}
+
+  /// @inheritdoc IBabylonSpoke
+  function updateBabylonLiquidationConfig(
+    address liquidationManager,
+    uint256 managedCollateralReserveId
+  ) external restricted {
+    _reserves.get(managedCollateralReserveId);
```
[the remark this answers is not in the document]
avniculae: I'd say maybe we want to keep the option to disable this, in extreme cases. wdyt?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -218,6 +218,17 @@ interface IAaveV4ConfigEngine {
     uint256 liquidationBonusFactor;
   }
 
+  /// @notice Parameters for updating the Babylon liquidation config on a BabylonSpoke.
+  /// @dev The update is performed directly on the spoke, not through a SpokeConfigurator.
+  /// @dev spoke The address of the BabylonSpoke.
+  /// @dev liquidationManager The only address allowed to perform liquidations on the Spoke.
+  /// @dev managedCollateralReserveId The identifier of the only reserve usable as collateral.
+  struct BabylonLiquidationConfigUpdate {
+    address spoke;
```
avniculae: should we name it babylonSpoke? it's already in a struct that is babylon specific, so maybe it's fine as is. thoughts?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -218,6 +218,17 @@ interface IAaveV4ConfigEngine {
     uint256 liquidationBonusFactor;
   }
 
+  /// @notice Parameters for updating the Babylon liquidation config on a BabylonSpoke.
+  /// @dev The update is performed directly on the spoke, not through a SpokeConfigurator.
```
avniculae: I have decided to not add support in the configurator as this would require us deploying and maintaining more custom contracts. thoughts?

src/deployments/utils/libraries/Roles.sol
```diff
@@ -56,6 +57,7 @@ library Roles {
   uint64 public constant SPOKE_DOMAIN_ADMIN_ROLE = 300;
   uint64 public constant SPOKE_CONFIGURATOR_ROLE = 301;
   uint64 public constant SPOKE_USER_POSITION_UPDATER_ROLE = 302;
+  uint64 public constant BABYLON_SPOKE_CONFIGURATOR_ROLE = 303;
```
avniculae: should we use some different namespace here? I think we should, but keen to hear suggestions/opinions

src/deployments/utils/MetadataLogger.sol
```diff
@@ -67,6 +67,31 @@ contract MetadataLogger is Logger {
...
+        value: report.babylonSpokeInstanceBatchReports[i].report.spokeImplementation
+      });
+      babylonOracleEntries[i] = Logger.AddressEntry({
+        label: report.babylonSpokeInstanceBatchReports[i].label,
+        value: report.babylonSpokeInstanceBatchReports[i].report.aaveOracle
+      });
+    }
+    _writeGroup('babylonSpoke', babylonSpokeEntries);
+    _writeGroup('babylonSpokeImplementation', babylonSpokeImplEntries);
+    _writeGroup('babylonOracle', babylonOracleEntries);
```
avniculae, line 93: does this sound weird?

src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,174 @@
...
+
+  /// @dev The managed collateral reserve is never borrowable: a user holds a single debt reserve,
+  /// which can never be the collateral being seized. A reserve being listed is skipped, since the
+  /// managed collateral reserve is always already listed.
+  function _validateReserveConfig(
+    uint256 reserveId,
+    ReserveConfig calldata config
+  ) internal view virtual override {
+    super._validateReserveConfig(reserveId, config);
+    require(
```
avniculae, line 144: wdyt of this? should we not bother to enforce this, or maybe it's fine as is?

src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,174 @@
...
+    uint256 managedCollateralReserveId
+  ) external restricted {
+    require(
+      !_reserves.get(managedCollateralReserveId).flags.borrowable(),
+      UnsupportedBorrowableCollateral()
+    );
+
+    BabylonSpokeStorage storage babylonSpokeStorage = _getBabylonSpokeStorage();
+    babylonSpokeStorage.liquidationManager = liquidationManager;
+    babylonSpokeStorage.managedCollateralReserveId = managedCollateralReserveId.toUint96();
```
avniculae: I don't like that, by default, managedCollateralReserveId is 0, and hence valid. I did not find any issues with it, but wdyt?

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
```diff
@@ -392,6 +403,12 @@ interface IAaveV4ConfigEngine {
     LiquidationConfigUpdate[] calldata updates
   ) external;
 
+  /// @notice Executes Babylon liquidation config updates on BabylonSpokes.
+  /// @param updates The Babylon liquidation config updates to execute.
+  function executeBabylonLiquidationConfigUpdates(
```
avniculae: @montyly made a good point that is worth discussing: do we want separate config engines instead? or should we keep code as is here, and reuse the same config engine?

src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,174 @@
...
+import {IBabylonSpoke} from 'src/spoke/interfaces/IBabylonSpoke.sol';
+import {ISpoke} from 'src/spoke/interfaces/ISpoke.sol';
+import {Spoke} from 'src/spoke/Spoke.sol';
+
+/// @title BabylonSpoke
+/// @author Aave Labs
+/// @notice Spoke variant for the Babylon integration: liquidations are restricted to a configured
+/// liquidation manager and sized by a collateral cap instead of a target health factor. Users hold
+/// at most one collateral and one debt reserve, the managed collateral reserve is never borrowable
+/// and reserves cannot charge a liquidation fee.
```
montyly, line 16: Maybe we should document here all the edge cases that have to be prevented due to the integration. On top of my head:

- liquidationFee is zero
- managedCollateralReserveId should not be borrowable
- receiveShares should not be enabled
- CF > 0

Having it in one place will make it easier to maintain over time imho

src/spoke/BabylonSpoke.sol
```diff
@@ -0,0 +1,174 @@
...
+
+  /// @dev Users hold at most one collateral and one debt reserve.
+  uint16 internal constant USER_RESERVES_LIMIT = 1;
+
+  /// @dev Constructor.
+  /// @param oracle_ The address of the oracle.
+  constructor(address oracle_) Spoke(oracle_, USER_RESERVES_LIMIT) {}
+
+  /// @inheritdoc IBabylonSpoke
+  function updateBabylonLiquidationConfig(
```
montyly: I am a bit unsure about allowing to change managedCollateralReserveId on the fly

What if the reserve is borrowable, someone borrow, then we update the reserve with updateReserveConfig to make it not borrowable anymore. and then we can use this reserve ID as the managed collateral one? 

I think if we have to change the reserve ID currently managed, its probably a significant event, that should either go through a new redeployment, or an upgrade. 

wdyt?

avniculae: fair point. how would you do it though? would you make the managed reserve id 0 by default? the cold start is a bit problematic here

montyly: If the goal is to allow init `managedCollateralReserveId` to a non-zero value, maybe we make it a separate function, that can only be called once? 

Ex:
- `initBabylonLiquidationConfig(liquidationManager, managedCollateralReserveId)` (check that liquidation manager is not set yet) 
- `updateBabylonLiquidationConfig(liquidationManager)` (only if we want to keep the ability to change the liquidation manager later on)

We also need to prevent updating liquidation manager to zero to prevent re-init in both functions

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
DhairyaSethi: no strong opinion, fine as is imo

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
DhairyaSethi: i think we should, this require a custom configurator which can be inherited here

src/deployments/utils/libraries/Roles.sol
DhairyaSethi: totally different namespace please, definitely not this. we need this expand regular spoke ops and babylon split is separate

src/spoke/BabylonSpoke.sol
DhairyaSethi: good to do imo

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
DhairyaSethi: i think it's purely additive and non blocking so fine to keep here, or go the normal route like configurator above and have separate, either works bc diff is small

snapshots/Spoke.Operations.json
```diff
@@ -8,7 +8,7 @@
   "liquidationCall: partial": "360696",
   "permitReserve + repay (multicall)": "166334",
   "permitReserve + supply (multicall)": "151663",
-  "permitReserve + supply + enable collateral (multicall)": "166090",
+  "permitReserve + supply + enable collateral (multicall)": "166114",
```
DhairyaSethi: these are just changing due to forge version change right

src/spoke/BabylonSpoke.sol
avniculae: The managed reserve is an immutable now and `setUsingAsCollateral` compares against it directly. The count-hook override is gone. c7d44746

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: Single debt reserve per call since 2ff7d85f, there is no list to scan.

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: One debt reserve per call, so one bonus by construction. 2ff7d85f

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: Collapsed into the single `BabylonLiquidationCall` event. 2ff7d85f

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: Deficit detection reuses the canonical `_evaluateDeficit`. 2ff7d85f

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: The loop is gone with the single debt reserve. 2ff7d85f

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: The loop is gone. The behaviour it protected, repaying debt when the priced removal floors to zero shares, is documented on `_executeLiquidation` and pinned by `test_calculateLiquidationAmounts_ZeroCollateral` and `test_executeLiquidation_ZeroCollateralRemoved`.

src/spoke/libraries/BabylonLiquidationLogic.sol
avniculae: One debt reserve means one `remove` per call, so the multiple-transfer question no longer arises. 2ff7d85f

src/spoke/interfaces/IBabylonSpoke.sol
avniculae: Settled on returning the bonus and the collateral removed, the account data return was dropped. 45318b3b

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
avniculae: The struct is gone, the engine is back to identical with main. c7d44746

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
avniculae: There is no runtime config left to route through a configurator, both values are constructor immutables now. c7d44746

src/deployments/utils/libraries/Roles.sol
avniculae: Role 303 is deleted. The Babylon spoke carries the canonical spoke roles. c7d44746

src/config-engine/interfaces/IAaveV4ConfigEngine.sol
avniculae: The engine is unchanged against main. c7d44746

src/spoke/BabylonSpoke.sol
avniculae: The id is a constructor immutable now, so there is no default that happens to be valid. It can still name a reserve that is not listed yet on a fresh deployment, which is why the borrowable check runs when the reserve gets listed, on every config update, and from `initialize` on upgrade, reading the stored flags rather than requiring the reserve to exist. c7d44746

src/spoke/BabylonSpoke.sol
avniculae: Went the way you suggested: both values are immutables, so a change is a new implementation. Listing, updating and upgrading all reject a borrowable managed reserve. c7d44746

snapshots/Spoke.Operations.json
avniculae: No. They came from the `_validateReserveConfig` signature change, which is reverted, and the spoke snapshots match main again. The one remaining hub line is a 1-gas artifact of the shared test fixture, hub and token bytecode are identical to main. c7d44746

tests/contracts/spoke/liquidation/BabylonSpoke.LiquidationCall.t.sol
avniculae: it's now down to 2h

src/spoke/BabylonSpoke.sol
avniculae: Done in 4b62425c. The `BabylonSpoke` header now lists the integration invariants in one place: one collateral and one debt reserve per user, only the managed collateral reserve enabled as collateral and never borrowable, no liquidation fee, and collateral always received in underlying assets. The collateral factor bound is the canonical liquidation validation's, so it stays documented there.

snapshots/Hub.Operations.json
```diff
@@ -11,7 +11,7 @@
   "remove: partial": "85702",
   "reportDeficit": "116908",
   "restore: full": "81488",
-  "restore: full - with transfer": "188919",
+  "restore: full - with transfer": "188918",
```
avniculae, line 14: 1 gas lower than main while the hub bytecode is byte-identical. This is the second snapshot window in `test_restore_with_transfer`, and forge also counts the test-frame gas spent since the previous window, so the value moves with the test contract bytecode (here one JUMPDEST fewer in a helper, because `Base` inherits `BatchTestProcedures`). Reported in foundry-rs/foundry#16941.

src/spoke/Spoke.sol
avniculae: Done in ff6c6591: every state-changing entry point in Spoke is virtual now, not only the ones BabylonSpoke overrides. SpokeInstance bytecode is unchanged.

src/spoke/BabylonSpoke.sol
avniculae: Superseded. The override this thread is on is gone; the check now lives in _validateManagedCollateralReserve, called after super from addReserve and updateReserveConfig and from initialize, so the managed collateral reserve can never be borrowable on listing, update or upgrade.

src/spoke/libraries/BabylonLiquidationLogic.sol
```diff
@@ -0,0 +1,361 @@
...
+          .collateralHub
+          .previewAddByAssets(params.collateralAssetId, params.maxCollateralToRemove)
+          .min(collateralUserPosition.suppliedShares)
+      })
+    );
+
+    // the liquidation fee is not charged: the liquidator receives the full removed collateral
+    LiquidationLogic.LiquidateCollateralResult memory liquidateCollateralResult = LiquidationLogic
+      ._liquidateCollateral(
+        collateralUserPosition,
```
avniculae, line 226: I didn't use named args here to be in line with canonical liquidation logic, although I would like to use since we have 2 consecutive params with same type, wdyt we should do? @DhairyaSethi

src/spoke/instances/BabylonSpokeInstance.sol
```diff
@@ -0,0 +1,41 @@
...
+    uint256 managedCollateralReserveId_
+  ) BabylonSpoke(oracle_, liquidationManager_, managedCollateralReserveId_) {
+    _disableInitializers();
+  }
+
+  /// @notice Initializer.
+  /// @dev The authority contract must implement the `AccessManaged` interface for access control.
+  /// @dev Reverts with `UnsupportedBorrowableCollateral` if the managed collateral reserve is already listed as borrowable.
+  /// @param authority The address of the authority contract which manages permissions.
+  function initialize(address authority) external override reinitializer(SPOKE_REVISION) {
```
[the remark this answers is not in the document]
avniculae: I am fine accepting this risk.

src/spoke/Spoke.sol
```diff
@@ -106,7 +106,7 @@ abstract contract Spoke is
   function initialize(address authority) external virtual;
 
   /// @inheritdoc ISpoke
-  function updateLiquidationConfig(LiquidationConfig calldata config) external restricted {
+  function updateLiquidationConfig(LiquidationConfig calldata config) external virtual restricted {
```
avniculae, line 109: Should all of these be `public virtual` rather than `external virtual`? @DhairyaSethi leaned that way earlier and I pushed back, but the Babylon overrides show the practical cost. A variant that wants to wrap an entry point has to call `super`, which is an internal call and only compiles when the base is `public`, which is exactly why `addReserve`, `updateReserveConfig` and `setUsingAsCollateral` changed in this PR. For every other entry point a variant has to duplicate the canonical body instead. Bytecode is identical either way, I measured it, so the only thing we would give up is the current convention that `public` in Spoke signals an internal caller.

DhairyaSethi: yes does spoke bytecode change with public for whatever reason?

avniculae: no, the consideration it's more around whether we want to allow internal calls to call those fns, since it can hide impl errors

PR #1349, opened 2026-08-28
docs: Add Whitepaper

CanonicalJP:


AlbertoCentonze (approved):
Why not call it simply whitepaper since this is the aave-v4 repo? Also just wanted to double check that we don't want to attach the raw latex file in case of edits or new features everything lives here.

PR #1353, opened 2026-09-01
test: cover same spoke add and draw accounting

Amlan2006:
Adds Hub draw coverage for the case where the same Spoke adds liquidity and then draws the same asset.

This test documents that Hub accounting keeps added shares and drawn shares as independent ledgers. Drawing liquidity does not auto-net against previously added liquidity from the same Spoke.

- Covers same-Spoke add/draw accounting
- Asserts added shares remain after draw
- Asserts drawn shares are created after draw
- Asserts Hub liquidity updates independently of share accounting

Testing:

```bash
forge test --match-path tests/contracts/hub/draw-restore/Hub.Draw.t.sol


Amlan2006:
@avniculae @DhairyaSethi

PR #1358, opened 2026-09-15
test: match exact SignatureGateway revert payloads

figtracer:
Prepare the SignatureGateway tests for Foundry v1.8.3 by removing trailing address arguments from three `expectRevert(bytes)` payloads. The exact raw-data matching introduced in https://github.com/foundry-rs/foundry/pull/16428 rejects these malformed expectations: `ERC20InsufficientAllowance` has three arguments and `SpokeNotRegistered` has none.


DhairyaSethi:
superseded by https://github.com/aave/aave-v4/pull/1360

Issue #853, opened 2025-10-03
Interest rate linearization from per-year to per-second allows users to inflate interest paid by borrowers


philatcertora:
## Description
The `AssetInterestRateStrategy.sol` stores interest rates on an annual basis.

When interest is accrued, the `MathUtils::calculateLinearInterest()` function calculates the interest generated over a time interval defined in seconds.

To align the units correctly (`per-year interest rate` vs `seconds passed`), the function converts the per-year interest rate into a per-second rate.

However, when a linearized interest rate is recompounded, the effective rate is potentially different:
- If the rate is recompounded **more frequently** than the original period, the effective rate is higher
- If the rate is recompounded at the same frequency as the original period, the effective rate is the same
- If the rate is recompounded **less frequently** than the original period, the effective rate is lower

The problem is: given interest accrual can be triggered permissionlessly by users, lenders can exploit this pattern to inflate the interest rate, causing borrowers to pay more interest than intended.

## Impact
Here is an example, assuming a maximum annual rate of 1,000% (please find all of the calculations [here](https://docs.google.com/spreadsheets/d/1ztOO1cMGcwD_1FXw1J_IHGoRkK5ypTDK-LXB_e_Ua5M/edit?usp=sharing))
- The linearized second rate is 3.17E-07 (spreadsheet ref 1)
- The recompounded annual rate is 2,202,543% (spreadsheet ref 2)

This would allow an attacker to inflate the annual interest rate from 1,000% to over 2 million percent, by forcing it to compound every second.

## Recommended mitigation
One way of mitigating this issue with minimal code changes is to first decompound the annual rate into a more granular rate, before linearizing it. I would suggest to use weekly rates.

This would reduce the impact as follows, assuming a maximum annual rate of 1,000%:
- The equivalent weekly rate is ~4.71% (spreadsheet ref 3)
- The linearized second rate for ~4.71% per week is 7.78E-08 (spreadsheet ref 4)
- The recompounded annual rate is 1,063% (spreadsheet ref 5)

As noted, the recommended mitigation is still exploitable, but the impact is limited to only a marginal increase in the resulting recompounded annual rate, from 1,000% to 1,063%.

This mitigation assumes that the asset will have at least one user interaction per week. This is because a downside of such mitigation is that, if the interest accrual is triggered less frequently than that, the resulting annual rate will be less than expected.

[...]

PR #1361, opened 2026-09-18
test(hub): reproduce and quantify accrual-frequency debt inflation from #853

Annkkitaaa:
## Summary

This does not fix anything. It adds a runnable Foundry reproduction for #853, confirming the issue is still present on current `main` and replacing the original spreadsheet estimate with numbers verified directly against the codebase.

Credit for the original finding goes to @philatcertora, and to @BhariGowda for pointing out that `AssetLogic.getDrawnIndex` is the debt-side function affected and that this mirrors a pattern Aave V2/V3 handled by using a separate compounded-interest formula for the variable debt index specifically. Both comments are on #853.

## What the test shows

`AssetLogic.getDrawnIndex` compounds `MathUtils.calculateLinearInterest` on every `accrue()` call. `accrue()` is reachable permissionlessly through ordinary Spoke actions (`add`, `draw`, `remove`, `restore`, ...), and `drawnIndex` is a single global value per asset, so anyone interacting with a live market can force extra accruals that affect every other borrower of that asset, not just themselves.

Because linear steps compounded repeatedly converge toward `e^x` while a single linear step only reaches `1 + x`, splitting a fixed elapsed period into more forced accruals strictly increases the resulting debt for the same nominal annual rate and the same wall-clock time. The new test (`tests/contracts/hub/misc/HubAccrueInterest.FrequencyInvariance.t.sol`) demonstrates this directly against the Hub rather than as a standalone spreadsheet:

- At the protocol's own configured rate ceiling used in tests (`MAX_ALLOWED_DRAWN_RATE`, 1000% APR), 500,000 DAI borrowed for one year accrues to 5,500,000 DAI with a single accrual, matching simple 1000% annual interest. With 180 forced accruals over that same year (about once every two days, ordinary gas cost, no per-block spamming needed), the same principal accrues to about 8.425 billion DAI, roughly 1,532x more debt than the single-accrual case for the identical elapsed time and nominal rate.
- At a more realistic 80% APR stress rate, the same 180 forced accruals produce about 1,110,800 DAI of debt against 900,000 DAI for a single accrual, about 23% more debt for no change in nominal rate or elapsed time.
- A fuzz test (`test_accrueInterest_fuzz_DebtIsMonotonicInForcedAccrualCount`) checks 1001 runs and confirms debt is monotonically non-decreasing in the number of forced accruals over a fixed period, for every split tried.

## On severity

The original report's headline figure (1000% nominal recompounding to roughly 2,202,543% if accrued every second for a full year) is the correct theoretical continuous-compounding limit, but it assumes per-second accrual sustained for an entire year, and `getDrawnIndex` already guards against updating more than once per block (`lastUpdateTimestamp == block.timestamp` short-circuits). A more grounded worst case is bounded by realistic attacker cadence and gas spend rather than the continuous limit.

What the test above shows is that this bound is not very reassuring in practice: at the protocol's own rate ceiling, a few hundred ordinary transactions spread over a year already realize about 76% of the theoretical continuous-compounding blow-up (16,850x principal reached, against a 22,026x asymptote), and at a realistic stress rate the effect is already close to its asymptotic value at the same modest cadence. No block-level spamming is required for the effect to be significant.

## Suggested direction (not implemented here)

Aave V2/V3 used `calculateLinearInterest` for the liquidity/supply index and a separate `calculateCompoundedInterest` (3-term binomial approximation of `e^(rate*t)`) specifically for the variable debt index, because debt accrual needs to be invariant to how often it is triggered. V4's Hub uses a single `drawnIndex` for both the debt side and (through `totalAddedAssets`) the value backing supplied shares, and there is no compounded-interest formula anywhere in the current codebase. Restoring a frequency-invariant formula for `getDrawnIndex` looks like the natural fix, but the exact approach (formula choice, whether the single-index design needs to change, how it interacts with the premium and deficit accounting) is a rate-model design decision that belongs to the team, so I did not attempt it here.

## Testing

- `forge test --mp 'tests/contracts/hub/misc/HubAccrueInterest.FrequencyInvariance.t.sol' -vv`: 3 passed, 0 failed (includes a 1001-run fuzz test).
- `forge test --mp 'tests/contracts/hub/**'`: 316 passed, 0 failed, no regressions in the existing Hub suite (including the pre-existing `HubAccrueInterest.t.sol`).
- No `src/` changes, so gas snapshots under `snapshots/` are unaffected.

Relates to #853.


BhariGowda:
Thanks for building this out properly  the monotonicity fuzz test and the realistic cadence framing (180 accruals over a year, no per-block spamming) make the severity much more concrete than the original spreadsheet estimate.

On the fix direction: the natural slot for a frequency-invariant formula is inside getDrawnIndex itself, replacing the calculateLinearInterest call with a binomial approximation of e^(rate * dt)  the same 3-term expansion Aave V2/V3 used for calculateCompoundedInterest in MathUtils. The key property is that e^(r*t1) * e^(r*t2) = e^(r*(t1+t2)), so compounding over two separate intervals gives exactly the same result as compounding over the combined interval, regardless of how many times accrue() is called. Linear interest doesn't have that property, which is exactly what your test demonstrates.

The harder question, as you note, is whether drawnIndex can safely diverge from the supply-side index without breaking the premium/deficit accounting. V2/V3 handled this by having two separate indices (liquidity index vs variable debt index), but V4's single-index design means that decision has wider surface area. Worth flagging explicitly for the team rather than assuming a drop-in formula swap is sufficient.

Annkkitaaa:
Thanks for the detailed breakdown, this is exactly the right formula direction.

To confirm the mechanism precisely: the property that makes this work is `e^(r*t1) * e^(r*t2) = e^(r*(t1+t2))`, so splitting a period into any number of accruals reproduces the same result as one accrual over the combined period. Linear interest has no equivalent identity, which is what the monotonicity fuzz test in this PR demonstrates empirically. The V2/V3 formula for `calculateCompoundedInterest` (a 3-term Taylor expansion of `e^x`, RAY-scaled) would carry over the same way here.

On the harder question, I traced the dependency and it goes further than premium and deficit accounting. `AssetLogic.totalAddedAssets()`, which determines what is owed to depositors, is computed from `drawnIndex` through `_calculateAggregatedOwedRay`. So `drawnIndex` isn't only the debt-side number, it also drives the supply-side accounting in this single-index design. Changing its growth formula would change what LPs are credited, not just what borrowers owe, and would also invalidate the expected-value calculations that the existing Hub test suite computes directly from `calculateLinearInterest`.

V2/V3 avoided this by keeping the debt index and the liquidity index separate. Given V4 merged them, I don't think a formula swap inside `getDrawnIndex` should be attempted as a community PR without the team confirming the intended design first, since the two paths are:

1. Keep the single-index model and verify a compounded formula is safe against premium, deficit, and added-assets accounting.
2. Reintroduce a second index for the debt side specifically, closer to the V2/V3 split.

Given this issue has been open since October 2025 with the regression risk now confirmed independently twice, would it make sense for a maintainer with authority over the Hub interest-rate design to weigh in on which direction is intended, before anyone invests time implementing either one?

cc @DhairyaSethi, since you've been active on the interest-rate and libraries side of the Hub.

